PR #6153, opened 2024-01-22
chore(admin-ui->orders): orders table pagination title use i18h

wangjue666:
image

The text in this place does not use **i18h**


olivermrbl:
@wangjue666, appreciate the contribution. However, our admin system is currently undergoing a redesign that includes improvements to and simplifying of translations. We expect to go live with the update next month. If you are still interested in contributing then, we would very much appreciate a new PR with these changes.

PR #6155, opened 2024-01-22
fix(medusa-js): correct invites resend path in js client

lukebui:


srindom (approved):
LGTM - thanks for the fix!

carlos-r-l-rodrigues (approved):

PR #6136, opened 2024-01-19
feat(cart): Item tax lines + shipping method tax lines

olivermrbl:

PR #6157, opened 2024-01-22
feat(cart): JoinerConfig

olivermrbl:


carlos-r-l-rodrigues (approved):
LGTM, but we are missing list methods for `LineItemTaxLines` and `ShippingMethodTaxLines` that is OK to have in another PR.

olivermrbl:
Ah yes nice catch @carlos-r-l-rodrigues. This PR is depending on #6136, which contains these methods 💪

PR #6159, opened 2024-01-22
feat(medusa,types): added buyget support for modules

riqwan:
what:
- adds buyget support for module create and update
- adds buy_rules and validations for application methods
- adds atomic operations for buy_rules

RESOLVES CORE-1697
RESOLVES CORE-1698

### DIFF
--- a/integration-tests/plugins/__tests__/promotion/admin/update-promotion.spec.ts
+++ b/integration-tests/plugins/__tests__/promotion/admin/update-promotion.spec.ts
@@ -132,4 +132,58 @@ describe("POST /admin/promotions/:id", () => {
       })
     )
   })
+
+  it("should update a buyget promotion successfully", async () => {
+    const createdPromotion = await promotionModuleService.create({
+      code: "PROMOTION_TEST",
+      type: PromotionType.BUYGET,
+      application_method: {
+        type: "fixed",
+        target_type: "items",
+        allocation: "across",
+        value: "100",
+        apply_to_quantity: 1,
+        buy_rules_min_quantity: 1,
+        buy_rules: [
+          {
+            attribute: "product_collection.id",
+            operator: "eq",
+            values: ["pcol_towel"],
+          },
+        ],
+        target_rules: [
+          {
+            attribute: "product.id",
+            operator: "eq",
+            values: "prod_mat",
+          },
+        ],
+      },
+    })
+
+    const api = useApi() as any
+    const response = await api.post(
+      `/admin/promotions/${createdPromotion.id}`,
+      {
+        code: "TEST_TWO",
+        application_method: {
+          value: "200",
+          buy_rules_min_quantity: 6,
+        },
+      },
+      adminHeaders
+    )
+
+    expect(response.status).toEqual(200)
+    expect(response.data.promotion).toEqual(
+      expect.objectContaining({
+        id: expect.any(String),
+        code: "TEST_TWO",
+        application_method: expect.objectContaining({
+          value: 200,
+          buy_rules_min_quantity: 6,
+        }),
+      })
+    )
+  })
 })
--- a/packages/medusa/src/api-v2/admin/promotions/validators.ts
+++ b/packages/medusa/src/api-v2/admin/promotions/validators.ts
@@ -58,8 +58,8 @@ export class AdminPostPromotionsReq {
 
   @IsNotEmpty()
   @ValidateNested()
-  @Type(() => ApplicationMethod)
-  application_method: ApplicationMethod
+  @Type(() => ApplicationMethodsPostReq)
+  application_method: ApplicationMethodsPostReq
 
   @IsOptional()
   @IsArray()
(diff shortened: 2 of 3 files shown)


packages/promotion/src/models/promotion-rule.ts
```diff
@@ -53,7 +53,13 @@ export default class PromotionRule {
...
     (applicationMethod) => applicationMethod.target_rules
   )
-  application_methods = new Collection<ApplicationMethod>(this)
+  method_target_rules = new Collection<ApplicationMethod>(this)
+
+  @ManyToMany(
+    () => ApplicationMethod,
+    (applicationMethod) => applicationMethod.buy_rules
+  )
+  method_buy_rules = new Collection<ApplicationMethod>(this)
```
olivermrbl, line 62: question: Could we omit the `method` prefix or do you think it makes them ambigious?

packages/promotion/src/services/promotion-module.ts
```diff
@@ -794,7 +823,7 @@ export default class PromotionModuleService<
   @InjectTransactionManager("baseRepository_")
   protected async createPromotionRulesAndValues_(
     rulesData: PromotionTypes.CreatePromotionRuleDTO[],
-    relationName: "promotions" | "application_methods",
+    relationName: "promotions" | "method_target_rules" | "method_buy_rules",
```
olivermrbl, line 868: suggestion: Goes with my other comment
```suggestion
    relationName: "promotions" | "target_rules" | "buy_rules",
```

packages/promotion/src/migrations/Migration20240122084316.ts
```diff
@@ -59,7 +59,11 @@ export class Migration20240122070028 extends Migration {
     )
 
     this.addSql(
-      'create table "application_method_promotion_rule" ("application_method_id" text not null, "promotion_rule_id" text not null, constraint "application_method_promotion_rule_pkey" primary key ("application_method_id", "promotion_rule_id"));'
+      'create table "application_method_target_rules" ("application_method_id" text not null, "promotion_rule_id" text not null, constraint "application_method_target_rules_pkey" primary key ("application_method_id", "promotion_rule_id"));'
```
carlos-r-l-rodrigues, line 62: Comment not specific to this PR, but for all the migrations of all modules.
We should add `if not exists` to create everything in the DB. Tables, indexes, constraints, add columns, etc.

packages/promotion/src/services/promotion-module.ts
```diff
@@ -952,9 +1022,10 @@ export default class PromotionModuleService<
     rulesData: PromotionTypes.RemovePromotionRuleDTO[],
     @MedusaContext() sharedContext: Context = {}
   ): Promise<PromotionTypes.PromotionDTO> {
-    await this.removePromotionTargetRules_(
+    await this.removeApplicationMethodTargetRules_(
       promotionId,
       rulesData,
+      "target_rules",
```
carlos-r-l-rodrigues: Probably a good ideia to have constants or a enum from these strings: "target_rules", "buy_rules"

packages/promotion/src/migrations/Migration20240122084316.ts
riqwan: Yes, good shout. I've added it to my pre-release tasks in linear. The reason i'm not doing it now is because we will still destroy existing migration files and create new ones a few more times to keep a singular migration file before release.

carlos-r-l-rodrigues (approved):

packages/promotion/src/models/application-method.ts
```diff
@@ -42,6 +44,12 @@ export default class ApplicationMethod {
   @Property({ columnType: "numeric", nullable: true, serializer: Number })
   max_quantity?: number | null
 
+  @Property({ columnType: "numeric", nullable: true, serializer: Number })
+  apply_to_quantity?: number | null
+
+  @Property({ columnType: "numeric", nullable: true, serializer: Number })
+  buy_rules_min_quantity?: number | null
```
olivermrbl: suggestion: Convert to
```suggestion
  @Property({ columnType: "numeric", nullable: true, serializer: Number })
  apply_to_quantity: number | null = null

  @Property({ columnType: "numeric", nullable: true, serializer: Number })
  buy_rules_min_quantity: number | null = null
```

olivermrbl: And rest of model

packages/promotion/src/models/promotion-rule.ts
riqwan: We could, but I added it there to make it more specific that it relations to the application method and not the rule itself. wdyt?

olivermrbl: I see. Let's keep it as is 👍

packages/promotion/src/models/application-method.ts
riqwan: changed in all models in the module.

riqwan: @olivermrbl also added some validations for the API in the new commit, take a look at those as well when you re-review.

olivermrbl (approved):

PR #6162, opened 2024-01-22
feat(dashboard): Pub Api Key domain

kasperkristensen:
**What**
- Adds Pub. Api Key domain

**Note**
- Pagination for sales channels associated with the pubkey is not implemented correctly as it is not supported by the API. Will open a separate issue on this, and revisit the impl. once fixed.

CLOSES CORE-1656

### DIFF
--- a/.changeset/young-forks-worry.md
+++ b/.changeset/young-forks-worry.md
@@ -1,0 +1,6 @@
+---
+"@medusajs/medusa": patch
+"@medusajs/types": patch
+---
+
+feat(medusa,types): added buyget support for modules
--- a/integration-tests/plugins/__tests__/promotion/admin/create-promotion.spec.ts
+++ b/integration-tests/plugins/__tests__/promotion/admin/create-promotion.spec.ts
@@ -62,7 +62,7 @@ describe("POST /admin/promotions", () => {
     )
   })
 
-  it("should create a promotion successfully", async () => {
+  it("should create a standard promotion successfully", async () => {
     const api = useApi() as any
     const response = await api.post(
       `/admin/promotions`,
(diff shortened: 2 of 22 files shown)


olivermrbl:
The modal does not close on submit:

https://github.com/medusajs/medusa/assets/59018053/ae588df9-6dc9-43eb-985a-6ee4b574778a

I rebuilt the `dashboard` project and tested again. Same result. Let me know if there is anything else I need to clear before trying again, if the issue is not present on your side.

olivermrbl:
Any particular reason for having a link to the user here: 

CleanShot 2024-01-22 at 19 59 19

To me it is a pretty uncommon use case to wanting to navigate from pub key -> user. If you agree, we could kill it

olivermrbl:
nitpicking(non-blocking): Should we remove "Revoke" when key is revoked?

CleanShot 2024-01-22 at 20 00 55

olivermrbl:
suggestion(non-blocking): Include "Created" in the details overview. If you have multiple active keys and you can't differentiate, the "Created" usually helps to give an indication of what is used and what is not. Ideally, we'd have a "Last used".

CleanShot 2024-01-22 at 20 02 25

packages/admin-next/dashboard/src/routes/api-key-management/api-key-management-detail/components/api-key-sales-channel-section/api-key-sales-channel-section.tsx
```diff
@@ -0,0 +1,370 @@
...
+}) => {
+  const { t } = useTranslation()
+  const prompt = usePrompt()
+
+  const { mutateAsync } = useAdminRemovePublishableKeySalesChannelsBatch(apiKey)
+
+  const handleDelete = async () => {
+    const res = await prompt({
+      title: t("general.areYouSure"),
+      description: t("locations.removeSalesChannelsWarning", { count: 1 }),
```
olivermrbl: todo: Message should be specific to API keys. Showing "locations" right now:
CleanShot 2024-01-22 at 20 04 39

olivermrbl:
Overall really nice @kasperkristensen. Left a couple of comments and suggestions 💪

kasperkristensen:
@olivermrbl Thanks I'll address the issues later today.

Re: Having the link to the user, I added that from the perspective that users should be able to navigate between related pages with ease. Just like how in Linear if I go to an issue you've created, then I can click on your avatar and go to a page with all your issues. So I'd like to keep the functionality, but I agree it seems a bit out of place as it currently is, I'll try something else.

Re: Having a `Created` field in the details, Ludvig is currently experimenting with how we can display `created_at` and `updated_at` on all pages in a nice uniform matter, so I'll add that here as well once we land on a pattern 😄

olivermrbl (approved):
Nice, LGTM!

PR #6164, opened 2024-01-22
feat(workflows-sdk): run, registerStepSuccess and registerStepFailure shortcut

adrien2p:
**What**

Currently, when exporting a workflow using the workflowExport util, it is mandatory to first call the resulted function passing the container before being able to call run, registerStepSuccess or failure on it. Now, it is possible to either continue that way, or to directly call the run, registerStepSuccess or failure on the exported workflow and at that moment it is possible to pass a container if needed

e.g
```ts
const workflow = exportWorkflow("id" as any, "result_step", prepare)
const wfRunner = workflow(container)
wfRunner.run(...) // Here the container is not expected
```
or
```ts
const workflow = exportWorkflow("id" as any, "result_step", prepare)
workflow.run(...) // Here we can now pass an optional container 
```


adrien2p:

carlos-r-l-rodrigues (approved):
LGTM

adrien2p:
@carlos-r-l-rodrigues do you think I should merge so that I can include it in our pr?

olivermrbl:
I would love to give it a spin 👍

adrien2p:
@olivermrbl sure, you can look at the tests to see the usage. But at somepoint everything will go through the engine, so it will always go through the enigine run method which among other things accept the container as part of the options 👍

olivermrbl:
This is great!

And if there's no container, it will try to use the loaded modules under the hood right?

So for example:
```ts
const createSomething = createSomethingWorkflow()

const { result, errors } = await createSomething.run({
  input: { ... }
})
```

adrien2p:
yes, or you can do

```ts
// Here it will use the loaded modules
const { result, errors } = await createSomethingWorkflow.run({
  input: { ... }
})
```

no need for the first step

olivermrbl:
Love it 😍

carlos-r-l-rodrigues:
We have to support both ways:
```typescript
const wf = myWorkflow();
wf.run({})
```
and
```typescript
myWorkflow.run({})
```

The first one is an instance of LocalWorkflows that is the only way you can extend it.
```typescript
const wf = myWorkflow();
wf.addStep(...)
wf.run({})
```

olivermrbl:
@carlos-r-l-rodrigues, makes sense!

PR #6166, opened 2024-01-22
Fix update product's collection

lukebui:
If user attempts to remove a product from a collection, then setting `collection_id` to `undefined` in the payload will do nothing. If present, the property must be a valid collection id or `null`.


riqwan:
@lukebui can you update the tests on this one? The builds are failing due to the new type. If they're fixed, looks like an easy merge. 

Closing for now, feel free to reopen if this is still a need 🤞🏻

PR #6184, opened 2024-01-23
feat(authentication, types): split authenticate method into two

pKorsholm:
**What**
- split `authtenticate` method into
  - `initializeAuthentication`
  - `authenticateCallback`

**Why**
- OAuth2 handlers require initialization of a request to a third party and handling a callback when the authentication is validated. This pr splits the two actions to ensure each method is only responsible for handling a specific part of the authentication flow


packages/authentication/src/providers/google.ts
```diff
@@ -91,58 +74,25 @@ class GoogleProvider extends AbstractAuthenticationModuleProvider {
...
 
-    let { callbackURL, clientID, clientSecret } = config
-
-    const meta: ProviderConfig = {
-      clientID,
-      callbackURL,
-      clientSecret,
-    }
-
     const code = (req.query && req.query.code) || (req.body && req.body.code)
```
carlos-r-l-rodrigues:
```suggestion
    const code = req.query?.code ?? req.body?.code
```

packages/authentication/src/providers/google.ts
```diff
@@ -169,6 +119,67 @@ class GoogleProvider extends AbstractAuthenticationModuleProvider {
...
+    if (!config.callbackURL) {
+      throw new Error("Google callbackUrl is required")
+    }
+  }
+
+  private originalURL(req: AuthenticationInput) {
+    const tls = req.connection.encrypted,
+      host = req.headers.host,
+      protocol = tls ? "https" : "http",
+      path = req.url || ""
```
carlos-r-l-rodrigues: nit: 1 const per line

carlos-r-l-rodrigues (approved):

olivermrbl:
Can I get you to rebase off of `develop`?

packages/authentication/src/providers/google.ts
```diff
@@ -0,0 +1,221 @@
...
+  protected readonly authProviderService_: AuthProviderService
+
+  constructor({ authUserService, authProviderService }: InjectedDependencies) {
+    super()
+
+    this.authUserSerivce_ = authUserService
+    this.authProviderService_ = authProviderService
+  }
+
+  async initiateAuthentication(
```
olivermrbl: thought(non-blocking): Should we call this`authorize`? I've seen this used in other auth APIs, so perhaps this would be more familiar to developers.
```suggestion
  async authorize(
```

packages/authentication/src/providers/google.ts
```diff
@@ -0,0 +1,221 @@
...
+    try {
+      config = await this.getProviderConfig(req)
+    } catch (error) {
+      return { success: false, error: error.message }
+    }
+
+    return this.getRedirect(config)
+  }
+
+  async authenticateCallback(
```
olivermrbl: thought, very-nit(non-blocking): Should we call this`callback`? For the same reason as above.
```suggestion
  async callback(
```

packages/authentication/src/providers/google.ts
adrien2p: usually we have authenticate and validate callback isnt it?

packages/authentication/src/providers/google.ts
```diff
@@ -0,0 +1,221 @@
...
+
+    const location = client.authorizeURL({
+      redirect_uri: callbackURL,
+      scope: "email profile",
+    })
+
+    return { success: true, location }
+  }
+
+  private getAuthorizationCodeHandler({
```
adrien2p: I think this can also be abstracted. when registering the provider in the container we have access to all the options etc, we could have a function that return a class, this function accept the provider specific config and return a class that could extend a function that takes all the specific info such as provider name, display name and config and in exchange returned a configured class, then the other method relying on the config will also be on that base class and will be able to consume those info. We can discuss that when we reach the abstraction

adrien2p: if it is not clear let me know, we could write a little poc together and discuss it. But most of it can be pre-configured before being registered in the container

packages/authentication/src/providers/google.ts
olivermrbl: Yes, maybe I was too hasty here. This is the method form the general authentication provider interface, so `authenticate` would probably be more appropriate – you're right. `authorize` is more specific to oauth

packages/authentication/src/providers/google.ts
pKorsholm: I did `validateCallback` instead to just have it be a bit more descriptive

packages/authentication/src/providers/google.ts
pKorsholm: yes, I think it's clear, but if needed I'll ping you 😄

PR #6185, opened 2024-01-23
docs(development): fix yarn version

xyzones:
medusa workspace uses 3.2.1 version if yarn is any higher it will raise errors and the build will fail.

added a section on how to check and install the version needed


xyzones:
found an error sry

PR #6188, opened 2024-01-23
chore: Abstract module service

adrien2p:
**What**
- Remove services that do not have any custom business and replace them with a simple interfaces
- Abstract module service provide the following base implementation
  - retrieve
  - list
  - listAndCount
  - delete
  - softDelete
  - restore

The above methods are created for the main model and also for each other models for which a config is provided

all method such as list, listAndCount, delete, softDelete and restore are pluralized with the model it refers to

**Migration**
- [x] product
- [x] pricing
- [x] promotion
- [x] cart
- [x] auth
- [x] customer
- [x] payment
- [x] Sales channel
- [x] Workflow-*

**Usage**

**Module**

The module service can now extend the ` ModulesSdkUtils.abstractModuleServiceFactory` which returns a class with the default implementation for each method and each model following the standard naming convention mentioned above.
This factory have 3 template arguments being the container, the main model DTO and an object representing the other model with a config object that contains at list the DTO and optionally a singular and plural property in case it needs to be set manually. It looks like the following:

```ts
export default class PricingModuleService</* ... */>
  extends ModulesSdkUtils.abstractModuleServiceFactory<
    InjectedDependencies,
    PricingTypes.PriceSetDTO,
    {
      Currency: { dto: PricingTypes.CurrencyDTO }
      MoneyAmount: { dto: PricingTypes.MoneyAmountDTO }
      PriceSetMoneyAmount: { dto: PricingTypes.PriceSetMoneyAmountDTO }
      PriceSetMoneyAmountRules: {
        dto: PricingTypes.PriceSetMoneyAmountRulesDTO
      }
      PriceRule: { dto: PricingTypes.PriceRuleDTO }
      RuleType: { dto: PricingTypes.RuleTypeDTO }
      PriceList: { dto: PricingTypes.PriceListDTO }
      PriceListRule: { dto: PricingTypes.PriceListRuleDTO }
    }
  >(PriceSet, generateMethodForModels, entityNameToLinkableKeysMap)
  implements PricingTypes.IPricingModuleService
{
// ...
}
```

In the above, the singular and plural can be inferred as there is no tricky naming. Also, the default implementation does not remove the fact that you need to provides all the overloads etc in your module service interface. The above will provide a default implementation following the interface `AbstractModuleService` which is also auto generated, hence you will have the following methods available:

**for the main model**
- list
- retrieve
- listAndCount 
- delete
- softDelete
- restore

**for the other models**
- list**MyModels**
- retrieve**MyModel**
- listAndCount**MyModels**
- delete**MyModels**
- softDelete**MyModels**
- restore**MyModels**

**Internal module service**

The internal module service can now extend `ModulesSdkUtils.internalModuleServiceFactory` which takes only one template argument which is the container type. 
All internal services provides a default implementation for all retrieve, list, listAndCount, create, update, delete, softDelete, restore methods which follow the following interface `ModulesSdkTypes.InternalModuleService`:

```ts
export interface InternalModuleService<
  TEntity extends {},
  TContainer extends object = object
> {
  get __container__(): TContainer

  retrieve(
    idOrObject: string,
    config?: FindConfig<any>,
    sharedContext?: Context
  ): Promise<TEntity>
  retrieve(
    idOrObject: object,
    config?: FindConfig<any>,
    sharedContext?: Context
  ): Promise<TEntity>

  list(
    filters?: FilterQuery<any> | BaseFilterable<FilterQuery<any>>,
    config?: FindConfig<any>,
    sharedContext?: Context
  ): Promise<TEntity[]>

  listAndCount(
    filters?: FilterQuery<any> | BaseFilterable<FilterQuery<any>>,
    config?: FindConfig<any>,
    sharedContext?: Context
  ): Promise<[TEntity[], number]>

  create(data: any[], sharedContext?: Context): Promise<TEntity[]>
  create(data: any, sharedContext?: Context): Promise<TEntity>

  update(data: any[], sharedContext?: Context): Promise<TEntity[]>
  update(data: any, sharedContext?: Context): Promise<TEntity>
  update(
    selectorAndData: {
      selector: FilterQuery<any> | BaseFilterable<FilterQuery<any>>
      data: any
    },
    sharedContext?: Context
  ): Promise<TEntity[]>
  update(
    selectorAndData: {
      selector: FilterQuery<any> | BaseFilterable<FilterQuery<any>>
      data: any
    }[],
    sharedContext?: Context
  ): Promise<TEntity[]>

  delete(idOrSelector: string, sharedContext?: Context): Promise<void>
  delete(idOrSelector: string[], sharedContext?: Context): Promise<void>
  delete(idOrSelector: object, sharedContext?: Context): Promise<void>
  delete(idOrSelector: object[], sharedContext?: Context): Promise<void>
  delete(
    idOrSelector: {
      selector: FilterQuery<any> | BaseFilterable<FilterQuery<any>>
    },
    sharedContext?: Context
  ): Promise<void>

  softDelete(
    idsOrFilter: string[] | InternalFilterQuery,
    sharedContext?: Context
  ): Promise<[TEntity[], Record<string, unknown[]>]>

  restore(
    idsOrFilter: string[] | InternalFilterQuery,
    sharedContext?: Context
  ): Promise<[TEntity[], Record<string, unknown[]>]>

  upsert(data: any[], sharedContext?: Context): Promise<TEntity[]>
  upsert(data: any, sharedContext?: Context): Promise<TEntity>
}
```

When a service is auto generated you can use that interface to type your class property representing the expected internal service.

**Repositories**

The repositories can now extend `DALUtils.mikroOrmBaseRepositoryFactory` which takes one template argument being the entity or the template entity and provides all the default implementation. If the repository is auto generated you can type it using the `RepositoryService` interface. Here is the new interface typings.

```ts
export interface RepositoryService<T = any> extends BaseRepositoryService<T> {
  find(options?: FindOptions<T>, context?: Context): Promise<T[]>

  findAndCount(
    options?: FindOptions<T>,
    context?: Context
  ): Promise<[T[], number]>

  create(data: any[], context?: Context): Promise<T[]>

  // Becareful here, if you have a custom internal service, the update data should never be the entity otherwise
 // both entity and update will point to the same ref and create issues with mikro orm
  update(data: { entity; update }[], context?: Context): Promise<T[]>

  delete(
    idsOrPKs: FilterQuery<T> & BaseFilterable<FilterQuery<T>>,
    context?: Context
  ): Promise<void>

  /**
   * Soft delete entities and cascade to related entities if configured.
   *
   * @param idsOrFilter
   * @param context
   *
   * @returns [T[], Record<string, string[]>] the second value being the map of the entity names and ids that were soft deleted
   */
  softDelete(
    idsOrFilter: string[] | InternalFilterQuery,
    context?: Context
  ): Promise<[T[], Record<string, unknown[]>]>

  restore(
    idsOrFilter: string[] | InternalFilterQuery,
    context?: Context
  ): Promise<[T[], Record<string, unknown[]>]>

  upsert(data: any[], context?: Context): Promise<T[]>
}
```

### DIFF
--- a/integration-tests/plugins/__tests__/customer/store/list-customer-addresses.ts
+++ b/integration-tests/plugins/__tests__/customer/store/list-customer-addresses.ts
@@ -9,6 +9,8 @@ import { createAuthenticatedCustomer } from "../../../helpers/create-authenticat
 
 const env = { MEDUSA_FF_MEDUSA_V2: true }
 
+jest.setTimeout(100000)
+
 describe("GET /store/customers/me/addresses", () => {
   let dbConnection
   let appContainer
@@ -16,13 +18,17 @@ describe("GET /store/customers/me/addresses", () => {
   let customerModuleService: ICustomerModuleService
 
   beforeAll(async () => {
-    const cwd = path.resolve(path.join(__dirname, "..", "..", ".."))
-    dbConnection = await initDb({ cwd, env } as any)
-    shutdownServer = await startBootstrapApp({ cwd, env })
-    appContainer = getContainer()
-    customerModuleService = appContainer.resolve(
-      ModuleRegistrationName.CUSTOMER
-    )
+    try {
+      const cwd = path.resolve(path.join(__dirname, "..", "..", ".."))
+      dbConnection = await initDb({ cwd, env } as any)
+      shutdownServer = await startBootstrapApp({ cwd, env })
+      appContainer = getContainer()
+      customerModuleService = appContainer.resolve(
+        ModuleRegistrationName.CUSTOMER
+      )
+    } catch (error) {
+      console.error(error)
+    }
   })
 
   afterAll(async () => {


adrien2p:
@srindom @olivermrbl could I have your opinion? Also, it doesn't seem a lot, but if I apply that to all module the footprint will be reduced by a lot. Also, it means that now all read and delete, soft, restore method are auto generated, meaning you can focus on create, update and other custom method which in most cases are the one having most of the business logic

cc @carlos-r-l-rodrigues

packages/utils/src/common/__tests__/pluralize.spec.ts
```diff
@@ -0,0 +1,31 @@
...
+      "potato",
+    ]
+
+    const expectedOutput = [
+      "apples",
+      "boxes",
+      "days",
+      "countries",
+      "babies",
+      "knifes",
```
srindom: Annoying one - should be knives 😬

Not a concern worth addressing though

srindom:
This is amazing!

olivermrbl:
LFG!! 🚀 🚀 

Really great

packages/utils/src/common/__tests__/pluralize.spec.ts
adrien2p: 🤣 ill do it, i havent ported all the rules, just some but we can improve it in time, the config also allow to specify those manually in case you are not using english or things like that. Thanks @carlos-r-l-rodrigues for the advice for non english people 🤣

carlos-r-l-rodrigues:
```typescript
const buildMethodNamesFromModel = (
    model:
      | Constructor<any>
      | { singular?: string; plural?: string; model: Constructor<any> },
    suffixed: boolean = true
  )
```

to avoid too much magic and exceptions, I would explicitly set the names we want for our services. And that is also valid in the future if someone is not using english names.

adrien2p: It is already supported if it is passed in the config, if not present it fallback to the auto pluralize

adrien2p: @srindom whats the rule between knife and strife 😂 I wanted to not use a library ahah and have just something minimal for our usecase since it is configurable as well

srindom: > @srindom whats the rule between knife and strife 😂 I wanted to not use a library ahah and have just something minimal for our usecase since it is configurable as well

Unfortunately, I don't think the English language isn't structured enough to have general rules for this :(

srindom: But honestly not worth it to support I think - we can just document that we are going to pluralize naively, so special cases like "knives" and "leaves" will be "knifes" and "leafs"

adrien2p:
I ve reached one limitation of typescript https://github.com/microsoft/TypeScript/issues/48125#issuecomment-1906591895 if anybody want to pitch in 😅

packages/customer/src/types/services/address.ts
srindom, line 1: what is the idea with these - could they not also be abstracted away?

packages/authentication/src/types/services/auth-provider.ts
```diff
@@ -1,3 +1,20 @@
+import { AbstractService } from "@medusajs/utils"
+import { IAuthProviderRepository } from "../repositories"
+import { AuthProvider } from "@models"
+
+export interface IAuthProviderService<
+  TEntity extends AuthProvider = AuthProvider
```
olivermrbl: question: What is the reason behind keeping these service interfaces?

Same question for the repositories that live in `packages/types/cart/repositories.ts` (from a previous PR).

olivermrbl: Sorry for the repeat. Seb and I was just discussing 😅

adrien2p: Because the service does not exists so we can't use it to type it in the module service same for the other things

packages/customer/src/types/services/address.ts
adrien2p: if we don't declare anything we can use the DTO to type the internal services for our usage and we loose the typings, we already removed a lot

packages/authentication/src/types/services/auth-provider.ts
olivermrbl: ~~Would it be possible to create the service interface using just the model or is there anything specific we need?~~

adrien2p: Based on our discussion, either use a base interface that does not enforce typings of the internal services or find a way to have a base typings based on TEntity (second step). same goes for repositories

packages/utils/src/modules-sdk/internal-module-service-factory.ts
```diff
@@ -0,0 +1,427 @@
...
+      const entities = await this[propertyRepositoryName].create(
+        data_,
+        sharedContext
+      )
+
+      return Array.isArray(data) ? entities : entities[0]
+    }
+
+    update(data: any[], sharedContext?: Context): Promise<TEntity[]>
+    update(data: any, sharedContext?: Context): Promise<TEntity>
```
srindom, line 176: suggestion: update should only work with selector + data. If you want to update with an entity you can use upsert. Wdyt?

adrien2p: Thats true, we can rework that 👌

packages/customer/src/services/customer-module.ts
```diff
@@ -95,13 +99,33 @@ export default class CustomerModuleService implements ICustomerModuleService {
...
   async create(
     dataOrArray:
       | CustomerTypes.CreateCustomerDTO
       | CustomerTypes.CreateCustomerDTO[],
     @MedusaContext() sharedContext: Context = {}
-  ) {
+  ): Promise<CustomerTypes.CustomerDTO | CustomerTypes.CustomerDTO[]> {
+    const customers = await this.create_(dataOrArray, sharedContext).catch(
+      this.handleDbErrors
+    )
```
adrien2p, line 111: @srindom In both places I have applied the current pattern that we are using and removed the flush for now. we can discuss it

packages/utils/src/modules-sdk/internal-module-service-factory.ts
adrien2p: maybe we can do that in a follow-up pr just to merge this one asap? wdyt?

srindom (commented):
LGTM! Suggest that we merge and start using it ASAP, fixes and adjustments can follow in separate PRs.

adrien2p:
@olivermrbl any reason this pr is mentioned in the dashboard pr?

olivermrbl:
No that was by mistake

olivermrbl (approved):
🔥 🔥 🔥 🔥

adrien2p:
@olivermrbl @srindom ready to be merged as soon as the pipeline is green, please seb have a look to the changes I ve done on the customer module though. I had to change quite a few things, I ve put some todos here and there in different modules for naming and stuff, they are not specific but later on we can do a cleanup round

packages/pricing/integration-tests/__tests__/services/currency/index.spec.ts
```diff
@@ -183,7 +183,7 @@ describe("Currency Service", () => {
         error = e
       }
 
-      expect(error.message).toEqual('"currencyCode" must be defined')
+      expect(error.message).toEqual("currency - code must be defined")
```
olivermrbl, line 186: question: How are these error messages generated?

adrien2p: You can find them in the internal service factory. But before they where trying to provide the name of the expected argument except that it doesn't work anymore, so now it shows the primary key(s) missing

packages/utils/src/modules-sdk/internal-module-service-factory.ts
```diff
@@ -315,10 +315,8 @@ export function internalModuleServiceFactory<
           MedusaError.Types.NOT_FOUND,
           `${
             primaryKeys.length === 1
-              ? `"${
-                  lowerCaseFirst(model.name) + upperCaseFirst(primaryKeys[0])
-                }"`
-              : `${lowerCaseFirst(model.name)} ${primaryKeys.join(", ")}`
+              ? `"${lowerCaseFirst(model.name) + " - " + primaryKeys[0]}"`
```
olivermrbl, line 332: thought(non-blocking): Can we do better than this?

Ideally, we have something like:
```
currencyCode
priceListId
moneyAmountId
```

adrien2p: Wouldn't it makes you think you need to provide { currencyCode } when we expect { code } ?

PR #6190, opened 2024-01-23
fix(medusa): Implement listAndCount for UserService and update list endpoint

### LINKED ISSUE #6182: bug(medusa): User service does not have a `listAndCount` method
### Describe the bug

The UserService does not have a `listAndCount` method, and the current `list` method lacks all standard params for list queries: `limit`, `offset`, `q`, `order`, and filters such as `created_at`, `updated_at`, etc.

### System information

Node.js version:
Database:
Operating system:
Browser (if relevant):

### Expected behavior

A clear and concise description of what you expected to happen

### Screenshots

If applicable, add screenshots to help explain your problem

### Code snippets

If applicable, add code samples to help explain your problem

### Additional context

Add any other context about the problem here

kasperkristensen:
**What**
- Adds a `listAndCount` method to the `UserService`.
- Updates the `list` method to use `listAndCount` and simply return the `users` only.
- Updates `GET /admin/users` to use the new method, and implement the required query params. 
- Makes required changes to medusa-js and medusa-react.

**Testing**
- Adds two new tests and updates existing test to not use snapshots.

Closes #6182

### DIFF
--- a/packages/medusa-plugin-sendgrid/.gitignore
+++ b/packages/medusa-plugin-sendgrid/.gitignore
@@ -13,3 +13,5 @@ node_modules
 /models
 /subscribers
 
+*.tgz
+
--- a/packages/medusa-plugin-sendgrid/src/services/__tests__/sendgrid.js
+++ b/packages/medusa-plugin-sendgrid/src/services/__tests__/sendgrid.js
@@ -193,4 +193,23 @@ describe("SendGridService", () => {
     })
     expect(mockedSendGrid.send).toBeCalled()
   })
+
+  it("should send message to non predefined template", async () => {
+    sendGridService = new SendGridService({}, { "send-otp": "test-template" })
+
+    await sendGridService.sendNotification("send-otp", {
+      otp: "test",
+      validity: "12-01-2020",
+    })
+    expect(mockedSendGrid.send).toBeCalled()
+    expect(mockedSendGrid.send).toHaveBeenCalledWith(
+      expect.objectContaining({
+        template_id: "test-template",
+        dynamic_template_data: {
+          otp: "test",
+          validity: "12-01-2020",
+        },
+      })
+    )
+  })
 })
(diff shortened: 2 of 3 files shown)


packages/medusa/src/api/routes/admin/users/index.ts
```diff
@@ -30,11 +31,31 @@ export default (app) => {
...
   return app
 }
 
+export const defaultAdminUserFields: (keyof User)[] = [
+  "id",
+  "email",
+  "first_name",
+  "last_name",
+  "role",
+  "api_token",
```
srindom, line 52: thought(non-blocking): maybe we don't want this to be exposed to everyone, but only the logged in user? Don't know how it's been in the past though

srindom (approved):

packages/medusa/src/api/routes/admin/users/index.ts
kasperkristensen: @srindom I thought so to, but the integration tests failed when I initially made it so the listAndCount endpoint didn't return it. ~~I am down to removing it, as I also think it is weird that this very "private" thing is available to all users, but it would be a breaking change in some sense.~~

It seems like quite a few of the user endpoints expect being able to work with `api_token` so removing it would be a breaking change for multiple endpoints. I think its best to just leave it for now, and consider not making it part of the User module, or at least maybe only making it accessible to `admin` users.

packages/medusa/src/services/user.ts
```diff
@@ -62,9 +63,53 @@ class UserService extends TransactionBaseService {
    * @param {Object} config - the configuration object for the query
    * @return {Promise} the result of the find operation
    */
-  async list(selector: FilterableUserProps, config = {}): Promise<User[]> {
+  async list(
+    selector: Selector<FilterableUserProps> & { q?: string } = {},
+    config: FindConfig<FilterableUserProps> = { skip: 0, take: 20 }
+  ): Promise<User[]> {
+    const [users] = await this.listAndCount(selector, config)
```
adrien2p: Todo: i ve just seen the email description and wanted to share that we shouldn't do that. It in fact perform one to two more queries under the hood to get the count which add up to the response time. So if we can it is better as a general approach to differentiate the two list and listAndCount method

kasperkristensen: All good, I have changed it to not use `listAndCount` I was just trying to follow the pattern we have in most of our other services ✌️

adrien2p: ahah no problem, unfortunatelly this one is not one to follow 😂

PR #6197, opened 2024-01-24
feat(medusa, stock-location, types): Add q and order params to list-stock-locations

### LINKED ISSUE #6189: bug(medusa): `GET /admin/stock-locations` does not have a `q` and `order` param
### Describe the bug

The `GET /admin/stock-locations` does not have a `q` and `order` param, making it difficult to filter the list of stock locations.

### System information

Node.js version:
Database:
Operating system:
Browser (if relevant):

### Expected behavior

A clear and concise description of what you expected to happen

### Screenshots

If applicable, add screenshots to help explain your problem

### Code snippets

If applicable, add code samples to help explain your problem

### Additional context

Add any other context about the problem here

pKorsholm:
**What**
- add `q` and `order` params to list-stock-location endpoint

closes #6189


packages/stock-location/src/utils/build-query.ts
```diff
@@ -159,7 +159,11 @@ function buildWhere<TWhereKeys extends object, TEntity>(
       continue
     }
 
-    where[key] = value
+    if (key === "q") {
+      where["name"] = ILike(`%${value}%`)
+    } else {
+      where[key] = value
+    }
```
kasperkristensen: Would it perhaps make sense to instead do like we do in `medusa`:

```
let q: string | undefined

if (selector.q) {
  q = selector.q
  delete selector.q
}

const query = buildQuery(selector, config)

if (q) {
  const where = query.where as ...

  delete where.name

  query.where = [
    {
      ...where,
      name: ILike(`%${q}%`)
    }
  ]
}

// ... find and count
```

Otherwise if you have `?q=something&name=something_else` then `q` will be overridden, and we usually have the opposite where `q` overrides any filter set on the searchable fields.

packages/types/src/stock-location/common.ts
```diff
@@ -1,4 +1,4 @@
-import { StringComparisonOperator } from "../common/common"
+import { StringComparisonOperator } from "../common/common";
```
kasperkristensen: lint: rm `;`

kasperkristensen (commented):
Thanks Phil! Looks great just added two small comments ✌️

olivermrbl (approved):

olivermrbl:
@pKorsholm can I get you to fix the integration tests, so we can merge?

PR #6204, opened 2024-01-24
fix(medusa): List Users return type

kasperkristensen:
**What**
- Missed updating the return type of the list request in the previous PR. This fixes the return type so it has count, limit and offset.


srindom (approved):
LGTM!

PR #6205, opened 2024-01-24
feat(customer): add migration file

srindom:


packages/customer/src/migrations/Migration20240124121016.ts
```diff
@@ -0,0 +1,51 @@
...
+    )
+    this.addSql(
+      'create index if not exists "IDX_customer_group_customer_group_id" on "customer_group_customer" ("customer_group_id");'
+    )
+    this.addSql(
+      'create index if not exists "IDX_customer_group_customer_customer_id" on "customer_group_customer" ("customer_id");'
+    )
+
+    this.addSql(
+      'alter table "customer" add constraint "customer_default_shipping_address_id_foreign" foreign key ("default_shipping_address_id") references "customer_address" ("id") on update cascade on delete set null;'
```
carlos-r-l-rodrigues: ### 
```suggestion
      'alter table "customer" add constraint if not exists "customer_default_shipping_address_id_foreign" foreign key ("default_shipping_address_id") references "customer_address" ("id") on update cascade on delete set null;'
```

packages/customer/src/migrations/Migration20240124121016.ts
```diff
@@ -0,0 +1,51 @@
...
+    )
+    this.addSql(
+      'alter table "customer" add constraint "customer_default_billing_address_id_foreign" foreign key ("default_billing_address_id") references "customer_address" ("id") on update cascade on delete set null;'
```
carlos-r-l-rodrigues:
```suggestion
      'alter table "customer" add constraint if not exists "customer_default_billing_address_id_foreign" foreign key ("default_billing_address_id") references "customer_address" ("id") on update cascade on delete set null;'
```

packages/customer/src/migrations/Migration20240124121016.ts
```diff
@@ -0,0 +1,51 @@
...
+    )
+
+    this.addSql(
+      'alter table "customer_address" add constraint "customer_address_customer_id_foreign" foreign key ("customer_id") references "customer" ("id") on update cascade;'
```
carlos-r-l-rodrigues:
```suggestion
      'alter table "customer_address" add constraint if not exists "customer_address_customer_id_foreign" foreign key ("customer_id") references "customer" ("id") on update cascade;'
```

packages/customer/src/migrations/Migration20240124121016.ts
```diff
@@ -0,0 +1,51 @@
...
+    )
+
+    this.addSql(
+      'alter table "customer_group_customer" add constraint "customer_group_customer_customer_group_id_foreign" foreign key ("customer_group_id") references "customer_group" ("id") on delete cascade;'
```
carlos-r-l-rodrigues:
```suggestion
      'alter table "customer_group_customer" add constraint if not exists "customer_group_customer_customer_group_id_foreign" foreign key ("customer_group_id") references "customer_group" ("id") on delete cascade;'
```

packages/customer/src/migrations/Migration20240124121016.ts
```diff
@@ -0,0 +1,51 @@
...
+    )
+    this.addSql(
+      'alter table "customer_group_customer" add constraint "customer_group_customer_customer_id_foreign" foreign key ("customer_id") references "customer" ("id") on delete cascade;'
```
carlos-r-l-rodrigues:
```suggestion
      'alter table "customer_group_customer" add constraint if not exists "customer_group_customer_customer_id_foreign" foreign key ("customer_id") references "customer" ("id") on delete cascade;'
```

packages/customer/src/migrations/Migration20240124121016.ts
```diff
@@ -0,0 +1,51 @@
...
+    )
+    this.addSql(
+      'create index if not exists "IDX_customer_default_shipping_address_id" on "customer" ("default_shipping_address_id");'
+    )
+    this.addSql(
+      'create index if not exists "IDX_customer_default_billing_address_id" on "customer" ("default_billing_address_id");'
+    )
+
+    this.addSql(
+      'create table if not exists "customer_address" ("id" text not null, "customer_id" text not null, "company" text null, "first_name" text null, "last_name" text null, "address_1" text null, "address_2" text null, "city" text null, "country_code" text null, "province" text null, "postal_code" text null, "phone" text null, "metadata" jsonb null, "created_at" timestamptz not null default now(), "updated_at" timestamptz not null default now(), constraint "customer_address_pkey" primary key ("id"));'
```
carlos-r-l-rodrigues: we can have an index on `customer_address`.`customer_id`

srindom:
@carlos-r-l-rodrigues redid the file please take another look

packages/customer/src/migrations/Migration20240124154000.ts
```diff
@@ -0,0 +1,79 @@
+import { Migration } from "@mikro-orm/migrations"
+
+export class Migration20240124154000 extends Migration {
+  async up(): Promise<void> {
+    // Adjust existing 'customer' table
+    this.addSql(
```
adrien2p, line 9: q: in this migration, do we need to take into account to drop unnecessary columns from v1 and their index? or do we plan to keep the existing columns and just alter/add the changed columns and then do the full migration separately? for the data, should we include that in the migration as well? e.g moving the billing_address to default_belling_address or things like that if necessary, I took this column as example, it is not necessarly true? what is the global plan?

packages/customer/src/migrations/Migration20240124154000.ts
```diff
@@ -0,0 +1,79 @@
...
+      'alter table "customer" add column if not exists "company_name" text null;'
```
adrien2p, line 10: Aren't we missing a create table if not exists here?

adrien2p: if you use the module on a separate database, the customer table won't exists

srindom: Yes you are right! See latest fix that should work right?

packages/customer/src/migrations/Migration20240124154000.ts
srindom: It's a good question for now I am working with the assumption that we only get tables "brought up to speed" with modules. Eventually, we will have to figure out a plan for data migration and cleanup of unused columns/tables, but I am not entirely sure if this should be the responsibility of the module - right now I am leaning more toward this being something the core handles (potentially with a workflow :))

packages/customer/src/migrations/Migration20240124154000.ts
adrien2p: we are still missing the create table in the last commit

adrien2p: the module can be started in a different database, in that case the user can use the runMigration function to spin up his table in his database

packages/customer/src/migrations/Migration20240124154000.ts
adrien2p: yes indeed, we could do something like that, indeed and that way it can work no matter where the database are in both parties 🎉

carlos-r-l-rodrigues (approved):
LGTM
It is just missing the create table customer if not exists.

carlos-r-l-rodrigues:
Sorry, I just spotted a couple of other details.
`customer_group` table is also not being created if not exists, and we have to add the `delete_at` to it. (indexes created on this table also need the `where delete_at is null`)

All the tables involved in public methods exposed by the module will have to be "soft deletable", that way the workflows can restore it if needed.

srindom:
> Sorry, I just spotted a couple of other details.
customer_group table is also not being created if not exists, and we have to add the delete_at to it. (indexes created on this table also need the where delete_at is null)
> All the tables involved in public methods exposed by the module will have to be "soft deletable", that way the workflows can restore it if needed.

@carlos-r-l-rodrigues - Is there something that needs to be done from a MikroORM perspective for this to work properly right now I just added it in the raw SQL?

srindom:
Seems like it's only supported through custom expressions https://github.com/mikro-orm/mikro-orm/issues/2837

So this should be sufficient right?

carlos-r-l-rodrigues:
> Seems like it's only supported through custom expressions [mikro-orm/mikro-orm#2837](https://github.com/mikro-orm/mikro-orm/issues/2837)
> 
> So this should be sufficient right?

Correct. if we want to auto generate that.
for the `if [not] exists` I haven't found anything

PR #6206, opened 2024-01-24
feat(customer): list customer

srindom:
**What**

- GET /admin/customers
- GET /admin/customer-groups


packages/medusa/src/api-v2/admin/customer-groups/route.ts
```diff
@@ -0,0 +1,24 @@
+import { ModuleRegistrationName } from "@medusajs/modules-sdk"
+import { ICustomerModuleService } from "@medusajs/types"
+import { MedusaRequest, MedusaResponse } from "../../../types/routing"
+
+export const GET = async (req: MedusaRequest, res: MedusaResponse) => {
+  const customerModuleService = req.scope.resolve<ICustomerModuleService>(
```
carlos-r-l-rodrigues, line 6: We can align how we want to build the GET endpoints.
I believe we should use the remoteQuery to be able to easily expand relations from here.

We have 2 utils that can help on that:
[remoteQueryObjectFromString](https://github.com/medusajs/medusa/blob/develop/packages/utils/src/common/remote-query-object-from-string.ts
) and [remoteQueryObjectToString](https://github.com/medusajs/medusa/blob/develop/packages/utils/src/common/remote-query-object-to-string.ts)

we can discuss it on Monday and come up with a pattern to follow in all the api-v2 endpoints. cc: @adrien2p @olivermrbl

srindom: Agree remote query should be the way to go - do we have an example implementation of the approach somewhere?

carlos-r-l-rodrigues: We have mixed examples that we need to normalize from now on.
I believe [this test](https://github.com/medusajs/medusa/blob/develop/packages/utils/src/common/__tests__/remote-query-object-from-string.spec.ts) is a good starting point, where we have a list of all default fields and relations to be listed. We need to think about and implement how to have a list of "allowed" relations to be expanded, and how to customize this list, adding custom relations to it.

from there it is used like this:

```typescript
const remoteQuery = container.resolve("remoteQuery")

const query = remoteQueryObjectFromString({
  entryPoint: "product",
  variables: {},
  fields,
})

const results = await remoteQuery(query)
```

variables are [handled here](https://github.com/medusajs/medusa/blob/develop/packages/modules-sdk/src/remote-query.ts#L155-L204) before calling `module.list(filters, options)`
We can send all the filters and options needed to the service from here.

As this will be replicate in most GET endpoints, I think it worth spending some time in it, even if in another PR.

srindom: I think I will need to see an example with filters and expands applied - I wasn't able to get this to work for these.

I suggest that we merge PRs without RemoteQuery until we have its feature set more standardized.

carlos-r-l-rodrigues (approved):
LGTM

PR #6212, opened 2024-01-24
feat(dashboard): Users domain

kasperkristensen:
**What**
- Adds users domain with list, detail, edit, and invite views


srindom:
This looks incorrect:
CleanShot 2024-01-26 at 1  54 17

When clicking edit from the table view, I think the experience would be better if the edit menu folds out directly over the table instead of navigating you to the user details page and then opening the edit side bar.
https://github.com/medusajs/medusa/assets/7554214/ce1b8a16-1043-4e68-b4eb-38d05abea4f3

olivermrbl:
Aside from @srindom's comments, LGTM!

kasperkristensen:
@srindom It seems like the staging server hasn't been redeployed with the latest updates to the /admin/users endpoint so it's not returning `count`, which is why it's showing 0 of 0. If you run it locally with the latest version of Medusa it should be good

srindom (approved):
> @srindom It seems like the staging server hasn't been redeployed with the latest updates to the /admin/users endpoint so it's not returning count, which is why it's showing 0 of 0. If you run it locally with the latest version of Medusa it should be good

Got it! LGTM

PR #6213, opened 2024-01-24
chore: move v2 api behind ff

srindom:


packages/medusa/src/loaders/api.ts
```diff
@@ -25,7 +34,26 @@ export default async ({ app, container, configModule }: Options) => {
   })
 
   app.use(bodyParser.json())
-  app.use("/", routes(container, configModule.projectConfig))
+
+  if (featureFlagRouter.isFeatureEnabled(FeatureFlagUtils.MedusaV2Flag.key)) {
+    // TODO: Figure out why this is causing issues with test when placed inside ./api.ts
```
olivermrbl, line 39: nit: I guess we can remove the todo, if this is no longer causing issues

olivermrbl (approved):
LGTM

packages/medusa/src/loaders/api.ts
adrien2p: Some integration tests are testing the route under API to validate the workflows and they have the v2 enabled, to the API is not loaded anymore and the tests fails

adrien2p: for the unit tests, it seams that in the api `featureFlagRouter` is not available

PR #6228, opened 2024-01-26
feat(authentication, types, utils): Add Authentication provider scopes

pKorsholm:
**What**
- Add providers as options for authentication module
- validate provider scope is defined when authenticating
- Type `authenticate` and `verifyCallback`

### DIFF
--- a/.changeset/shy-cars-turn.md
+++ b/.changeset/shy-cars-turn.md
@@ -1,0 +1,7 @@
+---
+"@medusajs/stock-location": patch
+"@medusajs/medusa": patch
+"@medusajs/types": patch
+---
+
+feat(stock-location, medusa, types): add `q` and `order` query parameters to stock locations list endpoint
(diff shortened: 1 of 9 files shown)


packages/authentication/integration-tests/__tests__/services/providers/username-password.spec.ts
```diff
@@ -91,6 +92,7 @@ describe("AuthenticationModuleService - AuthProvider", () => {
 
       const res = await service.authenticate("usernamePassword", {
```
srindom, line 93: would be nice to rename this to something like `userpass` or `user-pass`

olivermrbl: Or use `local`. Think people will find that familiar from passportjs

pKorsholm: rename in #6251

packages/types/src/authentication/common/provider.ts
```diff
@@ -2,4 +2,21 @@ export type AuthenticationResponse = {
...
   error?: string
+  location?: string
+}
+
+export type AuthModuleProviderConfig = {
+  name: string
+  scopes: Record<string, AuthProviderScope>
+}
+
+export type AuthProviderScope = { domain?: string } & Record<string, string>
```
olivermrbl: nit(non-blocking): Should this type be entirely open-ended? Are we using `domain` yet?
```suggestion
export type AuthProviderScope = Record<string, unknown>
```

olivermrbl (approved):
LGTM

PR #6231, opened 2024-01-26
feat(medusa, core-flows): add v2 endpoints for create-customer and customers/me

pKorsholm:
**What**
- Add `create-customer` endpoint to api-v2
- Add `customers/me` endpoint to api-v2

**Why**
- In order to properly develop and test auth middlewares we need endpoints for creating-customers and an authenticated endpoint. This could have worked with users but since the customer module is coming along I used that

### DIFF
--- a/.changeset/smart-ants-compete.md
+++ b/.changeset/smart-ants-compete.md
@@ -1,0 +1,5 @@
+---
+"@medusajs/medusa": patch
+---
+
+fix(medusa): improve error message on incorrect service export
--- a/packages/cart/src/joiner-config.ts
+++ b/packages/cart/src/joiner-config.ts
@@ -1,10 +1,26 @@
 import { Modules } from "@medusajs/modules-sdk"
 import { ModuleJoinerConfig } from "@medusajs/types"
 import { MapToConfig } from "@medusajs/utils"
-import { Cart } from "@models"
+import {
+  Address,
+  Cart,
+  LineItem,
+  LineItemAdjustment,
+  LineItemTaxLine,
+  ShippingMethod,
+  ShippingMethodAdjustment,
+  ShippingMethodTaxLine,
+} from "@models"
 
 export const LinkableKeys = {
   cart_id: Cart.name,
+  line_item_id: LineItem.name,
+  shipping_method_id: ShippingMethod.name,
+  address_id: Address.name,
+  line_item_adjustment_id: LineItemAdjustment.name,
+  shipping_method_adjustment_id: ShippingMethodAdjustment.name,
+  line_item_tax_line_id: LineItemTaxLine.name,
+  shipping_method_tax_line_id: ShippingMethodTaxLine.name,
 }
 
 const entityLinkableKeysMap: MapToConfig = {}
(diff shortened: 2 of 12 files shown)


packages/core-flows/src/definition/customer/create-customer.ts
```diff
@@ -0,0 +1,13 @@
...
+import { createWorkflow, WorkflowData } from "@medusajs/workflows-sdk"
+import { createStoreCustomerStep } from "../../handlers/customer"
+
+type WorkflowInput = { customersData: CreateCustomerDTO[] }
+
+export const createStoreCustomersWorkflowId = "create-store-customers"
+export const createStoreCustomersWorkflow = createWorkflow(
+  createStoreCustomersWorkflowId,
+  (input: WorkflowData<WorkflowInput>): WorkflowData<CustomerDTO[]> => {
+    return createStoreCustomerStep(input.customersData)
```
srindom: thought: there should be an additional step in this workflow along the lines of:

```js
  const [customer] = createStoreCustomerStep(input.customersData)
  associateCustomerWithAuthUser(input.authUserId, customer.id)
  return customer
```

packages/core-flows/src/definition/customer/create-customer.ts
```diff
@@ -0,0 +1,13 @@
+import { CreateCustomerDTO, CustomerDTO } from "@medusajs/types"
+import { createWorkflow, WorkflowData } from "@medusajs/workflows-sdk"
+import { createStoreCustomerStep } from "../../handlers/customer"
+
+type WorkflowInput = { customersData: CreateCustomerDTO[] }
+
+export const createStoreCustomersWorkflowId = "create-store-customers"
+export const createStoreCustomersWorkflow = createWorkflow(
```
srindom: would rename this to something that also signifies that there is an "account" creation

packages/core-flows/src/definition/customer/create-customer.ts
```diff
@@ -0,0 +1,13 @@
+import { CreateCustomerDTO, CustomerDTO } from "@medusajs/types"
+import { createWorkflow, WorkflowData } from "@medusajs/workflows-sdk"
+import { createStoreCustomerStep } from "../../handlers/customer"
+
+type WorkflowInput = { customersData: CreateCustomerDTO[] }
```
srindom, line 5:
```suggestion
type WorkflowInput = { customersData: CreateCustomerDTO, authUserId: string }
```

packages/medusa/src/api-v2/store/customers/route.ts
```diff
@@ -0,0 +1,14 @@
...
+import { createStoreCustomersWorkflow } from "@medusajs/core-flows"
+import { CreateCustomerDTO } from "@medusajs/types"
+
+export const POST = async (req: MedusaRequest, res: MedusaResponse) => {
+  const createCustomers = createStoreCustomersWorkflow(req.scope)
+  const customersData = [req.validatedBody as CreateCustomerDTO]
+
+  const { result } = await createCustomers.run({
+    input: { customersData },
+  })
```
srindom: thought: would do something like this

```
  const authUserId = // however the authUser.id is placed on req
  const createCustomerAccount = createCustomerAccountWorkflow(req.scope)
  const customersData = req.validatedBody as CreateCustomerDTO

  const { result } = await createCustomerAccount.run({
    input: { customersData, authUserId },
  })
```

packages/core-flows/src/definition/customer/create-customer.ts
srindom: we can reuse the step to create customers both in admin and "store account creation"

packages/medusa/src/api-v2/store/customers/query-config.ts
```diff
@@ -0,0 +1,38 @@
...
+  "groups",
+]
+export const allowedStoreCustomersRelations = [
+  ...defaultStoreCustomersRelations,
+]
+export const defaultStoreCustomersFields: (keyof CustomerDTO)[] = [
+  "id",
+  "email",
+  "default_billing_address_id",
+  "default_shipping_address_id",
```
srindom, line 16: these will be removed in one of my PRs

srindom (commented):
Just a few preliminary comments but looks good!

pKorsholm:
@srindom I agree with all the points, I omitted the auth-module part of this pr to establish a foundation to build the authentication flow on top of in a subsequent pr 😄 I don't mind "drafting" this and making one larger pr later, but I don't think the auth-integration here is essential and would prefer to defer the work to a subsequent pr

srindom:
In that case, I'd suggest you add TODOs in all the places where you intend to add things in a follow-up PR :)

PR #6235, opened 2024-01-26
feat(customer): admin addresses

srindom:
**What**
- GET /admin/customers/:id/addresses
- POST /admin/customers/:id/addresses
- POST /admin/customers/:id/addresses/:address_id
- DELETE /admin/customers/:id/addresses/:address_id

### DIFF
--- a/packages/medusa/src/api-v2/admin/customers/[id]/addresses/[address_id]/route.ts
+++ b/packages/medusa/src/api-v2/admin/customers/[id]/addresses/[address_id]/route.ts
@@ -11,7 +11,7 @@ export const GET = async (req: MedusaRequest, res: MedusaResponse) => {
     ModuleRegistrationName.CUSTOMER
   )
 
-  const customer = await customerModuleService.listAddresses(
+  const [address] = await customerModuleService.listAddresses(
     { id: req.params.address_id, customer_id: req.params.id },
     {
       select: req.retrieveConfig.select,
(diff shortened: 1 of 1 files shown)


fPolic (approved):
LGTM

packages/core-flows/src/customer/steps/update-addresses.ts
```diff
@@ -0,0 +1,54 @@
...
+    if (!prevCustomerAddresses) {
+      return
+    }
+
+    const service = container.resolve<ICustomerModuleService>(
+      ModuleRegistrationName.CUSTOMER
+    )
+
+    await promiseAll(
+      prevCustomerAddresses.map((c) => service.updateAddress(c.id, { ...c }))
```
carlos-r-l-rodrigues, line 51: is there a way to update that in bulk?

carlos-r-l-rodrigues (approved):
LGTM

packages/core-flows/src/customer/steps/update-addresses.ts
srindom: Not right now - we could add an `upsertAddress`, but would like the `update*` signatures to be `(selector, updateData)` instead of `({ id, ...updateData })`

@adrien2p - is `upsert` baked into the abstract services?

carlos-r-l-rodrigues: it is 💃🏻

PR #6237, opened 2024-01-26
docs: change process for generating docblocks through actions

shahednasser:
This PR changes the original process of generating docblocks through actions. The process now is:

1. When a PR is merged for the branch `changeset-release/develop`, the docblock generator tool us used to generate docblocks for the changed files. The changed files are determined by retrieving all comments since the last release and the files in each of those commits.
2. If there are changes after using the docblock generator tool, a PR is opened in the branch `chore/generate-tsdocs`.
3. Once the `chore/generate-tsdocs` is merged, it triggers an action that generates the references for the docs. This changes the previous behaviour of generating references on a new release.

Both actions (that runs the docblock generator tool and that generates references for the docs) can also be triggered manually.

### DIFF
--- a/.changeset/smart-ants-compete.md
+++ b/.changeset/smart-ants-compete.md
@@ -1,0 +1,5 @@
+---
+"@medusajs/medusa": patch
+---
+
+fix(medusa): improve error message on incorrect service export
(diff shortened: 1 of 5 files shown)


olivermrbl (approved):
Think this is great @shahednasser, let's give it a shot!

PR #6240, opened 2024-01-27
fix: improve error message on incorrect service export

srindom:
**What**
- Provide a more helpful error message if a user adds a file in `src/services` that doesn't export a service class.

**Why**
If you forget to `export default MyClass` in a custom service you can end up with this error:
CleanShot 2024-01-27 at 6  24 19

It's almost impossible to know how to recover from this which can be an issue for new users. The new error message informs the user that there is a missing class export and shows which file the error is related to.

### DIFF
--- a/.changeset/smart-ants-compete.md
+++ b/.changeset/smart-ants-compete.md
@@ -1,0 +1,5 @@
+---
+"@medusajs/medusa": patch
+---
+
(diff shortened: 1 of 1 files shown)


carlos-r-l-rodrigues (approved):
👌🏻 LGTM

PR #6243, opened 2024-01-28
chore(core-flows): reorganize folder structure for promotion workflows

riqwan:
workflows folder structure:

```
- src/
  - promotion/
    - workflows/
      - create-promotion.ts
    - steps/
      - prepare-create-promotion-data.ts
```

RESOLVES CORE-1688


.changeset/fluffy-balloons-eat.md
srindom, line 1: thought: does it make sense to include these for v2 work since it's not user facing yet? Or maybe we should at least indicate that this is for v2 so the changelog will reflect it. cc: @olivermrbl

srindom (approved):
LGTM

.changeset/fluffy-balloons-eat.md
olivermrbl: We can omit the changeset. 

However, since this package can contain both v1 and v2 updates, we can't guarantee these changes won't be part of a future release anyway. 

Also, not sure how noting the FF would work with PRs that touch several packages and with both v1 and v2 changes. Perhaps, breaking up the changesets could be an approach for having a sensible changelog for this case.

PR #6248, opened 2024-01-28
fix: updated russian tranlation file

elzaresen:


olivermrbl:
@elzaresen, appreciate the contribution. However, our admin system is currently undergoing a redesign that includes improvements to and simplifying of translations. We expect to go live with the update next month. If you are still interested in contributing then, we would very much appreciate a new PR with these changes.

PR #6251, opened 2024-01-29
feat(auth): Rename userpass

pKorsholm:
**What**
- rename usernamePassword to userpass


olivermrbl (commented):
thought(very-nit): Would it be more appropriate to call this `emailpass`? I know we use email as the username, so the name makes sense. However, one could imagine a different provider using handle (e.g. GH, Discord, etc.) as username. In that case, it would be handlepass (or something similar).

pKorsholm:
sure thing! emailpass makes sense

PR #6253, opened 2024-01-29
feat: Add BigNumber implementation

olivermrbl:
> This is a proposal - not necessarily the end result - to kick off the discussion about the implementation of the new totals utilities

### What
Introduces a BigNumber class implementation, enabling us to work with high-precision numeric values.

**Scope**
- Introduce the BigNumber class
- Remain somewhat backward-compatible (in behavior)
- Establish a foundation for handling high-precision values in more complex scenarios

**Not in scope**
- The implementation will not address complex use cases. However, the concept introduced now should be open for extensibility, so this can be added later without major changes to the calculation logic

### How
There are significant changes to three areas in this PR:
- Schemas
- (De)-Serialization
- Totals calculations

**Schemas**

Domains that need high-precision values will have two DB columns for each value in the database: a standard numeric column and a raw value column.

The standard column is for basic operations like sorting and filtering in the database and is what should be publicly exposed in our API.

The raw value is initially used solely for precise calculations and is stored as a JSONB column. Keeping it as JSONB is flexible and will allow us to extend the concept in future iterations. As of now, the raw value will only require a single property `value`.

**(De)-Serialization**

We cast the raw JSONB value to a `BigNumberRawValue` when reading from the database. 

We serialize the standard value to a `BigNumber` when reading from the database. 

We use the standard numeric value to construct the raw value upon writing to the database.

For example, the unit price and raw unit price on line items will be inserted as follows:
```ts
@BeforeCreate()
onCreate() {
  this.id = generateEntityId(this.id, "cali")
  
  const asBigNumber = new BigNumber(this.raw_unit_price ?? this.unit_price)
  
  this.unit_price = asBigNumber.numeric
  this.raw_unit_price = asBigNumber.raw
}
```

**Totals calculations**

For totals calculations, we will use the [`bignumber.js`](https://github.com/MikeMcl/bignumber.js/) library. The library ships with a `BigNumber` class with arithmetic methods for precise calculations. 

When we need to perform a calculation, we construct the BigNumber class from the library using the raw value from the database.

Let's have a look at an oversimplified example:
```ts
// create cart with line items
const [createdCart] = await service.create([
  {
    currency_code: "eur",
    items: [
      // li_1234
      {
        title: "test",
        quantity: 2,
        unit_price: 100,
      },
      // li_4321
      {
        title: "test",
        quantity: 3,
        // raw price creation
        unit_price: 200,
      },
    ],
  },
])
```

```ts
// calculating line item totals
import BN from "bignumber.js"

const lineItem1 = await service.retrieveLineItem("li_1234")
const lineItem2 = await service.retrieveLineItem("li_4321")

const bnUnitPrice1 = new BN(lineItem1.unit_price.raw)
const bnUnitPrice2 = new BN(lineItem2.unit_price.raw)

const line1Total = bnUnitPrice1.multipliedBy(lineItem1.quantity)
const line2Total = bnUnitPrice2.multipliedBy(lineItem2.quantity)

const total = line1Total.plus(line2Total)
```

**A note on backward compatibility**
Our BigNumber implementation is built to support the existing behavior of numeric values in the database. So even though we serialize the value to a BigNumber, you will still be able to treat it as a standard number, as we've always done.

For example, the following works perfectly fine:
```ts
const lineItem = await service.createLineItem({
  title: "test",
  quantity: 2,
  unit_price: 100,
})

console.log(lineItem.unit_price) // will print `100`
```

However, the type of `unit_price` will be `number | BigNumber`.


packages/utils/src/totals/index.ts
```diff
@@ -0,0 +1,280 @@
+import { CartDTO, CartShippingMethodDTO } from "@medusajs/types"
```
olivermrbl: comment: WIP - here for demoing purposes

packages/cart/src/models/line-item.ts
```diff
@@ -160,10 +144,33 @@ export default class LineItem {
...
+  onUpdate() {
+    const val = new BigNumber(this.raw_unit_price ?? this.unit_price)
+
+    this.unit_price = val.numeric
+    this.raw_unit_price = val.raw as BigNumberRawValue
+  }
+
+  @OnLoad()
+  onLoad() {
+    this.unit_price = new BigNumber(this.raw_unit_price)
```
riqwan: wouldn't this be undefined on load?

packages/cart/src/models/line-item.ts
```diff
@@ -104,11 +105,16 @@ export default class LineItem {
...
+  compare_at_unit_price?: BigNumber | number | null = null
 
-  @Property({ columnType: "numeric", serializer: Number })
-  @Check({ expression: "unit_price >= 0" }) // TODO: Validate that numeric types work with the expression
-  unit_price: number
+  @Property({ columnType: "jsonb", nullable: true })
+  raw_compare_at_unit_price: BigNumberRawValue | null = null
+
+  @Property({ columnType: "numeric" })
+  unit_price: BigNumber | number
```
riqwan, line 118: Not important, but wonder if we can do something like:

```
@CustomProperty(BigNumber)
@Property({ columnType: "numeric" })
unit_price: BigNumber | number
```

and it takes care of the rest of the stuff

packages/types/src/cart/common.ts
```diff
@@ -349,7 +366,7 @@ export interface CartLineItemDTO {
   /**
    * The price of the item
    */
-  unit_price: number
+  unit_price: BigNumber
```
riqwan: q: wdyt about external types being bound to string/number and the module handles bignumber internally?

riqwan (commented):
looks pretty clean and easy to work with!

packages/cart/integration-tests/__tests__/services/cart-module/index.spec.ts
```diff
@@ -38,6 +38,62 @@ describe("Cart Module Service", () => {
...
+              title: "one",
+              quantity: 1,
+              // proof of backward compatibility
+              unit_price: 100,
+            },
+            {
+              title: "two",
+              quantity: 1,
+              // raw price creation
+              unit_price: { value: "1234.1234" },
```
srindom: q: why can't I just do `unit_price: 1234.1234`?

packages/types/src/cart/common.ts
srindom: +1

srindom (commented):
Is the below a correct understanding of what's happening here?

The CartModule will allow line items to store unit_prices down to fractions of the lowest currency unit. E.g., a fraction of a cent, yen, penny, etc.

This could be relevant in industries that deal with items that are priced with such precision - e.g., gasoline, metered usage of a service, some b2b industries, etc. 

The fractional values are only used to ensure correct precision in totals calculations. I.e. if you have:

- 4023 x Item A price of 0.3 ~ $0.003
- 123 x Item B price of 1.3 ~ $0.013
- Tax rate of 19%

The item totals would end up being:
- Item A - 4023 * 0.3 = 1206.9 -> final charge of $12.07 - this would not be possible to charge in current scheme as lowest unit is 1 cent.
- Item B - 123 * 1.3 = 159.9 -> final charge of $1.60 - this would have been $1.23 without the opportunity to charge fractional cent.

---

One thing that is ofc quite important with this approach is that we also need to support this in the pricing module.

packages/cart/integration-tests/__tests__/services/cart-module/index.spec.ts
carlos-r-l-rodrigues: You can, like the example above and below that.
The BigNumber supports all these formats:

```
unit_price = 1234.1234
unit_price = "1234.1234"
unit_price = {
  value: "1234.1234"
}
unit_price = {
  value: "1234.1234",
  precision: 18,
  currency: "DOGE",
  exchange_rate: {
    "USD": "0.0753"
  }
}
```

then you always have the property `.numeric` and `.raw`
obviouly, when we have more properties than the value, we'll always set an object to it.

olivermrbl:
> The CartModule will allow line items to store unit_prices down to fractions of the lowest currency unit. E.g., a fraction of a cent, yen, penny, etc.

Yes, the numeric Postgres type already allows us to store plenty of digits before and after the decimal point, so we could actually store these fractions already. However, because of precision issues with floating point numbers we can't perform accurate calculations. This PR introduces an abstraction around the price concept, such that we can pull the raw value from the price and convert it into a `BigNumber` from `bignumber.js` in our totals utilities to perform the calculations while still allowing developers to consume the (in this case) `unit_price` as they are used to.

In the future, we could include additional information in the `raw_value`, such as what specific precision should be used.

packages/cart/integration-tests/__tests__/services/cart-module/index.spec.ts
olivermrbl: This was just to demonstrate the different ways of setting the price. You can indeed do that

olivermrbl:
We could get away with not having the abstraction, but this approach allows us to bake some stuff into the BigNumber in the future. 

For example, we could also just keep the two values as strings. Deserialize the value to a number, and only convert the raw value to a BigNumber (from string) when we need it.

packages/utils/src/totals/big-number.ts
```diff
@@ -0,0 +1,99 @@
...
+      /**
+       * Example: const unitPrice = "1234.1234"
+       */
+      const bigNum = new BigNumberJS(rawPrice)
+
+      this.numeric_ = bigNum.toNumber()
+      this.raw_ = this.raw_ = {
+        value: bigNum.toPrecision(BigNumber.DEFAULT_PRECISION),
+      }
+    } else if (isObject(rawPrice)) {
```
olivermrbl: todo: Create `isBigNumberRawValue`

packages/utils/src/totals/big-number.ts
```diff
@@ -0,0 +1,99 @@
...
+import { isObject, isString } from "../common"
+
+export class BigNumber {
+  static DEFAULT_PRECISION = 20
+
+  private numeric_: number
+  private raw_?: BigNumberRawValue
+
+  constructor(rawPrice: BigNumberRawValue | number | string | BigNumberJS) {
+    if (BigNumberJS.isBigNumber(rawPrice)) {
```
olivermrbl, line 16: todo: Extract constructor logic to separate method

packages/utils/src/totals/index.ts
```diff
@@ -0,0 +1,242 @@
...
+    original_tax_total: originalTaxTotal.toNumber(),
+    raw_original_tax_total: originalTaxTotal,
+  }
+
+  const isTaxInclusive = context.includeTax ?? shippingMethod.is_tax_inclusive
+
+  if (isTaxInclusive) {
+    const subtotal = BigNumberJs(shippingMethod.subtotal).minus(taxTotal)
+    totals.subtotal = subtotal.toNumber()
+    totals.raw_subtotal = subtotal
```
olivermrbl: todo: Early return

olivermrbl: todo: Extract `isTaxInclusive` and `isTaxExclusive` logic to separate methods.

packages/cart/src/models/line-item.ts
olivermrbl: I like the idea. However, while we are still working on the foundation, I would like to avoid introducing too many abstractions

packages/utils/src/totals/index.ts
```diff
@@ -0,0 +1,250 @@
...
+      {
+        includeTax,
+      }
+    )
+  }
+
+  return shippingMethodsTotals
+}
+
+export function getShippingMethodTotals_(
```
olivermrbl, line 60: comment: This is the old approach. Please look at `getTotalsForSingleLineItem` for an update one.

packages/cart/src/models/line-item.ts
```diff
@@ -144,10 +148,28 @@ export default class LineItem {
   @BeforeCreate()
   onCreate() {
     this.id = generateEntityId(this.id, "cali")
+
+    const val = new BigNumber(this.raw_unit_price ?? this.unit_price)
+
+    this.unit_price = val.numeric
+    this.raw_unit_price = val.raw!
```
carlos-r-l-rodrigues, line 155: I believe we can create a util to do that for 1 or N fields.

`modelToBigNumber(this, "unit_price")`
or
`modelToBigNumber(this, ["unit_price", "quantity"])` 

subliminal suggestion to set quantity as BigNumber 😆

packages/utils/src/totals/index.ts
```diff
@@ -0,0 +1,250 @@
...
+  return itemsTotals
+}
+
+function getTotalsForSingleLineItem(
+  item: GetItemTotalInput,
+  context: GetLineItemTotalsContext
+) {
+  const unitPrice = BigNumberJs(
+    item.unit_price.raw?.value ?? item.unit_price.numeric
+  )
```
carlos-r-l-rodrigues: a util to handle that might facilitate:

```
const [
  unitPrice,
  originalTaxTotal,
  taxTotal,
  quantity
] = toBigNumberJs(item, "unit_price", "original_tax_total", "tax_total", "quantity")
```

carlos-r-l-rodrigues (commented):
Amazing 🔥

carlos-r-l-rodrigues (approved):

packages/utils/src/totals/to-big-number-js.ts
```diff
@@ -0,0 +1,24 @@
...
+export function toBigNumberJs<T, V extends string>(
+  entity: InputEntity<T, V>,
+  fields: V[]
+): Output<V> {
+  return fields.reduce((acc, field: string) => {
+    if (isDefined(entity[field])) {
+      const entityField = entity[field]
+
+      const val = "raw" in entityField ? entityField.raw.value : entityField
+      acc[field] = new BigNumberJs(val)
```
carlos-r-l-rodrigues: maybe we can use the `toCamelCase` to transform the property name? so we don't use variables with snake case.

packages/cart/src/models/line-item.ts
olivermrbl: That would make the `raw_` prefix a strict requirement right? Otherwise, I am not sure how we can safely do `this.raw_... ?? this....`

PR #6255, opened 2024-01-29
feat(): added compute actions for buyget promotions

riqwan:
what:

- computes actions for buyget promotion (RESOLVES CORE-1700)


packages/promotion/integration-tests/__tests__/services/promotion-module/compute-actions.spec.ts
```diff
@@ -38,44 +38,40 @@ describe("Promotion Service: computeActions", () => {
...
+          },
+        },
+        items: [
+          {
+            id: "item_cotton_tshirt",
+            quantity: 1,
+            unit_price: 100,
+            product_category: {
+              id: "catg_cotton",
+            },
```
srindom, line 55: thought(non-blocking): would probably change this shape to be `product_categories: [{...}]` 

We should do this in a different API but let's add a task to settle on the final shape of the inputs to the promotion module.

packages/promotion/integration-tests/__tests__/services/promotion-module/compute-actions.spec.ts
```diff
@@ -2315,4 +2311,366 @@ describe("Promotion Service: computeActions", () => {
...
+              values: ["VIP", "top100"],
+            },
+          ],
+          application_method: {
+            type: "fixed",
+            target_type: "items",
+            allocation: "each",
+            max_quantity: 1,
+            apply_to_quantity: 1,
+            buy_rules_min_quantity: 1,
```
srindom, line 2377: thought: would want this to be part of the buy_rules instead. e.g., `{ attribute: "quantity", operator: "gt", values: 1 }`. All buy rules would then be evaluated with an `AND` operator on for each item. Wdyt of this idea?

packages/promotion/integration-tests/__tests__/services/promotion-module/compute-actions.spec.ts
riqwan: it works both ways already. You can pass product_categories: [] or product_category: {}. We've only added very few definition restrictions [here](https://github.com/medusajs/medusa/blob/develop/packages/types/src/promotion/common/compute-actions.ts#L64C1-L65C1). 

Let me know if thats not what you meant

packages/promotion/integration-tests/__tests__/services/promotion-module/compute-actions.spec.ts
riqwan: I think this might be quite limiting when we want to expand cases. It would only allow one BuyRule for item targeting and will also result in some relationship validations as min_qty is a required field. 

I think there are 2 cases here:

1. each rule needs to conform to a grouped quantity
ex: buy a cotton shirt and a silk shirt and get a sweater for free
or
ex: buy 2 cotton shirt and get a sweater for free

In this case, we need to use an OR operator.

```
buy_rules_type: "across",
buy_rules_min_quantity: 2,
buy_rules: [{
  attribute: "product_category.id",
  operator: "eq",
  values: ["catg_cotton"],
}, {
  attribute: "product_category.id",
  operator: "eq",
  values: ["catg_silk"],
}],
```

2. each rule conforms to its own min_quantity
ex: buy 2 cotton tshirts and 1 silk shirt and get a sweater free.


In this case, we need to use an AND operator.

```
buy_rules_type: "each",
buy_rules_min_quantity: null,
buy_rules: [{
  attribute: "product_category.id",
  operator: "eq",
  values: ["catg_cotton"],
  min_quantity: 2 
}, {
  attribute: "product_category.id",
  operator: "eq",
  values: ["catg_silk"],
  min_quantity: 1 
}],
```

wdyt?

srindom: That makes sense! Kinda hard to wrap your head around, but it makes sense - would be amazing if this could self-document somehow.

srindom (approved):
LGTM!

packages/promotion/integration-tests/__tests__/services/promotion-module/compute-actions.spec.ts
riqwan: Currently only case 1 is implemented. When implementing case 2, I can rework the code to make it more easily readable with some inline comments.

PR #6256, opened 2024-01-29
chore(cart): Clean up data models

olivermrbl:
Clean up data models in Cart Module according to latest established conventions


packages/cart/src/models/address.ts
```diff
@@ -20,37 +19,37 @@ export default class Address {
...
   @Property({ columnType: "text", nullable: true })
-  postal_code?: string | null
+  postal_code: string | null = null
 
   @Property({ columnType: "text", nullable: true })
-  phone?: string | null
+  phone: string | null = null
 
   @Property({ columnType: "jsonb", nullable: true })
   metadata?: Record<string, unknown> | null
```
riqwan:
```suggestion
  metadata: Record<string, unknown> | null = null
```

missed a spot

packages/cart/src/models/shipping-method.ts
```diff
@@ -24,18 +24,18 @@ export default class ShippingMethod {
   @Property({ columnType: "text" })
   cart_id: string
 
-  @ManyToOne(() => Cart, {
-    onDelete: "cascade",
+  @ManyToOne({
+    entity: () => Cart,
     index: "IDX_shipping_method_cart_id",
-    nullable: true,
+    cascade: [Cascade.REMOVE, Cascade.PERSIST],
```
riqwan, line 30: q: shouldn't we be cascading soft deletes on relations?

riqwan (approved):
lgtm!

packages/cart/src/models/shipping-method.ts
olivermrbl: Yes, I will need to make the models soft-deletable, but currently awaiting Adrien's, that will create all related methods under the hood.

PR #6258, opened 2024-01-29
fix(medusa): add support for `order` field on GET /admin/orders

kasperkristensen:
**What**
- Adds support for `order` field on `GET /admin/orders`


srindom (approved):
LGTM

PR #6100, opened 2024-01-16
docs: migrate guides to TSDoc references

shahednasser:
_Edit: Turning this into a PR holding all left migration of guides to references._

## Important

This PR includes changes to some plugins and abstract classes constructors. Check the changesets for more details.

---

This PR includes changes from the PRs #6121 , #6116 , #6115 + the following changes:

### Tax Provider

- Move documentation closer to code for tax-provider guide.
- Add typedoc configuration to generate the reference
- Generate the references

### Notification Provider

- Move documentation closer to code for notification-provider guide.
- Add typedoc configuration to generate the reference
- Fix issue in some return types not having description
- Changed typedoc sort order to be source-file based.
- Generate the references

### File Service

- Move documentation closer to code for file-service guide.
- Add typedoc configuration to generate the reference.
- Fix issue in return field duplicating the same field.
- Generate the references.

### Search Service

- Move documentation closer to code for search-service guide.
- Add typedoc configuration to generate the reference.
- Generate the references.

### Medusa Configurations

- Move documentation closer to code
- Add typedoc configuration to generate the reference
- Added support in the typedoc theme to expand properties within the pag
- Generate the references

PR #6259, opened 2024-01-29
fix(medusa): fix constructor container type for abstract services

shahednasser:
As discussed with @adrien2p before, the type of the container in the constructor should be `Record` rather than `MedusaContainer`. This also solves the issue mentioned in [this comment](https://github.com/medusajs/medusa/pull/6100#discussion_r1469711913)


olivermrbl:
I believe what we pass is the `cradle` from Awilix, so maybe it would make sense to note that. Wdyt?

PR #6262, opened 2024-01-29
feat(cart): `GET /store/carts/:id`

olivermrbl:


packages/medusa/src/joiner-configs/cart-service.ts
```diff
@@ -4,7 +4,7 @@ import { ModuleJoinerConfig } from "@medusajs/types"
 import { Cart } from "../models"
 
 export default {
-  serviceName: Modules.CART,
+  serviceName: "cartService",
```
srindom, line 7: q: is this meant to be like this?

srindom (approved):
LGTM w. comments - think the line I commented needs to be reverted? Otherwise good to go

packages/medusa/src/joiner-configs/cart-service.ts
olivermrbl: `Modules.CART` was introduced recently by mistake. These joiner configs are from when we attempted to add joiner configs for internal services, so this should apply to `cartService` as far as I can understand

PR #6263, opened 2024-01-29
chore: local workflow proxying methods to pass context

carlos-r-l-rodrigues:
What:
- When calling a module's method inside a Local Workflow the MedusaContext is passed as the last argument to the method if not provided
- Add `requestId` to req
- A couple of fixes on Remote Joiner and the data fetcher for internal services

Why:
- The context used to initialize the workflow has to be shared with all modules. properties like transactionId will be used to emit events and requestId to trace logs for example.


packages/orchestration/src/workflow/local-workflow.ts
```diff
@@ -60,7 +66,52 @@ export class LocalWorkflow {
       }
     }
 
-    this.container = container
+    this.container = this.contextualizedMedusaModules(container)
+  }
+
+  private contextualizedMedusaModules(container) {
```
olivermrbl, line 72: would be good to have @adrien2p's eye on this one

packages/utils/src/modules-sdk/decorators/inject-manager.ts
```diff
@@ -16,13 +16,28 @@ export function InjectManager(managerProperty?: string): MethodDecorator {
     const argIndex = target.MedusaContextIndex_[propertyKey]
 
     descriptor.value = function (...args: any[]) {
-      const context: SharedContext | Context = { ...(args[argIndex] ?? {}) }
+      const originalContext = args[argIndex] ?? {}
+      const copiedContext = {} as Context
+      for (const key in originalContext) {
+        if (key === "manager" || key === "transactionManager") continue
```
olivermrbl: nit: use curly

packages/utils/src/modules-sdk/decorators/inject-transaction-manager.ts
```diff
@@ -42,8 +43,22 @@ export function InjectTransactionManager(
         : this[managerProperty]
       ).transaction(
         async (transactionManager) => {
-          args[argIndex] = { ...(args[argIndex] ?? {}) }
-          args[argIndex].transactionManager = transactionManager
+          const copiedContext = {} as Context
+          for (const key in originalContext) {
+            if (key === "manager" || key === "transactionManager") continue
```
olivermrbl: nit: use curly

packages/medusa/src/utils/remote-query-fetch-data.ts
```diff
@@ -47,12 +86,12 @@ export function remoteQueryFetchData(container: MedusaContainer) {
...
     const result = await service[methodName](filters, options)
 
-    if (hasPagination) {
+    if (hasPagination_) {
       const [data, count] = result
       return {
         data: {
           rows: data,
-          metadata: {},
+          metadata: buildPagination(options, count),
```
olivermrbl, line 94: question: enables pagination and filters on `remoteQuery`?

olivermrbl (approved):
Just to be sure, this means these two code snippets are functionally identical right?

Explicit context:
```
createStep("some-step", (input, { container, context }) => {
  const moduleService = container.resolve("someModule")
  await moduleService.create(data, context)
})
```
Magical context
```ts
createStep("some-step", (input, { container }) => {
  const moduleService = container.resolve("someModule")
  await moduleService.create(data)
})
```

In the latter case, the context is passed as if we were doing it explicitly.

packages/medusa/src/utils/remote-query-fetch-data.ts
carlos-r-l-rodrigues: These are to correct identify pagination for internal services not loaded as modules.
To behave the same as it does for modules.

carlos-r-l-rodrigues:
> Just to be sure, this means these two code snippets are functionally identical right?

That is correct.

packages/medusa/src/utils/remote-query-fetch-data.ts
```diff
@@ -1,6 +1,23 @@
 import { MedusaModule, RemoteQuery } from "@medusajs/modules-sdk"
 import { MedusaContainer } from "@medusajs/types"
 
+function hasPagination(options: { [attr: string]: unknown }): boolean {
+  if (!options) {
+    return false
+  }
+
+  const attrs = ["skip"]
+  return Object.keys(options).some((key) => attrs.includes(key))
```
adrien2p, line 10: q: shouldn't we check for either skip or take?

packages/orchestration/src/joiner/remote-joiner.ts
```diff
@@ -338,7 +338,11 @@ export class RemoteJoiner {
       relationship
     )
     const isObj = isDefined(response.path)
-    const resData = isObj ? response.data[response.path!] : response.data
+    let resData = isObj ? response.data[response.path!] : response.data
+
+    if (!Array.isArray(resData)) {
+      resData = [resData]
+    }
```
adrien2p:
```suggestion
resData = Array.isArray(resData) ? resData : [resData]
```

packages/orchestration/src/workflow/local-workflow.ts
```diff
@@ -60,7 +66,52 @@ export class LocalWorkflow {
...
+                }
+
+                return await target[prop].apply(target, [...args])
+              }
+            },
+          })
+        }
+      },
+    })
   }
```
adrien2p:
```suggestion
 function contextualizedMedusaModules(container) {
  const medusaContext_ = this.medusaContext
  const originalResolver = container.resolve
  container.resolve = function (registrationName, opts) {
    const resolved = originalResolver(registrationName, opts)
    if (resolved?.constructor?.__type !== MedusaModuleType) {
      return resolved
    }

    return new Proxy(resolved, {
      get: function (target, prop) {
        if (typeof target[prop] !== "function") {
          return target[prop]
        }

        return async (...args) => {
          const ctxIndex =
            MedusaContext.getIndex(target, prop as string) ?? args.length - 1

          const hasContext = args[ctxIndex]?.__type === MedusaContextType

          if (!hasContext) {
            const context = medusaContext_

            if (context?.__type === MedusaContextType) {
              delete context?.manager
              delete context?.transactionManager
              args[ctxIndex] = context
            }
          }

          return await target[prop].apply(target, [...args])
        }
      },
    })
  }
}
```

packages/utils/src/decorators/inject-entity-manager.ts
```diff
@@ -1,5 +1,7 @@
 import { Context, SharedContext } from "@medusajs/types"
+import { MedusaContextType } from "../modules-sdk/decorators"
 
+// @deprecated Use InjectManager instead
```
adrien2p, line 4: we can't deprecate this one, it is only used in the old module and they can't use the new InjectManager because there is no notion of freshManager no? maybe we should just warn to not use it at all and then remove it once the old modules are migrated? such as the inventory and stocjk

packages/utils/src/modules-sdk/decorators/inject-manager.ts
```diff
@@ -16,13 +16,31 @@ export function InjectManager(managerProperty?: string): MethodDecorator {
     const argIndex = target.MedusaContextIndex_[propertyKey]
 
     descriptor.value = function (...args: any[]) {
-      const context: SharedContext | Context = { ...(args[argIndex] ?? {}) }
+      const originalContext = args[argIndex] ?? {}
```
adrien2p, line 19: nice 🙏 we get that back from the search right?

packages/utils/src/modules-sdk/decorators/inject-transaction-manager.ts
```diff
@@ -31,7 +31,8 @@ export function InjectTransactionManager(
     const argIndex = target.MedusaContextIndex_[propertyKey]
     descriptor.value = async function (...args: any[]) {
       const shouldForceTransactionRes = shouldForceTransaction(target)
-      const context: SharedContext | Context = args[argIndex] ?? {}
+      const context: Context = args[argIndex] ?? {}
+      const originalContext = args[argIndex] ?? {}
```
adrien2p, line 36: should we add the __type here too?

packages/utils/src/modules-sdk/decorators/context-parameter.ts
```diff
@@ -8,3 +8,12 @@ export function MedusaContext() {
     target.MedusaContextIndex_[propertyKey] = parameterIndex
   }
 }
+
+MedusaContext.getIndex = function (
```
adrien2p, line 12: should we add the __type line 7 as well?

packages/orchestration/src/workflow/local-workflow.ts
adrien2p: I have also added `const ctxIndex =
            MedusaContext.getIndex(target, prop as string) ?? args.length - 1` cause I believe the index start from 0 right?

packages/utils/src/decorators/inject-entity-manager.ts
carlos-r-l-rodrigues: we won't use it anymore and we must replace these modules asap. I don't see any issue flagging it as deprecated.

packages/utils/src/modules-sdk/decorators/inject-transaction-manager.ts
carlos-r-l-rodrigues: yeah. good catch

packages/utils/src/modules-sdk/decorators/context-parameter.ts
carlos-r-l-rodrigues: Not here. That is a static property added to the class, which holds the position of the Context argument in each method.

packages/orchestration/src/workflow/local-workflow.ts
carlos-r-l-rodrigues: yeah, better.

packages/medusa/src/utils/remote-query-fetch-data.ts
carlos-r-l-rodrigues: no because you can limit the results but not paginate them. skip indicates you are paginating it.

packages/utils/src/decorators/inject-entity-manager.ts
adrien2p: there is no issue, but you can't use InjectManager in the old module, I just want to be sure that people internally start to change it for InjectManager in the old module before migrating them

packages/utils/src/modules-sdk/decorators/context-parameter.ts
adrien2p: yes you are right my bad

olivermrbl:
Should we merge this one?

carlos-r-l-rodrigues:
> Should we merge this one?

I think so.
@adrien2p?

adrien2p:
I would say yes 🙏

PR #6265, opened 2024-01-30
feat(auth): add authentication endpoints

pKorsholm:
**What**
- Add authentication endpoints: 
  - `/auth/[scope]/[provider]` 
  - `/auth/[scope]/[provider]/callback`
- update authenticate-middleware handler
- Add scope field to user
- Add unique constraint on scope and entity_id

note: there's still some remaining work related to jwt auth to be handled, this is mainly focussed on session auth with endpoints


packages/auth/src/providers/email-password.ts
```diff
@@ -16,6 +16,14 @@ class EmailPasswordProvider extends AbstractAuthModuleProvider {
     this.authUserSerivce_ = authUserService
   }
 
+  private getHashConfig(scope: string) {
+    const scopeConfig = this.scopes_[scope].hashConfig as
```
riqwan, line 24: This is a lot of trust on the chaining 😄 

suggestion: throw error or silently fail when scope is not found

packages/auth/src/providers/email-password.ts
```diff
@@ -16,6 +16,14 @@ class EmailPasswordProvider extends AbstractAuthModuleProvider {
...
+      | Scrypt.ScryptParams
+      | undefined
+
+    return scopeConfig ?? { logN: 15, r: 8, p: 1 }
```
riqwan: question: can we add a descriptive variable for the objet or a comment on what that is?

packages/auth/src/providers/email-password.ts
```diff
@@ -34,11 +42,40 @@ class EmailPasswordProvider extends AbstractAuthModuleProvider {
...
-    )
+    let authUser
+
+    try {
+      authUser = await this.authUserSerivce_.retrieveByProviderAndEntityId(
+        email,
+        EmailPasswordProvider.PROVIDER
+      )
+    } catch (error) {
+      if (error.type === MedusaError.Types.NOT_FOUND) {
```
riqwan, line 60: question: wouldn't this try to create an authUser when an incorrect email is added?

packages/medusa/src/api-v2/auth/[scope]/[authProvider]/callback/route.ts
```diff
@@ -0,0 +1,36 @@
...
+import { MedusaError } from "@medusajs/utils"
+
+export const GET = async (req: MedusaRequest, res: MedusaResponse) => {
+  const { scope, authProvider } = req.params
+
+  const service: IAuthModuleService = req.scope.resolve(
+    ModuleRegistrationName.AUTH
+  )
+
+  const authData = { ...req } as unknown as AuthenticationInput
```
Comment (riqwan) [inline packages/medusa/src/api-v2/auth/[scope]/[authProvider]/callback/route.ts]: suggestion: should we scope the required attributes here since this is a fairly large object to clone?

packages/medusa/src/api-v2/store/customers/route.ts
```diff
@@ -1,14 +1,37 @@
...
+import { createCustomerAccountWorkflow } from "@medusajs/core-flows"
 
 export const POST = async (req: MedusaRequest, res: MedusaResponse) => {
+  if (req.auth_user?.app_metadata?.customer_id) {
+    const remoteQuery = req.scope.resolve(
+      ContainerRegistrationKeys.REMOTE_QUERY
+    )
+    const query = {
+      customer: {
+        __args: { id: req.session.auth_user.app_metadata.customer_id },
```
riqwan:
```suggestion
        __args: { id: req.auth_user?.app_metadata?.customer_id },
```
since you checked for existence for this already

packages/auth/src/models/auth-user.ts
```diff
@@ -34,11 +35,14 @@ export default class AuthUser {
...
   @Property({ columnType: "jsonb", nullable: true })
   user_metadata: Record<string, unknown> | null
 
-  @Property({ columnType: "jsonb", nullable: true })
-  app_metadata: Record<string, unknown> | null
+  @Property({ columnType: "jsonb" })
+  app_metadata: Record<string, unknown> = {}
 
   @Property({ columnType: "jsonb", nullable: true })
   provider_metadata: Record<string, unknown> | null
```
riqwan:
```suggestion
  provider_metadata: Record<string, unknown> | null = null
```

packages/medusa/src/api-v2/auth/[scope]/[authProvider]/callback/route.ts
```diff
@@ -0,0 +1,37 @@
...
+  authData.authScope = scope
+
+  const authResult = await service.validateCallback(authProvider, authData)
+
+  const { success, error, authUser, location } = authResult
+  if (location) {
+    res.redirect(location)
```
Comment (riqwan) [inline packages/medusa/src/api-v2/auth/[scope]/[authProvider]/callback/route.ts:27]: question: shouldn't the client be the one doing the redirect?

packages/medusa/src/api-v2/auth/[scope]/[authProvider]/route.ts
```diff
@@ -0,0 +1,37 @@
+import { AuthenticationInput, IAuthModuleService } from "@medusajs/types"
+import { MedusaRequest, MedusaResponse } from "../../../../types/routing"
+
+import { MedusaError } from "@medusajs/utils"
+import { ModuleRegistrationName } from "@medusajs/modules-sdk"
+
+export const GET = async (req: MedusaRequest, res: MedusaResponse) => {
+  const { scope, authProvider } = req.params
```
Comment (riqwan) [inline packages/medusa/src/api-v2/auth/[scope]/[authProvider]/route.ts:8]: suggestion: Maybe a middleware to validate the scope and provider so that the route is not open ended?

packages/auth/src/providers/email-password.ts
pKorsholm: yes, but that's what you would expect to happen if you create an account with a wrong email on any site

packages/auth/src/providers/email-password.ts
pKorsholm: we do that before going to the provider in the authentication or validateCallback module methods when we verify the provider here: 
```
 protected getRegisteredAuthenticationProvider(
    provider: string,
    { authScope }: AuthenticationInput
  ): AbstractAuthModuleProvider {
    let containerProvider: AbstractAuthModuleProvider
    try {
      containerProvider = this.__container__[`auth_provider_${provider}`]
    } catch (error) {
      ...
    }

    containerProvider.validateScope(authScope)

    return containerProvider
  }
```

we call the abstract class method: 
```
public validateScope(scope) {
    if (!this.scopes_[scope]) {
      throw new MedusaError(
        MedusaError.Types.INVALID_ARGUMENT,
        `Scope "${scope}" is not valid for provider ${this.provider}`
      )
    }
  }
```

packages/medusa/src/api-v2/auth/[scope]/[authProvider]/callback/route.ts
Comment (pKorsholm) [reply packages/medusa/src/api-v2/auth/[scope]/[authProvider]/callback/route.ts]: yes, we should 😅 now we should be able to serialize the request as well 🎉

packages/medusa/src/api-v2/auth/[scope]/[authProvider]/callback/route.ts
```diff
@@ -0,0 +1,42 @@
...
+  }
+
+  if (!success && error) {
+    throw new MedusaError(MedusaError.Types.UNAUTHORIZED, error)
+  } else if (success) {
+    req.session.auth_user = authUser
+    req.session.scope = authUser.scope
+
+    res.status(200).json({ authUser })
+  }
```
Comment (adrien2p) [inline packages/medusa/src/api-v2/auth/[scope]/[authProvider]/callback/route.ts]: q: if it is not success and there is no error then the request gets stuck here, should we have something?

packages/medusa/src/api-v2/auth/[scope]/[authProvider]/callback/route.ts
```diff
@@ -0,0 +1,42 @@
...
+    headers: req.headers,
+    query: req.query,
+    body: req.body,
+    authScope: scope,
+  } as AuthenticationInput
+
+  const authResult = await service.validateCallback(authProvider, authData)
+
+  const { success, error, authUser, location } = authResult
+  if (location) {
```
Comment (adrien2p) [inline packages/medusa/src/api-v2/auth/[scope]/[authProvider]/callback/route.ts:26]: if there is a location (the redirect could be configured) then we don't check any success or failure?

packages/medusa/src/api-v2/store/customers/route.ts
adrien2p: are the optional chaining necessary since the i already check its existence? I know in the latest TS version they have improved this

packages/medusa/src/api-v2/store/customers/route.ts
```diff
@@ -1,14 +1,37 @@
...
+import { ContainerRegistrationKeys } from "@medusajs/utils"
 import { CreateCustomerDTO } from "@medusajs/types"
+import { createCustomerAccountWorkflow } from "@medusajs/core-flows"
 
 export const POST = async (req: MedusaRequest, res: MedusaResponse) => {
+  if (req.auth_user?.app_metadata?.customer_id) {
+    const remoteQuery = req.scope.resolve(
+      ContainerRegistrationKeys.REMOTE_QUERY
+    )
+    const query = {
```
adrien2p: suggestion: should we start using `remoteQueryObjectFromString` everywhere to normalize and that way we don't have to change everywhere when we change the property names

```ts
const query = remoteQueryObjectFromString({ entryPoint: "customer", variables: { id: ... }, fields: [])
```

adrien2p (commented):
few questions :D

packages/medusa/src/api-v2/auth/[scope]/[authProvider]/callback/route.ts
Comment (pKorsholm) [reply packages/medusa/src/api-v2/auth/[scope]/[authProvider]/callback/route.ts:26]: yes, we could validate to ensure that `success` is required if you think that's appropriate? 😄 just figured that in case of a redirect you would want that no matter what

Comment (adrien2p) [reply packages/medusa/src/api-v2/auth/[scope]/[authProvider]/callback/route.ts:26]: in general you have a success redirect and a failure redirect that eventually could have custom query parameters, let see that later

packages/medusa/src/api-v2/auth/[scope]/[authProvider]/callback/route.ts
Comment (pKorsholm) [reply packages/medusa/src/api-v2/auth/[scope]/[authProvider]/callback/route.ts]: "flipped" the condition to always throw unless `success` is true

packages/auth/src/providers/email-password.ts
riqwan: ah fair, I thought this would be done for a login

PR #6267, opened 2024-01-30
feat(customer): Add create and retrieve customer from store side

srindom:
**What**
- GET /store/customers/me
- POST /store/customers
- Workflow for customer account creation
- Authentication middleware on customer routes


packages/auth/src/migrations/Migration20240122041959.ts
```diff
@@ -1,22 +1,30 @@
...
-
-    this.addSql('create table if not exists "auth_user" ("id" text not null, "entity_id" text not null, "provider_id" text null, "user_metadata" jsonb null, "app_metadata" jsonb null, "provider_metadata" jsonb null, constraint "auth_user_pkey" primary key ("id"));');
-    this.addSql('alter table "auth_user" add constraint "IDX_auth_user_provider_entity_id" unique ("provider_id", "entity_id");');
-
-    this.addSql('alter table "auth_user" add constraint if not exists "auth_user_provider_id_foreign" foreign key ("provider_id") references "auth_provider" ("provider") on delete cascade;');
+    this.addSql(
+      'create table if not exists "auth_provider" ("provider" text not null, "name" text not null, "domain" text check ("domain" in (\'all\', \'store\', \'admin\')) not null default \'all\', "config" jsonb null, "is_active" boolean not null default false, constraint "auth_provider_pkey" primary key ("provider"));'
```
srindom, line 6: domain should be open ended - and should probably be renamed to scope. Or maybe we need to find an entirely different term for it - maybe audience or auth group.

packages/auth/src/migrations/Migration20240122041959.ts
```diff
@@ -1,22 +1,30 @@
...
+    )
+
+    this.addSql(
+      'create table if not exists "auth_user" ("id" text not null, "entity_id" text not null, "provider_id" text null, "user_metadata" jsonb null, "app_metadata" jsonb null, "provider_metadata" jsonb null, constraint "auth_user_pkey" primary key ("id"));'
```
srindom, line 10: Auth user should probably have a `scope` column too.

packages/auth/src/migrations/Migration20240122041959.ts
```diff
@@ -1,22 +1,30 @@
...
+    )
+    this.addSql(
+      'alter table "auth_user" add constraint "IDX_auth_user_provider_entity_id" unique ("provider_id", "entity_id");'
```
srindom, line 13: there should be a unique constraint across scope too - maybe the provider_id holds a notion of scope already?

packages/auth/src/services/auth-module.ts
```diff
@@ -118,13 +133,50 @@ export default class AuthModuleService<
...
+    return jwt.sign({ id: authUser.id, scope }, this.options_.jwt_secret, {
+      expiresIn: "1d",
+    })
+  }
+
+  async retrieveAuthUserFromJwtToken(
+    token: string,
+    scope: string
+  ): Promise<AuthUserDTO> {
+    let decoded: AuthJWTPayload
```
srindom: Propose these to generate and validate jwt tokens - wdyt?

packages/medusa/src/utils/authenticate-middleware.ts
```diff
@@ -0,0 +1,79 @@
...
+  auth: {
+    [authScope: string]: {
+      user_id: string
+    }
+  }
+}
+
+type AuthType = "session" | "bearer"
+
+export default (
```
srindom: propose something like this instead of using passport custom strategies.

srindom (commented):
@pKorsholm - took the liberty to construct the /store/customer endpoints and build an auth middleware. There are some differences between your approach and mine - let's take a look at it together and settle on a path forward.

packages/medusa/src/api-v2/store/customers/middlewares.ts
```diff
@@ -0,0 +1,22 @@
+import { transformBody } from "../../../api/middlewares"
+import { MiddlewareRoute } from "../../../loaders/helpers/routing/types"
+import { StorePostCustomersReq } from "./validators"
+import authenticate from "../../../utils/authenticate-middleware"
+
+export const storeCustomerRoutesMiddlewares: MiddlewareRoute[] = [
+  {
+    method: "ALL",
+    matcher: "/store/customers*",
+    middlewares: [authenticate("store", ["session", "bearer"])],
```
srindom, line 11: Think this middleware is really nice to work with - you can imagine other users picking this up for custom routes:

e.g.:
```
import { authenticate } from "@medusajs/medusa"

// my route stuff that only allows session based auth for agent endpoints
... authenticate("agent", ["session"]) ...
```

packages/medusa/src/utils/authenticate-middleware.ts
pKorsholm: really like the api of this middleware, though, is there a reason you wan't to avoid using passport for handling middlewares? 😄

packages/core-flows/src/auth/steps/set-auth-app-metadata.ts
```diff
@@ -0,0 +1,59 @@
...
+import { ModuleRegistrationName } from "@medusajs/modules-sdk"
+import { isDefined } from "@medusajs/utils"
+
+type StepInput = {
+  authUserId: string
+  key: string
+  value: string
+}
+
+export const setAuthAppMetadata = "set-auth-app-metadata"
```
olivermrbl:
```suggestion
export const setAuthAppMetadataStepId = "set-auth-app-metadata"
```

olivermrbl (approved):
LGTM

PR #6272, opened 2024-01-30
feat(types): added computed actions for automatic promotions

riqwan:
what:

- compute actions account for automatic promotions (RESOLVES CORE-1701)

RESOLVES CORE-1689


packages/promotion/src/services/promotion-module.ts
```diff
@@ -231,6 +231,15 @@ export default class PromotionModuleService<
       PromotionTypes.ComputeActionAdjustmentLine
     >()
     const methodIdPromoValueMap = new Map<string, number>()
+    const automaticPromotions = preventAutoPromotions
+      ? []
+      : await this.list({ is_automatic: true }, { select: ["code"] })
```
srindom, line 236: thought: we might want to do something more clever than this at some point - imagine a store with thousands of automatic discounts. This would introduce a lot of overhead.

srindom (approved):
LGTM!

packages/promotion/src/services/promotion-module.ts
riqwan: Yes, agree. I'll create a ticket for revisiting this.

PR #6273, opened 2024-01-30
feat(cart): `POST /store/carts`

olivermrbl:
Depends on #6262


olivermrbl:
Would like us to consider changing to `zod` for API body validation: https://github.com/medusajs/medusa/compare/feat/cm-create-cart-zod?expand=1


More specifically, [this file](https://github.com/medusajs/medusa/blob/8ddddf9b28df8336c0a2a6ae6f6edf9fd47efdfd/packages/medusa/src/api-v2/store/carts/validators.ts) demonstrates how the schemas will look like with zod.

packages/medusa/src/types/routing.ts
srindom, line 1: nit: linting is off here

srindom (approved):
lgtm - w. a small nit.

+1 from my side on zod - the class-validator stuff is awfully hard to dance with.

PR #6274, opened 2024-01-30
feat(cart): `POST /store/carts/:id`

olivermrbl:
Depends on:
- #6262 
- #6273


integration-tests/plugins/__tests__/cart/store/update-cart.ts
```diff
@@ -0,0 +1,61 @@
+import { ModuleRegistrationName } from "@medusajs/modules-sdk"
```
adrien2p, line 1: should we add a changeset?

integration-tests/plugins/__tests__/cart/store/update-cart.ts
```diff
@@ -0,0 +1,61 @@
...
+    const response = await api.post(`/store/carts/${cart.id}`, {
+      email: "tony@stark.com",
+    })
+
+    expect(response.status).toEqual(200)
+    expect(response.data.cart).toEqual(
+      expect.objectContaining({
+        id: cart.id,
+        currency_code: "usd",
+        email: "tony@stark.com",
```
adrien2p, line 59: suggestion(nit): to simplify copy and update I would
move that to a const body
```ts
{
      email: "tony@stark.com",
    }
```

and here use body.email

packages/cart/src/services/cart-module.ts
```diff
@@ -190,35 +190,62 @@ export default class CartModuleService<
...
-    data: CartTypes.UpdateCartDTO,
+    selector: Partial<CartTypes.CartDTO>,
+    data: CartTypes.UpdateCartDataDTO,
     sharedContext?: Context
-  ): Promise<CartTypes.CartDTO>
+  ): Promise<CartTypes.CartDTO[]>
 
   @InjectManager("baseRepository_")
   async update(
-    data: CartTypes.UpdateCartDTO[] | CartTypes.UpdateCartDTO,
```
adrien2p, line 205: q: should we use InjectTransactionManager? also, maybe split in that case to update and update_ in order to not mess with populate within a transaction

packages/cart/src/types/cart.ts
```diff
@@ -14,12 +14,12 @@ export interface CreateCartDTO {
...
-  sales_channel_id?: string
-  email?: string
-  currency_code?: string
-  metadata?: Record<string, unknown>
+  region_id?: string | null
+  customer_id?: string | null
+  sales_channel_id?: string | null
+  email?: string | null
+  currency_code?: string | null
+  metadata?: Record<string, unknown> | null
```
adrien2p, line 22: suggestion: To represent the DTO, I wouldn't set them as optional, this should be handled automatically based on the select and relations but not here. If you fetch the cart, you will receive the curren prop marked as optional with a null value but they will be there unless configured differently

packages/core-flows/src/definition/cart/steps/update-carts.ts
```diff
@@ -0,0 +1,61 @@
...
+  async (previousCarts, { container }) => {
+    if (!previousCarts?.length) {
+      return
+    }
+
+    const service = container.resolve<ICartModuleService>(
+      ModuleRegistrationName.CART
+    )
+
+    const toRestore = previousCarts.map((c) => ({
```
adrien2p, line 49: todo: I think that we are missing the relations here, so it would partially work right?

packages/core-flows/src/definition/cart/workflows/update-carts.ts
```diff
@@ -0,0 +1,20 @@
+import {
+  CartDTO,
+  FilterableCartProps,
+  UpdateCartDataDTO,
+} from "@medusajs/types"
+import { WorkflowData, createWorkflow } from "@medusajs/workflows-sdk"
+import { updateCartsStep } from "../steps/update-carts"
+
+type WorkflowInput = {
```
adrien2p, line 9: suggestion: interface, plus same question/comment as for phil pr's in terms of location

packages/medusa/src/api-v2/store/carts/[id]/route.ts
```diff
@@ -17,3 +20,23 @@ export const GET = async (req: MedusaRequest, res: MedusaResponse) => {
 
   res.json({ cart })
 }
+
+export const POST = async (req: MedusaRequest, res: MedusaResponse) => {
+  const updateCartWorkflow = updateCartsWorkflow(req.scope)
```
Comment (adrien2p) [inline packages/medusa/src/api-v2/store/carts/[id]/route.ts:25]: suggestion: Normally you should be able to skip that step and pass the container in the run

packages/types/src/cart/mutations.ts
```diff
@@ -171,6 +174,11 @@ export interface UpdateLineItemWithSelectorDTO {
   data: Partial<UpdateLineItemDTO>
 }
 
+export interface UpdateCartWithSelectorDTO {
+  selector: Partial<CartDTO>
+  data: Partial<UpdateCartDTO>
```
adrien2p: q: would that mean the id is updatable>?

adrien2p (approved):
LGTM with a few comments/suggestions

packages/cart/src/types/cart.ts
olivermrbl: But this is an UpdateDTO. Isn't that a different case than what  you are describing?

packages/core-flows/src/definition/cart/steps/update-carts.ts
olivermrbl: I think we are yet to figure out a good approach for restoring the correct relations no? Or have we settled on a pattern? Because it would depend on what you pass right

packages/cart/src/types/cart.ts
adrien2p: My bad, I mixed it with the cart dto, discard

packages/core-flows/src/definition/cart/steps/update-carts.ts
adrien2p: we haven't discussed that at all so far. but we will need to think about it indeed

olivermrbl: Created a ticket for it

PR #6275, opened 2024-01-30
feat(modules-sdk): External Modules

carlos-r-l-rodrigues:
What:
 - Enable External Modules config
 - Serve MedusaApp as http or grpc servers (Experimental Only. No TLS)
 - http and gRPC clients to communicate with MedusaApp server
 
```typescript
// Server
import { MedusaApp, Modules } from "@medusajs/modules-sdk";

async function main() {
  const app = await MedusaApp({
    sharedResourcesConfig: {
      database: {
        clientUrl: "postgres://postgres:postgres@localhost:5432/medusa",
      },
    },
    modulesConfig: {
      [Modules.PRODUCT]: true,
    },
  });
  app.listen("http", 8000).then(() => {
    console.log("MedusaApp Server is running on port 8000");
  });
  app.listen("grpc", 8001).then(() => {
    console.log("Medusa gRPC Server is running on port 8001");
  });
}
main();
```

```typescript
// Client
import { MODULE_SCOPE, MedusaApp, Modules } from "@medusajs/modules-sdk";
import { IProductModuleService } from "@medusajs/types";

async function main() {
  const { modules, query } = await MedusaApp({
    sharedResourcesConfig: {
      database: {
        clientUrl: "postgres://postgres:postgres@localhost:5432/medusa",
      },
    },
    modulesConfig: {
      [Modules.PRODUCT]: {
        scope: MODULE_SCOPE.EXTERNAL,
        server: {
          type: "grpc",
          url: "localhost:8001",
        },
      },
    },
  });

  const productService =
    modules.productService as unknown as IProductModuleService;

  await productService.create(
    [
      {
        title: "Product " + Math.random(),
        description: "Product 1",
      },
    ]
  );

  const products = await query({
    product: {
      fields: ["id", "title", "description"],
    },
  });

  console.log(products);
}
main();
```

### DIFF
--- a/packages/core/modules-sdk/src/loaders/register-modules.ts
+++ b/packages/core/modules-sdk/src/loaders/register-modules.ts
@@ -1,4 +1,5 @@
 import {
+  CustomModuleDefinition,
   ExternalModuleDeclaration,
   InternalModuleDeclaration,
   ModuleDefinition,
@@ -22,12 +23,13 @@ export const registerMedusaModule = (
 ): Record<string, ModuleResolution> => {
   const moduleResolutions = {} as Record<string, ModuleResolution>
 
-  const isCustomModule = !definition && !ModulesDefinition[moduleKey]
+  const isCustomModule = !definition?.key && !ModulesDefinition[moduleKey]
 
   let customModuleConfig
   if (isCustomModule) {
     customModuleConfig = getCustomModuleResolution(
       moduleKey,
+      definition,
       moduleDeclaration as Partial<
         InternalModuleDeclaration | ExternalModuleDeclaration
       >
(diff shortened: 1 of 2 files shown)


.changeset/clean-pumpkins-move.md
```diff
@@ -0,0 +1,6 @@
+---
+"@medusajs/modules-sdk": patch
```
adrien2p, line 2: todo: to update

packages/modules-sdk/build.js
```diff
@@ -0,0 +1,7 @@
+const fs = require("fs")
```
adrien2p: q: don't the tsconfig include it? IMO it should already be in the dist after build no>?

packages/modules-sdk/package.json
```diff
@@ -40,8 +40,15 @@
...
+    "build": "rimraf dist && tsc --build && node build.js",
     "test": "jest --runInBand --bail --forceExit",
     "watch": "tsc --build --watch"
+  },
+  "optionalDependencies": {
+    "@fastify/compress": "^7.0.0",
+    "@grpc/grpc-js": "^1.9.14",
+    "@grpc/proto-loader": "^0.7.10",
+    "fastify": "^4.25.2",
+    "undici": "^6.4.0"
```
adrien2p: ❤️

adrien2p (approved):
Love it LGTM, I think we need to play a bit in a more context real like to have a better idea of improvements

packages/modules-sdk/build.js
carlos-r-l-rodrigues: nope. I couldn't make typescript to include the `.proto` file to be copied.
Apparently there isn't any simple config for that.

.changeset/clean-pumpkins-move.md
carlos-r-l-rodrigues: sorry, I didn't get it. to update what?

riqwan (approved):
This is v. cool, Carlos!

If we have both servers/dbs in the same network, this would actually be a very cool & quick way for systems to communicate with each other or build on top of each other. 

I think listing out some use-cases with this or a quick POC maybe with Strapi and Medusa might be able to add more weight to this concept. 💪🏻

PR #6276, opened 2024-01-30
feat(customer): add customer group customer management

srindom:
**What**
- GET /admin/customer-groups/:id/customers
- POST /admin/customer-groups/:id/customers/batch
- POST /admin/customer-groups/:id/customers/remove

Workflows


fPolic (approved):
LGTM

PR #6293, opened 2024-02-01
feat(user): Init user module

pKorsholm:
**What**
- Add blank slate for user module


packages/types/src/user/service.ts
```diff
@@ -0,0 +1,35 @@
...
+import { CreateUserDTO, UpdateUserDTO } from "./mutations"
+
+export interface IUserModuleService extends IModuleService {
+  retrieveUser(
+    id: string,
+    config?: FindConfig<UserDTO>,
+    sharedContext?: Context
+  ): Promise<UserDTO>
+
+  listUsers(
```
carlos-r-l-rodrigues: These are methods to retrieve and list from the main entity, I believe it is fine to keep them without the suffix (User).
`retrieve, list, listAndCount`

adrien2p: Yes indeed, like the auto generated methods will be. I suggest you to glance over my pr description for the abstraction, i added all those conventions as well 👍

olivermrbl: We should probably await your PR before merging this one?

packages/types/src/user/mutations.ts
```diff
@@ -0,0 +1,2 @@
+export type CreateUserDTO = { id?: string }
+export type UpdateUserDTO = { id: string }
```
adrien2p: suggestioin: Dont forget to move to interfaces later 👍

packages/user/src/migrations/Migration20240201081925.ts
```diff
@@ -0,0 +1,13 @@
+import { Migration } from '@mikro-orm/migrations';
+
+export class Migration20240201081925 extends Migration {
+
+  async up(): Promise<void> {
+    this.addSql('create table "user" ("id" text not null, constraint "user_pkey" primary key ("id"));');
```
adrien2p, line 6: comment: will probably change a lot later, but not to forget the if exists and dissociate the table creation and the columns

adrien2p (approved):
LGTM 🔥

packages/user/src/migrations/Migration20240201081925.ts
pKorsholm: yes, will update the table definitions in a subsequent pr (including `if exists`)

packages/user/integration-tests/__fixtures__/user.ts
```diff
@@ -0,0 +1,16 @@
+import { SqlEntityManager } from "@mikro-orm/postgresql"
```
srindom, line 1: thought(non-blocking): this approach of having one big seeder feels a bit off - think creating the data for the test within the test case is easier to wrap you your head around + you can use the service methods directly.

srindom (approved):
LGTM

packages/user/src/migrations/Migration20240201081925.ts
srindom: maybe worth leaving the migration out entirely for now

packages/user/integration-tests/__fixtures__/user.ts
pKorsholm: Will update when introducing models today 😄

PR #6295, opened 2024-02-01
feat(customer): manage default address selection

srindom:
**What**
- Catches unique constraints on customer_id, is_default_billing/is_default_shipping and reformats
- Adds an step to create and update of addresses that unsets the previous default shipping/billing address if necessary.
  - This creates a behavior in the API where you can always set an address to be default and it will automatically unset the previous one for you.


packages/customer/src/services/customer-module.ts
```diff
@@ -775,4 +782,30 @@ export default class CustomerModuleService implements ICustomerModuleService {
         )
       : void 0
   }
+
+  private async flush(context: Context) {
```
olivermrbl, line 786: thought: This could live as a MikroORM util no? It only depends on the context

olivermrbl (approved):
LGTM, I like the unsetting pattern

packages/customer/src/services/customer-module.ts
srindom: yes it could - let's leave it here until we meet the pattern in other places

PR #6297, opened 2024-02-01
feat(dashboard): DataTable component

kasperkristensen:
**What**
- Adds DataTable component.
- Updates to latest version of TW to gain access to new selectors and classes.

**Why**
- The DataTable component contains all logic for adding a new Table component to the admin. It allows for free text search, filtering, ordering and pagination. While this PR introduces a lot of new code for the component it will allow us to reduce the bundle size of the Admin, as we can replace all existing tables, and missing tables with this single component. It is also meant to help prevent us from running into the same problem we had in the previous version of Admin were each developer would have their own way of implementing a Table, resulting in code that was hard to maintain, and was using multiple versions of dependencies.
- The components query state is all handled via search params in the URL, making it possible to share links with for specific views of a table. If there are multiple tables on a page, a `prefix` can be passed to the component to differentiate the params for each table.
- The component also introduces a more responsive Table component, that allows users to scroll horizontally. When scrolling horizontally the first columns will be fixed, making it easier for users to keep track of which row they are looking at. If the row has a Select column (Checkbox) then the checkbox + the next columns will be fixed.

**Note**
- This PR only adds the DataTable component on the Order list page, and Customer details page, for demonstration. The rest of admin will be updated in subsequent PRs.

### DIFF
--- a/packages/admin-next/dashboard/src/components/table/data-table/data-table-filter/select-filter.tsx
+++ b/packages/admin-next/dashboard/src/components/table/data-table/data-table-filter/select-filter.tsx
@@ -104,7 +104,7 @@ export const SelectFilter = ({
             }
           }}
         >
-          <Command>
+          <Command className="h-full">
             {searchable && (
               <div className="border-b p-1">
                 <div className="grid grid-cols-[1fr_20px] gap-x-2 rounded-md px-2 py-1">
@@ -137,7 +137,7 @@ export const SelectFilter = ({
                 {t("general.noResultsTitle")}
               </span>
             </Command.Empty>
-            <Command.Group className="overflow-auto p-1 outline-none">
+            <Command.List className="h-full max-h-[163px] min-h-[0] overflow-auto p-1 outline-none">
               {options.map((option) => {
                 const isSelected = selectedParams
                   .get()
@@ -166,7 +166,7 @@ export const SelectFilter = ({
                   </Command.Item>
                 )
               })}
-            </Command.Group>
+            </Command.List>
           </Command>
         </Popover.Content>
       </Popover.Portal>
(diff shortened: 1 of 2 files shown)


packages/admin-next/dashboard/src/components/data-table/data-table-root/data-table-root.tsx
```diff
@@ -0,0 +1,272 @@
...
+/**
+ * TODO
+ *
+ * Add a sticky header to the table that shows the column name when scrolling through the table vertically.
+ *
+ * This is a bit tricky as we can't support horizontal scrolling and sticky headers at the same time, natively
+ * with CSS. We need to implement a custom solution for this. One solution is to render a duplicate table header
+ * using a DIV that, but it will require rerendeing the duplicate header every time the window is resized, to keep
+ * the columns aligned.
+ */
```
kasperkristensen: comment: This will be implemented in a follow up PR, most likely after the first iteration of Admin 3.0.

packages/admin-next/dashboard/src/hooks/table/query/use-order-table-query.tsx
```diff
@@ -0,0 +1,58 @@
...
+import { useQueryParams } from "../../use-query-params"
+
+type UseOrderTableQueryProps = {
+  prefix?: string
+  pageSize?: number
+}
+
+/**
+ * TODO: Enable `order` query param when staging is updated
+ */
```
kasperkristensen, line 11: comment: Currently the Order menu does nothing for the Order tables. Will enable it once staging has been updated, as it would currently crash the page.

olivermrbl (approved):
Really, really awesome work. Should we reduce the pagesize to 20, now that we know it doesn't look silly on larger screens? Think it makes for a better "whole" product experience.

olivermrbl:
Also, please don't merge before #6188 is in :)

srindom:
This is really pretty!

We should re-add the customer information in the order overview though

PR #6300, opened 2024-02-01
feat(): added percentage calculations for all cases

riqwan:
what:

- adds missing percentage calculations for items and shipping methods

### DIFF
--- a/packages/promotion/src/utils/compute-actions/items.ts
+++ b/packages/promotion/src/utils/compute-actions/items.ts
@@ -68,7 +68,8 @@ export function applyPromotionToItems(
         method.quantity,
         applicationMethod?.max_quantity!
       )
-      const totalItemValue = method.unit_price * quantityMultiplier
+      const totalItemValue =
+        (method.subtotal / method.quantity) * quantityMultiplier
       let promotionValue = parseFloat(applicationMethod!.value!)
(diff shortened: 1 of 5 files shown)


packages/promotion/src/utils/compute-actions/items.ts
```diff
@@ -67,10 +68,14 @@ export function applyPromotionToItems(
...
-      const applicableTotal =
-        method.unit_price * quantityMultiplier - appliedPromoValue
+      const totalItemValue = method.unit_price * quantityMultiplier
+      let promotionValue = parseFloat(applicationMethod!.value!)
+      const applicableTotal = totalItemValue - appliedPromoValue
+
+      if (applicationMethod?.type === ApplicationMethodType.PERCENTAGE) {
+        promotionValue = (promotionValue / 100) * applicableTotal
+      }
+
```
srindom: thought: would it make sense to move all this responsibility outside of the promotion module - instead the promotion module would expect `subtotal` fields on the items it gets passed?

srindom: todo: use `method.subtotal / method.quantity` instead of `method.unit_price`.

srindom (changes requested):
Unit price fix and we should be good to go

srindom (approved):
LGTM!

PR #6308, opened 2024-02-02
feat(payment): provider service

fPolic:
**What**
- introduce payment provider service
- refactor module methods that communicate with providers

**Note**
- please ignore TODOs related to amounts - this depends on the #6253

### DIFF
--- a/packages/generated/client-types/src/lib/models/AdminGetOrdersParams.ts
+++ b/packages/generated/client-types/src/lib/models/AdminGetOrdersParams.ts
@@ -156,7 +156,7 @@ export interface AdminGetOrdersParams {
    */
   fields?: string
   /**
-   * A order field to sort-order the retrieved orders by.
+   * Field to sort retrieved orders by.
    */
   order?: string
 }
--- a/packages/medusa-react/src/contexts/medusa.tsx
+++ b/packages/medusa-react/src/contexts/medusa.tsx
@@ -100,7 +100,8 @@ export interface MedusaProviderProps {
    */
   customHeaders?: Record<string, any>
   /**
-   * An instance of the Medusa JS Client. If you don't provide an instance, one will be created using the `baseUrl`, `apiKey`, `publishableApiKey`, `maxRetries`, and `customHeaders` props.
+   * An instance of the Medusa JS Client. If you don't provide an instance, one will be created using the `baseUrl`, `apiKey`, 
+   * `publishableApiKey`, `maxRetries`, and `customHeaders` props.
    */
   medusaClient: Medusa
 }
--- a/packages/medusa/src/api/routes/admin/orders/list-orders.ts
+++ b/packages/medusa/src/api/routes/admin/orders/list-orders.ts
@@ -152,7 +152,7 @@ import { cleanResponseData } from "../../../../utils/clean-response-data"
  *   - (query) limit=50 {integer} Limit the number of orders returned.
  *   - (query) expand {string} Comma-separated relations that should be expanded in the returned order.
  *   - (query) fields {string} Comma-separated fields that should be included in the returned order.
- *   - (query) order {string} A order field to sort-order the retrieved orders by.
+ *   - (query) order {string} Field to sort retrieved orders by.
  * x-codegen:
  *   method: list
  *   queryParams: AdminGetOrdersParams
@@ -280,7 +280,8 @@ export class AdminGetOrdersParams extends AdminListOrdersSelector {
   fields?: string
 
   /**
-   * The field to sort the data by. By default, the sort order is ascending. To change the order to descending, prefix the field name with `-`.
+   * The field to sort retrieved orders by. By default, the sort order is ascending.
+   * To change the order to descending, prefix the field name with `-`.
    */
   @IsOptional()
   @IsString()
--- a/packages/types/src/common/config-module.ts
+++ b/packages/types/src/common/config-module.ts
@@ -7,12 +7,12 @@ import {
 
 /**
  * @interface
- * 
+ *
  * Options to pass to `express-session`.
  */
 type SessionOptions = {
   /**
-   * The name of the session ID cookie to set in the response (and read from in the request). The default value is `connect.sid`. 
+   * The name of the session ID cookie to set in the response (and read from in the request). The default value is `connect.sid`.
    * Refer to [express-session’s documentation](https://www.npmjs.com/package/express-session#name) for more details.
    */
   name?: string
@@ -27,7 +27,7 @@ type SessionOptions = {
    */
   rolling?: boolean
   /**
-   * Whether a session that is "uninitialized" is forced to be saved to the store. The default value is `true`. 
+   * Whether a session that is "uninitialized" is forced to be saved to the store. The default value is `true`.
    * Refer to [express-session’s documentation](https://www.npmjs.com/package/express-session#saveUninitialized) for more details.
    */
   saveUninitialized?: boolean
@@ -45,7 +45,7 @@ type SessionOptions = {
 
 /**
  * @interface
- * 
+ *
  * HTTP compression configurations.
  */
 export type HttpCompressionOptions = {
@@ -54,17 +54,17 @@ export type HttpCompressionOptions = {
    */
   enabled?: boolean
   /**
-   * The level of zlib compression to apply to responses. A higher level will result in better compression but will take longer to complete. 
+   * The level of zlib compression to apply to responses. A higher level will result in better compression but will take longer to complete.
    * A lower level will result in less compression but will be much faster. The default value is `6`.
    */
   level?: number
   /**
-   * How much memory should be allocated to the internal compression state. It's an integer in the range of 1 (minimum level) and 9 (maximum level). 
+   * How much memory should be allocated to the internal compression state. It's an integer in the range of 1 (minimum level) and 9 (maximum level).
    * The default value is `8`.
    */
   memLevel?: number
   /**
-   * The minimum response body size that compression is applied on. Its value can be the number of bytes or any string accepted by the 
+   * The minimum response body size that compression is applied on. Its value can be the number of bytes or any string accepted by the
    * [bytes](https://www.npmjs.com/package/bytes) module. The default value is `1024`.
    */
   threshold?: number | string
@@ -72,36 +72,36 @@ export type HttpCompressionOptions = {
 
 /**
  * @interface
- * 
+ *
  * Essential configurations related to the Medusa backend, such as database and CORS configurations.
  */
 export type ProjectConfigOptions = {
   /**
    * The Medusa backend’s API Routes are protected by Cross-Origin Resource Sharing (CORS). So, only allowed URLs or URLs matching a specified pattern can send requests to the backend’s API Routes.
-   * 
-   * `store_cors` is a string used to specify the accepted URLs or patterns for store API Routes. It can either be one accepted origin, or a comma-separated list of accepted origins. 
-   * 
+   *
+   * `store_cors` is a string used to specify the accepted URLs or patterns for store API Routes. It can either be one accepted origin, or a comma-separated list of accepted origins.
+   *
    * Every origin in that list must either be:
-   * 
+   *
    * 1. A URL. For example, `http://localhost:8000`. The URL must not end with a backslash;
    * 2. Or a regular expression pattern that can match more than one origin. For example, `.example.com`. The regex pattern that the backend tests for is `^([\/~@;%#'])(.*?)\1([gimsuy]*)$`.
-   * 
+   *
    * @example
    * Some example values of common use cases:
-   * 
+   *
    * ```bash
    * # Allow different ports locally starting with 800
    * STORE_CORS=/http:\/\/localhost:800\d+$/
-   * 
+   *
    * # Allow any origin ending with vercel.app. For example, storefront.vercel.app
    * STORE_CORS=/vercel\.app$/
-   * 
+   *
    * # Allow all HTTP requests
    * STORE_CORS=/http:\/\/.+/
    * ```
-   * 
+   *
    * Then, set the configuration in `medusa-config.js`:
-   * 
+   *
    * ```js title="medusa-config.js"
    * module.exports = {
    *   projectConfig: {
@@ -111,9 +111,9 @@ export type ProjectConfigOptions = {
    *   // ...
    * }
    * ```
-   * 
+   *
    * If you’re adding the value directly within `medusa-config.js`, make sure to add an extra escaping `/` for every backslash in the pattern. For example:
-   * 
+   *
    * ```js title="medusa-config.js"
    * module.exports = {
    *   projectConfig: {
@@ -127,30 +127,30 @@ export type ProjectConfigOptions = {
   store_cors?: string
   /**
    * The Medusa backend’s API Routes are protected by Cross-Origin Resource Sharing (CORS). So, only allowed URLs or URLs matching a specified pattern can send requests to the backend’s API Routes.
-   * 
-   * `admin_cors` is a string used to specify the accepted URLs or patterns for admin API Routes. It can either be one accepted origin, or a comma-separated list of accepted origins. 
-   * 
+   *
+   * `admin_cors` is a string used to specify the accepted URLs or patterns for admin API Routes. It can either be one accepted origin, or a comma-separated list of accepted origins.
+   *
    * Every origin in that list must either be:
-   * 
+   *
    * 1. A URL. For example, `http://localhost:7001`. The URL must not end with a backslash;
    * 2. Or a regular expression pattern that can match more than one origin. For example, `.example.com`. The regex pattern that the backend tests for is `^([\/~@;%#'])(.*?)\1([gimsuy]*)$`.
-   * 
+   *
    * @example
    * Some example values of common use cases:
-   * 
+   *
    * ```bash
    * # Allow different ports locally starting with 700
    * ADMIN_CORS=/http:\/\/localhost:700\d+$/
-   * 
+   *
    * # Allow any origin ending with vercel.app. For example, admin.vercel.app
    * ADMIN_CORS=/vercel\.app$/
-   * 
+   *
    * # Allow all HTTP requests
    * ADMIN_CORS=/http:\/\/.+/
    * ```
-   * 
+   *
    * Then, set the configuration in `medusa-config.js`:
-   * 
+   *
    * ```js title="medusa-config.js"
    * module.exports = {
    *   projectConfig: {
@@ -160,9 +160,9 @@ export type ProjectConfigOptions = {
    *   // ...
    * }
    * ```
-   * 
+   *
    * If you’re adding the value directly within `medusa-config.js`, make sure to add an extra escaping `/` for every backslash in the pattern. For example:
-   * 
+   *
    * ```js title="medusa-config.js"
    * module.exports = {
    *   projectConfig: {
@@ -176,10 +176,10 @@ export type ProjectConfigOptions = {
   admin_cors?: string
   /**
    * A random string used to create cookie tokens. Although this configuration option is not required, it’s highly recommended to set it for better security.
-   * 
-   * In a development environment, if this option is not set, the default secret is `supersecret` However, in production, if this configuration is not set, an error is thrown and 
+   *
+   * In a development environment, if this option is not set, the default secret is `supersecret` However, in production, if this configuration is not set, an error is thrown and
    * the backend crashes.
-   * 
+   *
    * @example
    * ```js title="medusa-config.js"
    * module.exports = {
@@ -196,10 +196,10 @@ export type ProjectConfigOptions = {
 
   /**
    * A random string used to create authentication tokens. Although this configuration option is not required, it’s highly recommended to set it for better security.
-   * 
+   *
    * In a development environment, if this option is not set the default secret is `supersecret` However, in production, if this configuration is not set an error, an
    * error is thrown and the backend crashes.
-   * 
+   *
    * @example
    * ```js title="medusa-config.js"
    * module.exports = {
@@ -216,15 +216,15 @@ export type ProjectConfigOptions = {
 
   /**
    * The name of the database to connect to. If specified in `database_url`, then it’s not required to include it.
-   * 
-   * Make sure to create the PostgreSQL database before using it. You can check how to create a database in 
+   *
+   * Make sure to create the PostgreSQL database before using it. You can check how to create a database in
    * [PostgreSQL's documentation](https://www.postgresql.org/docs/current/sql-createdatabase.html).
-   * 
+   *
    * @example
    * ```js title="medusa-config.js"
    * module.exports = {
    *   projectConfig: {
-   *     database_database: process.env.DATABASE_DATABASE || 
+   *     database_database: process.env.DATABASE_DATABASE ||
    *       "medusa-store",
    *     // ...
    *   },
@@ -236,30 +236,30 @@ export type ProjectConfigOptions = {
 
   /**
    * The connection URL of the database. The format of the connection URL for PostgreSQL is:
-   * 
+   *
    * ```bash
    * postgres://[user][:password]@[host][:port]/[dbname]
    * ```
-   * 
+   *
    * Where:
-   * 
+   *
    * - `[user]`: (required) your PostgreSQL username. If not specified, the system's username is used by default. The database user that you use must have create privileges. If you're using the `postgres` superuser, then it should have these privileges by default. Otherwise, make sure to grant your user create privileges. You can learn how to do that in [PostgreSQL's documentation](https://www.postgresql.org/docs/current/ddl-priv.html).
    * - `[:password]`: an optional password for the user. When provided, make sure to put `:` before the password.
    * - `[host]`: (required) your PostgreSQL host. When run locally, it should be `localhost`.
    * - `[:post]`: an optional port that the PostgreSQL server is listening on. By default, it's `5432`. When provided, make sure to put `:` before the port.
    * - `[dbname]`: (required) the name of the database.
-   * 
+   *
    * You can learn more about the connection URL format in [PostgreSQL’s documentation](https://www.postgresql.org/docs/current/libpq-connect.html).
-   * 
+   *
    * @example
    * For example, set the following database URL in your environment variables:
-   * 
+   *
    * ```bash
    * DATABASE_URL=postgres://postgres@localhost/medusa-store
    * ```
-   * 
+   *
    * Then, use the value in `medusa-config.js`:
-   * 
+   *
    * ```js title="medusa-config.js"
    * module.exports = {
    *   projectConfig: {
@@ -273,11 +273,11 @@ export type ProjectConfigOptions = {
   database_url?: string
   /**
    * The database schema to connect to. This is not required to provide if you’re using the default schema, which is `public`.
-   * 
+   *
    * ```js title="medusa-config.js"
    * module.exports = {
    *   projectConfig: {
-   *     database_schema: process.env.DATABASE_SCHEMA || 
+   *     database_schema: process.env.DATABASE_SCHEMA ||
    *       "custom",
    *     // ...
    *   },
@@ -289,13 +289,13 @@ export type ProjectConfigOptions = {
 
   /**
    * This configuration specifies what database messages to log. Its value can be one of the following:
-   * 
+   *
    * - (default) A boolean value that indicates whether any messages should be logged.
    * - The string value `all` that indicates all types of messages should be logged.
    * - An array of log-level strings to indicate which type of messages to show in the logs. The strings can be `query`, `schema`, `error`, `warn`, `info`, `log`, or `migration`. Refer to [Typeorm’s documentation](https://typeorm.io/logging#logging-options) for more details on what each of these values means.
-   * 
+   *
    * If this configuration isn't set, its default value is `false`, meaning no database messages are logged.
-   * 
+   *
    * @example
    * ```js title="medusa-config.js"
    * module.exports = {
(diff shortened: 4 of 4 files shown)


olivermrbl:
Be careful about this PR when merging. It seems to have been based off another already merged branch.

fPolic:
> Be careful about this PR when merging. It seems to have been based off another already merged branch.

Should be resolved now

olivermrbl:
As discussed previously, I think we should go with making all methods that interface with 3rd party providers singular and handle the bulk operations (and the errors that may occur) separately in the API or workflows. This should mitigate the risk of unexpected behavior or an inconsistent state of payments between systems. That way, we can handle an issue with one payment operation in isolation from the others instead of having to deal with scenarios like reconciling payments because 4 captures succeeded in Stripe but none in Medusa.

To illustrate, this would look something like this:

Payment Module
```ts
async capturePayment(paymentId) {
  const payment = await this.retrievePayment(paymentId)
  const provider = await this.retrieveProvider(payment.provider_id)

  await provider.capturePayment(payment)

  return await paymentService.update({ captured_at: new Date() })	
}
```

API
```ts
// POST /admin/payment-collection/:id/capture
const route = (req, res) => {
  const workflow = capturePaymentsWorkflow(req.scope)

  const paymentCollection = await paymentModule.retrieve(req.query.id)

  const result = await workflow.run({ input: paymentCollection }, { throwOnError: false })

  // handle errors … e.g. one out of 5 captures might have succeeded
}
```

Or alternatively, just do five requests to a single API endpoint
```ts
// POST /admin/payments/:id/capture
const route = (req, res) => {
  const workflow = capturePaymentWorkflow(req.scope)

  const payment = await paymentModule.retrievePayment(req.query.id)

  const result = workflow.run({ input: p })

  // handle errors
}
```

Aside from making it less error-prone, it would also simply the Payment Module business logic quite a bit.

cc @adrien2p @carlos-r-l-rodrigues

fPolic:
@olivermrbl I agree! Converting this temporarily to a draft while I work on a refactor.

packages/payment/integration-tests/__tests__/services/payment-module/index.spec.ts
```diff
@@ -9,14 +12,134 @@ import {
...
+      service = await initialize({
+        database: {
+          clientUrl: DB_URL,
+          schema: process.env.MEDUSA_PAYMNET_DB_SCHEMA,
+        },
+      })
+
+      if (service.__hooks?.onApplicationStart) {
+        await service.__hooks.onApplicationStart()
+      }
```
adrien2p: todo: we should not do that and only use the initModules above, can you update that please. You can look what we have done in the region fif you need too

packages/payment/integration-tests/__tests__/services/payment-module/index.spec.ts
```diff
@@ -9,14 +12,134 @@ import {
...
+
+    afterAll(async () => {
+      await shutdownFunc()
+    })
+
+    beforeEach(async () => {
+      await MikroOrmWrapper.setupDatabase()
+      repositoryManager = await MikroOrmWrapper.forkManager()
+
+      service = await initialize({
```
adrien2p: suggestion: you might also take the time to remove the initialize function and maybe you can apply the new migration generator in the local module config

packages/payment/integration-tests/__tests__/services/payment-module/index.spec.ts
```diff
@@ -46,6 +169,10 @@ describe("Payment Module Service", () => {
         },
       })
 
+      if (service.__hooks?.onApplicationStart) {
+        await service.__hooks.onApplicationStart()
+      }
```
adrien2p: comment: same comment as above everywhere

packages/payment/src/models/payment-session.ts
```diff
@@ -33,8 +33,8 @@ export default class PaymentSession {
   @Property({ columnType: "text" })
   provider_id: string
 
-  @Property({ columnType: "jsonb", nullable: true })
-  data: Record<string, unknown> | null = null
+  @Property({ columnType: "jsonb" })
+  data: Record<string, unknown> = {}
```
adrien2p, line 37: q: what pushes you to do that?

packages/payment/src/services/payment-module.ts
```diff
@@ -160,305 +176,405 @@ export default class PaymentModuleService<
...
+    return this.retrievePayment(payment.id, {}, sharedContext)
   }
-  setPaymentSessions(
-    paymentCollectionId: string,
-    data: SetPaymentSessionsDTO[],
-    sharedContext?: Context | undefined
-  ): Promise<PaymentCollectionDTO> {
-    throw new Error("Method not implemented.")
+
+  private async createProvidersOnLoad() {
```
adrien2p: question: should we move that to a loader like oli did for the region? you have access to the services in the container too

packages/payment/src/services/payment-module.ts
```diff
@@ -95,13 +100,24 @@ export default class PaymentModuleService<
...
+  __hooks = {
+    onApplicationStart: async () => await this.createProvidersOnLoad(),
+  }
+
+  /**
+   *
+   * ********** PAYMENT COLLECTION **********
+   *
+   */
+
```
adrien2p: todo: pluralize methods that are bulk, it is part of the convention document, I let you have a look

packages/payment/src/services/payment-module.ts
```diff
@@ -95,13 +100,24 @@ export default class PaymentModuleService<
...
 
+  __hooks = {
+    onApplicationStart: async () => await this.createProvidersOnLoad(),
+  }
+
+  /**
+   *
+   * ********** PAYMENT COLLECTION **********
+   *
+   */
```
adrien2p: q: do we really need those block comments? since this is included in the method name I would say that it should be self explanatory

packages/payment/src/services/payment-module.ts
```diff
@@ -160,305 +176,405 @@ export default class PaymentModuleService<
...
-  ): Promise<PaymentDTO | PaymentDTO[]> {
-    const input = Array.isArray(data) ? data : [data]
+  @InjectTransactionManager("baseRepository_")
+  async updatePaymentSession(
+    data: UpdatePaymentSessionDTO,
+    @MedusaContext() sharedContext?: Context
+  ): Promise<PaymentSessionDTO> {
+    const session = await this.paymentSessionService_.retrieve(
+      data.id,
+      {},
```
adrien2p: todo: probably a good idea to add a select to not load everything from the entity. It is a good practice in general

packages/payment/src/services/payment-module.ts
```diff
@@ -160,305 +176,405 @@ export default class PaymentModuleService<
...
+  async deletePaymentSession(
+    id: string,
     @MedusaContext() sharedContext?: Context
-  ): Promise<Payment[]> {
-    let payments = await this.paymentService_.list(
-      { id: data.map((d) => d.payment_id) },
+  ): Promise<void> {
+    const session = await this.paymentSessionService_.retrieve(
+      id,
       {},
```
adrien2p: todo: same, config

adrien2p: everywhere it is possible

packages/payment/src/services/payment-module.ts
```diff
@@ -160,305 +176,405 @@ export default class PaymentModuleService<
...
+        authorized_at:
+          status === PaymentSessionStatus.AUTHORIZED ? new Date() : null,
+      },
+      sharedContext
+    )
+
+    if (status !== PaymentSessionStatus.AUTHORIZED) {
+      throw new MedusaError(
+        MedusaError.Types.NOT_ALLOWED,
+        "Session not authorized with the provider."
```
adrien2p: q: should we mention the provider for the log?

packages/payment/src/services/payment-module.ts
```diff
@@ -160,305 +176,405 @@ export default class PaymentModuleService<
...
+        payment_session: session.id,
+        payment_collection: session.payment_collection!.id,
+        provider_id: session.provider_id,
+        // customer_id: context.customer.id,
+        data,
       },
       sharedContext
     )
+
+    return this.retrievePayment(payment.id, {}, sharedContext)
```
adrien2p: todo: missing await

adrien2p: to be checked everywhere

packages/payment/src/services/payment-provider.ts
```diff
@@ -0,0 +1,235 @@
...
+  CreatePaymentInput,
+  PaymentProcessorContext,
+} from "@medusajs/types/src"
+
+type InjectedDependencies = {
+  paymentProviderRepository: DAL.RepositoryService
+}
+
+export default class PaymentProviderService implements IPaymentProviderService {
+  paymentProviderRepository_: DAL.RepositoryService
```
adrien2p: todo: protected readonly

packages/payment/src/services/payment-provider.ts
```diff
@@ -0,0 +1,235 @@
...
+    @MedusaContext() sharedContext?: Context
+  ): Promise<PaymentProvider[]> {
+    return await this.paymentProviderRepository_.create(data, sharedContext)
+  }
+
+  @InjectManager("paymentProviderRepository_")
+  async list(
+    @MedusaContext() sharedContext?: Context
+  ): Promise<PaymentProvider[]> {
+    return await this.paymentProviderRepository_.find(undefined, sharedContext)
```
adrien2p, line 55: q: passing `{}` instead of `undefined` is the same?

packages/payment/src/services/payment-provider.ts
```diff
@@ -0,0 +1,235 @@
...
+    ) {
+      throw new MedusaError(
+        MedusaError.Types.INVALID_ARGUMENT,
+        "`currency_code` and `amount` are required to create payment session."
+      )
+    }
+
+    const paymentResponse = await provider.initiatePayment(sessionInput)
+
+    if ("error" in paymentResponse) {
```
adrien2p: suggestion: use isPaymentProcessorError

packages/payment/src/services/payment-provider.ts
```diff
@@ -0,0 +1,235 @@
...
+      data: Record<string, unknown>
+      provider_id: string
+    },
+    sessionInput: PaymentProcessorContext
+  ): Promise<Record<string, unknown> | undefined> {
+    const provider = this.retrieveProvider(paymentSession.provider_id)
+
+    const paymentResponse = await provider.updatePayment(sessionInput)
+
+    if ((paymentResponse as PaymentProcessorError)?.error) {
```
adrien2p: suggestion: use isPaymentProcessorError

packages/payment/src/services/payment-provider.ts
```diff
@@ -0,0 +1,235 @@
...
+    const { payment_session, provider_id } = data
+    const providerId = provider_id ?? payment_session.provider_id
+
+    const provider = this.retrieveProvider(providerId)
+
+    /**
+     * NOTE: JUST RETRIEVE
+     */
+    const paymentData = await provider.retrievePayment(payment_session.data)
+    if ("error" in paymentData) {
```
adrien2p: comment :everywhere we can we should use isPaymentProcessorError

packages/payment/src/services/payment-provider.ts
```diff
@@ -0,0 +1,235 @@
...
+
+    const res = await provider.authorizePayment(paymentSession.data, context)
+    if ("error" in res) {
+      this.throwFromPaymentProcessorError(res)
+    }
+
+    return {
+      data: (res as PaymentProcessorAuthorizeResponse).data,
+      status: (res as PaymentProcessorAuthorizeResponse).status,
+    }
```
adrien2p:
```suggestion
    const { data, status } = res as PaymentProcessorAuthorizeResponse
    return {
      data,
      status,
    }
```

packages/payment/src/services/payment-provider.ts
```diff
@@ -0,0 +1,235 @@
...
+
+    const res = await provider.refundPayment(payment.data, amount)
+    if (isPaymentProcessorError(res)) {
+      this.throwFromPaymentProcessorError(res as PaymentProcessorError)
+    }
+
+    return res as Record<string, unknown>
+  }
+
+  throwFromPaymentProcessorError(error: PaymentProcessorError) {}
```
adrien2p: todo

adrien2p (commented):
preliminary review 🚀

adrien2p:
As a general rule we should activate our IDE missing await error, there is a lot of missing await in the modules here and there. Also, it would be nice to explain what is being done in the pr and what have been discarded to be done later. It feels like the work is not finished here and not sure what to review or not review in some places :)

packages/payment/src/services/payment-provider.ts
```diff
@@ -0,0 +1,235 @@
...
+import { PaymentProvider } from "@models"
+
+import { isDefined, MedusaError } from "medusa-core-utils"
+import {
+  CreatePaymentInput,
+  PaymentProcessorContext,
+} from "@medusajs/types/src"
+
+type InjectedDependencies = {
+  paymentProviderRepository: DAL.RepositoryService
```
adrien2p, line 26: todo: missing the providers
```ts
[key: `payment_providers_${string}`]: IPaymentProcessor
```

packages/payment/src/loaders/providers.ts
```diff
@@ -3,9 +3,11 @@ import { moduleProviderLoader } from "@medusajs/modules-sdk"
 import { LoaderOptions, ModuleProvider, ModulesSdkTypes } from "@medusajs/types"
 import { Lifetime, asFunction } from "awilix"
 
+import * as providers from "../providers"
+
 const registrationFn = async (klass, container, pluginOptions) => {
   container.register({
-    [`payment_provider_${klass.prototype}`]: asFunction(
+    [`payment_provider_${klass.identifier}`]: asFunction(
```
olivermrbl: comment: This will change slightly in #6311

packages/payment/src/loaders/providers.ts
```diff
@@ -33,6 +35,12 @@ export default async ({
   const pluginProviders =
     options?.providers?.filter((provider) => provider.resolve) || []
 
+  // Local providers
+  for (const provider of Object.values(providers)) {
+    // TODO: pass options
+    await registrationFn(provider, container, {})
+  }
```
olivermrbl: thought: If the only local provider is `system`, I think we could leave this out and just inject the service directly into the payment provider. Similar to Google and EmailPass in the auth module

packages/payment/src/services/payment-module.ts
```diff
@@ -95,13 +100,24 @@ export default class PaymentModuleService<
...
+    this.paymentProviderService_ = paymentProviderService
     this.paymentCollectionService_ = paymentCollectionService
   }
 
   __joinerConfig(): ModuleJoinerConfig {
     return joinerConfig
   }
 
+  __hooks = {
+    onApplicationStart: async () => await this.createProvidersOnLoad(),
```
olivermrbl: comment: This will be moved to a loader in #6311

packages/payment/src/services/payment-module.ts
olivermrbl: comment: Yes, this will be moved to a loader in #6311

packages/payment/src/services/payment-provider.ts
olivermrbl: comment: The registration key will be `pp_[string]` in #6311

packages/payment/src/loaders/providers.ts
fPolic: could you elaborate please, I think we are doing similar thing here as with the auth module

packages/payment/src/models/payment-session.ts
fPolic: this is just for consistency since we expect the provider to return a record from `createSession` so this shouldn't be `null`

packages/payment/src/services/payment-provider.ts
adrien2p: frane, to type accordingly please :)

fPolic: did push the change a while ago but seems like a cache issue with github, didn't see the change on PR until hard refresh

packages/utils/src/payment/abstract-payment-provider.ts
```diff
@@ -0,0 +1,146 @@
+import {
+  IPaymentProcessor,
+  MedusaContainer,
+  PaymentProcessorContext,
+  PaymentProcessorError,
+  PaymentProcessorSessionResponse,
+  PaymentSessionStatus,
+} from "@medusajs/types"
+
+export abstract class AbstractPaymentProcessor implements IPaymentProcessor {
```
olivermrbl: thought(non-blocking): Should we move back to `AbstractPaymentProvider` and `IPaymentProvider`, as we have had previously? I feel like there is something off about having processor and provider as two separate concepts.

olivermrbl: If we, at some point, want to introduce a provider for checkout, we can introduce a CheckoutProvider.

packages/types/src/payment/processors.ts
```diff
@@ -0,0 +1,310 @@
...
+   * The error code.
+   */
+  code?: string
+  /**
+   * Any additional helpful details.
+   */
+  detail?: any
+}
+
+export interface IPaymentProcessor {
```
olivermrbl: See other comment.

olivermrbl: todo: Update JSDoc to not be coupled to other concepts e.g. Region and Cart. The payment provider should be capable of powering payments in any context.

packages/types/src/payment/processors.ts
```diff
@@ -0,0 +1,310 @@
...
+ */
+export type PaymentProcessorContext = {
+  /**
+   * The payment's billing address.
+   */
+  billing_address?: Record<string, unknown> | null // TODO: revisit types
+  /**
+   * The customer's email.
+   */
+  email: string
```
olivermrbl: question: Should this be optional?

packages/types/src/payment/processors.ts
```diff
@@ -0,0 +1,310 @@
...
+  customer?: Record<string, unknown> // TODO: type
+  /**
+   * The cart's context.
+   */
+  context: Record<string, unknown>
+  /**
+   * If the payment session hasn't been created or initiated yet, it'll be an empty object.
+   * If the payment session exists, it'll be the value of the payment session's `data` field.
+   */
+  paymentSessionData: Record<string, unknown>
```
olivermrbl: todo: Snake case this property or camel all others.
```suggestion
  payment_session_data: Record<string, unknown>
```

packages/types/src/payment/mutations.ts
```diff
@@ -69,17 +69,26 @@ export interface CreateRefundDTO {
...
 
-  cart_id?: string
-  resource_id?: string
-  customer_id?: string
+export interface UpdatePaymentSessionDTO {
+  id: string
+  providerContext: PaymentProcessorContext
 }
 
 export interface SetPaymentSessionsDTO {
```
olivermrbl:
```suggestion
export interface SetPaymentSessionDTO {
```

packages/payment/src/services/payment-provider.ts
```diff
@@ -0,0 +1,238 @@
...
+    const provider = this.retrieveProvider(payment.provider_id)
+
+    const error = await provider.cancelPayment(payment.data)
+    if (isPaymentProcessorError(error)) {
+      this.throwFromPaymentProcessorError(error)
+    }
+  }
+
+  async refundFromPayment(
+    payment: { data: Record<string, unknown>; provider_id: string },
```
olivermrbl: todo: Create type

packages/payment/src/services/payment-provider.ts
```diff
@@ -0,0 +1,238 @@
...
+  }): Promise<void> {
+    const provider = this.retrieveProvider(payment.provider_id)
+
+    const error = await provider.cancelPayment(payment.data)
+    if (isPaymentProcessorError(error)) {
+      this.throwFromPaymentProcessorError(error)
+    }
+  }
+
+  async refundFromPayment(
```
olivermrbl: question: Do we need the `From`?
```suggestion
  async refundPayment(
```

packages/payment/src/services/payment-provider.ts
```diff
@@ -0,0 +1,238 @@
...
+
+    const res = await provider.refundPayment(payment.data, amount)
+    if (isPaymentProcessorError(res)) {
+      this.throwFromPaymentProcessorError(res)
+    }
+
+    return res as Record<string, unknown>
+  }
+
+  private throwFromPaymentProcessorError(errObj: PaymentProcessorError) {
```
olivermrbl: question: Goes along with my other comment, the naming is a bit off

packages/payment/src/services/payment-provider.ts
```diff
@@ -0,0 +1,238 @@
...
+    } catch (e) {
+      throw new MedusaError(
+        MedusaError.Types.NOT_FOUND,
+        `Could not find a payment provider with id: ${providerId}`
+      )
+    }
+  }
+
+  async createSession(
+    provider_id: string,
```
olivermrbl: todo:
```suggestion
    providerId: string,
```

packages/payment/src/services/payment-provider.ts
```diff
@@ -0,0 +1,238 @@
...
+
+    return (paymentResponse as PaymentProcessorSessionResponse).session_data
+  }
+
+  async updateSession(
+    paymentSession: {
+      id: string
+      data: Record<string, unknown>
+      provider_id: string
+    },
```
olivermrbl: suggestion: Create type

packages/payment/src/services/payment-provider.ts
```diff
@@ -0,0 +1,238 @@
...
+      this.throwFromPaymentProcessorError(paymentResponse)
+    }
+
+    return (paymentResponse as PaymentProcessorSessionResponse)?.session_data
+  }
+
+  async deleteSession(paymentSession: {
+    provider_id: string
+    data: Record<string, unknown>
+  }): Promise<void> {
```
olivermrbl: suggestion: Create type

packages/payment/src/services/payment-provider.ts
```diff
@@ -0,0 +1,238 @@
...
+    data: CreatePaymentInput
+  ): Promise<Record<string, unknown>> {
+    const { payment_session, provider_id } = data
+    const providerId = provider_id ?? payment_session.provider_id
+
+    const provider = this.retrieveProvider(providerId)
+
+    /**
+     * NOTE: JUST RETRIEVE
+     */
```
olivermrbl: Do we need to revisit this?

packages/payment/src/services/payment-provider.ts
```diff
@@ -0,0 +1,238 @@
...
+    if (isPaymentProcessorError(error)) {
+      this.throwFromPaymentProcessorError(error)
+    }
+  }
+
+  async createPayment(
+    data: CreatePaymentInput
+  ): Promise<Record<string, unknown>> {
+    const { payment_session, provider_id } = data
+    const providerId = provider_id ?? payment_session.provider_id
```
olivermrbl: thought: I wonder if this was added for backward compatibility purposes. Would it make sense to just require a provider ID in `data`?

packages/payment/src/services/payment-provider.ts
```diff
@@ -0,0 +1,238 @@
...
+    }
+
+    return paymentData as Record<string, unknown>
+  }
+
+  async authorizePayment(
+    paymentSession: {
+      provider_id: string
+      data: Record<string, unknown>
+    },
```
olivermrbl: suggestion: Create type

packages/payment/src/services/payment-provider.ts
```diff
@@ -0,0 +1,238 @@
...
+    context: Record<string, unknown>
+  ): Promise<{ data: Record<string, unknown>; status: PaymentSessionStatus }> {
```
olivermrbl, line 121: suggestion: Create type

packages/payment/src/services/payment-provider.ts
```diff
@@ -0,0 +1,238 @@
...
+    const { data, status } = res as PaymentProcessorAuthorizeResponse
+    return { data, status }
+  }
+
+  async updateSessionData(
+    paymentSession: {
+      id: string
+      provider_id: string
+      data: Record<string, unknown>
+    },
```
olivermrbl: suggestion: Create type

packages/payment/src/services/payment-provider.ts
```diff
@@ -0,0 +1,238 @@
...
+    const res = await provider.updatePaymentData(paymentSession.id, data)
+    if (isPaymentProcessorError(res)) {
+      this.throwFromPaymentProcessorError(res)
+    }
+
+    return (res as PaymentProcessorSessionResponse).session_data
+  }
+
+  async getStatus(payment: {
+    provider_id: string
```
olivermrbl:
```suggestion
    providerId: string
```

packages/payment/src/services/payment-provider.ts
```diff
@@ -0,0 +1,238 @@
...
+    data: Record<string, unknown>
+  }): Promise<PaymentSessionStatus> {
+    const provider = this.retrieveProvider(payment.provider_id)
+    return await provider.getPaymentStatus(payment.data)
+  }
+
+  async capturePayment(paymentObj: {
+    provider_id: string
+    data: Record<string, unknown>
```
olivermrbl: suggestion: Create type

packages/utils/src/payment/abstract-payment-provider.ts
```diff
@@ -0,0 +1,146 @@
...
+    paymentSessionData: Record<string, unknown>
+  ): Promise<
+    PaymentProcessorError | PaymentProcessorSessionResponse["session_data"]
+  >
+
+  abstract updatePayment(
+    context: PaymentProcessorContext
+  ): Promise<PaymentProcessorError | PaymentProcessorSessionResponse | void>
+
+  abstract updatePaymentData(
```
olivermrbl: suggestion(non-blocking): `updateSessionData`?

packages/types/src/payment/processors.ts
```diff
@@ -0,0 +1,310 @@
...
+export type PaymentProcessorSessionResponse = {
+  /**
+   * Used to specify data that should be updated in the Medusa backend.
+   */
+  update_requests?: {
+    /**
+     * Specifies a new value of the `metadata` field of the customer associated with the payment.
+     */
+    customer_metadata?: Record<string, unknown>
+  }
```
olivermrbl: thought: Think we should try to see if we can eliminate this. I don't think the pattern is great. If we are not using it right now, should we leave it out?

packages/types/src/payment/processors.ts
```diff
@@ -0,0 +1,310 @@
...
+  /**
+   * The data to be stored in the `data` field of the Payment Session to be created.
+   * The `data` field is useful to hold any data required by the third-party provider to process the payment or retrieve its details at a later point.
+   */
+  session_data: Record<string, unknown>
```
olivermrbl: suggestion:
```suggestion
  data: Record<string, unknown>
```

packages/payment/src/services/payment-module.ts
```diff
@@ -160,305 +166,396 @@ export default class PaymentModuleService<
...
+  completePaymentCollections(
+    paymentCollectionId: string[],
     sharedContext?: Context
-  ): Promise<PaymentDTO[]>
+  ): Promise<PaymentCollectionDTO[]>
 
   @InjectTransactionManager("baseRepository_")
-  async createPayment(
-    data: CreatePaymentDTO | CreatePaymentDTO[],
+  async completePaymentCollections(
```
olivermrbl, line 179: thought(non-blocking): I wonder if we should just remove `completePaymentCollection` and let the full authorization be the completed state.

packages/payment/src/services/payment-module.ts
```diff
@@ -160,305 +166,396 @@ export default class PaymentModuleService<
...
-
-      // TODO: revisit when https://github.com/medusajs/medusa/pull/6253 is merged
-      // if (payment.captured_amount + input.amount === payment.amount) {
-      //   fullyCapturedPaymentsId.push(payment.id)
-      // }
+    if (session.authorized_at) {
+      throw new MedusaError(
+        MedusaError.Types.NOT_ALLOWED,
+        "Session is already authorized."
+      )
```
olivermrbl: thought: Should we return the payment instead and make it idempotent?

olivermrbl (commented):
Great work. Have left comments and todos.

Also, the BigNumber class has been merged. We can tackle the integration of that in a follow-up PR :)

packages/utils/src/payment/abstract-payment-provider.ts
fPolic: Yes 👍 I will rename

packages/payment/src/services/payment-provider.ts
fPolic: Just kept the old method name for consistency with the current provider service but we can rename

packages/payment/src/services/payment-provider.ts
fPolic: yes, passing just the provider_id in data should be enough

packages/payment/src/services/payment-module.ts
fPolic: I think that would make sense. ATM the method is just setting the timestamp on a payment collection which can be done implicitly upon full authorization/capture as well.

EDIT: Will revisit when amounts are implemented

olivermrbl (approved):
LGTM, think this is good to merge now. I imagine things will change a bit when we integrate Stripe and the BigNumber class, so let's push forward 💪

packages/payment/src/services/payment-provider.ts
fPolic: I pushed a [commit](https://github.com/medusajs/medusa/pull/6308/commits/e950a3227b42081cdf11184dd762cbe8a66272ab) where I removed `createPayment` API. It's not needed since `authorizeSession` will create payment.

olivermrbl:
@adrien2p or @carlos-r-l-rodrigues – do any of you want to go through this before we merge?

PR #6310, opened 2024-02-03
feat(dashboard): Log in, reset password, and accept invite pages

kasperkristensen:
**What**
- Adds the following pages to the new admin: /login, /reset-password, /reset-password/:token, and /invite.
- Adds `<Alert />` component to `@medusajs/ui`


srindom:
CleanShot 2024-02-03 at 11  44 42@2x

I tried opening the invite link with an invalid token - maybe we should have an error message that the invite is invalid or something like that?

srindom:
praise: love the animation when you sign up 😍

packages/admin-next/dashboard/src/routes/invite/invite.tsx
```diff
@@ -0,0 +1,341 @@
...
+  const decodedToken: {
+    invite_id: string
+    role: UserRoles
+    user_email: string
+    iat: number
+  } | null = token ? decodeToken(token) : null
+
+  if (!decodedToken) {
+    tokenStatus = "invalid"
+  }
```
srindom: Have you tested this with a token that is not a JWT token - e.g. some random characters. Doesn't seem to result in an invalid token status

srindom (commented):
Looks great - only issue I came across was the token status thing

packages/admin-next/dashboard/src/routes/invite/invite.tsx
kasperkristensen: I did test it, and on my end it did fail when the token was not a valid JWT token, but I guess it was possible to have false positives, as long as decodeToken didn't return `null`. 

I have added a parse with zod, that validates that the decoded token has all the expected fields, e.g. a valid email, invite_id, etc, and if any one is missing then we show a invalid state screen instead 👍

srindom: Nice! LGTM 💪

srindom (approved):
LGTM!

PR #6311, opened 2024-02-04
feat(payment, payment-stripe): Add Stripe module provider

olivermrbl:
Depends on #6308. There is some finessing to be done after having merged the depending PR, as this one doesn't ship the payment provider service of the module.

**What**
- Introduce Stripe module provider `@medusajs/payment-stripe`
- Rework the events architecture of payment providers

This PR introduces a new payment plugin compatible with the Payment Module, `@medusajs/payment-stripe`.

However, the more significant change in this PR is rearchitecting how webhooks events are handled in payment plugins.

Until now, we've relied on payment plugins to ship with their own API endpoints and subscribers to perform business logic for webhook events. With the introduction of module plugins, this needs changing as the module should not ship with tools that rely on the file system. For example, developers should be able to initiate the payment module with Stripe in a serverless function.

With this PR, Medusa's core API will expose a single endpoint that all payment providers should use to process their webhook events.
```
POST /hooks/payment/[provider]
```

In this API route, we will do nothing but emit an event with the data received and respond with 200, as this is generally deemed best practice in webhook handlers.

In addition to the API route, Medusa's core will ship a subscriber handling this payment event. In the subscriber, we resolve the payment module service and call a method dedicated to performing the business logic for the different providers.  

Let's look at a basic example with Stripe:

1. Checkout flow complete
2. Stripe sends a webhook event to `POST /hooks/payment/stripe`
3. Our core API route extracts the provider from the query and emits an event:
```ts
const { provider } = req.params
const eventData = { provider, data: req.body } 

evnetBus.emit("payment.webhook_received", eventData)
```
4. Our core subscriber picks up the event and calls the payment module service
```ts
const paymentModuleService = container.resolve(ModuleRegistrationName.PAYMENT)
await paymentModuleService.onWebhookReceived(data)
```
5. The payment module propagates the method call to the correct payment provider service
```ts
async processEvent(eventData) {
   // Payment providers are responsible for processing the event and returning normalized action + data
   const {action, data} = await this.paymentProviderService_.processEvent(
      eventData.provider,
      eventData.payload
    )

  switch (action) {
      case PaymentActions.CAPTURED:
         // ... call capture payment
      case PaymentActions.AUTHORIZED:
        // ... call authorize session
    }
}
```

Finally, this PR also establishes a new format for how providers are installed in the Payment Module. We want to support multiple installations (i.e. with different credentials) of the same plugin, e.g. different Stripe accounts for different regions. 

The module config is as follows:
```
{
  payment: {
    resolve: "@medusajs/payment",
    options: {
      providers: [
        {
           resolve: "@medusajs/payment-stripe",
           options: {
             config: {
               dkk: { ...dkkCredentials },
               usd: { ...usdCredentials },
               eur: { ...eurCredentials },
             }
           }
        }
      ]
    }
}
```

This example configuration above will enable three installations of each Stripe provider and the provider IDs will be formatted and registered in the DB as follows: `stripe-dkk`, `stripe-usd`, `stripe-eur`. 

-----------

Separate from the webhook event stuff, it is also worth noting that aside from prohibiting API endpoints and subscribers in payment plugins, we will also no longer be able to ship admin UI routes and widgets. It is possible that we will support UI in modules and module providers at some point in the future.

### DIFF
--- a/.changeset/loud-mirrors-perform.md
+++ b/.changeset/loud-mirrors-perform.md
@@ -1,0 +1,5 @@
+---
+"@medusajs/types": patch
+---
+
+feat(fulfillment): List shipping options filtered by context and rules
--- a/packages/fulfillment/integration-tests/__fixtures__/index.ts
+++ b/packages/fulfillment/integration-tests/__fixtures__/index.ts
@@ -1,0 +1,39 @@
+import { CreateShippingOptionDTO } from "@medusajs/types"
+
+export function generateCreateShippingOptionsData({
+  name,
+  service_zone_id,
+  shipping_profile_id,
+  service_provider_id,
+  price_type,
+  rules,
+  type,
+  data,
+}: Omit<CreateShippingOptionDTO, "name" | "price_type" | "type"> & {
+  price_type?: CreateShippingOptionDTO["price_type"]
+  name?: string
+  type?: CreateShippingOptionDTO["type"]
+}): Required<CreateShippingOptionDTO> {
+  return {
+    service_zone_id: service_zone_id,
+    shipping_profile_id: shipping_profile_id,
+    service_provider_id: service_provider_id,
+    type: type ?? {
+      code: "test-type",
+      description: "test-description",
+      label: "test-label",
+    },
+    data: data ?? {
+      amount: 1000,
+    },
+    name: name ?? Math.random().toString(36).substring(7),
+    price_type: price_type ?? "flat",
+    rules: rules ?? [
+      {
+        attribute: "weight",
+        operator: "eq",
+        value: "test",
+      },
+    ],
+  }
+}
--- a/packages/fulfillment/integration-tests/__tests__/fulfillment-module-service.spec.ts
+++ b/packages/fulfillment/integration-tests/__tests__/fulfillment-module-service.spec.ts
@@ -1,4 +1,4 @@
-import { Modules } from "@medusajs/modules-sdk"
+import {Modules} from "@medusajs/modules-sdk"
 import {
   CreateFulfillmentSetDTO,
   CreateGeoZoneDTO,
@@ -12,10 +12,20 @@ import {
   UpdateGeoZoneDTO,
   UpdateServiceZoneDTO,
 } from "@medusajs/types"
-import { GeoZoneType } from "@medusajs/utils"
-import { moduleIntegrationTestRunner, SuiteOptions } from "medusa-test-utils"
+import {GeoZoneType} from "@medusajs/utils"
+import {moduleIntegrationTestRunner, SuiteOptions} from "medusa-test-utils"
+import {generateCreateShippingOptionsData} from "../__fixtures__"
 
 jest.setTimeout(100000)
+
+// TODO: Temporary until the providers are sorted out
+const createProvider = async (MikroOrmWrapper, providerId: string) => {
+  const [{ id }] = await MikroOrmWrapper.forkManager().execute(
+    `insert into service_provider (id) values ('${providerId}') returning id`
+  )
+
+  return id
+}
 
 moduleIntegrationTestRunner({
   moduleName: Modules.FULFILLMENT,
@@ -247,6 +257,177 @@ moduleIntegrationTestRunner({
                 expect.objectContaining({ id: createdZone1.id }),
               ])
             )
+          })
+        })
+
+        describe("shipping options", () => {
+          it("should list shipping options with a filter", async function () {
+            const fulfillmentSet = await service.create({
+              name: "test",
+              type: "test-type",
+              service_zones: [
+                {
+                  name: "test",
+                },
+              ],
+            })
+
+            const shippingProfile = await service.createShippingProfiles({
+              name: "test",
+              type: "default",
+            })
+
+            const providerId = await createProvider(
+              MikroOrmWrapper,
+              "sp_jdafwfleiwuonl"
+            )
+
+            const [shippingOption1] = await service.createShippingOptions([
+              generateCreateShippingOptionsData({
+                service_zone_id: fulfillmentSet.service_zones[0].id,
+                shipping_profile_id: shippingProfile.id,
+                service_provider_id: providerId,
+                rules: [
+                  {
+                    attribute: "test-attribute",
+                    operator: "in",
+                    value: ["test"],
+                  },
+                ],
+              }),
+              generateCreateShippingOptionsData({
+                service_zone_id: fulfillmentSet.service_zones[0].id,
+                shipping_profile_id: shippingProfile.id,
+                service_provider_id: providerId,
+                rules: [
+                  {
+                    attribute: "test-attribute",
+                    operator: "eq",
+                    value: "test",
+                  },
+                  {
+                    attribute: "test-attribute2.options",
+                    operator: "in",
+                    value: ["test", "test2"],
+                  },
+                ],
+              }),
+            ])
+
+            const listedOptions = await service.listShippingOptions({
+              name: shippingOption1.name,
+            })
+
+            expect(listedOptions).toHaveLength(1)
+            expect(listedOptions[0].id).toEqual(shippingOption1.id)
+          })
+
+          it("should list shipping options with a context", async function () {
+            const fulfillmentSet = await service.create({
+              name: "test",
+              type: "test-type",
+              service_zones: [
+                {
+                  name: "test",
+                },
+              ],
+            })
+
+            const shippingProfile = await service.createShippingProfiles({
+              name: "test",
+              type: "default",
+            })
+
+            const providerId = await createProvider(
+              MikroOrmWrapper,
+              "sp_jdafwfleiwuonl"
+            )
+
+            const [shippingOption1, , shippingOption3] =
+              await service.createShippingOptions([
+                generateCreateShippingOptionsData({
+                  service_zone_id: fulfillmentSet.service_zones[0].id,
+                  shipping_profile_id: shippingProfile.id,
+                  service_provider_id: providerId,
+                  rules: [
+                    {
+                      attribute: "test-attribute",
+                      operator: "in",
+                      value: ["test"],
+                    },
+                  ],
+                }),
+                generateCreateShippingOptionsData({
+                  service_zone_id: fulfillmentSet.service_zones[0].id,
+                  shipping_profile_id: shippingProfile.id,
+                  service_provider_id: providerId,
+                  rules: [
+                    {
+                      attribute: "test-attribute",
+                      operator: "in",
+                      value: ["test-test"],
+                    },
+                  ],
+                }),
+                generateCreateShippingOptionsData({
+                  service_zone_id: fulfillmentSet.service_zones[0].id,
+                  shipping_profile_id: shippingProfile.id,
+                  service_provider_id: providerId,
+                  rules: [
+                    {
+                      attribute: "test-attribute",
+                      operator: "eq",
+                      value: "test",
+                    },
+                    {
+                      attribute: "test-attribute2.options",
+                      operator: "in",
+                      value: ["test", "test2"],
+                    },
+                  ],
+                }),
+              ])
+
+            let listedOptions = await service.listShippingOptions({
+              context: {
+                "test-attribute": "test",
+                "test-attribute2": {
+                  options: "test2",
+                },
+              },
+            })
+
+            expect(listedOptions).toHaveLength(2)
+            expect(listedOptions).toEqual(
+              expect.arrayContaining([
+                expect.objectContaining({ id: shippingOption1.id }),
+                expect.objectContaining({ id: shippingOption3.id }),
+              ])
+            )
+
+            listedOptions = await service.listShippingOptions({
+              fulfillment_set_id: { $ne: fulfillmentSet.id },
+              context: {
+                "test-attribute": "test",
+                "test-attribute2": {
+                  options: "test2",
+                },
+              },
+            })
+
+            expect(listedOptions).toHaveLength(0)
+
+            listedOptions = await service.listShippingOptions({
+              fulfillment_set_type: "non-existing-type",
+              context: {
+                "test-attribute": "test",
+                "test-attribute2": {
+                  options: "test2",
+                },
+              },
+            })
+
+            expect(listedOptions).toHaveLength(0)
           })
         })
       })
@@ -788,34 +969,16 @@ moduleIntegrationTestRunner({
               fulfillment_set_id: fulfillmentSet.id,
             })
 
-            // TODO: change that for a real provider instead of fake data manual inserted data
-            const [{ id: providerId }] =
-              await MikroOrmWrapper.forkManager().execute(
-                "insert into service_provider (id) values ('sp_jdafwfleiwuonl') returning id"
-              )
+            const providerId = await createProvider(
+              MikroOrmWrapper,
+              "sp_jdafwfleiwuonl"
+            )
 
-            const createData: CreateShippingOptionDTO = {
-              name: "test-option",
-              price_type: "flat",
+            const createData: CreateShippingOptionDTO = generateCreateShippingOptionsData({
               service_zone_id: serviceZone.id,
               shipping_profile_id: shippingProfile.id,
               service_provider_id: providerId,
-              type: {
-                code: "test-type",
-                description: "test-description",
-                label: "test-label",
-              },
-              data: {
-                amount: 1000,
-              },
-              rules: [
-                {
-                  attribute: "test-attribute",
-                  operator: "in",
-                  value: ["test-value"],
-                },
-              ],
-            }
+            })
 
             const createdShippingOption = await service.createShippingOptions(
               createData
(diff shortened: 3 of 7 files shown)


packages/utils/src/payment/abstract-payment-provider.ts
```diff
@@ -0,0 +1,31 @@
+import { PaymentProcessorError } from "@medusajs/types"
```
olivermrbl: todo: Revisit when #6308 has been merged

packages/types/src/payment/payment-provider.ts
```diff
@@ -0,0 +1,88 @@
+// import { Address, Customer, PaymentSessionStatus } from "../models"
```
olivermrbl: todo: Revisit when #6308 has been merged

packages/payment/src/services/payment-module.ts
```diff
@@ -271,4 +279,29 @@ export default class PaymentModuleService implements IPaymentModuleService {
   ): Promise<PaymentCollectionDTO> {
     throw new Error("Method not implemented.")
   }
+
+  private async createProvidersOnLoad() {
```
olivermrbl: todo: Should use payment provider service from #6308

packages/payment/src/loaders/providers.ts
```diff
@@ -1,24 +1,18 @@
...
 import { LoaderOptions, ModuleProvider, ModulesSdkTypes } from "@medusajs/types"
-import { Lifetime, asFunction } from "awilix"
+import { Lifetime, asFunction, asValue } from "awilix"
 
 const registrationFn = async (klass, container, pluginOptions) => {
-  container.register({
-    [`payment_provider_${klass.prototype}`]: asFunction(
-      (cradle) => new klass(cradle, pluginOptions),
-      {
+  Object.entries(pluginOptions.config).map(([k, v]) => {
```
olivermrbl: todo: Needs documentation. Perhaps jsdoc will suffice

packages/payment-stripe/src/core/stripe-base.ts
```diff
@@ -0,0 +1,346 @@
...
+  }
+
+  /**
+   * Constructs Stripe Webhook event
+   * @param {object} data - the data of the webhook request: req.body
+   * @param {object} signature - the Stripe signature on the event, that
+   *    ensures integrity of the webhook event
+   * @return {object} Stripe Webhook event
+   */
+  constructWebhookEvent(data, signature) {
```
olivermrbl: todo: Revisit when #6308 has been merged. Not functional in its current state.

olivermrbl:
todo:
- Add `onWebhookReceived` to `IPaymentModuleService`
- Add `onWebhookReceived` to `AbstractPaymentModuleProvider`
- Make Stripe flow fully functional -> test it

packages/medusa/src/api-v2/hooks/payment/[provider]/route.ts
```diff
@@ -0,0 +1,25 @@
...
+    const { provider } = req.params
+
+    const options = req.scope.resolve(ModuleRegistrationName.PAYMENT).options
+
+    const event = { provider, data: req.body }
+
+    const eventBus = req.scope.resolve("eventBusService")
+
+    // we delay the processing of the event to avoid a conflict caused by a race condition
+    await eventBus.emit("payment.webhook_received", event, {
```
Comment (adrien2p) [inline packages/medusa/src/api-v2/hooks/payment/[provider]/route.ts]: todo: those event string should probably be in the utils package in an enum or something. To error prone otherwise, wdyt?

packages/payment-stripe/src/index.ts
```diff
@@ -0,0 +1,24 @@
+import { ModuleProviderExports } from "@medusajs/types"
+import {
+  StripeBancontactService,
+  StripeBlikService,
+  StripeGiropayService,
+  StripeIdealService,
+  StripeProviderService,
+  StripePrzelewy24Service,
+} from "./services/index"
```
adrien2p: nit: shorten

packages/payment/src/loaders/providers.ts
adrien2p: todo: check that config is present, also, remove `k`, `v` and use instead `name`, `config`

packages/payment/src/loaders/providers.ts
```diff
@@ -1,24 +1,18 @@
...
         lifetime: klass.LIFE_TIME || Lifetime.SINGLETON,
-      }
-    ),
-  })
-
-  container.registerAdd(
-    "payment_providers",
-    asFunction((cradle) => new klass(cradle, pluginOptions), {
-      lifetime: klass.LIFE_TIME || Lifetime.SINGLETON,
+      }).inject(() => ({ config: v })),
```
adrien2p: todo: since we already have the config at hand, I would probably pass it in the plugin options associated to a specifi key such as `currentProviderConfig` or similar. The inject can lead to conflict with other registration that module dependencies could include. wdyt?

packages/payment/src/services/payment-module.ts
adrien2p: should we move that to a loader instead of the hook?

packages/auth/src/providers/google.ts
```diff
@@ -1,12 +1,12 @@
-import { AbstractAuthModuleProvider, MedusaError } from "@medusajs/utils"
```
adrien2p, line 1: todo: missing changeset?

packages/payment-stripe/src/utils/webhook-handlers.ts
```diff
@@ -0,0 +1,144 @@
...
+      err?.detail ?? err?.message
+    }`
+  }
+  if (err?.code === "409") {
+    message = `Stripe webhook ${event} handle failed.${EOL}${
+      err?.detail ?? err?.message
+    }`
+  }
+
+  return message
```
adrien2p: todo: we should probably revisit

packages/payment/integration-tests/__tests__/loaders/providers.spec.ts
```diff
@@ -0,0 +1,99 @@
...
+                apiKey: "pk_test_123",
+              },
+              usd: {
+                apiKey: "pk_test_456",
+              },
+            },
+          },
+        },
+      ],
+    })
```
adrien2p: todo: not sure why we have that, shouldn't it be part of the initModules?

packages/types/src/modules-sdk/module-provider.ts
```diff
@@ -0,0 +1,11 @@
+import { Constructor } from "./index"
+
+export type ModuleProviderExports = {
+  services: Constructor<any>[]
+}
+
+export type ModuleProvider = {
+  resolve: string | ModuleProviderExports
+  provider_name?: string
+  options: Record<string, unknown>
```
adrien2p, line 10: todo: probably some doc would be good to explain to the user what impact it has and what kind of options are expected

packages/types/src/payment/provider.ts
```diff
@@ -29,7 +32,7 @@ export type PaymentProviderContext = {
   /**
    * The customer associated with this payment.
    */
-  customer?: Record<string, unknown> // TODO: type
+  customer?: CustomerDTO
```
adrien2p: suggestion: The payment should have his own address and customer DTO to define what is expected from those DTO, they are not necessarely entire custom or address no? here it looks like we need to provide all proprs from those DTO that are coming from another domain, is that true?

packages/payment-stripe/src/types/index.ts
```diff
@@ -0,0 +1,54 @@
...
+  /**
+   * The delay in milliseconds before processing the webhook event.
+   * @defaultValue 5000
+   */
+  webhook_delay?: number
+  /**
+   * The number of times to retry the webhook event processing in case of an error.
+   * @defaultValue 3
+   */
+  webhook_retries?: number
```
olivermrbl: See other comment. Think these are general payment module options. Or core options

packages/types/src/payment/provider.ts
fPolic: Yep agree, I created separate DTOs which are, for now, just partials from the original DTOs.

olivermrbl:
Attaching an illustration of what @fPolic and I discussed about the payment webhook architecture. Hope, this will ease the review process. 

image

packages/medusa/src/api-v2/hooks/payment/[provider]/route.ts
```diff
@@ -0,0 +1,31 @@
...
+    const config = req.scope.resolve(ModuleRegistrationName.PAYMENT).config
+
+    const event = {
+      provider,
+      payload: { data: req.body, headers: req.headers },
+    }
+
+    const eventBus = req.scope.resolve("eventBusService")
+
+    // const validated = await paymentModuleService.validateWebhook(data)
```
Comment (olivermrbl) [inline packages/medusa/src/api-v2/hooks/payment/[provider]/route.ts]: todo: rm

packages/medusa/src/api-v2/hooks/payment/[provider]/route.ts
```diff
@@ -0,0 +1,31 @@
...
+
+    // we delay the processing of the event to avoid a conflict caused by a race condition
+    await eventBus.emit(PaymentWebhookEvents.WebhookReceived, event, {
+      delay: config.webhook_delay || 5000,
+      attempts: config.webhook_retries || 3,
```
Comment (olivermrbl) [inline packages/medusa/src/api-v2/hooks/payment/[provider]/route.ts]: thought: I would put these in the payment module options. Can't see why these should differ per provider.

packages/medusa/src/subscribers/payment-webhook.ts
```diff
@@ -0,0 +1,35 @@
...
+    this.eventBusService_ = eventBusService
+    this.paymentModuleService_ = paymentModuleService
+
+    this.eventBusService_.subscribe(
+      PaymentWebhookEvents.WebhookReceived,
+      this.processEvent as Subscriber
+    )
+  }
+
+  processEvent = async (data: ProviderWebhookPayload): Promise<void> => {
```
olivermrbl, line 38: comment: I think it would make sense to add a TODO, that we should consider using a workflow for this flow

packages/payment/src/services/payment-provider.ts
```diff
@@ -173,6 +175,15 @@ export default class PaymentProviderService {
     return res as Record<string, unknown>
   }
 
+  async processEvent(
```
olivermrbl: suggestion: Rename `getWebhookActionAndData`

I would consider doing the same with the method in the provider interface.

We are only really processing the event in the payment module service method.

packages/medusa/src/api-v2/hooks/middlewares.ts
```diff
@@ -0,0 +1,12 @@
+import { raw } from "body-parser"
+
+import { MiddlewareRoute } from "../../types/middlewares"
+
+export const hooksRoutesMiddlewares: MiddlewareRoute[] = [
+  {
+    method: ["POST"],
+    bodyParser: false,
```
olivermrbl: thought: This is actually a bit tricky. Stripe requires us to disable the `bodyParser` middleware, but this might differ from provider to provider. 

Perhaps, we need to think of ways of performing webhook event operations that are specific to providers. E.g. through a module plugin option.

packages/medusa/src/subscribers/payment-webhook.ts
```diff
@@ -0,0 +1,48 @@
...
+
+  /**
+   * TODO: consider moving this to a workflow
+   */
+  processEvent = async (data: ProviderWebhookPayload): Promise<void> => {
+    if ((data.payload.data as unknown as SerializedBuffer).type === "Buffer") {
+      data.payload.data = Buffer.from(
+        (data.payload.data as unknown as SerializedBuffer).data
+      )
+    }
```
olivermrbl: comment: Goes along with my other thought. This also seems a bit off to perform here, if we do it because of what specific providers expect.

olivermrbl:
@medusajs/engineering, would appreciate a review here. I can't do it myself, as I authored the pr, but will leave comments :)

packages/medusa/src/subscribers/payment-webhook.ts
carlos-r-l-rodrigues: +1. This decode should be at the plugin provider level if there is something to do with the provider.

In case this is a SerializedBuffer that BullMQ has created, I think it is fine to keep it here, I would just add a comment.

olivermrbl: @fPolic, why was this added here?

fPolic: Since we are dispatching the webhook payload via the event bus, the body, which is a buffer, is serialized. Stripe expects a buffer or a string of the request body so we are reconstructing a buffer in the subscriber.

Alternatively, one way of unifying this across provides could be to always pass the raw body(string/buffer) to a plugin which would be responsible for parsing the webhook payload?

olivermrbl: I see, makes sense.

> Alternatively, one way of unifying this across provides could be to always pass the raw body(string/buffer) to a plugin

This sounds like a reasonable approach to me. I would perhaps even add both the raw body and parsed body, so providers generally wouldn't need to do anything with the payload. I believe the way Stripe handles it, it actually an edge case, which is weird coming from them but I am sure they have good reasons.

So in practice, I would remove the `bodyParser` from the middleware and add a custom middleware that attaches both types of parsed payloads. The API route would end up looking something like:
```
  try {
   ...
    const event = {
      provider,
      payload: { data: req.body, headers: req.headers, rawData: req.rawBody },
    }
  ...
```

Wdyt?

fPolic: Sounds like a good approach - I will add the middleware 👍

olivermrbl:
Perhaps, it would make more sense to get this merged, so we can start integrating the payment module into workflows, and then focus on the provider webhook architecture in a follow-up PR? That would also make it easier to review, I believe.

If we choose to go down this path, then @fPolic could I get you to resolve the last issues in the tests?

fPolic:
@olivermrbl should be fixed now. I believe the only thing left to change here is the middleware as you suggested but I'm also fine with merging and tackling this in a follow up PR to unblock integration into workflows.

packages/medusa/src/api-v2/hooks/middlewares.ts
adrien2p: maybe we can just not perform any operations and expect raw body to reach the module? We also need to consider when a module is external

packages/medusa/src/loaders/api.ts
```diff
@@ -52,6 +50,7 @@ export default async ({
       throw Error("An error occurred while registering Medusa Core API Routes")
     }
   } else {
+    app.use(bodyParser.json())
```
adrien2p, line 53: Q: wouldn't this break the v1 webhooks?

packages/payment/src/services/payment-module.ts
```diff
@@ -206,23 +207,38 @@ export default class PaymentModuleService<
...
         currency_code: data.providerContext.currency_code,
         payment_collection: paymentCollectionId,
-        data: sessionData,
       },
       sharedContext
     )
 
-    return await this.baseRepository_.serialize(created, { populate: true })
+    try {
+      const sessionData = await this.paymentProviderService_.createSession(
```
adrien2p, line 221: suggestion: communicating with an external third party API shouldn't happen within a transaction. We should probably split that in a workflow. should we revisit that ? eventually, for now it could be moved one level higher and then call a protected `createPaymentSession_` which would be wrapped in a transaction

packages/types/src/payment/provider.ts
```diff
@@ -1,4 +1,33 @@
 import { PaymentSessionStatus } from "./common"
+import { CustomerDTO } from "../customer"
+import { AddressDTO } from "../address"
+import { ProviderWebhookPayload } from "./mutations"
+
+/**
+ * Normalized events from payment provider to internal payment module events.
+ */
+export enum PaymentActions {
```
adrien2p, line 13: todo: enum should not be part of the type package but moved to the utils package

packages/medusa/src/loaders/api.ts
fPolic: Currently, we are calling this before v1 and v2 in the API loader so everything stayed the same for v1. I just moved it here to be called just before v1 routes since for V2 routes this is handled in RoutesLoader.

packages/medusa/src/api-v2/hooks/middlewares.ts
olivermrbl: Agree. Suggested a hybrid [here](https://github.com/medusajs/medusa/pull/6311#discussion_r1500941303).

packages/medusa/src/loaders/api.ts
adrien2p: I mean that the webhook in v1 will have the body parsed no?

packages/medusa/src/api-v2/hooks/middlewares.ts
adrien2p: sorry yeah seen it after in the review process

packages/types/src/payment/provider.ts
fPolic: what if a type definition in `types` depends on that enum? since `utils` can't be a dep of `types` we were writing duplicate declarations in some cases one in `utils` and one in `types`

packages/medusa/src/loaders/api.ts
fPolic: No since in the plugins loader we only call `RoutesLoader` which will apply body parser depending on the config

packages/types/src/payment/provider.ts
olivermrbl: I guess we would have to create a dedicated type for the type definition in `@medusajs/types` (lot of types here :D)

packages/payment/src/services/payment-module.ts
olivermrbl: Yes, this will be revisited as we introduce workflows.

packages/medusa/src/subscribers/payment-webhook.ts
fPolic: @olivermrbl - I've added support for `rawBody` but with a slightly different approach in [this](https://github.com/medusajs/medusa/pull/6311/commits/4ec6d16b963bd1e276d542d196ba2a9eb2c24e93) commit. Let me know wdyt.

olivermrbl: This looks good to me. @kasperkristensen, can you glance over this commit too? 🙏

olivermrbl: However, I would still leave the buffer transformation to Stripe. Is that possible?

fPolic: You mean the deserialization that we do in the subscriber? Since we have a serialized buffer because we transmit data through the event bus I think we should "fix" the data back in the subscriber. From the plugin perspective, I'm not sure why we would expect to receive a serialized buffer instead of a plain buffer there.

PR #6314, opened 2024-02-05
Feat(auth): Remove auth provider entity

pKorsholm:
**What**
- remove auth provider entity

**Why**
- The auth provider entity was not really used anywhere

**How**
- Keeping loader behavior as is but removing the

### DIFF
--- a/packages/auth/src/services/auth-module.ts
+++ b/packages/auth/src/services/auth-module.ts
@@ -80,7 +80,7 @@ export default class AuthModuleService<TAuthUser extends AuthUser = AuthUser>
       populate: true,
     })
 
-    return Array.isArray(data) ? serializedUsers : serializedUsers[0]
+    return serializedUsers
   }
 
   update(
--- a/integration-tests/plugins/__tests__/customer/store/create-customer.spec.ts
+++ b/integration-tests/plugins/__tests__/customer/store/create-customer.spec.ts
@@ -49,7 +49,7 @@ describe("POST /store/customers", () => {
       ModuleRegistrationName.AUTH
     )
     const { jwt_secret } = appContainer.resolve("configModule").projectConfig
-    const authUser = await authService.createAuthUser({
+    const authUser = await authService.create({
       entity_id: "store_user",
       provider: "emailpass",
       scope: "store",
--- a/integration-tests/plugins/__tests__/customer/store/delete-customer-address.spec.ts
+++ b/integration-tests/plugins/__tests__/customer/store/delete-customer-address.spec.ts
@@ -28,14 +28,6 @@ describe("DELETE /store/customers/me/addresses/:address_id", () => {
     )
   })
 
-  // TODO: delete with removal of authProvider
-  beforeEach(async () => {
-    const onStart =
-      appContainer.resolve(ModuleRegistrationName.AUTH).__hooks
-        .onApplicationStart ?? (() => Promise.resolve())
-    await onStart()
-  })
-
   afterAll(async () => {
     const db = useDb()
     await db.shutdown()
--- a/integration-tests/plugins/__tests__/customer/store/update-customer-address.spec.ts
+++ b/integration-tests/plugins/__tests__/customer/store/update-customer-address.spec.ts
@@ -26,14 +26,6 @@ describe("POST /store/customers/:id/addresses/:address_id", () => {
     )
   })
 
-  // TODO: delete with removal of authProvider
-  beforeEach(async () => {
-    const onStart =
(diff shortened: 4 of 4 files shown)


srindom (approved):
LGTM

packages/auth/integration-tests/__tests__/services/module/auth-user.spec.ts
```diff
@@ -42,23 +40,23 @@ describe("AuthModuleService - AuthUser", () => {
 
   describe("listAuthUsers", () => {
     it("should list authUsers", async () => {
-      const authUsers = await service.listAuthUsers()
+      const authUsers = await service.list()
 
       expect(authUsers).toEqual([
         expect.objectContaining({
-          provider: { provider: "manual" },
+          provider: "manual",
```
adrien2p, line 47: q: why didn't we kept the object here? It can be a bit weird to get the id on the relation property wdyt? normally, the util we are using to serialize have the forceObject set to true by default.

packages/auth/integration-tests/__tests__/services/module/providers.spec.ts
```diff
@@ -42,53 +41,17 @@ describe("AuthModuleService - AuthProvider", () => {
...
-          authScope: "non-existing",
-        } as any
+        "notRegistered",
+        {} as any
       )
 
       expect(success).toBe(false)
       expect(error).toEqual(
-        `Scope "non-existing" is not valid for provider emailpass`
+        "AuthenticationProvider with for provider: notRegistered wasn't registered in the module. Have you configured your options correctly?"
```
adrien2p: suggestion: I think there is a typo

packages/auth/integration-tests/__tests__/services/module/auth-user.spec.ts
adrien2p: my bad, just saw it became a text prop 😂

packages/auth/src/providers/email-password.ts
```diff
@@ -20,8 +20,8 @@ class EmailPasswordProvider extends AbstractAuthModuleProvider {
     this.authUserSerivce_ = authUserService
```
adrien2p, line 23: suggestion: I think line 12/12 should be something to pass to the super constructor as argumet such that the parent class does not depends on potentially miss configured child class. wdyt?

packages/auth/src/services/auth-module.ts
```diff
@@ -26,138 +24,57 @@ import {
...
-  ): Promise<AuthUserDTO>
+  create(data: CreateAuthUserDTO, sharedContext?: Context): Promise<AuthUserDTO>
 
   @InjectManager("baseRepository_")
-  async createAuthUser(
+  async create(
     data: CreateAuthUserDTO[] | CreateAuthUserDTO,
     @MedusaContext() sharedContext: Context = {}
   ): Promise<AuthTypes.AuthUserDTO | AuthTypes.AuthUserDTO[]> {
     const input = Array.isArray(data) ? data : [data]
```
adrien2p: q: The base service already manages that part plus return a single entity or an array of it if it receives an array. So potentially we could just pass the data to the create_ which does the same to the inner service and return the result that we serialize and return. Just mentionning but we can keep it manual if we prefer. cc @olivermrbl

packages/auth/src/services/auth-module.ts
```diff
@@ -168,25 +85,30 @@ export default class AuthModuleService<
...
   @InjectManager("baseRepository_")
-  async updateAuthUser(
+  async update(
     data: UpdateAuthUserDTO | UpdateAuthUserDTO[],
     @MedusaContext() sharedContext: Context = {}
   ): Promise<AuthTypes.AuthUserDTO | AuthTypes.AuthUserDTO[]> {
     const input = Array.isArray(data) ? data : [data]
 
-    const updatedUsers = await this.updateAuthUsers_(input, sharedContext)
+    const updatedUsers = await this.update_(input, sharedContext)
```
adrien2p: same as previous

packages/utils/src/auth/abstract-auth-provider.ts
```diff
@@ -17,10 +22,11 @@ export abstract class AbstractAuthModuleProvider {
   }
 
   protected constructor({ scopes }) {
```
adrien2p: suggestion: Accept PROVIDER and DISPLAY NAME as part of the constructor to not rely on the child class setting it on the static props

adrien2p: depending on that, you can simplify L 16 and 20

PR #6315, opened 2024-02-05
feat: Region Module (basic CRUD)

olivermrbl:


packages/region/src/services/region-module.ts
```diff
@@ -0,0 +1,230 @@
...
+    TCurrency extends Currency = Currency
+  >
+  extends ModulesSdkUtils.abstractModuleServiceFactory<
+    InjectedDependencies,
+    RegionDTO,
+    {
+      Country: {
+        dto: RegionCountryDTO
+        singular: "Country"
+        plural: "Countries"
```
srindom: nit: can leave out - the `pluralize` helper takes care of this automatically

packages/region/src/services/region-module.ts
```diff
@@ -0,0 +1,230 @@
...
+      }
+      Currency: {
+        dto: RegionCurrencyDTO
+        singular: "Currency"
+        plural: "Currencies"
```
srindom: nit: can leave out - the pluralize helper takes care of this automatically

packages/region/src/services/region-module.ts
```diff
@@ -0,0 +1,230 @@
...
+    super(...arguments)
+    this.baseRepository_ = baseRepository
+    this.regionService_ = regionService
+    this.countryService_ = countryService
+    this.currencyService_ = currencyService
+  }
+
+  __hooks = {
+    onApplicationStart: async () => await this.createCountriesAndCurrencies(),
+  }
```
srindom: thought(non-blocking): this looks super ugly - the `__` makes it feel like it's something you really shouldn't use. Gives me these vibes: `React.__SECRET_DOM_DO_NOT_USE_OR_YOU_WILL_BE_FIRED`.

Let's revisit at some point to make the API for this better. Same goes for joinerConfig

packages/region/integration-tests/__tests__/services/region-module.spec.ts
```diff
@@ -0,0 +1,93 @@
...
+    await MikroOrmWrapper.setupDatabase()
+    repositoryManager = await MikroOrmWrapper.forkManager()
+
+    service = await initialize({
+      database: {
+        clientUrl: DB_URL,
+        schema: process.env.MEDUSA_REGION_DB_SCHEMA,
+      },
+    })
+
```
adrien2p: todo: we should use the unify loading approach 

```ts
    const initModulesConfig = getInitModuleConfig()
    const { medusaApp, shutdown } = await initModules(initModulesConfig)
    service = medusaApp.modules[Modules.AUTH]
```

packages/region/integration-tests/__tests__/services/region-module.spec.ts
```diff
@@ -0,0 +1,93 @@
...
+    testManager = await MikroOrmWrapper.forkManager()
+  })
+
+  afterEach(async () => {
+    await MikroOrmWrapper.clearDatabase()
```
adrien2p: suggestion: this would become 
```ts
await shutdownFunc()
```

packages/region/src/services/region-module.ts
```diff
@@ -0,0 +1,230 @@
...
+    if (!currency) {
+      currsToCreate = Object.entries(DefaultsUtils.defaultCurrencies).map(
+        ([code, currency]) => ({
+          code: code.toLowerCase(),
+          symbol: currency.symbol,
+          symbol_native: currency.symbol_native,
+          name: currency.name,
+        })
+      )
+    }
```
srindom: suggestion(non-blocking): compare count in db with `defaultCurrencies` count:

```
const [, count] = await this.currencyService_.listAndCount({}, {select: ["id"], take: 1 })

if (count !== Object.entries(defaultCurrencies).length) {
  ...
}
```
This way we can add currencies by updating defaultCurrencies. 

Note: the creation would have to account for this somehow to ensure it's only the difference that is created.

packages/region/integration-tests/__tests__/services/region-module.spec.ts
```diff
@@ -0,0 +1,93 @@
...
+  })
+
+  beforeEach(async () => {
+    await MikroOrmWrapper.setupDatabase()
+    testManager = MikroOrmWrapper.forkManager()
+
+    if (service.__hooks?.onApplicationStart) {
+      await service.__hooks.onApplicationStart()
+    }
```
adrien2p: todo: rm

packages/types/src/region/common.ts
```diff
@@ -1,11 +1,63 @@
...
+  currency: RegionCurrencyDTO
+  tax_code?: string
   tax_rate?: number
-  tax_code?: string | null
-  gift_cards_taxable?: boolean
-  automatic_taxes?: boolean
-  tax_provider_id?: string | null
-  metadata?: Record<string, unknown>
-  includes_tax?: boolean
+  tax_provider_id?: string
```
srindom: tax related fields can be removed

packages/region/src/models/country.ts
```diff
@@ -0,0 +1,54 @@
...
+  num_code: number
+
+  @Property({ columnType: "text" })
+  name: string
+
+  @Property({ columnType: "text" })
+  display_name: string
+
+  @Property({ columnType: "text", nullable: true })
+  region_id: string | null
```
adrien2p: q: on prop we should have = null right?

packages/region/src/services/region-module.ts
adrien2p: todo: this can be removed, it will be infered, this option is more when wehave a tricky word or we want to force the values

packages/region/src/services/region-module.ts
adrien2p: todo: same

packages/region/src/services/region-module.ts
```diff
@@ -0,0 +1,230 @@
...
+  async create(
+    data: CreateRegionDTO,
+    sharedContext?: Context
+  ): Promise<RegionDTO>
+  @InjectManager("baseRepository_")
+  async create(
+    data: CreateRegionDTO | CreateRegionDTO[],
+    @MedusaContext() sharedContext: Context = {}
+  ): Promise<RegionDTO | RegionDTO[]> {
+    const input = Array.isArray(data) ? data : [data]
```
adrien2p, line 98: suggestion: same as for [phil](https://github.com/medusajs/medusa/pull/6314/files#r1477841729), as you like

packages/region/src/services/region-module.ts
```diff
@@ -0,0 +1,230 @@
...
+
+    const result = await this.regionService_.update(input, sharedContext)
+
+    return await this.baseRepository_.serialize<RegionDTO[]>(
+      Array.isArray(data) ? result : result[0],
+      {
+        populate: true,
+      }
```
adrien2p, line 158: suggestion: same as above, phil open a pr for the update, I forgot in the base class

packages/types/src/region/mutations.ts
```diff
@@ -0,0 +1,27 @@
+import { RegionCurrencyDTO } from "./common"
+
+export interface CreateRegionDTO {
+  name: string
+  currency_code: string
+  currency?: RegionCurrencyDTO
+  tax_code?: string
+  tax_rate?: number
+  tax_provider_id?: string
```
adrien2p, line 9: q: will the tax provider ids stay here?

packages/types/src/region/service.ts
```diff
@@ -0,0 +1,69 @@
...
+    filters?: FilterableRegionCountryProps,
+    config?: FindConfig<RegionCountryDTO>,
+    sharedContext?: Context
+  ): Promise<RegionCountryDTO[]>
+
+  listCurrencies(
+    filters?: FilterableRegionCurrencyProps,
+    config?: FindConfig<RegionCurrencyDTO>,
+    sharedContext?: Context
+  ): Promise<RegionCurrencyDTO[]>
```
adrien2p: q: should we add retrieve for those too?

PR #6320, opened 2024-02-05
feat(region): Add migration

olivermrbl:


packages/region/integration-tests/__tests__/region-module.spec.ts
```diff
@@ -63,7 +63,6 @@ describe("Region Module Service", () => {
       expect.objectContaining({
         id: createdRegion.id,
         name: "Europe",
-        currency_code: "EUR",
```
srindom, line 66: question: wouldn't this field still be on the data model? That would in any case be ideal.

olivermrbl: Should be good now.

srindom (approved):
LGTM!

PR #6322, opened 2024-02-05
feat(region): Add admin region get + list endpoints

olivermrbl:
**What**
Add `GET /admin/regions`
Add `GET /admin/regions/:id`

Blocked by #6320


olivermrbl:
@carlos-r-l-rodrigues, @adrien2p – it seems we have an issue with loaders running before migrations are applied. In the region module, we want to create default currencies and countries in a loader, but because the migration is yet to be applied, this fails. Do you have any suggestions for how we should proceed?

If my assumption is incorrect and I am simply missing a step, please do let me know.

integration-tests/environment-helpers/use-db.js
```diff
@@ -172,18 +175,30 @@ module.exports = {
       const pgConnection = await pgConnectionLoader({ configModule, container })
       instance.setPgConnection(pgConnection)
 
-      const { runMigrations } = await medusaAppLoader(
-        { configModule, container },
-        { registerInContainer: false }
-      )
+      const migrationPromises = []
+      for (const [, options] of Object.entries(configModule.modules)) {
```
olivermrbl: @adrien2p, something like this for running module migrations before loaders?

cc @carlos-r-l-rodrigues

carlos-r-l-rodrigues: What is the issue happening here? 
if we do that, we'll miss link-modules' migration.

olivermrbl: `MedusaApp` will run module loaders before migrations. I am creating default currencies and countries in the loader and this will fail with a "table doesn't exist" error.

olivermrbl: But I can see several suites are failing now. Let me rethink this approach and revert back. If you have any input, I am curious to hear.

adrien2p:
@olivermrbl you also still have an issue with the migration, the tax_rate, tax_code etc are still non nullable column, meaning that you can't create a region from the module 👍

olivermrbl:
> you also still have an issue with the migration, the tax_rate, tax_code etc are still non nullable column, meaning that you can't create a region from the module 👍

This is a bit of a tricky issue actually. We cannot do `ALTER COLUMN IF EXISTS` in Postgres to drop the null constraints on the columns, so I've had to come up with a "nasty" script instead. Curious to hear what you think. I can't immediately see other, better ways.

carlos-r-l-rodrigues:
> We cannot do `ALTER COLUMN IF EXISTS` in Postgres to drop the null constraints on the columns.

I'm fine with that script. an alternative to it, would be a `deprecateColumn` helper that we can run at the beginning of the migration if the table exists.
```typescript
deprecateColumns(tableName, ...columns)
```

```SQL
-- Similar checks if the table exists
THEN
  alter table ${tableName} alter column ${column} drop not null;
  comment on column ${tableName}.${column} is 'Deprecated in Medusa 2.0';
END IF
```

adrien2p:
@olivermrbl @carlos-r-l-rodrigues you can just do `ALTER TABLE IF EXISTS tableNam ALTER COLUMN columnName DROP  NOT NULL;` and it will work even the constraint is not applied cause it is not a named constraint. On a named constraint it is possible to have `IF EXISTS`.

PS: the current script is more future proof indeed for both cases

carlos-r-l-rodrigues:
The table won't exist if it is a fresh install.

carlos-r-l-rodrigues:
Nevermind, my suggestion also won't work if there is a module table, and it is being migrated to another database to run the modules isolated. we have to check if the column is there anyway.

olivermrbl:
@adrien2p, can you summarise your thoughts on the script I added? I am not sure if you are OK with it or you are suggesting a different approach

adrien2p:
It is the best approach to support both cases. The module is isolated or shared with the core 👍

packages/medusa/src/api-v2/admin/regions/route.ts
```diff
@@ -0,0 +1,20 @@
...
+
+  const variables = { filters: req.filterableFields }
+
+  const queryObject = remoteQueryObjectFromString({
+    entryPoint: "region",
+    variables,
+    fields: defaultAdminRegionFields,
+  })
+
+  // TODO: Add count, offset, limit
```
srindom, line 16: assuming this will be part of a future PR?

packages/utils/src/common/alter-columns-helper.ts
```diff
@@ -0,0 +1,32 @@
+export function generatePostgresAlterColummnsStatement(
```
srindom: suggestion: maybe get if exists in here somehow to make it's purpose clearer

srindom (approved):

Issue #6264, opened 2024-01-30
Docs(Code Issue): Code Issue in /modules/products/admin/import-products


irab:
### Preliminary Checks

- [X] This issue is not a duplicate. Before opening a new issue, please search existing issues: https://github.com/medusajs/medusa/issues

### Issue Summary

**_Using latest Admin package release version: `"@medusajs/admin": "7.1.10"`_**

The [current cURL doco for importing products](https://docs.medusajs.com/modules/products/admin/import-products) specifies the API Token is provided with `-H 'Authorization: Bearer <API_TOKEN>'` but calls made with this header fail with an `Unauthorized` response. 

Calling the API with the following cURL header option works: `-H 'x-medusa-access-token: <API_TOKEN>'`

I can possibly raise a PR to fix the documentation if I get confirmation that this is the correct approach?

### How can this issue be resolved?

The API doco for cURL commands are updated to `-H 'x-medusa-access-token: <API_TOKEN>'` wherever `-H 'Authorization: Bearer <API_TOKEN>'` occurs.

### Are you interested in working on this issue?

- [X] I would like to fix this issue

shahednasser:
Thank you for opening this issue! That's true, it should be `-H 'x-medusa-access-token: <API_TOKEN>'`. I think this is a change we missed to make in the main docs but have made in the API reference.

Would you be up to make the change in all files under `www/apps/docs/content` as this is probably used in different places? It should be a quick find-and-replace

irab:
Yes, happy to ❤️

Will raise a PR in a bit.

PR #6326, opened 2024-02-06
Update all curl documentation examples and references with new x-medusa-access-token header

irab:
## What

This is to update incorrect documentation in regards to authentication to the Admin API - raised in https://github.com/medusajs/medusa/issues/6264.

## Why

Because the current documentation has been incorrect since the September 2023 release of [v1.17.0](https://github.com/medusajs/medusa/releases/tag/v1.17.0), which had breaking changes to API token usage.

## How

Simple search and replace. I was asked to replace occurrences under `www/apps/docs/content/` but there were also additional places where I thought references should also be updated:

- `packages/medusa/src/api/`
- `www/apps/api-reference/`

Feel free to revert them as needed.

There is also some inconsistency between the format shown in examples e.g. `<API_TOKEN>` vs `{api_token}` vs `{access_token}`.

I have kept the format the same in all cases as the original, as surrounding documentation text would not have format updated as well. I suggest maybe reviewing the documentation and keeping to a consistent format e.g. `<API_TOKEN>`.

 
## Testing 

I have not tested these changes. I would assume the `packages/medusa/src/api/` changes may need more thorough testing?


www/apps/docs/content/modules/products/storefront/use-categories.mdx
shahednasser, line 1: Those API routes are store routes so please revert the changes.

shahednasser (changes requested):
Thanks for your contribution! A couple of things to note:

- This header is only used for admin API Routes. No changes should be done to store API Routes. So, please revert the changes under the `packages/medusa/src/api/routes/store` directory.
- Files under the `www/apps/api-reference/specs` directory are generated automatically, so make sure to revert those changes as well. Instead, you can run the `yarn openapi:generate` command in the root of the monorepository and commit the generated changes. Make sure to do this after you revert the changes mentioned in the first point.

I also added a comment below.

www/apps/docs/content/modules/products/storefront/use-categories.mdx
irab: Done

shahednasser (approved):
LGTM

PR #6327, opened 2024-02-06
feat(user, types): add invite and user properties + migration

pKorsholm:
**What**
- Add invite model 
- Populate user model

### DIFF
--- a/www/apps/docs/content/development/workflows/index.mdx
+++ b/www/apps/docs/content/development/workflows/index.mdx
@@ -40,29 +40,31 @@ Next, you can create a workflow using the `createWorkflow` function:
 
 ```ts title=src/workflows/hello-world.ts
 import {
-  createStep, 
+  createStep,
   StepResponse,
   createWorkflow,
 } from "@medusajs/workflows-sdk"
 
 type WorkflowOutput = {
-  message: string;
-};
+  message: string
+}
 
 const step1 = createStep("step-1", async () => {
   return new StepResponse(`Hello from step one!`)
 })
 
-const myWorkflow = createWorkflow<any, WorkflowOutput>(
+const myWorkflow = createWorkflow<unknown, WorkflowOutput>(
   "hello-world",
   function () {
-    const str1 = step1({})
+    const str1 = step1()
 
     return {
       message: str1,
     }
   }
 )
+
+export default myWorkflow
 ```
 
 This creates a `hello-world` workflow. When you create a workflow, it’s constructed but not executed yet.
@@ -202,29 +204,52 @@ import {
 // ...
 
 const myWorkflow = createWorkflow<
-  WorkflowInput, WorkflowOutput
->(
-  "hello-world",
-  function (input) {
-    const str1 = step1({})
-    const str2 = step2(input)
+  WorkflowInput,
+  WorkflowOutput
+>("hello-world", function (input) {
+  const str1 = step1()
+  const str2 = step2(input)
 
-    const result = transform(
-      {
-        str1,
-        str2,
-      },
-      (input) => ({
-        message: `${input.str1}\n${input.str2}`,
-      })
-    )
+  const result = transform(
+    {
+      str1,
+      str2,
+    },
+    (input) => ({
+      message: `${input.str1}\n${input.str2}`,
+    })
+  )
 
-    return result
-  }
-)
+  return result
+})
+
+export default myWorkflow
 ```
 
 Notice that to use the results of the steps, you must use the `transform` utility function. It gives you access to the real-time results of the steps once the workflow is executed.
+
+Then, pass the necessary input to the workflow when you execute it. For example, in the API Route executing the workflow:
+
+```ts title="src/api/store/custom/route.ts"
+import type {
+  MedusaRequest,
+  MedusaResponse,
+} from "@medusajs/medusa"
+import myWorkflow from "../../../workflows/hello-world"
+
+export async function GET(
+  req: MedusaRequest,
+  res: MedusaResponse
+) {
+  const { result } = await myWorkflow(req.scope).run({
+    input: {
+      name: req.validatedQuery.name as string,
+    },
+  })
+
+  res.send(result)
+}
+```
 
 If you execute the workflow again, you’ll see:
 
@@ -240,18 +265,22 @@ The compensation function only runs if an error occurs throughout the Workflow. 
 For example, change step one to add a compensation function and step two to throw an error:
 
 ```ts title=src/workflows/hello-world.ts
-const step1 = createStep("step-1", async () => {
-  const message = `Hello from step one!`
-    
-  console.log(message)
+const step1 = createStep(
+  "step-1",
+  async () => {
+    const message = `Hello from step one!`
 
-  return new StepResponse(message)
-}, async () => {
-  console.log("Oops! Rolling back my changes...")
-})
+    console.log(message)
+
+    return new StepResponse(message)
+  },
+  async () => {
+    console.log("Oops! Rolling back my changes...")
+  }
+)
 
 const step2 = createStep(
-  "step-2", 
+  "step-2",
   async ({ name }: WorkflowInput) => {
     throw new Error("Throwing an error...")
   }
@@ -317,7 +346,7 @@ const updateProduct = createStep(
         select: ["title", "description", "images"],
       }
     )
-    const product = productService.update(id, updateData)
+    const product = await productService.update(id, updateData)
 
     return new StepResponse(product, {
       id,
@@ -327,7 +356,7 @@ const updateProduct = createStep(
     const productService: ProductService = 
       context.container.resolve("productService")
 
-    productService.update(id, previousProductData)
+    await productService.update(id, previousProductData)
   }
 )
 ```
@@ -364,7 +393,7 @@ const myWorkflow = createWorkflow<WorkflowInput, WorkflowOutput>
 - Since the constructor function only defines how the workflow works, you can’t directly manipulate data within the function. To do that, you must use the `transform` function:
 
 ```ts
-1// Don't
+// Don't
 const myWorkflow = createWorkflow<
   WorkflowInput, 
   WorkflowOutput


packages/user/src/services/user-module.ts
```diff
@@ -17,20 +14,36 @@ import {
...
-const generateMethodForModels = []
-
-export default class UserModuleService<TUser extends User = User>
+const generateMethodForModels = [
+  {
+    model: Invite,
+    singular: "Invite",
+    plural: "Invites",
+  },
+]
```
srindom:
```suggestion
const generateMethodForModels = [Invite]
```

packages/user/src/services/user-module.ts
```diff
@@ -17,20 +14,36 @@ import {
...
   extends ModulesSdkUtils.abstractModuleServiceFactory<
     InjectedDependencies,
-    UserDTO,
-    {}
+    UserTypes.UserDTO,
+    {
+      Invite: {
+        dto: UserTypes.InviteDTO
+        singular: "Invite"
+        plural: "Invites"
```
srindom:
```suggestion
```

packages/user/src/models/invite.ts
```diff
@@ -0,0 +1,74 @@
...
+
+  @PrimaryKey({ columnType: "text" })
+  id: string
+
+  @Index({
+    name: "invite_user_identifier_index",
+    expression: `create unique index "invite_user_identifier_index" on "invite" ("user_identifier") where deleted_at is null;`,
+  })
+  @Property({ columnType: "text" })
+  user_identifier: string
```
srindom: think it is reasonable to just stick with `email`. Flow would be something like this:

1. User invites new user - enters their email.
2. New user receives invite link on email with a token.
3. If user wants to authenticate with e.g. Google they complete sign in flow.
4. Once authenticated with Google they pass the invite token as part of the user creation call.
5. AuthUser is associated with the new User. 

I.e., the email is not used to tie a the AuthUser's entity id, but simply ensures there is a trusted destination the invite token can be sent to.

srindom: Email-password could also be facilitated through the user creation request with an additional step in the workflow to create the auth user.

packages/types/src/user/common.ts
```diff
@@ -1,5 +1,38 @@
+import { DateComparisonOperator } from "../common"
+
 export type UserDTO = {
```
adrien2p: suggestion: Privilege interface over type here

packages/types/src/user/common.ts
```diff
@@ -1,5 +1,38 @@
...
+
+export type FilterableUserProps = {
+  id?: string | string[]
+  email?: string | string[]
+  first_name?: string | string[]
+  last_name?: string | string[]
 }
 
-export type FilterableUserProps = {}
+export type InviteDTO = {
```
adrien2p: suggestion: Privilege interface over type here

packages/types/src/user/common.ts
```diff
@@ -1,5 +1,38 @@
...
+  id: string
+  email: string
+  accepted: boolean
+  token: string
+  expires_at: Date
+  metadata?: Record<string, unknown> | null
+  created_at: Date
+  updated_at: Date
+  deleted_at?: Date | null
```
adrien2p: todo: The DTO is meant to represent the data that can transfert, deleted at and metadata should be present and non optional. In a later future, we can adapt the response type of our method by dynamically pick from the DTO based on the select and relations

adrien2p: same for the UserDTO

packages/types/src/user/mutations.ts
```diff
@@ -1,6 +1,34 @@
 export interface CreateUserDTO {
   id?: string
```
adrien2p: q: do we allow a user to set a specific id upon the user creation?

packages/types/src/user/common.ts
```diff
@@ -1,5 +1,38 @@
...
+}
+
+export type FilterableInviteProps = {
```
adrien2p: todo: interface and extends BaseFilterable to allow $and and $or as well

packages/types/src/user/mutations.ts
```diff
@@ -1,6 +1,34 @@
...
   id: string
+  email?: string
+  first_name?: string
+  last_name?: string
+  avatar_url?: string
+  metadata?: Record<string, unknown>
+}
+
+export interface CreateInviteDTO {
+  id?: string
```
adrien2p: q: same as the previous one

packages/types/src/user/mutations.ts
```diff
@@ -1,6 +1,34 @@
 export interface CreateUserDTO {
   id?: string
+  email: string
+  first_name?: string
+  last_name?: string
+  avatar_url?: string
+  metadata?: Record<string, unknown>
 }
 export interface UpdateUserDTO {
```
adrien2p: suggestion: `extends Partial<CreateUserDTO>`

packages/types/src/user/mutations.ts
```diff
@@ -1,6 +1,34 @@
...
+export interface CreateInviteDTO {
+  id?: string
+  email?: string
+  accepted?: boolean
+  token: string
+  expires_at: Date
+  metadata?: Record<string, unknown>
+}
+
+export interface UpdateInviteDTO {
```
adrien2p: suggestion: `extends Partial<CreateInviteDTO>`

packages/types/src/user/service.ts
```diff
@@ -33,4 +38,44 @@ export interface IUserModuleService extends IModuleService {
...
+    sharedContext?: Context
+  ): Promise<InviteDTO[]>
+
+  listAndCountInvites(
+    filters?: FilterableUserProps,
+    config?: FindConfig<InviteDTO>,
+    sharedContext?: Context
+  ): Promise<[InviteDTO[], number]>
+
+  createInvite(
```
adrien2p: todo: plural

packages/types/src/user/service.ts
```diff
@@ -33,4 +38,44 @@ export interface IUserModuleService extends IModuleService {
...
+    data: CreateInviteDTO[],
+    sharedContext?: Context
+  ): Promise<InviteDTO[]>
+
+  createInvite(
```
adrien2p: todo: same as previoujs

packages/types/src/user/service.ts
```diff
@@ -33,4 +38,44 @@ export interface IUserModuleService extends IModuleService {
...
+  updateInvite(
+    data: UpdateInviteDTO[],
+    sharedContext?: Context
+  ): Promise<InviteDTO[]>
+
+  updateInvite(
+    data: UpdateInviteDTO,
+    sharedContext?: Context
+  ): Promise<InviteDTO>
+
```
adrien2p: todo: same as previous

packages/user/integration-tests/__tests__/services/module/invite.spec.ts
```diff
@@ -0,0 +1,251 @@
...
+  afterAll(async () => {
+    await shutdownFunc()
+  })
+
+  describe("listInvites", () => {
+    it("should list invites", async () => {
+      await createInvites(testManager, defaultInviteData)
+
+      const invites = await service.listInvites()
+      const serialized = JSON.parse(JSON.stringify(invites))
```
adrien2p: todo: this shouldn't be necessary. can be removed

packages/user/integration-tests/__tests__/services/module/invite.spec.ts
```diff
@@ -0,0 +1,251 @@
...
+        }),
+      ])
+    })
+  })
+
+  describe("listAndCountInvites", () => {
+    it("should list and count invites", async () => {
+      await createInvites(testManager, defaultInviteData)
+      const [invites, count] = await service.listAndCountInvites()
+      const serialized = JSON.parse(JSON.stringify(invites))
```
adrien2p: todo: same as previous one

packages/user/integration-tests/__tests__/services/module/invite.spec.ts
```diff
@@ -0,0 +1,251 @@
...
+        }),
+      ])
+    })
+  })
+
+  describe("retrieveInvite", () => {
+    const id = "1"
+
+    it("should return an invite for the given id", async () => {
+      await createInvites(testManager, [
```
adrien2p: suggestion: we could reuse `defaultInviteData`

packages/user/integration-tests/__tests__/services/module/invite.spec.ts
```diff
@@ -0,0 +1,251 @@
...
+    })
+
+    it("should throw an error when an invite with the given id does not exist", async () => {
+      let error
+
+      try {
+        await service.retrieveInvite("does-not-exist")
+      } catch (e) {
+        error = e
+      }
```
adrien2p:
```suggestion
        const error = await service.retrieveInvite("does-not-exist").catch(e => e)
```

adrien2p: if you like you can use that everywhere

packages/user/integration-tests/__tests__/services/module/invite.spec.ts
```diff
@@ -0,0 +1,251 @@
...
+          email: "user_1@test.com",
+          token: "test",
+          expires_at: expireDate,
+        },
+      ])
+      const invite = await service.retrieveInvite(id, {
+        select: ["id"],
+      })
+
+      const serialized = JSON.parse(JSON.stringify(invite))
```
adrien2p: todo: rm

packages/user/integration-tests/__tests__/services/module/invite.spec.ts
```diff
@@ -0,0 +1,251 @@
...
+        await service.retrieveInvite(undefined as unknown as string)
+      } catch (e) {
+        error = e
+      }
+
+      expect(error.message).toEqual("invite - id must be defined")
+    })
+
+    it("should return invite based on config select param", async () => {
+      await createInvites(testManager, [
```
adrien2p: suggestion: we could reuse the default ones?

packages/user/integration-tests/__tests__/services/module/invite.spec.ts
```diff
@@ -0,0 +1,251 @@
...
+        id,
+      })
+    })
+  })
+
+  describe("deleteInvite", () => {
+    const id = "1"
+
+    it("should delete the Invite given an id successfully", async () => {
+      await createInvites(testManager, [
```
adrien2p: suggestion: same here

packages/user/src/migrations/Migration20240208015217.ts
```diff
@@ -0,0 +1,20 @@
...
+
+    this.addSql('create table "user" ("id" text not null, "first_name" text null, "last_name" text null, "email" text not null, "avatar_url" text null, "metadata" jsonb null, "created_at" timestamptz not null default now(), "updated_at" timestamptz not null default now(), "deleted_at" timestamptz null, constraint "user_pkey" primary key ("id"));');
+    this.addSql('create index "IDX_user_deleted_at" on "user" ("deleted_at");');
+  }
+
+  async down(): Promise<void> {
+    this.addSql('drop table if exists "invite" cascade;');
+
+    this.addSql('drop table if exists "user" cascade;');
+  }
```
adrien2p: todo: we are missing plenty of `if exists` `if not exists`, you can also look at this [file](https://github.com/medusajs/medusa/blob/develop/packages/fulfillment/mikro-orm.config.dev.ts) where there is a custom migrator which should handle a good portion of those missing ones.

packages/user/src/migrations/Migration20240208015217.ts
```diff
@@ -0,0 +1,20 @@
+import { Migration } from '@mikro-orm/migrations';
+
+export class Migration20240208015217 extends Migration {
+
+  async up(): Promise<void> {
+    this.addSql('create table "invite" ("id" text not null, "email" text not null, "accepted" boolean not null default false, "token" text not null, "expires_at" timestamptz not null, "metadata" jsonb null, "created_at" timestamptz not null default now(), "updated_at" timestamptz not null default now(), "deleted_at" timestamptz null, constraint "invite_pkey" primary key ("id"));');
+    this.addSql('create index "IDX_invite_deleted_at" on "invite" ("deleted_at");');
+    this.addSql('create unique index "IDX_invite_email" on "invite" ("email") where deleted_at is null;;');
```
adrien2p: todo: double `;` this migration can't work. Would be nice to run the migrations locally to test them everytime we generate them as a common practice

packages/user/src/models/invite.ts
```diff
@@ -0,0 +1,74 @@
...
+@Filter(DALUtils.mikroOrmSoftDeletableFilterOptions)
+export default class Invite {
+  [OptionalProps]: OptionalFields
+
+  @PrimaryKey({ columnType: "text" })
+  id: string
+
+  @Index({
+    name: "invite_user_identifier_index",
+    expression: `create unique index "IDX_invite_email" on "invite" ("email") where deleted_at is null;`,
```
adrien2p: todo: I suggest you to use this [utils](https://github.com/medusajs/medusa/blob/develop/packages/utils/src/common/create-psql-index-helper.ts) as mentioned in slack. examples [here](https://github.com/medusajs/medusa/blob/develop/packages/fulfillment/src/models/fulfillment-item.ts#L23)

packages/user/src/models/invite.ts
```diff
@@ -0,0 +1,74 @@
...
+  })
+  @Property({ columnType: "text" })
+  email: string
```
adrien2p, line 61: todo: email is probably a column that can be used for querying, I suggest to have an index on it

adrien2p: maybe same for token let me know

packages/user/src/models/user.ts
```diff
@@ -3,16 +3,63 @@ import {
...
   @PrimaryKey({ columnType: "text" })
   id!: string
 
+  @Property({ columnType: "text", nullable: true })
+  first_name: string
+
+  @Property({ columnType: "text", nullable: true })
+  last_name: string
+
+  @Property({ columnType: "text" })
```
adrien2p, line 53: todo: index

packages/user/src/models/user.ts
```diff
@@ -3,16 +3,63 @@ import {
   Entity,
   OnInit,
   PrimaryKey,
+  Property,
```
adrien2p, line 6: todo: this file, convention issues

packages/user/src/models/user.ts
```diff
@@ -3,16 +3,63 @@ import {
...
+
+  @Property({
+    onCreate: () => new Date(),
+    onUpdate: () => new Date(),
+    columnType: "timestamptz",
+    defaultRaw: "now()",
+  })
+  updated_at: Date
+
+  @Index({ name: "IDX_user_deleted_at" })
```
adrien2p: todo: index building based on previous example

packages/user/src/models/invite.ts
```diff
@@ -0,0 +1,74 @@
...
+
+  @Property({
+    onCreate: () => new Date(),
+    onUpdate: () => new Date(),
+    columnType: "timestamptz",
+    defaultRaw: "now()",
+  })
+  updated_at: Date
+
+  @Index({ name: "IDX_invite_deleted_at" })
```
adrien2p: todo: index changes based on example

packages/user/src/services/user-module.ts
```diff
@@ -105,9 +124,66 @@ export default class UserModuleService<TUser extends User = User>
...
   ): Promise<TUser[]> {
     return await this.userService_.update(data, sharedContext)
   }
+
+  createInvite(
+    data: UserTypes.CreateInviteDTO[],
+    sharedContext?: Context
+  ): Promise<UserTypes.InviteDTO[]>
+  createInvite(
+    data: UserTypes.CreateInviteDTO,
```
adrien2p: todo: review naming based on convention and the comments on the interface

packages/user/src/services/user-module.ts
```diff
@@ -105,9 +124,66 @@ export default class UserModuleService<TUser extends User = User>
...
+    sharedContext?: Context
+  ): Promise<UserTypes.InviteDTO>
+
+  @InjectManager("baseRepository_")
+  async createInvite(
```
adrien2p: todo: wrong decorator used

packages/user/src/services/user-module.ts
```diff
@@ -105,9 +124,66 @@ export default class UserModuleService<TUser extends User = User>
...
+    data: UserTypes.UpdateInviteDTO[],
+    sharedContext?: Context
+  ): Promise<UserTypes.InviteDTO[]>
+  updateInvite(
+    data: UserTypes.UpdateInviteDTO,
+    sharedContext?: Context
+  ): Promise<UserTypes.InviteDTO>
+
+  @InjectManager("baseRepository_")
+  async updateInvite(
```
adrien2p: todo: wrong decorator used

.changeset/tame-kings-battle.md
```diff
@@ -0,0 +1,8 @@
+---
+"@medusajs/core-flows": patch
+"@medusajs/medusa": patch
+"@medusajs/types": patch
+"@medusajs/utils": patch
```
adrien2p: q: I think the utils shouldn't be here right?

integration-tests/plugins/__tests__/users/create-user.spec.ts
```diff
@@ -0,0 +1,60 @@
...
+    const body = {
+      email: "test_member@test.com",
+    }
+
+    const response = await api.post(`/admin/users`, body, adminHeaders)
+
+    expect(response.status).toEqual(200)
+    expect(response.data).toEqual({
+      user: expect.objectContaining({
+        email: "test_member@test.com",
```
adrien2p: suggestion: In those context, to facilitate future changes or copy, I would use `body.email` wdyt?

integration-tests/plugins/__tests__/users/delete-user.spec.ts
```diff
@@ -0,0 +1,71 @@
...
+    expect(response.status).toEqual(200)
+    expect(response.data).toEqual({
+      id: user.id,
+      object: "user",
+      deleted: true,
+    })
+
+    const { response: deletedResponse } = await api
+      .get(`/admin/users/${user.id}`, adminHeaders)
+      .catch((e) => e)
```
adrien2p, line 66: todo: I think we can rm the catch since the test would fail if the call fail

packages/core-flows/src/user/steps/update-users.ts
```diff
@@ -0,0 +1,38 @@
...
+export const updateUsersStepId = "update-users-step"
+export const updateUsersStep = createStep(
+  updateUsersStepId,
+  async (input: UpdateUserDTO[], { container }) => {
+    const service: IUserModuleService = container.resolve(
+      ModuleRegistrationName.USER
+    )
+
+    const originalUsers = await service.list({
+      id: input.map((u) => u.id),
```
adrien2p, line 18: suggestion: For safety guard, as I have seen it occured in the modules, I would store the ids separately and then id ids.length then call the lists and the update wdyt?

packages/core-flows/src/user/workflows/create-users.ts
```diff
@@ -0,0 +1,15 @@
+import { CreateUserDTO, UserDTO } from "@medusajs/types"
+import { WorkflowData, createWorkflow } from "@medusajs/workflows-sdk"
+import { createUsersStep } from "../steps"
+
+interface CreateUsersInput {
```
adrien2p: q: This interfaces for me should be in the types package. Because this is a public facing API. As a user, I could be consuming a workflow and be able to type my input before passing it to the workflow. I think it should moves in general. wdyt? Also, when we will allow for extension the users might need to access this interface so it should be at least exported.

Same question for all workflows interfaces, we did start to put them in the types package from memories

packages/core-flows/src/user/workflows/delete-users.ts
```diff
@@ -0,0 +1,14 @@
+import { WorkflowData, createWorkflow } from "@medusajs/workflows-sdk"
+import { deleteUsersStep } from "../steps"
+
+interface WorkflowInput {
+  ids: string[]
+}
+
+export const deleteUsersWorkflowId = "delete-user"
+export const deleteUsersWorkflow = createWorkflow(
+  "delete-user",
```
adrien2p: todo: use the const id above

packages/types/src/user/common.ts
```diff
@@ -1,5 +1,39 @@
...
+export interface UserDTO {
   id: string
+  email: string
+  first_name?: string | null
+  last_name?: string | null
+  avatar_url?: string | null
+  metadata?: Record<string, unknown> | null
+  created_at: Date
+  updated_at: Date
+  deleted_at?: Date | null
```
adrien2p: suggestion: IMO, the DTO should represent the data corresponding to the User, properties should not be optional, if you retrieve the user all properties will be returned and eventually some will be null. Then comes the question of dynamic DTO's based on the select and relations but that is another topic :)

cc @olivermrbl what do you think?

packages/types/src/user/common.ts
```diff
@@ -1,5 +1,39 @@
...
+}
+
+export interface FilterableUserProps extends BaseFilterable<UserDTO> {
```
adrien2p:
```suggestion
export interface FilterableUserProps extends BaseFilterable<FilterableUserProps> {
```

packages/types/src/user/common.ts
```diff
@@ -1,5 +1,39 @@
...
+  id?: string | string[]
+  email?: string | string[]
+  first_name?: string | string[]
+  last_name?: string | string[]
 }
 
-export type FilterableUserProps = {}
+export interface InviteDTO {
```
adrien2p, line 24: same comment as before

packages/types/src/user/common.ts
```diff
@@ -1,5 +1,39 @@
...
+  accepted: boolean
+  token: string
+  expires_at: Date
+  metadata?: Record<string, unknown> | null
+  created_at: Date
+  updated_at: Date
+  deleted_at?: Date | null
+}
+
+export interface FilterableInviteProps {
```
adrien2p:
```suggestion
export interface FilterableInviteProps extends BaseFilterable<FilterableInviteProps> {
```

packages/types/src/user/mutations.ts
```diff
@@ -1,6 +1,22 @@
 export interface CreateUserDTO {
-  id?: string
+  email: string
+  first_name?: string
+  last_name?: string
+  avatar_url?: string
+  metadata?: Record<string, unknown>
```
adrien2p: todo: nullable fields should be marked as nullable

packages/types/src/user/mutations.ts
```diff
@@ -1,6 +1,22 @@
...
 }
-export interface UpdateUserDTO {
+export interface UpdateUserDTO extends Partial<Omit<CreateUserDTO, "email">> {
+  id: string
+}
+
+export interface CreateInviteDTO {
```
adrien2p, line 12: todo: same comment as before

packages/user/integration-tests/__tests__/services/module/invite.spec.ts
```diff
@@ -0,0 +1,207 @@
+import { IUserModuleService } from "@medusajs/types/dist/user"
```
adrien2p, line 1: suggestion: I took the part of not testing a service that do not have a customisation. It will be tested as a side effect of testing the module service. Since the internal service derived from the abstraction, no need if no customisation. If something would alter the behaviour of those, then the integration of the modules service should be impacted as well.

packages/user/integration-tests/__tests__/services/module/user.spec.ts
```diff
@@ -8,6 +8,17 @@ import { initModules } from "medusa-test-utils"
 
 jest.setTimeout(30000)
 
+const defaultUserData = [
```
adrien2p, line 11: same comment as before

packages/user/src/migrations/Migration20240213092656.ts
```diff
@@ -0,0 +1,54 @@
...
+export class Migration20240213092656 extends Migration {
+  async up(): Promise<void> {
+    this.addSql(
+      'create table if not exists "invite" ("id" text not null, "email" text not null, "accepted" boolean not null default false, "token" text not null, "expires_at" timestamptz not null, "metadata" jsonb null, "created_at" timestamptz not null default now(), "updated_at" timestamptz not null default now(), "deleted_at" timestamptz null, constraint "invite_pkey" primary key ("id"));'
+    )
+    this.addSql(
+      'alter table "invite" add column if not exists "email" text not null;'
+    )
+
+    this.addSql(`
```
adrien2p: todo: should we use `alter-columns-helper.ts` here

packages/user/src/migrations/Migration20240213092656.ts
```diff
@@ -0,0 +1,54 @@
...
+        FROM information_schema.columns
+        WHERE table_name='invite' and column_name='user_email')
+      THEN
+          ALTER TABLE "public"."invite" alter column "user_email" drop not null;
+      END IF;
+    END $$;
+`)
+
+    this.addSql(
+      'create index if not exists "IDX_invite_deleted_at" on "invite" ("deleted_at");'
```
adrien2p: todo: where deleted_at is not null

packages/user/src/migrations/Migration20240213092656.ts
```diff
@@ -0,0 +1,54 @@
...
+    )
+    this.addSql(
+      'alter table "user" add column if not exists "avatar_url" text null;'
+    )
+
+    this.addSql(
+      'CREATE INDEX IF NOT EXISTS "IDX_user_email" ON "user" (email) WHERE deleted_at IS NULL;'
+    )
+    this.addSql(
+      'CREATE INDEX IF NOT EXISTS "IDX_user_deleted_at" ON "user" (deleted_at);'
```
adrien2p: same as the previous one

packages/user/src/migrations/Migration20240213092656.ts
```diff
@@ -0,0 +1,54 @@
...
+    )
+  }
+
+  async down(): Promise<void> {
+    this.addSql('drop table if exists "invite" cascade;')
+
+    this.addSql('drop table if exists "user" cascade;')
```
adrien2p: todo: very dangerous, the down should revert what have been done in the up, here the risk is that if you run the migration of the module and immediately revert, the you lose all the table and data from the core 

cc @olivermrbl @carlos-r-l-rodrigues something to apply to all modules IMO

adrien2p: probably copilot could generate the invert of the up

packages/user/src/models/invite.ts
```diff
@@ -0,0 +1,99 @@
...
```
adrien2p: todo: apply partial index

packages/user/src/models/invite.ts
```diff
@@ -0,0 +1,99 @@
...
+  @Property({ columnType: "timestamptz", nullable: true })
+  deleted_at?: Date
```
adrien2p:
```suggestion
  deleted_at?: Date | null = null
```

packages/user/src/models/user.ts
```diff
@@ -3,16 +3,87 @@ import {
...
 @Entity()
+@Filter(DALUtils.mikroOrmSoftDeletableFilterOptions)
 export default class User {
+  [OptionalProps]?: OptionalFields
+
   @PrimaryKey({ columnType: "text" })
   id!: string
 
+  @Property({ columnType: "text", nullable: true })
+  first_name: string
```
adrien2p: todo: everywhere where it is needed `| null = null`

packages/user/src/models/user.ts
```diff
@@ -3,16 +3,87 @@ import {
...
 
 @Entity()
+@Filter(DALUtils.mikroOrmSoftDeletableFilterOptions)
 export default class User {
+  [OptionalProps]?: OptionalFields
+
   @PrimaryKey({ columnType: "text" })
   id!: string
 
+  @Property({ columnType: "text", nullable: true })
```
adrien2p, line 47: todo(maybe): if a user is searchable by text search then add GIN index on first and last name. but you tell me

packages/user/src/models/user.ts
```diff
@@ -3,16 +3,87 @@ import {
...
+  tableName: "user",
+  columns: "email",
+  where: "deleted_at IS NULL",
+})
+
+const userDeletedAtIndexName = "IDX_user_deleted_at"
+const userDeletedAtIndexStatement = createPsqlIndexStatementHelper({
+  name: userDeletedAtIndexName,
+  tableName: "user",
+  columns: "deleted_at",
```
adrien2p, line 28: todo: where deleted_at is not null

packages/user/src/services/user-module.ts
```diff
@@ -41,29 +46,37 @@ export default class UserModuleService<TUser extends User = User>
...
+  @InjectTransactionManager("baseRepository_")
   async create(
-    data: CreateUserDTO[] | CreateUserDTO,
+    data: UserTypes.CreateUserDTO[] | UserTypes.CreateUserDTO,
     @MedusaContext() sharedContext: Context = {}
   ): Promise<UserTypes.UserDTO | UserTypes.UserDTO[]> {
     const input = Array.isArray(data) ? data : [data]
 
-    const users = await this.create_(input, sharedContext)
+    const users = await this.userService_.create(input, sharedContext)
```
adrien2p, line 79: comment: Just to ensure everybody is aware, this does not work when relations are involved. In that case it is better to have the create calling the create_ and then serialize in the create. Otherwise the populate could not function properly. wdyt?

cc @olivermrbl

adrien2p: same for update

packages/user/src/services/user-module.ts
```diff
@@ -103,11 +114,60 @@ export default class UserModuleService<TUser extends User = User>
...
+
+    const invites = await this.inviteService_.create(input, sharedContext)
+
+    const serializedInvites = await this.baseRepository_.serialize<
+      UserTypes.InviteDTO[]
+    >(invites, {
+      populate: true,
+    })
+
+    return serializedInvites
```
adrien2p: todo: missing Array.isArray etc

packages/core-flows/src/user/workflows/create-users.ts
olivermrbl: Yeah, we can start to put these in the types package. I don't have any strong opinion on whether we do it now or later. A larger clean-up of types is due, but that could all be tackled after 2.0rc.

packages/types/src/user/common.ts
olivermrbl: Good point, but I think this is in the same category as my other comment. I would leave it up to @pKorsholm to decide if we are tackling it now or later.

packages/user/src/services/user-module.ts
olivermrbl: I am not sure I understand. As long as we are not doing anything but creating/updating, shouldn't this be fine?

.changeset/tame-kings-battle.md
pKorsholm: yes, leftover from before extracting the psql index fix

packages/user/src/migrations/Migration20240213092656.ts
pKorsholm: well, the up creates the table if it isn't there, then running down wouldn't perform the expected behavior either, a different option would be to create tables like `module_user` instead of `user`? wdyt?

packages/user/src/services/user-module.ts
pKorsholm: That is handled in the create method of the abstract internal service factory

olivermrbl:
@adrien2p, would it be possible to have you review this one? Would like to see if we can get #6395 merged today too.

packages/types/src/user/common.ts
```diff
@@ -1,5 +1,41 @@
...
+export interface InviteDTO {
+  id: string
+  email: string
+  accepted: boolean
+  token: string
+  expires_at: Date
+  metadata: Record<string, unknown> | null
+  created_at: Date
+  updated_at: Date
+  deleted_at?: Date | null
```
adrien2p: nit: we can remove the ?

adrien2p (approved):
LGTM

PR #6328, opened 2024-02-06
feat(fulfillment): Initialize models work

adrien2p:
**What**
Initialize work on the fulfillment module entities.

This pr finally also include the indexes as i was working on some utilities i though it would make sense to test them directly.

Also this pr add a new utility to generate proper index for our entity properties. It also include a new monkey patch for the migration generator to handle most of if exists/not exists cases. The monkey patch is a workaround the fact that the transpilation does work well with the ECMA used by mikro orm and therefore we end up with some constructor issue when mikro orm try to instanciate the custom generator class extension.

**Comment**

- The rule part will likely evolved when we reach the point of rule filtering based data, so no need for details review I believe

FIXES CORE-1714
FIXES CORE-1715
FIXES CORE-1718
FIXES CORE-1722
FIXES CORE-1723

Current schema diagram
fulfillment-models


adrien2p:
@olivermrbl I think now is a good time to have a look to the preliminary entities work to be sure it is align with the desing. Of course it might evolve as we move forward. Also, there is a comment in regards to the fulfillment entities part in the figma document, so I have created those entities but I ve left them out for now. We can discuss that part too.
Everything should be explained in the description

packages/fulfillment/src/models/address.ts
```diff
@@ -0,0 +1,88 @@
...
+
+@Entity({ tableName: "fulfillment_address" })
+export default class Address {
+  [OptionalProps]: OptionalAddressProps
+
+  @PrimaryKey({ columnType: "text" })
+  id!: string
+
+  @Property({ columnType: "text", nullable: true })
+  fulfillment_id: string | null = null
```
carlos-r-l-rodrigues, line 46: index

packages/fulfillment/src/models/fulfillment-item.ts
```diff
@@ -0,0 +1,70 @@
...
+  sku: string
+
+  @Property({ columnType: "text" })
+  barcode: string
+
+  @Property({ columnType: "numeric", serializer: Number })
+  quantity: number // TODO: probably allow big numbers here
+
+  @Property({ columnType: "text", nullable: true })
+  line_item_id: string | null = null
```
carlos-r-l-rodrigues, line 80: index

packages/fulfillment/src/models/shipping-option.ts
```diff
@@ -0,0 +1,111 @@
...
+  service_zone_id: string
+
+  @Property({ columnType: "text" })
+  shipping_profile_id: string
+
+  @Property({ columnType: "text" })
+  service_provider_id: string
+
+  @Property({ columnType: "text", nullable: true })
+  shipping_option_type_id: string | null = null
```
carlos-r-l-rodrigues, line 117: index

packages/fulfillment/src/models/shipping-option.ts
```diff
@@ -0,0 +1,111 @@
...
+    items: () => ShippingOptionPriceType,
+    default: ShippingOptionPriceType.CALCULATED,
+  })
+  price_type: ShippingOptionPriceType
+
+  @Property({ columnType: "text" })
+  service_zone_id: string
+
+  @Property({ columnType: "text" })
+  shipping_profile_id: string
```
carlos-r-l-rodrigues, line 103: index

packages/fulfillment/src/models/shipping-option.ts
```diff
@@ -0,0 +1,111 @@
...
+  name: string
+
+  @Enum({
+    items: () => ShippingOptionPriceType,
+    default: ShippingOptionPriceType.CALCULATED,
+  })
+  price_type: ShippingOptionPriceType
+
+  @Property({ columnType: "text" })
+  service_zone_id: string
```
carlos-r-l-rodrigues, line 96: index

packages/fulfillment/src/models/shipping-option.ts
```diff
@@ -0,0 +1,111 @@
...
+
+  @Property({ columnType: "text" })
+  shipping_profile_id: string
+
+  @Property({ columnType: "text" })
+  service_provider_id: string
```
carlos-r-l-rodrigues, line 110: index

packages/fulfillment/src/models/shipping-profile.ts
```diff
@@ -0,0 +1,65 @@
...
+@Entity()
+@Filter(DALUtils.mikroOrmSoftDeletableFilterOptions)
+export default class ShippingProfile {
+  [OptionalProps]?: ShippingProfileOptionalProps
+
+  @PrimaryKey({ columnType: "text" })
+  id: string
+
+  @Property({ columnType: "text" })
+  shipping_option_id: string
```
carlos-r-l-rodrigues: index

packages/fulfillment/src/models/address.ts
adrien2p: this model existence need to be validated first before I finish it. it is a bear entity for now until further validation

packages/fulfillment/src/models/shipping-option.ts
adrien2p: All indexes will be part of a separate PR once all the schema is validated

olivermrbl:
I think we should go with having Fulfillment, Address FulfillmentLabel, and FulfillmentLine in the DB too. 

The alternative approach we were considering was if we could map a fulfillment and the related concepts + operations to a long-running workflow, but I fear we will have to shoehorn it to make it work. It is not immediately obvious to me that the possible benefits of having it as a workflow outweigh the risks of not having it persisted in the database. 

I believe it would also make the Fulfillment Module less usable in a standalone context, as you would no longer have these concepts at your disposal. The workflow would become a strict requirement regardless of the mode you are running the module in.

Curious to hear what you think?



But I would need @srindom to pitch in here before we proceed with either approach.

adrien2p:
> I think we should go with having Fulfillment, Address FulfillmentLabel, and FulfillmentLine in the DB too.
> 
> The alternative approach we were considering was if we could map a fulfillment and the related concepts + operations to a long-running workflow, but I fear we will have to shoehorn it to make it work. It is not immediately obvious to me that the possible benefits of having it as a workflow outweigh the risks of not having it persisted in the database.
> 
> I believe it would also make the Fulfillment Module less usable in a standalone context, as you would no longer have these concepts at your disposal. The workflow would become a strict requirement regardless of the mode you are running the module in.
> 
> Curious to hear what you think?
> 
> But I would need @srindom to pitch in here before we proceed with either approach.

I agree to actually make them part of the module and be manageable through the module. That was what I thought was the idea but the note made me doubt. I ll wait seb response as well 👍 
Also, I am missing the return shipping options model (thanks @carlos-r-l-rodrigues for spotting it) I ll add it before we finalize that pr, in parallel, I am working on the preliminary version of the DTO's. The main idea is to have the preliminary foundation ready to start the work on the core logic. Then, all the rest (services logic, filtering, and eventual adjustments) will be part of a single pr to keep the context of everything when reviewing.

packages/fulfillment/src/models/fulfillment.ts
```diff
@@ -0,0 +1,91 @@
...
+  delivered_at: Date | null = null
+
+  @Property({ columnType: "jsonb", nullable: true })
+  data: Record<string, unknown> | null = null
+
+  @Property({ columnType: "text" })
+  provider_id: string
+
+  @Property({ columnType: "text", nullable: true })
+  shipping_option_id: string | null = null
```
srindom: note: `shipping_option_id` is nullable here to enable fulfillments without requiring a shipping option to exist ahead of time.

The critical pieces for fulfillment are provider_id and data.

packages/fulfillment/src/models/fulfillment.ts
```diff
@@ -0,0 +1,91 @@
...
+  PrimaryKey,
+  Property,
+} from "@mikro-orm/core"
+import { DAL } from "@medusajs/types"
+
+type FulfillmentOptionalProps = DAL.SoftDeletableEntityDateColumns
+
+@Entity()
+@Filter(DALUtils.mikroOrmSoftDeletableFilterOptions)
+export default class Fulfillment {
```
srindom, line 65: thought: `canceled_at` could make sense on this too.

packages/fulfillment/src/models/geo-zone.ts
```diff
@@ -0,0 +1,96 @@
...
+  })
+  updated_at: Date
+
+  @Index({ name: "IDX_geo_zone_deleted_at" })
+  @Property({ columnType: "timestamptz", nullable: true })
+  deleted_at: Date | null = null
+
+  @BeforeCreate()
+  onCreate() {
+    this.id = generateEntityId(this.id, " geozo")
```
srindom:
```suggestion
    this.id = generateEntityId(this.id, " fgz")
```

packages/fulfillment/src/models/geo-zone.ts
```diff
@@ -0,0 +1,96 @@
...
+  }
+
+  @OnInit()
+  onInit() {
+    this.id = generateEntityId(this.id, "geozo")
```
srindom:
```suggestion
    this.id = generateEntityId(this.id, "fgz")
```

packages/fulfillment/src/models/shipping-option-type.ts
```diff
@@ -0,0 +1,64 @@
...
+  })
+  updated_at: Date
+
+  @Index({ name: "IDX_shipping_option_type_deleted_at" })
+  @Property({ columnType: "timestamptz", nullable: true })
+  deleted_at: Date | null = null
+
+  @BeforeCreate()
+  onCreate() {
+    this.id = generateEntityId(this.id, "shoptty")
```
srindom:
```suggestion
    this.id = generateEntityId(this.id, "sotype")
```

packages/fulfillment/src/models/shipping-option-type.ts
```diff
@@ -0,0 +1,64 @@
...
+  }
+
+  @OnInit()
+  onInit() {
+    this.id = generateEntityId(this.id, "shoptty")
```
srindom:
```suggestion
    this.id = generateEntityId(this.id, "sotype")
```

packages/fulfillment/src/models/shipping-option.ts
```diff
@@ -0,0 +1,111 @@
...
+  })
+  updated_at: Date
+
+  @Index({ name: "IDX_shipping_option_deleted_at" })
+  @Property({ columnType: "timestamptz", nullable: true })
+  deleted_at: Date | null = null
+
+  @BeforeCreate()
+  onCreate() {
+    this.id = generateEntityId(this.id, "shopt")
```
srindom:
```suggestion
    this.id = generateEntityId(this.id, "so")
```

packages/fulfillment/src/models/shipping-option.ts
```diff
@@ -0,0 +1,111 @@
...
+  }
+
+  @OnInit()
+  onInit() {
+    this.id = generateEntityId(this.id, "shopt")
```
srindom:
```suggestion
    this.id = generateEntityId(this.id, "so")
```

srindom (commented):
Only have a few smaller comments and a note explaining why shipping_option_id is nullable in case someone is curious.

Wrt. a model vs a workflow I would definitely keep the model within the module.

packages/fulfillment/src/models/fulfillment.ts
adrien2p: yes I thought it was the case :) it gives more possibilities for the future I believe

adrien2p:
Thanks for review guys, I ll finalize the rest of the models that was standing by, and re review it before making it reviewable again 👍

packages/fulfillment/src/models/fulfillment.ts
```diff
@@ -0,0 +1,167 @@
...
+  createPsqlIndexStatementHelper({
+    name: fulfillmentShippingOptionIdIndexName,
+    tableName: "fulfillment",
+    columns: "shipping_option_id",
+    where: "deleted_at IS NULL",
+  })
+
+@Entity()
+@Filter(DALUtils.mikroOrmSoftDeletableFilterOptions)
+export default class Fulfillment {
```
carlos-r-l-rodrigues, line 65: I think we can add a `metadata` field here for quick customizations.

packages/fulfillment/src/models/service-provider.ts
```diff
@@ -0,0 +1,77 @@
...
+const deletedAtIndexStatement = createPsqlIndexStatementHelper({
+  name: deletedAtIndexName,
+  tableName: "service_provider",
+  columns: "deleted_at",
+  where: "deleted_at IS NOT NULL",
+})
+
+@Entity()
+@Filter(DALUtils.mikroOrmSoftDeletableFilterOptions)
+export default class ServiceProvider {
```
carlos-r-l-rodrigues, line 34: `metadata` field to add providers config related data?

packages/fulfillment/src/models/geo-zone.ts
```diff
@@ -0,0 +1,131 @@
...
+const cityIndexStatement = createPsqlIndexStatementHelper({
+  name: cityIndexName,
+  tableName: "geo_zone",
+  columns: "city",
+  where: "deleted_at IS NULL AND city IS NOT NULL",
+})
+
+@Entity()
+@Filter(DALUtils.mikroOrmSoftDeletableFilterOptions)
+export default class GeoZone {
```
carlos-r-l-rodrigues, line 60: do we plan to have native support for lat/long points? @srindom

if not it is probably fine to use the `metadata` field if someone wants to use [PostGIS](https://postgis.net/) to customize it.

packages/fulfillment/src/models/shipping-option-rule.ts
```diff
@@ -0,0 +1,87 @@
...
+  @Index({
+    name: deletedAtIndexName,
+    expression: deletedAtIndexStatement,
+  })
+  @Property({ columnType: "timestamptz", nullable: true })
+  deleted_at: Date | null = null
+
+  @BeforeCreate()
+  onCreate() {
+    this.id = generateEntityId(this.id, "shoptty")
```
carlos-r-l-rodrigues:
```suggestion
    this.id = generateEntityId(this.id, "shoptru")
```

carlos-r-l-rodrigues:
It is missing things related to Teturns, but I just assume this will be added later

adrien2p:
> It is missing things related to Teturns, but I just assume this will be added later

Yes my idea is to handle the return once everything is working as it will be similar to the rest and i d like to wait for the classic path to be settled completely

packages/fulfillment/src/models/geo-zone.ts
srindom: Yes think that would be super cool to have over time - don't think it's worth including in the MVP though as most providers don't base pricing off of coordinates anyway. 

But would definitely be nice for things like last mile delivery and similar.

srindom (approved):
LGTM!

packages/fulfillment/src/models/fullfilment-set.ts
```diff
@@ -25,6 +40,20 @@ export default class FulfillmentSet {
   @Property({ columnType: "text" })
   name: string
 
+  @Property({ columnType: "text" })
+  type: string
```
olivermrbl, line 44: question: Do we want this to be fully open-ended?

packages/fulfillment/src/models/shipping-option-rule.ts
```diff
@@ -0,0 +1,88 @@
...
+  deleted_at: Date | null = null
+
+  @BeforeCreate()
+  onCreate() {
+    this.id = generateEntityId(this.id, "shoptru")
+  }
+
+  @OnInit()
+  onInit() {
+    this.id = generateEntityId(this.id, "shoptru")
```
olivermrbl:
```suggestion
    this.id = generateEntityId(this.id, "sorul")
  }

  @OnInit()
  onInit() {
    this.id = generateEntityId(this.id, "sorul")
```

packages/fulfillment/src/models/shipping-profile.ts
```diff
@@ -0,0 +1,80 @@
...
+  deleted_at: Date | null = null
+
+  @BeforeCreate()
+  onCreate() {
+    this.id = generateEntityId(this.id, "shpro")
+  }
+
+  @OnInit()
+  onInit() {
+    this.id = generateEntityId(this.id, "shpro")
```
olivermrbl:
```suggestion
    this.id = generateEntityId(this.id, "sp")
  }

  @OnInit()
  onInit() {
    this.id = generateEntityId(this.id, "sp")
```

packages/fulfillment/src/models/fullfilment-set.ts
adrien2p: for now it is not decided but we can come up with a set of applicable values

olivermrbl: 👍

olivermrbl:
Let's slap that automerge label on it, shall we?

adrien2p:
> Let's slap that automerge label on it, shall we?

For sure sir 🙏 any feedback on the util, indexes and the generator for now?

olivermrbl:
> any feedback on the util, indexes and the generator for now?

I think it is a lovely addition. No feedback now. Will have to give it a spin my self :)

PR #6330, opened 2024-02-06
feat(medusa): workflow engine api

carlos-r-l-rodrigues:
What:

   Workflow Engine API.
   Endpoints for:
     - List workflow executions
     - Run a workflow
     - Set async steps as success or failure
     - Retrieve the details of a workflow run


srindom:
Can we add these API endpoints to v1 API too :)

packages/workflows-sdk/src/types/service.ts
```diff
@@ -31,7 +31,16 @@ export type IdempotencyKeyParts = {
...
-export interface IWorkflowsModuleService extends IModuleService {
+export interface IWorkflowEngineService extends IModuleService {
+  retrieveWorkflowExecution(
+    idOrObject: {
+      workflow_id: string
+      transaction_id: string
+    },
+    config?: FindConfig<WorkflowExecutionDTO>,
+    sharedContext?: Context
+  )
```
olivermrbl: todo: Missing return type

packages/workflows-sdk/src/types/service.ts
```diff
@@ -31,7 +31,16 @@ export type IdempotencyKeyParts = {
   action: "invoke" | "compensate"
 }
 
-export interface IWorkflowsModuleService extends IModuleService {
+export interface IWorkflowEngineService extends IModuleService {
+  retrieveWorkflowExecution(
+    idOrObject: {
```
olivermrbl: question: It's called `idOrObject` but typed as an object only. Should we create a dedicated type?

packages/workflow-engine-redis/src/services/workflows-module.ts
```diff
@@ -49,6 +49,42 @@ export class WorkflowsModuleService
     return joinerConfig
   }
 
+  @InjectManager("baseRepository_")
+  async retrieveWorkflowExecution(
+    idOrObject: {
```
olivermrbl: Goes with my other comment. Doens't seem like we are treating the case of `id`; if this is intentional, let's change the name to reflect that

packages/workflow-engine-inmemory/src/services/workflows-module.ts
```diff
@@ -49,6 +49,42 @@ export class WorkflowsModuleService
     return joinerConfig
   }
 
+  @InjectManager("baseRepository_")
+  async retrieveWorkflowExecution(
+    idOrObject: {
```
olivermrbl: Same as other comment

packages/medusa/src/api/routes/admin/workflows/set-step-success.ts
```diff
@@ -0,0 +1,39 @@
...
+    stepResponse,
+    options: {
+      container: req.scope,
+      context: {
+        requestId: req.requestId,
+      },
+    },
+  })
+
+  return res.status(200).json({ success: true })
```
olivermrbl: thought(non-blocking): Would it make sense to return the workflow execution in this route + `setStepFailure`?

packages/medusa/src/api/routes/admin/workflows/index.ts
```diff
@@ -0,0 +1,74 @@
...
+}
+
+const listTransformQueryConfig = {
+  ...retrieveTransformQueryConfig,
+  defaultFields: defaultAdminWorkflowExecutionsFields,
+  isList: true,
+}
+
+export default (app) => {
+  app.use("/workflows", route)
```
olivermrbl: thought(non-blocking): Should we call this route `/workflow-executions` instead, knowing that we will have both a Workflows and Executions overview in the Admin in the future?

So the routes become:

```
GET /admin/workflow-executions

GET /admin/workflow-executions/:id
{ transaction_id: "..." }

POST /admin/workflow-executions/:id/run

POST /admin/workflow-executions/:id/steps/:step_id/success
{ transaction_id: "..." }

POST /admin/workflow-executions/:id/steps/:step_id/failure
{ transaction_id: "..." }

```

olivermrbl: @carlos-r-l-rodrigues, what do you think of this change? 

I know we don't expose all existing workflows + definitions yet, but I think we will some day.

carlos-r-l-rodrigues: That makes sense. And executions can also be fetched by their id. Hence the `idOrObject`.
I am fixing that.

olivermrbl: Cool, sounds good 👍

packages/medusa/src/api/routes/admin/workflows/set-step-success.ts
carlos-r-l-rodrigues: The current execution from where it will resume the `workflow + transaction`?
we can do that.

packages/medusa/src/api-v2/admin/workflows-executions/[workflow_id]/[transaction_id]/[step_id]/success/route.ts
```diff
@@ -0,0 +1,42 @@
...
+  MedusaResponse,
+} from "../../../../../../../types/routing"
+import { AdminPostWorkflowsAsyncResponseReq } from "../../../../validators"
+
+export const POST = async (req: MedusaRequest, res: MedusaResponse) => {
+  const workflowEngineService: IWorkflowEngineService = req.scope.resolve(
+    ModuleRegistrationName.WORKFLOW_ENGINE
+  )
+
+  const { workflow_id, transaction_id, step_id } = req.params
```
Comment (olivermrbl) [inline packages/medusa/src/api-v2/admin/workflows-executions/[workflow_id]/[transaction_id]/[step_id]/success/route.ts]: thought(non-blocking): I am not super fond of creating these lengthy URLs. It makes the endpoint hard to reason about (and less importantly, not super pretty).

An alternative approach would be to include some of the params as an object:
```
POST /admin/workflow-executions/:id/steps/success
{ "transaction_id": "tx_1234", "step_id": "step_1234" }

POST /admin/workflow-executions/:id/steps/failure
{ "transaction_id": "tx_1234", "step_id": "step_1234" }
```

And the GET endpoints would be:
```
GET /admin/workflow-executions/:id

GET /admin/workflow-executions
{ "transaction_id": "tx_1234", "step_id": "step_1234" }
```

What do you think of this?

packages/medusa/src/api-v2/admin/workflows-executions/middlewares.ts
```diff
@@ -0,0 +1,57 @@
...
+import { MiddlewareRoute } from "../../../loaders/helpers/routing/types"
+import * as QueryConfig from "./query-config"
+import {
+  AdminGetWorkflowExecutionDetailsParams,
+  AdminGetWorkflowExecutionsParams,
+  AdminPostWorkflowsAsyncResponseReq,
+  AdminPostWorkflowsRunReq,
+} from "./validators"
+
+export const adminWorkflowsExecutionsMiddlewares: MiddlewareRoute[] = [
```
olivermrbl, line 11: See comment on URLs

packages/medusa/src/api-v2/admin/workflows-executions/route.ts
```diff
@@ -0,0 +1,43 @@
...
+  )
+
+  const listConfig = req.listConfig
+
+  const { transaction_id, workflow_id } = req.filterableFields
+
+  const transactionIds = Array.isArray(transaction_id)
+    ? transaction_id
+    : [transaction_id]
+  const workflowIds = Array.isArray(workflow_id) ? workflow_id : [workflow_id]
```
olivermrbl: thought(non-blocking): Isn't this handled by the module service?

packages/medusa/src/api/routes/admin/workflows-executions/get-execution.ts
```diff
@@ -0,0 +1,26 @@
...
+  const workflowEngineService: IWorkflowEngineService = req.scope.resolve(
+    ModuleRegistrationName.WORKFLOW_ENGINE
+  )
+
+  const { id, workflow_id, transaction_id } = req.params
+
+  const execution = await workflowEngineService.retrieveWorkflowExecution(
+    id ?? {
+      workflow_id,
+      transaction_id,
```
olivermrbl, line 15: See comment on URLs. We could use a standard filter approach for this query

packages/medusa/src/api-v2/admin/workflows-executions/[workflow_id]/[transaction_id]/[step_id]/success/route.ts
Comment (carlos-r-l-rodrigues) [reply packages/medusa/src/api-v2/admin/workflows-executions/[workflow_id]/[transaction_id]/[step_id]/success/route.ts]: for POST endpoints that is ok, we can add to the body, but for GET endpoints it won't change much, in the params is probably better than in the req.query.
But I don't have any strong opinion about it.

Comment (olivermrbl) [reply packages/medusa/src/api-v2/admin/workflows-executions/[workflow_id]/[transaction_id]/[step_id]/success/route.ts]: Cool, lets keep the GET endpoints as is 👍

Comment (adrien2p) [reply packages/medusa/src/api-v2/admin/workflows-executions/[workflow_id]/[transaction_id]/[step_id]/success/route.ts]: with the params there is no typo issue that can occur whereas with query param the user need to build it and be careful of any typo, thats one of the advantage i would say

packages/medusa/src/api-v2/admin/workflows-executions/route.ts
carlos-r-l-rodrigues: Good call, I'll bring it to the service layer.

packages/medusa/src/api/routes/admin/workflows-executions/get-execution.ts
carlos-r-l-rodrigues: not sure I get it. What is the standard filter for the retrieve?
here the primary key is workflow_id + transaction_id, but we also have an id.

olivermrbl: My thinking was that we could just use a regular list endpoint when you want to query using the two IDs:
```
GET /admin/workflow-executions?workflow_id=,transaction_id=,
```
Instead of having this multi-purpose endpoint.

But I am fine to keep it as is 👍

carlos-r-l-rodrigues: for a retrieve where just one single entity is expected, I believe all our endpoints are using just the url, no querystring.
But I'm not aware of other endpoint that could be `/something/_id` or `/something/pk_1/pk_2`

packages/medusa/src/api/routes/admin/workflows/set-step-success.ts
carlos-r-l-rodrigues: @olivermrbl can you confirm it?

olivermrbl (approved):
LGTM!

packages/medusa/src/api/routes/admin/workflows/set-step-success.ts
olivermrbl: I am fine to keep it as is 👍

carlos-r-l-rodrigues: alright. we can revisit that later anyway.

PR #6332, opened 2024-02-06
feat(region): Region create, delete, update admin endpoints

olivermrbl:
**What**
Add `POST /admin/regions`
Add `POST /admin/regions/:id`
Add `DELETE /admin/regions/:id`

All are added for v2 using API Routes and workflows.

In follow-up PRs, I will add support for passing countries in update and create.

Update: First follow-up PR is #6372


packages/medusa/src/api-v2/admin/regions/validators.ts
```diff
@@ -69,3 +69,21 @@ export class AdminGetRegionsParams extends extendedFindParamsMixin({
   @Type(() => AdminGetRegionsParams)
   $or?: AdminGetRegionsParams[]
 }
+
+export class AdminPostRegionsReq {
```
olivermrbl, line 73: comment: Countries will be added in follow up PR

packages/core-flows/src/region/steps/update-regions.ts
```diff
@@ -0,0 +1,59 @@
...
+    }
+
+    const service = container.resolve<IRegionModuleService>(
+      ModuleRegistrationName.REGION
+    )
+
+    await promiseAll(
+      prevData.map(
+        async (region) =>
+          await service.update(region.id, {
```
adrien2p: q: don't we have the ability to provide all the data such that we pass an array of [{ id, name, currency, meta }]?

packages/medusa/src/api-v2/admin/regions/[id]/route.ts
```diff
@@ -17,3 +22,41 @@ export const GET = async (req: MedusaRequest, res: MedusaResponse) => {
 
   res.status(200).json({ region })
 }
+
+export const POST = async (req: MedusaRequest, res: MedusaResponse) => {
+  const updateRegions = updateRegionsWorkflow(req.scope)
```
Comment (adrien2p) [inline packages/medusa/src/api-v2/admin/regions/[id]/route.ts]: suggestion: should we pass the container as part of the run to remove this extra step?

packages/medusa/src/api-v2/admin/regions/[id]/route.ts
```diff
@@ -17,3 +22,41 @@ export const GET = async (req: MedusaRequest, res: MedusaResponse) => {
...
+  if (Array.isArray(errors) && errors[0]) {
+    throw errors[0].error
+  }
+
+  res.status(200).json({ region: result[0] })
+}
+
+export const DELETE = async (req: MedusaRequest, res: MedusaResponse) => {
+  const id = req.params.id
+  const deleteRegions = deleteRegionsWorkflow(req.scope)
```
Comment (adrien2p) [inline packages/medusa/src/api-v2/admin/regions/[id]/route.ts]: same here

packages/medusa/src/api-v2/admin/regions/route.ts
```diff
@@ -18,3 +20,23 @@ export const GET = async (req: MedusaRequest, res: MedusaResponse) => {
...
+
+export const POST = async (req: MedusaRequest, res: MedusaResponse) => {
+  const workflow = createRegionsWorkflow(req.scope)
+  const input = [
+    {
+      ...(req.validatedBody as CreateRegionDTO),
+    },
+  ]
+
+  const { result, errors } = await workflow.run({
```
adrien2p: same here

packages/region/src/services/region-module.ts
```diff
@@ -188,26 +190,36 @@ export default class RegionModuleService<
...
   async update(
-    data: UpdateRegionDTO,
+    regionId: string,
+    data: UpdatableRegionFields,
     sharedContext?: Context
   ): Promise<RegionDTO>
   @InjectTransactionManager("baseRepository_")
   async update(
-    data: UpdateRegionDTO | UpdateRegionDTO[],
+    idOrSelector: string | FilterableRegionProps,
```
adrien2p: suggestion: being able to provide an array of dto would be helpful for most cases like you your steps

srindom: Think we should use upsert for this instead

adrien2p: but if you don't want to create but only update, if it does not exists you might not want to create it under the hood but maybe be aware that it does not exists no?

adrien2p: for example, in the workflow during the rollback we update them, if for some reason, they have been deleted elsewhere, you might re create them accidentally when you didn't want to. So the update in that case would be better than an upsert. Also, it is common in our modules that the update takes an array of data that includes the primary key(s) to perform the bulk update

olivermrbl: Perhaps, we should formalize the signatures, so we know exactly when to use what method and how to implement them. I would also like to see an _update by id_ signature for many of the modules.

@adrien2p, can you let me know if the approach taken in the latest commit is along the lines of what you would like to see?

adrien2p: Yes it looks like it is 👍

packages/types/src/region/mutations.ts
```diff
@@ -1,24 +1,20 @@
...
 export interface UpdateRegionDTO {
   id: string
   currency_code?: string
-  currency?: RegionCurrencyDTO
-  countries: string[]
+  countries?: string[]
+  name?: string
+}
+
+export interface UpdatableRegionFields {
```
sradevski, line 16: Why aren't countries updatable?, similarly, why isn't `metadata` part of `UpdateRegionDTO` and `CreateRegionDTO`?

packages/medusa/src/models/payment.ts
```diff
@@ -10,11 +10,11 @@ import {
 import { DbAwareColumn, resolveDbType } from "../utils/db-aware-column"
 
 import { BaseEntity } from "../interfaces/models/base-entity"
+import { generateEntityId } from "../utils/generate-entity-id"
```
sradevski, line 13: Really not important, but it's better not to change unrelated files 😄

packages/medusa/src/api-v2/admin/regions/validators.ts
```diff
@@ -69,3 +69,21 @@ export class AdminGetRegionsParams extends extendedFindParamsMixin({
...
+  @IsString()
+  name: string
+
+  @IsString()
+  currency_code: string
+}
+
+export class AdminPostRegionsRegionReq {
```
sradevski, line 81: The naming here is a bit strange, looking from the code this is for updates only? Any reason we can't use simple names, like `UpdateRegionReq`?

packages/medusa/src/api-v2/admin/regions/validators.ts
```diff
@@ -69,3 +69,21 @@ export class AdminGetRegionsParams extends extendedFindParamsMixin({
   @Type(() => AdminGetRegionsParams)
   $or?: AdminGetRegionsParams[]
 }
+
+export class AdminPostRegionsReq {
+  @IsString()
```
sradevski, line 74: Just a clarification, the existing tickets are only for updating a region if I understood correctly, and this object is for creating a new Region. Should I also take care of setting the countries during creation of the region as well?

packages/medusa/src/api-v2/admin/regions/route.ts
```diff
@@ -1,20 +1,47 @@
...
 export const GET = async (req: MedusaRequest, res: MedusaResponse) => {
   const remoteQuery = req.scope.resolve("remoteQuery")
 
-  const variables = { filters: req.filterableFields }
-
   const queryObject = remoteQueryObjectFromString({
     entryPoint: "region",
-    variables,
-    fields: defaultAdminRegionFields,
+    variables: {
```
sradevski, line 12: Not necessarily something we need to change in this PR, but wouldn't it make sense for `remoteQueryObjectFromString` to take a `MedusaRequest`, and handle this variable setting internally? Since MedusaRequest is standardized, I don't see any harm in doing that, and it will eliminate this sort of boilerplate.

Eg. the signature will then be `remoteQueryObjectFromString(req, "region")`

packages/medusa/src/api-v2/admin/regions/route.ts
```diff
@@ -1,20 +1,47 @@
...
+    count: metadata.count,
+    offset: metadata.skip,
+    limit: metadata.take,
+  })
+}
+
+export const POST = async (req: MedusaRequest, res: MedusaResponse) => {
+  const input = [
+    {
+      ...(req.validatedBody as CreateRegionDTO),
```
sradevski, line 34: Instead of doing this, would it make sense to make `MedusaRequest` a generic, and then we specify it as `req: MedusaRequest<CreateRegionDTO>`?

packages/medusa/src/api-v2/admin/regions/route.ts
```diff
@@ -1,20 +1,47 @@
...
+    },
+  ]
+
+  const { result, errors } = await createRegionsWorkflow(req.scope).run({
+    input: { regionsData: input },
+    throwOnError: false,
```
sradevski, line 40: I think having too many configuration options makes the code a bit more verbose, and it's better to just do it either through exceptions, or through error object (I prefer the second one, but either is fine). WDYT?

packages/medusa/src/api-v2/admin/regions/[id]/route.ts
```diff
@@ -17,3 +22,38 @@ export const GET = async (req: MedusaRequest, res: MedusaResponse) => {
...
+export const POST = async (req: MedusaRequest, res: MedusaResponse) => {
+  const { result, errors } = await updateRegionsWorkflow(req.scope).run({
+    input: {
+      selector: { id: req.params.id },
+      update: req.validatedBody as UpdatableRegionFields,
+    },
+    throwOnError: false,
+  })
+
+  if (Array.isArray(errors) && errors[0]) {
```
Comment (sradevski) [inline packages/medusa/src/api-v2/admin/regions/[id]/route.ts:35]: Is there a middleware that catches these errors and handles them gracefully? Wouldn't doing something like `res.status(500)` be better if not?

packages/core-flows/src/region/workflows/update-regions.ts
```diff
@@ -0,0 +1,22 @@
...
+
+type UpdateRegionsStepInput = {
+  selector: FilterableRegionProps
+  update: UpdatableRegionFields
+}
+
+type WorkflowInput = UpdateRegionsStepInput
+
+export const updateRegionsWorkflowId = "update-regions"
+export const updateRegionsWorkflow = createWorkflow(
```
sradevski, line 17: Maybe we can have a simpler approach to running simple CRUD workflows, like some default 
`updateWorkflow("region", selector, update)` that does what you've written here under the hood? This is just an optimization of course, just something to keep in mind as we'll have quite a few simple CRUD operations

packages/core-flows/src/region/steps/update-regions.ts
```diff
@@ -0,0 +1,54 @@
...
+  async (data: UpdateRegionsStepInput, { container }) => {
+    const service = container.resolve<IRegionModuleService>(
+      ModuleRegistrationName.REGION
+    )
+
+    const { selects, relations } = getSelectsAndRelationsFromObjectArray([
+      data.update,
+    ])
+
+    const prevData = await service.list(data.selector, {
```
sradevski, line 27: Are this and the operation below atomic? if not, we can easily end up with a race condition where you list the previous data, another workflow updates the same region, then this update fails, and you revert to the oldest data that is no longer correct.

I have quite a few questions around race conditions and resilience of workflows that maybe I'll talk to Carlos about :)

integration-tests/plugins/__tests__/regions/admin/create-region.spec.ts
```diff
@@ -0,0 +1,65 @@
...
+  beforeEach(async () => {
+    await adminSeeder(dbConnection)
+  })
+
+  afterEach(async () => {
+    const db = useDb()
+    await db.teardown()
+  })
+
+  it("should create a region", async () => {
```
sradevski: Adding few more tests, like an empty body, using an invalid `currency_code` or an empty name (if that is not allowed) would be nice to have.

integration-tests/plugins/__tests__/regions/admin/update-region.spec.ts
```diff
@@ -0,0 +1,70 @@
...
+import adminSeeder from "../../../../helpers/admin-seeder"
+
+jest.setTimeout(50000)
+
+const env = { MEDUSA_FF_MEDUSA_V2: true }
+const adminHeaders = {
+  headers: { "x-medusa-access-token": "test_token" },
+}
+
+describe("POST /admin/regions/:id", () => {
```
sradevski: I think having a single `regions/admin/regions.spec.ts` file will reduce all the boilerplate around setting up the before/after hooks, and it also makes sense for CRUD operations to be next to each other. Since these are integration tests, you can have a more detailed flow in a single test where you:
1. Create a region
2. You update it
3. You delete it

This tests the entire lifecycle of the region without duplicating a lot of code.

packages/medusa/src/api-v2/admin/regions/route.ts
adrien2p: The idea is that the util can be used by our users as well, and shouldn't rely on express necessarely, that is why we are enforcing the API of the util to accept a specific object instead of knowing about the underlying http server. this util can be used outside of the Http layer

packages/medusa/src/api-v2/admin/regions/route.ts
adrien2p: there is another PR where we talked about doing that

packages/medusa/src/api-v2/admin/regions/route.ts
adrien2p: why don't we let the workflow throw the error, that a good question I don't remember

packages/medusa/src/api-v2/admin/regions/[id]/route.ts
Comment (adrien2p) [reply packages/medusa/src/api-v2/admin/regions/[id]/route.ts:35]: we have a middleware that handle that, and we generally throw a medusa error type and depending on the type it respondwith the appropriate status

packages/core-flows/src/region/steps/update-regions.ts
adrien2p: we have an on going discussion around loack, you can discuss it with carlos too

packages/medusa/src/api-v2/admin/regions/route.ts
sradevski: @adrien2p from what I understand this is part of the medusa core package, so customers won't really do changes here? Or is there another use-case for remoteQueryObjectFromString?

packages/medusa/src/api-v2/admin/regions/validators.ts
olivermrbl: This goes far back, and I don't think we are actively using these names for much right now.

The idea was that payloads and query params should be derived from the request's path to ensure uniqueness and readability. This could prove useful in case we want to autogenerate stuff that uses these types. And it would allow the reader to always know exactly where a specific type is used, regardless of its location.

So for example: `POST /admin/products`

The types for the above endpoint would be:
- Payload: `AdminPostProductsReq`
- Query: `AdminPostProductsParams`
- General response: `AdminProductsRes`
- Delete response: `AdminProductsDeleteRes`

And you would know where this is used, even if placed in a separate folder or package.

packages/medusa/src/api-v2/admin/regions/validators.ts
olivermrbl: > Should I also take care of setting the countries during creation of the region as well?

That would be great!

packages/types/src/region/mutations.ts
olivermrbl: `metadata` should most definitly be updatable. I will add that in now. 

Updating countries in `service.update` is not supported yet. Should be tackled in a separate ticket.

packages/core-flows/src/region/workflows/update-regions.ts
olivermrbl: Love this idea! I imagine we'll find more of these cases when we start tackling workflows more aggresively

packages/medusa/src/api-v2/admin/regions/route.ts
adrien2p: this utils is part of the utils package, they could use it in a lambda, or in a next function etc for example. modules and remote query can be used in various places

integration-tests/plugins/__tests__/regions/admin/regions.spec.ts
```diff
@@ -0,0 +1,224 @@
+import { ModuleRegistrationName } from "@medusajs/modules-sdk"
```
olivermrbl, line 1: @sradevski, consolidated tests here.

sradevski: @olivermrbl looks much better, thanks!

sradevski (approved):

olivermrbl:
@medusajs/engineering, this one is longing for a review 😄

integration-tests/plugins/__tests__/regions/admin/regions.spec.ts
```diff
@@ -0,0 +1,224 @@
...
+      id: updated.data.region.id,
+      object: "region",
+      deleted: true,
+    })
+
+    const deletedRegion = await service.retrieve(updated.data.region.id, {
+      withDeleted: true,
+    })
+
+    // @ts-ignore
```
adrien2p, line 101: q: why do we need that ignore here?

integration-tests/plugins/__tests__/regions/admin/regions.spec.ts
```diff
@@ -0,0 +1,224 @@
...
+  it("should throw on missing required properties in create", async () => {
+    const api = useApi() as any
+    try {
+      await api.post(`/admin/regions`, {}, adminHeaders)
+    } catch (error) {
+      expect(error.response.status).toEqual(400)
+      expect(error.response.data.message).toEqual(
+        "name must be a string, currency_code must be a string"
+      )
+    }
```
adrien2p: I suggest the following here, in general it is better to not have any expectation in the catch because if someone breaks the API and it doesnt throw anymore, then this test will never rich the expectation and will succeed
```suggestion
     const api = useApi() as any
      const err = await api.post(`/admin/regions`, {}, adminHeaders).catch(e => e)
      expect(error.response.status).toEqual(400)
      expect(error.response.data.message).toEqual(
        "name must be a string, currency_code must be a string"
      )
```

integration-tests/plugins/__tests__/regions/admin/regions.spec.ts
```diff
@@ -0,0 +1,224 @@
...
+  })
+
+  it("should throw on unknown currency in create", async () => {
+    const api = useApi() as any
```
adrien2p, line 118: todo: same commend as before

integration-tests/plugins/__tests__/regions/admin/regions.spec.ts
```diff
@@ -0,0 +1,224 @@
...
+        `/admin/regions`,
+        {
+          foo: "bar",
+          currency_code: "usd",
+          name: "Test Region",
+        },
+        adminHeaders
+      )
+    } catch (error) {
+      expect(error.response.status).toEqual(400)
```
adrien2p: todo: same for all try catch block

packages/medusa/src/api-v2/admin/regions/query-config.ts
```diff
@@ -18,6 +20,9 @@ export const defaultAdminRegionFields = [
 ]
 
 export const retrieveTransformQueryConfig = {
+  defaultFields: defaultAdminRegionFields,
+  defaultRelations: defaultAdminRegionRelations,
+  allowedRelations: allowedAdminRegionRelations,
```
adrien2p, line 25: suggestion(nit): we could just have `defaultAdminRegionFields` which already include the relations. In the route (for now but in the middleware would be better) we can use the util to get the select and relations from the fields. wdyt? Just in order to start the cleanup and not have to maintain two array etc and facilitate the migration later

adrien2p (approved):
LGTM, few todo's left

PR #6342, opened 2024-02-08
feat(region): Add store region get + list endpoints

olivermrbl:
**What**
Add `GET /store/regions`
Add `GET /store/regions/:id`

Both added as API Routes using remoteQuery for v2.

~~Depends on #6322. Should not be reviewed before that has been merged.~~


fPolic (approved):
LGTM

PR #6353, opened 2024-02-08
docs-util: fix release scripts

shahednasser:
- Fix GitHub action to run on push and check if the commit message is "chore: Release". Only then are TSDocs generated and a PR is opened.
- Add an option to pass to the `run:release` method of the docblock generator a release tag. This is helpful in cases when the GitHub action fails for any reason.
- Add scripts that checks the message of a commit.


olivermrbl (approved):
LGTM!

PR #6372, opened 2024-02-11
feat(region): Create region with countries

olivermrbl:
**What**
Create region with countries

Validation is added for the following two cases:
- Passing a country that doesn't exist e.g. creating a region with "Wonderland" as country
- Passing a country that is already assigned to a different region

### DIFF
--- a/packages/region/src/services/region-module.ts
+++ b/packages/region/src/services/region-module.ts
@@ -125,13 +125,16 @@ export default class RegionModuleService<
       sharedContext
     )
 
-    const countriesInDbMap = new Map(countriesInDb.map((c) => [c.iso_2, c]))
+    const countriesInDbMap = new Map<string, Country>(
+      countriesInDb.map((c) => [c.iso_2, c])
+    )
 
     const currencyMap = new Map(
       currencies.map((c) => [c.code.toLowerCase(), c])
     )
 
     const toCreate: InternalCreateRegionDTO[] = []
+    const seenCountries = new Set<string>()
     for (const region of data) {
       const reg = { ...region } as InternalCreateRegionDTO
       const countriesToAdd = region.countries || []
@@ -151,12 +154,14 @@ export default class RegionModuleService<
         const regionCountries = countriesToAdd.map((code) => {
           const country = countriesInDbMap.get(code.toLowerCase())
 
-          if (country?.region_id) {
+          if (country?.region_id || seenCountries.has(code.toLowerCase())) {
             throw new MedusaError(
               MedusaError.Types.INVALID_DATA,
               `Country with code ${code} is already assigned to a region`
             )
           }
+
+          seenCountries.add(code.toLowerCase())
 
           return country
         }) as unknown as TCountry[]
--- a/packages/region/integration-tests/__tests__/region-module.spec.ts
+++ b/packages/region/integration-tests/__tests__/region-module.spec.ts
@@ -148,6 +148,25 @@ describe("Region Module Service", () => {
     )
   })
 
+  it("should throw when country is being assigned to multiple regions", async () => {
+    await expect(
+      service.create([
+        {
+          name: "United States",
+          currency_code: "USD",
+          countries: ["us"],
+        },
+        {
+          name: "North America",
+          currency_code: "USD",
+          countries: ["us"],
+        },
+      ])
+    ).rejects.toThrowError(
+      "Country with code us is already assigned to a region"
+    )
+  })
+
   it("should fail when currency does not exist", async () => {
     await expect(
       service.create({


packages/region/src/services/region-module.ts
```diff
@@ -115,17 +151,41 @@ export default class RegionModuleService<
       sharedContext
     )
 
-    let currencyMap = new Map(currencies.map((c) => [c.code.toLowerCase(), c]))
-    for (const reg of data) {
-      const lowerCasedCurrency = reg.currency_code.toLowerCase()
+    const regionCountriesMap = new Map<string, string[]>(
+      data.map((d) => [d.name, d?.countries || []])
+    )
```
pKorsholm: todo: having this map doesn't increase performance as you are using it to loop through data and access `.countries` of each element when looping through making it O(n)
```suggestion
```

packages/region/src/services/region-module.ts
```diff
@@ -115,17 +151,41 @@ export default class RegionModuleService<
...
+      sharedContext
+    )
+
+    const countriesInDbMap = new Map(countriesInDb.map((c) => [c.iso_2, c]))
+    const currencyMap = new Map(
+      currencies.map((c) => [c.code.toLowerCase(), c])
+    )
+
+    for (const region of data) {
+      const countriesToAdd = regionCountriesMap.get(region.name)
```
pKorsholm:
```suggestion
```

packages/region/src/services/region-module.ts
```diff
@@ -115,17 +151,41 @@ export default class RegionModuleService<
...
+
+      if (countriesToAdd) {
```
pKorsholm, line 142:
```suggestion
      if (region.countries?.length) {
```

packages/region/src/services/region-module.ts
```diff
@@ -104,6 +104,42 @@ export default class RegionModuleService<
...
+      throw new MedusaError(
+        MedusaError.Types.INVALID_DATA,
+        `Countries with codes ${countriesToAdd.join(", ")} does not exist`
+      )
+    }
+
+    const regionCountries = countriesToAdd.map((code) => {
+      const country = countriesInDb.get(code.toLowerCase())
+
+      if (country?.region_id) {
```
pKorsholm: This should always exist at this point given the validation above right?
```suggestion
      if (country!.region_id) {
```

packages/region/src/services/region-module.ts
```diff
@@ -115,17 +151,41 @@ export default class RegionModuleService<
...
+    )
+
+    for (const region of data) {
+      const countriesToAdd = regionCountriesMap.get(region.name)
+
+      if (countriesToAdd) {
+        this.validateAndSetCountriesForRegion(
+          region,
+          countriesInDbMap,
+          countriesToAdd
```
pKorsholm: todo: remove this parameret in favor of `region.countries`

packages/region/src/services/region-module.ts
```diff
@@ -115,17 +151,41 @@ export default class RegionModuleService<
...
+      data.map((d) => [d.name, d?.countries || []])
+    )
+
+    const countriesInDb = await this.countryService_.list(
+      { iso_2: data.map((d) => d.countries).flat() },
+      { select: ["iso_2", "region_id"] },
+      sharedContext
+    )
+
+    const countriesInDbMap = new Map(countriesInDb.map((c) => [c.iso_2, c]))
```
pKorsholm: I understand having this outside the validation method to avoid the N database queries, but it feels like we're bleeding behavior from the validation method into the creation method (same for currencies).

I think it would make sense to increase the scope validation method instead and pass along the input data as a whole instead  to keep this method cleaner. You could also consider creating an overload for the method to take either an array or the arguments here. 

suggestion: 
```typescript
private async validateCurrecyAndCountries(
    countriesToAdd: CreateRegionDTO[],
    sharedContext?: Context
  ): Promise<void>
  private async validateCurrecyAndCountries(
    region: CreateRegionDTO,
    sharedContext?: Context
  ): Promise<void>
  private async validateCurrecyAndCountries(
    regionOrData: CreateRegionDTO | CreateRegionDTO[],
    sharedContext?: Context
  ): Promise<void> {
    const data = Array.isArray(regionOrData) ? regionOrData : [regionOrData]

    const countriesInDb: CountryDTO[] = await this.countryService_.list(
      { iso_2: data.map((d) => d.countries).flat() },
      { select: ["iso_2", "region_id"] },
      sharedContext
    )

    const countriesInDbMap = new Map(countriesInDb.map((c) => [c.iso_2, c]))

    const currencies = await this.currencyService_.list(
      { code: data.map((d) => d.currency_code.toLowerCase()) },
      {},
      sharedContext
    )

    const currencyMap = new Map(
      currencies.map((c) => [c.code.toLowerCase(), c])
    )

    for (const region of data) {
      const lowerCasedCurrency = region.currency_code.toLowerCase()
      if (!currencyMap.has(lowerCasedCurrency)) {
        throw new MedusaError(
          `Currency with code: ${region.currency_code} was not found`
        )
      }
      region.currency_code = lowerCasedCurrency

      if (!region.countries?.length) {
        continue
      }

      const notInDb = region.countries.filter(
        (c) => !countriesInDbMap.has(c.toLowerCase())
      )

      if (notInDb.length) {
        throw new MedusaError(
          `Countries with codes ${notInDb.join(", ")} does not exist`
        )
      }

      const regionCountries = region.countries.map((code) => {
        const country = countriesInDbMap.get(code.toLowerCase())! as any
        // TODO: cleanup types
        if (country.region_id as string) {
          throw new MedusaError(
            `Country with code ${code} is already assigned to a region`
          )
        }

        return country
      }) as unknown as TCountry[]

      // TODO: Fix TS error
      // @ts-ignore
      region.countries = regionCountries
    }
  }

  async create_(
    data: CreateRegionDTO[],
  ): Promise<Region[]> {
    await this.validateCurrecyAndCountries(data)

    const result = await this.regionService_.create(data, sharedContext)

    return result
  }
```

packages/region/src/models/country.ts
```diff
@@ -30,13 +30,15 @@ export default class Country {
   @Property({ columnType: "text" })
   display_name: string
 
+  @Property({ columnType: "text", nullable: true })
+  region_id?: string
```
pKorsholm:
```suggestion
  region_id?: string | null
```

packages/region/src/services/region-module.ts
```diff
@@ -104,6 +104,42 @@ export default class RegionModuleService<
...
+        throw new MedusaError(
+          MedusaError.Types.INVALID_DATA,
+          `Country with code ${code} is already assigned to a region`
+        )
+      }
+
+      return country
+    }) as unknown as TCountry[]
+
+    if (regionCountries?.length) {
```
pKorsholm: This check shouldn't be necessary given the `notInDb` check above. If you want to validate that countries has a length I suggest adding an early return before the validation

packages/region/src/services/region-module.ts
olivermrbl: Which validation point?

First is checking if the country exists e.g. if you try to add Wonderland it will throw. 

Second is checking if you are assigning a country, that's already assigned to a region e.g. if you create a region "US" with country "us", but region "North America" already exists (and includes "us").

packages/region/src/services/region-module.ts
olivermrbl: Yeah, I was a bit on the fence about the separate validation method. It didn't feel right. And you confirmed this.

However, with the above suggestion, we are effectively replacing the create method, so I'd rather just unpack the validation directly in the create method for now, and revisit if we need to.

packages/region/src/models/country.ts
```diff
@@ -3,15 +3,20 @@ import {
...
 } from "@mikro-orm/core"
 
 import { generateEntityId } from "@medusajs/utils"
 import Region from "./region"
 
+type CountryOptionalProps = "region" | "region_id"
+
 @Entity({ tableName: "region_country" })
 export default class Country {
+  [OptionalProps]?: CountryOptionalProps
```
fPolic: q: do we need `region/region_id` here since they are marked with optional with `?`

fPolic (approved):
LGTM

packages/region/src/services/region-module.ts
olivermrbl: @pKorsholm, ping :D

packages/region/src/models/country.ts
```diff
@@ -30,13 +30,15 @@ export default class Country {
   @Property({ columnType: "text" })
   display_name: string
 
+  @Property({ columnType: "text", nullable: true })
+  region_id?: string | null = null
```
adrien2p, line 34: The ? Is not required to be here, since there is a default initializer it is by default optional

olivermrbl: You're right. More generally, this should not be a blocking comment (I know you haven't done that, but just thought I'd point it out)

adrien2p: Yes most definitely, hence why i just slapped a comment only

packages/region/src/services/region-module.ts
pKorsholm: I meant you know the country is in the database at this point, you've validated that the country exists using the `notInDb` check, i.e. the map will never return undefined. It's a nit comment at best, feel free to disregard 😄

packages/region/src/services/region-module.ts
```diff
@@ -115,20 +119,65 @@ export default class RegionModuleService<
...
+
+          if (country?.region_id) {
+            throw new MedusaError(
+              MedusaError.Types.INVALID_DATA,
+              `Country with code ${code} is already assigned to a region`
+            )
+          }
+
+          return country
+        }) as unknown as TCountry[]
```
pKorsholm: suggestion: split the mapping of countries to db entities from the validation of region_id for readability. 

We're not doing any db queries and everything is constant operation so the operation would still be linear in cost (though 2N instead of just N): 

```
 const alreadyHasRegion = countriesToAdd.filter(
          (c) => countriesInDbMap.get(c.toLowerCase())?.region_id
        )

        if (alreadyHasRegion.length) {
          throw new MedusaError(
            `Country with code ${alreadyHasRegion[0]} is already assigned to a region`
          )
        }

        reg.countries = countriesToAdd.map(
          (code) => countriesInDbMap.get(code.toLowerCase())!
        )
```

pKorsholm (commented):
we should check for duplicate countries being assigned when creating multiple regions. This test fails because both regions are created (the last will have the country): 

```
it("should throw when country is being assigned to multiple regions", async () => {
    await expect(
      service.create([
        {
          name: "United States",
          currency_code: "USD",
          countries: ["us"],
        },
        {
          name: "North America",
          currency_code: "USD",
          countries: ["us"],
        },
      ])
    ).rejects.toThrowError(
      "Country with code us is already assigned to a region"
    )
  })
```

PR #6377, opened 2024-02-12
chore(docs): add empty changeset step to actions

shahednasser:
- Creates a script that runs the `yarn changeset --empty` command if there are file changes. This is important to not create PRs of just empty changesets if there are no file changes.
- Run script in all docs-generation PRs
- Fix to condition in `generate-docblock` action


olivermrbl:
> This is important to not create PRs of just empty changesets if there are no file changes.

I am not sure I understand this part. I know what the command does, but what exactly do we need it for here?

shahednasser:
@olivermrbl for the actions that generate docs (specs, references, etc...) the process is usually like this:

1. Run the command that generates the docs
2. Run the create PR step. In this step, if there are git changes, it creates a PR. otherwise, no PR is created as there was nothing new generated.

If we add the `yarn changeset --empty` command before the create PR step as-is, it'll add a changeset regardless of whether there were actual changes (generated docs). So, the create PR step will always create a PR, even if there are no changes, which would result in PRs having just empty changesets.

With the condition step, we only run `yarn changeset --empty` if there are git changes. This maintains the behavior of point 2.

olivermrbl:
Got it, makes sense. Thx for explaining!

olivermrbl (approved):
LGTM

PR #6381, opened 2024-02-12
feat(fulfillment): Module service implementation first iteration

adrien2p:
**What**

> [!TIP]
> The review should particularly focus on the module service implementation and the integration tests to ensure that the behavior conforms to the expectations. In Sus, the adjustments made to the models and DTO's

> [!WARNING]
> The delete methods are kept aside as it will require manually cascading since according to the doc, cascade remove on collection will perform one query per item. In that case I rather prefer deleting them manually in bulk.

- Preliminary migration file
   - Does not include changes to be integrated with existing tables from the core, so probably not to review unless for indexes and modules schema expectations
- Events will be managed separately later on
- Update some indexes and unique indexes
- Start implementing the module service with the following
  - create - To create fulfillment set including service zones and geo zones
  - createServiceZones - To create service zones and geo zones that can be attached to a fulfillment set on update
  - update - To update a fulfillment set and allow to replace/add/remove service zones
  - updateServiceZones - Update service zones with their geo zones to either attach an existing zones or create a new one or replace existing ones
  - create geo zones
  - update geo zones
  - soft delete are based from the abstraction
  - list/retrieve/listAndCount are based from the abstraction
- Few adjustments in regards to the DTO's
- Integration tests for the above module service

FIXES CORE-1716
FIXES CORE-1734


packages/customer/src/migrations/Migration20240124154000.ts
```diff
@@ -23,7 +23,7 @@ export class Migration20240124154000 extends Migration {
       'create index if not exists "IDX_customer_address_customer_id" on "customer_address" ("customer_id");'
     )
     this.addSql(
-      'create unique index "IDX_customer_address_unqiue_customer_billing" on "customer_address" ("customer_id") where "is_default_billing" = true;'
+      'create unique index "IDX_customer_address_unique_customer_billing" on "customer_address" ("customer_id") where "is_default_billing" = true;'
```
adrien2p: @srindom by trully running the migrations as part of the integration tests, I encounter this one 😂 until now, the test-utils wasn't running the migration at all. It might help forcing everybody to test migrations as well

adrien2p: It means that until now, even if the integration tests was green, it could be not working in truth cause of the discrepancy between the definition and the migrations.

Will see if there is other like this

adrien2p: fixed in another pr

packages/fulfillment/src/repositories/fulfillment-set.ts
```diff
@@ -0,0 +1,59 @@
+/*
```
adrien2p, line 1: TODO: not necessary but keeping it for now just in case I need that piece of code in the future, more of history keeping

packages/fulfillment/src/migrations/Migration20240213165503.ts
```diff
@@ -0,0 +1,88 @@
+import { Migration } from '@mikro-orm/migrations';
+
+export class Migration20240213165503 extends Migration {
+
+  async up(): Promise<void> {
+    this.addSql('create table if not exists "fulfillment_address" ("id" text not null, "fulfillment_id" text null, "company" text null, "first_name" text null, "last_name" text null, "address_1" text null, "address_2" text null, "city" text null, "country_code" text null, "province" text null, "postal_code" text null, "phone" text null, "metadata" jsonb null, "created_at" timestamptz not null default now(), "updated_at" timestamptz not null default now(), "deleted_at" timestamptz null, constraint "fulfillment_address_pkey" primary key ("id"));');
```
olivermrbl: thought: One thing I would like us to do at a later point is format these, so they are easier to read. 

I pasted mine in ChatGPT and asked it to format it for me.

packages/fulfillment/src/services/fulfillment-module-service.ts
```diff
@@ -30,27 +39,176 @@ export default class FulfillmentModuleService<
...
+   * extracting the service zones and geo zones from the data and then from that
+   * data extract the ids of the service zones and geo zones that are already
+   * existing in the database. Then it will fetch the existing service zones and
+   * geo zones from the database and return them.
+   *
+   * @param data
+   * @param sharedContext
+   * @protected
+   */
+  protected async prepareCreateData(
```
olivermrbl: praise: nice!

packages/fulfillment/src/services/fulfillment-module-service.ts
```diff
@@ -151,7 +808,130 @@ export default class FulfillmentModuleService<
...
+    if (geoZoneIds.length) {
+      existingGeoZones.push(
+        ...(await this.geoZoneService_.list(
+          {
+            id: geoZoneIds,
+          },
+          {},
+          sharedContext
+        ))
+      )
```
olivermrbl: suggestion: In these case, I'd rather have a separate API call. Think it makes for cleaner code, even thought it increases the footprint

packages/fulfillment/src/services/fulfillment-module-service.ts
```diff
@@ -151,7 +808,130 @@ export default class FulfillmentModuleService<
...
+        const geoZoneTuple: [
+          string,
+          FulfillmentTypes.CreateGeoZoneDTO | { id: string }
+        ][] = geo_zones.map((geoZone) => {
+          if (!("id" in geoZone)) {
+            geoZonesToCreate.push(geoZone)
+          }
+
+          const existingZone =
+            "id" in geoZone ? existingGeoZonesMap.get(geoZone.id)! : geoZone
```
olivermrbl: question: Even though we queried the database for GeoZones, couldn't we be in a situation where the ID didn't exist in the database i.e. the map would not contain the GeoZone? In that case, `existingZone` will be undefined. 

This comment is relevant for all places where we have this pattern of "query the DB and add them to a map for later use"

packages/fulfillment/src/services/fulfillment-module-service.ts
```diff
@@ -30,27 +39,176 @@ export default class FulfillmentModuleService<
...
+  /**
+   * Returns the identifier of a geo zone. The identifier is a string that is
+   * generated based on the type of the geo zone and the properties of the geo
+   * zone. The identifier is used for map building and retrieval.
+   *
+   * @param geoZone
+   * @param preventIdUsage
+   * @protected
+   */
+  protected static getGeoZoneIdentifier(
```
carlos-r-l-rodrigues: suggestion: identifier can be an `[]` and each value has a string to identify the attribute. The method returns `identifier.join("_")`. 
```
if ("country_code" in geoZone && geoZone.country_code) {
  identifier.push("c:" + geoZone.country_code)
}
```
that will avoid clashing in case both properties has the same value. like `city: SP, province_code: SP`

adrien2p: With that example and the current code, the identifier will be spsp, which shouldn't conflict no? But i can do that still 😊

packages/fulfillment/src/services/fulfillment-module-service.ts
adrien2p: 🤣 you are right, i usually dont do that but i went crazy, i will store it separately before pushing

packages/fulfillment/src/services/fulfillment-module-service.ts
adrien2p: You are right, i can have an extra step that query then do a diff to ensure that 💪 ill add that in

packages/fulfillment/src/services/fulfillment-module-service.ts
carlos-r-l-rodrigues: they are optional, aren't they?
if you have city: sp, and other one province_code: sp, they will become the same.

packages/fulfillment/src/models/fullfilment-set.ts
```diff
@@ -51,6 +64,7 @@ export default class FulfillmentSet {
     pivotTable: "fulfillment_set_service_zones",
     joinColumn: "fulfillment_set_id",
     inverseJoinColumn: "service_zone_id",
+    fixedOrder: true,
   })
   service_zones = new Collection<ServiceZone>(this)
```
srindom, line 67: thought: think this should be a OneToMany relationship - each service zone will be be defined in the context of the location that the fulfillment set is for, and it is quite unlikely that two locations will share exactly the same service zones.

packages/fulfillment/src/services/fulfillment-module-service.ts
```diff
@@ -30,27 +39,176 @@ export default class FulfillmentModuleService<
...
+   */
+  protected async prepareCreateData(
+    data: (
+      | FulfillmentTypes.CreateFulfillmentSetDTO
+      | FulfillmentTypes.UpdateFulfillmentSetDTO
+    )[],
+    sharedContext?: Context
+  ): Promise<{
+    existingServiceZones: TServiceZoneEntity[]
+    existingServiceZonesMap: Map<string, TServiceZoneEntity>
```
srindom: thought: we can get rid of the existingServiceZone notion here since there is no ManyToMany between fulfillment sets and service zones - i.e. the service zone will always be created in the DB as part of the fulfillment set creation or added at a later point in time.

packages/fulfillment/src/services/fulfillment-module-service.ts
```diff
@@ -69,7 +227,194 @@ export default class FulfillmentModuleService<
...
+                FulfillmentModuleService.getGeoZoneIdentifier(geoZone)
+
+              return [geoZoneIdentifier, existingGeoZone]
+            })
+          }
+
+          const existingZone =
+            "id" in serviceZone
+              ? existingServiceZonesMap.get(serviceZone.id)!
+              : serviceZone
```
srindom: this can be disregarded.

packages/fulfillment/src/models/service-zone.ts
```diff
@@ -57,6 +70,7 @@ export default class ServiceZone {
     pivotTable: "service_zone_geo_zones",
     joinColumn: "service_zone_id",
     inverseJoinColumn: "geo_zone_id",
+    fixedOrder: true,
   })
   geo_zones = new Collection<GeoZone>(this)
```
srindom, line 85: thought: I am actually also leaning toward making this a OneToMany relationship instead. The thinking is that each ServiceZone may evolve independently of each other. Take this example:

1. I have Warehouse A with Service Zone A that ships to Europe.
2. I have Warehouse B with Service Zone B which also ships to Europe. 
3. I add the Europe GeoZone to Service Zone A and B.
4. I now learn that Warehouse B is geographically placed so that shipments to a particular country in Europe is extremely expensive, so I want to remove that country for Warehouse B. 
5. I go to Warehouse B and remove the country.
6. **But** I have now also removed that country from Warehouse A which was not my intention.

When you think about the flows here – both from module service and eventually from UI perspectives – it is better to have the geo_zones separated.

srindom (commented):
Very nice work - have a few comments on the data model. Might make sense to address as part of this PR, but up to you.

packages/fulfillment/src/models/fullfilment-set.ts
adrien2p: Can't we have the same service zone for different type of set?

packages/fulfillment/src/models/service-zone.ts
adrien2p: I see your point, what i was thinking is if you want to remove a geo from a zone then we detach it from the zone. But i can go with that as well i dont have a strong opinion for that. It will also simplify things 👍

srindom: Let's do it 👍

packages/fulfillment/src/models/fullfilment-set.ts
adrien2p: I think you are right, same as geo, if they have the same set but a different type they can create a separate zone as well

srindom: If it's a different type of set e.g. pickup vs. shipping, you will have different shipping options. For example, you don't have a shipping option with FedEx when you are coming down to pick up the package in a store.

adrien2p: Yes your are right 👍

packages/fulfillment/src/services/fulfillment-module-service.ts
adrien2p: I will add a validator to ensure the data, but the type is included in the identifier.
If one has city: sp, then it is probably of type city since there is no province code, and province code would have to be provided if it was of type province, but in that case city is not expected. But in any case, it would make clearer identifier to go with it so i ll include it

srindom (approved):
LGTM!

packages/fulfillment/src/models/fullfilment-set.ts
```diff
@@ -46,11 +60,9 @@ export default class FulfillmentSet {
...
   metadata: Record<string, unknown> | null = null
 
-  @ManyToMany(() => ServiceZone, "fulfillment_sets", {
-    owner: true,
-    pivotTable: "fulfillment_set_service_zones",
-    joinColumn: "fulfillment_set_id",
-    inverseJoinColumn: "service_zone_id",
+  @OneToMany(() => ServiceZone, "fulfillment_set", {
+    cascade: [Cascade.PERSIST, "soft-remove"] as any,
+    orphanRemoval: true,
```
olivermrbl, line 65: question: I thought we didn't use this because of its "aggressiveness". I am fine to keep it in, but just wanted to point it out in case it was an oversight

packages/fulfillment/src/models/service-zone.ts
```diff
@@ -41,22 +59,28 @@ export default class ServiceZone {
...
+  fulfillment_set: FulfillmentSet
 
-  @ManyToMany(() => GeoZone, "service_zones", {
-    owner: true,
-    pivotTable: "service_zone_geo_zones",
-    joinColumn: "service_zone_id",
-    inverseJoinColumn: "geo_zone_id",
+  @OneToMany(() => GeoZone, "service_zone", {
+    cascade: [Cascade.PERSIST, "soft-remove"] as any,
+    orphanRemoval: true,
```
olivermrbl, line 83: Same here

olivermrbl (approved):

olivermrbl:
@adrien2p feel free to merge 🚢

packages/fulfillment/src/models/fullfilment-set.ts
adrien2p: Here it a requirements. Because since i am bulk deleting the entities, then when mikro orm manage the collection under the hood it will complain that thos configuration is missing. Also, i am in a case where if the fulfillment set is removed completely, then we can have orphan service zones since the foreign key is non nullable. Same goes between service zones and geo zones

PR #6386, opened 2024-02-13
Feat(utils): psql unique index instead of constraint

pKorsholm:
**What**
- always return an index expression, also for unique constraints

**why**
- constraints can't be partial, meaning `UNIQUE ... WHERE` is not possible with constraints
- constraints are indicies under the hood so it doesn't change the behavior of the system when we're not using constraint specific features but just using them for `UNIQUE`

### DIFF
--- a/packages/utils/src/common/__tests__/create-psql-index-helper.ts
+++ b/packages/utils/src/common/__tests__/create-psql-index-helper.ts
@@ -73,9 +73,9 @@ describe("createPsqlIndexStatementHelper", function () {
 
     const indexStatement = createPsqlIndexStatementHelper(options)
     expect(indexStatement).toEqual(
-      `ALTER TABLE IF EXISTS "${options.tableName}" ADD CONSTRAINT "${
-        options.name
-      }" UNIQUE (${options.columns.join(", ")}) WHERE ${options.where}`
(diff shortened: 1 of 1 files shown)


packages/utils/src/common/create-psql-index-helper.ts
```diff
@@ -45,10 +45,7 @@ export function createPsqlIndexStatementHelper({
...
   const typeStr = type ? ` USING ${type}` : ""
   const optionsStr = where ? ` WHERE ${where}` : ""
+  const uniqueStr = unique ? "UNIQUE " : ""
 
-  if (!unique) {
-    return `CREATE INDEX IF NOT EXISTS "${name}" ON "${tableName}"${typeStr} (${columns})${optionsStr}`
-  } else {
-    return `ALTER TABLE IF EXISTS "${tableName}" ADD CONSTRAINT "${name}" UNIQUE (${columns})${optionsStr}`
-  }
+  return `CREATE ${uniqueStr}INDEX IF NOT EXISTS ${name} ON ${tableName}${typeStr} (${columns})${optionsStr}`
```
adrien2p: Todo: re add all double quotes and update tests and then merge 😁 thanks man

adrien2p (approved):
💪🎉

PR #6388, opened 2024-02-13
Feat(medusa, core-flows, utils, types, user): add user endpoints to api-v2

pKorsholm:
**What**
- Add users endpoints for simple CRUD

_note_: this is NOT for handling sign ups, authentication and creation of users as auth-users, a separate pr will follow with handlers dealing with auth

### DIFF
--- a/packages/medusa/src/api-v2/admin/users/validators.ts
+++ b/packages/medusa/src/api-v2/admin/users/validators.ts
@@ -104,10 +104,6 @@ export class AdminUpdateUserRequest {
 
   @IsString()
   @IsOptional()
-  email?: string
-
-  @IsString()
-  @IsOptional()
   last_name?: string
 
   @IsString()
--- a/packages/user/src/migrations/Migration20240213092656.ts
+++ b/packages/user/src/migrations/Migration20240213092656.ts
@@ -8,7 +8,18 @@ export class Migration20240213092656 extends Migration {
     this.addSql(
       'alter table "invite" add column if not exists "email" text not null;'
     )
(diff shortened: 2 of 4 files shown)


packages/core-flows/src/user/steps/update-users.ts
```diff
@@ -0,0 +1,38 @@
...
+    }
+
+    const service = container.resolve(ModuleRegistrationName.USER)
+
+    await service.update(
+      originalUsers.map((u) => ({
+        id: u.id,
+        first_name: u.first_name,
+        last_name: u.last_name,
+        email: u.email,
```
olivermrbl, line 32: thought: we shouldn't allow email updates like this

packages/medusa/src/api-v2/admin/users/validators.ts
```diff
@@ -0,0 +1,116 @@
...
+}
+
+export class AdminUpdateUserRequest {
+  @IsString()
+  @IsOptional()
+  first_name?: string
+
+  @IsString()
+  @IsOptional()
+  email?: string
```
olivermrbl: suggestion: remove email from allowed update fields

packages/user/src/migrations/Migration20240213092656.ts
```diff
@@ -1,22 +1,43 @@
-import { Migration } from '@mikro-orm/migrations';
+import { Migration } from "@mikro-orm/migrations"
```
olivermrbl, line 1: comment,very-nit(this-file): Think we should be consistent in casing, but we can clean up later on

olivermrbl (approved):
LGTM, nice work!

PR #6392, opened 2024-02-13
feat: CartRegion link, definition + workflow

olivermrbl:


packages/link-modules/src/loaders/container.ts
```diff
@@ -1,5 +1,5 @@
-import { BaseRepository, getLinkRepository } from "@repositories"
-import { LinkService, getModuleService } from "@services"
+import { BaseRepository, getLinkRepository } from "../repositories"
```
olivermrbl: question: @carlos-r-l-rodrigues, for some reason, the link module was throwing a `ModuleNotFoundError: Cannot find module '@repositories'` error upon booting the server in my integration tests. Same goes for `@services`.

Do you have any idea why?

packages/link-modules/src/definitions/cart-region.ts
```diff
@@ -0,0 +1,58 @@
...
+    tableName: "cart_region",
+    idPrefix: "cart_reg",
+  },
+  alias: [
+    {
+      name: "cart_region",
+    },
+    {
+      name: "cart_regions",
+    },
```
carlos-r-l-rodrigues:
```suggestion
    {
      name: ["cart_region", "cart_regions"]
    },
```

packages/medusa/src/api-v2/store/carts/route.ts
```diff
@@ -1,23 +1,32 @@
...
+  const variables = { id: result[0].id }
+
+  const query = {
+    cart: {
+      __args: variables,
+      ...defaultStoreCartRemoteQueryObject,
+    },
+  }
+
+  const [cart] = await remoteQuery(query)
```
carlos-r-l-rodrigues: ### 
```suggestion
  const variables = { id: result[0].id }

  const query = {
    cart: {
      ...defaultStoreCartRemoteQueryObject,
    },
  }

  const [cart] = await remoteQuery(query, { cart: variables })
```

srindom (commented):
Should be changed to readonly link

integration-tests/plugins/__tests__/cart/store/create-cart.spec.ts
```diff
@@ -41,6 +46,38 @@ describe("POST /store/carts", () => {
...
+        email: "tony@stark.com",
+        region: expect.objectContaining({
+          id: region.id,
+          currency_code: "usd",
+        }),
+      })
+    )
+  })
+
+  it("should use any region", async () => {
```
sradevski, line 75: Is the region selection deterministic when creating the cart? If so, it would be good to create 2 regions, and assert that eg. the first one is used. If not, we can still assert that the region is present in the response.

packages/core-flows/src/definition/cart/steps/find-one-or-any-region.ts
```diff
@@ -0,0 +1,27 @@
...
+export const findOneOrAnyRegionStepId = "find-one-or-any-region"
+export const findOneOrAnyRegionStep = createStep(
+  findOneOrAnyRegionStepId,
+  async (data: { regionId?: string }, { container }) => {
+    const service = container.resolve<IRegionModuleService>(
+      ModuleRegistrationName.REGION
+    )
+
+    if (!data.regionId) {
+      const regions = await service.list({})
```
sradevski, line 14: Making this deterministic by eg. sorting (maybe there already is some default sorting?) would be a bit better IMO.

packages/core-flows/src/definition/cart/steps/link-cart-region.ts
```diff
@@ -0,0 +1,36 @@
+import { Modules } from "@medusajs/modules-sdk"
+import { StepResponse, createStep } from "@medusajs/workflows-sdk"
+
+export const linkCartRegionStepId = "link-cart-region"
+export const linkCartRegionStep = createStep(
+  linkCartRegionStepId,
+  async (data: { regionId: string; cartId: string }, { container }) => {
```
sradevski: Let's extract the type and use it below as well, so we don't have to do `data!`

packages/medusa/src/api-v2/store/carts/route.ts
```diff
@@ -1,23 +1,31 @@
...
   })
 
   if (Array.isArray(errors) && errors[0]) {
     throw errors[0].error
   }
 
-  res.status(200).json({ cart: result[0] })
+  const remoteQuery = req.scope.resolve("remoteQuery")
+
+  const variables = { id: result[0].id }
```
sradevski, line 20: Is it worth asserting the result length in such cases?

sradevski (commented):
I am still not 100% sure how all of this ties together, so the comments are mostly on style/correctness of the code itself

packages/core-flows/src/definition/cart/steps/find-one-or-any-region.ts
adrien2p: there is no default sorting indeed a default would be nice to have one indeed.

packages/core-flows/src/definition/cart/steps/link-cart-region.ts
adrien2p: The type of data in the rollback function is inferred by the second argument of the step response, so here data will be `{ cartId, regionId } | undefined`. Now, we decided to make it also undefined, because it can happen that the rollback end up being called with no data in case the invoke function fail. I think here it is mostly an oversight, the first thing to do in that rollback is checking if that exists or return early, then data will not required `!`

packages/medusa/src/api-v2/store/carts/route.ts
adrien2p: theoratically at this point, there will always be an item in the collection

packages/core-flows/src/definition/cart/steps/find-one-or-any-region.ts
olivermrbl: Adding a ticket for this @adrien2p

integration-tests/plugins/__tests__/cart/store/create-cart.spec.ts
olivermrbl: @sradevski, can I get you to expand on this? Since we don't have a default sorting, I have kept the test as is.

packages/link-modules/src/definitions/cart-region.ts
```diff
@@ -0,0 +1,28 @@
+import { Modules } from "@medusajs/modules-sdk"
+import { ModuleJoinerConfig } from "@medusajs/types"
+
+export const CartRegion: ModuleJoinerConfig = {
```
olivermrbl, line 4: comment: Since `region_id` lives on the Cart model, we don't need a link table to associate the two entities, so a read-only link is sufficient.

integration-tests/plugins/__tests__/cart/store/create-cart.spec.ts
sradevski: @olivermrbl yeah that is fine until sorting is deterministic, but I guess it's still good to check that the cart.region.id is set in the assertions.

olivermrbl: Good push, this was an oversight from my side. Will add in

packages/link-modules/src/links.ts
```diff
@@ -14,6 +14,12 @@ export const LINKS = {
     Modules.PRICING,
     "price_set_id"
   ),
+  CartRegion: composeLinkName(
```
srindom: q: should this be deleted now that it's read-only?

srindom: I don't completely understand what this file does, but just saw the reference to a table name in the `composeLinkName` function. Since there's no join-table anymore, I thought it might not be needed anymore?

carlos-r-l-rodrigues: this file is only a definition of all known links among Medusa modules.
We could bring these link names to the link definition itself, but there is the case where someone needs to replace a link module by a custom module and can use this to reference the original module name.

olivermrbl: But @srindom is correct in that we don't need it for read-only links right?

carlos-r-l-rodrigues: yes he is. I was just explaing what the file is haha. sorry.

carlos-r-l-rodrigues: read only links do not generate a module. The definition is only used to compsoe the graph.

olivermrbl: > yes he is. I was just explaing what the file is haha. sorry.

All good. I needed a refresher too, so thx! :D

olivermrbl:
Made a couple of additional changes in 1597853:
- Removed old create-cart workflow from the create cart endpoint, because it was messing with some test suites. This is theoretically a breaking change, as people could be using the it, but since the feature is in beta, and we've communicated that things are subject to change, I decided to get rid of it
- Added a transformer to add region currency to cart in case currency code is not passed to the workflow
- Moved workflow input type to `@medusajs/types`

packages/link-modules/src/definitions/cart-region.ts
```diff
@@ -0,0 +1,28 @@
...
+    },
+    {
+      serviceName: Modules.REGION,
+      relationship: {
+        serviceName: Modules.CART,
+        primaryKey: "id",
+        foreignKey: "cart_id",
+        alias: "carts",
+        isList: true,
+      },
```
carlos-r-l-rodrigues:
```suggestion
      serviceName: Modules.REGION,
      relationship: {
        serviceName: Modules.CART,
        primaryKey: "region_id",
        foreignKey: "id",
        alias: "carts",
        isList: true,
      },
```

carlos-r-l-rodrigues: Maybe we can cover these configs on the tests, querying these new properties.

olivermrbl: Something like this: https://github.com/medusajs/medusa/pull/6392/commits/2225f157fe49fdd428128dc778779d38164d15d9

carlos-r-l-rodrigues (approved):
LGTM

PR #6395, opened 2024-02-14
feat(medusa, types, core-flows): Add invite endpoints for api-v2

pKorsholm:
**What**
- Add invite endpoints for simple operations on invites

### DIFF
--- a/.changeset/rude-shirts-act.md
+++ b/.changeset/rude-shirts-act.md
@@ -1,0 +1,5 @@
+---
+"@medusajs/types": patch
+---
+
+feat(cart): `POST /store/carts/:id`
--- a/integration-tests/plugins/__tests__/cart/store/update-cart.ts
+++ b/integration-tests/plugins/__tests__/cart/store/update-cart.ts
@@ -1,0 +1,63 @@
+import { ModuleRegistrationName } from "@medusajs/modules-sdk"
+import { ICartModuleService } from "@medusajs/types"
+import path from "path"
+import { startBootstrapApp } from "../../../../environment-helpers/bootstrap-app"
+import { useApi } from "../../../../environment-helpers/use-api"
+import { getContainer } from "../../../../environment-helpers/use-container"
+import { initDb, useDb } from "../../../../environment-helpers/use-db"
+import adminSeeder from "../../../../helpers/admin-seeder"
+
+jest.setTimeout(50000)
+
+const env = { MEDUSA_FF_MEDUSA_V2: true }
+
+describe("POST /store/carts/:id", () => {
+  let dbConnection
+  let appContainer
+  let shutdownServer
+  let cartModuleService: ICartModuleService
+
+  beforeAll(async () => {
+    const cwd = path.resolve(path.join(__dirname, "..", "..", ".."))
+    dbConnection = await initDb({ cwd, env } as any)
+    shutdownServer = await startBootstrapApp({ cwd, env })
+    appContainer = getContainer()
+    cartModuleService = appContainer.resolve(ModuleRegistrationName.CART)
+  })
+
+  afterAll(async () => {
+    const db = useDb()
+    await db.shutdown()
+    await shutdownServer()
+  })
+
+  beforeEach(async () => {
+    await adminSeeder(dbConnection)
+  })
+
+  afterEach(async () => {
+    const db = useDb()
+    await db.teardown()
+  })
+
+  it("should create a cart", async () => {
+    const api = useApi() as any
+
+    const cart = await cartModuleService.create({
+      currency_code: "usd",
+    })
+
+    const response = await api.post(`/store/carts/${cart.id}`, {
+      email: "tony@stark.com",
+    })
+
+    expect(response.status).toEqual(200)
+    expect(response.data.cart).toEqual(
+      expect.objectContaining({
+        id: cart.id,
+        currency_code: "usd",
+        email: "tony@stark.com",
+      })
+    )
+  })
+})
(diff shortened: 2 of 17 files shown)


packages/core-flows/src/invite/steps/update-invites.ts
```diff
@@ -0,0 +1,50 @@
...
+  async (originalInvites, { container }) => {
+    if (!originalInvites?.length) {
+      return
+    }
+
+    const service = container.resolve(ModuleRegistrationName.USER)
+
+    await service.updateInvites(
+      originalInvites.map((invite) => ({
+        id: invite.id,
```
olivermrbl: suggestion: I would probably also remove email from the updatable fields. In the case of admins writing the email wrong, they would revoke and create a new one. That way, we only have to handle email triggers on creation (and resending). Wdyt?
```suggestion
```

olivermrbl: And perhaps, also expires at 😊

packages/core-flows/src/invite/steps/create-invites.ts
```diff
@@ -0,0 +1,30 @@
...
+    }
+
+    const service: IUserModuleService = container.resolve(
+      ModuleRegistrationName.USER
+    )
+
+    await service.deleteInvites(
+      createdInvites.map((invite: InviteDTO) => invite.id)
+    )
+  }
```
olivermrbl: todo: Missing compensation input
```suggestion
    return new StepResponse(createdInvites, createdInvites.map(inv => inv.id))
  },
  async (createdInvitesIds, { container }) => {
    if (!createdInvitesIds?.length) {
      return
    }

    const service: IUserModuleService = container.resolve(
      ModuleRegistrationName.USER
    )

    await service.deleteInvites(createdInvitesIds)
  }
```

packages/medusa/src/api-v2/admin/invites/[id]/route.ts
```diff
@@ -0,0 +1,41 @@
...
+import { ModuleRegistrationName } from "../../../../../../modules-sdk/dist"
+
+// Get invite
+export const GET = async (req: MedusaRequest, res: MedusaResponse) => {
+  const { id } = req.params
+
+  const moduleService: IUserModuleService = req.scope.resolve(
+    ModuleRegistrationName.USER
+  )
+  const invite = await moduleService.retrieveInvite(id, req.retrieveConfig)
```
Comment (olivermrbl) [inline packages/medusa/src/api-v2/admin/invites/[id]/route.ts]: suggestion: Use remoteQuery

packages/user/src/services/user-module.ts
```diff
@@ -141,6 +141,25 @@ export default class UserModuleService<
     return Array.isArray(data) ? serializedInvites : serializedInvites[0]
   }
 
+  @InjectTransactionManager("baseRepository_")
+  private async createInvites_(
+    data: UserTypes.CreateInviteDTO[],
+    @MedusaContext() sharedContext: Context = {}
+  ): Promise<TInvite[]> {
+    // expiration date in 10 days
+    const expirationDate = new Date().setDate(new Date().getDate() + 10)
```
olivermrbl, line 150: question: Should this be configurable?

olivermrbl (changes requested):

packages/core-flows/src/invite/steps/update-invites.ts
pKorsholm: yes, I agree, I've not created an update endpoint either as all updates will exist in terms of re-issuing the invite 😅 I'll remove the workflow entirely as it's not used

packages/user/src/services/user-module.ts
pKorsholm: we could make it configurable with options, I'll include that when looking at the invite flow instead of here so this can land on dev asap?

olivermrbl: Let's do it 👍

packages/core-flows/src/invite/steps/delete-invites.ts
```diff
@@ -0,0 +1,28 @@
...
+
+export const deleteInvitesStepId = "delete-invites-step"
+export const deleteInvitesStep = createStep(
+  deleteInvitesStepId,
+  async (input: string[], { container }) => {
+    const service: IUserModuleService = container.resolve(
+      ModuleRegistrationName.USER
+    )
+
+    await service.softDeleteInvites(input)
```
adrien2p, line 13: todo: I would suggest to be careful here, just to be sure, but if you pass an empty array it will soft delete all the invites table. So maybe like the other one we should check that there is length before

packages/medusa/src/api-v2/admin/invites/[id]/route.ts
Comment (adrien2p) [reply packages/medusa/src/api-v2/admin/invites/[id]/route.ts]: we will have to migrate to remote query indeed, maybe another pr to not block

adrien2p (commented):
LGTM, one todo but let me know, very clean work 🔥

adrien2p (approved):

packages/core-flows/src/invite/steps/delete-invites.ts
olivermrbl: To me, sounds like we should have this check further down the stack. The operation of deleting all rows should not be "hidden away". Would want that to require a very intentional action 😅

adrien2p: yes true done https://github.com/medusajs/medusa/pull/6396

olivermrbl (approved):

PR #6399, opened 2024-02-14
feat(order): module foundation

carlos-r-l-rodrigues:
What:
Initial Order module foundation.
Basic models and methods to manage entites like in Cart Module

### DIFF
--- a/packages/order/README.md
+++ b/packages/order/README.md
@@ -1,1 +1,1 @@
-# Fulfillment Module
+# Order Module
--- a/packages/order/src/models/adjustment-line.ts
+++ b/packages/order/src/models/adjustment-line.ts
@@ -1,4 +1,5 @@
-import { DAL } from "@medusajs/types"
+import { BigNumberRawValue, DAL } from "@medusajs/types"
+import { BigNumber } from "@medusajs/utils"
 import { OptionalProps, PrimaryKey, Property } from "@mikro-orm/core"
 
 type OptionalAdjustmentLineProps = DAL.EntityDateColumns
@@ -25,8 +26,11 @@ export default abstract class AdjustmentLine {
   @Property({ columnType: "text", nullable: true })
   code: string | null = null
 
-  @Property({ columnType: "numeric", serializer: Number })
-  amount: number
+  @Property({ columnType: "numeric" })
+  amount: BigNumber | number
+
+  @Property({ columnType: "jsonb" })
+  raw_amount: BigNumberRawValue
 
   @Property({ columnType: "text", nullable: true })
   provider_id: string | null = null
--- a/packages/medusa/src/models/order.ts
+++ b/packages/medusa/src/models/order.ts
@@ -19,6 +19,7 @@ import {
   FeatureFlagDecorators,
 } from "../utils/feature-flag-decorators"
 
+import { MedusaV2Flag } from "@medusajs/utils"
 import { BaseEntity } from "../interfaces/models/base-entity"
 import { generateEntityId } from "../utils/generate-entity-id"
 import { manualAutoIncrement } from "../utils/manual-auto-increment"
@@ -41,7 +42,6 @@ import { Return } from "./return"
 import { SalesChannel } from "./sales-channel"
 import { ShippingMethod } from "./shipping-method"
 import { Swap } from "./swap"
-import { MedusaV2Flag } from "@medusajs/utils"
 
 /**
  * @enum
--- a/packages/order/src/models/line-item.ts
+++ b/packages/order/src/models/line-item.ts
@@ -66,8 +66,11 @@ export default class LineItem {
   @Property({ columnType: "text", nullable: true })
   thumbnail: string | null = null
 
-  @Property({ columnType: "integer" })
-  quantity: number
+  @Property({ columnType: "numeric" })
+  quantity: BigNumber | number
+
+  @Property({ columnType: "jsonb" })
+  raw_quantity: BigNumberRawValue
 
   @Property({
     columnType: "text",
--- a/packages/order/src/models/tax-line.ts
+++ b/packages/order/src/models/tax-line.ts
@@ -1,3 +1,5 @@
+import { BigNumberRawValue } from "@medusajs/types"
+import { BigNumber } from "@medusajs/utils"
 import { PrimaryKey, Property } from "@mikro-orm/core"
 
 /**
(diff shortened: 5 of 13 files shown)


packages/order/README.md
```diff
@@ -0,0 +1 @@
+# Fulfillment Module
```
olivermrbl:
```suggestion
# Order Module
```

packages/order/src/models/adjustment-line.ts
```diff
@@ -0,0 +1,48 @@
...
+    columnType: "text",
+    nullable: true,
+  })
+  promotion_id: string | null = null
+
+  @Property({ columnType: "text", nullable: true })
+  code: string | null = null
+
+  @Property({ columnType: "numeric", serializer: Number })
+  amount: number
```
olivermrbl: thought(non-blocking): Should we introduce BigNumbers? I still have some refactoring to do in the cart module. Not all models have been updated yet.

packages/utils/src/common/create-psql-index-helper.ts
```diff
@@ -35,17 +37,38 @@ export function createPsqlIndexStatementHelper({
...
+    name,
+    expression,
+    MikroORMIndex: (options = {}) => {
+      return Index({
+        name,
+        expression,
+        ...options,
+      })
+    },
+  }
```
olivermrbl, line 73: question: What's the idea here? And why did we change the fulfillment (among others) module indexes? Seems weird to have a discrepancy here.

adrien2p (approved):
As we discuss, we can continue from there, things will change later but it gives us a structure to work from 👍

packages/utils/src/common/create-psql-index-helper.ts
adrien2p: The idea is to be able to construct the index object and immediately apply it to the props without having to explicitly declared a name and build the expression and then having to apply them all.

```ts
const Index = createPsqlIndexHelper({
  name: "IDX",
  ...
})

class Entity {
  @Index.MikroORMIndex()
  prop: something
}
```

instead of 

```ts
const indexName = 'IDX'
const index = createPsqlIndexHelper({
  name: indexName,
  ...
})

class Entity {
  @Index({
    name: indexName,
    expression: index
  })
  prop: something
}
```

olivermrbl: Sorry, I realise I phrased my question wrong. Why do we need the different return values and why are we not using it the same why in the modules?

adrien2p: You can have a need for the values for example like in the customer module where you need to index name in order to handle the error manually, in that case this index is exported and you can extract the name from it. The expression could be also needed why not. The idea was to not fully changed the other modules in order to reduce the amount of conflict. So the other modules are just extracting the expression so that only the indexes are touched and not the entities

packages/order/src/services/order-module-service.ts
```diff
@@ -0,0 +1,1280 @@
...
+    data: OrderTypes.CreateOrderLineItemForOrderDTO[]
+  ): Promise<OrderTypes.OrderLineItemDTO[]>
+  addLineItems(
+    orderId: string,
+    items: OrderTypes.CreateOrderLineItemDTO[],
+    sharedContext?: Context
+  ): Promise<OrderTypes.OrderLineItemDTO[]>
+
+  @InjectManager("baseRepository_")
+  async addLineItems(
```
srindom, line 247: thought: we should probably have a check here eventually that verifies that the order is in a state where it's allowed to have line items added externally. For example, once the order is placed adding line items should go through the mechanism we come up with for edits.

packages/utils/src/order/status.ts
```diff
@@ -0,0 +1,31 @@
+/**
```
srindom, line 1: q: shouldn't this be in types?

packages/order/src/models/order.ts
```diff
@@ -0,0 +1,168 @@
...
+})
+
+const currencyCodeIndex = createPsqlIndexStatementHelper({
+  tableName: "order",
+  columns: "currency_code",
+  where: "deleted_at IS NOT NULL",
+})
+
+@Entity({ tableName: "order" })
+export default class Order {
```
srindom, line 60: todo: add a canceled_at field here.

srindom: We might also want to have a `placed_at` here. E.g., to indicate when an order moves from a draft state to being placed/confirmed.

packages/order/src/services/order-module-service.ts
srindom: Similar thinking for some of the other functions.

srindom (approved):
LGTM w. comments

packages/order/README.md
carlos-r-l-rodrigues: my IDE is ignoring .md files to search 😆

packages/order/src/models/adjustment-line.ts
carlos-r-l-rodrigues: yes, good catch I will review all the entities.

packages/utils/src/common/create-psql-index-helper.ts
carlos-r-l-rodrigues: I didn't want to change the default behavior of the helper, and regardless the ORM used, the expression in plain sql might be needed.
I will refactor all the other modules in a next PR. as Adrien said, I'm avoiding a big number of conflicts for now.

packages/utils/src/order/status.ts
carlos-r-l-rodrigues: unfortunately enums are transpiled to objects. and things that generate code we have to move to utils.

packages/order/src/services/order-module-service.ts
carlos-r-l-rodrigues: I agree, I am just avoiding in this PR handle things that are uncertain.
When we define the order edits, and what will be the rules, all these services will have to be revisited anyway.
In theory we could have multiple versions of the same order and the service will have to handle/validate that.
Same for the field `placed_at`. I will wait more definition how we'll manage order changes to change that.

olivermrbl (approved):
LGTM, great piece of work!

packages/utils/src/order/status.ts
adrien2p: the approach so far is having an enum in the utils and a type in the types

packages/order/src/services/order-module-service.ts
srindom: Yep all good!

packages/utils/src/order/status.ts
srindom: Got it!

PR #6401, opened 2024-02-14
docs(plugin): paypal integration missing createOrder callback

raiyansarker:
This PR adds missing information in paypal plugin docs which is crucial to its use. Previously, docs didn't mention about createOrder callback that is required by PayPal sdk to connect with backend for details like currency and amount to be charged. Medusa backend responds with required id for the PayPal sdk to use. Related to https://github.com/medusajs/nextjs-starter-medusa/issues/260#issuecomment-1944157985


shahednasser (approved):
Sorry for leaving this PR hanging, thank you for your contribution!

PR #6409, opened 2024-02-15
docs(payments): Fix incorrect reference in payment processor tsdoc

### LINKED ISSUE #6394: The payment processor is wrongly referred to as fullfilment provider
### Preliminary Checks

- [X] This issue is not a duplicate. Before opening a new issue, please search existing issues: https://github.com/medusajs/medusa/issues

### Issue Summary

The subsection **parameters** of constructor section of the document **How to Create a Payment Processor** wrongly refers it to  fulfillment provider instead of payment processor. Please see the attachment.
Error

1. _Edit the documentation and change the text to read If this payment processor is created in a plugin, the plugin's options are passed in this parameter._
2.
3.
...

- [ ] I would like to fix this issue

sradevski:
Resolves #6394

### DIFF
--- a/packages/medusa/src/interfaces/payment-processor.ts
+++ b/packages/medusa/src/interfaces/payment-processor.ts
@@ -653,7 +653,7 @@ export abstract class AbstractPaymentProcessor implements PaymentProcessor {
    * you can access it in the constructor. The options are passed as a second parameter.
    *
    * @param {Record<string, unknown>} container - An instance of `MedusaContainer` that allows you to access other resources, such as services, in your Medusa backend through [dependency injection](https://docs.medusajs.com/development/fundamentals/dependency-injection)
-   * @param {Record<string, unknown>} config - If this fulfillment provider is created in a plugin, the plugin's options are passed in this parameter.
+   * @param {Record<string, unknown>} config - If this payment processor is created in a plugin, the plugin's options are passed in this parameter.
    *
    * @example
    * ```ts
--- a/packages/utils/src/payment/abstract-payment-provider.ts
+++ b/packages/utils/src/payment/abstract-payment-provider.ts
@@ -18,7 +18,7 @@ export abstract class AbstractPaymentProvider implements IPaymentProvider {
    * you can access it in the constructor. The options are passed as a second parameter.
    *
    * @param {MedusaContainer} container - An instance of `MedusaContainer` that allows you to access other resources, such as services, in your Medusa backend through [dependency injection](https://docs.medusajs.com/development/fundamentals/dependency-injection)
-   * @param {Record<string, unknown>} config - If this fulfillment provider is created in a plugin, the plugin's options are passed in this parameter.
+   * @param {Record<string, unknown>} config - If this payment processor is created in a plugin, the plugin's options are passed in this parameter.
    *
    * @example
    * ```ts


packages/medusa/src/interfaces/payment-processor.ts
```diff
@@ -653,7 +653,7 @@ export abstract class AbstractPaymentProcessor implements PaymentProcessor {
    * you can access it in the constructor. The options are passed as a second parameter.
    *
    * @param {Record<string, unknown>} container - An instance of `MedusaContainer` that allows you to access other resources, such as services, in your Medusa backend through [dependency injection](https://docs.medusajs.com/development/fundamentals/dependency-injection)
-   * @param {Record<string, unknown>} config - If this fulfillment provider is created in a plugin, the plugin's options are passed in this parameter.
+   * @param {Record<string, unknown>} options - If this payment processor is created in a plugin, the plugin's options are passed in this parameter.
```
adrien2p: Q: you renamed it to options in the TSDoc, but the constructor argument is named config, what is the reason?

packages/utils/src/payment/abstract-payment-provider.ts
```diff
@@ -18,7 +18,7 @@ export abstract class AbstractPaymentProvider implements IPaymentProvider {
    * you can access it in the constructor. The options are passed as a second parameter.
    *
    * @param {MedusaContainer} container - An instance of `MedusaContainer` that allows you to access other resources, such as services, in your Medusa backend through [dependency injection](https://docs.medusajs.com/development/fundamentals/dependency-injection)
-   * @param {Record<string, unknown>} config - If this fulfillment provider is created in a plugin, the plugin's options are passed in this parameter.
+   * @param {Record<string, unknown>} options - If this payment processor is created in a plugin, the plugin's options are passed in this parameter.
```
shahednasser: Same comment as adrien, would revert `options` back to `config`.

shahednasser (changes requested):

packages/medusa/src/interfaces/payment-processor.ts
sradevski: @adrien2p changing the constructor to `options` would be a breaking change, but everywhere else in the docs we refer to it as options, so like this it is a bit more correct (if you see, just in the paragraph above we talk about options). WDYT? cc @shahednasser

adrien2p: It wouldn't really be a breaking change as the name of the argument is not enforced, it can be named with whatever name and would work as the important thing in a constructor is the number of args and their type. I agree that it would be a breaking change if we change the type or if it was an object with named properties. Don you agree?

sradevski: @adrien2p typically that'd be true, but since in eg. `medusa-payments-stripe` we override the `options` property it fails when running the tests.

```
 Type 'StripeOptions' is not assignable to type 'Record<string, unknown>'.
medusa-payment-stripe:build:     Index signature for type 'string' is missing in type 'StripeOptions'.
medusa-payment-stripe:build: src/core/stripe-base.ts(43,7): error TS2611: 'options' is defined as a property in class 'AbstractPaymentProcessor', but is overridden here in 'StripeBase' as an accessor.
```

This is the error we get.

adrien2p: let's discuss that, but in the plugin, we are assigning the second parameter to the class property options. no matter the name of the second argument, when passed to super, it will be the options in the base service class do you agree? The options you are referring to, is the getter of the class property options which returns this.option_ that has been set from the constructor using the second argument.

So the real issue here, is that in the doc it is auto assigning the second argument on the class by setting the visibility from the argument, which try to assign options on the class except that the class base option is only a getter accessor and not a property. So we shouldn't say to do `constructor(protected readonly options)` since the accessor already exists. In fact, I don't think we should suggest to shortcut property definition here and let the user choose how to define it and see that if he tries to set options on the class it can't because it is an accessor and not a prop. Let's discuss that in our call if you want

adrien2p: So in fact it become breaking change because of the auto assignement on the class due to the usage of protected readonly in the constructor 😂 which we shouldn't show in the doc and let the user choose the way he want to assign it on the class, it could be that the user would like to assign it to __options__ for example

adrien2p: cc @shahednasser why do we tell the user to auto set the argument prop to the class? It can be a breaking thing in those scenario, we should let the user choose how to assign it on his side no?

shahednasser: @adrien2p can you clarify what you're referring to? I'm not sure I understand 😅

adrien2p: No worries, I can explain you later ahah

shahednasser: So what is the next action now? Are we renaming to `options` or keeping it as `config`?

adrien2p: we can just rename the tsdoc as option to match the actual doc, and keep config in the constructor for now

shahednasser: We can't because when generating the reference it'll look for `config` in the TSDoc to find its description. So, if we decide to keep it as config now we should change it back in the TSDoc as well

sradevski: @shahednasser we discussed in a call with Adrien, but basically we cannot change the constructor now because it will cause a breaking change, so that one we'll have to keep named as config. Would you be happy to proceed with the other changes, since as I said, everywhere else in the document we refer to `options`, and this only means the constructor object is called `config`, but we use `options` everywhere else. The other option is to rename everything to `config`, but I think that is a step in the wrong direction. WDYT?

packages/utils/src/payment/abstract-payment-provider.ts
sradevski: @shahednasser I guess this became overly complicated, I just reverted it 😄

PR #6418, opened 2024-02-16
feat: Cart SalesChannel link + clean-up

olivermrbl:
**What**
- Change CartSalesChannel link to read-only
- Remove CartSalesChannel link table
- Remove CartSalesChannel migration
- Remove transaction decorators from sales channel module service methods**
- Remove old CreateCart workflow + all related handlers
- Consolidates cart links tests
- Consolidates cart integration tests

**Doing so creates the transaction a level deeper than usual - namely in the `this.salesChannelService.[method]` - so the serialize-call is outside of the transaction.

### DIFF
--- a/.github/teams.yml
+++ b/.github/teams.yml
@@ -10,3 +10,4 @@
   - "@kasperkristensen"
   - "@pKorsholm"
   - "@sradevski"
+  - "@edast"


packages/medusa/src/services/cart.ts
```diff
@@ -61,7 +61,7 @@ import {
 } from "../types/common"
 import { buildQuery, isString, setMetadata } from "../utils"
 
-import { Modules, RemoteLink } from "@medusajs/modules-sdk"
+import { RemoteLink } from "@medusajs/modules-sdk"
```
fPolic: todo: we can remove `RemoteLink` completely from this file

carlos-r-l-rodrigues (approved):
🥇 LGTM

PR #6420, opened 2024-02-16
Update portuguese translations

joevtap:
- **What** - Portuguese translations update
- **Why** - Admin UI translations for portuguese were outdated
- **How** - Merged english translations file with current portuguese translations file and made the update over values for the diff keys
- **Testing** - Couldn't run the project locally, but I basically just changed values for keys from the translations file, it should be working


JoKenPo:
Great!
Eager for the approval 👍🏽

joevtap:
I added a chore commit reformatting the file to use 4 spaces instead of 2 (because the diff view was messy with thousands of changes) and notice some typos in translations tokens, for example: 

- "Calculated" spelled as "Calcualted" both in the token name from translation file and in the source code using said token
- Same thing with "Occurred" being spelled as "Occured"

Since the scope of this PR is translation only, I kept the typos and "duplication" of keys (quoted because just the spelling is different, the key is tecnically the same).

olivermrbl:
@joevtap, we appreciate the contribution. However, our admin system is currently undergoing a redesign that includes improvements to and simplifying of translations and we will therefore not accept this PR. 

We expect to go live with the update next month. If you are still interested in contributing then, we would very much appreciate a new PR with changes that follow the new format.

I apologize for the slow response time.

joevtap:
> @joevtap, we appreciate the contribution. However, our admin system is currently undergoing a redesign that includes improvements to and simplifying of translations and we will therefore not accept this PR.
> 
> We expect to go live with the update next month. If you are still interested in contributing then, we would very much appreciate a new PR with changes that follow the new format.
> 
> I apologize for the slow response time.

I see. Then, when the time comes, I will be happy to make another PR.

PR #6421, opened 2024-02-16
feat(tax): Add TaxRegion

srindom:
**What**
- Adds a TaxRegion entity.

**For context: High-level design of the Tax module**
- A TaxRegion scopes tax rates to a geographical place.
- You can define tax regions on two levels: country-level, province-level (this corresponds to state in US contexts)
- Each Tax Region can have a default Tax Rate.
-  [not yet done] - Each Tax Region can also have granularly defined tax rates for different products and shipping rates. For example, California can have a base rate for default products, but a reduced rate for groceries.
- Tax Rates specify if they can be combined with other rates - it's always the lowest level rate that wins.

The above allows a merchant to define their tax settings along the lines of this:

- Denmark (Region)
  - Default rate: 25% (TaxRate)
- Germany (Region)
  - Default rate: 19% (TaxRate)
  - Reduced rate (books): 9% (TaxRate w. rule)
- United States (Region)
  - Default rate: 0% (TaxRate)
  - California: (Region)
    - Default rate: 7.25% (TaxRate)
  - Arkansas: (Region)
    - Default rate: 6.5%
    - Reduced rate (groceries): 0.125% (TaxRate w. rule)

The TaxModule can then receive a list of products and the shipping address to determine what tax rates apply to the line items.


packages/tax/src/services/tax-module-service.ts
```diff
@@ -78,4 +84,48 @@ export default class TaxModuleService<TTaxRate extends TaxRate = TaxRate>
   ) {
     return await this.taxRateService_.create(data, sharedContext)
   }
+
+  async createTaxRegions(
+    data: TaxTypes.CreateTaxRegionDTO[],
```
olivermrbl, line 89: nit: Allow the creation of a single tax region. Don't think it's important; would only be for consistency purposes

olivermrbl (approved):
LGTM

packages/tax/src/services/tax-module-service.ts
srindom: Yep would add that later as well!

PR #6422, opened 2024-02-16
feat(tax): introduce tax override data models

srindom:
**What**
- Adds Tax rules to allow overrides of tax rates for certain products, product types or shipping options.

**Punted to future PR**
- Currently, the creation methods only support bulk operations. A later PR will include support for singular operations, too.
- It should be possible to add products, types, and shipping options to a tax rate after creating it. Add, remove, update will come in a later PR.


carlos-r-l-rodrigues:
If we make them specific for product, shipping, product_type, wouldn't it make sense to have them as link definitions?

If we are to keep them in the module, we could just create a single generic table, where you specify a `reference` and a `reference_id` and enable these specific taxes for anything.

srindom:
> If we make them specific for product, shipping, product_type, wouldn't it make sense to have them as link definitions?

I wanted to include them into the tax module because this gives the tax module the feature of having tax overrides. If the rules only exist as relations we make it a task for the consumer to find the right tax rates (as opposed to exposing a `taxModuleService.getItemTaxLines(arrayOfItemLikeStructures, calculationContextEgAddress)`). It would probably make sense to have readonly links here to make them queriable with remote query, but wouldn't consider it strictly necessary.

> If we are to keep them in the module, we could just create a single generic table, where you specify a reference and a reference_id and enable these specific taxes for anything.

Yeah this might actually be better - is it possible to have links like this?

carlos-r-l-rodrigues:
> Yeah this might actually be better - is it possible to have links like this?

Other modules are copying the TaxRate internally at the moment. I am not sure I get which link you mean.
These would be used to calculate a tax rate given a context no? where, type is included.

srindom:
>  I am not sure I get which link you mean.

The one I was thinking about was one that could be used in remoteQuery to allow a query like this:

```
product {
  tax_rate_rules {
    tax_rate {
      rate
    }
  }
}
```

I.e., if it were possible for remote query to understand a link like "reference_type + reference_id".

But this was more out of curiosity, not something that is required at all.

packages/tax/src/models/tax-rate-rule.ts
```diff
@@ -0,0 +1,55 @@
...
+  @PrimaryKey({ columnType: "text" })
+  tax_rate_id!: string
+
+  @PrimaryKey({ columnType: "text" })
+  reference_id!: string;
+
+  [PrimaryKeyProp]?: ["tax_rate_id", "reference_id"]
+
+  @Property({ columnType: "text" })
+  reference_type: string
```
carlos-r-l-rodrigues: Depending on how filters will be passed to fetch data from this table, probably individual indexes on tax_rate_id, reference_id and reference_type will be a better choice than this composite primary key.
Keeping the composite key and adding additional indexes is also an option.

carlos-r-l-rodrigues (approved):

carlos-r-l-rodrigues:
> The one I was thinking about was one that could be used in remoteQuery to allow a query like this:
> 
> ```
> product {
>   tax_rate_rules {
>     tax_rate {
>       rate
>     }
>   }
> }
> ```
> 
> I.e., if it were possible for remote query to understand a link like "reference_type + reference_id".
> 
> But this was more out of curiosity, not something that is required at all.

This isn't implemented at the moment, but it's definitely doable

packages/tax/src/services/tax-module-service.ts
```diff
@@ -112,7 +128,7 @@ export default class TaxModuleService<
       sharedContext
     )
 
-    const rates = regions.map((region, i) => {
+    const rates = regions.map((region: TaxRegionDTO, i: number) => {
```
sradevski: Doesn't taxRegionService_.create return a typed object? If so, the typings inside the map shouldn't be necessary as they will be inferred.

packages/types/src/tax/common.ts
```diff
@@ -73,3 +73,29 @@ export interface FilterableTaxRegionProps
   updated_at?: OperatorMap<string>
   created_by?: string | string[] | OperatorMap<string>
 }
+
+export interface TaxRateRuleDTO {
+  reference_type: string
```
sradevski: I saw in Carlos' PR that this is called reference, might be better to stick to one for consistency.

PR #6423, opened 2024-02-17
Fix payment plugin loader

lukebui:
In the current version, PaymentService no longer has the isPaymentService function.


olivermrbl:
@lukebui, can I get you to explain the changes here?

The `AbstractPaymentService` still ships with the static helper:
https://github.com/medusajs/medusa/blob/1ba35b02dd52eeca9f3e1bee073c5e7a17edbc33/packages/medusa/src/interfaces/payment-service.ts#L154-L156

Althought, this abstract class + interface is deprecated and will be removed in 2.0.

Closing until more has been added.

PR #6426, opened 2024-02-18
feat: Cart Customer link + create cart with customer

olivermrbl:
**What**
- Add Cart Customer read-only link
- Add `findOrCreateCustomer` step to `CreateCartWorkflow`. Three scenarios are covered:
  - Retrieve customer from passed ID
  - Retrieve customer from passed email
  - Create customer with passed email
- Use signed-in customer id in `POST /store/carts`
- Restructure tests

**Tests**
- Cart Customer link queries
- Create cart with existing customer
- Create cart creates customer using email
- Compensate: Should delete customer, if created from email
- Compensate: Should not delete customer, if customer already existed

_Note: Compensation flows are tested using the local workflow instead of the API, as we cannot modify the global workflow (yet)_

### DIFF
--- a/integration-tests/plugins/__tests__/cart/store/carts.spec.ts
+++ b/integration-tests/plugins/__tests__/cart/store/carts.spec.ts
@@ -15,6 +15,7 @@ import { useApi } from "../../../../environment-helpers/use-api"
 import { getContainer } from "../../../../environment-helpers/use-container"
 import { initDb, useDb } from "../../../../environment-helpers/use-db"
 import adminSeeder from "../../../../helpers/admin-seeder"
+import { createAuthenticatedCustomer } from "../../../helpers/create-authenticated-customer"
 
 jest.setTimeout(50000)
 
@@ -76,17 +77,13 @@ describe("Store Carts API", () => {
         name: "Webshop",
       })
 
-      const customer = await customerModule.create({
-        email: "tony@stark.com",
-      })
-
       const api = useApi() as any
 
       const created = await api.post(`/store/carts`, {
+        email: "tony@stark.com",
         currency_code: "usd",
         region_id: region.id,
         sales_channel_id: salesChannel.id,
-        customer_id: customer.id,
       })
 
       expect(created.status).toEqual(200)
@@ -101,7 +98,6 @@ describe("Store Carts API", () => {
           }),
           sales_channel_id: salesChannel.id,
           customer: expect.objectContaining({
-            id: customer.id,
             email: "tony@stark.com",
           }),
         })
@@ -180,6 +176,32 @@ describe("Store Carts API", () => {
       )
     })
 
+    it("should create cart with logged-in customer", async () => {
+      const { customer, jwt } = await createAuthenticatedCustomer(appContainer)
+
+      const api = useApi() as any
+      const response = await api.post(
+        `/store/carts`,
+        {},
+        {
+          headers: { authorization: `Bearer ${jwt}` },
+        }
+      )
+
+      expect(response.status).toEqual(200)
+      expect(response.data.cart).toEqual(
+        expect.objectContaining({
+          id: response.data.cart.id,
+          currency_code: "dkk",
+          email: customer.email,
+          customer: expect.objectContaining({
+            id: customer.id,
+            email: customer.email,
+          }),
+        })
+      )
+    })
+
     it("should throw when no regions exist", async () => {
       const api = useApi() as any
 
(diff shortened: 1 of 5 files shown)


packages/core-flows/src/definition/cart/steps/find-or-create-customer.ts
```diff
@@ -0,0 +1,77 @@
...
+
+export interface StepOutput {
+  customer?: any
+  email?: string
+}
+
+interface StepCompensateInput {
+  customer?: any
+  customerWasCreated: boolean
+}
```
olivermrbl: comment: TODO

integration-tests/plugins/__tests__/cart/store/carts.spec.ts
```diff
@@ -41,167 +50,331 @@ describe("POST /store/carts", () => {
...
+      const created = await api.post(`/store/carts`, {
         currency_code: "usd",
-        email: "tony@stark.com",
-        region: expect.objectContaining({
-          id: region.id,
-          currency_code: "usd",
-        }),
+        region_id: region.id,
         sales_channel_id: salesChannel.id,
+        customer_id: customer.id,
```
srindom: thought: I don't think we should allow passing the customer id directly form the client. Either you get a guest customer or you use the logged in customer's id. Otherwise, you would be able to impersonate others.

integration-tests/plugins/__tests__/cart/store/carts.spec.ts
```diff
@@ -41,167 +50,331 @@ describe("POST /store/carts", () => {
...
+          "throw",
+          findOrCreateCustomerStepId,
+          {
+            invoke: async function failStep() {
+              throw new Error(`Failed to create cart`)
+            },
+          },
+          {
+            noCompensation: true,
+          }
```
srindom: q: why is this necessary? Doesn't the step just fail and only previous steps compensate?

packages/medusa/src/api-v2/store/carts/route.ts
```diff
@@ -1,14 +1,19 @@
...
+import { StorePostCartsCartReq } from "./validators"
 
 export const POST = async (req: MedusaRequest, res: MedusaResponse) => {
-  const workflow = createCartWorkflow(req.scope)
+  const input = req.validatedBody as StorePostCartsCartReq
 
-  const { result, errors } = await workflow.run({
-    input: req.validatedBody as CreateCartDTO,
+  // If the customer is logged in, we auto-assign them to the cart
+  if (req?.user?.customer_id) {
```
srindom: thought: not sure if we made changes but the customer id from the authuser has been dug out like so in the /me route:

https://github.com/medusajs/medusa/blob/develop/packages/medusa/src/api-v2/store/customers/me/route.ts

integration-tests/plugins/__tests__/cart/store/carts.spec.ts
olivermrbl: Do you mean they will be able to impersonate others by "guessing" their customer ID? Many other platforms' Cart APIs allow customer ID. Are you sure we should remove it?

We've also always had `customer_id` exposed in the update cart endpoint.

packages/medusa/src/api-v2/store/carts/route.ts
olivermrbl: Nice catch, I forgot to add in the new middleware.

integration-tests/plugins/__tests__/cart/store/carts.spec.ts
srindom: My thinking is that if someone has an account then that person must be logged in, in order to have their customer id associated with the cart. 

Guessing is close to being mathematically impossible but if you somehow have your customer id exposed it should still be protected by a login to be used. 

For guest customers it is less important but I don't really see a need for a customer id being editable from outside in regular use cases (i.e., our default apis scope).

olivermrbl: > Guessing is close to being mathematically impossible

Yeah, this is why I was questioning it, but I think your other points make sense. Will remove it.

integration-tests/plugins/__tests__/cart/store/carts.spec.ts
olivermrbl: Made a c/p. Think this can def. be removed.

srindom (approved):
LGTM!

packages/medusa/src/api-v2/store/carts/middlewares.ts
```diff
@@ -8,6 +9,15 @@ import {
 } from "./validators"
 
 export const storeCartRoutesMiddlewares: MiddlewareRoute[] = [
+  {
+    method: "ALL",
```
olivermrbl, line 13: @pKorsholm, can I get you to verify the setup here is correct?

PR #6427, opened 2024-02-18
feat(core-flow, medusa): Create cart with Sales Channel

olivermrbl:
**What**
- Add `findSalesChannelStep` to `CreateCartWorkflow`. There are a couple of todos:
  - Find default sales channel configured on the store
  - Implement publishable API key work
- Parallelize the data populating steps in `CreateCartWorkflow`

### DIFF
--- a/packages/core-flows/src/definition/cart/steps/find-sales-channel.ts
+++ b/packages/core-flows/src/definition/cart/steps/find-sales-channel.ts
@@ -5,9 +5,6 @@ import { StepResponse, createStep } from "@medusajs/workflows-sdk"
 
 interface StepInput {
   salesChannelId?: string
-  publishableApiKeyScopes?: {
-    salesChannelIds?: string[]
-  }
 }
 
 export const findSalesChannelStepId = "find-sales-channel"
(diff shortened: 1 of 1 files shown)


packages/core-flows/src/definition/cart/steps/find-sales-channel.ts
```diff
@@ -0,0 +1,51 @@
...
+  async (data: StepInput, { container }) => {
+    const salesChannelService = container.resolve<ISalesChannelModuleService>(
+      ModuleRegistrationName.SALES_CHANNEL
+    )
+
+    let { salesChannelId, publishableApiKeyScopes } = data
+
+    // TODO: Revisit when API key module is implemented
+    if (salesChannelId && publishableApiKeyScopes?.salesChannelIds?.length) {
+      if (publishableApiKeyScopes.salesChannelIds.length > 1) {
```
srindom: thought: would probably make the workflow independent of the publishable key and just handle this in the API layer instead.

srindom (approved):
LGTM - although think we should move sales channel resolution from a publishable key out of the workflow and into the API layer instead. We can revisit this when the API key module is done.

packages/core-flows/src/definition/cart/steps/find-sales-channel.ts
olivermrbl: Good push. I wanted to wait for the API key module to put thought into this part of the flow, but I'll just remove it for now.

packages/core-flows/src/definition/cart/steps/find-sales-channel.ts
```diff
@@ -0,0 +1,34 @@
...
+    const salesChannelService = container.resolve<ISalesChannelModuleService>(
+      ModuleRegistrationName.SALES_CHANNEL
+    )
+
+    let salesChannel: SalesChannelDTO | undefined
+    if (data.salesChannelId) {
+      salesChannel = await salesChannelService.retrieve(data.salesChannelId)
+    } else {
+      // TODO: Find default sales channel from store
+    }
```
srindom, line 23: thought(non-blocking): we could even just make it a requirement that the sales channel is passed to the workflow and avoid any fallback. But let's revisit the question when we have publishable keys.

srindom (approved):
LGTM!

PR #6431, opened 2024-02-19
feat(modules-sdk, types, user, utils):init user module events

pKorsholm:
**What**
- Add events handling to the user module
- Add `emitEvents` decorator to inject message aggregator into medusa context and emit events after method execution
- Add `emitEvents_` method to `abstractModuleService` in abstract module service factory


packages/user/integration-tests/__fixtures__/event-bus.ts
```diff
@@ -0,0 +1,43 @@
+import {
+  EmitData,
+  EventBusTypes,
+  IEventBusModuleService,
+  Message,
+  Subscriber,
+} from "@medusajs/types"
+
+export class EventBusService implements IEventBusModuleService {
```
olivermrbl: nit(non-blocking): Could be moved to be general-purpose. I imagine the same will be needed for all modules

packages/user/src/services/user-module.ts
```diff
@@ -96,6 +101,20 @@ export default class UserModuleService<
       populate: true,
     })
 
+    sharedContext.messageAggregator?.save(
```
olivermrbl: question: I think we should look into exposing a simpler API for the `messageAggregator`. If we will be using `buildEventMessages` everywhere, we might as well think in ways of baking it in.

Ideally, we can do something like:
```ts
 sharedContext.messageAggregator?.save(
        users.map((user) => ({
          eventName: UserEvents.CREATED,
          metadata: {
            service: this.constructor.name,
            action: "create",
            object: "user",
          },
          data: user.id,
        })
      )
    )
```

And also allow for a single event to be passed, as we currently do with our event bus:
```ts
sharedContext.messageAggregator?.save({
  eventName: UserEvents.CREATED,
  metadata: {
    service: this.constructor.name,
    action: "create",
    object: "user",
  },
  data: user.id,
})
```

cc @carlos-r-l-rodrigues

olivermrbl (approved):

packages/medusa-test-utils/src/module-test-runner.ts
```diff
@@ -61,6 +63,7 @@ export function moduleIntegrationTestRunner({
   const moduleOptions: InitModulesOptions = {
     injectedDependencies: {
       [ContainerRegistrationKeys.PG_CONNECTION]: connection,
+      eventBusService: new MockEventBusService(),
```
adrien2p, line 72: Suggestion: i suggest that we add dependencies as a config of the test runner and spread it here

packages/medusa-test-utils/src/mock-event-bus-service.ts
```diff
@@ -0,0 +1,43 @@
...
+
+  unsubscribe(
+    event: string | symbol,
+    subscriber: Subscriber,
+    context?: EventBusTypes.SubscriberContext
+  ): this {
+    return this
+  }
+
+  withTransaction() {}
```
adrien2p: q: is that needed?

packages/types/src/user/events.ts
```diff
@@ -0,0 +1,4 @@
+export const CREATED = "user.created"
+export const UPDATED = "user.updated"
+export const INVITE_CREATED = "invite.created"
+export const INVITE_UPDATED = "invite.updated"
```
adrien2p: todo: move to utils. Also, you can probably follow the current pattern we have created
```ts
const UserEvents = {
  created: 'user.' + CommonEvents.CREATED
}
```

packages/user/integration-tests/__tests__/services/module/invite.spec.ts
```diff
@@ -188,5 +214,26 @@ describe("UserModuleService - Invite", () => {
...
+
+    it("should emit invite created events", async () => {
+      const eventBusSpy = jest.spyOn(MockEventBusService.prototype, "emit")
+      await service.createInvites(defaultInviteData)
+
+      expect(eventBusSpy).toHaveBeenCalledTimes(1)
+      expect(eventBusSpy).toHaveBeenCalledWith([
+        expect.objectContaining({
+          body: expect.objectContaining({
+            data: "1",
```
adrien2p: suggestion: data should probably be an object which contains the id property no?

packages/user/integration-tests/__tests__/services/module/invite.spec.ts
```diff
@@ -171,6 +173,30 @@ describe("UserModuleService - Invite", () => {
...
+        },
+      ])
+
+      expect(eventBusSpy).toHaveBeenCalledTimes(1)
+      expect(eventBusSpy).toHaveBeenCalledWith([
+        expect.objectContaining({
+          body: expect.objectContaining({
+            data: "1",
+          }),
+          eventName: "invite.updated",
```
adrien2p: todo: use the const for the events everywhere to be future proof

packages/user/src/services/user-module.ts
```diff
@@ -96,6 +101,18 @@ export default class UserModuleService<
...
 
+    sharedContext.messageAggregator?.saveRawMessageData(
+      users.map((user) => ({
+        eventName: UserEvents.CREATED,
+        metadata: {
+          service: this.constructor.name,
+          action: "create",
+          object: "user",
+        },
+        data: user.id,
```
adrien2p: suggestion: `{ id: user.id }`

packages/user/src/services/user-module.ts
```diff
@@ -96,6 +101,18 @@ export default class UserModuleService<
       populate: true,
     })
 
+    sharedContext.messageAggregator?.saveRawMessageData(
+      users.map((user) => ({
```
adrien2p, line 106: suggestion: maybe we should use `composeMessage` so that we normalize events

packages/user/src/services/user-module.ts
```diff
@@ -96,6 +101,18 @@ export default class UserModuleService<
...
+        eventName: UserEvents.CREATED,
+        metadata: {
+          service: this.constructor.name,
+          action: "create",
```
adrien2p: todo: use CommonEvents,

packages/utils/src/event-bus/message-aggregator.ts
```diff
@@ -23,6 +26,15 @@ export class MessageAggregator implements IMessageAggregator {
     }
   }
 
+  saveRawMessageData<T>(
+    messageData:
+      | EventBusTypes.MessageFormat<T>
+      | EventBusTypes.MessageFormat<T>[],
+    options?: Record<string, unknown>
+  ): void {
+    this.save(buildEventMessages(messageData, options))
```
adrien2p, line 35: comment: maybe rename `composeMessage` to `composeMessages` and sligthly change the API, we already have discrepency between the two util

packages/user/src/services/user-module.ts
pKorsholm: I added `saveRawMessageData` to take the message input for `composeMessage` and call `eventAggregator.save` per. Oli's comment here: https://github.com/medusajs/medusa/pull/6431#discussion_r1494647141

I'm fine with just changing the `save` method as well

olivermrbl (approved):
Can we merge this @adrien2p? This has been hanging for too long

adrien2p:
@olivermrbl we can, there is just one thing i d like to discuss

PR #6432, opened 2024-02-19
feat(regions): Add support for updating countries in a region

sradevski:
**What**
Add missing support for updating countries on the POST /admin/regions/:id route.
Furthermore, the PR refactors how the normalization and validation was handled in both create and update scenarios.
Finally, the PR adds a unique index on the country entity, which ensures at the DB level that we cannot have a duplicate country in the DB for a single region.


integration-tests/plugins/__tests__/regions/admin/regions.spec.ts
```diff
@@ -63,14 +65,21 @@ describe("Regions - Admin", () => {
         id: created.data.region.id,
         name: "Test Region",
         currency_code: "usd",
+        metadata: { foo: "bar" },
       })
     )
+    expect(created.data.region.countries.map((c) => c.iso_2).sort()).toEqual([
```
adrien2p: comment: eventually you can also use arrayContaining to not have to sort. but I am fine with it

packages/region/src/services/region-module.ts
```diff
@@ -117,78 +115,31 @@ export default class RegionModuleService<
...
-      {},
-      sharedContext
-    )
-
-    const countriesInDb = await this.countryService_.list(
-      { iso_2: data.map((d) => d.countries).flat() },
-      { select: ["iso_2", "region_id"] },
-      sharedContext
-    )
+    let normalizedInput = this.normalizeInput(data)
```
adrien2p: nit: when a function doesnt rely on anything from the class context, I would say it could be a private static method on the class as an internal util.

packages/region/src/services/region-module.ts
```diff
@@ -223,37 +183,167 @@ export default class RegionModuleService<
...
+        ...region,
+        currency_code: region.currency_code?.toLowerCase(),
+        name: region.name?.trim(),
+        countries: region.countries?.map((country) => country.toLowerCase()),
       })
+    )
+  }
 
-      result = await this.regionService_.update(toUpdate, sharedContext)
+  @InjectTransactionManager("baseRepository_")
```
adrien2p: todo: you can probably rm the decorator here and rely on the shared context provided. Also, when using those decorator, you should apply @MedusaContext on the sharedContext argument

packages/region/src/services/region-module.ts
```diff
@@ -117,78 +115,31 @@ export default class RegionModuleService<
...
+      ),
+      this.validateCountries(
+        normalizedInput.map((r) => r.countries ?? []).flat(),
+        sharedContext
+      ),
+    ] as const
+
+    // Assign the full country object so the ORM updates the relationship
+    const [, dbCountries] = await Promise.all(validations)
+    const dbCountriesMap = asMap(dbCountries, "iso_2")
```
adrien2p: suggestion: we could create an inline map here wdyt? not sure the util is very useful
```ts
const dbCountriesMap = new Map(dbCountries.map(d => [d.id, d]))
```

packages/region/src/services/region-module.ts
```diff
@@ -208,11 +159,20 @@ export default class RegionModuleService<
...
+    const updateResult = await this.update_(
+      idOrSelectorOrData,
+      data,
+      sharedContext
+    )
+    const resultWithRelations = await this.regionService_.list(
+      { id: updateResult.map((r) => r.id) },
+      { relations: ["currency", "countries"] },
+      sharedContext
+    )
```
adrien2p: suggestion: wouldn't it be better to let the consumer decide on that? I mean the consumer could list or retrieve after updating the region. He will be the one knowing if he needs to retrive the region and if he needs the relations. that way it will remove some queries that could be unnecessary

packages/region/src/services/region-module.ts
```diff
@@ -223,37 +183,167 @@ export default class RegionModuleService<
...
+
+    // If countries are being updated for a region, first make previously set countries' region to null to get to a clean slate.
+    // Somewhat less efficient, but region operations will be very rare, so it is better to go with a simple solution
+    await this.countryService_.update(
+      {
+        selector: {
+          region_id: normalizedInput
+            .filter((region) => !!region.countries)
+            .map((region) => region.id)
+            .flat(),
```
adrien2p: todo: extract outside plus add an if, if the filter result in 0 items, then we will have an empty array here

packages/region/src/services/region-module.ts
```diff
@@ -223,37 +183,167 @@ export default class RegionModuleService<
...
+      throw new MedusaError(
+        MedusaError.Types.INVALID_DATA,
+        `Currencies with codes: "${missingCurrencies.join(
+          ", "
+        )}" were not found`
+      )
+    }
+  }
+
+  @InjectTransactionManager("baseRepository_")
```
adrien2p: todo: same as previous

packages/region/src/services/region-module.ts
```diff
@@ -223,37 +183,167 @@ export default class RegionModuleService<
...
+  }
 
-      result = await this.regionService_.update(toUpdate, sharedContext)
+  @InjectTransactionManager("baseRepository_")
+  private async validateCurrencies(
+    currencyCodes: (string | undefined)[] | undefined,
+    sharedContext: Context
+  ): Promise<void> {
+    const normalizedCurrencyCodes = currencyCodes
+      ?.filter((c) => c !== undefined)
```
adrien2p, line 264: todo: here null will go to the map step. 
```ts
currencyCodes
      ?.filter((c) => !!c)
      .map(c => c.toLowerCase()
```

packages/region/src/services/region-module.ts
```diff
@@ -223,37 +183,167 @@ export default class RegionModuleService<
...
+      .map((c) => c!.toLowerCase())
+
+    if (!normalizedCurrencyCodes || !normalizedCurrencyCodes.length) {
```
adrien2p:
```suggestion
    if (!normalizedCurrencyCodes?.length) {
```

packages/region/src/services/region-module.ts
```diff
@@ -223,37 +183,167 @@ export default class RegionModuleService<
...
+      )
+    }
+  }
+
+  @InjectTransactionManager("baseRepository_")
+  private async validateCountries(
+    countries: string[] | undefined,
+    sharedContext: Context
+  ): Promise<TCountry[]> {
+    if (!countries || !countries.length) {
```
adrien2p:
```suggestion
    if (!countries?.length) {
```

packages/region/src/services/region-module.ts
```diff
@@ -325,3 +415,35 @@ export default class RegionModuleService<
...
 }
+
+// microORM complains if undefined fields are present in the passed data object
+const removeUndefined = <T extends Record<string, any>>(obj: T): T => {
+  return Object.fromEntries(
+    Object.entries(obj).filter(([_, v]) => v !== undefined)
+  ) as T
+}
+
+const asMap = <T extends Record<string, any>>(
```
adrien2p: suggestion: rm not sure it is necessary as creating a Map object is equivalent and one liner

integration-tests/plugins/__tests__/regions/admin/regions.spec.ts
sradevski: @adrien2p as we talked on Oli's PR, I'd prefer if we have deterministic sorting as part of the internal service, so we don't even need to do the sorting.

packages/region/src/services/region-module.ts
sradevski: Yup agreed, will do it now.

packages/region/src/services/region-module.ts
sradevski: Yeah I agree, I actually didn't know new Map takes a tuple 👍

packages/region/src/services/region-module.ts
sradevski: @adrien2p well I think it really depends on the use-case. In this case, calling `Create` will always return the countries (somewhat by coincidence), and I think updates should map a similar behavior.

This is really up to us to define what should the heuristics be for each module, and follow them everywhere. I just went with this as it seemed more intuitive and simpler for consumers, but we can discuss it later.

packages/region/src/services/region-module.ts
sradevski: Good point, we can avoid a DB request 👍

packages/region/src/services/region-module.ts
sradevski: Ah good catch, I moved the checks around and didn't refactor 👍

packages/region/src/services/region-module.ts
sradevski: Well actually I was trying to handle the scenario where someone passes an empty string, for example. If they do, the validation below will catch it, but if we filter it out, it won't, so checking undefined only is better.

packages/region/src/services/region-module.ts
adrien2p: so far we have let the populate: true manage the behaviour and later we will look into it deeper but at least it is centralised. So I would say we could rm this part

packages/region/src/migrations/UpdateUniqueIndexCountry20240216151707.ts
```diff
@@ -0,0 +1,15 @@
+import { Migration } from '@mikro-orm/migrations';
```
olivermrbl: nit: Include this in the existing migration. As long as the modules are not released, we should add, modify or delete from a single migration file to not have redundant migrations when they are published

packages/region/src/services/region-module.ts
olivermrbl: And we do this because we want to throw when empty strings are passed, because that is, in fact, invalid data?

packages/region/src/services/region-module.ts
```diff
@@ -325,3 +420,25 @@ export default class RegionModuleService<
     }
   }
 }
+
+// microORM complains if undefined fields are present in the passed data object
+const removeUndefined = <T extends Record<string, any>>(obj: T): T => {
```
olivermrbl, line 418: question: Don't we have one in `@medusajs/utils`? If not, we should move it there for reusability purposes

packages/region/src/services/region-module.ts
```diff
@@ -325,3 +420,25 @@ export default class RegionModuleService<
...
+  return Object.fromEntries(
+    Object.entries(obj).filter(([_, v]) => v !== undefined)
+  ) as T
+}
+
+const getDuplicateEntries = (collection: string[]): string[] => {
```
olivermrbl, line 424: Same for this one

packages/region/src/services/region-module.ts
olivermrbl: For the sake of convention, I am also leaning toward using `populate: true`, but I don't find it blocking for the PR

packages/region/src/migrations/UpdateUniqueIndexCountry20240216151707.ts
sradevski: @olivermrbl got it, I assumed we actually don't want to do that, but it makes sense 👍

packages/region/src/services/region-module.ts
sradevski: Yeah I've taken a note to do it, but I wouldn't do it as part of this PR.

olivermrbl: Fine with me!

packages/region/src/services/region-module.ts
sradevski: Same, I'll do it in a follow-up PR

packages/region/src/services/region-module.ts
sradevski: Yes exactly @olivermrbl

integration-tests/plugins/__tests__/regions/admin/regions.spec.ts
adrien2p: todo: can you rm the sorting everywhere as it is by default on the relation. Normally the default sql sorting should be enough

packages/region/src/services/region-module.ts
```diff
@@ -223,37 +178,171 @@ export default class RegionModuleService<
...
+    normalizedInput = RegionModuleService.normalizeInput(normalizedInput)
+
+    // If countries are being updated for a region, first make previously set countries' region to null to get to a clean slate.
+    // Somewhat less efficient, but region operations will be very rare, so it is better to go with a simple solution
+    const regionsWithCountryUpdate = normalizedInput
+      .filter((region) => !!region.countries)
+      .map((region) => region.id)
+      .flat()
+
+    if (regionsWithCountryUpdate.length > 0) {
```
adrien2p: nit: checking > 0 is not necessary, 0 will be considered as false and otherwise true. But I am fine keeping it

packages/region/src/services/region-module.ts
```diff
@@ -223,37 +178,171 @@ export default class RegionModuleService<
...
+        sharedContext
+      ),
+      this.validateCountries(
+        normalizedInput.map((d) => d.countries ?? []).flat(),
+        sharedContext
+      ),
+    ] as const
+
+    // Assign the full country object so the ORM updates the relationship
+    const [, dbCountries] = await Promise.all(validations)
```
adrien2p: todo: use the `promiseAll` util instead of promise.all, it is not a problem here but just by convention. It happened that using promise.all was breaking the promise chain when one of the promise failed and therefore it was breaking the rollback chain

packages/region/src/services/region-module.ts
```diff
@@ -223,37 +178,171 @@ export default class RegionModuleService<
...
+        currency_code: region.currency_code?.toLowerCase(),
+        name: region.name?.trim(),
+        countries: region.countries?.map((country) => country.toLowerCase()),
+      })
+    )
+  }
+
+  private async validateCurrencies(
+    currencyCodes: (string | undefined)[] | undefined,
+    @MedusaContext() sharedContext: Context
```
adrien2p: todo: rm the decorator in the method where the Inject* is not used

packages/region/src/services/region-module.ts
```diff
@@ -325,3 +414,25 @@ export default class RegionModuleService<
     }
   }
 }
+
+// microORM complains if undefined fields are present in the passed data object
+const removeUndefined = <T extends Record<string, any>>(obj: T): T => {
```
adrien2p, line 418: maybe just put a todo here in order to not forget just in case 👍

adrien2p (approved):
LGTM 🔥 few comments

packages/region/src/services/region-module.ts
sradevski: @adrien2p I have a todo in my notion, no worries :) Once it's merged I'll open a PR to clean it up

packages/region/src/services/region-module.ts
sradevski: No I think it's fine without it, I'm just so used to typing it in golang 😄

packages/region/src/services/region-module.ts
adrien2p: 🚀

integration-tests/plugins/__tests__/regions/admin/regions.spec.ts
adrien2p: @sradevski could you remove the sort in all the files please

PR #6434, opened 2024-02-19
Introduced schema to the check of the migrations table

joaorb2:
What: database_schema included on the check for the migrations;
Why: because when running the application against a different schema from the public one, it searches for the migrations on the public schema and not on the new specified schema;
How: Including the namespace of the schema on the "Select * from migrations" query;
Testing: change your public schema to "test_schema", for example, connect to the new schema and reproduce the error.


riqwan:

PR #6435, opened 2024-02-19
Feat(order): order changes

carlos-r-l-rodrigues:
What:
 - Order DB schema migration
 - BigNumberField Decorator
  
order_schema

### DIFF
--- a/packages/admin-next/dashboard/src/lib/currencies.ts
+++ b/packages/admin-next/dashboard/src/lib/currencies.ts
@@ -298,7 +298,7 @@ export const currencies: Record<string, CurrencyInfo> = {
   INR: {
     code: "INR",
     name: "Indian Rupee",
-    symbol_native: "টকা",
+    symbol_native: "₹",
     decimal_digits: 2,
   },
   IQD: {
--- a/packages/admin-ui/ui/src/utils/currencies.ts
+++ b/packages/admin-ui/ui/src/utils/currencies.ts
@@ -448,7 +448,7 @@ export const currencies: CurrenciesType = {
   INR: {
     symbol: "Rs",
     name: "Indian Rupee",
-    symbol_native: "টকা",
+    symbol_native: "₹",
     decimal_digits: 2,
     rounding: 0,
     code: "INR",
--- a/packages/fulfillment/integration-tests/__tests__/fulfillment-module-service.spec.ts
+++ b/packages/fulfillment/integration-tests/__tests__/fulfillment-module-service.spec.ts
@@ -13,7 +13,7 @@ import {
   UpdateServiceZoneDTO,
 } from "@medusajs/types"
 import { GeoZoneType } from "@medusajs/utils"
-import { moduleIntegrationTestRunner, SuiteOptions } from "medusa-test-utils"
+import { SuiteOptions, moduleIntegrationTestRunner } from "medusa-test-utils"
 
 jest.setTimeout(100000)
 
@@ -1004,19 +1004,24 @@ moduleIntegrationTestRunner({
             )
 
             const updatedFulfillmentSets = await service.update(updateData)
+            const fullfillmentSets = await service.list({
+              id: updateData.map((ud) => ud.id),
+            })
 
             expect(updatedFulfillmentSets).toHaveLength(2)
 
-            let i = 0
             for (const data_ of updateData) {
-              expect(updatedFulfillmentSets[i]).toEqual(
+              const currentFullfillmentSet = fullfillmentSets.find(
+                (fs) => fs.id === data_.id
+              )
+
+              expect(currentFullfillmentSet).toEqual(
                 expect.objectContaining({
-                  id: createdFulfillmentSets[i].id,
+                  id: data_.id,
                   name: data_.name,
                   type: data_.type,
                 })
               )
-              ++i
             }
           })
 
--- a/packages/medusa/src/utils/currencies.ts
+++ b/packages/medusa/src/utils/currencies.ts
@@ -444,7 +444,7 @@ export const currencies: Record<string, Currency> = {
   INR: {
     symbol: "Rs",
     name: "Indian Rupee",
-    symbol_native: "টকা",
+    symbol_native: "₹",
     decimal_digits: 2,
     rounding: 0,
     code: "INR",
--- a/packages/tax/integration-tests/__tests__/index.ts
+++ b/packages/tax/integration-tests/__tests__/index.ts
@@ -92,4 +92,50 @@ describe("TaxModuleService", function () {
       ])
     )
   })
+
+  it("should create a tax rate rule", async () => {
+    const [region] = await service.createTaxRegions([
+      {
+        country_code: "US",
+        default_tax_rate: {
+          name: "Test Rate",
+          rate: 0.2,
+        },
+      },
+    ])
+
+    const rate = await service.create({
+      tax_region_id: region.id,
+      name: "Shipping Rate",
+      rate: 8.23,
+    })
+
+    await service.createTaxRateRules([
+      {
+        tax_rate_id: rate.id,
+        reference: "product",
+        reference_id: "prod_1234",
+      },
+    ])
+
+    const listedRules = await service.listTaxRateRules(
+      {},
+      {
+        relations: ["tax_rate"],
+      }
+    )
+    expect(listedRules).toEqual(
+      expect.arrayContaining([
+        expect.objectContaining({
+          reference: "product",
+          reference_id: "prod_1234",
+          tax_rate: expect.objectContaining({
+            tax_region_id: region.id,
+            name: "Shipping Rate",
+            rate: 8.23,
+          }),
+        }),
+      ])
+    )
+  })
 })
--- a/packages/tax/src/models/index.ts
+++ b/packages/tax/src/models/index.ts
@@ -1,2 +1,3 @@
 export { default as TaxRate } from "./tax-rate"
 export { default as TaxRegion } from "./tax-region"
+export { default as TaxRateRule } from "./tax-rate-rule"
--- a/packages/tax/src/models/tax-rate-rule.ts
+++ b/packages/tax/src/models/tax-rate-rule.ts
@@ -1,0 +1,55 @@
+import {
+  Cascade,
+  Entity,
+  ManyToOne,
+  PrimaryKey,
+  PrimaryKeyProp,
+  Property,
+} from "@mikro-orm/core"
+import TaxRate from "./tax-rate"
+
+const TABLE_NAME = "tax_rate_rule"
+
+const taxRateIdIndexName = "IDX_tax_rate_rule_tax_rate_id"
+
+@Entity({ tableName: TABLE_NAME })
+export default class TaxRateRule {
+  @PrimaryKey({ columnType: "text" })
+  tax_rate_id!: string
+
+  @PrimaryKey({ columnType: "text" })
+  reference_id!: string;
+
+  [PrimaryKeyProp]?: ["tax_rate_id", "reference_id"]
+
+  @Property({ columnType: "text" })
+  reference: string
+
+  @ManyToOne(() => TaxRate, {
+    fieldName: "tax_rate_id",
+    index: taxRateIdIndexName,
+    cascade: [Cascade.REMOVE, Cascade.PERSIST],
+  })
+  tax_rate: TaxRate
+
+  @Property({ columnType: "jsonb", nullable: true })
+  metadata: Record<string, unknown> | null = null
+
+  @Property({
+    onCreate: () => new Date(),
+    columnType: "timestamptz",
+    defaultRaw: "now()",
+  })
+  created_at: Date
+
+  @Property({
+    onCreate: () => new Date(),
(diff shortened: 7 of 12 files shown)


packages/order/src/migrations/Migration20240219102530.ts
```diff
@@ -0,0 +1,462 @@
+import { generatePostgresAlterColummnIfExistStatement } from "@medusajs/utils"
+import { Migration } from "@mikro-orm/migrations"
+
+export class Migration20240219102530 extends Migration {
```
sradevski, line 4: Have there been any changes to the model, or is it pretty much the same with v1?

packages/order/src/migrations/Migration20240219102530.ts
```diff
@@ -0,0 +1,462 @@
...
+          customer_id
+      )
+      WHERE deleted_at IS NOT NULL;
+
+      CREATE INDEX IF NOT EXISTS "IDX_order_currency_code" ON "order" (
+          currency_code
+      )
+      WHERE deleted_at IS NOT NULL;
+
+      CREATE INDEX IF NOT EXISTS "IDX_order_shipping_address_id" ON "order" (
```
sradevski, line 130: Do we need all these indices? For example, I think, for example, shipping address ID is not something you will search orders by.

packages/order/src/migrations/Migration20240219102530.ts
```diff
@@ -0,0 +1,462 @@
...
+          "updated_at" TIMESTAMPTZ NOT NULL DEFAULT Now(),
+          CONSTRAINT "order_address_pkey" PRIMARY KEY ("id")
+      );
+
+      CREATE INDEX IF NOT EXISTS "IDX_order_address_customer_id" ON "order_address" (
+          customer_id
+      );
+
+      CREATE TABLE IF NOT EXISTS "order" (
+          "id" TEXT NOT NULL,
```
sradevski, line 31: What are some constraints that we can/want to enforce at the DB level on an order?

packages/order/src/migrations/Migration20240219102530.ts
```diff
@@ -0,0 +1,462 @@
...
+          "raw_return_dismissed_quantity" JSONB NOT NULL,
+          "written_off_quantity" NUMERIC NOT NULL,
+          "raw_written_off_quantity" JSONB NOT NULL,
+          "summary" JSONB NOT NULL,
+          "created_at" TIMESTAMPTZ NOT NULL DEFAULT Now(),
+          "updated_at" TIMESTAMPTZ NOT NULL DEFAULT Now(),
+          CONSTRAINT "order_detail_pkey" PRIMARY KEY ("id")
+      );
+
+      CREATE INDEX IF NOT EXISTS "IDX_order_detail_id" ON "order_detail" (id);
```
sradevski: I don't think this is needed, since Postgres will create an index for the primary key.

packages/order/src/migrations/Migration20240219102530.ts
```diff
@@ -0,0 +1,462 @@
...
+          "compare_at_unit_price" NUMERIC NULL,
+          "raw_compare_at_unit_price" JSONB NULL,
+          "unit_price" NUMERIC NOT NULL,
+          "raw_unit_price" JSONB NOT NULL,
+          "created_at" TIMESTAMPTZ NOT NULL DEFAULT Now(),
+          "updated_at" TIMESTAMPTZ NOT NULL DEFAULT Now(),
+          CONSTRAINT "order_line_item_pkey" PRIMARY KEY ("id")
+      );
+
+      CREATE INDEX IF NOT EXISTS "IDX_order_line_item_variant_id" ON "order_line_item" (
```
sradevski, line 263: Do we ever fetch orders only by variant id? Or is it always product_id/variant_id combo? If it's the later, we can make this a composite index instead.

packages/order/src/migrations/Migration20240219102530.ts
```diff
@@ -0,0 +1,462 @@
...
+          "promotion_id" TEXT NULL,
+          "code" TEXT NULL,
+          "amount" NUMERIC NOT NULL,
+          "raw_amount" JSONB NOT NULL,
+          "provider_id" TEXT NULL,
+          "created_at" TIMESTAMPTZ NOT NULL DEFAULT Now(),
+          "updated_at" TIMESTAMPTZ NOT NULL DEFAULT Now(),
+          "item_id" TEXT NULL,
+          CONSTRAINT "order_line_item_adjustment_pkey" PRIMARY KEY ("id"),
+          CONSTRAINT order_line_item_adjustment_check CHECK (amount >= 0)
```
sradevski, line 299: Can't an adjustment be negative?

packages/order/src/migrations/Migration20240219102530.ts
```diff
@@ -0,0 +1,462 @@
...
+          "created_at" TIMESTAMPTZ NOT NULL DEFAULT Now(),
+          "updated_at" TIMESTAMPTZ NOT NULL DEFAULT Now(),
+          CONSTRAINT "order_transaction_pkey" PRIMARY KEY ("id")
+      );
+
+      CREATE INDEX IF NOT EXISTS "IDX_order_transaction_order_id" ON "order_transaction" (
+          order_id
+      );
+
+      CREATE INDEX IF NOT EXISTS "IDX_order_transaction_currency_code" ON "order_transaction" (
```
sradevski, line 382: Here as well, do we need this many indices?

packages/order/src/models/adjustment-line.ts
```diff
@@ -27,6 +27,7 @@ export default abstract class AdjustmentLine {
   code: string | null = null
 
   @Property({ columnType: "numeric" })
+  @BigNumberField()
```
sradevski, line 30: Do we need the ambiguity in typings for amount?

packages/utils/src/order/status.ts
```diff
@@ -29,3 +29,31 @@ export enum OrderStatus {
    */
   REQUIRES_ACTION = "requires_action",
 }
+
+/**
+ * @enum
+ *
+ * The order change's status.
+ */
+export enum OrderChangeStatus {
```
sradevski, line 38: I am just wondering, is utils the best place for things like enums? Wouldn't this be better placed as part of a module (and exported from it in some way)? Alternatively, it can be part of the API definition, but like this It feels like we are leaking domain models to the public

packages/order/src/models/transaction.ts
```diff
@@ -0,0 +1,103 @@
...
+    nullable: true,
+  })
+  reference: string | null = null
+
+  @Property({
+    columnType: "text",
+    nullable: true,
+  })
+  @ReferenceIdIndex.MikroORMIndex()
+  reference_id: string | null = null
```
sradevski, line 77: What does this column represent? If it is the ID in the external system, `external_id` is typically a better name, as `reference` typically means something else and `reference_id` sounds like the two are related.

sradevski (commented):
I am still just exploring, so some questions might sounds a bit uninformed (and they are) 😄

packages/order/src/migrations/Migration20240219102530.ts
carlos-r-l-rodrigues: There are changes to the current model. the `order` entity is using less fields and adding a `deleted_at` on it. The other tables are either the same or new so they won't conflict.

packages/order/src/migrations/Migration20240219102530.ts
carlos-r-l-rodrigues: Not normally, but we are defining all the graph in our joiner configs + link modules. So, from a shipping address, one can expand orders.
If in the future any index size become a concern, we can review that, for now traversing the entities from any entrypoint is a nice to have feature.

packages/order/src/migrations/Migration20240219102530.ts
carlos-r-l-rodrigues: I wouldn't add any at the DB level. The ones from LineItemAdjustment and ShippingMethod are inherited from Cart. But very likely I will drop them soon.

packages/order/src/migrations/Migration20240219102530.ts
carlos-r-l-rodrigues: Indeed. Before the PK was a composite key, and the ID a regular index.. I inverted it but didn't remove this index.

packages/order/src/migrations/Migration20240219102530.ts
carlos-r-l-rodrigues: These ids fall in the same category as things that can be expanded by other services via RemoteJoiner.

packages/order/src/migrations/Migration20240219102530.ts
carlos-r-l-rodrigues: No in this context. But as I mentioned, I prefer to keep business logic in the app layer. That will probably be removed

packages/order/src/models/adjustment-line.ts
carlos-r-l-rodrigues: Not sure I understand.
While BigNumber is not supported everywhere I believe it makes sense to keep it as `BigNumber | number`.

packages/order/src/models/transaction.ts
carlos-r-l-rodrigues: It is the ID of whatever `reference` is.
I'm ok with renaming it, but I would like to also rename `reference`. any suggestion?

packages/utils/src/order/status.ts
carlos-r-l-rodrigues: These are public exposed properties of the module, so I do not consider this a "leak".
We don't have a clean solution to place enums, that behave like a type, but can't be added to the types package.
One shouldn't need the module to have access to their public types.

packages/order/src/migrations/Migration20240219102530.ts
sradevski: Got it 👍

packages/order/src/migrations/Migration20240219102530.ts
sradevski: OK cool, so I guess we can remove these checks

packages/order/src/models/adjustment-line.ts
sradevski: I was thinking `number` might not be necessary, but I guess it all depends on how the module is used.

packages/utils/src/order/status.ts
sradevski: I think discussing this here will require several paragraphs so we can talk about the different approaches in defining module definitions some other time. But either way, putting enums in utils is a bit dirty in my opinion. The types package could be a better place where we expose enums, WDYT?

packages/order/src/migrations/Migration20240219102530.ts
sradevski: @carlos-r-l-rodrigues I think there are some that are better placed on the db, namely uniqueness. Unless you run all transactions with snapshot isolation there is always a chance that you'll have a race condition that will mess up your data :)

packages/order/src/migrations/Migration20240219102530.ts
sradevski: @carlos-r-l-rodrigues one more question 😄 is it required that we have an index when doing the linking, or we do it for performance? If it is purely from a performance POV, I think if you have an index on a column that will always have only 2-3 values (eg. shipping method), Postgres might decide to not use the index so you end up with wasted storage (and it impacts write performance). Just a thought 😄

packages/order/src/models/transaction.ts
sradevski: We discussed with Carlos, this is not a payment reference, but a reference to the originating module/source.

A name we discussed was `originates_from` and `origin_id` as an alternative.

olivermrbl: I believe `reference` + `reference_id` is a pattern that we've used elsewhere for a similar purpose, so for consistency, I think we should keep it as is

packages/utils/src/order/status.ts
carlos-r-l-rodrigues: `@medusajs/types` should be strictly a dev dependency, as enums will generate an object, they can't be referenced at the code. They won't exist in production.
That's the main reason it is not placed there.

packages/order/src/models/transaction.ts
sradevski: @olivermrbl yeah that's what we discussed with carlos, we were just talking if we were to rename it everywhere, what would a better name be.

packages/order/src/migrations/Migration20240219102530.ts
carlos-r-l-rodrigues: I agree. but it's rare the case where we have a unique check that is just a constraint and not an index.

packages/utils/src/order/status.ts
sradevski: Got it, thanks @carlos-r-l-rodrigues . I guess we can look for a better place later on. My previous experience was that the sdk would carry types, enums, and wrappers around the backend API, but I think here we want them to be more independent.

srindom (approved):
LGTM!

PR #6436, opened 2024-02-19
chore(utils): Add default ordering to the internal service factory list/listAndCount

adrien2p:
**What**
Default ordering. By default, only the top level entity ordering is applied using the primary keys, the relations are default ordered by the foreign keys.

It include tests fixing for deterministic data ordering

### DIFF
--- a/packages/pricing/integration-tests/__tests__/services/pricing-module/price-set-money-amount-rules.spec.ts
+++ b/packages/pricing/integration-tests/__tests__/services/pricing-module/price-set-money-amount-rules.spec.ts
@@ -273,16 +273,14 @@ describe("PricingModule Service - PriceSetMoneyAmountRules", () => {
         },
       ])
 
-      const priceSetMoneyAmountRules =
-        await service.listPriceSetMoneyAmountRules(
-          {},
-          {
-            relations: ["price_set_money_amount", "rule_type"],
-          }
-        )
-
-      const created =
-        priceSetMoneyAmountRules[priceSetMoneyAmountRules.length - 1]
+      const [created] = await service.listPriceSetMoneyAmountRules(
+        {
+          value: ["New priceSetMoneyAmountRule"],
+        },
+        {
+          relations: ["price_set_money_amount", "rule_type"],
+        }
+      )
 
       expect(created).toEqual(
         expect.objectContaining({


packages/utils/src/modules-sdk/internal-module-service-factory.ts
```diff
@@ -63,6 +63,48 @@ export function internalModuleServiceFactory<
       return keys.map((k) => data[k]).join("_")
     }
 
+    static applyDefaultOrdering(config: FindConfig<any>) {
+      if (config.order) {
+        return
+      }
+
+      const order: object[] = []
```
sradevski: We can make the type stricter, where the key is `keyof model`, and the value is "ASC" or "DESC"

packages/utils/src/modules-sdk/internal-module-service-factory.ts
```diff
@@ -63,6 +63,48 @@ export function internalModuleServiceFactory<
...
+      const primaryKeys = AbstractService_.retrievePrimaryKeys(model)
+      const mainEntityOrder = {}
+      primaryKeys.forEach((pk) => {
+        mainEntityOrder[pk] = "ASC"
+      })
+      order.push(mainEntityOrder)
+
+      if (config.relations?.length) {
+        config.relations.forEach((relation) => {
+          const relation_ = (
```
sradevski: Rather than using "private" models, an easier to read approach would be to be more concrete about the variables. For example, the `relation` in the array would be a `relationKey` or `relationName`, and not a relation object. Then you can just name this `relation` without any underscores.

packages/utils/src/modules-sdk/internal-module-service-factory.ts
```diff
@@ -63,6 +63,48 @@ export function internalModuleServiceFactory<
...
+            model.__meta ?? model.prototype.__meta
+          ).relations.find((r) => r.name === relation)
+          const relationPrimaryKeys = relation_.targetMeta.primaryKeys ?? ["id"]
+
+          // create order deep object based on relation segment
+          const relationOrder = {}
+          relation.split(".").reduce((acc, segment, index, segments) => {
```
sradevski: This part is really tough to understand. I ran through it with an example and I feel like I'm missing something :D

sradevski (approved):

carlos-r-l-rodrigues (approved):
LGTM

PR #6439, opened 2024-02-19
fix #6438

### LINKED ISSUE #6438: Capture method does not automatically mark payment as captured.
## Bug report
Payment Plugin support `capture` a boolean value that indicates whether payment should be captured manually or automatically.
This option is working perfectly in plugin (e.g PayPal, Stripe etc) but problem is new order is always created with payment status `awaiting` it should not be awaiting state when it is already captured this problem is due to hard coded status in newly created order as you see below.
https://github.com/medusajs/medusa/blob/1d91b7429beebd6f09d5027f7f7e1fe74ce3a8ff/packages/medusa/src/services/order.ts#L685-L688

### System information

Medusa version (including plugins): v1.20.2
Node.js version: v21.6.1
Database: postgres
Operating system: macOS
Browser (if relevant):

1. install medusa2. 
2. set `capture` true in medusa-config for any payment provider
3. place order and you will see order is always created with `awaiting` payment status

AxeemHaider:


SapioBeasley:
Is there a plan to have this merged? I myself have had issues with auto capture of payment on the medusa side and this this would solve my issue. If there has been another solution or insight into why orders are not auto capturing please point me in the correct direction

carlos-r-l-rodrigues (dismissed):

carlos-r-l-rodrigues:
Could you please rebase your PR to `v1.x`? 
We also have to execute the method `capture` of the payment service to create the entity and trigger the event.
cc: @olivermrbl @adrien2p

riqwan:

lukegreaves5:
Could we reopen this for v2? The same issue is happening in v2

PR #6440, opened 2024-02-19
feat(tax): adds getItemTaxLines

srindom:
**What**
- Selects the correct tax line for an item given a calculation context.

**For later PR**
- Consider optimizations. Some thoughts:
  - Even with global sales the number of rates in the DB is not likely to grow beyond ~1000.
  - Can large orders with hundreds of items optimize somehow?
  - Does it make sense to write a custom SQL query to do this?
- Support combined rate.

**Test cases covered**
The selection of tax rates take the following priority:

1. specific product rules - province
2. specific product type rules - province
3. default province rules
4. specific product rules - country
5. specific product type rules - country
6. default country rules

There are test cases for each of them under the following data seed structure:

### **US** 
- **Default Rate**: 2%
- **Sub-Regions**
  - CA 
    - Default Rate: 5%
    - Overrides
      - Reduced rate (for 3 product ids): 3%
      - Reduced rate (for product type): 1%
  - NY 
    - Default rate: 6%
  - FL 
    - Default rate: 4%
- **Overrides**
  - None

### **Denmark** 
- **Default rate:** 25% 
- **Sub-Regions**
  - None
- **Overrides**
  -  None

### **Germany** 
- **Default Rate:** 19%
- **Sub-Regions**
  - None
- **Overrides:**
  - Reduced Rate (for product type) - 7%

### **Canada** 
- **Default rate**: 5%
- **Sub-Regions**
  - QC 
    - Default rate: 2%
    - Overrides:
      - Reduced rate (for same product type as country reduced rate): 1%
  - BC 
    - Default rate: 2%
- **Overrides**
  - Reduced rate (for product id) - 3%
  - Reduced rate (for product type) - 3.5%


packages/tax/src/services/tax-module-service.ts
```diff
@@ -166,4 +167,145 @@ export default class TaxModuleService<
...
+            country_code: calculationContext.address.country_code,
+            province_code: calculationContext.address.province_code,
+          },
+        ],
+      },
+      {},
+      sharedContext
+    )
+
+    const toReturn = await promiseAll(
```
carlos-r-l-rodrigues, line 194: Not a blocker, but ideally we should query the DB just once here.
First aggregating all the items to compose a single query to call `this.taxRateService_.list(`, then we map the items with the their priority rates.

packages/tax/src/services/tax-module-service.ts
```diff
@@ -166,4 +167,145 @@ export default class TaxModuleService<
...
   }
+
+  async getTaxLines(
+    items: (TaxTypes.TaxableItemDTO | TaxTypes.TaxableShippingDTO)[],
+    calculationContext: TaxTypes.TaxCalculationContext,
+    @MedusaContext() sharedContext: Context = {}
+  ): Promise<(TaxTypes.ItemTaxLineDTO | TaxTypes.ShippingTaxLineDTO)[]> {
+    const regions = await this.taxRegionService_.list(
+      {
+        $or: [
```
sradevski, line 179: Wouldn't 
```
{
  country_code: calculationContext.address.country_code
}
```

have the same result? Or is the country optional, so we would like to only fetch by province if it's not provided?

packages/tax/src/services/tax-module-service.ts
```diff
@@ -166,4 +167,145 @@ export default class TaxModuleService<
...
+        let ruleQuery = isShipping
+          ? [
+              {
+                reference: "shipping_option",
+                reference_id: item.shipping_option_id,
+              },
+            ]
+          : [
+              {
+                reference: "product",
```
sradevski: Are we sure that product_id and product_type_id will exist? Might be good to do some validation beforehand.

packages/tax/src/services/tax-module-service.ts
```diff
@@ -166,4 +167,145 @@ export default class TaxModuleService<
...
+        ],
+      },
+      {},
+      sharedContext
+    )
+
+    const toReturn = await promiseAll(
+      items.map(async (item) => {
+        const isShipping = "shipping_option_id" in item
+        let ruleQuery = isShipping
```
sradevski: We have quite a bit of nesting here, I'd suggest we break this method down with a couple of helper methods (eg. `getRulesForItem` or `getRatesForItem`, and the rest can stay in-line)

packages/tax/src/services/tax-module-service.ts
```diff
@@ -166,4 +167,145 @@ export default class TaxModuleService<
...
+          rate: rate.rate,
+          code: rate.code,
+          name: rate.name,
+        }
+
+        if (isShipping) {
+          return [{ shipping_line_id: item.id, ...toReturn }]
+        }
+
+        return [{ line_item_id: item.id, ...toReturn }]
```
sradevski: Why do we return an array, just to flatten it afterwards?

packages/tax/src/services/tax-module-service.ts
```diff
@@ -166,4 +167,145 @@ export default class TaxModuleService<
...
+              rule.reference_id === item.product_type_id)
+          )
+        })
+
+        if (matchingRules.length !== 0) {
+          if (matchingRules.some((rule) => rule.reference === "product")) {
+            isProductMatch = true
+          } else {
+            isProductTypeMatch = true
+          }
```
riqwan: could also be shipping here:

```suggestion
          } else if (matchingRules.some((rule) => rule.reference === "product_type")) {
            isProductTypeMatch = true
          }
```

packages/tax/src/services/tax-module-service.ts
sradevski: Actually looking at the code below as well, a better approach might be to simply do:
```
if (isShipping)
 return getShippingRate

return productRate
```

And then the two calculation are independent. It will result in a bit of code duplication, but you will have far less conditionals, so it will be much easier to read and follow.

packages/tax/src/services/tax-module-service.ts
```diff
@@ -166,4 +167,145 @@ export default class TaxModuleService<
...
+    item: TaxTypes.TaxableItemDTO | TaxTypes.TaxableShippingDTO
+  ) {
+    const isShipping = "shipping_option_id" in item
+
+    const decoratedRates: (TTaxRate & {
+      priority_score: number
+    })[] = rates.map((rate) => {
+      let isProductMatch = false
+      let isProductTypeMatch = false
+      if (rate.rules.length !== 0) {
```
sradevski: Doing an early return reduces the nesting, eg.
```
if (!rate.rules.length)
 return
...
```

packages/tax/src/services/tax-module-service.ts
```diff
@@ -166,4 +167,145 @@ export default class TaxModuleService<
...
+        }
+
+        return [{ line_item_id: item.id, ...toReturn }]
+      })
+    )
+
+    return toReturn.flat()
+  }
+
+  private prioritizeRates(
```
carlos-r-l-rodrigues: Is it possible to have something a bit more generic here?
Something a bit like this:

```typescript
const taxProperties = {
  province_code: { weight: 50 },
  is_default: { weight: 10 },
}

const itemProperties = {
  product_id: { weight: 100 },
  product_type_id: { weight: 50 },
  shipping_option_id: { weight: 100 },
  // any other item property
}

function prioritizeRates(rates, item) {
  const calculateWeight = (rate) => {
    let weight = rate.is_default ? itemProperties.is_default.weight : 0

    for (const property in taxProperties) {
      if (rate.tax_region[property] === null) {
        continue
      }

      const ruleConfig = taxProperties[property]
      weight += ruleConfig.weight
    }

    for (const rule of rate.rules) {
      const itemProperty = `${rule.reference}_id`
      if (!(itemProperty in itemProperties) || !(itemProperty in item)) {
        continue
      }

      if (rule.reference_id === item[itemProperty]) {
        const ruleConfig = itemProperties[itemProperty]
        weight += ruleConfig.weight
      }
    }

    return weight
  }

  return rates
    .map((rate) => ({ ...rate, weight: calculateWeight(rate) }))
    .sort((a, b) => b.weight - a.weight) // weight DESC
}
```
forgive any mistake here, it was a quick implementation of an idea. If not applicable, just disregard it.

packages/tax/src/services/tax-module-service.ts
```diff
@@ -166,4 +167,145 @@ export default class TaxModuleService<
...
+              { $or: [{ is_default: true }, { rules: { $or: ruleQuery } }] },
+            ],
+          },
+          { relations: ["rules", "tax_region"] },
+          sharedContext
+        )
+
+        const prioritizedRates = this.prioritizeRates(rates, item)
+
+        const rate = prioritizedRates[0]
```
riqwan: suggestion: to add a validation here if rate is not found.

packages/tax/src/services/tax-module-service.ts
```diff
@@ -166,4 +167,145 @@ export default class TaxModuleService<
...
+        const toReturn = {
+          rate_id: rate.id,
+          rate: rate.rate,
+          code: rate.code,
+          name: rate.name,
+        }
+
+        if (isShipping) {
+          return [{ shipping_line_id: item.id, ...toReturn }]
```
riqwan: q: can this have multiple returns?

packages/tax/src/services/tax-module-service.ts
```diff
@@ -166,4 +167,145 @@ export default class TaxModuleService<
   ) {
     return await this.taxRateRuleService_.create(data, sharedContext)
   }
+
+  async getTaxLines(
```
adrien2p, line 172: todo: InjectManager

packages/tax/src/services/tax-module-service.ts
adrien2p, line 179: curiours to see if this would work instead
```ts
{
  country_code: calculationContext.address.country_code,
  province_code: [null, calculationContext.address.province_code]
}
```

packages/tax/src/services/tax-module-service.ts
```diff
@@ -166,4 +167,145 @@ export default class TaxModuleService<
...
+      }
+      if (isDefault && !isProvince) {
+        return { ...rate, priority_score: 6 }
+      }
+
+      return { ...rate, priority_score: 7 }
+    })
+
+    return decoratedRates.sort(
+      (a, b) => (a as any).priority_score - (b as any).priority_score
```
riqwan, line 376: suggestion: add a local type and reuse here to remove any. 

```
(TTaxRate & {
      priority_score: number
    })
```

riqwan (approved):
Readability takes a bit of a hit here with the multi-nests. LGTM otherwise with some comments.

packages/tax/src/services/tax-module-service.ts
```diff
@@ -166,4 +167,145 @@ export default class TaxModuleService<
...
+    const isShipping = "shipping_option_id" in item
+
+    const decoratedRates: (TTaxRate & {
+      priority_score: number
+    })[] = rates.map((rate) => {
+      let isProductMatch = false
+      let isProductTypeMatch = false
+      if (rate.rules.length !== 0) {
+        const matchingRules = rate.rules.filter((rule) => {
+          if (isShipping) {
```
adrien2p: there is quite a few early return that can be made here

packages/tax/src/services/tax-module-service.ts
```diff
@@ -166,4 +167,145 @@ export default class TaxModuleService<
...
+        return { ...rate, priority_score: 4 }
+      }
+      if (isProductTypeMatch && !isProvince) {
+        return { ...rate, priority_score: 5 }
+      }
+      if (isDefault && !isProvince) {
+        return { ...rate, priority_score: 6 }
+      }
+
+      return { ...rate, priority_score: 7 }
```
adrien2p:
```suggestion
const result = { ...rate, priority_score: 7 }

if ((isShipping || isProductMatch) && isProvince) {
  result.priority_score = 1
}
if (isProductTypeMatch && isProvince) {
  result.priority_score = 2
}
if (isDefault && isProvince) {
  result.priority_score = 3
}
if (isProductMatch && !isProvince) {
  result.priority_score = 4
}
if (isProductTypeMatch && !isProvince) {
  result.priority_score = 5
}
if (isDefault && !isProvince) {
  result.priority_score = 6
}

return result
```

adrien2p (commented):
to complete other comments

packages/tax/src/services/tax-module-service.ts
srindom: The idea is that we as a maximum only get the country region and the province region. That maps to:
1. A region where the country code is set and the province code is not set.
2. A region where both the country code and the province code is set.

If we just listed all regions with the country code we would get a all province regions too. I.e. NY in US should only give:
- US
- US, NY

Your suggestion would give:
- US
- US, NY
- US, CA
- etc. etc.

srindom: This doesn't work - but don't understand why? It seems like it should work

adrien2p: yeah I would have expected that to work too, nevermind we can see that later

packages/tax/src/services/tax-module-service.ts
srindom: Some rates may be combinable - this is coming in a future PR :)

packages/tax/src/services/tax-module-service.ts
srindom: Love this idea!

srindom: After playing around with it I think I will stick with what's there for now. Reason being:
1. It's tricky to figure out what the exact priority of combinations are with this.
2. Right now we pull out all rules for an override rate (which might be a lot), this should change to only pull out the rates relevant to the item. This will simplify the current model more and maybe make this approach less relevant.

packages/tax/src/services/tax-module-service.ts
srindom: I would be really keen to have just one query to the DB that does all this stuff in one shot.

packages/tax/src/services/tax-module-service.ts
srindom: We know that the reference_id will exist and we know that the item will exist so we may have:
```
"somereferenceid" === undefined
```
but that should be alright. right?

packages/tax/src/services/tax-module-service.ts
sradevski: Yes you are right, understood what the intention is 👍

carlos-r-l-rodrigues (approved):

PR #6441, opened 2024-02-19
feat(medusa-react,medusa,utils): add users/me endpoint + add missing specs

riqwan:
**what:**

- adds /me endpoint
- adds fixes to routes
- adds specs for auth endpoint
- updates dotenv package versions

### DIFF
--- a/integration-tests/modules/__tests__/cart/store/cart.workflows.spec.ts
+++ b/integration-tests/modules/__tests__/cart/store/cart.workflows.spec.ts
@@ -1,9 +1,11 @@
 import {
   addToCartWorkflow,
   createCartWorkflow,
+  createPaymentCollectionForCartWorkflow,
   deleteLineItemsStepId,
   deleteLineItemsWorkflow,
   findOrCreateCustomerStepId,
+  linkCartAndPaymentCollectionsStepId,
   updateLineItemInCartWorkflow,
   updateLineItemsStepId,
 } from "@medusajs/core-flows"
@@ -11,6 +13,7 @@ import { ModuleRegistrationName } from "@medusajs/modules-sdk"
 import {
   ICartModuleService,
   ICustomerModuleService,
+  IPaymentModuleService,
   IPricingModuleService,
   IProductModuleService,
   IRegionModuleService,
@@ -37,7 +40,8 @@ describe("Carts workflows", () => {
   let customerModule: ICustomerModuleService
   let productModule: IProductModuleService
   let pricingModule: IPricingModuleService
-  let remoteLink
+  let paymentModule: IPaymentModuleService
+  let remoteLink, remoteQuery
 
   let defaultRegion
 
@@ -52,7 +56,9 @@ describe("Carts workflows", () => {
     customerModule = appContainer.resolve(ModuleRegistrationName.CUSTOMER)
     productModule = appContainer.resolve(ModuleRegistrationName.PRODUCT)
     pricingModule = appContainer.resolve(ModuleRegistrationName.PRICING)
+    paymentModule = appContainer.resolve(ModuleRegistrationName.PAYMENT)
     remoteLink = appContainer.resolve("remoteLink")
+    remoteQuery = appContainer.resolve("remoteQuery")
   })
 
   afterAll(async () => {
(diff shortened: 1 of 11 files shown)


olivermrbl:
@riqwan, I guess there are some of these changes, we can still get in even though we are parking the promotions UI for now right?

riqwan:
Yes, if we remove the admin-next changes, this should be mergeable. I can do a cleanup and reopen later today.

olivermrbl:
Sg!

olivermrbl:
> Yes, if we remove the admin-next changes, this should be mergeable. I can do a cleanup and reopen later today.

@riqwan, can I get you to attend to this one, so it doesn't hang for too long?

srindom:
Should we maybe remove the changes in medusa-react and medusa-js for now too? 

Also think it would be good to get @sradevski and @pKorsholm to review this since it's auth related.

packages/medusa/src/utils/authenticate-middleware.ts
```diff
@@ -67,12 +67,27 @@ export const authenticate = (
       }
     }
 
-    if (authUser) {
+    const isMedusaScope =
```
sradevski: I think after you rebase you'll see that this code was added already by Philip 😄

packages/medusa/src/loaders/helpers/routing/index.ts
```diff
@@ -601,6 +606,21 @@ export class RoutesLoader {
         )
       }
 
+      if (descriptor.config.shouldAppendAuthCors) {
```
sradevski, line 620: Can you explain the logic behind having a different CORS setting for auth? Is this for the cases where the auth module can be hosted on a separate domain?

packages/medusa/src/api/middlewares/authenticate.ts
```diff
@@ -1,12 +1,19 @@
...
 export default (): RequestHandler => {
   return (req: Request, res: Response, next: NextFunction): void => {
-    passport.authenticate(["admin-session", "admin-bearer", "admin-api-token"], { session: false })(
-      req,
-      res,
-      next
+    const featureFlagRouter = req.scope.resolve(
+      ContainerRegistrationKeys.FEATURE_FLAG_ROUTER
     )
+    if (featureFlagRouter.isFeatureEnabled(MedusaV2Flag.key)) {
```
sradevski: This was also added by Philip I see

packages/medusa/src/api/middlewares/authenticate-customer.ts
```diff
@@ -1,11 +1,20 @@
...
 import { NextFunction, Request, RequestHandler, Response } from "express"
+
 import passport from "passport"
 
 // Optional customer authentication
 // If authenticated, middleware attaches customer to request (as user) otherwise we pass through
 // If you want to require authentication, use `requireCustomerAuthentication` in `packages/medusa/src/api/middlewares/require-customer-authentication.ts`
 export default (): RequestHandler => {
   return (req: Request, res: Response, next: NextFunction): void => {
+    const featureFlagRouter = req.scope.resolve(
```
sradevski: Same here

packages/medusa/src/api-v2/auth/[scope]/[authProvider]/route.ts
```diff
@@ -23,16 +23,18 @@ export const GET = async (req: MedusaRequest, res: MedusaResponse) => {
...
     res.redirect(location)
     return
   }
 
   if (success) {
     const { jwt_secret } = req.scope.resolve("configModule").projectConfig
-
     const token = jwt.sign(authUser, jwt_secret)
 
+    req.session.auth_user = authUser
```
Comment (sradevski) [inline packages/medusa/src/api-v2/auth/[scope]/[authProvider]/route.ts]: Doesn't this happen in the middleware (and if not, maybe it should happen there)?

packages/medusa/src/api-v2/admin/users/middlewares.ts
```diff
@@ -31,9 +34,24 @@ export const adminUserRoutesMiddlewares: MiddlewareRoute[] = [
...
+        QueryConfig.retrieveTransformQueryConfig
+      ),
+    ],
+  },
   {
     method: ["POST"],
     matcher: "/admin/users/:id",
     middlewares: [transformBody(AdminUpdateUserRequest)],
   },
+  {
```
sradevski: NIT: Let's put this on top, it's a bit easier to read and discover

packages/medusa/src/api-v2/admin/users/me/route.ts
```diff
@@ -0,0 +1,30 @@
+import {
+  ContainerRegistrationKeys,
+  MedusaError,
+  remoteQueryObjectFromString,
+} from "@medusajs/utils"
+import { MedusaRequest, MedusaResponse } from "../../../../types/routing"
+
+export const GET = async (req: MedusaRequest, res: MedusaResponse) => {
+  const userId = req.auth_user?.app_metadata.user_id
```
sradevski: I'd say it's good to check if the userId is defined before we call remoteQuery

packages/medusa/src/api-v2/admin/users/me/route.ts
```diff
@@ -0,0 +1,30 @@
...
+    },
+    fields: req.retrieveConfig.select as string[],
+  })
+
+  const [user] = await remoteQuery(query)
+
+  if (!user) {
+    throw new MedusaError(
+      MedusaError.Types.NOT_FOUND,
+      `Invite with id: ${userId} was not found`
```
carlos-r-l-rodrigues:
```suggestion
      `User with id: ${userId} was not found`
```

packages/medusa/src/api-v2/auth/[scope]/[authProvider]/route.ts
Comment (riqwan) [reply packages/medusa/src/api-v2/auth/[scope]/[authProvider]/route.ts]: This was added to make it work for the admin as the admin worked with sessions. The authenticate middleware doesn't "create" sessions. We probably need a create session endpoint or a config to enable that in the authenticate endpoint at some point or rework the admin to use token auth. 

I'll remove this from here for now as I'm sure there are better ways to achieve this.

packages/medusa/src/loaders/helpers/routing/index.ts
riqwan: This is to accept requests from whitelisted 3rd party redirects coming in from google and the lot.

sradevski (approved):

riqwan:
@olivermrbl might need your auth powers on this one. 🤞🏻 😄

packages/medusa/src/api-v2/admin/users/me/route.ts
```diff
@@ -0,0 +1,38 @@
...
+  ContainerRegistrationKeys,
+  MedusaError,
+  remoteQueryObjectFromString,
+} from "@medusajs/utils"
+import {
+  AuthenticatedMedusaRequest,
+  MedusaResponse,
+} from "../../../../types/routing"
+
+export const GET = async (
```
olivermrbl, line 11: thought(non-blocking): Would it make sense to throw a 401 on both undefined `app_metadata.user_id` and non-existing user? Otherwise, we would be leaking information about the user base no? Theoretically impossible for an adversary to do anything about it, but thought I'd bring it up

olivermrbl (approved):
LGTM

packages/medusa/src/api-v2/admin/users/me/route.ts
riqwan: This route is already protected by the auth middleware that will throw a 401, these guards will probably never be hit. Can still throw a 401 here if it still makes sense, wdyt?

PR #6447, opened 2024-02-20
feat: Update authentication middleware

pKorsholm:
**What**
- disabled `authenticate` and `authenticate-customer` middlewares if medusa-v2 is enabled
- add `isRegistered` check to authentication middleware to validate that a user is validated in authentication with respect to scope (default for scopes outside `admin` and `store` is that a user is registered, requiring a manual check to validate that an id exists for the auth-user of the given scope)
- add authentication middlewares for the remaining api-v2 endpoints 

**Why**
- disable: because passport middlewares were interfering with api-v2 and we are moving away from passport
- validating a user/customer is registered and not just that the auth-user exists is important to know they have a profile

### DIFF
--- a/integration-tests/plugins/__tests__/api-key/admin/api-key.spec.ts
+++ b/integration-tests/plugins/__tests__/api-key/admin/api-key.spec.ts
@@ -1,12 +1,14 @@
-import { ModuleRegistrationName } from "@medusajs/modules-sdk"
+import { initDb, useDb } from "../../../../environment-helpers/use-db"
+
+import { ApiKeyType } from "@medusajs/utils"
 import { IApiKeyModuleService } from "@medusajs/types"
+import { ModuleRegistrationName } from "@medusajs/modules-sdk"
+import adminSeeder from "../../../../helpers/admin-seeder"
+import { createAdminUser } from "../../../helpers/create-admin-user"
+import { getContainer } from "../../../../environment-helpers/use-container"
 import path from "path"
 import { startBootstrapApp } from "../../../../environment-helpers/bootstrap-app"
 import { useApi } from "../../../../environment-helpers/use-api"
-import { getContainer } from "../../../../environment-helpers/use-container"
-import { initDb, useDb } from "../../../../environment-helpers/use-db"
-import adminSeeder from "../../../../helpers/admin-seeder"
-import { ApiKeyType } from "@medusajs/utils"
 
 jest.setTimeout(50000)
 
@@ -36,7 +38,7 @@ describe("API Keys - Admin", () => {
   })
 
   beforeEach(async () => {
-    await adminSeeder(dbConnection)
+    await createAdminUser(dbConnection, adminHeaders)
   })
 
   afterEach(async () => {
@@ -60,7 +62,7 @@ describe("API Keys - Admin", () => {
       expect.objectContaining({
         id: created.data.apiKey.id,
         title: "Test Secret Key",
-        created_by: "test",
+        created_by: "admin_user",
       })
     )
     // On create we get the token in raw form so we can store it.
@@ -92,7 +94,7 @@ describe("API Keys - Admin", () => {
     expect(revoked.data.apiKey).toEqual(
       expect.objectContaining({
         id: created.data.apiKey.id,
-        revoked_by: "test",
+        revoked_by: "admin_user",
       })
     )
     expect(revoked.data.apiKey.revoked_at).toBeTruthy()
--- a/packages/medusa/src/api-v2/admin/api-keys/middlewares.ts
+++ b/packages/medusa/src/api-v2/admin/api-keys/middlewares.ts
@@ -1,25 +1,21 @@
-import { transformBody, transformQuery } from "../../../api/middlewares"
-import { MiddlewareRoute } from "../../../loaders/helpers/routing/types"
 import * as QueryConfig from "./query-config"
+
 import {
-  AdminGetApiKeysParams,
   AdminGetApiKeysApiKeyParams,
-  AdminPostApiKeysReq,
+  AdminGetApiKeysParams,
   AdminPostApiKeysApiKeyReq,
+  AdminPostApiKeysReq,
   AdminRevokeApiKeysApiKeyReq,
 } from "./validators"
+import { transformBody, transformQuery } from "../../../api/middlewares"
+
+import { MiddlewareRoute } from "../../../loaders/helpers/routing/types"
+import { authenticate } from "../../../utils/authenticate-middleware"
 
 export const adminApiKeyRoutesMiddlewares: MiddlewareRoute[] = [
   {
     matcher: "/admin/api-keys*",
-    // middlewares: [authenticate("admin", ["bearer", "session"])],
-    // TODO: Apply authentication middleware correctly once https://github.com/medusajs/medusa/pull/6447 is merged.
-    middlewares: [
-      (req, res, next) => {
-        req.auth_user = { id: "test" }
-        next()
-      },
-    ],
+    middlewares: [authenticate("admin", ["bearer", "session"])],
   },
   {
     method: ["GET"],


packages/medusa/src/utils/authenticate-middleware.ts
```diff
@@ -67,7 +67,21 @@ export const authenticate = (
...
+    const isMedusaScope =
+      stringEqualsOrRegexMatch(authScope, "admin") ||
+      stringEqualsOrRegexMatch(authScope, "store")
+
+    const isRegistered =
+      !isMedusaScope ||
+      (authUser?.app_metadata?.user_id &&
+        stringEqualsOrRegexMatch(authScope, "admin")) ||
+      (authUser?.app_metadata?.customer_id &&
+        stringEqualsOrRegexMatch(authScope, "store"))
```
olivermrbl, line 84: thought: Does this clash with our goal of using this middleware for more than just customers and users?

For example, if I create a new entity `Agent` that stores an `agent_id` in my `app_metadata`, and I would like agents to have admin access too. How would that be achieved?

pKorsholm: the agent auth_user would have to have a medusa user still. We have to validate for admin that a user exists though. If you have registered custom scope  that's not "store" or "admin" we expect the person developing the extension to manually handle validation of the auth_user's logged in status

With access control in the future this would be more easily handled but for now I think we need to ensure that a user exists. ie. an invite has been accepted

packages/medusa/src/api-v2/admin/campaigns/middlewares.ts
```diff
@@ -1,22 +1,28 @@
...
+import { MedusaV2Flag } from "@medusajs/utils"
+import { MiddlewareRoute } from "../../../loaders/helpers/routing/types"
+import { authenticate } from "../../../utils/authenticate-middleware"
 
 export const adminCampaignRoutesMiddlewares: MiddlewareRoute[] = [
   {
     matcher: "/admin/campaigns*",
-    middlewares: [isFeatureFlagEnabled(MedusaV2Flag.key)],
+    middlewares: [
+      isFeatureFlagEnabled(MedusaV2Flag.key),
```
olivermrbl: nit: remove flag, can be done separtely

packages/medusa/src/api-v2/admin/invites/accept/route.ts
```diff
@@ -23,7 +24,17 @@ export const POST = async (req: MedusaRequest, res: MedusaResponse) => {
     user: req.validatedBody as AdminPostInvitesInviteAcceptReq,
   } as InviteWorkflow.AcceptInviteWorkflowInputDTO
 
-  const { result: users } = await workflow.run({ input })
+  let users
+  try {
+    const { result: userss } = await workflow.run({ input })
+    users = userss
```
olivermrbl: todo: 
```suggestion
    const { result } = await workflow.run({ input })
    users = result
```

packages/medusa/src/api-v2/admin/invites/accept/route.ts
```diff
@@ -23,7 +24,17 @@ export const POST = async (req: MedusaRequest, res: MedusaResponse) => {
...
+  } catch (e) {
+    if (e.message.startsWith("jwt malformed")) {
+      res.status(401).json({ message: "Unauthorized" })
+      return
+    }
```
olivermrbl: thought: I wonder if we should just always respond with 401 in case of errors to not leak information about the authentication?

packages/medusa/src/api-v2/admin/promotions/middlewares.ts
```diff
@@ -1,23 +1,28 @@
...
+import { MedusaV2Flag } from "@medusajs/utils"
+import { MiddlewareRoute } from "../../../loaders/helpers/routing/types"
+import { authenticate } from "../../../utils/authenticate-middleware"
 
 export const adminPromotionRoutesMiddlewares: MiddlewareRoute[] = [
   {
     matcher: "/admin/promotions*",
-    middlewares: [isFeatureFlagEnabled(MedusaV2Flag.key)],
+    middlewares: [
+      isFeatureFlagEnabled(MedusaV2Flag.key),
```
olivermrbl: todo: rm flag

packages/medusa/src/utils/authenticate-middleware.ts
```diff
@@ -67,7 +67,21 @@ export const authenticate = (
...
+
+    if (
+      authUser &&
+      (isRegistered || (!isRegistered && options.allowUnregistered))
+    ) {
       req.auth_user = {
```
olivermrbl: `req.auth_user` will likely change. See internal discussion.

olivermrbl (changes requested):

packages/medusa/src/utils/authenticate-middleware.ts
pKorsholm: updated this pr to reflect internal discussion

packages/medusa/src/utils/authenticate-middleware.ts
```diff
@@ -97,3 +98,12 @@ export const authenticate = (
...
   }
 }
+
+const getActorId = (authUser: AuthUserDTO, scope: string | RegExp) => {
+  if (stringEqualsOrRegexMatch(scope, "admin")) {
+    return authUser.app_metadata.user_id
+  } else if (stringEqualsOrRegexMatch(scope, "store")) {
+    return authUser.app_metadata.customer_id
+  }
+  return authUser.app_metadata.medusa_id
```
sradevski: What does this ID represent?

packages/medusa/src/utils/authenticate-middleware.ts
```diff
@@ -82,8 +82,9 @@ export const authenticate = (
       authUser &&
       (isRegistered || (!isRegistered && options.allowUnregistered))
     ) {
-      req.auth_user = {
-        id: authUser.id,
+      req.auth = {
+        actor_id: getActorId(authUser, authScope) as string,
```
sradevski: NIT: Instead of casting as string here, you can set the return type of `getActorId` as a string

packages/medusa/src/types/routing.ts
```diff
@@ -1,14 +1,19 @@
...
 
 export interface MedusaRequest extends Request {
   user?: (User | Customer) & { customer_id?: string; userId?: string }
   scope: MedusaContainer
   session?: any
   requestId?: string
-  auth_user?: { id: string; app_metadata: Record<string, any>; scope: string }
+  auth?: {
+    actor_id: string
+    auth_user_id: string
```
sradevski, line 24: Wouldn't this and the actor_id point to the same ID?

packages/medusa/src/api-v2/store/customers/route.ts
```diff
@@ -29,7 +29,7 @@ export const POST = async (req: MedusaRequest, res: MedusaResponse) => {
   const customersData = req.validatedBody as CreateCustomerDTO
 
   const { result } = await createCustomers.run({
-    input: { customersData, authUserId: req.auth_user!.id },
+    input: { customersData, authUserId: req.auth!.auth_user_id },
```
sradevski: Let's check that the user ID is present before calling createCustomers

packages/medusa/src/api-v2/store/customers/me/addresses/route.ts
```diff
@@ -29,7 +30,7 @@ export const GET = async (req: MedusaRequest, res: MedusaResponse) => {
 }
 
 export const POST = async (req: MedusaRequest, res: MedusaResponse) => {
-  const customerId = req.auth_user!.app_metadata.customer_id
+  const customerId = req.auth!.actor_id!
```
sradevski: Let's check that there indeed is an actor ID, instead of forcing the typings.

sradevski (commented):
Few comments, one that I see repeating though is the usage of `!` for the types, which is better to not rely on and instead do an undefined check.

packages/medusa/src/api-v2/admin/invites/accept/route.ts
```diff
@@ -19,11 +20,18 @@ export const POST = async (req: MedusaRequest, res: MedusaResponse) => {
...
     user: req.validatedBody as AdminPostInvitesInviteAcceptReq,
   } as InviteWorkflow.AcceptInviteWorkflowInputDTO
 
-  const { result: users } = await workflow.run({ input })
+  let users
+  try {
+    const { result } = await workflow.run({ input })
+    users = result
+  } catch (e) {
+    res.status(401).json({ message: "Unauthorized" })
```
adrien2p, line 38: suggestion: we should probably log something here, the reason being that it can fail for many reason other than unauthorized and the log wouldn't show that information and make difficult to debug. wdyt?

packages/medusa/src/api-v2/store/customers/me/addresses/route.ts
adrien2p: I think the problem here is the lack of typings on the MedusaRequest, since this route is guarded by the auth middleware, I suppose the auth.actor_id will be present. When we will change the MedusaRequest to accept some template arugment we should be able to get rid of those wdyt?

packages/medusa/src/api-v2/store/customers/me/route.ts
```diff
@@ -3,7 +3,7 @@ import { MedusaRequest, MedusaResponse } from "../../../../types/routing"
 import { ModuleRegistrationName } from "@medusajs/modules-sdk"
 
 export const GET = async (req: MedusaRequest, res: MedusaResponse) => {
-  const id = req.auth_user!.app_metadata?.customer_id
+  const id = req.auth?.actor_id
```
adrien2p: todo: I think it comes back to the previous comment right? maybe `!` instead as this route is guarded by the auth middleware and id should be a string but not undefined right?

packages/medusa/src/types/routing.ts
```diff
@@ -1,14 +1,19 @@
+import type { Customer, User } from "../models"
 import type { NextFunction, Request, Response } from "express"
 
-import type { Customer, User } from "../models"
 import type { MedusaContainer } from "./global"
 
 export interface MedusaRequest extends Request {
```
adrien2p: Suggestion: shouldn't we break this request in two, one for the authenticated route and one for the non authenticated route?

I am thinking of something like the following:
```ts
export interface MedusaRequest<Body = unknown> extends Request {
  validatedBody: Body
  validatedQuery: RequestQueryFields & Record<string, unknown>
  listConfig: FindConfig<unknown>
  retrieveConfig: FindConfig<unknown>
  filterableFields: Record<string, unknown>
  allowedProperties: string[]
  includes?: Record<string, boolean>
  errors: string[]
}

export interface AuthenticatedMedusaRequest<Body = never>
  extends MedusaRequest<Body> {
  user: (User | Customer) & { customer_id?: string; userId?: string }
  auth: {
    actor_id: string
    auth_user_id: string
    app_metadata: Record<string, any>
    scope: string
  }
}
```

and later on we can probably rm the express augmentation too

cc @olivermrbl

packages/medusa/src/api-v2/store/customers/me/addresses/route.ts
sradevski: But if you forget to apply the middleware I guess it could be undefined, so as you suggested I guess it would be nice to have 2 types, Unauthenticated and Authenticated, which will solve this issue. But I'm fine with it as it is for now 👍

adrien2p: fine too, it is more to improve the DX, but even though the type says it is present it might not be if the middleware is not applied. I ve suggested the split elsewhere and it can serve as reference

packages/medusa/src/types/routing.ts
srindom: When we have 2.0 done, I'd remove `user` from here and just base everything on what's in `auth`.

Maybe worth keeping in mind that we should never user `req.user` in api-v2

packages/medusa/src/types/routing.ts
olivermrbl: The `auth_user_id` would be the ID of the authenticated user from the Auth module, while as the `actor_id` will be the ID of the entity the authenticated user is possibly associated with.

Example: User <> Auth
```
auth_user_id: "authusr_1234"
actor_id: "usr_1234"
```

Example: Customer <> Auth
```
auth_user_id: "authusr_1234"
actor_id: "cus_1234"
```

Example: Secret API key
```
auth_user_id: undefined
actor_id: "sk_1234"
```

packages/medusa/src/utils/authenticate-middleware.ts
pKorsholm: rm'ed, no need to think about other scopes rn 😄

olivermrbl (approved):
Changes LGTM, just need tests to pass. We might need to revisit the `AuthenticatedRequest` in the future, I am not sure having two types makes for a better DX. Internally yes, but not sure for external devs.

packages/medusa/src/types/routing.ts
sradevski: @olivermrbl ok understood the difference, but the naming is still confusing in my opinion. For example, tokens and users should be treated equally (they are both actors, as an actor is the one performing the operation), but an entity (eg. customer, or store) would be something completely different. Anyway it's not a blocker, but I think we need to rethink our naming and approach here.

olivermrbl: In the examples above, we are treating the users and tokens equally wrt the actor. Auth user is a separate concept from entities Customer or User and is specifically designed to provide authentication to those. Do you think the auth user should be renamed? I am not sure I understand what is confusing here.

sradevski: @olivermrbl Ok I think the part that got me confused is what `customer` means, I thought it represents an `organization`, but now I remembered that it's simply a `storefront user` basically. All clear now 👍

olivermrbl:
@pKorsholm, could I get you to fix the tests, so we can merge? 🙏

PR #6449, opened 2024-02-20
feat: Create cart with line items

olivermrbl:
**What**
- Add support for creating a cart with items
- Add endpoint `POST /store/carts/:id/line-items`
- Add `CreateCartWorkflow`
- Add `AddToCartWorkflow`
- Add steps for both workflows

**Testing**
- Endpoints
- Workflows

I would still call this a first iteration, as we are missing a few pieces of the full flow, such as payment sessions, discounts, and taxes.

### DIFF
--- a/integration-tests/plugins/__tests__/cart/store/cart.workflows.spec.ts
+++ b/integration-tests/plugins/__tests__/cart/store/cart.workflows.spec.ts
@@ -382,7 +382,7 @@ describe("Carts workflows", () => {
       ])
     })
 
-    it("should throw if variant doesnt exist sets for variant exist", async () => {
+    it("should throw if variant does not exist", async () => {
       const cart = await cartModuleService.create({
         currency_code: "usd",
       })
--- a/packages/core-flows/src/definition/cart/steps/add-to-cart.ts
+++ b/packages/core-flows/src/definition/cart/steps/add-to-cart.ts
@@ -17,5 +17,15 @@ export const addToCartStep = createStep(
     const items = await cartService.addLineItems(data.items)
 
     return new StepResponse(items)
+  },
+  async (createdLineItems, { container }) => {
+    const cartService: ICartModuleService = container.resolve(
+      ModuleRegistrationName.CART
+    )
+    if (!createdLineItems?.length) {
+      return
+    }
+
+    await cartService.removeLineItems(createdLineItems.map((c) => c.id))
   }
 )


packages/core-flows/src/definition/cart/steps/get-variant-price-sets.ts
```diff
@@ -0,0 +1,65 @@
...
+import { StepResponse, createStep } from "@medusajs/workflows-sdk"
+
+interface StepInput {
+  variantIds: string[]
+  // TODO: Expand the context passed to calculatePrices
+  region: RegionDTO
+}
+
+export const getVariantPriceSetsStepId = "get-variant-price-sets"
+export const getVariantPriceSetsStep = createStep(
```
olivermrbl, line 12: @riqwan @pKorsholm, would appreciate your eyes on this implementation. Is there a more appropriate way of retrieving variant prices?

Please ignore the context for now. This will be revisited later.

packages/core-flows/src/definition/cart/steps/get-variant-price-sets.ts
```diff
@@ -0,0 +1,65 @@
...
+    const priceSetToVariant: Map<string, string> = new Map(
+      variantPriceSets.map((variantPriceSet) => [
+        variantPriceSet.price_set_id,
+        variantPriceSet.variant_id,
+      ])
+    )
+
+    const calculatedPriceSets = await pricingModuleService.calculatePrices(
+      { id: [...priceSetToVariant.keys()] },
+      { context: { currency_code: data.region.currency_code } } // TODO: Should allow for more context properties
```
riqwan: overall looks correct on how you'd retrieve prices.

My only should would be the pricing context. If you want to retrieve the correct price here, you need the exact context to be passed when viewing the product. 

Edit: Just saw your comment on ignoring the context :D

packages/core-flows/src/definition/cart/steps/get-variant-price-sets.ts
```diff
@@ -0,0 +1,65 @@
...
+      return new StepResponse({})
+    }
+
+    const pricingModuleService = container.resolve<IPricingModuleService>(
+      ModuleRegistrationName.PRICING
+    )
+
+    const variables = {
+      variant_id: data.variantIds,
+      take: null, // TODO: We probably want to add a limit here
```
riqwan: we added `null` on the take here specifically to get all variants, since by default the limit is 15 and we miss out calculating some variant prices.

Other option is to paginate them until the last item, which we haven't done.

packages/core-flows/src/definition/cart/steps/get-variant-price-sets.ts
```diff
@@ -0,0 +1,70 @@
...
+      return new StepResponse({})
+    }
+
+    const pricingModuleService = container.resolve<IPricingModuleService>(
+      ModuleRegistrationName.PRICING
+    )
+
+    const remoteQuery = container.resolve("remoteQuery")
+
+    const variantPriceSets = await remoteQuery(
```
olivermrbl, line 25: @carlos-r-l-rodrigues, could also make sense to have your eyes on this. What do you think about this approach?

packages/core-flows/src/definition/cart/steps/get-variant-price-sets.ts
```diff
@@ -0,0 +1,74 @@
...
+      `
+      query {
+        variant (id: $id) {
+          id
+          price {
```
olivermrbl: @riqwan, this can return multiple prices per variant, right?

And that in turn will mean, `calculatePrices` below will return multiple calculated prices for single variant

packages/core-flows/src/definition/cart/steps/get-variant-price-sets.ts
carlos-r-l-rodrigues: you can use variables with the json syntax as well.
so we don't have to use gql strings

```
await remoteQuery({
  variant {
    fields: ["id"],
    price {
      fields: ["price_set_id"]
    }
  }
}, {
   variant: {
     id: data.variantIds,
   }
})
```

carlos-r-l-rodrigues: also I would like to have this implemented in the RemoteQuery:

```
await remoteQuery({
  variant {
    fields: ["id"],
    price {
      fields: ["price_set_id"]
    }
  }
}, {
   variant: {
     id: data.variantIds,
   },
   "variant.price": {
     context: data.context
   }
})
```

in case we have a parameter context, instead of calling `list`, it could call a `calculated` method on the module

olivermrbl: > you can use variables with the json syntax as well.

Yeah, I just thought the gql approach reads better, but I can change it if you prefer the json syntax. We should probably be consistent with how we use it. 

> also I would like to have this implemented in the RemoteQuery:

I think this is an excellent idea. Does that mean the `calculatePrice` method would somehow be a "resolver" of the `variant.price` field? Where do you imagine this is defined? In the link def.?

carlos-r-l-rodrigues: I also think the GQL looks easier to read, but when we get to the point of using it in endpoints, with the option to extend fields, it will be harder to work with strings.
But this is being used internally, so I don't really have a strong opinion.
you'll just have to add the variable on the `query ($var: Any)` on the gql query. 

Regarding `calculate` method, it is up to us to decide where to set this. The same way we call `listAndCount` if there is any pagination parameter, we could call calculate if we have the property `context`.

olivermrbl: Makes sense. I guess these additional methods on the `remoteQuery` vary from link to link, which is why I thought it would make sense to put them in the link-definition.

integration-tests/plugins/__tests__/cart/store/carts.spec.ts
```diff
@@ -224,93 +285,6 @@ describe("Store Carts API", () => {
         })
       ).rejects.toThrow()
     })
-
-    it("should throw if sales channel is disabled", async () => {
```
olivermrbl, line 228: I moved these tests to `cart.workflows.spec.ts` to match against proper error messages.

packages/core-flows/src/definition/cart/steps/get-variant-price-sets.ts
carlos-r-l-rodrigues: we can find somewhere in the joiner config to set these methods. Because this is more a config of the module itself to calculate something based on parameters, not their links.
This case the link variant -> price_set doesn't change. what changes is just the amount of prices that will return depending on the context passed.

packages/core-flows/src/definition/cart/steps/get-variant-price-sets.ts
riqwan: does it? I remember adding a config in the link module to return just a singular entity. Let me know if it does return multiple, thats a bug.

olivermrbl:
@adrien2p, thanks for helping out 🥇

integration-tests/plugins/__tests__/cart/store/carts.spec.ts
```diff
@@ -77,13 +80,74 @@ describe("Store Carts API", () => {
...
+              title: "Test variant",
+            },
+            {
+              title: "Test variant 2",
+            },
+          ],
+        },
+      ])
+
+      const priceSet = await pricingModule.create({
```
sradevski: I assume we can do `const [first, second] = create([first, second]`? At least it's supported like that in the other modules

packages/core-flows/src/definition/cart/steps/get-variant-price-sets.ts
```diff
@@ -0,0 +1,82 @@
+import { ModuleRegistrationName } from "@medusajs/modules-sdk"
```
sradevski, line 1: A general question - Should getters be a step in the workflow, or should they just be helper functions? Since we won't have a compensation function for getters, and usually they are just as a support role, performance will likely be worse if they are a step in the workflow.

packages/core-flows/src/definition/cart/workflows/create-carts.ts
```diff
@@ -26,9 +40,27 @@ export const createCartWorkflow = createWorkflow(
       findOrCreateCustomerStep({
         customerId: input.customer_id,
         email: input.email,
-      })
+      }),
+      validateVariantsExistStep({ variantIds })
     )
 
+    // TODO: This is on par with the context used in v1.*, but we can be more flexible.
+    const pricingContext = transform(
```
sradevski, line 48: What does the transform function do?

olivermrbl: The `transform` function was added as a middleware between steps where the input and output don't fit perfectly together. We don't want data transformations to live as separate steps because they are typically dependent on the specific flow of events in a workflow, so we added the transformer as a utility function that pipes the data to other steps in the correct shape.

olivermrbl: Imagine steps created independently of each other, e.g. as part of two different workflows, that you now want to use in a new third workflow. Ideally, you will just dump these steps in, but it is unlikely that you can do that, as they were created separately and typically as generic as possible.

integration-tests/plugins/__tests__/cart/store/carts.spec.ts
olivermrbl: Yes, that is possible. Will update.

packages/core-flows/src/definition/cart/workflows/create-carts.ts
sradevski: @olivermrbl ah ok got it, I just checked the code and I guess the main point here is that they are lazily executed, compared to eg. here just saying
`pricingContext = {currency_code: input.currency_code, region_id: region.id...}`

packages/workflows-sdk/src/utils/composer/type.ts
```diff
@@ -10,6 +10,15 @@ import { Context, MedusaContainer } from "@medusajs/types"
 export type StepFunctionResult<TOutput extends unknown | unknown[] = unknown> =
```
olivermrbl, line 10: @adrien2p, it seems these changes are possibly causing some issues for the Redis workflow engine tests. See [here](https://github.com/medusajs/medusa/actions/runs/8021017524/job/21912240155?pr=6449). Does that sound plausible? Can't see what else in the PR should cause it.

adrien2p: yes I ve sent you a message the otherday that some tests will fail and need fixing, mainly becuse of symbole vs string. Do you want me to jump on it after i finish my thing?

olivermrbl: I thought you referred to tests in the context of the cart module. I don't think I am in a position to solve the workflow-related issues, so that would be great - thanks 💪

integration-tests/plugins/__tests__/cart/store/cart.workflows.spec.ts
```diff
@@ -0,0 +1,412 @@
...
+          pricingService: {
+            price_set_id: priceSet.id,
+          },
+        },
+      ])
+
+      const { result } = await createCartWorkflow(appContainer).run({
+        input: {
+          email: "tony@stark.com",
+          currency_code: "usd",
```
riqwan, line 121: suggestion: If the variant does not have a "usd" price, does it throw an error? If so, might be good to have that as a test.

integration-tests/plugins/__tests__/cart/store/cart.workflows.spec.ts
```diff
@@ -0,0 +1,412 @@
...
+          action: "get-variant-price-sets",
+          handlerType: "invoke",
+          error: new Error(
+            `Variants with IDs ${product.variants[0].id} do not have a price`
+          ),
+        },
+      ])
+    })
+
+    it("should throw if variant doesnt exist sets for variant exist", async () => {
```
riqwan:
```suggestion
    it("should throw if variant does not exist", async () => {
```

packages/core-flows/src/definition/cart/steps/add-to-cart.ts
```diff
@@ -0,0 +1,21 @@
...
+  addToCartStepId,
+  async (data: StepInput, { container }) => {
+    const cartService = container.resolve<ICartModuleService>(
+      ModuleRegistrationName.CART
+    )
+
+    const items = await cartService.addLineItems(data.items)
+
+    return new StepResponse(items)
+  }
```
riqwan, line 30: missing compensation

```suggestion
  },
  async (createdLineItems, { container }) => {
    const cartService: ICartModuleService = container.resolve(
      ModuleRegistrationName.CART
    )

    if (!createdLineItems?.length) {
      return
    }

    await cartService.removeLineItems(
      createdLineItems.map((c) => c.id)
    )
  }
```

packages/core-flows/src/definition/cart/steps/find-one-or-any-region.ts
```diff
@@ -14,7 +15,10 @@ export const findOneOrAnyRegionStep = createStep(
       const regions = await service.list({})
```
riqwan, line 15: question: shouldn't this be scoped to a currency_code or something else?

packages/core-flows/src/definition/cart/workflows/add-to-cart.ts
```diff
@@ -0,0 +1,75 @@
...
+    })
+
+    validateVariantsExistStep({ variantIds })
+
+    // TODO: This is on par with the context used in v1.*, but we can be more flexible.
+    const pricingContext = transform({ cart: input.cart }, (data) => {
+      return {
+        currency_code: data.cart.currency_code,
+        region_id: data.cart.region_id,
+        customer_id: data.cart.customer_id,
```
riqwan, line 39: Not important for now, but this might need customer_group_id as well to activate price_list scoped queries.

packages/core-flows/src/definition/cart/workflows/create-carts.ts
```diff
@@ -51,11 +83,53 @@ export const createCartWorkflow = createWorkflow(
...
+        ...data.cartInput,
+        items: data.lineItems,
+      }
+    })
+
+    const carts = createCartsStep([cartToCreate])
 
-    // @ts-ignore
-    const cart = createCartsStep([cartInput])
+    const cart = transform({ carts }, (data) => data.carts?.[0])
```
riqwan, line 131: completely throwable, but with the new type changes, we should be able to do just this:

```suggestion
    return carts[0]
```

integration-tests/plugins/__tests__/cart/store/cart.workflows.spec.ts
olivermrbl: I believe this case is covered here: 

https://github.com/medusajs/medusa/blob/0ed6dac32b8a0076af2f4d1ff3ea0d8ed5280d26/integration-tests/plugins/__tests__/cart/store/cart.workflows.spec.ts#L345-L383

packages/core-flows/src/definition/cart/steps/add-to-cart.ts
olivermrbl: Nice, thanks!

packages/core-flows/src/definition/cart/steps/find-one-or-any-region.ts
olivermrbl: We actually expect you to pass a region ID when creating a cart. But in the unlikely case, that you do not, we assign you any region. The currency code on the cart would likely come from the region, you have selected, so we can't scope it by that.

packages/core-flows/src/definition/cart/workflows/create-carts.ts
olivermrbl: Are you sure? This doesn't work for me. I am seeing the following error:

CleanShot 2024-02-26 at 13 57 19


And I believe this is because we want to access an element of an unresolved variable. If we were just returning `carts` then it would be a different case, but I am not sure.

riqwan: ah no, you're right there.

PR #6455, opened 2024-02-20
feat(fulfillment): Shipping options, rules CRUD + rules based context filtering

adrien2p:
**What**
- Update shipping options with its rules and type
- create/update rules independently
- context based validation fundation
- 🔴 list shipping options with context rules fitlering will come in a separate pr to keep this one smaller

FIXES CORE-1743
FIXES CORE-1764

### DIFF
--- a/integration-tests/plugins/__tests__/api-key/admin/api-key.spec.ts
+++ b/integration-tests/plugins/__tests__/api-key/admin/api-key.spec.ts
@@ -1,0 +1,109 @@
+import { ModuleRegistrationName } from "@medusajs/modules-sdk"
+import { IApiKeyModuleService } from "@medusajs/types"
+import path from "path"
+import { startBootstrapApp } from "../../../../environment-helpers/bootstrap-app"
+import { useApi } from "../../../../environment-helpers/use-api"
+import { getContainer } from "../../../../environment-helpers/use-container"
+import { initDb, useDb } from "../../../../environment-helpers/use-db"
+import adminSeeder from "../../../../helpers/admin-seeder"
+import { ApiKeyType } from "@medusajs/utils"
+
+jest.setTimeout(50000)
+
+const env = { MEDUSA_FF_MEDUSA_V2: true }
+const adminHeaders = {
+  headers: { "x-medusa-access-token": "test_token" },
+}
+
+describe("API Keys - Admin", () => {
+  let dbConnection
+  let appContainer
+  let shutdownServer
+  let service: IApiKeyModuleService
+
+  beforeAll(async () => {
+    const cwd = path.resolve(path.join(__dirname, "..", "..", ".."))
+    dbConnection = await initDb({ cwd, env } as any)
+    shutdownServer = await startBootstrapApp({ cwd, env })
+    appContainer = getContainer()
+    service = appContainer.resolve(ModuleRegistrationName.API_KEY)
+  })
+
+  afterAll(async () => {
+    const db = useDb()
+    await db.shutdown()
+    await shutdownServer()
+  })
+
+  beforeEach(async () => {
+    await adminSeeder(dbConnection)
+  })
+
+  afterEach(async () => {
+    const db = useDb()
+    await db.teardown()
+  })
+
+  it("should correctly implement the entire lifecycle of an api key", async () => {
+    const api = useApi() as any
+    const created = await api.post(
+      `/admin/api-keys`,
+      {
+        title: "Test Secret Key",
+        type: ApiKeyType.SECRET,
+      },
+      adminHeaders
+    )
+
+    expect(created.status).toEqual(200)
+    expect(created.data.apiKey).toEqual(
+      expect.objectContaining({
+        id: created.data.apiKey.id,
+        title: "Test Secret Key",
+        created_by: "test",
+      })
+    )
+    // On create we get the token in raw form so we can store it.
+    expect(created.data.apiKey.token).toContain("sk_")
+
+    const updated = await api.post(
+      `/admin/api-keys/${created.data.apiKey.id}`,
+      {
+        title: "Updated Secret Key",
+      },
+      adminHeaders
+    )
+
+    expect(updated.status).toEqual(200)
+    expect(updated.data.apiKey).toEqual(
+      expect.objectContaining({
+        id: created.data.apiKey.id,
+        title: "Updated Secret Key",
+      })
+    )
+
+    const revoked = await api.post(
+      `/admin/api-keys/${created.data.apiKey.id}/revoke`,
+      {},
+      adminHeaders
+    )
+
+    expect(revoked.status).toEqual(200)
+    expect(revoked.data.apiKey).toEqual(
+      expect.objectContaining({
+        id: created.data.apiKey.id,
+        revoked_by: "test",
+      })
+    )
+    expect(revoked.data.apiKey.revoked_at).toBeTruthy()
+
+    const deleted = await api.delete(
+      `/admin/api-keys/${created.data.apiKey.id}`,
+      adminHeaders
+    )
+    const listedApiKeys = await api.get(`/admin/api-keys`, adminHeaders)
+
+    expect(deleted.status).toEqual(200)
+    expect(listedApiKeys.data.apiKeys).toHaveLength(0)
+  })
+})
--- a/integration-tests/plugins/medusa-config.js
+++ b/integration-tests/plugins/medusa-config.js
@@ -102,5 +102,10 @@ module.exports = {
       resources: "shared",
       resolve: "@medusajs/region",
     },
+    [Modules.API_KEY]: {
+      scope: "internal",
+      resources: "shared",
+      resolve: "@medusajs/api-key",
+    },
   },
 }
--- a/integration-tests/plugins/package.json
+++ b/integration-tests/plugins/package.json
@@ -9,6 +9,7 @@
     "build": "babel src -d dist --extensions \".ts,.js\""
   },
   "dependencies": {
+    "@medusajs/api-key": "workspace:^",
     "@medusajs/auth": "workspace:*",
     "@medusajs/cache-inmemory": "workspace:*",
     "@medusajs/customer": "workspace:^",
--- a/packages/api-key/integration-tests/__tests__/api-key-module-service.spec.ts
+++ b/packages/api-key/integration-tests/__tests__/api-key-module-service.spec.ts
@@ -105,10 +105,12 @@ moduleIntegrationTestRunner({
 
         it("should allow for at most two tokens, where one is revoked", async function () {
           const firstApiKey = await service.create(createSecretKeyFixture)
-          await service.revoke({
-            id: firstApiKey.id,
-            revoked_by: "test",
-          })
+          await service.revoke(
+            { id: firstApiKey.id },
+            {
+              revoked_by: "test",
+            }
+          )
 
           await service.create(createSecretKeyFixture)
           const err = await service
@@ -123,8 +125,7 @@ moduleIntegrationTestRunner({
       describe("revoking API keys", () => {
         it("should have the revoked at and revoked by set when a key is revoked", async function () {
           const firstApiKey = await service.create(createSecretKeyFixture)
-          const revokedKey = await service.revoke({
-            id: firstApiKey.id,
+          const revokedKey = await service.revoke(firstApiKey.id, {
             revoked_by: "test",
           })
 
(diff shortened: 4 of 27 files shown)


packages/fulfillment/src/migrations/Migration20240221115805.ts
```diff
@@ -1,6 +1,6 @@
 import { Migration } from '@mikro-orm/migrations';
 
-export class Migration20240219115644 extends Migration {
+export class Migration20240221115805 extends Migration {
```
sradevski: It's nice to name migrations so they are more easily discoverable (at least when we have more than 1)

packages/fulfillment/src/services/fulfillment-module-service.ts
```diff
@@ -235,6 +252,11 @@ export default class FulfillmentModuleService<
       return []
     }
 
+    const rules = data_.flatMap((d) => d.rules)
```
sradevski, line 256: Maybe it's nice to validate the individual requests as well, as you can have
[[], [rule]]
and I assume an empty ruleset isn't really valid? Maybe it is, just wanted to bring it up

packages/fulfillment/src/services/fulfillment-module-service.ts
```diff
@@ -328,6 +350,59 @@ export default class FulfillmentModuleService<
...
+      | FulfillmentTypes.CreateShippingOptionRuleDTO,
+    @MedusaContext() sharedContext: Context = {}
+  ): Promise<TShippingOptionRuleEntity | TShippingOptionRuleEntity[]> {
+    const data_ = Array.isArray(data) ? data : [data]
+
+    if (!data_.length) {
+      return []
+    }
+
+    const createdShippingOptionRules =
```
sradevski, line 401: Any validation of the data we need to do? Eg. make sure the rule has at least one condition, etc?

packages/fulfillment/src/services/fulfillment-module-service.ts
```diff
@@ -671,7 +746,137 @@ export default class FulfillmentModuleService<
...
+      throw new MedusaError(
+        MedusaError.Types.NOT_FOUND,
+        `The following shipping options do not exist: ${Array.from(
+          missingShippingOptionIds
+        ).join(", ")}`
+      )
+    }
+
+    const ruleIdsToDelete: string[] = []
+    data_.forEach((shippingOption) => {
```
sradevski: This piece of code is a great candidate for a refactor. Let's try to break it down into helper functions, as it's really difficult to follow the logic as it is.

packages/fulfillment/src/utils/utils.ts
```diff
@@ -0,0 +1,144 @@
+import {isString, MedusaError, pickValueFromObject} from "@medusajs/utils"
+
+export type Rule = {
+  attribute: string
+  operator: RuleOperator
+  value: string | string[]
+}
+
+export enum RuleOperator {
```
sradevski, line 17: Should this be part of the top-level utils?

packages/utils/src/modules-sdk/internal-module-service-factory.ts
```diff
@@ -60,7 +60,7 @@ export function internalModuleServiceFactory<
     }
 
     static buildUniqueCompositeKeyValue(keys: string[], data: object) {
-      return keys.map((k) => data[k]).join("_")
+      return keys.map((k) => data[k]).join(":")
```
sradevski, line 63: Just curious, why did we do this change?

sradevski (commented):
I want to put in a bit more time to review the rules engine separately, as I think it's quite easy to mess things up there :)

packages/fulfillment/src/migrations/Migration20240221115805.ts
adrien2p: the name is auto generated, do you have strong opinion to rename it since it is the only one?

packages/fulfillment/src/services/fulfillment-module-service.ts
adrien2p: yes am empty rule set is accepted, at least from my understanding, the shipping options could be filtered only by its location (geo zone) but has not other rules to apply for it. Also, the flatMap for `[[], [rule]` will return `[rule]`

packages/fulfillment/src/services/fulfillment-module-service.ts
adrien2p: my bad, I ve missed to apply the validation here, only the db will reject for non nullable columns. I ll add that in

packages/fulfillment/src/utils/utils.ts
adrien2p: at somepoint, all those utils might be moved higher, for now I am constraining it to the module to not create any confusion. But we will have to discuss something about all the module having the same approach to standardize it and re use those utils. For now, i ll keep them here as there is inconsistency between the modules on that area

packages/utils/src/modules-sdk/internal-module-service-factory.ts
adrien2p: it happen that all our id's are composed of a prefix and a suffix separated byt `_`. Here were joing with `_` and then replace `_` by `-` which break the ids

packages/fulfillment/src/utils/utils.ts
adrien2p: I can add a todo if you think it could be good to have?

packages/fulfillment/src/migrations/Migration20240221115805.ts
sradevski: @adrien2p I know it's autogenerated, but all of them are named the same, so when the second one comes we'll need to rename it then :) Not a big deal at all, I just thought it's nicer if this is called eg `InitialSetup...` for example

packages/fulfillment/src/utils/utils.ts
sradevski: Yeah at least a todo would be nice 👍

packages/utils/src/modules-sdk/internal-module-service-factory.ts
sradevski: Ah I see 👍

packages/fulfillment/src/migrations/Migration20240221115805.ts
adrien2p: ahah got it 👍

adrien2p: I ve changed my script option so that everytime I generate the initial migration it will have a specific name

packages/fulfillment/src/models/shipping-option-rule.ts
```diff
@@ -31,6 +32,8 @@ const ShippingOptionIdIndex = createPsqlIndexStatementHelper({
   where: "deleted_at IS NULL",
 })
 
+type ruleOperator = "in" | "eq" | "ne" | "gt" | "gte" | "lt" | "lte" | "nin"
```
olivermrbl:
```suggestion
type RuleOperator = "in" | "eq" | "ne" | "gt" | "gte" | "lt" | "lte" | "nin"
```

packages/fulfillment/src/models/shipping-option-type.ts
```diff
@@ -48,12 +48,9 @@ export default class ShippingOptionType {
...
   code: string
 
-  @Property({ columnType: "text" })
-  @ShippingOptionIdIndex.MikroORMIndex()
-  shipping_option_id: string
-
   @OneToOne(() => ShippingOption, (so) => so.type, {
-    persist: false,
+    type: "text",
+    cascade: [],
```
olivermrbl: rm?
```suggestion
```

packages/fulfillment/src/services/fulfillment-module-service.ts
```diff
@@ -671,7 +748,117 @@ export default class FulfillmentModuleService<
...
+      {
+        id: shippingOptionIds,
+      },
+      {
+        relations: ["rules"],
+      },
+      sharedContext
+    )
+
+    FulfillmentModuleService.validateMissingShippingOptions_(
```
olivermrbl, line 798: Why is this a static method? For validation, we typically use regular methods. You've done that with `validateRules`, so I would consider changing this to be consistent

packages/fulfillment/src/services/fulfillment-module-service.ts
```diff
@@ -671,7 +748,117 @@ export default class FulfillmentModuleService<
...
+      if (!shippingOption.rules) {
+        return
+      }
+
+      const existingShippingOption = shippingOptions.find(
+        (s) => s.id === shippingOption.id
+      )!
+      const existingRules = existingShippingOption.rules
+
+      FulfillmentModuleService.validateMissingShippingOptionRules(
```
olivermrbl, line 814: Same as other comment.

packages/fulfillment/src/services/fulfillment-module-service.ts
```diff
@@ -671,7 +748,117 @@ export default class FulfillmentModuleService<
...
+
+    const ruleIdsToDelete: string[] = []
+    dataArray.forEach((shippingOption) => {
+      if (!shippingOption.rules) {
+        return
+      }
+
+      const existingShippingOption = shippingOptions.find(
+        (s) => s.id === shippingOption.id
+      )!
```
olivermrbl: nit: Create map before `forEach` and access in constant time

packages/fulfillment/src/services/fulfillment-module-service.ts
```diff
@@ -671,7 +748,117 @@ export default class FulfillmentModuleService<
...
+        existingShippingOption,
+        shippingOption
+      )
+
+      const existingRulesMap = new Map(
+        existingRules.map((rule) => [rule.id, rule])
+      )
+      const updatedRules = shippingOption.rules
+      const toDeleteRuleIds = getSetDifference(
+        new Set(existingRules.map((r) => r.id)),
```
olivermrbl:
```suggestion
        new Set(existingRulesMap.keys()),
```

packages/fulfillment/src/services/fulfillment-module-service.ts
```diff
@@ -671,7 +748,117 @@ export default class FulfillmentModuleService<
...
+      FulfillmentModuleService.validateMissingShippingOptionRules(
+        existingShippingOption,
+        shippingOption
+      )
+
+      const existingRulesMap = new Map(
+        existingRules.map((rule) => [rule.id, rule])
+      )
+      const updatedRules = shippingOption.rules
+      const toDeleteRuleIds = getSetDifference(
```
olivermrbl: question: Could we just use `arrayDifference` instead of converting them to sets?

packages/fulfillment/src/services/fulfillment-module-service.ts
```diff
@@ -724,4 +911,111 @@ export default class FulfillmentModuleService<
...
+
+  protected static validateMissingShippingOptions_(
+    shippingOptions: ShippingOption[],
+    shippingOptionsData: FulfillmentTypes.UpdateShippingOptionDTO[]
+  ) {
+    const shippingOptionSet = new Set(shippingOptions.map((s) => s.id))
+    const expectedShippingOptionSet = new Set(
+      shippingOptionsData.map((s) => s.id)
+    )
+    const missingShippingOptionIds = getSetDifference(
```
olivermrbl: question: Can we just use `arrayDifference` instead?

packages/fulfillment/src/services/fulfillment-module-service.ts
adrien2p: validateRules is a utils that leaves with the "rule engine" it is not a method. Also, I use static as it is not using anything from the class context but is tight to the module data

packages/fulfillment/src/services/fulfillment-module-service.ts
adrien2p: yes def 👍

packages/fulfillment/src/services/fulfillment-module-service.ts
olivermrbl: > validateRules is a utils that leaves with the "rule engine" it is not a method

Sorry, this was an oversight from my side. 

I am still not super fond of the static methods. Haven't we generally avoided those in the past?

olivermrbl (approved):
Slapping an approval on this one, so you can merge after addressing the comments above 👋

PR #6458, opened 2024-02-20
wip: payment module big number

fPolic:
**WIP**
- big number properties for payment module models
- add payment module migrations

### DIFF
--- a/packages/utils/src/dal/mikro-orm/big-number-field.ts
+++ b/packages/utils/src/dal/mikro-orm/big-number-field.ts
@@ -1,4 +1,5 @@
 import { BigNumber } from "../../totals/big-number"
+import { isDefined } from "../../common"
 
 const bigNumberFields = new WeakMap<
   object,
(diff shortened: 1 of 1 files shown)


packages/payment/src/models/payment-session.ts
```diff
@@ -65,10 +72,52 @@ export default class PaymentSession {
...
+    console.log("!!! ON INIT", typeof this.amount)
+
+    const amount = this.raw_amount ?? this.amount
+
+    // OnInit can be sometimes fired twice, once when an entity reference is created, and once after its populated.
+    // The first call will only contain the id and amounts will be undefined so calling the BigNumber constructor will fail.
+    // On the second call with populated data amounts will exist.
+    if (typeof amount === "undefined") {
+      return
+    }
```
olivermrbl: @riqwan, this does sound an awful lot like what you experienced no?

fPolic: ~~We should probably move BigNumber stuff from `@OnInit` to `@OnLoad` which will only fire when the entity is fully loaded.~~ Nvm: we need on init so this could work when `.create` is called

olivermrbl:
Should we update the PR with the `BigNumberField` decorator that @carlos-r-l-rodrigues introduced?

PR #6462, opened 2024-02-21
chore(medusa-test-utils): Add debug option to the module test runner

adrien2p:
**What**
Add a debug options to the test runner, for now it will only switch the db to log the query but it could evolve to do more debug.


adrien2p:
wait before merging, I ve pushed something from another pr 😂

PR #6463, opened 2024-02-21
feat(api-key): Add CRUD functionalities to the api key module

sradevski:
**What?**

Adds implementation of CRUD functionalities to the api key module.
Adds support for both publishable and secret keys.

**What next?**
The tokens need to be utilized in the middleware, and the authenticate endpoint needs to be implemented.

### DIFF
--- a/packages/api-key/src/models/api-key.ts
+++ b/packages/api-key/src/models/api-key.ts
@@ -17,12 +17,19 @@ const TypeIndex = createPsqlIndexStatementHelper({
   columns: "type",
 })
 
+const TokenIndex = createPsqlIndexStatementHelper({
+  tableName: "api_key",
+  columns: "token",
+  unique: true,
+})
+
 @Entity()
 export default class ApiKey {
   @PrimaryKey({ columnType: "text" })
   id: string
 
   @Property({ columnType: "text" })
+  @TokenIndex.MikroORMIndex()
   token: string
 
   @Property({ columnType: "text" })
--- a/packages/api-key/src/services/api-key-module-service.ts
+++ b/packages/api-key/src/services/api-key-module-service.ts
@@ -89,6 +89,7 @@ export default class ApiKeyModuleService<TEntity extends ApiKey = ApiKey>
       token:
         generatedTokens.find((t) => t.hashedToken === key.token)?.rawToken ??
         key.token,
+      salt: undefined,
     }))
 
     return Array.isArray(data) ? responseWithRawToken : responseWithRawToken[0]
@@ -99,7 +100,7 @@ export default class ApiKeyModuleService<TEntity extends ApiKey = ApiKey>
     data: ApiKeyTypes.CreateApiKeyDTO[],
     @MedusaContext() sharedContext: Context = {}
   ): Promise<[TEntity[], TokenDTO[]]> {
-    await this.validateCreateApiKeys(data, sharedContext)
+    await this.validateCreateApiKeys_(data, sharedContext)
 
     const normalizedInput: CreateApiKeyDTO[] = []
     const generatedTokens: TokenDTO[] = []
@@ -201,7 +202,7 @@ export default class ApiKeyModuleService<TEntity extends ApiKey = ApiKey>
       sharedContext
     )
 
-    return this.baseRepository_.serialize<ApiKeyTypes.ApiKeyDTO[]>(
+    return await this.baseRepository_.serialize<ApiKeyTypes.ApiKeyDTO[]>(
       apiKeys.map(omitToken),
       {
         populate: true,
@@ -215,17 +216,15 @@ export default class ApiKeyModuleService<TEntity extends ApiKey = ApiKey>
     config?: FindConfig<ApiKeyTypes.ApiKeyDTO>,
     sharedContext?: Context
   ): Promise<[ApiKeyTypes.ApiKeyDTO[], number]> {
-    const result = await this.apiKeyService_.listAndCount(
+    const [apiKeys, count] = await this.apiKeyService_.listAndCount(
       filters,
       config,
       sharedContext
     )
-    const withoutToken = result[0].map(omitToken)
-    const count = result[1]
 
     return [
       await this.baseRepository_.serialize<ApiKeyTypes.ApiKeyDTO[]>(
-        withoutToken,
+        apiKeys.map(omitToken),
         {
           populate: true,
         }
@@ -267,7 +266,7 @@ export default class ApiKeyModuleService<TEntity extends ApiKey = ApiKey>
     data: ApiKeyTypes.RevokeApiKeyDTO[],
     @MedusaContext() sharedContext: Context = {}
   ): Promise<TEntity[]> {
-    await this.validateRevokeApiKeys(data)
+    await this.validateRevokeApiKeys_(data)
 
     const updateRequest = data.map((k) => ({
       id: k.id,
@@ -292,7 +291,7 @@ export default class ApiKeyModuleService<TEntity extends ApiKey = ApiKey>
     return Promise.resolve(false)
   }
 
-  protected async validateCreateApiKeys(
+  protected async validateCreateApiKeys_(
     data: ApiKeyTypes.CreateApiKeyDTO[],
     sharedContext: Context = {}
   ): Promise<void> {
@@ -327,12 +326,19 @@ export default class ApiKeyModuleService<TEntity extends ApiKey = ApiKey>
     }
   }
 
-  protected async validateRevokeApiKeys(
+  protected async validateRevokeApiKeys_(
     data: ApiKeyTypes.RevokeApiKeyDTO[],
     sharedContext: Context = {}
   ): Promise<void> {
     if (!data.length) {
       return
+    }
+
+    if (data.some((k) => !k.id)) {
+      throw new MedusaError(
+        MedusaError.Types.INVALID_DATA,
+        `You must provide an api key id field when revoking a key.`
+      )
     }
 
     if (data.some((k) => !k.revoked_by)) {
@@ -387,9 +393,12 @@ export default class ApiKeyModuleService<TEntity extends ApiKey = ApiKey>
 }
 
 // We are mutating the object here as what microORM relies on non-enumerable fields for serialization, among other things.
-const omitToken = (key: ApiKey): ApiKey => {
+const omitToken = (
+  // We have to make salt optional before deleting it (and we do want it required in the DB)
+  key: Omit<ApiKey, "salt"> & { salt?: string }
+): Omit<ApiKey, "salt"> => {
   key.token = key.type === ApiKeyType.SECRET ? "" : key.token
-  key.salt = ""
+  delete key.salt
   return key
 }
 
--- a/packages/api-key/integration-tests/__tests__/api-key-module-service.spec.ts
+++ b/packages/api-key/integration-tests/__tests__/api-key-module-service.spec.ts
@@ -53,7 +53,7 @@ moduleIntegrationTestRunner({
             expect.objectContaining({
               title: "Test API Key",
               type: ApiKeyType.PUBLISHABLE,
-              salt: "",
+              salt: undefined,
               created_by: "test",
               last_used_at: null,
               revoked_by: null,
@@ -75,7 +75,7 @@ moduleIntegrationTestRunner({
             expect.objectContaining({
               title: "Secret key",
               type: ApiKeyType.SECRET,
-              salt: "44de31ebcf085fa423fc584aa8540670",
+              salt: undefined,
               created_by: "test",
               last_used_at: null,
               revoked_by: null,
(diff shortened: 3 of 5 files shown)


packages/api-key/src/services/api-key-module-service.ts
```diff
@@ -79,13 +86,42 @@ export default class ApiKeyModuleService<TEntity extends ApiKey = ApiKey>
...
+      })
+    }
 
     const createdApiKeys = await this.apiKeyService_.create(
-      data_,
+      normalizedInput,
       sharedContext
     )
 
+    // TODO: This seems to affect what is ultimately written in the DB, fix.
```
sradevski: @adrien2p I wanted to discuss this with you. Unfortunately, microORM depends on some non-enumerable fields from the object it returns for serialization, so we have to mutate here.

However, due to the unit of work approach it follows, mutating the data here means that what actually gets written in the DB is the raw token, and not the hashed token as expected.

What do you think the best approach would be here (and similarly, below in retrievals)? I thought of doing is:
1. Do the token reassigning after the serialization happens
2. Something else you are aware of? :D

packages/api-key/src/services/api-key-module-service.ts
```diff
@@ -117,22 +153,238 @@ export default class ApiKeyModuleService<TEntity extends ApiKey = ApiKey>
...
+
+    return {
+      rawToken: token,
+      hashedToken: token,
+      salt: "",
+      redacted: redactKey(token),
+    }
+  }
+
+  protected static async generateSecretKey(): Promise<TokenDTO> {
```
sradevski, line 381: I'd like people to pay more attention to this part, as it can have serious security implications. Let me know if you think we should make the token longer.

packages/api-key/src/services/api-key-module-service.ts
adrien2p: based on our discussion, the best approach here would be for the protected create method to return a map (or similar) with the api keys and the raw token and perform the assign after serialization

packages/api-key/src/models/api-key.ts
```diff
@@ -1,33 +1,68 @@
...
+  salt: string
+
+  @Property({ columnType: "text" })
+  redacted: string
+
+  @Property({ columnType: "text" })
+  title: string
+
+  @Property({ columnType: "text" })
+  @Enum({ items: ["publishable", "secret"] })
```
sradevski, line 45: I am not using the enum here as I think we should not couple the public interface with the DB model. What do others think?

packages/api-key/src/models/api-key.ts
```diff
@@ -1,33 +1,68 @@
...
+  columns: "type",
+})
+
 @Entity()
 export default class ApiKey {
   @PrimaryKey({ columnType: "text" })
   id: string
 
+  @Property({ columnType: "text" })
+  token: string
```
sradevski, line 33: I introduced a token field, so we don't rely on the ID being the token as well. This worked fine for publishable keys, but naturally it doesn't work for secret and restricted keys, so it's better for all of them to rely on a single column - token.

packages/api-key/src/models/api-key.ts
riqwan: Any particular downsides you think this will cause by using enums here?

The way I see it, these values will be used within application code quite tightly, if the enums in the db change, the application logic will be affected. The model is still part of the application logic.

olivermrbl: Fine by me. However, if the enums are the same in the public interface and the model, I guess it doesn't hurt to import it from utils and use it here?

packages/api-key/src/models/api-key.ts
```diff
@@ -1,33 +1,68 @@
...
+
+  @Property({ columnType: "text" })
+  salt: string
```
riqwan: q: Shouldn't these be unique?

packages/api-key/src/services/api-key-module-service.ts
```diff
@@ -65,28 +72,60 @@ export default class ApiKeyModuleService<TEntity extends ApiKey = ApiKey>
...
 
   @InjectTransactionManager("baseRepository_")
   protected async create_(
-    data: ApiKeyTypes.CreateApiKeyDTO | ApiKeyTypes.CreateApiKeyDTO[],
+    data: ApiKeyTypes.CreateApiKeyDTO[],
     @MedusaContext() sharedContext: Context = {}
-  ): Promise<TEntity | TEntity[]> {
-    const data_ = Array.isArray(data) ? data : [data]
+  ): Promise<[TEntity[], TokenDTO[]]> {
+    await this.validateCreateApiKeys(data, sharedContext)
```
riqwan: suggestion: I think we have a convention of calling protected methods with a `_` in the end
```suggestion
    await this.validateCreateApiKeys_(data, sharedContext)
```

packages/api-key/src/services/api-key-module-service.ts
```diff
@@ -103,36 +142,257 @@ export default class ApiKeyModuleService<TEntity extends ApiKey = ApiKey>
...
+      hashedToken: token,
+      salt: "",
+      redacted: redactKey(token),
+    }
+  }
+
+  protected static async generateSecretKey(): Promise<TokenDTO> {
+    const token = "sk_" + crypto.randomBytes(32).toString("hex")
+    const salt = crypto.randomBytes(16).toString("hex")
+    const hashed = ((await scrypt(token, salt, 64)) as Buffer).toString("hex")
```
riqwan, line 384: thought: The module could provide fairly standard defaults and could accept options to make this longer

packages/api-key/src/services/api-key-module-service.ts
```diff
@@ -103,36 +142,257 @@ export default class ApiKeyModuleService<TEntity extends ApiKey = ApiKey>
...
+      salt,
+      redacted: redactKey(token),
+    }
+  }
+}
+
+// We are mutating the object here as what microORM relies on non-enumerable fields for serialization, among other things.
+const omitToken = (key: ApiKey): ApiKey => {
+  key.token = key.type === ApiKeyType.SECRET ? "" : key.token
+  key.salt = ""
```
riqwan: suggestion: we could just delete these entirely from the response?

olivermrbl (approved):
LGTM, nice work

riqwan (approved):
Nice work! LGTM

packages/api-key/src/models/api-key.ts
sradevski: Well actually the salt barely matters in the case where the password (or in this case, a random key) is completely random, so not only uniqueness doesn't matter much, the salt itself has very little value (other than to follow good practices). So I think we don't have to enforce strict uniqueness on the DB level @riqwan

packages/api-key/src/services/api-key-module-service.ts
sradevski: Fine with that if we prefer it 👍

packages/api-key/src/services/api-key-module-service.ts
sradevski: For now let's keep it as it is (less options - less decisions to make), and if we see the need for it it's very easy to expand it to accept such options. Good point though 👍

packages/api-key/src/services/api-key-module-service.ts
sradevski: @riqwan I didn't want to make the `token` optional in a bunch of places, so I opted for setting an empty string. I can change that you feel strongly about it though, I am fine either way.

packages/api-key/src/models/api-key.ts
riqwan: the same applies for token as well?

packages/api-key/src/services/api-key-module-service.ts
riqwan: none are strongly held. :) Just suggestions.

packages/api-key/src/models/api-key.ts
sradevski: It really boils down to decoupling the database layer from the "API" layer. Accidentally, these two hold the same values, but we might want to rename them to something else externally (eg. `master` vs `secret`), but we won't necessarily want to change how the value is stored in the DB (a common scenario when you want to migrate data, and you want to do it one step at a time). In that case, you will mistakenly start saving `master` (since the enum will change) in the DB when you didn't really intend to do so, resulting in a bug.

If you are familiar with onion-layer architecture, this is the typical scenario where you want to decouple dependencies between your public-facing API and your database model

packages/api-key/src/models/api-key.ts
sradevski: @riqwan I guess we could have a unique index on the token, but it really is impossible to have a collision. I don't mind setting an index though, let's do it.

packages/api-key/src/services/api-key-module-service.ts
```diff
@@ -103,36 +142,257 @@ export default class ApiKeyModuleService<TEntity extends ApiKey = ApiKey>
     data: ApiKeyTypes.UpdateApiKeyDTO[] | ApiKeyTypes.UpdateApiKeyDTO,
     @MedusaContext() sharedContext: Context = {}
   ): Promise<ApiKeyTypes.ApiKeyDTO[] | ApiKeyTypes.ApiKeyDTO> {
```
adrien2p, line 145: q: looks like the DTO does not expect the salt to be returned, should we just remove it entirely? I join riqwan on that. Otherwise it feels weird that the salt is present but not expected to be, you might wonder if something wrong happen?

packages/api-key/src/services/api-key-module-service.ts
```diff
@@ -103,36 +142,257 @@ export default class ApiKeyModuleService<TEntity extends ApiKey = ApiKey>
...
+    config?: FindConfig<ApiKeyTypes.ApiKeyDTO>,
+    sharedContext?: Context
+  ): Promise<ApiKeyTypes.ApiKeyDTO[]> {
+    const apiKeys = await this.apiKeyService_.list(
+      filters,
+      config,
+      sharedContext
+    )
+
+    return this.baseRepository_.serialize<ApiKeyTypes.ApiKeyDTO[]>(
```
adrien2p: todo: missing await

packages/api-key/src/services/api-key-module-service.ts
```diff
@@ -103,36 +142,257 @@ export default class ApiKeyModuleService<TEntity extends ApiKey = ApiKey>
...
+
+    return [
+      await this.baseRepository_.serialize<ApiKeyTypes.ApiKeyDTO[]>(
+        withoutToken,
+        {
+          populate: true,
+        }
+      ),
+      count,
+    ]
```
adrien2p:
```suggestion
const [apiTokens, count] = await this.apiKeyService_.listAndCount(
      filters,
      config,
      sharedContext
    )
    const withoutToken = apiTokens.map(omitToken)

    return [
      await this.baseRepository_.serialize<ApiKeyTypes.ApiKeyDTO[]>(
        withoutToken,
        {
          populate: true,
        }
      ),
      count,
    ]
```

packages/api-key/src/services/api-key-module-service.ts
```diff
@@ -103,36 +142,257 @@ export default class ApiKeyModuleService<TEntity extends ApiKey = ApiKey>
...
+    if (data.some((k) => !k.revoked_by)) {
+      throw new MedusaError(
+        MedusaError.Types.INVALID_DATA,
+        `You must provide a revoked_by field when revoking a key.`
+      )
+    }
+
+    const revokedApiKeys = await this.apiKeyService_.list(
+      {
+        id: data.map((k) => k.id),
```
adrien2p, line 353: suggestion: I would extract that map into a const and filter(Boolean), then check if there is a length before ding the list. If the user do not provide the ids, we will list all secret keys

packages/api-key/src/services/api-key-module-service.ts
adrien2p: q: when the serialization have been applied, the salt will still be part of the response has an empty string, I would remove it entirely for the reason mentioned in my previous comment

packages/api-key/src/services/api-key-module-service.ts
```diff
@@ -103,36 +142,257 @@ export default class ApiKeyModuleService<TEntity extends ApiKey = ApiKey>
...
+}
+
+// We are mutating the object here as what microORM relies on non-enumerable fields for serialization, among other things.
+const omitToken = (key: ApiKey): ApiKey => {
+  key.token = key.type === ApiKeyType.SECRET ? "" : key.token
+  key.salt = ""
+  return key
+}
+
+const redactKey = (key: string): string => {
```
adrien2p, line 405: suggestion: static method of the class?

packages/api-key/src/services/api-key-module-service.ts
sradevski: It was convenient in the test, but I agree on deleting at least the salt 👍

packages/api-key/src/services/api-key-module-service.ts
sradevski: Well, it's not really required here but let's be consistent 👍

packages/api-key/src/services/api-key-module-service.ts
sradevski: @adrien2p well it's not really tied to the module in any way, it simply redacts a key, so I'd rather keep it as a pure function.

packages/api-key/src/services/api-key-module-service.ts
sradevski: @adrien2p wait what? Shouldn't an empty array of ids (or ids that are null/undefined) should result in 0 results? Let me write a quick test for it as well.

sradevski: Ok I confirmed, an empty array will not affect anything. It's fine to do what you said as an optimization however.

sradevski: Ok it will actually fail, so what we need to do is add validation and throw an error for invalid requests.

PR #6464, opened 2024-02-21
feat(tax): singular creates and deletes of regions, rates, rules

srindom:
**What**
- Adds support for creating single rates, regions, rules.
- Adds delete methods.


packages/tax/src/services/tax-module-service.ts
```diff
@@ -210,6 +278,41 @@ export default class TaxModuleService<
...
+    const parentIds = regionsToVerify.map((i) => i.parent_id).filter(isDefined)
+    if (parentIds.length > 0) {
+      const parentRegions = await this.taxRegionService_.list(
+        { id: { $in: parentIds } },
+        { select: ["id", "country_code"] },
+        sharedContext
+      )
+
+      for (const region of regionsToVerify) {
+        if (isDefined(region.parent_id)) {
```
riqwan, line 294: could return early here if not defined and save a level of nesting

riqwan (approved):
lgtm!

PR #6466, opened 2024-02-21
chore(utils): Move generic utils from region module to utils

sradevski:
Moves some generic utilities from the region module to the top-level utils package


packages/utils/src/common/remove-undefined.ts
```diff
@@ -0,0 +1,6 @@
+// useful in cases where presence of undefined is not desired (eg. in microORM operations)
```
carlos-r-l-rodrigues, line 1: should we create a `clone` util that only does `JSON.parse(JSON.stringify(obj))` and will have the same effect but also works with nested objects?

adrien2p: I do agree, it would work in most cases, at the moment, it only removes the top level undefined and skip the relation which either carlos approach would solve or you can use a recursive approach or alternatively a reduce approach

adrien2p (approved):
LGTM, comments to review

packages/utils/src/common/remove-undefined.ts
sradevski: @carlos-r-l-rodrigues now that you say it, we can definitely do that, I completely forgot that Json stringify will remove undefined, and I've even done it in the past 😄 It is less performant, but as you said, it does work on nested objects, so we can definitely do it like that. 

Now, even though it clones an object, I'd still prefer to call it `removeUndefined` as that is the purpose of the utility, even though it's achieved through "cloning". WDYT?

carlos-r-l-rodrigues: That works for me. The only downside of this approach is in case we need to remove `undefined`s of a complex object with functions and references. But if we need that in the future we can just do your implementation recursive.

sradevski: Yeah agreed, I think supporting at least nested objects is a nice to have 👍

PR #6467, opened 2024-02-21
feat(order): order change actions engine

carlos-r-l-rodrigues:
What:
 - Engine to apply order changes based on actions
 - Methods to managa Order, line items and shipping methods
 - Change Mikro-ORM select strategy to JOINED when there isn't pagination.

This is only the initial implementation of the action engine and it isn't yet integrated with the order changes the module will support.

### DIFF
--- a/packages/pricing/integration-tests/__tests__/services/pricing-module/money-amount.spec.ts
+++ b/packages/pricing/integration-tests/__tests__/services/pricing-module/money-amount.spec.ts
@@ -96,6 +96,7 @@ describe("PricingModule Service - MoneyAmount", () => {
       expect(serialized).toEqual([
         {
           id: "money-amount-USD",
+          amount: null,
           min_quantity: "1",
           currency_code: "USD",
         },


packages/order/src/utils/calculate-order-change.ts
```diff
@@ -0,0 +1,355 @@
...
+  constructor({
+    order,
+    transactions,
+    actions,
+  }: {
+    order: VirtualOrder
+    transactions: OrderTransaction[]
+    actions: InternalOrderChangeEvent[]
+  }) {
+    this.order = JSON.parse(JSON.stringify(order))
```
sradevski, line 43: Is the purpose here to remove the undefined fields? If so, we can use the util now

packages/order/src/utils/calculate-order-change.ts
```diff
@@ -0,0 +1,355 @@
...
+    this.transactions = JSON.parse(JSON.stringify(transactions ?? []))
+    this.actions = JSON.parse(JSON.stringify(actions ?? []))
+
+    const transactionTotal = transactions.reduce((acc, transaction) => {
```
sradevski, line 47: NIT: A `sum` util would be nice, as it's a pretty common operation

packages/order/src/utils/calculate-order-change.ts
```diff
@@ -0,0 +1,355 @@
...
+    }, 0)
+
+    this.summary = {
+      futureDifference: 0,
+      futureTemporaryDifference: 0,
+      temporaryDifference: 0,
+      pendingDifference: 0,
+      futureTemporarySum: 0,
+      differenceSum: 0,
+      currentOrderTotal: order.total as number,
```
sradevski, line 58: What is the `order.total` type?

packages/order/src/utils/calculate-order-change.ts
```diff
@@ -0,0 +1,355 @@
...
+    const status = action.status
+    return (
+      status === undefined ||
+      status === EVENT_STATUS.PENDING ||
+      status === EVENT_STATUS.DONE
+    )
+  }
+  private isEventDone(action: InternalOrderChangeEvent): boolean {
+    const status = action.status
+    return status === EVENT_STATUS.DONE
```
sradevski, line 74:
```suggestion
    return action.status === EVENT_STATUS.DONE
```

packages/order/src/utils/calculate-order-change.ts
```diff
@@ -0,0 +1,355 @@
...
+  }
+
+  private isEventPending(action: InternalOrderChangeEvent): boolean {
+    const status = action.status
+    return status === undefined || status === EVENT_STATUS.PENDING
```
sradevski, line 79: Shall we also check nulls, i.e `!status || status === EVENT_STATUS.PENDING`?

packages/order/src/utils/calculate-order-change.ts
```diff
@@ -0,0 +1,355 @@
...
+  }
+
+  public processActions(): void {
```
sradevski: NIT: No need to specify the return type here

packages/order/src/utils/calculate-order-change.ts
```diff
@@ -0,0 +1,355 @@
...
+    for (const action of this.actions) {
+      this.processAction_(action)
+    }
+
+    const summary = this.summary
+    for (let idx = 0; idx < this.actions.length; idx++) {
+      const action = this.actions[idx]
```
sradevski:
```suggestion
    for (let action of this.actions){
```

packages/order/src/utils/calculate-order-change.ts
```diff
@@ -0,0 +1,355 @@
...
+        summary.futureDifference += amount
+      } else {
+        if (!this.isEventDone(action) && !action.group_id) {
+          summary.differenceSum += amount
+        }
+        summary.currentOrderTotal += amount
+      }
+    }
+
+    const groupSum = Object.values(this.groupTotal).reduce((acc, amount) => {
```
sradevski, line 123: Ah, few more instances of summing 😄

packages/order/src/utils/calculate-order-change.ts
```diff
@@ -0,0 +1,355 @@
...
+    }, 0)
+
+    summary.differenceSum += groupSum
+
+    summary.transactionTotal = this.transactions.reduce((acc, transaction) => {
+      return acc + transaction.amount
+    }, 0)
+
+    summary.futureTemporaryDifference =
+      -summary.futureTemporarySum + summary.futureDifference
```
sradevski: NIT: Inverting the operands makes it a bit easier to read (futureDifference - futureTemporarySum)

packages/order/src/utils/calculate-order-change.ts
```diff
@@ -0,0 +1,355 @@
...
+
+    if (!isReplay) {
+      this.actionsProcessed[action.action].push(action)
+    }
+
+    let previousEvents: InternalOrderChangeEvent[] | undefined
+    if (type.commitsAction) {
+      previousEvents = (this.actionsProcessed[type.commitsAction] ?? []).filter(
+        (action) =>
+          action.reference_id === action.reference_id &&
```
sradevski: I imagine this check should have been something else

packages/order/src/utils/calculate-order-change.ts
```diff
@@ -0,0 +1,355 @@
...
+          action.status !== EVENT_STATUS.VOIDED
+      )
+    }
+
+    let calculatedAmount: number = action.amount ?? 0
```
sradevski, line 167: Is the `: number` typing necessary here?

packages/order/src/utils/actions/shipping-add.ts
```diff
@@ -0,0 +1,47 @@
...
+    const shipping = Array.isArray(currentOrder.shipping_methods)
+      ? currentOrder.shipping_methods
+      : [currentOrder.shipping_methods]
+
+    const existingIndex = shipping.findIndex(
+      (item) => item.id === action.reference_id
+    )
+
+    if (existingIndex > -1) {
+      shipping.splice(existingIndex, 1)
```
sradevski, line 29: Should we return or do something with the result here? Also, maybe we can stick to non-mutating methods when possible

packages/order/src/utils/calculate-order-change.ts
carlos-r-l-rodrigues: Here the purpose is to clone the originals and not mutate them outside this scope

packages/order/src/utils/calculate-order-change.ts
carlos-r-l-rodrigues: they will be big numbers. The calculations are only using number for now.
will change that in a next PR

packages/order/integration-tests/__tests__/create-order.ts
```diff
@@ -0,0 +1,210 @@
...
+              expect.objectContaining({
+                id: expect.stringContaining("orddetail_"),
+                version: 1,
+                item: expect.objectContaining({
+                  id: expect.stringContaining("ordli_"),
+                  tax_lines: [],
+                  adjustments: [],
+                }),
+              }),
+            ],
```
srindom: I really don't like this structure - specifically:
- `items.item` nesting is not pretty
- it doesn't follow the same pattern as cart
- it's not intuitive what the point of "orddetail_" is, which will confuse an external developer

Two options I'd be fine with:

1. invert how orddetail_ is stored - i.e. something like this

```
item: [
  {
    id: "ordli_12313",
    tax_lines: [..],
    adjustments: [...],
    ...,
    item_detail: {
      id: "orddetail_123132",
      version: 1
    }
  }
]
```

2. Create an `item_details` field on the order item that can store in the current format. But then keep `items` for pure items.

```
item: [
  {
    id: "ordli_12313",
    tax_lines: [..],
    adjustments: [...],
    ...,
  }
],
item_details: [
  {
    id: "orddetail_123132",
    version: 1,
    item: { ... }
  }
]
```

srindom: Also, what is the idea with order detail's many to one relationship with items?

carlos-r-l-rodrigues: Order details will handle versioning and all the totals of the item for that specific version.
I will transform it before returning, where do you think these quantities like `return_requested_quantity`, `shipped_quantity` should be placed?
also, there is the metadata inside these details where custom quantities/values could be placed.

can it be something like this?
```
{
  order: {
    items: [
      {
        id: "ordli_12313",
        tax_lines: [...],
        adjustments: [...],
        detail: {
          version: 1,
          shipped_quantity: 1,
          ...,
        }
      }
    ]
  }
}
```

carlos-r-l-rodrigues: I will rename it to `order_item` internally to make more sense. `order_detail` is not really a good name for what it does

srindom: Think this is perfect 👍 

> can it be something like this?
> ```
> {
>  order: {
>    items: [
>      {
>        id: "ordli_12313",
>        tax_lines: [...],
>        adjustments: [...],
>        detail: {
>          version: 1,
>          shipped_quantity: 1,
>          ...,
>        }
>      }
>    ]
>  }
>}
>```

srindom:
@carlos-r-l-rodrigues - is there more work going into this PR or can it be reviewed in the current state?

packages/order/src/models/shipping-method.ts
```diff
@@ -31,23 +30,41 @@ const OrderIdIndex = createPsqlIndexStatementHelper({
...
+    cascade: [Cascade.REMOVE],
+  })
   @OrderIdIndex.MikroORMIndex()
   order_id: string
 
   @ManyToOne({
     entity: () => Order,
     fieldName: "order_id",
-    cascade: [Cascade.REMOVE, Cascade.PERSIST],
+    cascade: [Cascade.REMOVE],
```
srindom, line 57: suggestion: `onDelete: "cascade"`

packages/order/src/models/shipping-method.ts
```diff
@@ -31,23 +30,41 @@ const OrderIdIndex = createPsqlIndexStatementHelper({
...
   @PrimaryKey({ columnType: "text" })
   id: string
 
-  @Property({ columnType: "text" })
+  @ManyToOne({
+    entity: () => Order,
+    columnType: "text",
+    fieldName: "order_id",
+    mapToPk: true,
+    cascade: [Cascade.REMOVE],
```
srindom, line 49: suggestion: `onDelete: "cascade"`

packages/order/src/models/transaction.ts
```diff
@@ -41,14 +41,18 @@ export default class Transaction {
...
-  @Property({ columnType: "text" })
-  @OrderIdIndex.MikroORMIndex()
-  order_id: string
-
   @ManyToOne({
     entity: () => Order,
+    columnType: "text",
     fieldName: "order_id",
-    cascade: [Cascade.REMOVE, Cascade.PERSIST],
+    cascade: [Cascade.REMOVE],
```
srindom, line 48: suggestion: `onDelete: "cascade"`

packages/order/integration-tests/__tests__/create-order.ts
```diff
@@ -0,0 +1,243 @@
...
+            "shipping_methods",
+            "shipping_methods.tax_lines",
+            "shipping_methods.adjustments",
+          ],
+        })
+
+        expect(getOrder).toEqual(expectation)
+      })
+
+      it.skip("should transform where clause to match the db schema and return the order", async function () {
```
srindom, line 226: q: is this skip still necessary?

packages/order/src/models/order-item.ts
```diff
@@ -1,59 +1,75 @@
...
+  columns: ["version"],
+})
+
+const ItemIdIndex = createPsqlIndexStatementHelper({
+  tableName: "order_item",
+  columns: ["item_id"],
+})
+
+@Entity({ tableName: "order_item" })
+export default class OrderItem {
```
srindom, line 38: thought(non-blocking): maybe `OrderItemDetail` or similar would be better `OrderItem` is very close to `OrderLineItem`

packages/order/src/models/shipping-method-adjustment.ts
```diff
@@ -20,24 +19,30 @@ const ShippingMethodIdIdIndex = createPsqlIndexStatementHelper({
...
+  })
+  shipping_method: ShippingMethod
+
   @ManyToOne({
     entity: () => ShippingMethod,
+    columnType: "text",
     fieldName: "shipping_method_id",
-    cascade: [Cascade.REMOVE, Cascade.PERSIST],
+    mapToPk: true,
+    cascade: [Cascade.REMOVE],
```
srindom, line 32: suggestion: `onDelete: "cascade"`

packages/order/src/models/shipping-method-tax-line.ts
```diff
@@ -20,24 +19,30 @@ const ShippingMethodIdIdIndex = createPsqlIndexStatementHelper({
...
+  })
+  shipping_method: ShippingMethod
+
   @ManyToOne({
     entity: () => ShippingMethod,
     fieldName: "shipping_method_id",
-    cascade: [Cascade.REMOVE, Cascade.PERSIST],
+    columnType: "text",
+    mapToPk: true,
+    cascade: [Cascade.REMOVE],
```
srindom, line 32: suggestion: `onDelete: "cascade"`

packages/order/src/repositories/order.ts
srindom, line 1: thought(non-blocking): could materialized views in postgres make sense instead of this and some of the work in `./utils/transform-order`?

carlos-r-l-rodrigues:
> @carlos-r-l-rodrigues - is there more work going into this PR or can it be reviewed in the current state?

@srindom - it can be reviewed as it is. All the actions thing will very likley change a lot when I actually integrate it with the order edits based on actions.

packages/order/src/utils/actions/receive-damaged-return-item.ts
```diff
@@ -0,0 +1,88 @@
...
+          toReturn -= ret
+
+          previousEvent.details.quantity -= ret
+          if (previousEvent.details.quantity <= 0) {
+            previousEvent.status = EVENT_STATUS.DONE
+          }
+        }
+      }
+
+      return existing.unit_price * action.details.quantity
```
srindom, line 36: note: these calculations will have to be done with some sort of `unit_total` that takes discounts and taxes into account as well. Doesn't have to go into this PR but worth keeping in mind.

packages/order/src/utils/transform-order.ts
srindom, line 1: see comment in repositories/order

packages/payment/integration-tests/__tests__/services/payment-module/index.spec.ts
```diff
@@ -472,25 +472,7 @@ moduleIntegrationTestRunner({
...
-                    captured_at: null,
-                    canceled_at: null,
-                    refunds: [],
-                    captures: [],
-                    amount: 100,
-                    currency_code: "usd",
-                    provider_id: "pp_system_default",
-                  }),
-                },
+                }),
```
srindom, line 475: q: what is this about?

packages/pricing/integration-tests/__tests__/services/pricing-module/money-amount.spec.ts
srindom, line 1: some of these files seem unrelated to Order?

packages/product/integration-tests/__tests__/services/product-variant/index.ts
```diff
@@ -1,22 +1,23 @@
...
 } from "../../../__fixtures__/product"
 import { productsData, variantsData } from "../../../__fixtures__/product/data"
 import { buildProductVariantOnlyData } from "../../../__fixtures__/variant/data/create-variant"
-import { createMedusaContainer } from "@medusajs/utils"
-import { asValue } from "awilix"
-import ContainerLoader from "../../../../src/loaders/container"
+import { TestDatabase } from "../../../utils"
 
-describe("ProductVariant Service", () => {
+// TODO: fix tests
```
srindom, line 19: note: seems unrelated to order, but just making sure you are aware

srindom (approved):
A bit hard to review, but think it's okay to merge and then move to smaller PRs for following work.

packages/payment/integration-tests/__tests__/services/payment-module/index.spec.ts
carlos-r-l-rodrigues: All these other files are fixes because of changing the mikro-orm strategy to JOINED. That should be another PR and I'm to blame for it.

packages/order/src/utils/actions/receive-damaged-return-item.ts
carlos-r-l-rodrigues: Yeah, here is not the only place where this will happen. I believe we'll have a util shared with multiple modules that can calculate theses things.
I will revisit that anyway when calculating using BigNumbers

packages/order/integration-tests/__tests__/create-order.ts
carlos-r-l-rodrigues: in this PR yes. it's fixed but will put it in another one.

packages/order/src/repositories/order.ts
carlos-r-l-rodrigues: I will keep that in mind when order versioning is more mature I can think if we can simplify the queries internally reshaping things in materialized views for reading.

packages/order/src/utils/actions/receive-damaged-return-item.ts
srindom: Yep sounds good

PR #6468, opened 2024-02-21
fix(workflows-sdk): Fix StepFunction typings and custom step name

adrien2p:
**What**
- Fix step function typings
- Fix step name when custom config is passed


packages/workflows-sdk/package.json
```diff
@@ -40,6 +40,6 @@
     "build": "rimraf dist && tsc --build",
     "watch": "tsc --build --watch",
     "test": "jest --runInBand --bail --forceExit",
-    "test:run": "../../node_modules/.bin/ts-node ./src/utils/_test.ts"
+    "test:run": "../../node_modules/.bin/ts-node ./src/utils/_playground.ts"
```
olivermrbl, line 43: question: intentional?

adrien2p:
@olivermrbl should we merge?

olivermrbl (approved):

packages/workflows-sdk/package.json
adrien2p: yes _test does not exists and was replace bu this one

PR #6469, opened 2024-02-22
feat: Refresh invite

pKorsholm:


packages/types/src/user/service.ts
```diff
@@ -5,18 +5,23 @@ import {
...
   validateInviteToken(
     token: string,
     sharedContext?: Context
   ): Promise<InviteDTO>
 
+  resendInvites(
+    inviteIds: string[],
+    sharedContext?: Context
+  ): Promise<InviteDTO[]>
+
```
srindom: thought: not sure that this should live in the user module. 

"Sending" an invite is an opinionated way of sharing an invite token with someone. But the UserModule should not really care or even know about the possible methods of transfer between the application and the one needing the invite. 

I would instead move all of the logic related to "sending" to a workflow that simply retrieves the token from the user module and pushes it to whatever "sending" logic has been configured.

srindom: We could still have a function in the UserService that can refresh an invite token, but what it is being used for afterwards should not be the module's concern.

srindom: When reading the implementation this seems to be what is happening so maybe it's just a question of renaming this method to `refreshInvite`

srindom (commented):
Suggest we rename the token

packages/user/src/services/invite.ts
```diff
@@ -109,8 +152,10 @@ export default class InviteService<
 
     return jwt.sign(data, jwtSecret, {
       expiresIn,
+      jwtid: crypto.randomUUID(),
```
olivermrbl, line 167: question: Why was this added?

olivermrbl (approved):
LGTM!

packages/user/src/services/invite.ts
```diff
@@ -84,6 +87,46 @@ export default class InviteService<
...
+    context: Context = {}
+  ): Promise<TEntity[]> {
+    const [invites, count] = await super.listAndCount(
+      { id: inviteIds },
+      {},
+      context
+    )
+
+    if (count !== inviteIds.length) {
+      const invitesInDbSet = new Set(invites.map((invite) => invite.id))
```
sradevski: You can use `arrayDifference` util here I believe

integration-tests/plugins/__tests__/invites/resend-invite.spec.ts
```diff
@@ -0,0 +1,67 @@
...
+
+    const api = useApi()! as AxiosInstance
+
+    const response = await api.post(
+      `/admin/invites/${invite.id}/resend`,
+      {},
+      adminHeaders
+    )
+
+    expect(response.status).toEqual(200)
```
sradevski, line 61: I would be nice to test that the event was also emitted, as that is probably more important here for now

packages/user/src/services/invite.ts
pKorsholm: adding a jwtid ensures that the token is actually regenerated instead of the same token being issued for the same data. We could also include the expiry_date to achieve this, but an id like this seemed better

olivermrbl: Got it, thanks. Makes sense. Feel free to merge as you please :)

integration-tests/plugins/__tests__/invites/resend-invite.spec.ts
pKorsholm: Added a test in the integration test suite that the event is emitted

PR #6471, opened 2024-02-22
feat(api-key): Add the endpoints and workflows for api key module

sradevski:
Added endpoints for managing API keys, as well as the workflows and related necessary changes.

### DIFF
--- a/packages/api-key/integration-tests/__tests__/api-key-module-service.spec.ts
+++ b/packages/api-key/integration-tests/__tests__/api-key-module-service.spec.ts
@@ -105,10 +105,12 @@ moduleIntegrationTestRunner({
 
         it("should allow for at most two tokens, where one is revoked", async function () {
           const firstApiKey = await service.create(createSecretKeyFixture)
-          await service.revoke({
-            id: firstApiKey.id,
-            revoked_by: "test",
-          })
+          await service.revoke(
+            { id: firstApiKey.id },
+            {
+              revoked_by: "test",
+            }
+          )
 
           await service.create(createSecretKeyFixture)
           const err = await service
@@ -123,8 +125,7 @@ moduleIntegrationTestRunner({
       describe("revoking API keys", () => {
         it("should have the revoked at and revoked by set when a key is revoked", async function () {
           const firstApiKey = await service.create(createSecretKeyFixture)
-          const revokedKey = await service.revoke({
-            id: firstApiKey.id,
+          const revokedKey = await service.revoke(firstApiKey.id, {
             revoked_by: "test",
           })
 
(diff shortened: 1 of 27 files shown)


packages/api-key/integration-tests/__tests__/api-key-module-service.spec.ts
```diff
@@ -180,8 +178,7 @@ moduleIntegrationTestRunner({
         it("should not reflect any updates on other fields", async function () {
           const createdApiKey = await service.create(createSecretKeyFixture)
 
-          const updatedApiKey = await service.update({
-            id: createdApiKey.id,
+          const updatedApiKey = await service.update(createdApiKey.id, {
```
adrien2p, line 181: q: why did we move the id here?

packages/types/src/api-key/service.ts
```diff
@@ -15,11 +15,31 @@ export interface IApiKeyModuleService extends IModuleService {
...
+  update(
+    id: string,
+    data: Omit<UpdateApiKeyDTO, "id">,
+    sharedContext?: Context
+  ): Promise<ApiKeyDTO>
+  /**
+   * Update an api key
+   * @param data
+   */
+  update(data: UpdateApiKeyDTO[]): Promise<ApiKeyDTO[]>
```
adrien2p, line 42: q: missing shared context, also, is the selector based approach needed here?

adrien2p: also, it would be better to have the shared context at always the same position as the index of the shared context is stored on the class through the decorator and then used to identify where to put it. Here the second argument of the method could then be data or shared context and can create issues. Maybe an huddle would be easier if you want to discuss it

adrien2p (approved):
Very good 🔥 a couple of comments to look at 👍

packages/api-key/integration-tests/__tests__/api-key-module-service.spec.ts
sradevski: @adrien2p I just followed the API pattern of the region, as I assumed that's how we want all modules to behave

packages/types/src/api-key/service.ts
sradevski: @adrien2p similar to above, I just followed the API pattern of the region, as I assumed that's how we want all modules to behave.

I think we just need to standardize how all modules behave basically, as I am not sure what the "expected" public API for each module is. 

I would suggest we do the discussion separately from the PR, and we go with this one as it is, and then we modify all modules at once that don't follow the convention.

adrien2p: I do agree on the standardization for sure 👍 though, you will have an issue with the context coming from a workflow for example. This should be handled now, @olivermrbl you also have the issue in the region I conclude. so it should be handled too cc @carlos-r-l-rodrigues

sradevski: @adrien2p let me merge this and I'll update it in the follow-up PR where I enforce the usage of the authentication, since this already built and passed 😄

PR #6473, opened 2024-02-22
feat(utils): Fix big number decorator and cleanup

adrien2p:
**What**
- Fix big number decorator and cleanup


riqwan (approved):
Much more cleaner. 🔥 +1 on adding mikroorm prefix

packages/utils/src/dal/mikro-orm/big-number-field.ts
```diff
@@ -1,78 +1,46 @@
...
+    const targetColumn = columnName + "_"
+    const rawColumnName = options.rawColumnName ?? `raw_${columnName}`
+
+    Object.defineProperty(target, columnName, {
+      get() {
+        if (!isDefined(this[rawColumnName]) && !isDefined(this[targetColumn])) {
+          return null
+        }
+
+        return this[targetColumn] instanceof BigNumber
```
fPolic: question: aren't we setting `this[targetColumn]` on L:28 as a number so this check will never pass i.e. we will always call BigNumber constructor

adrien2p: should be all good man 👍

packages/utils/src/dal/mikro-orm/big-number-field.ts
```diff
@@ -1,78 +1,54 @@
...
-      return
-    }
-
-    const val = new BigNumber(rawValue ?? value)
-    entity[prop] = val.numeric
-    entity[`raw_${prop}`] = val.raw
+import { Property } from "@mikro-orm/core"
+import { BigNumberInput } from "@medusajs/types"
+
+export function MikroOrmBigNumberProperty(
```
sradevski, line 5: Since this is already under `mikro-orm` folder, I think naming it `BigNumberProperty` is sufficient, WDYT? It will make it a bit nicer to read, and the prefix feels unnecessary.

carlos-r-l-rodrigues (approved):

sradevski (commented):
Let's make a final decision internally on how to handle money and big numbers before we merge this

packages/utils/src/dal/mikro-orm/big-number-field.ts
adrien2p: I do agree, though, when you import it, you import from utils, not necessarily knowing where it is located in the package ☺️ so i preferred to be extra clear

adrien2p:
> Let's make a final decision internally on how to handle money and big numbers before we merge this

This pr is meant to fix current issues, i think it could still be merged to resolve those issues in the different other pr's and we can come back to it when we settled? No strong opinion though

packages/utils/src/dal/mikro-orm/big-number-field.ts
riqwan: Just for added context, the model/repository layer is designed to be completely replaceable. Which is why you'll see a lot of places have a MikroORM prefix.

riqwan:
+1 this fixes an issue I have in my PR. We can still do a followup after we've aligned.

packages/payment/src/models/payment-session.ts
```diff
@@ -22,13 +28,15 @@ export default class PaymentSession {
   id: string
 
   @Property({ columnType: "text" })
-  currency_code: string
+  currency_code: string | number
```
olivermrbl: question: Why have we added `number` here?

adrien2p: My mistake, completely involuntary

olivermrbl:
I would also be OK with merging this now. It is blocking a few other PRs.

PR #6474, opened 2024-02-22
feat(workflows-sdk,core-flows,medusa,types): add workflow to update promotions to cart

riqwan:
what:

- adds API + workflow to add/remove promotions in a cart
- minor fixes in promotions module
- minor type fixes in cart module
- typing fix in workflows-sdk (Thanks @adrien2p)
- fix step result in workflows-sdk (Thanks @adrien2p)

RESOLVES CORE-1768


packages/workflows-sdk/src/utils/composer/create-step.ts
```diff
@@ -202,6 +202,8 @@ function applyStep<
         ret.__step__ = newStepName
         WorkflowManager.update(this.workflowId, this.flow, this.handlers)
 
+        ret.__step__ = newStepName
```
sradevski: Why is this needed again?

packages/promotion/src/utils/compute-actions/shipping-methods.ts
```diff
@@ -95,14 +99,18 @@ export function applyPromotionToShippingMethods(
     const totalApplicableValue = shippingMethods!.reduce((acc, method) => {
       const appliedPromoValue = methodIdPromoValueMap.get(method.id) || 0
 
-      return acc + method.subtotal - appliedPromoValue
+      return acc + (method.subtotal || 0) - appliedPromoValue
```
sradevski: I know it doesn't matter in this case, but it's nice practice to rely on `??` instead :)

packages/promotion/src/utils/compute-actions/items.ts
```diff
@@ -112,14 +117,18 @@ export function applyPromotionToItems(
...
-      return (
-        acc +
-        (method.subtotal / method.quantity) * method.quantity -
-        appliedPromoValue
-      )
+      const perItemCost = method.subtotal
+        ? method.subtotal / method.quantity
+        : 0
+
+      return acc + perItemCost * method.quantity - appliedPromoValue
```
sradevski, line 124: This is one of the scenarios where you'd want to do some rounding (or maybe higher-up at the caller of this function, but we'll need to round somewhere)

integration-tests/plugins/__tests__/cart/store/add-promotions-to-cart.spec.ts
```diff
@@ -0,0 +1,309 @@
...
+              operator: "eq",
+              values: "prod_mat",
+            },
+          ],
+        },
+      })
+
+      const cart = await cartModuleService.create({
+        currency_code: "usd",
+        items: [
```
sradevski, line 91: Maybe we can create some fixtures that you can reuse, esp. in tests where the creation is just a matter of preparing for the actual test.

packages/workflows-sdk/src/utils/composer/create-step.ts
adrien2p: this how we identified the relation between the steps and the data, where they are coming from etc, here with a custom name the step name wasnt replaced leading to issues. In a workflow composer, even though through the typing we tried to give the idea that you are manipulating the data++ in fact they are only object representation of the flow and relations between the steps

riqwan: @adrien2p seems like you already added it in another PR line:202

packages/promotion/src/utils/compute-actions/items.ts
riqwan: Correct, we will be removing this from the promotion module entirely once we have a totals utility, which should take care of the precision and rounding issues more generally.

packages/workflows-sdk/src/utils/composer/create-step.ts
riqwan: removed this, good catch!

integration-tests/plugins/__tests__/cart/store/add-promotions-to-cart.spec.ts
```diff
@@ -0,0 +1,309 @@
...
+              adjustments: expect.arrayContaining([
+                expect.objectContaining({
+                  promotion_id: createdPromotion.id,
+                  code: createdPromotion.code,
+                  amount: 1000,
+                }),
+              ]),
+            }),
+            expect.objectContaining({
+              id: expect.not.stringContaining(lineItemAdjustment.id),
```
olivermrbl: question: Isn't this the wrong id to assert on? You are testing for `items` here but use a line item adjustment ID, so I think this is a false positive no?

packages/core-flows/src/definition/cart/steps/create-line-item-adjustments.ts
```diff
@@ -0,0 +1,41 @@
...
+  async (createdLineItemAdjustments, { container }) => {
+    const cartModuleService: ICartModuleService = container.resolve(
+      ModuleRegistrationName.CART
+    )
+
+    if (!createdLineItemAdjustments?.length) {
+      return
+    }
+
+    await cartModuleService.removeLineItemAdjustments(
```
olivermrbl: comment: Will be changed to `softDeleteLineItemAdjustments` in #6491

packages/core-flows/src/definition/cart/steps/create-shipping-method-adjustments.ts
```diff
@@ -0,0 +1,42 @@
...
+  async (createdShippingMethodAdjustments, { container }) => {
+    const cartModuleService: ICartModuleService = container.resolve(
+      ModuleRegistrationName.CART
+    )
+
+    if (!createdShippingMethodAdjustments?.length) {
+      return
+    }
+
+    await cartModuleService.removeShippingMethodAdjustments(
```
olivermrbl: Same as other comment.

packages/core-flows/src/definition/cart/steps/get-actions-to-compute-from-promotions.ts
```diff
@@ -0,0 +1,54 @@
...
+  cart: CartDTO
+  promoCodes: string[]
+  removePromotions: boolean
+}
+
+export const getActionsToComputeFromPromotionsStepId =
+  "get-actions-to-compute-from-promotions"
+export const getActionsToComputeFromPromotionsStep = createStep(
+  getActionsToComputeFromPromotionsStepId,
+  async (data: StepInput, { container }) => {
```
olivermrbl, line 16: thought(non-blocking): Would it make sense for this step to be even more generic i.e. just accept `codes` and return the computed actions? All the data extraction can be performed in transformers. Wdyt?

packages/core-flows/src/definition/cart/steps/retrieve-cart.ts
```diff
@@ -0,0 +1,43 @@
...
+    const cartModuleService = container.resolve<ICartModuleService>(
+      ModuleRegistrationName.CART
+    )
+
+    const cart = await cartModuleService.retrieve(data.cartId, {
+      relations: [
+        "items",
+        "items.adjustments",
+        "shipping_methods",
+        "shipping_methods.adjustments",
```
olivermrbl: todo: Remove in favor of `...data.config`

packages/core-flows/src/definition/cart/steps/retrieve-cart.ts
```diff
@@ -0,0 +1,43 @@
+import { ModuleRegistrationName } from "@medusajs/modules-sdk"
+import { ICartModuleService } from "@medusajs/types"
+import { StepResponse, createStep } from "@medusajs/workflows-sdk"
+
+interface StepInput {
+  cartId: string
```
olivermrbl, line 6: todo: Add config to increase reusability of the step
```suggestion
  cartId: string
  config: FindConfig<CartDTO>
```

packages/core-flows/src/definition/cart/workflows/update-cart-promotions.ts
```diff
@@ -0,0 +1,57 @@
...
+  promoCodes: string[]
+  cartId: string
+  removePromotions?: boolean
+}
+
+export const updateCartPromotionsWorkflowId = "update-cart-promotions"
+export const updateCartPromotionsWorkflow = createWorkflow(
+  updateCartPromotionsWorkflowId,
+  (input: WorkflowData<WorkflowInput>): WorkflowData<CartDTO> => {
+    const cart = retrieveCartStep(input)
```
olivermrbl: todo: Use config
```suggestion
    const cart = retrieveCartStep(input, { relations: [
       "items",
        "items.adjustments",
        "shipping_methods",
        "shipping_methods.adjustments",
    ] })
```

packages/medusa/src/api-v2/store/carts/[id]/promotions/route.ts
```diff
@@ -0,0 +1,42 @@
+import { updateCartPromotionsWorkflow } from "@medusajs/core-flows"
+import { MedusaRequest, MedusaResponse } from "../../../../../types/routing"
+import { StorePostCartsCartPromotionsReq } from "../../validators"
+
+export const POST = async (req: MedusaRequest, res: MedusaResponse) => {
+  const workflow = updateCartPromotionsWorkflow(req.scope)
+  const requestParams = req.validatedBody as StorePostCartsCartPromotionsReq
```
Comment (olivermrbl) [inline packages/medusa/src/api-v2/store/carts/[id]/promotions/route.ts]: todo: Rename
```suggestion
  const payload = req.validatedBody as StorePostCartsCartPromotionsReq
```

packages/medusa/src/api-v2/store/carts/[id]/promotions/route.ts
```diff
@@ -0,0 +1,42 @@
...
+
+  const { result, errors } = await workflow.run({
+    input: {
+      promoCodes: requestParams.promo_codes,
```
Comment (olivermrbl) [inline packages/medusa/src/api-v2/store/carts/[id]/promotions/route.ts]:
```suggestion
      promoCodes: payload.promo_codes,
```

packages/medusa/src/api-v2/store/carts/[id]/promotions/route.ts
```diff
@@ -0,0 +1,42 @@
...
+  if (Array.isArray(errors) && errors[0]) {
+    throw errors[0].error
+  }
+
+  res.status(200).json({ cart: result })
+}
+
+export const DELETE = async (req: MedusaRequest, res: MedusaResponse) => {
+  const workflow = updateCartPromotionsWorkflow(req.scope)
+  const requestParams = req.validatedBody as StorePostCartsCartPromotionsReq
```
Comment (olivermrbl) [inline packages/medusa/src/api-v2/store/carts/[id]/promotions/route.ts]: todo: Rename
```suggestion
  const payload = req.validatedBody as StorePostCartsCartPromotionsReq
```

packages/medusa/src/api-v2/store/carts/[id]/promotions/route.ts
```diff
@@ -0,0 +1,42 @@
...
+
+  const { result, errors } = await workflow.run({
+    input: {
+      promoCodes: requestParams.promo_codes,
```
Comment (olivermrbl) [inline packages/medusa/src/api-v2/store/carts/[id]/promotions/route.ts]:
```suggestion
      promoCodes: payload.promo_codes,
```

packages/medusa/src/api-v2/store/carts/middlewares.ts
```diff
@@ -38,4 +39,9 @@ export const storeCartRoutesMiddlewares: MiddlewareRoute[] = [
     matcher: "/store/carts/:id",
     middlewares: [transformBody(StorePostCartsCartReq)],
   },
+  {
+    method: ["POST", "DELETE"],
+    matcher: "/store/carts/:id/promotions",
+    middlewares: [transformBody(StorePostCartsCartPromotionsReq)],
```
olivermrbl, line 46: suggestion(non-blocking): Create separate type for `DELETE`

olivermrbl (changes requested):
Nice work. Left a couple of comments and todos.

integration-tests/plugins/__tests__/cart/store/add-promotions-to-cart.spec.ts
riqwan: ah yes, good catch. This should go under adjustments

packages/core-flows/src/definition/cart/steps/get-actions-to-compute-from-promotions.ts
riqwan: Can you elaborate on this one? I don't fully understand.

For us to be able to generate actions to compute, we need the full cart along with the promo codes.

we also need a third parameter - `action`, which can be:

- `add`: adds new promo codes along with already applied promo codes in the cart (case: add promo code to cart)
- `remove`: removes only one promo code from the cart, keeps all existing promo codes (case: remove promo code from cart)
- `replace`: replace the given promo codes, i.e. removes all existing ones and add only the ones in this given promo code array (case: update cart)

The promo codes to pass into the compute actions will vary depending on each of these actions.

packages/core-flows/src/definition/cart/steps/retrieve-cart.ts
riqwan: nice!

packages/core-flows/src/definition/cart/steps/get-actions-to-compute-from-promotions.ts
olivermrbl: Yes, I am not sure what I was thinking actually. I take it back :D

olivermrbl (approved):
LGTM, nice work!

PR #6475, opened 2024-02-22
chore(cart): Make all cart entities soft-deletable

olivermrbl:


riqwan (approved):
just in time, lgtm!

adrien2p:
Dont we need cascade on soft delete? Maybe not but just wanted to check

olivermrbl:
> Dont we need cascade on soft delete? Maybe not but just wanted to check

I followed the approach taken in the fulfillment module. Did I miss something? 🤔

adrien2p:
> > Dont we need cascade on soft delete? Maybe not but just wanted to check
> 
> 
> 
> I followed the approach taken in the fulfillment module. Did I miss something? 🤔 

There is a task on linear about cascade etc for the fulfilment, it is not finished yet on that part until that ticket is being resolved. But if you want the soft deletion to cascade on the relations it needs to be specified

olivermrbl:
> There is a task on linear about cascade etc for the fulfilment, it is not finished yet on that part until that ticket is being resolved. But if you want the soft deletion to cascade on the relations it needs to be specified

Ah sorry. I wasn't aware of this.

packages/cart/src/models/adjustment-line.ts
```diff
@@ -16,12 +16,6 @@ export default abstract class AdjustmentLine {
   @Property({ columnType: "text", nullable: true })
   description: string | null = null
 
-  @Property({
-    columnType: "text",
-    nullable: true,
-  })
-  promotion_id: string | null = null
-
```
adrien2p, line 24: is that intended? feels a bit out of scope no?

packages/cart/src/models/line-item-tax-line.ts
```diff
@@ -11,17 +11,32 @@ import LineItem from "./line-item"
...
   @ManyToOne({
     entity: () => LineItem,
-    index: "IDX_tax_line_item_id",
-    cascade: [Cascade.REMOVE, Cascade.PERSIST]
+    persist: false,
   })
   item: LineItem
 
-  @Property({ columnType: "text" })
+  @Property({ columnType: "text", index: "IDX_tax_line_item_id" })
```
adrien2p: suggestion: deletes at is null

packages/cart/src/models/line-item-tax-line.ts
```diff
@@ -11,17 +11,32 @@ import LineItem from "./line-item"
...
   item_id: string
 
+  @Property({
+    columnType: "text",
+    nullable: true,
+    index: "IDX_line_item_tax_line_tax_rate_id"
```
adrien2p: suggestion: deleted at is null and tax rate id is not null

packages/cart/src/models/line-item.ts
```diff
@@ -23,23 +29,22 @@ type OptionalLineItemProps =
...
 @Entity({ tableName: "cart_line_item" })
+@Filter(DALUtils.mikroOrmSoftDeletableFilterOptions)
 export default class LineItem {
   [OptionalProps]?: OptionalLineItemProps
 
   @PrimaryKey({ columnType: "text" })
   id: string
 
-  @Property({ columnType: "text" })
+  @Property({ columnType: "text", index: "IDX_line_item_cart_id" })
```
adrien2p: suggestion: deleted at is null maybe to review in all new indexes?

adrien2p: todo: since the relation is persist false, it will remove the foreign key constraint from the db, maybe that we don't really care but otherwise you need to apply the many to one on the foregin key prop as well, we can discuss if needed

adrien2p (approved):
LGTM, couple of suggestions to look at 💪

packages/cart/src/models/adjustment-line.ts
olivermrbl: It is intended, as we need separate indexes for the inheriting classes

packages/cart/src/models/line-item.ts
olivermrbl: We prob. don't want this persist flag here actually. Will revert.

sradevski (approved):

PR #6478, opened 2024-02-22
feat: Line Items API Routes

olivermrbl:
**What**
- `POST /store/carts/:id/line-items`
- `POST /store/carts/:id/line-items/:id`
- `DELETE /store/carts/:id/line-items/:id`

**Outstanding**
- Integration tests
- Module integrations: Payment, Fulfillment, Promotions

Depends on #6475 and #6449.


packages/cart/src/services/cart-module.ts
```diff
@@ -636,7 +636,7 @@ export default class CartModuleService<
         ? methodIdsOrSelector
         : [methodIdsOrSelector]
     }
-    await this.shippingMethodService_.delete(toDelete, sharedContext)
+    await this.shippingMethodService_.softDelete(toDelete, sharedContext)
```
adrien2p, line 639: question: Shouldn't we just use the abstract soft delete implementation that does that for each entity? also, in that case, we would have to remove all those implementation and rename the service interface methods

adrien2p: I think soft remove is more appropriate for self documentation POV but also because in other places you might have remove which then collides in terms of behaviour with other modules

olivermrbl: I'll do this in a follow-up PR to not increase the scope more than it already has.

packages/core-flows/src/definition/line-item/steps/delete-line-items.ts
```diff
@@ -0,0 +1,27 @@
...
+
+export const deleteLineItemsStepId = "delete-line-items"
+export const deleteLineItemsStep = createStep(
+  deleteLineItemsStepId,
+  async (ids: string[], { container }) => {
+    const service = container.resolve<ICartModuleService>(
+      ModuleRegistrationName.CART
+    )
+
+    await service.removeLineItems(ids)
```
adrien2p, line 13: question: why don't we use the native `deleteLineItems`?

packages/core-flows/src/definition/line-item/steps/delete-line-items.ts
```diff
@@ -0,0 +1,27 @@
...
+  },
+  async (ids, { container }) => {
+    if (!ids?.length) {
+      return
+    }
+    const service = container.resolve<ICartModuleService>(
+      ModuleRegistrationName.CART
+    )
+
+    await service.restoreLineItems(ids)
```
adrien2p, line 25: todo: the items are hard removed above, the restore wont work here

packages/medusa/src/api-v2/store/carts/[id]/line-items/[line_id]/route.ts
```diff
@@ -0,0 +1,84 @@
...
+
+export const POST = async (req: MedusaRequest, res: MedusaResponse) => {
+  const cartModuleService = req.scope.resolve<ICartModuleService>(
+    ModuleRegistrationName.CART
+  )
+
+  const cart = await cartModuleService.retrieve(req.params.id, {
+    select: ["id", "region_id", "currency_code"],
+    relations: ["region", "items", "items.variant_id"],
+  })
```
Comment (adrien2p) [inline packages/medusa/src/api-v2/store/carts/[id]/line-items/[line_id]/route.ts:20]: suggestion: should we use the remote query here?

packages/medusa/src/api-v2/store/carts/[id]/line-items/[line_id]/route.ts
```diff
@@ -0,0 +1,84 @@
...
+
+  const remoteQuery = req.scope.resolve("remoteQuery")
+
+  const query = remoteQueryObjectFromString({
+    entryPoint: "cart",
+    fields: defaultStoreCartFields,
+  })
+
+  const [cart] = await remoteQuery(query, {
+    cart: { id: req.params.id },
```
Comment (adrien2p) [inline packages/medusa/src/api-v2/store/carts/[id]/line-items/[line_id]/route.ts:80]: suggestion: this could be part of the `remoteQueryObjectFromString`

packages/types/src/cart/common.ts
```diff
@@ -400,6 +400,10 @@ export interface CartLineItemDTO extends CartLineItemTotalsDTO {
    * When the line item was updated.
    */
   updated_at?: Date
+  /**
+   * When the line item was deleted.
+   */
+  deleted_at?: Date
```
adrien2p, line 406: nit: nullable and rm `?` but the DTO's will need some rework later so you don't have to do it here

packages/types/src/cart/workflows.ts
```diff
@@ -32,6 +33,12 @@ export interface CreateCartCreateLineItemDTO {
   metadata?: Record<string, unknown>
 }
 
+export interface UpdateLineItemInCartWorkflowInputDTO {
+  cart: CartDTO
```
adrien2p, line 37: question: when doing an update of the item we need the entire DTO for which reason? I think, unless necessary, we should describe only what is expected and required

adrien2p (approved):
LGTM 🔥 one todo around the deletion and few comments

packages/core-flows/src/definition/line-item/steps/delete-line-items.ts
olivermrbl: Will be updated in #6491

packages/core-flows/src/definition/line-item/steps/delete-line-items.ts
olivermrbl: Will be updated in #6491

packages/medusa/src/api-v2/store/carts/[id]/line-items/[line_id]/route.ts
Comment (olivermrbl) [reply packages/medusa/src/api-v2/store/carts/[id]/line-items/[line_id]/route.ts:20]: I chose to use the module service directly, as I don't need any other resources than what it internal to the cart module. Any specific reason you want to use the remote query?

packages/types/src/cart/workflows.ts
olivermrbl: The workflow is still missing some other modules, so I'd like to keep it as the full DTO until we know with certainty what properties are required.

olivermrbl: But I like the idea!

packages/medusa/src/api-v2/store/carts/[id]/line-items/[line_id]/route.ts
Comment (olivermrbl) [reply packages/medusa/src/api-v2/store/carts/[id]/line-items/[line_id]/route.ts:80]: I had that in a previous PR and I recall a suggestion about changing it to this. To avoid going too much back and forth, I will not update this now. We can the usage at a later point.

PR #6484, opened 2024-02-23
feat(api-key): Allow revoking in the future, and enforce the secret key

sradevski:
Since there is quite a bit of code here already, I'll do the middleware changes in a separate PR

### DIFF
--- a/packages/api-key/src/services/api-key-module-service.ts
+++ b/packages/api-key/src/services/api-key-module-service.ts
@@ -13,7 +13,12 @@ import {
 } from "@medusajs/types"
 import { entityNameToLinkableKeysMap, joinerConfig } from "../joiner-config"
 import { ApiKey } from "@models"
-import { CreateApiKeyDTO, TokenDTO } from "@types"
+import {
+  CreateApiKeyDTO,
+  RevokeApiKeyInput,
+  TokenDTO,
+  UpdateApiKeyInput,
+} from "@types"
 import {
   ApiKeyType,
   InjectManager,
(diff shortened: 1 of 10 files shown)


integration-tests/plugins/__tests__/api-key/admin/api-key.spec.ts
```diff
@@ -106,4 +106,29 @@ describe("API Keys - Admin", () => {
     expect(deleted.status).toEqual(200)
     expect(listedApiKeys.data.apiKeys).toHaveLength(0)
   })
+
+  it.skip("can use a secret api key for authentication", async () => {
```
sradevski, line 112: I am waiting for #6447 to get merged first before we can actually enforce the API keys. I'll add the enforcement changes there

adrien2p (approved):
:firte

packages/api-key/src/services/api-key-module-service.ts
```diff
@@ -132,61 +138,109 @@ export default class ApiKeyModuleService<TEntity extends ApiKey = ApiKey>
...
+  async upsert(
+    data: ApiKeyTypes.UpsertApiKeyDTO[],
     sharedContext?: Context
   ): Promise<ApiKeyTypes.ApiKeyDTO[]>
+  async upsert(
+    data: ApiKeyTypes.UpsertApiKeyDTO,
+    sharedContext?: Context
+  ): Promise<ApiKeyTypes.ApiKeyDTO>
+  @InjectTransactionManager("baseRepository_")
+  async upsert(
```
adrien2p: nit: some spaces would be good :)

packages/api-key/src/services/api-key-module-service.ts
```diff
@@ -132,61 +138,109 @@ export default class ApiKeyModuleService<TEntity extends ApiKey = ApiKey>
...
+
+    const operations: Promise<ApiKeyTypes.ApiKeyDTO[]>[] = []
+
+    if (forCreate.length) {
+      const op = async () => {
+        const [createdApiKeys, generatedTokens] = await this.create_(
+          forCreate,
+          sharedContext
+        )
+        const serializedResponse = await this.baseRepository_.serialize<
```
adrien2p, line 171: suggestion: I think it is fine, but in general it is better to serialize outside a transaction to avoid some side effects

packages/api-key/src/services/api-key-module-service.ts
```diff
@@ -329,13 +372,52 @@ export default class ApiKeyModuleService<TEntity extends ApiKey = ApiKey>
...
+    const secretKeys = await this.apiKeyService_.list(
+      {
+        type: ApiKeyType.SECRET,
+        // If the revoke date is set in the future, it means the key is still valid.
+        $or: [
+          { revoked_at: { $eq: null } },
+          { revoked_at: { $gt: new Date() } },
+        ],
+      },
+      {},
```
adrien2p: todo: If we want to get all tokens I would suggest to pass `take: null` to the config options, otherwise the default limit is 15, so you could miss some tokens

packages/api-key/src/services/api-key-module-service.ts
```diff
@@ -329,13 +372,52 @@ export default class ApiKeyModuleService<TEntity extends ApiKey = ApiKey>
...
+      return false
+    }
+
+    const serialized =
+      await this.baseRepository_.serialize<ApiKeyTypes.ApiKeyDTO>(
+        matchedKeys[0],
+        { populate: true }
+      )
+
+    return serialized
```
adrien2p: suggestion: same as the previous serialize

packages/api-key/src/services/api-key-module-service.ts
```diff
@@ -359,7 +441,10 @@ export default class ApiKeyModuleService<TEntity extends ApiKey = ApiKey>
     const dbSecretKeys = await this.apiKeyService_.list(
       {
         type: ApiKeyType.SECRET,
-        revoked_at: null,
+        $or: [
+          { revoked_at: { $eq: null } },
+          { revoked_at: { $gt: new Date() } },
+        ],
       },
       {},
```
adrien2p: todo: same comment for auto pagnation

adrien2p: in fact everywhere you are expecting to get all tokens and not the default 15 limit

adrien2p (approved):
LGTM, few todo's and suggestions

PR #6486, opened 2024-02-23
feat(tax): soft deletes

srindom:
**What**
- Adds soft deletes


packages/tax/src/models/tax-rate.ts
```diff
@@ -18,7 +20,10 @@ import {
 import TaxRegion from "./tax-region"
 import TaxRateRule from "./tax-rate-rule"
 
-type OptionalTaxRateProps = DAL.EntityDateColumns
+type OptionalTaxRateProps =
+  | "tax_region"
```
adrien2p: you can rm the tax_region here

packages/tax/src/models/tax-rate.ts
```diff
@@ -56,17 +62,20 @@ export default class TaxRate {
...
 
-  @Property({ columnType: "text" })
-  tax_region_id: string
-
   @ManyToOne(() => TaxRegion, {
     fieldName: "tax_region_id",
     index: taxRegionIdIndexName,
-    cascade: [Cascade.REMOVE, Cascade.PERSIST],
+    mapToPk: true,
+    cascade: [Cascade.REMOVE],
```
adrien2p, line 72: todo: not sure this cascade should be here, I don't see a region being removed when we remove one of the tax rates

packages/tax/src/models/tax-rate.ts
```diff
@@ -56,17 +62,20 @@ export default class TaxRate {
...
   })
+  tax_region_id: string
+
+  @ManyToOne({ entity: () => TaxRegion, persist: false })
   tax_region: TaxRegion
 
-  @OneToMany(() => TaxRateRule, (rule) => rule.tax_rate)
+  @OneToMany(() => TaxRateRule, (rule) => rule.tax_rate, {
```
adrien2p, line 80: question: can we create a tax_rate from a rule? if not (which I expect), in that case, the relation needs some rework in the rule model

packages/tax/src/models/tax-rate.ts
```diff
@@ -56,17 +62,20 @@ export default class TaxRate {
   @Property({ columnType: "bool", default: false })
   is_combinable = false
 
-  @Property({ columnType: "text" })
-  tax_region_id: string
-
   @ManyToOne(() => TaxRegion, {
     fieldName: "tax_region_id",
```
adrien2p, line 70: todo: columnType and migration generation just to ensure everything is alright

packages/tax/src/models/tax-region.ts
```diff
@@ -49,18 +55,21 @@ export default class TaxRegion {
...
   province_code: string | null = null
 
-  @Property({ columnType: "text", nullable: true })
-  parent_id: string | null = null
-
   @ManyToOne(() => TaxRegion, {
     index: "IDX_tax_region_parent_id",
-    cascade: [Cascade.PERSIST],
+    fieldName: "parent_id",
     onDelete: "set null",
```
adrien2p: todo: rm, the nullable does that already, also, I suggest to regenerate the migration to validate all the configuration

packages/tax/src/models/tax-region.ts
```diff
@@ -49,18 +55,21 @@ export default class TaxRegion {
   @Property({ columnType: "text", nullable: true })
   province_code: string | null = null
 
-  @Property({ columnType: "text", nullable: true })
-  parent_id: string | null = null
-
   @ManyToOne(() => TaxRegion, {
     index: "IDX_tax_region_parent_id",
```
adrien2p, line 56: suggestion: Maybe a partial index applied on the prop directly

packages/tax/src/models/tax-region.ts
```diff
@@ -18,7 +20,10 @@ import {
 } from "@mikro-orm/core"
 import TaxRate from "./tax-rate"
 
-type OptionalTaxRegionProps = DAL.EntityDateColumns
+type OptionalTaxRegionProps =
+  | "parent_id"
```
adrien2p: todo: rm

packages/tax/src/models/tax-region.ts
```diff
@@ -18,7 +20,10 @@ import {
...
+  | "tax_rates"
```
adrien2p: todo: rm

adrien2p (commented):
Very good, few things to look at, we can discuss if needed 💪

packages/tax/src/models/tax-rate.ts
srindom: don't think I understand optional props fully - can you elaborate on what they are used for?

packages/tax/src/models/tax-rate.ts
srindom: Nope that's not possible - what needs to be reworked?

packages/tax/src/models/tax-rate.ts
adrien2p: the tax region is not persisted and goes through the region_id, in that case the tax_region is a virtual prop that holds the relation so it is optional by default normally

packages/tax/src/models/tax-rate.ts
srindom: It's the other way around - when we remove a TaxRegion, its associated TaxRates should be removed - that's what this cascade ensures.

I thought about why the "soft-remove" doesn't implement it this way around also?

packages/tax/src/models/tax-rate.ts
srindom: And what about rules here?

adrien2p: can be removed to as it is a collection and the prop is always initialized

packages/tax/src/models/tax-rate.ts
adrien2p: in the tax rate rule there is a cascade persist on the tax tate, I believe it should be removed right?

srindom: Yep you are right - take a look at the latest changes. wdyt?

packages/tax/src/models/tax-rate.ts
adrien2p: just found out that those cascade are application level cascade, and since we do not populate the relations and use native delete, the cascade is not performed 🤔 The db level cascade is the onDelete: cascade

packages/tax/src/models/tax-rate.ts
adrien2p: I would say lets go with that as all the tests are green, though the cascade are set at the application level we might need to revisit some of the relation definition any way. I would say lets re have a look at the details later on am move forward 🚀

adrien2p (approved):

PR #6487, opened 2024-02-23
feat: Update signature of region module to match latest spec

sradevski:


packages/region/src/services/region-module.ts
```diff
@@ -143,54 +143,72 @@ export default class RegionModuleService<
...
-    )
-
-    const regions = await this.baseRepository_.serialize<
-      RegionDTO[] | RegionDTO
-    >(updateResult)
-
-    return isString(idOrSelectorOrData) ? regions[0] : regions
-  }
+    const input = Array.isArray(data) ? data : [data]
+    const forUpdate = input.filter((region) => region.id)
```
adrien2p:
```suggestion
    const forUpdate = input.filter((region): region is  UpsertRegionDTO  => region.id)
```

packages/region/src/services/region-module.ts
```diff
@@ -143,54 +143,72 @@ export default class RegionModuleService<
...
+    const forCreate = input.filter((region) => !region.id)
```
adrien2p: same as above

packages/region/src/services/region-module.ts
```diff
@@ -143,54 +143,72 @@ export default class RegionModuleService<
...
+        MedusaError.Types.INVALID_ARGUMENT,
+        `You cannot mix create and update in the same call`
+      )
     }
 
-    if (Array.isArray(idOrSelectorOrData)) {
-      normalizedInput = idOrSelectorOrData
+    if (forCreate.length) {
+      const result = await this.create_(
+        input as unknown as CreateRegionDTO[],
```
adrien2p:
```suggestion
        forCreate,
```

packages/region/src/services/region-module.ts
```diff
@@ -143,54 +143,72 @@ export default class RegionModuleService<
...
+        sharedContext
+      )
+      return await this.baseRepository_.serialize<RegionDTO[]>(
+        Array.isArray(data) ? result : result[0]
+      )
     }
 
-    if (isObject(idOrSelectorOrData)) {
+    const result = await this.update_(
+      forUpdate as unknown as UpdateRegionInput[],
```
adrien2p:
```suggestion
      forUpdate,
```

packages/region/src/services/region-module.ts
sradevski: @adrien2p this doesn't quite work since the `UpsertRegionDTO` object has several undefined values that are required in `UpdateRegionDTO` and `CreateRegionDTO`, so we have to force the type.

 I guess what I can do is also apply some validation on the input since of course you can never trust just the typings, but I thought that's better reserved for the create and update functions themselves.
 
 I am not a Typescript expert so there might be another better way to do it?

adrien2p: sorry I wanted to put `UpdateRegionInput` instead, of course we can also have some validation but in theory the subsequent method will handle it

olivermrbl (approved):
LGTM

PR #6489, opened 2024-02-23
chore(util): Detect circular dependencies on soft delete

adrien2p:
**What**
Detect circular soft remove configuration and throw. This prevents from having infinite loop and result into the call to hang. Also should prevent any mistake in relation configuration


carlos-r-l-rodrigues (approved):
🔥

olivermrbl (approved):

PR #6491, opened 2024-02-23
chore(cart): Use module native soft-delete/delete methods

olivermrbl:


srindom:
I would still expect a hard delete of entities to exist on the cart module for people to clean up or to make real deletes if they want to. Although, we might want to give them different names e.g. `deleteShippingMethod` instead of `removeShippingMethod`. 

Any thoughts around this?

adrien2p:
> I would still expect a hard delete of entities to exist on the cart module for people to clean up or to make real deletes if they want to. Although, we might want to give them different names e.g. `deleteShippingMethod` instead of `removeShippingMethod`.
> 
> Any thoughts around this?

I think it would make sense to keep the typings, by default, the abstract class have a default implementation of delete* methods and removing the types does not remove the fact that those method still exists. Also, I link my last messageabout cascade here https://medusacorp.slack.com/archives/C025KMS13SA/p1708936203349889

srindom:
> I think it would make sense to keep the typings, by default, the abstract class have a default implementation of delete* methods and removing the types does not remove the fact that those method still exists.

I'm not sure I follow this completely - in my opinion, the delete methods should still exist on the module (and thereby also on the ICartModuleService) so consumers can call hard deletes.

adrien2p:
> > I think it would make sense to keep the typings, by default, the abstract class have a default implementation of delete* methods and removing the types does not remove the fact that those method still exists.
> 
> I'm not sure I follow this completely - in my opinion, the delete methods should still exist on the module (and thereby also on the ICartModuleService) so consumers can call hard deletes.

Yes this is what I said 😂 I just added a note saying that removing it from the interface does not remove it from the class

olivermrbl:
@adrien2p @srindom, you are right. I added back that methods in the interface.

olivermrbl:
Converting to draft until #6478 is merged

srindom (approved):
LGTM

PR #6499, opened 2024-02-25
fix(utils): bignumber util considers nullable options when setting value

riqwan:
what:

- when setting null for nullable columns, big number utils should not throw error
- remove circular soft delete depenedencies for cart module
- trims zeros on big number values


carlos-r-l-rodrigues:
I am doing [similar changes](https://github.com/medusajs/medusa/pull/6467/files#diff-830d2d12fe2a3f7a17746929bbee34a91eff9e45845eccb19042b097cfda8efc) to it + some additional ones.
do you mind to bring that to this PR?

riqwan:
@carlos-r-l-rodrigues sounds good, I can do that once I'm back. 👍🏻

riqwan:
@carlos-r-l-rodrigues added them in here, let me know if i missed anything.

carlos-r-l-rodrigues (approved):
LGTM

PR #6504, opened 2024-02-26
chore(utils): Soft delete should allow self referencing circular deps

adrien2p:
**What**
Self referencing circular dep should not be an error and should be accepted


srindom (approved):
LGTM!

carlos-r-l-rodrigues (approved):

Issue #6502, opened 2024-02-26
Typo in Configurations/database_url [:post] instead of [:port]


nkhar:
### Preliminary Checks

- [X] This issue is not a duplicate. Before opening a new issue, please search existing issues: https://github.com/medusajs/medusa/issues

### Issue Summary

There is a typo:
https://docs.medusajs.com/development/backend/configurations#database_url
instead of [:port] there is [:post]
[:post]: an optional port that the PostgreSQL server is listening on. By default, it's 5432. When provided, make sure to put : before the port.

### How can this issue be resolved?

1.
2.
3.
...

### Are you interested in working on this issue?

- [X] I would like to fix this issue

shahednasser:
Nice catch! You can fix it [here](https://github.com/medusajs/medusa/blob/3b18f399c120d5bf2e5ede3b043b20271d65b31a/packages/types/src/common/config-module.ts#L249)

nkhar:
@shahednasser thank you for helping with the fix. 

This is my first open source contribution to Medusa JS and in fact this is my first contribution to JS/TS repository. There are likely to be mistakes in the PR. Please feel free to provide any feedback so that I can improve my contribution.

https://github.com/medusajs/medusa/pull/6505

PR #6505, opened 2024-02-26
docs: Fixed typo in database_url [:post] instead of [:port]

nkhar:
https://docs.medusajs.com/development/backend/configurations#database_url
There the typo :
instead of [:port] there is [:post]
[:post]: an optional port that the PostgreSQL server is listening on. By default, it's 5432. When provided, make sure to put : before the port.

https://github.com/medusajs/medusa/issues/6502


shahednasser (approved):
LGTM! Thanks for your contribution!

PR #6507, opened 2024-02-26
feat(fulfillment): List shipping options filtered by context anmd rules

adrien2p:
**What**
Should be able to list the shipping options with or without a context, when a context is provided all the rules of the shipping options must be valid for the shipping options to be returned.

FIXES CORE-1765

### DIFF
--- a/packages/store/integration-tests/__fixtures__/index.ts
+++ b/packages/store/integration-tests/__fixtures__/index.ts
@@ -1,0 +1,10 @@
+import { StoreTypes } from "@medusajs/types"
+
+export const createStoreFixture: StoreTypes.CreateStoreDTO = {
+  name: "Test store",
+  default_sales_channel_id: "test-sales-channel",
+  default_region_id: "test-region",
+  metadata: {
+    test: "test",
+  },
+}
--- a/packages/store/integration-tests/__tests__/store-module-service.spec.ts
+++ b/packages/store/integration-tests/__tests__/store-module-service.spec.ts
@@ -1,6 +1,7 @@
 import { Modules } from "@medusajs/modules-sdk"
 import { IStoreModuleService } from "@medusajs/types"
 import { moduleIntegrationTestRunner, SuiteOptions } from "medusa-test-utils"
+import { createStoreFixture } from "../__fixtures__"
 
 jest.setTimeout(100000)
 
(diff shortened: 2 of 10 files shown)


packages/fulfillment/src/services/fulfillment-module-service.ts
```diff
@@ -113,6 +117,97 @@ export default class FulfillmentModuleService<
...
+    // Apply rules context filtering
+    if (context) {
+      shippingOptions = shippingOptions.filter((shippingOption) => {
+        if (!shippingOption.rules?.length) {
+          return true
+        }
+
+        return isContextValid(
+          context,
+          shippingOption.rules.map((r) => r)
```
riqwan, line 202:
```suggestion
          shippingOption.rules
```
this should suffice right?

riqwan (approved):
lgtm!

packages/fulfillment/src/services/fulfillment-module-service.ts
adrien2p: It is a collection under the hood, I just do that to extract them, the other options would be to force cast them ahah

riqwan: ah, i see!

packages/fulfillment/src/services/fulfillment-module-service.ts
```diff
@@ -113,6 +117,100 @@ export default class FulfillmentModuleService<
...
+  async listShippingOptions(
+    filters: FilterableShippingOptionProps = {},
+    config: FindConfig<ShippingOptionDTO> = {},
+    sharedContext?: Context
+  ): Promise<FulfillmentTypes.ShippingOptionDTO[]> {
+    const {
+      filters: normalizedFilters,
+      config: normalizedConfig,
+      context,
+    } = FulfillmentModuleService.normalizeFiltersAndConfigForShippingOptionsFiltering(
```
adrien2p: Q: do you have a better naming than that? `shippingOptionFilteringNormalization` maybe?

packages/fulfillment/integration-tests/__tests__/fulfillment-module-service.spec.ts
```diff
@@ -249,6 +249,200 @@ moduleIntegrationTestRunner({
...
+              ],
+            })
+
+            const shippingProfile = await service.createShippingProfiles({
+              name: "test",
+              type: "default",
+            })
+
+            // TODO: change that for a real provider instead of fake data manual inserted data
+            const [{ id: providerId }] =
```
sradevski: NIT: We can have a helper func (inside the .spec file is just fine) with this exact code, as it will be cleaner IMO

packages/fulfillment/integration-tests/__tests__/fulfillment-module-service.spec.ts
```diff
@@ -249,6 +249,200 @@ moduleIntegrationTestRunner({
...
+                  },
+                ],
+              },
+            ])
+
+            const listedOptions = await service.listShippingOptions({
+              name: shippingOption1.name,
+            })
+
+            expect(listedOptions).toEqual(
```
sradevski: You can also just do `expect(listedOptions[0].id).toEqual(shippingOption1.id)`, it's a bit easier to read. Also we can just check `expect(listedOptions).toHaveLength(1)` as we are not doing that check currently.

packages/fulfillment/integration-tests/__tests__/fulfillment-module-service.spec.ts
```diff
@@ -249,6 +249,200 @@ moduleIntegrationTestRunner({
...
+              ],
+            })
+
+            const shippingProfile = await service.createShippingProfiles({
+              name: "test",
+              type: "default",
+            })
+
+            // TODO: change that for a real provider instead of fake data manual inserted data
+            const [{ id: providerId }] =
```
sradevski: Ah, a helper func is definitely nicer 😄

packages/fulfillment/integration-tests/__tests__/fulfillment-module-service.spec.ts
```diff
@@ -249,6 +249,200 @@ moduleIntegrationTestRunner({
...
+                description: "test-description",
+                label: "test-label",
+              },
+              data: {
+                amount: 1000,
+              },
+            }
+
+            const [shippingOption1, ,shippingOption3] = await service.createShippingOptions([{
+              name: "test-option",
```
sradevski: We can create some fixtures so we don't have to repeat the same creation options again and again. Then if you want to change them, just destructure and change only the fields you care about.

packages/fulfillment/src/services/fulfillment-module-service.ts
riqwan: `normalizeShippingOptionsList(Params|Args)` since its not just filters?

packages/fulfillment/src/services/fulfillment-module-service.ts
```diff
@@ -113,6 +117,100 @@ export default class FulfillmentModuleService<
     return joinerConfig
   }
 
+  protected static normalizeFiltersAndConfigForShippingOptionsFiltering(
```
sradevski: Can we maybe split the filter and config normalization in two functions, and then you simply call

```
const filters = normalizeFilters(...)
const config = normalizeConfig(...)
```

packages/fulfillment/src/services/fulfillment-module-service.ts
sradevski: See above 😄

packages/fulfillment/integration-tests/__tests__/fulfillment-module-service.spec.ts
adrien2p: I don't want to just check the length to be sure it is the right one that is returned, but I can definately just check the id ahah

packages/fulfillment/src/services/fulfillment-module-service.ts
adrien2p: the config is dependent on what is in the filter as well, that is why I kept them together. I prefer having only one function for this case, but I like riqwan naming suggestion

packages/fulfillment/integration-tests/__tests__/fulfillment-module-service.spec.ts
sradevski: @adrien2p I think it's still good to check the length so we are sure there isn't more than 1 item returned, don't you think so?

packages/fulfillment/src/services/fulfillment-module-service.ts
sradevski: Yup the name is definitely better. Just to decouple the two a bit (if possible), can't we do
```
const filters = normalizeFilters(...)
const config = normalizeConfig(filters, ...)
```

So like that `normalizeConfig` can operate on top of normalized filters? Just thinking out loud, I'm fine with them being together if you prefer it 👍

packages/fulfillment/integration-tests/__tests__/fulfillment-module-service.spec.ts
adrien2p: Yes sorry if I wasn't clear, but yeah I ll do both, I just didn't want to `just` check the length :D

sradevski: Ah yes, definitely!

sradevski (approved):

PR #6509, opened 2024-02-26
feat(payment): Add migration

olivermrbl:
Depends on:
- #6449 
- #6311

Some columns will be updated when #6311 has landed and BigNumber introduced.


packages/payment/src/models/payment.ts
```diff
@@ -30,11 +35,11 @@ export default class Payment {
   @PrimaryKey({ columnType: "text" })
   id: string
 
-  @Property({
-    columnType: "numeric",
-    serializer: Number,
-  })
-  amount: number
+  @MikroOrmBigNumberProperty()
+  amount: BigNumber | number
```
fPolic, line 38: tests are failing because we end up with the amounts as strings

packages/payment/src/migrations/Migration20240225134525.ts
```diff
@@ -0,0 +1,162 @@
+import { generatePostgresAlterColummnIfExistStatement } from "@medusajs/utils"
+import { Migration } from "@mikro-orm/migrations"
+
+export class Migration20240225134525 extends Migration {
+  async up(): Promise<void> {
+    this.addSql(
+      `
+CREATE TABLE IF NOT EXISTS "payment_collection" (
```
sradevski: Let's try to break this apart in multiple `addSql` statements to aid readability.

One approach we can take here is to separate a clean v2 install, from the one that migrates from v1 to v2, see [this PR](https://github.com/medusajs/medusa/pull/6515/files#diff-e20b7a5cf6444556be24a5c3985f01388e9e4e3ecdc12f1001503657f871c57f).

The benefit of this is two-fold:
1. It makes the code more readable
2. We can remove the v1-to-v2 migration code later on when we fully deprecate v1.

adrien2p: 👍

olivermrbl: ~We will need to go through all migrations before releasing v2 to ensure everything is correct and they are clean and easy to read. Therefore, I would like to punt this type of work so that we do not do double the work.~

~The idea of getting v1 tables up to speed with v2 schemas was introduced after the product and pricing module, so I know we already have some work to do in this area.~

Think this is fairly quick to do, so I'll add in the split. We still have some work to do and the payment module is a work in progress, so this is subject to change.

packages/payment/src/models/payment.ts
adrien2p: I ll have a look @olivermrbl

adrien2p: @fPolic @olivermrbl you can have a look [here](https://github.com/medusajs/medusa/pull/6509/commits/f86bf38934a466530a558d3e117873e50f9b13bf)

packages/payment/src/migrations/Migration20240225134525.ts
sradevski: @olivermrbl yeah I think we'll need to revisit every single migration and make sure we are not missing anything, but it's good to at least have a solid initial setup in place, but of course feel free to do as much as you think it's necessary :)

packages/payment/src/models/payment.ts
fPolic: Looks good @adrien2p 👍 

Also, should we define the amount as `number` only? From the current implementation of the decorator, the field is always set as bignumber.numeric so it will never be a BigNumber?

packages/payment/src/models/refund.ts
```diff
@@ -7,20 +7,21 @@ import {
...
 export default class Refund {
   @PrimaryKey({ columnType: "text" })
   id: string
 
-  @Property({
-    columnType: "numeric",
-    serializer: Number,
-  })
+  @MikroOrmBigNumberProperty()
   amount: number
```
fPolic: todo: BigNumber | number

fPolic (approved):
LGTM

PR #6510, opened 2024-02-26
feat: Add basic CRUD functionality to store module

sradevski:
Adds the core model and basic CRUD around it.

Note: The store model is not complete, but I prefer doing smaller PRs so it's easier to do a proper review. Adding currencies will be a follow-up PR.

### DIFF
--- a/packages/types/src/store/common/store.ts
+++ b/packages/types/src/store/common/store.ts
@@ -1,3 +1,15 @@
+import { BaseFilterable } from "../../dal"
+
 export interface StoreDTO {
   id: string
+  name: string
+  default_sales_channel_id?: string
+  default_region_id?: string
+  default_location_id?: string
+  metadata: Record<string, any> | null
+}
+export interface FilterableStoreProps
+  extends BaseFilterable<FilterableStoreProps> {
+  id?: string | string[]
+  name?: string | string[]
 }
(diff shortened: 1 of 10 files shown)


riqwan (approved):
lgtm! 🔥

packages/types/src/store/common/store.ts
```diff
@@ -1,3 +1,14 @@
+import { BaseFilterable } from "../../dal"
+
 export interface StoreDTO {
   id: string
+  name: string
+  default_sales_channel_id?: string
+  default_region_id?: string
+  metadata?: Record<string, any>
```
adrien2p: todo: rm `?` and add `| null`

adrien2p (approved):
LGTM, just one todo 🔥

packages/types/src/store/common/store.ts
sradevski: @adrien2p that's fine with me, but why do we prefer null over an optional?

adrien2p: Because it will be returned but with `null` as the default value is null no? the only moment it wont be returned will be if you don't select it. But this will be part of a dynamic typing i ll do at a later point.

adrien2p: and the foreign keys too normally

sradevski: ah ok you mean that's how we define it on the DB model, got it 👍 We can definitely work on the typings being more accurate and we need to settle on working with nulls vs undefined :D

PR #6513, opened 2024-02-26
feat(tax): normalize country and province code

srindom:
**What**
- Ensures that country and province codes are always stored as lower case.
- Ensures that calculation context normalizes input before getting region rates.

### DIFF
--- a/www/apps/docs/content/development/events/modules/redis.md
+++ b/www/apps/docs/content/development/events/modules/redis.md
@@ -76,7 +76,7 @@ Other available options include:
 
 - `queueName`: a string indicating the name of the BullMQ queue. By default, it’s `events-queue`.
 - `queueOptions`: an object containing options for the BullMQ queue. You can learn about available options in [BullMQ’s documentation](https://api.docs.bullmq.io/interfaces/QueueOptions.html). By default, it’s an empty object.
-- `redisOptions`: an object containing options for the Redis instance. You can learn about available options in [io-redis’s documentation](https://luin.github.io/ioredis/index.html#RedisOptions). By default, it’s an empty object.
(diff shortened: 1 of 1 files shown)


packages/tax/src/services/tax-module-service.ts
```diff
@@ -246,12 +227,20 @@ export default class TaxModuleService<
       {
         $or: [
           {
-            country_code: calculationContext.address.country_code,
+            country_code: this.normalizeRegionCodes(
```
sradevski: A nicer approach is to have a single `normalizeTaxLinesCalculationContext` function that will normalize all input at the beginning of the method, and then you don't need to worry about forgetting to normalize in one place anymore.

packages/tax/src/services/tax-module-service.ts
```diff
@@ -278,6 +267,38 @@ export default class TaxModuleService<
     return toReturn.flat()
   }
 
+  private prepareTaxRegionInputForCreate(
+    data: TaxTypes.CreateTaxRegionDTO | TaxTypes.CreateTaxRegionDTO[]
+  ) {
+    const input = Array.isArray(data) ? data : [data]
+    return input.reduce(
```
sradevski: This code is really difficult to read and understand. Maybe we can use a simple loop instead (usually long `reduce` functions make code less readable), and try to structure the logic a bit better?

packages/tax/src/services/tax-module-service.ts
srindom: Great push! Thanks

packages/tax/src/services/tax-module-service.ts
srindom: Gave it another shot - wdyt?

olivermrbl (approved):
LGTM

sradevski (approved):
Much better, thanks!

PR #6514, opened 2024-02-26
feat(core-flows,medusa,types,utils): adds update cart API with promotions

riqwan:
what:

- adds update cart API
  - workflow
  - promotions
  - sales channel
  - region
  - customer

### DIFF
--- a/.changeset/polite-papayas-exercise.md
+++ b/.changeset/polite-papayas-exercise.md
@@ -1,0 +1,5 @@
+---
+"@medusajs/medusa": patch
+---
+
+fix(medusa): add `order` param to discounts list
--- a/integration-tests/plugins/__tests__/api-key/admin/api-key.spec.ts
+++ b/integration-tests/plugins/__tests__/api-key/admin/api-key.spec.ts
@@ -108,4 +108,29 @@ describe("API Keys - Admin", () => {
     expect(deleted.status).toEqual(200)
     expect(listedApiKeys.data.apiKeys).toHaveLength(0)
   })
+
+  it.skip("can use a secret api key for authentication", async () => {
+    const api = useApi() as any
+    const created = await api.post(
+      `/admin/api-keys`,
+      {
+        title: "Test Secret Key",
+        type: ApiKeyType.SECRET,
+      },
+      adminHeaders
+    )
+
+    const createdRegion = await api.post(
+      `/admin/regions`,
+      {
+        name: "Test Region",
+        currency_code: "usd",
+        countries: ["us", "ca"],
+      },
+      { headers: { Authorization: `Bearer ${created.token}` } }
+    )
+
+    expect(createdRegion.status).toEqual(200)
+    expect(createdRegion.data.region.name).toEqual("Test Region")
+  })
 })
--- a/packages/admin-next/dashboard/public/locales/en-US/translation.json
+++ b/packages/admin-next/dashboard/public/locales/en-US/translation.json
@@ -187,7 +187,14 @@
     "domain": "Draft Orders"
   },
   "discounts": {
-    "domain": "Discounts"
+    "domain": "Discounts",
+    "deleteWarning": "You are about to delete the discount {{title}}. This action cannot be undone.",
+    "discountStatus": {
+      "scheduled": "Scheduled",
+      "expired": "Expired",
+      "active": "Active",
+      "disabled": "Disabled"
+    }
   },
   "pricing": {
     "domain": "Pricing"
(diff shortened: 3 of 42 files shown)


integration-tests/plugins/__tests__/cart/store/carts.spec.ts
```diff
@@ -288,6 +292,214 @@ describe("Store Carts API", () => {
...
+            expect.objectContaining({
+              id: "item-1",
+              adjustments: [],
+            }),
+          ],
+        })
+      )
+    })
+
+    it("should update a cart's region", async () => {
```
olivermrbl: suggestion(non-blocking): Squash the region, customer, and sales channel tests.

packages/core-flows/src/definition/cart/steps/update-carts.ts
```diff
@@ -1,61 +1,44 @@
...
-  selector: FilterableCartProps
-  update: UpdateCartDataDTO
-}
-
 export const updateCartsStepId = "update-carts"
 export const updateCartsStep = createStep(
   updateCartsStepId,
-  async (data: UpdateCartsStepInput, { container }) => {
-    const service = container.resolve<ICartModuleService>(
+  async (data: CartWorkflow.UpdateCartWorkflowInputDTO, { container }) => {
```
olivermrbl: thought(non-blocking): Can we keep this as a bulk step? Unless there is a specific reason for converting it to a single cart update, I'd like that.

packages/core-flows/src/definition/cart/workflows/update-carts.ts
```diff
@@ -1,20 +1,114 @@
...
+  findSalesChannelStep,
+  getActionsToComputeFromPromotionsStep,
+  prepareAdjustmentsFromPromotionActionsStep,
+  removeLineItemAdjustmentsStep,
+  removeShippingMethodAdjustmentsStep,
+  retrieveCartStep,
+  updateCartsStep,
+} from "../steps"
 
 export const updateCartsWorkflowId = "update-carts"
```
olivermrbl: nit(non-blocking): Rename to `updateCartWorkflow` -> operation on a single cart

packages/medusa/src/api-v2/store/carts/validators.ts
```diff
@@ -81,14 +81,19 @@ export class StorePostCartsCartReq {
...
   sales_channel_id?: string
 
   @IsObject()
   @IsOptional()
   metadata?: Record<string, unknown>
 
+  @IsOptional()
+  @IsArray()
+  @Type(() => String)
+  promo_codes: string[]
```
olivermrbl: nit:
```suggestion
  promo_codes?: string[]
```

packages/types/src/workflow/cart/update-cart.ts
```diff
@@ -0,0 +1,10 @@
+export interface UpdateCartWorkflowInputDTO {
```
olivermrbl: nit: I believe I introduced cart workflow types in `types/src/cart/workflows.ts`. Perhaps, we can consolidate. Nothing important, so consider this non-blocking

packages/core-flows/src/definition/cart/workflows/update-carts.ts
```diff
@@ -1,20 +1,114 @@
+import { CartDTO, CartWorkflow } from "@medusajs/types"
```
olivermrbl: nit(non-blocking): Rename file `update-cart.ts` -> operation on a single cart

packages/types/src/cart/workflows.ts
```diff
@@ -65,3 +65,14 @@ export interface AddToCartWorkflowInputDTO {
   items: CreateCartCreateLineItemDTO[]
   cart: CartDTO
 }
+
+export interface UpdateCartWorkflowInputDTO {
+  id: string
+  promo_codes?: string[]
+  region_id?: string
```
sradevski, line 72: Can you even update the region of a cart? I am still exploring what is a possibility or not so bear with me :D

The same question on customer_id, sales_channel_id and currency_code I guess.

riqwan: Good question 😄 I'm not 100% sure either.

From the customer_id perspective, I think its possible that you'd want to move an anonymous customer to a customer thats logged in mid cart. I suppose a customer could also update region/currency/sales channels, but that should have some consequences on the items that are already present in the cart, which is not done in this PR. 

Why these things are done are entirely dependent on the consumer's usecases and workflows.

sradevski: @riqwan I think adding some rail guards is still part of the module's responsibility, so we should either not allow eg. modifying a region, or ensure that if one does do that, we bring the cart into a consistent state (updating what items are in the cart, etc.). I guess we could shift the responsibility to a workflow, but this seems like a core responsibility of the cart.

Anyway, these are discussions that can happen later on as well, so it's not a blocker for sure, it's all about where responsibilities lie ultimately.

riqwan: Yup, completely agree there.

PR #6515, opened 2024-02-26
feat: Add basic endpoints and workflows for Store module

sradevski:


packages/core-flows/src/store/steps/create-stores.ts
```diff
@@ -0,0 +1,40 @@
...
+
+export const createStoresStepId = "create-stores"
+export const createStoresStep = createStep(
+  createStoresStepId,
+  async (data: CreateStoresStepInput, { container }) => {
+    const service = container.resolve<IStoreModuleService>(
+      ModuleRegistrationName.STORE
+    )
+
+    // We only allow a singleton Store through the HTTP API, so we return the store if there is one already.
```
riqwan: wee bit confused here, does this mean we only allow creating only one instance of a store?

If so:

1. Can't we just throw an error here?
2. Should we allow bulk create in this case?

packages/store/src/migrations/InitialSetup20240227075933.ts
```diff
@@ -0,0 +1,48 @@
+import { Migration } from "@mikro-orm/migrations"
+
+export class InitialSetup20240226130829 extends Migration {
+  async up(): Promise<void> {
+    // TODO: The migration needs to take care of moving data before dropping columns, among other things
+    const storeTables = await this.execute(
```
olivermrbl, line 6: question: I am seeing the following error in module integration tests (using the module test runner) following the same approach:
```
relation "information_schema.tables" does not exist
```

Did you encounter this? It looks as if the database is not set up properly. We might need a try catch here.

packages/core-flows/src/store/steps/create-stores.ts
sradevski: @riqwan thanks for pointing this out, I just noticed I have a bug 😄 

The context is this: We want to allow for people to build marketplaces with Medusa, where you'd like to be able to create multiple stores. Therefore the module should support multiple instances of a Store without any restrictions.

However, our HTTP API by default should only allow you to create one, and if there already is a store we want to return that one (therefore returning the existing store). The bug is, we pass `data.storesData` to the create request, but we should only take the first one in this case.

Does that make sense?

sradevski: However as I am explaining this, it does seem to me that it's better to just throw, rather than return the existing one... WDYT @olivermrbl?

riqwan: Yes, the bulk create question was for the workflow and not the module.

Since the core-flows are core-opinionated, I think it can safely keep these restrictions. If users want to extend it, they can override the workflow or the step.

packages/medusa/src/api-v2/admin/stores/route.ts
```diff
@@ -0,0 +1,44 @@
...
+    offset: metadata.skip,
+    limit: metadata.take,
+  })
+}
+
+export const POST = async (req: MedusaRequest, res: MedusaResponse) => {
+  const input = [req.validatedBody as CreateStoreDTO]
+
+  const { result, errors } = await createStoresWorkflow(req.scope).run({
+    input: { storesData: input },
```
olivermrbl: todo: As per our latest discussion, workflow input should be snake_cased. Or  alternatively renamed:
```suggestion
    input: { stores: input },
```

packages/medusa/src/api-v2/admin/stores/middlewares.ts
```diff
@@ -0,0 +1,53 @@
...
+  AdminGetStoresStoreParams,
+  AdminPostStoresReq,
+  AdminPostStoresStoreReq,
+} from "./validators"
+
+export const adminStoreRoutesMiddlewares: MiddlewareRoute[] = [
+  {
+    matcher: "/admin/stores*",
+    // middlewares: [authenticate("admin", ["bearer", "session"])],
+    // TODO: Apply authentication middleware correctly once https://github.com/medusajs/medusa/pull/6447 is merged.
```
olivermrbl: todo: https://github.com/medusajs/medusa/pull/6447 is merged, so we can update

packages/core-flows/src/store/workflows/create-stores.ts
```diff
@@ -0,0 +1,13 @@
+import { StoreDTO, CreateStoreDTO } from "@medusajs/types"
+import { WorkflowData, createWorkflow } from "@medusajs/workflows-sdk"
+import { createStoresStep } from "../steps"
+
+type WorkflowInput = { storesData: CreateStoreDTO[] }
```
olivermrbl: todo: snake_case or rename
```suggestion
type WorkflowInput = { data: CreateStoreDTO[] }
```

packages/core-flows/src/store/steps/create-stores.ts
olivermrbl: I think you both have valid points. 

However, I would probably enforce the opinionated logic in the Rest API and not in the workflow. So the `createStores` workflow and step would allow you to create multiple stores, but the API won't and will on attempts throw an error. This makes for reusable workflows, while remaining opinionated for the average commerce case. As you mentioned, it is easy for people to repurpose the steps or workflow to achieve a multi-store setup through a custom API route.

Now as I was writing this, it occurred to me that in the vast majority of cases, the store will be created in a _defaults_ loader, as we do in v1, leaving the create-store endpoint redundant. So the alternative would be to leave it out and use the workflow in a loader?

Let me know what you think.

riqwan: A guard outside before running the workflow also sgtm!

sradevski: @olivermrbl Ok let's not apply any restrictions to the workflow, and let's remove the `create/POST` REST endpoint.

When you say `defaults` loader, do we expect to load defaults the same way in v2? Maybe a better way would be to have a "initialize" workflow that will create resources in all modules where we want a default? If that's the case maybe I can leave the workflow and the step existing, but it won't be used anywhere just yet?

Also, if we don't allow creating a store through the API, should we also not allow deleting one?

olivermrbl: > Maybe a better way would be to have a "initialize" workflow that will create resources in all modules where we want a default?

Yeah, this will likely be the approach, we will take. And the workflow would be called in a core loader.

> If that's the case maybe I can leave the workflow and the step existing, but it won't be used anywhere just yet?

Yes sure, we can do that and revisit if need be.

> Also, if we don't allow creating a store through the API, should we also not allow deleting one?

Yes, I think we can leave that out too.

sradevski: Great, I'll apply those changes now 👍

olivermrbl (approved):

PR #6516, opened 2024-02-26
feat(tax): add support for updating tax rates

srindom:
**What**
- Adds update


packages/tax/integration-tests/__tests__/index.spec.ts
```diff
@@ -8,6 +8,53 @@ moduleIntegrationTestRunner({
...
+          rate: 8.23,
+        })
+
+        await service.update(rate.id, {
+          name: "Updated Rate",
+          code: "TEST",
+          rate: 8.25,
+        })
+
+        await service.update(
```
sradevski: It's nice to check that the returned value from `update` and `create` matches what we expect, i.e the model that got created/updated

packages/tax/src/services/tax-module-service.ts
```diff
@@ -136,6 +136,49 @@ export default class TaxModuleService<
     })
   }
 
+  async update(
+    id: string,
+    data: TaxTypes.UpdateTaxRateDTO,
+    sharedContext?: Context
+  ): Promise<TaxTypes.TaxRateDTO>
+  async update(
```
sradevski, line 154: I think the code will get simpler if we don't have this particular override, as it's just a special case of the `selector` approach, but it is not as common of an use-case as the `id` signature. WDYT?

packages/tax/src/services/tax-module-service.ts
```diff
@@ -136,6 +136,49 @@ export default class TaxModuleService<
...
+  }
+
+  @InjectTransactionManager("baseRepository_")
+  protected async update_(
+    idOrSelector: string | string[] | TaxTypes.FilterableTaxRateProps,
+    data: TaxTypes.UpdateTaxRateDTO,
+    @MedusaContext() sharedContext: Context = {}
+  ) {
+    const selector =
+      Array.isArray(idOrSelector) || typeof idOrSelector === "string"
```
adrien2p: todo: we can use isString instead of typeof here

packages/tax/src/services/tax-module-service.ts
```diff
@@ -136,6 +136,49 @@ export default class TaxModuleService<
...
+  async update(
+    selector: string | string[] | TaxTypes.FilterableTaxRateProps,
+    data: TaxTypes.UpdateTaxRateDTO,
+    @MedusaContext() sharedContext: Context = {}
+  ): Promise<TaxTypes.TaxRateDTO | TaxTypes.TaxRateDTO[]> {
+    const rates = await this.update_(selector, data, sharedContext)
+    const serialized = await this.baseRepository_.serialize<
+      TaxTypes.TaxRateDTO[]
+    >(rates, { populate: true })
+    return typeof selector === "string" ? serialized[0] : serialized
```
adrien2p: todo: isString

adrien2p (approved):
LGTM

packages/tax/src/services/tax-module-service.ts
srindom: Is your suggestion to just remove the override and leave the rest or what is your thinking?

PR #6518, opened 2024-02-27
Update redis.md to show an additional step

leocabeza:
Adding `redis_url` in `projectConfig` key was needed in order to enable redis


shahednasser:
Thanks for your contribution! You shouldn't need to add redis to the project configurations, but you need to add it to the module configurations as explained [here](https://docs.medusajs.com/development/events/modules/redis#step-3-add-configuration). If you're facing issues with this, can you open an issue instead?

PR #6521, opened 2024-02-27
feat(api-key): Add api-key authentication to middleware

sradevski:
Also did a bit of a cleanup on the auth middleware. There should be no behavioral changes, just moved code around.

### DIFF
--- a/integration-tests/plugins/__tests__/api-key/admin/api-key.spec.ts
+++ b/integration-tests/plugins/__tests__/api-key/admin/api-key.spec.ts
@@ -131,7 +131,11 @@ describe("API Keys - Admin", () => {
         currency_code: "usd",
         countries: ["us", "ca"],
       },
-      { headers: { Authorization: `Bearer ${created.data.apiKey.token}` } }
+      {
+        auth: {
+          username: created.data.apiKey.token,
+        },
+      }
     )
 
     expect(createdRegion.status).toEqual(200)
@@ -149,7 +153,7 @@ describe("API Keys - Admin", () => {
       adminHeaders
     )
 
-    const revoked = await api.post(
+    await api.post(
       `/admin/api-keys/${created.data.apiKey.id}/revoke`,
       {},
       adminHeaders
@@ -163,7 +167,11 @@ describe("API Keys - Admin", () => {
           currency_code: "usd",
           countries: ["us", "ca"],
         },
-        { headers: { Authorization: `Bearer ${created.data.apiKey.token}` } }
+        {
+          auth: {
+            username: created.data.apiKey.token,
+          },
+        }
       )
       .catch((e) => e.message)
 
@@ -175,10 +183,10 @@ describe("API Keys - Admin", () => {
         countries: ["us", "ca"],
       },
       {
-        headers: {
-          Authorization: `Bearer ${created.data.apiKey.token}`,
-          ...adminHeaders.headers,
+        auth: {
+          username: created.data.apiKey.token,
         },
+        ...adminHeaders,
       }
     )
 
--- a/packages/medusa/src/utils/authenticate-middleware.ts
+++ b/packages/medusa/src/utils/authenticate-middleware.ts
@@ -103,12 +103,23 @@ const getApiKeyInfo = async (req: MedusaRequest): Promise<ApiKeyDTO | null> => {
   }
 
   const [tokenType, token] = authHeader.split(" ")
-  if (tokenType.toLowerCase() !== "bearer" || !token) {
+  if (tokenType.toLowerCase() !== "basic" || !token) {
     return null
   }
 
-  // Secret tokens start with 'sk_', and if it doesn't it could be a user JWT or a malformed token
+  // The token could have been base64 encoded, we want to decode it first.
+  let normalizedToken = token
   if (!token.startsWith("sk_")) {
+    normalizedToken = Buffer.from(token, "base64").toString("utf-8")
+  }
+
+  // Basic auth is defined as a username:password set, and since the token is set to the username we need to trim the colon
+  if (normalizedToken.endsWith(":")) {
+    normalizedToken = normalizedToken.slice(0, -1)
+  }
+
+  // Secret tokens start with 'sk_', and if it doesn't it could be a user JWT or a malformed token
+  if (!normalizedToken.startsWith("sk_")) {
     return null
   }
 
@@ -116,7 +127,7 @@ const getApiKeyInfo = async (req: MedusaRequest): Promise<ApiKeyDTO | null> => {
     ModuleRegistrationName.API_KEY
   ) as IApiKeyModuleService
   try {
-    const apiKey = await apiKeyModule.authenticate(token)
+    const apiKey = await apiKeyModule.authenticate(normalizedToken)
     if (!apiKey) {
       return null
     }


packages/medusa/src/utils/authenticate-middleware.ts
```diff
@@ -32,50 +35,40 @@ export const authenticate = (
...
 
-    // @ts-ignore
-    const session: MedusaSession = req.session || {}
+    // We only allow authenticating using a secret API key on the admin
+    if (authTypes.includes(API_KEY_AUTH) && isAdminScope(authScope)) {
+      const apiKey = await getApiKeyInfo(req)
+      if (apiKey) {
+        ;(req as AuthenticatedMedusaRequest).auth = {
+          actor_id: apiKey.id,
+          // TODO: Should we set this to the creator of the token (apiKey.created_by), or do we leave it empty?
```
olivermrbl: I would just leave it empty for now

packages/medusa/src/utils/authenticate-middleware.ts
```diff
@@ -32,50 +35,40 @@ export const authenticate = (
...
+        }
 
-    let authUser: AuthUserDTO | null = null
-    if (authTypes.includes(SESSION_AUTH)) {
-      if (
-        session.auth_user &&
-        stringEqualsOrRegexMatch(authScope, session.auth_user.scope)
-      ) {
-        authUser = session.auth_user
+        return next()
```
olivermrbl, line 50: question: Having `return next()` here means you cannot authenticate as a user on requests that carry a bearer token. Problably an edge case, but do we want to allow for that?

I would have thought just passing through all the configured auth types would make sense. Obviously, returning early if `req.auth` is set in one of them.

packages/medusa/src/utils/authenticate-middleware.ts
```diff
@@ -104,6 +97,96 @@ export const authenticate = (
...
 }
 
+const getApiKeyInfo = async (req: MedusaRequest): Promise<ApiKeyDTO | null> => {
+  const authHeader = req.headers.authorization
+  if (!authHeader) {
+    return null
+  }
+
+  const [tokenType, token] = authHeader.split(" ")
+  if (tokenType.toLowerCase() !== "bearer" || !token) {
```
olivermrbl: thought(non-blocking): Regarding your comment on Basic vs Bearer vs Something Third, I wanted to link a [discussion](https://github.com/medusajs/medusa/discussions/4037) opened in the community some time ago. 

This discussion actually prompted us to add a specific header `x-medusa-access-token` for API key authentication, instead of the Bearer which commonly carries a JWT.

However, I like the idea of moving back to Bearer or Basic for API key authentication, because those are native authentication headers and are therefore by default handled appropriately by clients and proxies, e.g. avoiding any type of caching. 

Would be curios to hear what you think?

olivermrbl (approved):
LGTM, clean PR @sradevski. Left a couple of comments.

packages/medusa/src/utils/authenticate-middleware.ts
sradevski: @olivermrbl This was intentional, but whether we want it like that or not I am happy to discuss 😄 

My thought-process was that we need to have some ordering/precendence when doing authentication. I thought an API key to be the most specialized case, and as such, should be the first on the list. If you haven't authenticated with an API key, then we fall back to other modes of authentication.

I think there is no right or wrong solution here, I personally just like not having too many options as it adds confusion to both the implementation side, as well as the user side. Let me know what you think though, I am not set on this approach :)

olivermrbl: > I thought an API key to be the most specialized case, and as such, should be the first on the list

Agree with the precedence and your priority. 

> then we fall back to other modes of authentication.

This is what I want to allow, but with the current implementation, I don't think this is possible, because we return `next()` regardless of the result of the API key auth no? In case API key authentication fails, we should proceed to try the other types of authentication.

One scenario I can think of where returning `next()` early could be an issue would be if you've configured your SDK in your client with an API key (and this client has regular user sign in too), and you choose to revoke that API key, then all your users would get 401s on attempted sign-in. Only when you have removed or updated the API key will your users be able to sign in again.

As I said, this is probably an edge case since you are unlikely to use the bearer token, if you also have regular user session sign-in, but thought I'd highlight it.

integration-tests/plugins/__tests__/api-key/admin/api-key.spec.ts
```diff
@@ -39,6 +40,9 @@ describe("API Keys - Admin", () => {
 
   beforeEach(async () => {
     await createAdminUser(dbConnection, adminHeaders)
+
+    // Used for testing cross-module authentication checks
+    regionService.createDefaultCountriesAndCurrencies()
```
olivermrbl: question: Don't we need an `await` here?

packages/medusa/src/utils/authenticate-middleware.ts
sradevski: @olivermrbl ok went through that discussion, I think some points are valid.

However, one thing that people miss about Bearer tokens is that they are not defined anywhere on what exactly they should carry. The do commonly carry JWT tokens as Bearer was added for Oauth 2.0, but they can literally carry anything that says "Give access to the bearer of this token". 

Now, I think it's still good to align with what people expect, rather than what is technically correct. I checked Stripe and they suggest Basic auth with the token, which I think is fine and we can do. This will just need another middleware to extract the basic credentials before calling the auth middleware, but it's not a big deal.

When it comes to session vs JWT authentication, it all depends on which tradeoffs you want to take. I think a pure cookie-based session auth is a bit outdated as it limits how this API can be called from eg. mobile apps (as that discussion points out). However, if you do want to allow revoking JWT tokens, for example, or you need a fine-grained authorization management, you'll end up with some session state regardless, so JWT token authentication gets a different meaning. That, however, is something we can discuss at a later point.

For now, I can just follow Stripe's approach and just use basic auth?

integration-tests/plugins/__tests__/api-key/admin/api-key.spec.ts
sradevski: Good catch, we do!

packages/medusa/src/utils/authenticate-middleware.ts
sradevski: Sounds good 👍

packages/medusa/src/utils/authenticate-middleware.ts
sradevski: @olivermrbl If you see, we only call `return next()` only if the `if(apiKey)` is true, and that is only true if the API key is valid, so it will fallback to other modes of authentication, but I'll add it in the test scenario.

packages/medusa/src/utils/authenticate-middleware.ts
olivermrbl: > For now, I can just follow Stripe's approach and just use basic auth?

This sounds good to me.

packages/medusa/src/utils/authenticate-middleware.ts
olivermrbl: Oh man, I'm so sorry for going back and forth on this. Missed a curly bracket.

But I am glad that we agree 😄

sradevski: Added one more test where we try to use a revoked key

sradevski: @olivermrbl no worries, it could also have been a bug so thanks for pushing 😄

sradevski:
@olivermrbl please check the last commit for the basic auth approach.

olivermrbl:
LGTM

PR #6522, opened 2024-02-27
feat(types): promotion module uses big number

riqwan:
what: 

- promotion modules (application method and campaign budget) uses big number fields instead of numeric fields
- refreshes the migrations to include new fields for big numbers
- adds a promotion_ prefix to promotion models
- uses `take: null` within the module to prevent the default pagination on performing actions within the module


srindom (approved):
LGTM

PR #6524, opened 2024-02-27
Feat(fulfillment): service provider registration + fulfillment management

adrien2p:
**What**
- Create the fulfillment manual package with a first iteration API 
- Create a new `AbstractFulfillmentProviderService` and `IFulfillmentProvider`
- Modify the module service interface to add new methods to manipulate the fulfillment and the communication with the external provider
  - create (no bulk)
  - cancel (no bulk)
  - update (no bulk)
  - list
  - listAndCount
  - retrieve
- Add new methods to the service provider service to include communication with the third party provider
  - get options
  - create
  - cancel
  - validate data
  - validate option
- Update/create interfaces and DTO's
- fix repository serializer to allow non entity to be passed without throwing
- split module tests into multiple files to simplify navigation
- Add integration tests to validate fulfillments manipulation and external provider loading + communication

FIXES CORE-1729
FIXES CORE-1785
FIXES CORE-1784
FIXES CORE-1766


adrien2p:
@olivermrbl in order to not end up with a too big pr I believe a review at this stage would be nice to have already

adrien2p:
@olivermrbl only git guardians is being red, should I do something?

packages/utils/src/fulfillment/provider.ts
```diff
@@ -0,0 +1,69 @@
...
+    throw Error("cancelFulfillment must be overridden by the child class")
+  }
+
+  async getFulfillmentDocuments(data) {
+    return []
+  }
+
+  async createOrder(fromData): Promise<any> {
+    throw Error("createReturn must be overridden by the child class")
+  }
```
olivermrbl: todo: I believe we can remove this method from the interface no?

packages/utils/src/fulfillment/provider.ts
```diff
@@ -0,0 +1,69 @@
...
+
+  async createReturn(fromData): Promise<any> {
+    throw Error("createReturn must be overridden by the child class")
+  }
```
olivermrbl: suggestion(non-blocking): Should we rename this to `createReturnFulfillment`

packages/types/src/fulfillment/provider.ts
```diff
@@ -0,0 +1,102 @@
...
+   *
+   * Get the documents for the given fulfillment data.
+   */
+  getFulfillmentDocuments(data: Record<string, unknown>): Promise<any>
+  /**
+   * @ignore
+   *
+   * Create an order for the given data.
+   */
+  createOrder(fromData: Record<string, unknown>): Promise<any>
```
olivermrbl: todo: rm

packages/types/src/fulfillment/provider.ts
```diff
@@ -0,0 +1,102 @@
...
+  /**
+   * @ignore
+   *
+   * Create a return for the given data.
+   */
+  createReturn(fromData: Record<string, unknown>): Promise<any>
```
olivermrbl: see other comment

packages/fulfillment/src/services/service-provider.ts
```diff
@@ -0,0 +1,103 @@
+import {
```
olivermrbl: question: Should we call this `FulfillmentProvider` as in V1, or are there some specific thoughts behind the choice of name here? Think it gets a bit messy with `ServiceProviderService`.

packages/fulfillment/src/services/fulfillment-module-service.ts
```diff
@@ -42,6 +47,7 @@ const generateMethodForModels = [
   ShippingProfile,
   ShippingOptionRule,
   ShippingOptionType,
+  // Not adding Fulfillment to not auto generate the methods under the hood and only provide the methods we want to expose8
```
olivermrbl, line 49: question: I can't remember, what methods do we not want to expose?

nitpicking: Also, minor typo in the last word -> `expose8`

packages/fulfillment/src/models/fulfillment-item.ts
```diff
@@ -100,10 +110,12 @@ export default class FulfillmentItem {
   @BeforeCreate()
   onCreate() {
     this.id = generateEntityId(this.id, "fulit")
+    this.fulfillment_id ??= this.fulfillment.id
   }
 
   @OnInit()
   onInit() {
     this.id = generateEntityId(this.id, "fulit")
+    this.fulfillment_id ??= this.fulfillment.id
```
olivermrbl, line 119: question: I guess this pattern will need to be applied in all modules?

packages/fulfillment/src/loaders/providers.ts
```diff
@@ -0,0 +1,86 @@
...
+  await moduleProviderLoader({
+    container,
+    providers: options?.providers || [],
+    registerServiceFn: registrationFn,
+  })
+
+  await syncDatabaseProviders({
+    container,
+    providerIdentifiersRegistrationKey: FulfillmentIdentifiersRegistrationName,
+    providerServiceRegistrationKey: lowerCaseFirst(ServiceProviderService.name),
```
olivermrbl: question: Why do we need to pass these as arguments? Couldn't we just use them directly in `syncDatabaseProviders`?

olivermrbl (commented):
Nice work. Submitting partial review

packages/fulfillment/src/services/service-provider.ts
adrien2p: I just kept what we aggreed on in the schema definition, fine to change it

packages/fulfillment/src/services/fulfillment-module-service.ts
adrien2p: 😂 thanks, we don't want createFulfillments, deleteFulfillments, softDeleteFulfillments etc that are all bulk and we agreed on non supporting bulk operation when it involves external providers

packages/fulfillment/src/models/fulfillment-item.ts
adrien2p: this is dependent on the relation you are creating, but if you are persisting from a parent then mikro orm assign to the relation and we transfert the id as we do not persist the relation

packages/fulfillment/src/services/fulfillment-module-service.ts
olivermrbl: That's right. Thanks for reminding me. I'll have to update payment sessions to be in line with this.

packages/fulfillment/src/models/fulfillment-item.ts
olivermrbl: I believe we have many of these relationships, which is why I thought we might have some cleaning up to do later.

adrien2p: yes definately, in some modules, the configuration is different and the repository and some service are managing the transposition, so yes some cleanup will be good for sure

packages/types/src/fulfillment/service.ts
```diff
@@ -337,6 +351,18 @@ export interface IFulfillmentModuleService extends IModuleService {
     sharedContext?: Context
   ): Promise<ShippingOptionTypeDTO>
 
+  /**
```
sradevski, line 354: NIT: Grouping the methods by model, rather than by function (eg. you put retrieve, list, create, and update next to each other, followed by retrieveGeoZone, listGeoZone...) makes it a bit easier to find things

packages/types/src/fulfillment/service.ts
```diff
@@ -337,6 +351,18 @@ export interface IFulfillmentModuleService extends IModuleService {
...
+   * Retrieve a fulfillment
+   * @param id
+   * @param config
+   * @param sharedContext
+   */
+  retrieveFulfillment(
```
sradevski, line 360: Should this be named `retrieve` only?

packages/types/src/fulfillment/service.ts
```diff
@@ -572,4 +622,31 @@ export interface IFulfillmentModuleService extends IModuleService {
   ): Promise<Record<string, string[]> | void>
 
   // TODO define needed soft delete/delete/restore methods
+
+  /**
+   * Retrieve the available fulfillment options for the given data.
+   */
+  retrieveFulfillmentOptions(
```
sradevski, line 629: To be consistent with the payments module, should we call these `Providers` rather than `Options`? Or are they different in any way?

packages/utils/src/fulfillment/provider.ts
```diff
@@ -0,0 +1,69 @@
...
+
+  static isFulfillmentService(obj) {
+    return obj?.constructor?._isFulfillmentService
+  }
+
+  getIdentifier() {
+    return (this.constructor as any).identifier
+  }
+
+  async getFulfillmentOptions(): Promise<Record<string, unknown>[]> {
```
sradevski, line 18: I can see the file is called provider, by the methods `Options`, so we definitely need to settle for one.

packages/fulfillment/src/services/service-provider.ts
sradevski: Agreed with Oli here

packages/fulfillment/src/services/fulfillment-module-service.ts
```diff
@@ -1074,6 +1207,82 @@ export default class FulfillmentModuleService<
...
+  ): Promise<FulfillmentDTO> {
+    const canceledAt = new Date()
+
+    let fulfillment = await this.fulfillmentService_.retrieve(
+      id,
+      {},
+      sharedContext
+    )
+
+    if (fulfillment.shipped_at) {
```
sradevski: Let's move all validations to a separate function, it's a bit easier to read when we separate validation from other business logic

packages/fulfillment/src/services/fulfillment-module-service.ts
```diff
@@ -211,6 +227,74 @@ export default class FulfillmentModuleService<
...
+    return await this.serviceProviderService_.getFulfillmentOptions(providerId)
+  }
+
+  @InjectManager("baseRepository_")
+  async listFulfillments(
+    filters: FulfillmentTypes.FilterableFulfillmentProps = {},
+    config: FindConfig<FulfillmentTypes.FulfillmentDTO> = {},
+    @MedusaContext() sharedContext: Context = {}
+  ): Promise<FulfillmentTypes.FulfillmentDTO[]> {
+    const fulfillments = await this.fulfillmentService_.list(
```
sradevski, line 262: What's the reason we have to override the list method here? Same with fulfillment options above?

packages/fulfillment/src/models/fulfillment.ts
```diff
@@ -88,28 +89,37 @@ export default class Fulfillment {
   @Property({ columnType: "jsonb", nullable: true })
   data: Record<string, unknown> | null = null
 
-  @Property({ columnType: "text" })
+  @ManyToOne(() => ServiceProvider, {
```
sradevski: Don't forget this one if we go with FulfillmentProvider 😄

packages/fulfillment/src/loaders/providers.ts
```diff
@@ -0,0 +1,86 @@
...
+    const providersToCreate: any[] = []
+    for (const identifier of providerIdentifiers) {
+      if (loadedProvidersMap.has(identifier)) {
+        continue
+      }
+
+      providersToCreate.push({ id: identifier })
+    }
+
+    await providerService.create(providersToCreate)
```
sradevski, line 83: Should this be an upsert instead?

sradevski (commented):
From a user's perspective, how would loading plugins look like?

Also, do we want to keep the plugins as a flat list of folders (eg. `fulfillment-manual`) in the repo, or do we want to prefix/suffix them with something so we distinguish between modules and plugins?

packages/types/src/fulfillment/service.ts
adrien2p: no the main entity entry point is fulfillmentSet which differ from fulfillment

packages/types/src/fulfillment/service.ts
adrien2p: they are different, the options are return by the provider, for example, webshiper will return a list of available options

packages/types/src/fulfillment/service.ts
sradevski: Ah ok, that confused me for a second 😄

packages/utils/src/fulfillment/provider.ts
adrien2p: this is just the base provider class a user need to use to create a fulfillment provider, but the options are the fulfillment options, not the provider "options" in the sense of config but in the sense of fulfillment, for example
- ups
- laposte
- etc
they are the available options

packages/fulfillment/src/services/fulfillment-module-service.ts
adrien2p: I do not override the method here, I did explicitly not generated the method for this entity as mention in the comment I ve put at the top of the file :). You can read the explanation I gave to oli [here](https://github.com/medusajs/medusa/pull/6524#discussion_r1511251963)

packages/fulfillment/src/models/fulfillment.ts
adrien2p: yes ofcourse thanks :) but since I test before pushing I can't miss it ahah

packages/types/src/fulfillment/service.ts
sradevski: Ah ok, got it. @adrien2p in payments I see we use `methods` for it, maybe we can align the two. It's not necessary, but if it makes sense, why not.

packages/fulfillment/src/loaders/providers.ts
adrien2p: yes we could, in that case I can also simplify this function I believe

packages/fulfillment/src/services/fulfillment-module-service.ts
sradevski: Got it, I missed that comment while reviewing 😄

adrien2p:
> From a user's perspective, how would loading plugins look like?
> 
> Also, do we want to keep the plugins as a flat list of folders (eg. `fulfillment-manual`) in the repo, or do we want to prefix/suffix them with something so we distinguish between modules and plugins?

@sradevski loading module providers looks like [this](https://github.com/medusajs/medusa/pull/6524/files#diff-9cb9ba71f5d6e4bcf698a868cd344bb08e830a594939824384888cf7555fda10R59)

sradevski:
> @sradevski loading module providers looks like [this](https://github.com/medusajs/medusa/pull/6524/files#diff-9cb9ba71f5d6e4bcf698a868cd344bb08e830a594939824384888cf7555fda10R59)

@adrien2p great thanks, good with me. How about the location/placement of these plugins, should we make it a bit more obvious that they are indeed plugins? Or do you think just prefixing them with the module they are a plugin for is enough?

adrien2p:
> > @sradevski loading module providers looks like [this](https://github.com/medusajs/medusa/pull/6524/files#diff-9cb9ba71f5d6e4bcf698a868cd344bb08e830a594939824384888cf7555fda10R59)
> 
> @adrien2p great thanks, good with me. How about the location/placement of these plugins, should we make it a bit more obvious that they are indeed plugins? Or do you think just prefixing them with the module they are a plugin for is enough?

I don't have a strong opinion on this for now, I believe there is some cleanup between the plugins, the module provider plugins and the module todo. Also, maybe reorganising the monorepo structure could help as we start having so many thing it become hard to navigate

packages/types/src/fulfillment/service.ts
adrien2p: I believe we could indeed as we have the notion of shipping method in v1 for something else, wdyt @olivermrbl

olivermrbl: `retrieveFulfillmentMethods` would be OK for me. 

I believe the reason we use options is that these are associated with a `ShippingOption`, not a `ShippingMethod`. We use the concept to define the "service" a shipping option will eventually be fulfilled with at the provider e.g. "Express" or "Access point". You pass the fulfillment option data when creating a shipping option. With that in mind, both method signatures could make sense.

cc @srindom

packages/types/src/fulfillment/provider.ts
```diff
@@ -0,0 +1,102 @@
...
+  ): Promise<any>
+  /**
+   * @ignore
+   *
+   * Create a fulfillment for the given data.
+   */
+  createFulfillment(
+    data: object,
+    items: object[],
+    order: object,
```
olivermrbl, line 54: thought(non-blocking): I wonder if we should update the signature to expect a single object with the required properties?

This can be tackled in a follow-up PR, when we start to build out an actual provider.

packages/fulfillment/src/services/fulfillment-module-service.ts
```diff
@@ -504,6 +588,55 @@ export default class FulfillmentModuleService<
...
+
+    let fulfillmentThirdPartyData!: any
+    try {
+      fulfillmentThirdPartyData =
+        await this.serviceProviderService_.createFulfillment(
+          provider_id,
+          fulfillmentData || {},
+          items.map((i) => i),
+          order,
+          fulfillmentRest
```
olivermrbl, line 618: thought: See other comment. Perhaps, we should rethink the shape of objects passed to the provider here. For example, the properties we need from `order` could be extracted, so it is decoupled from the actual order concept.

packages/types/src/fulfillment/provider.ts
adrien2p: yes that was my take on that too, I ve put a todo somewhere for that

packages/fulfillment/src/services/fulfillment-module-service.ts
adrien2p: yes I think we need to rethink the entire API, I ve a todo somewhere in the abstract of the interface to rework all dtos and stuff

packages/types/src/fulfillment/service.ts
adrien2p: I ll keep this one for later until we agree on something

adrien2p:
@olivermrbl should I slap an automerge now?

sradevski (approved):

olivermrbl:
@adrien2p sure!

PR #6525, opened 2024-02-27
chore(utils): Improve big number decorator

adrien2p:
**What**
- Re add the ability to use the original column name as part of a select in both select in and joined strategy
- The undefined error message will now display the original column name instead of the underscored one


adrien2p:
@olivermrbl this should solve your message too

carlos-r-l-rodrigues (approved):
🤝

PR #6526, opened 2024-02-27
fix(medusa-cli): Logger arguments

xyzones:
- winston.Logger.log requires level as param

### DIFF
--- a/packages/medusa-cli/src/reporter/index.ts
+++ b/packages/medusa-cli/src/reporter/index.ts
@@ -311,8 +311,16 @@ export class Reporter {
    * A wrapper around winston's log method.
    */
   log = (...args) => {
-    // @ts-ignore
-    this.loggerInstance_.log(...args)
+    if (args.length > 1) {
+      // @ts-ignore
+      this.loggerInstance_.log(...args)
+    } else {
+      let message = args[0]
+      this.loggerInstance_.log({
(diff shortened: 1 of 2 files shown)


adrien2p (commented):
Should we instead update the log method from the logger to pass (`info', ... args) wdyt? cc @olivermrbl

xyzones:
Well, we could extend the
Log wrapper however as we do not know where this is used defaulting to info might not be the best. 

I would maybe do something like 
```typescript
if (args.length > 1) { 
  logger.log(...args)
} else {
  logger.info(args[0])
}
```

adrien2p:
> Well, we could extend the Log wrapper however as we do not know where this is used defaulting to info might not be the best.
> 
> I would maybe do something like
> 
> ```ts
> if (args.length > 1) { 
>   logger.log(...args)
> } else {
>   logger.info(args[0])
> }
> ```

yeah lets go with

```ts
if (args.length > 1) { 
  return logger.log(...args)
}

logger.info(args[0])
```

xyzones:
I have updated the medusa-cli reporter that is used by the migrate-inventory-items

olivermrbl:
LGTM!

gancena (approved):

sradevski (approved):

PR #6527, opened 2024-02-27
feat: Add payment collection creation for cart

olivermrbl:
**What**
- Add `POST /store/carts/:id/payment-collection` API route
- Add `CreatePaymentCollectionForCartWorkflow` workflow
   - `createPaymentCollectionsStep`
   - `linkCartAndPaymentCollectionsStep`
   - `retrieveCartStep`

**A note on checkout flows**
This is the first piece in assembling the checkout flow. 

At a high level, I imagine the flow will be as follows:
1. Create cart (Carts API)
2. Add item to cart (Carts API)
3. Create payment collection for cart (Carts API)
4. Select and create payment session (Payment Sessions API)
5. Authorize payment session (Payment Sessions API)
6. Complete cart (Carts API)

Step 4 and 5 must be repeated every time a cart update is performed. We destroy the currently active payment sessions if the cart is updated. This is to ensure that the correct payment is collected for a cart.

The flow is illustrated below:
image

What's missing?
- Totals decoration for carts
- More tests

### DIFF
--- a/packages/core-flows/src/definition/cart/workflows/create-payment-collection-for-cart.ts
+++ b/packages/core-flows/src/definition/cart/workflows/create-payment-collection-for-cart.ts
@@ -2,7 +2,11 @@ import {
   CartDTO,
   CreatePaymentCollectionForCartWorkflowInputDTO,
 } from "@medusajs/types"
-import { WorkflowData, createWorkflow } from "@medusajs/workflows-sdk"
+import {
+  WorkflowData,
+  createWorkflow,
+  transform,
+} from "@medusajs/workflows-sdk"
 import { retrieveCartStep } from "../steps"
 import { createPaymentCollectionsStep } from "../steps/create-payment-collection"
 import { linkCartAndPaymentCollectionsStep } from "../steps/link-cart-payment-collection"
@@ -16,16 +20,16 @@ export const createPaymentCollectionForCartWorkflow = createWorkflow(
   ): WorkflowData<CartDTO> => {
     const created = createPaymentCollectionsStep([input])
 
-    const linksInput = {
+    const link = transform({ cartId: input.cart_id, created }, (data) => ({
       links: [
         {
-          cart_id: input.cart_id,
-          payment_collection_id: created[0].id,
+          cart_id: data.cartId,
+          payment_collection_id: data.created[0].id,
         },
       ],
-    }
+    }))
 
-    linkCartAndPaymentCollectionsStep(linksInput)
+    linkCartAndPaymentCollectionsStep(link)
 
     const cart = retrieveCartStep({ id: input.cart_id })
 


packages/core-flows/src/definition/cart/workflows/create-payment-collection-for-cart.ts
```diff
@@ -0,0 +1,38 @@
...
+    const link = transform({ cartId: input.cart_id, created }, (data) => ({
+      links: [
+        {
+          cart_id: data.cartId,
+          payment_collection_id: data.created[0].id,
+        },
+      ],
+    }))
+
+    linkCartAndPaymentCollectionsStep(link)
```
adrien2p, line 32:
```suggestion
    const linksInput = {
      links: [
        {
          cart_id: input.cart_id,
          payment_collection_id: created[0].id,
        },
      ],
    }

    linkCartAndPaymentCollectionsStep(linksInput)
```

packages/medusa/src/api-v2/store/carts/[id]/payment-collections/route.ts
```diff
@@ -0,0 +1,51 @@
...
+  })
+
+  if (Array.isArray(errors) && errors[0]) {
+    throw errors[0].error
+  }
+
+  const query = remoteQueryObjectFromString({
+    entryPoint: "cart",
+    fields: defaultStoreCartFields,
+  })
```
Comment (adrien2p) [inline packages/medusa/src/api-v2/store/carts/[id]/payment-collections/route.ts]:
```suggestion
  const query = remoteQueryObjectFromString({
    entryPoint: "cart",
    variables: { id: req.pqrqms.id },
    fields: defaultStoreCartFields,
  })
```

packages/medusa/src/api-v2/store/carts/[id]/payment-collections/route.ts
```diff
@@ -0,0 +1,51 @@
...
+
+  ;[cart] = await remoteQuery(query, {
+    cart: { id: req.params.id },
+  })
```
Comment (adrien2p) [inline packages/medusa/src/api-v2/store/carts/[id]/payment-collections/route.ts]:
```suggestion
  ;[cart] = await remoteQuery(query)
```

packages/medusa/src/api-v2/store/carts/[id]/payment-collections/route.ts
```diff
@@ -0,0 +1,51 @@
...
+  )
+
+  const { errors } = await workflow.run({
+    input: {
+      cart_id: req.params.id,
+      region_id: cart.region_id,
+      currency_code: cart.currency_code,
+      amount: cart.total ?? 0, // TODO: This should be calculated from the cart when totals decoration is introduced
+    },
+    throwOnError: false,
```
Comment (adrien2p) [inline packages/medusa/src/api-v2/store/carts/[id]/payment-collections/route.ts:34]: discussion: what do you think of having an onError callback handler instead of this configuration?
it would look like:

```ts
import { workflowApiErrorHandler } from '@medusajs/utils'
const { errors } = await workflow.run({
    input: {
      cart_id: req.params.id,
      region_id: cart.region_id,
      currency_code: cart.currency_code,
      amount: cart.total ?? 0, // TODO: This should be calculated from the cart when totals decoration is introduced
    },
   onError: workflowApiErrorHandler
})
```

```ts
export function workflowApiErrorHandler(errors) {
  if  (Array.isArray(errors) && errors[0]) {
    throw errors[0].error
  }
}
```

packages/medusa/src/api-v2/store/carts/query-config.ts
```diff
@@ -42,6 +42,12 @@ export const defaultStoreCartFields = [
   "region.name",
   "region.currency_code",
   "sales_channel_id",
+
+  // TODO: To be updated when payment sessions are introduces in the Rest API
+  "payment_collection",
```
adrien2p: suggestion: rm, first level relation are not needed if we specifiy relation plus columns

packages/medusa/src/api-v2/store/carts/[id]/payment-collections/route.ts
Comment (olivermrbl) [reply packages/medusa/src/api-v2/store/carts/[id]/payment-collections/route.ts:34]: I think I like this idea. It is def. better than what we do now. 

We could export good error handlers for when we use workflows in different places:
- Rest API -> workflowsApiRouteErrorHandler
- Modules- > workflowsModulesErrorHandler
- ...


However, we should still find a good way to throw an appropriate Medusa error in an API Route. Right now, if an error is thrown in a workflow executed in an API Route, I believe we respond with a 500 "unknown error", because our error handler doesn't know how to treat workflow errors.

Comment (adrien2p) [reply packages/medusa/src/api-v2/store/carts/[id]/payment-collections/route.ts:34]: no in theory if you re throw the error it should be a medusa error if it was a medusa error thrown, you can check what the error is to be sure

Comment (olivermrbl) [reply packages/medusa/src/api-v2/store/carts/[id]/payment-collections/route.ts:34]: Let's punt this discussion for when we've built some more workflows. Think it would be good to have experienced some more cases before we proceed with a solution

packages/medusa/src/api-v2/store/carts/query-config.ts
olivermrbl: Apparently this didn't doesn't work. My latest commit fails on an undefined payment collection in the cart response 🤔

packages/core-flows/src/definition/cart/workflows/create-payment-collection-for-cart.ts
olivermrbl: Am I wrong in thinking this is not possible? The `created[0].id` is not resolved yet

integration-tests/modules/__tests__/cart/store/cart.workflows.spec.ts
```diff
@@ -669,4 +675,142 @@ describe("Carts workflows", () => {
...
+        items: [
+          {
+            quantity: 1,
+            unit_price: 5000,
+            title: "Test item",
+          },
+        ],
+      })
+
+      await createPaymentCollectionForCartWorkflow(appContainer).run({
```
sradevski, line 698: Should we only rely on REST API endpoints in the test, considering it's an integration test? Or is it because the endpoints are not yet available?

sradevski (commented):
Code-wise looks good, I just have a few questions:
1. Should we create the payment collection at the same time we create the cart (in a workflow when you hit the `POST /carts` endpoint)? It might be a bit wasteful, but at least it simplifies things for the user
2. Is there a risk of race condition between completing the cart payment and someone adding an item to the cart in the meantime? Do we even need to worry about it?
3. I remember there was a feature request for this, and I needed the same thing :D I wanted to allow creating an order without paying, and then I the payment can be collected afterwards. Is this something we'll be doing in v2?
4. Can a step trigger a workflow? What I am thinking is that the `update-cart` workflow should trigger the payment session deletion workflow.

sradevski:
NIT: The steps you described and the diagram differ a bit, in the diagram we only create the payment session after you select the provider, but in the description's step 4 you wrote we will create the payment sessions. I assume the diagram is the correct flow we want?

olivermrbl:
@sradevski, answering in-line here:

> Should we create the payment collection at the same time we create the cart (in a workflow when you hit the POST /carts endpoint)? It might be a bit wasteful, but at least it simplifies things for the user

That could make sense. If you don't mind, I'll leave it as is until the full checkout flow has been built. There are still a few unknowns about the flow that I'd like to have clarified before we proceed with optimizing for the end customer.

> Is there a risk of race condition between completing the cart payment and someone adding an item to the cart in the meantime? Do we even need to worry about it?

Too early to tell if we'll have an issue with this. This is one of the unknowns mentioned above. My current idea is to destroy the sessions every time an update is made. With that in mind, I guess we could end up in a situation where a customer attempts to complete and update the cart at the same time, leading to a race condition between the destruction and completion of sessions. Let's address this when building the cart-completion flow.

> I remember there was a feature request for this, and I needed the same thing :D I wanted to allow creating an order without paying, and then I the payment can be collected afterwards. Is this something we'll be doing in v2?

That should be possible, given the cart-completion logic lives in a workflow. Users can either build their own or modify ours. The latter would require us to open the one we ship to customizations, and I am unsure if we want to do that in the first iteration. I imagine those customizations/injection zones would have to be controlled so users cannot mess up their cart-completion flow, which is one of the most business-critical ones.

> Can a step trigger a workflow? What I am thinking is that the update-cart workflow should trigger the payment session deletion workflow.

Yes, this is what I plan to do.

integration-tests/modules/__tests__/cart/store/cart.workflows.spec.ts
olivermrbl: I intentionally split the integration tests into API (`/store/carts.spec.ts`) and Workflows (`/store/cart.workflows.ts`). It didn't feel right to place tests for different inputs and compensation flows of workflows in the API folder.

adrien2p:
I would add that we also have the hook API on the workflow composer 💪

olivermrbl:
> I would add that we also have the hook API on the workflow composer 💪

Yup, that's the one I was referring to with the "injection zones" comment. Think that could make sense to use in the cart-completion workflows

integration-tests/modules/__tests__/cart/store/cart.workflows.spec.ts
sradevski: Ah missed the .workflows. part, 👍

sradevski (approved):

fPolic (approved):

PR #6531, opened 2024-02-28
feat(utils): consolidate promotion utils + refactor + fixes

riqwan:


packages/promotion/src/services/promotion-module.ts
```diff
@@ -251,6 +253,8 @@ export default class PromotionModuleService<
           sharedContext
         )
 
+    // Promotions we need to apply includes all the codes thats passed as an argument
+    // of this method, along with any automatic promotions that can be applied to the context
```
sradevski:
```suggestion
    // Promotions we need to apply includes all the codes that are passed as an argument
    // to this method, along with any automatic promotions that can be applied to the context
```

packages/promotion/src/utils/compute-actions/line-items.ts
```diff
@@ -0,0 +1,183 @@
...
+) {
+  if (!context) {
+    throw new MedusaError(
+      MedusaError.Types.INVALID_DATA,
+      `"${contextKey}" should be present as an array in the context for computeActions`
+    )
+  }
+}
+
+export function getComputedActionsForItems(
```
sradevski, line 27: It feels like the logic can just be here rather than creating a new function, since all we do here is just call validateContext before we call the actual function that holds the logic. I leave it up to you, it just feels like an unnecessary additional step

packages/promotion/src/utils/compute-actions/line-items.ts
```diff
@@ -0,0 +1,183 @@
...
+    | PromotionTypes.ComputeActionContext[ApplicationMethodTargetType.ITEMS]
+    | PromotionTypes.ComputeActionContext[ApplicationMethodTargetType.SHIPPING_METHODS],
+  appliedPromotionsMap: Map<string, number>,
+  allocationOverride?: ApplicationMethodAllocationValues
+): PromotionTypes.ComputeActions[] {
+  const { application_method: applicationMethod } = promotion
+  const allocation = applicationMethod?.allocation! || allocationOverride
+  const computedActions: PromotionTypes.ComputeActions[] = []
+  const applicableItems = getValidItemsForPromotion(items, promotion)
+  const isTargetShippingMethod =
```
sradevski: Adding some newline space eg. here can aid with readability.

packages/promotion/integration-tests/__tests__/services/promotion-module/compute-actions.spec.ts
```diff
@@ -3424,13 +3424,13 @@ describe("Promotion Service: computeActions", () => {
           {
             action: "addShippingMethodAdjustment",
             shipping_method_id: "shipping_method_express",
-            amount: 83.33333333333331,
+            amount: 166.66666666666666,
```
srindom, line 3427: q: what happened here?

riqwan: There was a bug in the calculations that was fixed here. It was an incorrect test to begin with. It incorrectly inferred that part of the value was already used.

sradevski (approved):

srindom (approved):
LGTM

PR #6533, opened 2024-02-28
feat(tax): add endpoints to create tax regions and tax rates

srindom:
**What**
Adds:
- POST /admin/tax-regions
- POST /admin/tax-rates
- GET /admin/tax-rates
- `createTaxRegionsWorkflow`
- `createTaxRatesWorkflow`


packages/medusa/src/api-v2/admin/tax-regions/validators.ts
```diff
@@ -0,0 +1,122 @@
...
+}) {
+  /**
+   * Search parameter for regions.
+   */
+  @IsString({ each: true })
+  @IsOptional()
+  id?: string | string[]
+
+  /**
+   * Filter by currency code
```
sradevski: The comment here and below are incorrect 🙈

packages/medusa/src/api-v2/admin/tax-regions/validators.ts
```diff
@@ -0,0 +1,122 @@
...
+  $or?: AdminGetTaxRegionsParams[]
+}
+
+class CreateDefaultTaxRate {
+  @IsOptional()
+  @IsType([Number])
+  rate?: number | null
+
+  @IsOptional()
+  @IsType([String])
```
sradevski:
```suggestion
@IsString
```

packages/medusa/src/api-v2/admin/tax-regions/validators.ts
```diff
@@ -0,0 +1,122 @@
...
+
+  @IsOptional()
+  @ValidateNested({ each: true })
+  @Type(() => AdminGetTaxRegionsParams)
+  $or?: AdminGetTaxRegionsParams[]
+}
+
+class CreateDefaultTaxRate {
+  @IsOptional()
+  @IsType([Number])
```
sradevski:
```suggestion
@IsNumber
```

packages/types/src/tax/common.ts
```diff
@@ -22,6 +22,18 @@ export interface TaxRateDTO {
...
    */
   metadata: Record<string, unknown> | null
+  /**
+   * The id of the Tax Region the rate is associated with.
+   */
+  tax_region_id: string
+  /**
+   * Flag to indicate if the Tax Rate should be combined with parent rates.
+   */
+  is_combinable: boolean
```
sradevski, line 32: I am completely clueless on this one, what does it mean to be able to combine a tax rate with a parent rate? If this is explained somewhere in Notion feel free to just share a link to that 😄

packages/medusa/src/api-v2/admin/tax-rates/route.ts
```diff
@@ -0,0 +1,58 @@
...
+export const GET = async (
+  req: AuthenticatedMedusaRequest,
+  res: MedusaResponse
+) => {
+  const remoteQuery = req.scope.resolve("remoteQuery")
+
+  const variables = { id: req.params.id }
+
+  const queryObject = remoteQueryObjectFromString({
+    entryPoint: "tax_rates",
```
sradevski: We have it as plural here and singular above, I think using just one for consistency is better (I assume due to the aliases this works as well, right?)

packages/core-flows/src/tax/steps/create-tax-regions.ts
```diff
@@ -0,0 +1,31 @@
...
+      created.map((region) => region.id)
+    )
+  },
+  async (createdIds, { container }) => {
+    if (!createdIds?.length) {
+      return
+    }
+
+    const service = container.resolve<ITaxModuleService>(
+      ModuleRegistrationName.REGION
```
sradevski:
```suggestion
      ModuleRegistrationName.TAX
```

packages/types/src/tax/common.ts
srindom: In some countries, multiple authorities can charge tax. For example, in Canada, you pay a federal sales tax and a provincial sales tax. 

The way that would be set up would be something like:
- COUNTRY Region
  - Tax Rate: X
  - PROVINCE Region
    - parent_id: state region id
    - Tax Rate: Y

The is_combinable flag indicates whether the end customer should be charged X+Y or just Y. 

Does this make more sense?

sradevski: Yeah makes perfect sense, thanks @srindom !

PR #6536, opened 2024-02-28
feat: Add currency module and remove currency models from region and pricing modules

sradevski:
What:
- Creates a new currency module
- Removes currency model from the pricing module
- Removes currency model from region module

### DIFF
--- a/packages/currency/integration-tests/__tests__/currency-module-service.spec.ts
+++ b/packages/currency/integration-tests/__tests__/currency-module-service.spec.ts
@@ -11,9 +11,187 @@ moduleIntegrationTestRunner({
     service,
   }: SuiteOptions<ICurrencyModuleService>) => {
     describe("Currency Module Service", () => {
-      describe("noop", function () {
-        it("should run", function () {
-          expect(true).toBe(true)
+      describe("list", () => {
+        it("list currencies", async () => {
+          const currenciesResult = await service.list({}, { take: null })
+          expect(currenciesResult).toEqual(
+            expect.arrayContaining([
+              expect.objectContaining({
+                code: "cad",
+                name: "Canadian Dollar",
+              }),
+              expect.objectContaining({
+                code: "usd",
+                name: "US Dollar",
+              }),
+            ])
+          )
+        })
+
+        it("list currencies by code", async () => {
+          const currenciesResult = await service.list(
+            { code: ["usd"] },
+            { take: null }
+          )
+
+          expect(currenciesResult).toEqual([
+            expect.objectContaining({
+              code: "usd",
+              name: "US Dollar",
+            }),
+          ])
+        })
+
+        it("list currencies by code regardless of case-sensitivity", async () => {
+          const currenciesResult = await service.list(
+            { code: ["Usd"] },
+            { take: null }
+          )
+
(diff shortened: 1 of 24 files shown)


packages/currency/integration-tests/__tests__/currency-module-service.spec.ts
```diff
@@ -0,0 +1,185 @@
...
+jest.setTimeout(100000)
+
+moduleIntegrationTestRunner({
+  moduleName: Modules.CURRENCY,
+  testSuite: ({
+    MikroOrmWrapper,
+    service,
+  }: SuiteOptions<ICurrencyModuleService>) => {
+    describe("Currency Module Service", () => {
+      // TODO: Remove this and load currencies in loader
```
sradevski: This is the only todo that needs to be done before the PR is merged, but I decided to open it in a non-draft mode so others have the chance to review it in the meantime. The currencies will be loaded through a loader based on the final discussion we had with Adrien.

packages/currency/src/models/currency.ts
srindom, line 1: thought: would it make sense to include decimals here? e.g. 2 for most, 0 for jpy, etc?

sradevski: @srindom it definitely makes sense, but I thought of not adding things in this PR, so as to keep the model unchanged (and the number of changes to a minimum).

Once we have a use for it it will be really straightforward to add it

packages/currency/src/loaders/initial-data.ts
```diff
@@ -0,0 +1,23 @@
...
+import { Currency } from "@models"
+
+export default async ({
+  container,
+  options,
+}: LoaderOptions<
+  | ModulesSdkTypes.ModuleServiceInitializeOptions
+  | ModulesSdkTypes.ModuleServiceInitializeCustomDataLayerOptions
+>): Promise<void> => {
+  const {
```
sradevski: @adrien2p this works, but I'd like to hear how we can improve it if possible (eg. how to get the currencyModuleService name, etc.

packages/currency/src/models/currency.ts
srindom: Makes sense!

srindom (approved):
LGTM

packages/currency/src/migrations/InitialSetup20240228133303.ts
```diff
@@ -0,0 +1,9 @@
+import { Migration } from "@mikro-orm/migrations"
+
+export class InitialSetup20240228133303 extends Migration {
+  async up(): Promise<void> {
+    this.addSql(`create table "currency" 
+    ("code" text not null, "symbol" text not null, "symbol_native" text not null, "name" text not null, 
+    constraint "currency_pkey" primary key ("code"));`)
```
srindom: todo: this should probably be changed so it doesn't break instances with v1 tables - i.e. `create table if not exists`.

sradevski: Good point, this was generated by migraiton:initial and I didn't double-check it, I'll add it now!

sradevski:
@srindom done!

srindom (approved):

PR #6537, opened 2024-02-28
feat(tax): add support for updating tax rates

srindom:


packages/tax/src/services/tax-module-service.ts
```diff
@@ -196,6 +196,27 @@ export default class TaxModuleService<
...
+    data: TaxTypes.UpsertTaxRateDTO,
+    sharedContext?: Context
+  ): Promise<TaxTypes.TaxRateDTO>
+
+  @InjectTransactionManager("baseRepository_")
+  async upsert(
+    data: TaxTypes.UpsertTaxRateDTO | TaxTypes.UpsertTaxRateDTO[],
+    @MedusaContext() sharedContext: Context = {}
+  ): Promise<TaxTypes.TaxRateDTO | TaxTypes.TaxRateDTO[]> {
+    const result = await this.taxRateService_.upsert(data, sharedContext)
```
sradevski, line 213: This will vary from module to module, but many times we'd like to do some validation and normalization to the input before doing update/create, so calling the internal create_ and update_ methods might be better. It might be fine here though, just wanted to mention it

sradevski (approved):

PR #6539, opened 2024-02-28
feat(medusa,core-flows): update cart adjustments on item updates

riqwan:


packages/core-flows/src/definition/cart/steps/refresh-cart-promotions.ts
```diff
@@ -0,0 +1,30 @@
...
+    | PromotionActions.REPLACE
+}
+
+export const refreshCartPromotionsStepId = "refresh-cart-promotions"
+export const refreshCartPromotionsStep = createStep(
+  refreshCartPromotionsStepId,
+  async (data: StepInput, { container }) => {
+    const { promo_codes = [], id, action = PromotionActions.ADD } = data
+
+    await updateCartPromotionsWorkflow(container).run({
```
olivermrbl, line 20: love this! 🎆

packages/core-flows/src/definition/cart/workflows/update-cart.ts
```diff
@@ -79,34 +62,24 @@ export const updateCartWorkflow = createWorkflow(
...
-      removeShippingMethodAdjustmentsStep({
-        shippingMethodAdjustmentIdsToRemove,
-      })
-    )
-
-    parallelize(
-      createLineItemAdjustmentsStep({ lineItemAdjustmentsToCreate }),
-      createShippingMethodAdjustmentsStep({ shippingMethodAdjustmentsToCreate })
-    )
+    const retrieveCartInput = {
```
olivermrbl, line 71: question: Can we do this? I didn't know that. Thought everything had to be either a step or transformer.

riqwan: We changed it recently to allow plain objects as well. You only need the transformer if its an unresolved object.

olivermrbl: That's great

carlos-r-l-rodrigues: one can't keep up with so many improvements 😆

olivermrbl: Haha, let's keep it that way 🏎️

PR #6546, opened 2024-02-29
feat: Modify api key and sales channel link to use modules and add test

sradevski:


integration-tests/modules/__tests__/link-modules/publishable-key-sales-channel.spec.ts
```diff
@@ -0,0 +1,105 @@
...
+    await db.shutdown()
+    await shutdownServer()
+  })
+
+  afterEach(async () => {
+    const db = useDb()
+    await db.teardown()
+  })
+
+  it("should query carts, sales channels, customers, regions with remote query", async () => {
```
olivermrbl: nitpicking:
```suggestion
  it("should query api keys and sales channels with remote query", async () => {
```

packages/medusa/src/api-v2/admin/api-keys/validators.ts
```diff
@@ -34,6 +34,13 @@ export class AdminGetApiKeysParams extends extendedFindParamsMixin({
   @IsOptional()
   title?: string | string[]
 
+  /**
+   * Filter by title
```
olivermrbl: nitpicking:
```suggestion
   * Filter by token
```

packages/link-modules/src/definitions/publishable-api-key-sales-channel.ts
```diff
@@ -48,16 +49,5 @@ export const PublishableApiKeySalesChannel: ModuleJoinerConfig = {
         isList: true,
       },
     },
-    {
```
olivermrbl, line 51: question: Should this be migrate to use the sales channel module instead of removed?

olivermrbl (approved):
LGTM

packages/link-modules/src/definitions/publishable-api-key-sales-channel.ts
sradevski: @olivermrbl what I was thinking is that there is no point going from a sales channel to an API key, which is why I removed the relationship. Let me know if you think differently

integration-tests/modules/__tests__/link-modules/publishable-key-sales-channel.spec.ts
sradevski: copy-paste issues 😄

packages/link-modules/src/definitions/publishable-api-key-sales-channel.ts
olivermrbl: You're right. I had a similar thought for Sales Channels and Carts. However, I actually ended up including the extension, because we don't really lose anything by doing so, and therefore I thought we might as well. Who knows, someone might need this one day 😅 

That said, fine to keep it as is.

carlos-r-l-rodrigues: +1
Having the graph definition and being able to traverse it from any entry point might be useful in the future.

sradevski: I don't mind adding it back, let's add it and revisit.

PR #6549, opened 2024-02-29
feat: Create payment sessions

olivermrbl:
~~Opening a draft PR to discuss a couple of implementation details that we should align on~~

**What**

Add workflow and API endpoint for creating payment sessions for a payment collection. Endpoint is currently `POST /store/payment-collection/:id/payment-sessions`. I suggested an alternative in a comment below.

Please note, we intentionally do not want to support creating payment sessions in bulk, as this would become a mess when having to manage multiple calls to third-party providers.

### DIFF
--- a/packages/medusa/src/api-v2/store/carts/route.ts
+++ b/packages/medusa/src/api-v2/store/carts/route.ts
@@ -1,22 +1,19 @@
+import { createCartWorkflow } from "@medusajs/core-flows"
+import { CreateCartWorkflowInputDTO } from "@medusajs/types"
+import { remoteQueryObjectFromString } from "@medusajs/utils"
 import {
   AuthenticatedMedusaRequest,
   MedusaResponse,
 } from "../../../types/routing"
-
-import { createCartWorkflow } from "@medusajs/core-flows"
-import { CreateCartWorkflowInputDTO } from "@medusajs/types"
-import { remoteQueryObjectFromString } from "@medusajs/utils"
 import { defaultStoreCartFields } from "../carts/query-config"
 
 export const POST = async (
   req: AuthenticatedMedusaRequest<CreateCartWorkflowInputDTO>,
   res: MedusaResponse
 ) => {
-  const workflowInput = req.validatedBody
-
-  // If the customer is logged in, we auto-assign them to the cart
-  if (req.auth?.actor_id) {
-    workflowInput.customer_id = req.auth.actor_id
+  const workflowInput = {
+    ...req.validatedBody,
+    customer_id: req.auth?.actor_id,
   }
 
   const { result, errors } = await createCartWorkflow(req.scope).run({


packages/payment/src/services/payment-module.ts
```diff
@@ -207,38 +207,33 @@ export default class PaymentModuleService<
     data: CreatePaymentSessionDTO,
     @MedusaContext() sharedContext?: Context
   ): Promise<PaymentSessionDTO> {
-    const created = await this.paymentSessionService_.create(
-      {
+    const { result, errors } = await paymentSessionWorkflow(
+      this.__container__ as any
```
olivermrbl: discussion: Using `this__container__` is not super elegant. We should find a nicer way of passing the container. Do you have any ideas?

One alternative that came to mind was: 

```
await paymentSessionWorkflow({
  paymentProviderService: this.paymentProviderService_,
  paymentSessionService: this.paymentSessionService_,
})
```

I.e. passing the required dependencies explicitly to the workflow runner, but I don't know if I like that more.

Ideally, we don't need to pass it at all from within a service, that already has the injected dependencies needed to carry out the flow.

cc @carlos-r-l-rodrigues @adrien2p

packages/orchestration/src/workflow/local-workflow.ts
```diff
@@ -50,10 +50,13 @@ export class LocalWorkflow {
...
     let container
-
     if (!Array.isArray(modulesLoaded) && modulesLoaded) {
       if (!("cradle" in modulesLoaded)) {
-        container = createMedusaContainer(modulesLoaded)
+        container = createMedusaContainer()
+
+        for (const [name, registration] of Object.entries(modulesLoaded)) {
+          container.register(name, asValue(registration))
+        }
```
olivermrbl: comment: I can't tell you exactly why I needed to add this workaround, but it seems the `awilix.createContainer` call in the `createMedusaContainer` util failed to properly handle the conversion in this specific case:

https://github.com/medusajs/medusa/blob/e61c46c36da51d24e2294ce2b1d5b568bfe9526c/packages/utils/src/common/medusa-container.ts#L43-L56

The one conversion that happens on line 44.

@carlos-r-l-rodrigues, this is what I ended up doing. Let me know what you think.

For context, this case of the `modulesLoaded` state is reached when using the workflow in a service and passing `__this.container__`.

packages/payment/src/services/payment-module.ts
```diff
@@ -207,38 +207,33 @@ export default class PaymentModuleService<
...
+    }
 
-      await this.paymentSessionService_.update(
-        {
-          id: created.id,
-          data: sessionData,
-        },
-        sharedContext
-      )
+    const paymentSession = await this.paymentSessionService_.retrieve(
```
olivermrbl: comment: The result from the workflow execution cannot be directly serialized because it's missing some required properties. Perhaps, this is expected but thought I'd highlight, since this means we need to introduce some sort of "pattern" here. 

CleanShot 2024-02-29 at 20 28 31

packages/payment/src/services/payment-module.ts
```diff
@@ -207,38 +207,33 @@ export default class PaymentModuleService<
...
-      const sessionData = await this.paymentProviderService_.createSession(
-        data.provider_id,
-        {
-          ...data.providerContext,
-          resource_id: created.id,
-        }
-      )
+    if (errors.length) {
+      throw errors[0].error
+    }
```
olivermrbl: comment: Might make sense to come up with a pattern for handling errors from workflows within services

packages/payment/src/services/payment-module.ts
adrien2p: I dont see a way of doing that as the context is not shared at any point, it is an outside function and unless binding this and making the workflow aware of this property i can't see a nice and clean way of doing this. You would have to do something at somepoint as we already discussed that for the route as well 👍

packages/payment/src/services/payment-module.ts
adrien2p: The workflow return an already serialised entity, so it is not an entity that can be serialised anymore, i can improve the util to check that though

packages/payment/src/services/payment-module.ts
adrien2p: Maybe instead of having a throwOnError configuration, we could have an onError callback. If it is provided you receive all the errors and can process them as you want and if not provided then it will throw. Then we can have default onError handler? 🤔

packages/payment/src/services/payment-module.ts
olivermrbl: Yeah, I think you are right. Hard to avoid passing the container.

But it would be cool to see if we can find something more intuitive than `this.__container__`. 

I feel like properties that have this type of naming signals "don't touch me unless you are 100% aware of what you are doing" 😅

adrien2p: I can name it differently but it was more to avoid conflicts if a user want to store the container on this.container_ 🤣

packages/orchestration/src/workflow/local-workflow.ts
carlos-r-l-rodrigues: Have you tried to use `createContainerLike(modulesLoaded)`?
otherwise we can bring that loop to register all the keys using asValue to the createMedusaContainer helper if the "cradle" is not present in the first arg.

adrien2p: I believe if you use the createContainerLike util before calling the createContainer it would work, could you try?

adrien2p: I would avoid that loop as we might loose the life time of the object that are registered by doing it like that i think

packages/payment/src/workflows/steps/create-payment-provider-session.ts
```diff
@@ -0,0 +1,39 @@
...
+}
+
+export const createPaymentProviderSessionStepId =
+  "module:create-payment-provider-session"
+export const createPaymentProviderSessionStep = createStep(
+  createPaymentProviderSessionStepId,
+  async (input: StepInput, { container }) => {
+    const service = container.resolve("paymentProviderService")
+
+    const session = await service.createSession(input.provider_id, input.data)
```
carlos-r-l-rodrigues, line 20: here, as this is a workflow scoped to this module we could pass the sharedContext to call the service, and use the same transaction.
That could even allow us to remove the deleteSession on the compensate action, as that would be rollback by the same transaction.

packages/payment/src/services/payment-module.ts
carlos-r-l-rodrigues: At the moment I don't have an anwers how we could skip from passing the container here. maybe sharing the "this" at the sharedContext, so you could get the instance from the context in the workflow step, instead of getting the container

adrien2p: Yeah but then we mix responsibilities and the workflow start to know what the service is holding which seems weird

carlos-r-l-rodrigues: but this is literally for the internal workflow. we can't do that outside this scope.

adrien2p: Yes true, it is a local module worklow

packages/payment/src/workflows/steps/create-payment-session.ts
```diff
@@ -0,0 +1,39 @@
...
+}
+
+export const createPaymentSessionStepId = "module:create-payment-session"
+export const createPaymentSessionStep = createStep(
+  createPaymentSessionStepId,
+  async (data: StepInput, { container }) => {
+
+    const service = container.resolve("paymentSessionService")
+
+    const created = await service.create({
```
carlos-r-l-rodrigues: Here we can pass the sharedContext to the create use the same DB transaction

packages/payment/src/workflows/steps/create-payment-session.ts
```diff
@@ -0,0 +1,39 @@
...
+      provider_id: data.provider_id,
+      amount: data.amount,
+      currency_code: data.currency_code,
+      payment_collection: data.payment_collection_id,
+      data: data.data,
+    })
+
+    return new StepResponse(created, created.id)
+  },
+  async (createdSession, { container }) => {
```
carlos-r-l-rodrigues: if using the same db transaction we don't need to compensate it

packages/payment/src/workflows/steps/create-payment-provider-session.ts
carlos-r-l-rodrigues: maybe we can better handle this in case of error to decide if we are throwing a regular error that can be retried by the workflow or if it's a permanent error.

packages/payment/src/workflows/steps/create-payment-session.ts
adrien2p: I wouldn't do that, plus we still have a LRT here and it would just increase it, lets discuss tomorrow

carlos-r-l-rodrigues: yes, disregard it competely. that was the reason to use workflows to begin with..
🤦🏻‍♂️

adrien2p: Ahah no worries, even though the current LRT issue remain with the workflow in the current state of the pr. We can huddle tomorrow and ill expose my points 💪

packages/payment/src/services/payment-module.ts
carlos-r-l-rodrigues: we can also remove the `InjectTransactionManager` from this method so we don't have a long transaction running

adrien2p: Yes we have to, thats why i though the workflow wasn't really necessary here as we can do that in the function directly, lets discuss further tomorrow

packages/payment/src/workflows/steps/create-payment-session.ts
carlos-r-l-rodrigues: we just need to remove the InjectTransactionManager no?

adrien2p: Yea that should suffice, and replace it with Inject manager instead. The call to the inner service will start a transaction block for it, and if we have multiple actions to do then we can use this.baseRepository_.transaction to wrap multiple calls in a single transaction which can also take a transaction if one already started but it shouldn't be the case here

packages/orchestration/src/workflow/local-workflow.ts
olivermrbl: You mean like so:
```
const containerLike = createContainerLike(modulesLoaded)
container = createMedusaContainer(containerLike)
```

packages/payment/src/services/payment-module.ts
```diff
@@ -201,44 +201,67 @@ export default class PaymentModuleService<
...
 
     try {
-      const sessionData = await this.paymentProviderService_.createSession(
-        data.provider_id,
-        {
-          ...data.providerContext,
-          resource_id: created.id,
-        }
-      )
+      const providerSessionSession =
```
olivermrbl, line 213: @adrien2p, wdyt?

adrien2p: sounds good to me 💪

carlos-r-l-rodrigues: the drawback is that we lose the built-in ability of the workflow to retry calling the provider, and in the future when we handle errors to compensate, we'll miss it here

olivermrbl: If you had to choose, what approach would you go with? I think we should settle on one and use it everywhere.

adrien2p: The only problem with retry mechanism handled in that place is that it could timeout the requests, probably the retry should happen from the client in that context. Like we do with 409

carlos-r-l-rodrigues: I don't have a strong opinion, I just feel that by using the workflow we could benefit from future improvements and handling things using config, like the retries.
But that can be applied in the future as well, so I'm good with that too.
If we had more than 1 external call in the same operation then I would say the workflow would be a better choice.

adrien2p: yes with multiple external `writes` call I do agree especially if they are separated with internal actions. But for the retries in the service method through the workflow I am not entirely sure it would be wise to have that, for me it should be handled differently, in async workflows its fine, but during the request lifespan it can create other issues. We can see that later in the futre

carlos-r-l-rodrigues: yeah, we can discuss that later but this is an external call, it is common to have built in retries when doing network calls as issues are somewhat common to happen.

adrien2p: Yeah I do agree, my question is where this should happen, because if the request end up timing out while it is still retrying then you are loosing the context no? we can go with server retries I don't have a strong opinion I just wanted to raise this though ahah :)

olivermrbl: Alright, I will keep it as is, and we can revisit in the future.

packages/medusa/src/api-v2/store/payment-collections/[id]/payment-sessions/middlewares.ts
```diff
@@ -0,0 +1,11 @@
+import { transformBody } from "../../../../../api/middlewares"
+import { MiddlewareRoute } from "../../../../../loaders/helpers/routing/types"
+import { StorePostPaymentCollectionsPaymentSessionReq } from "./validators"
+
+export const storeCartRoutesMiddlewares: MiddlewareRoute[] = [
+  {
+    method: ["POST"],
+    matcher: "/store/payment-collections/:id/payment-sessions",
```
Comment (olivermrbl) [inline packages/medusa/src/api-v2/store/payment-collections/[id]/payment-sessions/middlewares.ts:8]: comment: An alternative path could be `/store/payment-sessions`
`{ payment_collection_id: "paycol_1234", ... }`

packages/types/src/payment/provider.ts
```diff
@@ -43,31 +43,31 @@ export type PaymentProviderContext = {
...
+  context: PaymentProviderContext
+  amount: number
+  currency_code: string
+}
+
+export type UpdatePaymentProviderSession = {
+  context: PaymentProviderContext
+  data: Record<string, unknown>
+  amount: number
+  currency_code: string
```
olivermrbl, line 70: comment:

I replaced the `context` of the payment provider context DTO to not have `context.context`.

Instead, we now have `extra` which can be used by developers to pass additional non-required details to the payment session creation, e.g. a description of the payment.

Additionally, I extracted the `payment_session_data` of the context to be a standalone property used when creating/updating a payment session.

packages/orchestration/src/workflow/local-workflow.ts
```diff
@@ -53,7 +54,8 @@ export class LocalWorkflow {
 
     if (!Array.isArray(modulesLoaded) && modulesLoaded) {
       if (!("cradle" in modulesLoaded)) {
-        container = createMedusaContainer(modulesLoaded)
+        const containerLike = createContainerLike(modulesLoaded)
+        container = createMedusaContainer(containerLike)
```
olivermrbl, line 58: @adrien2p, was this what you had in mind?

srindom:
Think this is great - also like the endpoint for initializing sessions 👍

packages/medusa/src/api-v2/store/payment-collections/[id]/payment-sessions/middlewares.ts
Comment (sradevski) [reply packages/medusa/src/api-v2/store/payment-collections/[id]/payment-sessions/middlewares.ts:8]: If we always know the collection ID (and we have to know it), then the existing endpoint makes sense. If you can operate on payment sessions without knowing the payment collection id (eg. you know the session ID and you want to delete it, but you don't know the collection ID), then what you proposed makes more sense.

I actually have a similar case in the product endpoints that I wanted to discuss, but we can do it once I open a PR

packages/medusa/src/api-v2/store/payment-collections/[id]/payment-sessions/route.ts
```diff
@@ -0,0 +1,64 @@
...
+
+export const POST = async (
+  req: AuthenticatedMedusaRequest<StorePostPaymentCollectionsPaymentSessionReq>,
+  res: MedusaResponse
+) => {
+  const { id } = req.params
+  const { context, provider_id, data } = req.body
+
+  // If the customer is logged in, we auto-assign them to the payment collection
+  if (req.auth?.actor_id) {
```
Comment (sradevski) [inline packages/medusa/src/api-v2/store/payment-collections/[id]/payment-sessions/route.ts:18]: Should this happen in some middleware instead?

packages/medusa/src/api-v2/store/payment-collections/[id]/payment-sessions/route.ts
```diff
@@ -0,0 +1,64 @@
...
+      ...context.customer,
+      id: req.auth.actor_id,
+    }
+  }
+
+  const paymentService = req.scope.resolve<IPaymentModuleService>(
+    ModuleRegistrationName.PAYMENT
+  )
+
+  const paymentCollection = await paymentService.retrievePaymentCollection(id, {
```
Comment (sradevski) [inline packages/medusa/src/api-v2/store/payment-collections/[id]/payment-sessions/route.ts]: Should we use remoteQuery instead? Alternatively, should the data fetching happen in the workflow itself?

packages/medusa/src/api-v2/store/payment-collections/[id]/payment-sessions/route.ts
```diff
@@ -0,0 +1,64 @@
...
+    select: ["id", "amount", "currency_code"],
+  })
+
+  const workflowInput = {
+    payment_collection_id: id,
+    provider_id: provider_id,
+    data,
+    context,
+    amount: paymentCollection.amount, // Question for the future: How do we handle split payments?
```
Comment (sradevski) [inline packages/medusa/src/api-v2/store/payment-collections/[id]/payment-sessions/route.ts]: One way to handle split payments is to not really care whether a payment is split or not. Let me explain:
1. You create a draft order / order awaiting payment
2. You do a payment of whatever amount you want, as many times as you want. (we check that you don't overpay)
3. The order is marked as paid only when the order amount = the amounts you've paid for this particular order.

Like this, you remove all complexity of managing split payments as part of the payment module, and any orchestration you need to do there.

There are different approaches and I think this will require a written document with proposals, but this is roughly how I've worked with split payments before

packages/payment/src/services/payment-module.ts
```diff
@@ -201,44 +201,67 @@ export default class PaymentModuleService<
...
+        input,
         sharedContext
       )
-
-      return await this.baseRepository_.serialize(created, { populate: true })
-    } catch (e) {
-      await this.paymentSessionService_.delete([created.id], sharedContext)
-      throw e
+    } catch (error) {
+      // In case the session is created at the provider, but fails to be created in Medusa,
```
sradevski, line 231: Note that this is not enough if we want to always clean up, since what can happen is:
1. Create session in provider
2. Your node crashes at any point between the session being created, and `deleteSession` being executed

So you'll end up with an orphaned session. I guess in this case it doesn't matter as sessions expire I assume.

With that in mind, do we even need this cleanup step? Failures will be rare on our side anyway, so not sure if we need the added complexity and false security that it will fully clean up.

sradevski (commented):
@olivermrbl can I find a high level overview of how the interactions between collection, session, cart, payment etc. look like?

One thing that I find strange with sessions is that we are exposing implementation detail to the user. Users don't really care about sessions, that's just how it happens some of the payment providers to be implemented. So in that sense, I feel like we are leaking abstractions to the users and it's better if we can avoid it.

Now I can't suggest how it should look exactly without going deeper and understanding the entire flow, but I still wanted to put it out here as food for thought. WDYT?

olivermrbl:
> can I find a high level overview of how the interactions between collection, session, cart, payment etc. look like?

You can find the Payment module design document [here](https://www.notion.so/medusajs/Payment-38608f6bf39b4fb8a5cc06acfef182c3?pvs=4) and a very messy Figma board [here](https://www.figma.com/file/AGdYJXTy40oJjeJ9kImn41/PaymentModule-Data-Models-and-Flows?type=whiteboard&node-id=1-2&t=BLdPH8iBbRTO32qn-4). The latter is maybe not the most useful, but thought I'd link it anyway. 

I am not sure I understand your comment on the "leakiness" of the payment session concept, so I would suggest you read through this documentation first and then you can let me know if you think we should still consider a redesign.

Does that sound OK?

packages/medusa/src/api-v2/store/payment-collections/[id]/payment-sessions/route.ts
Comment (olivermrbl) [reply packages/medusa/src/api-v2/store/payment-collections/[id]/payment-sessions/route.ts]: I decided to use the service directly, since we don't need any resources outside of the payment module's domain. We could use the `remoteQuery` for consistency.

That said, I think I'll move the retrieval to the workflow for now, since that would eliminate this concern.

packages/payment/src/services/payment-module.ts
olivermrbl: Orphaned sessions do not really matter, no. However, I don't find this to add much complexity to the operation, so I'll keep this in for now. I am open to revisiting it in the future.

packages/medusa/src/api-v2/store/payment-collections/[id]/payment-sessions/route.ts
Comment (olivermrbl) [reply packages/medusa/src/api-v2/store/payment-collections/[id]/payment-sessions/route.ts:18]: I could introduce a generic middleware `maybeAttachCustomerIdFromActor` for reusability purposes. Wdyt?

Comment (sradevski) [reply packages/medusa/src/api-v2/store/payment-collections/[id]/payment-sessions/route.ts:18]: @olivermrbl yeah I think that would make a lot more sense, eg. `attachActorToCustomer`. I am a bit uncertain about the exact logic we want here, but would it even make sense if it's part of the auth middleware?

packages/medusa/src/api-v2/store/payment-collections/[id]/payment-sessions/route.ts
Comment (sradevski) [reply packages/medusa/src/api-v2/store/payment-collections/[id]/payment-sessions/route.ts]: Sounds good 👍

sradevski:
@olivermrbl went through the document, and while waiting for access to the FIgma file this is where I find a bit of leaking concepts:

This is my understanding of the flow in Notion, and how it can be tweaked:
1. Customer goes to the checkout page - we call `createPaymentCollection`
2. Customer selects payment method - we call `createPaymentSession`. This is where the leakage is - not every provider uses sessions to manage payments. What the user expresses here is `selectPaymentProvider`. Internally we might still create a session, and store its ID, but the end user shouldn't care about it.
3. Customer completes payment and creates order - Throughout the flow I think using the `paymentCollection` ID, rather than the `paymentSession` ID is better, since the `paymentCollection` is what the customer intended to do.

Let me know what you think, I might be missing some information where exposing the session makes more sense.

olivermrbl:
> 2\. Customer selects payment method - we call `createPaymentSession`. This is where the leakage is - not every provider uses sessions to manage payments. What the user expresses here is `selectPaymentProvider`. Internally we might still create a session, and store its ID, but the end user shouldn't care about it.

Thanks for sharing. I get what you are alluding to, and I believe we might be able to find a good solution by introducing minor tweaks to the payment module interface.

I agree the `createPaymentSession` method is slightly misleading when used in the context of a traditional checkout flow where a storefront customer selects a single method of payment, and that's it. However, since this is at the module level, I would like to keep the interface generic enough to make sense for any use case where you would like to collect one or more payment(s) for _something_ through payment sessions. What do you think about renaming it to `selectPaymentSession` or `initiatePaymentSession` (I would probably still just create wrappers around `createPaymentSession`)?

I am not convinced we should "decouple" the module interface from the session concept – if we want to do this, it should probably be at a higher level in the stack e.g. in the Rest API. As a developer consuming the module directly, I think it would cause more harm than good to expose a method `selectPaymentProvider`, and then what happens under the hood is actually payment session creation. This is where I think `selectPaymentSession` would be good to have.

Not every provider uses sessions to manage payments, but we use sessions to manage payments for all providers. I don't see how we can get around a one-fits-all concept here. We need an abstraction to describe a process that might differ in naming from provider to provider, and we happen to use payment sessions for it. 

So to summarise, I don't see the need for any immediate changes to the PR, except for maybe a new endpoint and an aliased interface method to make it more intuitive to consume in a storefront.

Let me know what you think.

sradevski:
@olivermrbl ok I should have placed my comment at the right place 😄 I completely agree that the module can expose the concept of a session, but in this PR we created a REST endpoint called `payment_sessions`, and that is where I think we're leaking info. I thought that could be something more high-level, like selecting a payment provider, but the workflow would take care of creating the session, and everything in-between.

olivermrbl:
Ahh I see. But then I believe we are aligned

> but in this PR we created a REST endpoint called `payment_sessions`, and that is where I think we're leaking info.

We can introduce a new endpoint `POST /store/payment-collections/:id/[select]` or similar, that would be more intuitive and not leak the session concept through.

olivermrbl:
@sradevski, after giving this some thought, I actually want to keep the API as it is. At least for the time being.

To me, it makes sense that you initiate a payment session similar to how you would initiate a checkout session, when you want to collect a payment for something. I believe the concept of a session in the checkout domain is well-known, and most developers should be familiar with it. 

One thing we could still consider is introducing an endpoint like so: `POST /payment-collections/:id/sessions/:provider_id`. I believe this would make it more obvious that you are selecting a provider for your sessions.

sradevski (approved):
@olivermrbl sounds good, we can definitely keep it as-is, it really isn't a major problem, I just wanted to share another perspective of looking at the API. 🍾

packages/medusa/src/api-v2/store/payment-collections/[id]/payment-sessions/route.ts
Comment (olivermrbl) [reply packages/medusa/src/api-v2/store/payment-collections/[id]/payment-sessions/route.ts:18]: Went with a slightly simpler approach in f0c9aa0f100cbc3153de14e2c. I checked other routes that use the actor ID and there are no commonalities yet, so a middleware seems overkill. Some use `customer.id`, others `created_by`, and here `customer_id`.

Comment (olivermrbl) [reply packages/medusa/src/api-v2/store/payment-collections/[id]/payment-sessions/route.ts:18]: Went with a slightly simpler approach in f0c9aa0f100cbc3153de14e2c for cart creation and kept this as is. I checked other routes that use the actor ID and there are no commonalities yet, so a middleware seems overkill. Some use `customer_id`, others `created_by`, and here `customer.id`.

PR #6550, opened 2024-02-29
feat(dashboard): Reservations and inventory item list pages

kasperkristensen:
**What**
- Adds list page for reservations and inventory items
- Adds new String filter type, that allows users to filter by a string, eg. "material === 'metal'"
- Adds new Number filter type, that allows users to filter by a number or numerical comparator, eg. quantity === 10 / quantity is gt 10 and lt 50.

### DIFF
--- a/.changeset/plenty-bikes-drop.md
+++ b/.changeset/plenty-bikes-drop.md
@@ -1,0 +1,6 @@
+---
+"@medusajs/ui": patch
+"@medusajs/medusa": patch
+---
+
+fix(medusa,ui) Export param types for workflow endpoints. Add support for JSON to CodeBlock component.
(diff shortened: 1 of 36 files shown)


olivermrbl:
Clicking the # order link takes you to the reservation item (which is currently not done) instead of the order: 


https://github.com/medusajs/medusa/assets/59018053/d9ed63a6-0b2a-4361-be6c-32a0dbee4a72

olivermrbl:
Should we show location in the list view too?

olivermrbl (approved):

kasperkristensen:
@olivermrbl unfortunately we can't as the reservations don't have a relation to their location, just a location_id. This means we would have to fire of a request for each table row (20 per page) in order to display the location. You can filter by location which is the best we can do with the current data structure.

PR #6557, opened 2024-03-01
feat(tax): add endpoints to manage tax rate rules

srindom:
**What**
Adds endpoints to manage tax rules on a tax rate:
- Create a tax rule: POST /admin/tax-rates/:id/rules 
- Delete a tax rule: DELETE /admin/tax-rates/:id/rules/:rule_id
- Replace tax rules: POST /admin/tax-rates/:id -- with { rules: [...] } in body.

### Noteworthy things I bumped into

**Updating nested relationships**
A TaxRate can have multiple TaxRules and in this PR we enable users to replace all TaxRules associated with a TaxRate in one operation. If working with the module directly this can be done with:

```javascript
taxModuleService.update(rateId, { rules: [{ ... }] })
```

Internally in the `update` function the TaxModule first soft deletes any TaxRules that exist on the TaxRate and then creates new TaxRules for the passed rules ([see test](https://github.com/medusajs/medusa/pull/6557/files#diff-cdcbab80ac7928b80648088ec57a3ab09dddd4409d6afce034f2caff08ee022bR78)).

A challenge arises when doing this in a compensatable way in a workflow. To see this imagine the following:
1. `updateTaxRatesWorkflow` gets the current data for the tax rates to update. This includes the tax rates' rules.
2. `updateTaxRatesWorkflow` calls `taxModuleService.update` with new rules. 
3. Internally, the tax module deletes the rules in 1. and creates new rules.
4. Imagine an error happens in a following step and the workflow has to compensate.
5. The workflow uses the data from 1. and calls upsert. The tax module may correctly update the previous tax rules so they are no longer soft deleted. However, upsert (at least not by default) doesn't delete the new rules that were created in 2.

As illustrated by 5. compensating the update is not pretty. To get around this I instead opted to let the workflow handle setting the rules for a rate that makes the compensation more straightforward to handle. [See workflow here](https://github.com/medusajs/medusa/pull/6557/files#diff-ff19e1f2fa32289aefff90d33c05c154f9605a3c5da6a62683071a1fcaedfd7bR89).

**Using nested workflows**
Initially, I wanted to use the `setTaxRateRulesWorkflow` within the `updateTaxRatesWorkflow`. And this worked great for the invoke phase. However, when I needed to compensate the update workflow (and hence also had to compensate the set rules workflow), I found that the workflow engine no longer had the set rules transaction in memory and therefore could not roll it back. ([This is where I try to rollback](https://github.com/medusajs/medusa/pull/6557/files#diff-ff19e1f2fa32289aefff90d33c05c154f9605a3c5da6a62683071a1fcaedfd7bR62), but the transaction id can't be found).

I therefore opted to copy the steps from the set tax rate rules workflow into the update tax rates workflow; however, once we figure out a good way to ensure we can compensate nested workflows we should move to the nested workflow instead. 

This also made me realize that the current implementation of workflows that use `refreshCartPromotions` may create inconsistencies in case of failures (cc: @riqwan).


packages/medusa/src/api-v2/admin/tax-rates/[id]/rules/batch/set/route.ts
```diff
@@ -0,0 +1,40 @@
...
+  res: MedusaResponse
+) => {
+  const taxRateId = req.params.id
+  const rateRules = req.validatedBody.rules.map((r) => ({
+    ...r,
+    created_by: req.auth.actor_id,
+  }))
+
+  const { errors } = await setTaxRateRulesWorkflow(req.scope).run({
+    input: { taxRateId, rules: rateRules },
```
Comment (riqwan) [inline packages/medusa/src/api-v2/admin/tax-rates/[id]/rules/batch/set/route.ts]: we decided on recently to snake case all workflow inputs

riqwan (approved):
lgtm!

packages/core-flows/src/tax/workflows/set-tax-rate-rules.ts
```diff
@@ -0,0 +1,49 @@
...
+} from "@medusajs/workflows-sdk"
+import { createTaxRateRulesStep, deleteTaxRateRulesStep } from "../steps"
+import { ModuleRegistrationName } from "@medusajs/modules-sdk"
+
+type WorkflowInput = {
+  taxRateId: string
+  rules: Omit<CreateTaxRateRuleDTO, "tax_rate_id">[]
+}
+
+const listRulesStep = createStep(
```
adrien2p: suggestion: we usually split the workflow and the steps in different files as well as export the step in order to be reusable for a user

adrien2p: should we name it `listRuleIdsStep`?

adrien2p (approved):
Very good, few comment and snake case to handle

packages/medusa/src/api-v2/admin/tax-rates/middlewares.ts
```diff
@@ -36,4 +38,14 @@ export const adminTaxRateRoutesMiddlewares: MiddlewareRoute[] = [
       ),
     ],
   },
+  {
+    method: "POST",
+    matcher: "/admin/tax-rates/:id/rules/batch/set",
```
olivermrbl: question: Do we need the `/set`?

packages/core-flows/src/tax/workflows/set-tax-rate-rules.ts
```diff
@@ -0,0 +1,49 @@
...
+  WorkflowData,
+  createStep,
+  createWorkflow,
+  transform,
+} from "@medusajs/workflows-sdk"
+import { createTaxRateRulesStep, deleteTaxRateRulesStep } from "../steps"
+import { ModuleRegistrationName } from "@medusajs/modules-sdk"
+
+type WorkflowInput = {
+  taxRateId: string
```
olivermrbl:
```suggestion
  tax_rate_id: string
```

packages/core-flows/src/tax/workflows/set-tax-rate-rules.ts
```diff
@@ -0,0 +1,49 @@
...
+  rules: Omit<CreateTaxRateRuleDTO, "tax_rate_id">[]
+}
+
+const listRulesStep = createStep(
+  "set-tax-rate-rules-list-rules",
```
olivermrbl:
```suggestion
  "list-tax-rate-rules",
```

olivermrbl (approved):
LGTM, just a few comments

packages/medusa/src/api-v2/admin/tax-rates/middlewares.ts
sradevski: Also as a general rule, do we want to have `batch` operations on a separate route from the non-batch ones, or should it be the same, but the data format will differ?
Eg. you can just have it as `:id/rules` and the data can be `data: {}` or `data: [{}]`?

sradevski (approved):

packages/medusa/src/api-v2/admin/tax-rates/middlewares.ts
srindom: > question: Do we need the /set?

The idea here was that we could have different batch operations e.g.: `/set`, `/add`, `/remove` - but maybe there's a better approach.

sradevski: I think the HTTP methods are sufficient to distinguish between those actions, but we can always revisit if that's no longer true

srindom: I don't think this works well a few examples:

- POST /admin/tax-rates/:id/rules -> does this create new rules or override existing rules
- DELETE /admin/tax-rates/:id/rules -> DELETE typically doesn't take a body; we have done it in previous endpoints but it's a non-standard pattern.

--
For this particular case I think the right approach would probably be to support `rules` on the update resource. E.g.,:

- POST /admin/tax-rates/:id 
  { rules: [{...}, ...], }

Will change to that approach here.

PR #6559, opened 2024-03-01
fix: Move country loading for region to the loader, fix a bug with cascade

sradevski:


integration-tests/environment-helpers/use-db.js
```diff
@@ -70,11 +70,8 @@ const DbTestUtil = {
         continue
       }
 
-      await manager.query(`DELETE
-                           FROM "${table_name}";`)
+      await manager.query(`DELETE FROM "${table_name}";`)
     }
-
-    await manager.query(`SET session_replication_role = 'origin';`)
```
sradevski: @olivermrbl (I see you were the last one to change this line) do you remember why we were manually changing the replication role? It was messing up the cascade delete, and it seems to be just fine without doing it.

packages/currency/src/loaders/initial-data.ts
```diff
@@ -10,20 +10,23 @@ export default async ({
   | ModulesSdkTypes.ModuleServiceInitializeOptions
   | ModulesSdkTypes.ModuleServiceInitializeCustomDataLayerOptions
 >): Promise<void> => {
-  try {
-    const {
-      currencyService_,
-    }: { currencyService_: ModulesSdkTypes.InternalModuleService<Currency> } =
-      container.resolve(ModuleRegistrationName.CURRENCY)
+  // TODO: Add default logger to the container when running tests
```
sradevski, line 13: I'll do this in a separate PR

integration-tests/environment-helpers/use-db.js
olivermrbl: Hmm, I can't recall. If everything is working fine, I am fine to remove it.

olivermrbl (approved):

olivermrbl:
Just need tests to pass and we should be good to merge 🏁

integration-tests/environment-helpers/use-db.js
sradevski: @olivermrbl ok I see what we're doing here, we are not cascading operations correctly, so this is a "hack" to get around the constraint checks. I'll revert it for now, but this reduces the confidence in our tests doing the right thing :D

adrien2p: Probably other module constraints are not set up correctly, it will be part of the work we need to do to ensure everything is right. I believe most cascade are using application level instead of db level (both in some cases, db only in best case) and probably some constraints might be missing.

sradevski: Yup that's why I'll just do a manual delete on my data for now, and added todo to revisit how the teardown is done.

PR #6561, opened 2024-03-01
feat(core-flows,link-modules,modules-sdk): add cart <> promotion link as source of truth

riqwan:
what:

- adds promotion cart link
- update steps to create and remove links


packages/core-flows/src/definition/cart/steps/update-cart-promotions.ts
```diff
@@ -0,0 +1,107 @@
...
+    if (action === PromotionActions.REPLACE) {
+      for (const link of existingCartPromotionLinks) {
+        linksToDismiss.push({
+          [Modules.CART]: { cart_id: link.cart_id },
+          [Modules.PROMOTION]: { promotion_id: link.promotion_id },
+        })
+      }
+    }
+
+    if (linksToDismiss.length) {
```
sradevski: Let's parallelize these calls, no reason not to IMO

packages/link-modules/src/definitions/cart-promotion.ts
```diff
@@ -0,0 +1,55 @@
+import { Modules } from "@medusajs/modules-sdk"
+import { ModuleJoinerConfig } from "@medusajs/types"
+import { LINKS } from "../links"
+
+export const CartPromotion: ModuleJoinerConfig = {
+  serviceName: LINKS.CartPromotion,
+  isLink: true,
+  databaseConfig: {
+    tableName: "cart_promotion",
+    idPrefix: "capromo",
```
sradevski: We might as well just make it `cartpromo` 😄

sradevski (approved):

packages/link-modules/src/definitions/cart-promotion.ts
riqwan: I was joining the prefix convention of both modules. 😄 why not, lets do it.
```suggestion
    idPrefix: "cartpromo",
```

packages/core-flows/src/definition/cart/steps/update-cart-promotions.ts
srindom: You can also pass a list of link defs here instead of calling them one by one.

riqwan: @srindom Can you check the latest commit? I think you might be on an older version

srindom: You are right sorry about that!

carlos-r-l-rodrigues (approved):

PR #6562, opened 2024-03-01
feat: Add supported currencies to store model

sradevski:
I ultimately went with having a flat list of settings for the store. It wouldn't be very difficult to change it if we wish to do so, but for now this keeps the codebase simpler

### DIFF
--- a/packages/medusa/src/api-v2/admin/stores/validators.ts
+++ b/packages/medusa/src/api-v2/admin/stores/validators.ts
@@ -1,5 +1,11 @@
 import { Type } from "class-transformer"
-import { IsObject, IsOptional, IsString, ValidateNested } from "class-validator"
+import {
+  IsArray,
+  IsObject,
+  IsOptional,
+  IsString,
+  ValidateNested,
+} from "class-validator"
 import { FindParams, extendedFindParamsMixin } from "../../../types/common"
 
 export class AdminGetStoresStoreParams extends FindParams {}
@@ -40,6 +46,10 @@ export class AdminPostStoresStoreReq {
   @IsOptional()
   @IsString()
   name?: string
+
+  @IsOptional()
+  @IsArray()
+  supported_currency_codes?: string[]
 
   @IsOptional()
   @IsString()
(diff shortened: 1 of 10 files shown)


packages/store/src/models/store.ts
```diff
@@ -35,6 +35,9 @@ export default class Store {
   @Property({ columnType: "text", default: "Medusa Store" })
   name: string
 
+  @Property({ type: "array", default: "{}" })
```
adrien2p, line 38: Suggestion: i would default to [] as it is a valid jsonb value

adrien2p (approved):

packages/store/src/models/store.ts
sradevski: @adrien2p why do we want it as jsonb, where Postgres has native support for primitive type arrays (and IIRC "{}" is how you represent an empty array, but maybe mikroorm does some magic if you pass []?)

adrien2p: My bad, mostly a habit, also jsonb areay are declare like this [] and with array you can either use {} or ARRAY[]. Feel free to discard

sradevski: @adrien2p No worries, I am always wondering if there is some magic happening that I should know of 😄

packages/medusa/src/api-v2/admin/stores/validators.ts
```diff
@@ -41,6 +47,10 @@ export class AdminPostStoresStoreReq {
   @IsString()
   name?: string
 
+  @IsOptional()
+  @IsArray()
+  supported_currency_codes?: string[]
```
riqwan, line 52: Just a nit, but I think we can drop the `supported_` prefix, discardable.

riqwan (approved):
lgtm!

packages/medusa/src/api-v2/admin/stores/validators.ts
sradevski: @riqwan I was a bit torn TBH, and I decided to add it to distinguish between "all currency codes" and the ones that the store decided to support.

PR #6564, opened 2024-03-01
feat(dashboard): Workflow executions

kasperkristensen:
**What**
- v1 of executions domain

**TODO in follow up PR**
I think it might be a good idea for Carlos or Adrien to go over the design with Ludvig, and make sure we are displaying what is relevant to developers, and how things should be displayed. So this PR is just the initial implementation, and does not handle displaying things like compensation steps (not part of the current design), step input (not supported by the API but part of the design), etc.

Closes CORE-1751, CORE-1755


srindom:
A global status would be helpful:

CleanShot 2024-03-05 at 11  42 53

---

nit: the caret should turn 180 degrees when folded out

CleanShot 2024-03-05 at 11  43 29

--- 
The code blocks feel too muted - would like them to have a dark background and syntax highlighting

CleanShot 2024-03-05 at 11  44 07

---
In output we might want to filter out stuff like this:

CleanShot 2024-03-05 at 11  45 32

kasperkristensen:
I think everything should be addressed now, still things to be done with this, but I'll revisit it again when we get SSE for real time updates, and hopefully export/improve the types for the executions, as it's a bit tricky for me to entirely understand what is returned and in what cases things are returned and not.

olivermrbl:
> I think everything should be addressed now, still things to be done with this, but I'll revisit it again when we get SSE for real time updates, and hopefully export/improve the types for the executions, as it's a bit tricky for me to entirely understand what is returned and in what cases things are returned and not.

Sounds good, let's get this in 💪

olivermrbl (approved):

PR #6565, opened 2024-03-01
chore: Make module tests run concurrently

sradevski:
Locally this makes the test run go from ~300s to ~50s. If we see this working well, we can apply it to all other test run configurations.

We will just need to be disciplined and make sure all test suites are independent and run in their own DB, for example

### DIFF
--- a/integration-tests/modules/package.json
+++ b/integration-tests/modules/package.json
@@ -5,7 +5,7 @@
   "license": "MIT",
   "private": true,
   "scripts": {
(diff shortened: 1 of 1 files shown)


sradevski:
Also upgrading to Jest 28/29 can bring us some performance improvements, but there are breaking changes we need to deal with

carlos-r-l-rodrigues:
now that we removed probably all the missing `await` in the code, we might be lucky and see that working. 🤞🏻 🍀

sradevski:
Of course it had to fail in the runners, locally it seems fine 😄 I'll try to get it working

PR #6577, opened 2024-03-04
feat: Region PaymentProvider link

olivermrbl:
**What**

- Introduce link between Region and PaymentProvider
- Introduce API endpoint `GET /store/regions/:id/payment-providers` for retrieving providers by region
- Add tests for both

### DIFF
--- a/www/packages/docs-ui/src/hooks/index.tsx
+++ b/www/packages/docs-ui/src/hooks/index.tsx
@@ -1,4 +1,5 @@
 export * from "./use-copy"
+export * from "./use-is-browser"
 export * from "./use-keyboard-shortcut"
 export * from "./use-scroll-utils"
 export * from "./use-search-navigation"
--- a/www/packages/docs-ui/src/hooks/use-is-browser/index.ts
+++ b/www/packages/docs-ui/src/hooks/use-is-browser/index.ts
@@ -1,0 +1,13 @@
+"use client"
+
+import { useEffect, useState } from "react"
+
+export const useIsBrowser = () => {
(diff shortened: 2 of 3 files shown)


packages/types/src/payment/service.ts
```diff
@@ -387,13 +389,11 @@ export interface IPaymentModuleService extends IModuleService {
    */
   cancelPayment(paymentId: string, sharedContext?: Context): Promise<PaymentDTO>
 
-  /**
-   * This method creates providers on load.
-   *
-   * @example
-   * {example-code}
-   */
-  createProvidersOnLoad(): Promise<void>
```
olivermrbl, line 403: comment: Removed this from the module service interface, as this should not be accessible to developers consuming the module. Instead, this now lives directly in the loader.

packages/payment/src/loaders/defaults.ts
```diff
@@ -1,10 +1,32 @@
...
 export default async ({ container }: LoaderOptions): Promise<void> => {
   const paymentModuleService: IPaymentModuleService = container.resolve(
     ModuleRegistrationName.PAYMENT
   )
 
-  await paymentModuleService.createProvidersOnLoad()
+  const providersToLoad = container.resolve("payment_providers")
+  const paymentProviderService = container.resolve("paymentProviderService")
+
+  const providers = await paymentModuleService.listPaymentProviders({
```
fPolic: nit: we can list these with `paymentProviderService` so we don't need to resolve `paymentModuleService` in the loader

fPolic (approved):
LGTM!

PR #6579, opened 2024-03-04
feat: add product admin v2 endpoints

sradevski:
This implementation obviously lacks a lot of things, and there are a lot of TODOs. However, there are already a lot of questions I'd rather get answered soon, so I figured it's much easier to do the implementation in steps.

I wrote down all breaking changes, suggested changes, and new additions with comments (TODO and Note).

In a follow-up PR I will:

Add the remaining/missing models
Make the workflows handle all interactions between the different models/modules
Add integration tests


packages/medusa/src/api-v2/admin/products/middlewares.ts
```diff
@@ -0,0 +1,146 @@
...
+    matcher: "/admin/products/:id",
+    middlewares: [transformBody(AdminPostProductsProductReq)],
+  },
+  {
+    method: ["DELETE"],
+    matcher: "/admin/products/:id",
+    middlewares: [],
+  },
+
+  // TODO: Part of v1 API. Remove in favor of just updating the metadata through as standard update
```
olivermrbl: +1

packages/medusa/src/api-v2/admin/products/[id]/variants/[variant_id]/route.ts
```diff
@@ -0,0 +1,84 @@
...
+  }
+
+  res.status(200).json({ product_variant: result[0] })
+}
+
+export const DELETE = async (
+  req: AuthenticatedMedusaRequest,
+  res: MedusaResponse
+) => {
+  // TODO: Should we allow fetching a variant without knowing the product ID? In such case we'll need to change the route to /admin/products/variants/:id
```
Comment (olivermrbl) [inline packages/medusa/src/api-v2/admin/products/[id]/variants/[variant_id]/route.ts:65]: With this specific endpoint path structure, I think it is fine to require the product ID. 

An alternative approach, if we want to support this, would be to introduce: `/admin/product-variants/:id`. That would make for pretty and consistent endpoonts.

Comment (olivermrbl) [reply packages/medusa/src/api-v2/admin/products/[id]/variants/[variant_id]/route.ts:65]: Same goes for all other comments on this.

packages/medusa/src/api-v2/admin/products/middlewares.ts
```diff
@@ -0,0 +1,146 @@
...
+      transformQuery(
+        AdminGetProductsOptionsParams,
+        QueryConfig.retrieveTransformQueryConfig
+      ),
+    ],
+  },
+  // Note: New endpoint in v2
+  {
+    method: ["GET"],
+    matcher: "/admin/products/:id/options/:option_id",
```
olivermrbl, line 116: question: Do we need this?

olivermrbl (approved):
Nice work. Think we should get this merged, so we can get started adding the remaining elements in follow up PRs.

A couple of comments:

We could move the `WorkflowInput` types to the types package, as we've done elsewhere. 

Also, listing below some additional things, that we will need to address in the product workflows (that you might already be aware of): 
- Shipping profile assignment (from fulfillment module)
- Sales channel assignment (from sales channel module)
- Product variant prices (from pricing module)
- Product variant inventory (from inventory + stock location modules)

packages/medusa/src/api-v2/admin/products/middlewares.ts
sradevski: @olivermrbl I added it for consistency, as it adds very little overhead. I know the options model is pretty slim but I think it helps if we have consistency where possible

packages/medusa/src/api-v2/admin/products/[id]/variants/[variant_id]/route.ts
Comment (sradevski) [reply packages/medusa/src/api-v2/admin/products/[id]/variants/[variant_id]/route.ts:65]: @olivermrbl a use-case I was thinking of is a case where you know the variant (eg. in the cart) but not the product, and you want to expand the relationship from variant -> product. Now if we have an endpoint like `/admin/product-variants/:id` (which I think works well), then do we even need the `product/:id/variant/:variant_id`? Anyway I am fine to leave things as they are for now, I am writing down all TODOs anyway and we can discuss this later on, just wanted to bring it up as a probably common scenario for nested models

sradevski:
@olivermrbl just FYI, the steps I am taking are:
1. Create the basic models first (no relationships yet). Add e2e tests, make it all work
2. Add extra models from the product module (types, collections, categories). Add tests for them
3. Add relationships and more complex workflows. Add tests

I'll add the remaining pieces of 1. in a follow-up PR, but the relationships as you mentioned will come a bit later

packages/medusa/src/api-v2/admin/products/[id]/variants/[variant_id]/route.ts
Comment (olivermrbl) [reply packages/medusa/src/api-v2/admin/products/[id]/variants/[variant_id]/route.ts:65]: Makes sense.

> Now if we have an endpoint like /admin/product-variants/:id (which I think works well), then do we even need the product/:id/variant/:variant_id

If we introduce `/admin/product-variants/:id`, I would leave out `/product/:id/variant/:variant_id`. 

Sure, let's leave things as is and revisit later on.

PR #6581, opened 2024-03-04
Chore/integration tests modules utils

adrien2p:
new wrapper for medusa integration tests.
for now it is only applied to the modules directory, but it could be used in the api integration tests or any other integrations that requires a db and a server up and running. 

It is not perfect, but I wanted to have something working and centralised before improving it, also avoiding too many conflicts with other prs


adrien2p:
i ll try to play quickly with some config

sradevski (approved):

adrien2p:
I ve been caught by things merged in develop, I need to fix all the new integration tests, I ll continue tomorrow

PR #6586, opened 2024-03-05
fix: Add some tests and test cleanup for product module

sradevski:


packages/medusa/src/api-v2/admin/products/route.ts
```diff
@@ -22,7 +22,8 @@ export const GET = async (
       skip: req.listConfig.skip,
       take: req.listConfig.take,
     },
-    fields: defaultAdminProductFields,
+    // TODO: How do we include the relations as well as defined in query-config? Or will remote query work differently?
```
sradevski: @adrien2p let's say I want to get the variants relationship from the FE, how is that expressed with remoteQuery?

packages/medusa/src/api-v2/admin/products/query-config.ts
```diff
@@ -30,6 +30,10 @@ export const defaultAdminProductFields = [
   // TODO: Handle in next iteration
   // "collection_id",
   // "type_id",
+  // TODO: Can I expand the fields list from the HTTP request, or does it get replaced instead? What purpose do the allowed and default relations fields have?
```
sradevski: I don't want these in the defaults so we have 1:1 compatability with v1, I just added them so my test passes until I get an answer on how these relationships should be expanded with remoteQuery

packages/medusa/src/api-v2/admin/products/route.ts
adrien2p: you can just add it in the remote query, the transform to remote query object should do it

packages/medusa/src/api-v2/admin/products/query-config.ts
adrien2p: As I mentioned somewhere, in the future the idea will be that maybe a specific prop on the req could allow a user to alter allowed and default. Also, from the http currently, there are fields and expand, but the idea would be to only keep one and also have something (i proposed +,-, nothing as the prefix of props and relation if needed) to alter the default fields and not just replace all unless there is no prefix. Once we have that and only a single array instead of default relations and default fields which can be merged, then we can update the transformer to always provide both `select, relations` and `fields` which can be used with the remote query object utils to transform it and it will auto include the user request

adrien2p: don't know if I explained it well but we can discuss that later (I can whisper only ahah)

sradevski: Yeah I understand, I'll just need a concrete example to see how all of this would look like, I'll get lunch now let's chat later 😄

integration-tests/helpers/product-seeder.js
```diff
@@ -16,11 +16,11 @@ const {
 module.exports = async (dataSource, data = {}) => {
   const manager = dataSource.manager
 
-  const defaultProfile = await manager.findOne(ShippingProfile, {
+  const defaultProfile = (await manager.findOne(ShippingProfile, {
     where: {
       type: ShippingProfileType.DEFAULT,
     },
-  })
+  })) || { id: "default-profile" }
```
adrien2p, line 23: q: why do we need that?

integration-tests/modules/__tests__/product/admin/create-product.spec.ts
```diff
@@ -0,0 +1,494 @@
...
+          expect.objectContaining({
+            id: expect.any(String),
+            url: "test-image-2.png",
+            created_at: expect.any(String),
+            updated_at: expect.any(String),
+          }),
+        ])
+      )
+
+      console.log(response?.data.product)
```
adrien2p, line 138: todo: rm

integration-tests/modules/__tests__/product/admin/create-product.spec.ts
```diff
@@ -0,0 +1,494 @@
...
+
+      expect(response?.data.product.variants).toEqual(
```
adrien2p, line 140: todo: for all array it would be nice to validate length too

packages/medusa/src/api-v2/admin/products/query-config.ts
```diff
@@ -82,4 +83,30 @@ export const defaultAdminProductsVariantFields = [
   "barcode",
 ]
 
+export const retrieveVariantConfig = {
+  defaultFields: defaultAdminProductsVariantFields,
+  defaultRelations: [],
+  allowedRelations: [],
```
adrien2p, line 89: suggestion: I believe in the admin we don't event need to specify those options

packages/medusa/src/api-v2/admin/products/query-config.ts
```diff
@@ -82,4 +83,30 @@ export const defaultAdminProductsVariantFields = [
...
+  defaultLimit: 50,
+  isList: true,
+}
+
 export const defaultAdminProductsOptionFields = ["id", "title"]
+
+export const retrieveOptionConfig = {
+  defaultFields: defaultAdminProductsOptionFields,
+  defaultRelations: [],
+  allowedRelations: [],
```
adrien2p, line 104: same

adrien2p (approved):
I believe we can start from there, I left just a couple of comment but feel free to manage that in a follow up pr

integration-tests/helpers/product-seeder.js
sradevski: The tests were failing when running the product-seeder (the tests were skipped before), so I just added it as a fallback.

PR #6588, opened 2024-03-05
chore(tests): Try to use the api integration tests for v2

sradevski:
Few things to keep in mind:
1. You need to set MEDUSA_FF_MEDUSA_V2 to true before running the tests to run with the v2 API
2. You can use the `breaking` function to differentiate between v1 and v2 differences. This can help us identify what was breaking pretty quickly afterwards
3. You will need to run specific tests for now instead of all if you want to target v2. I think that's fine though, as we don't really need these to run on every PR until we have feature parity (and by then, all tests would be both v1 and v2 compatible)

**note: Adrien** 
- add a new way to load modules only to run their loaders comparable to the way to run the migrations only
- improve tests runner to cleanup the data properly as well as re running all loaders and core defaults


adrien2p (approved):
Sounds pretty good to me 🔥

adrien2p:
@carlos-r-l-rodrigues I ve added a new medusa app loader just to be able to run all modules loader only a bit like we did for the migration only. We can discuss that tomorrow if you want :)

sradevski:
I'll get this merged, and we can discuss the new loader separately if needed, since it only affects the tests

PR #6590, opened 2024-03-05
feat(dashboard) admin 3.0 create discounts form

fPolic:
**WHAT**
- added create discounts form

**NOTE**
- conditions sections will be added in a separate PR

### DIFF
--- a/packages/admin-next/dashboard/src/routes/discounts/create/components/create-discount-form/create-discount-form.tsx
+++ b/packages/admin-next/dashboard/src/routes/discounts/create/components/create-discount-form/create-discount-form.tsx
@@ -138,6 +138,16 @@ export const CreateDiscountForm = () => {
       return getDbAmount(amount, region!.currency_code)
     }
 
+    const duration = {
+      years: values.years,
+      months: values.months,
+      days: values.days,
+      hours: values.hours,
+      minutes: values.minutes,
+    }
+
+    const isDurationEmpty = Object.values(duration).every((v) => !v)
+
     await mutateAsync(
       {
         code: values.code,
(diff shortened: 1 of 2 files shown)


olivermrbl:
`Redemptions limit` and `Duration of the discount` are not included in the request to the backend

fPolic:
> `Redemptions limit` and `Duration of the discount` are not included in the request to the backend

@olivermrbl - mixup with filed names, should be fixed now

packages/admin-next/dashboard/src/routes/discounts/create/components/create-discount-form/create-discount-form.tsx
```diff
@@ -135,6 +135,8 @@ export const CreateDiscountForm = () => {
       return getDbAmount(amount, region!.currency_code)
     }
 
+    console.log(values)
```
olivermrbl: todo: rm

olivermrbl (approved):
Very nice! Will let @kasperkristensen glance over the implementation too

packages/admin-next/dashboard/src/routes/discounts/create/components/create-discount-form/create-discount-details.tsx
```diff
@@ -11,56 +27,642 @@ type CreateDiscountPropsProps = {
...
+                        />
+                      </Form.Control>
+                    </div>
+                    <Form.Hint className="!mt-1">
+                      {t("discounts.startDateHint")}
+                    </Form.Hint>
+                    <Form.ErrorMessage />
+                  </Form.Item>
+                )}
+              />
```
kasperkristensen, line 326: This applies to all "sections" with a Switch. 

The intended design is that you click the switch to show the input. If you then disable the Switch it should hide the input and then the value should be undefined when submitting.

Then in the Edit form if the value exists on the Discount then it should be checked per default.

kasperkristensen (commented):
LGTM, just the "Switch" sections need to be modified to have the same behaviour as everywhere else

packages/admin-next/dashboard/src/routes/discounts/create/components/create-discount-form/create-discount-details.tsx
fPolic: Should be good now

olivermrbl (commented):
nitpicking: If I flip the "Availability duration" switch but don't enter any details it will send them as `P0Y0M0DT0H0M0S` i.e. be expired.

olivermrbl:
Nice, should be good now

PR #6592, opened 2024-03-06
feat(pricing,medusa,utils): added list + get endpoints for price lists

riqwan:
what:

- brings back price list GET endpoints to v2 endpoints

### DIFF
--- a/integration-tests/api/__tests__/store/orders.js
+++ b/integration-tests/api/__tests__/store/orders.js
@@ -290,27 +290,29 @@ describe("/store/carts", () => {
         "/store/orders/order_test?fields=status&expand=billing_address"
       )
 
-      expect(Object.keys(response.data.order)).toEqual([
-        // fields
-        "status",
+      expect(Object.keys(response.data.order).sort()).toEqual(
+        [
+          // fields
+          "status",
 
-        // selected relations
-        "billing_address",
+          // selected relations
+          "billing_address",
 
-        // totals
-        "shipping_total",
-        "discount_total",
-        "tax_total",
-        "refunded_total",
-        "total",
-        "subtotal",
-        "paid_total",
-        "refundable_amount",
-        "gift_card_total",
-        "gift_card_tax_total",
-        "item_tax_total",
-        "shipping_tax_total",
-      ])
+          // totals
+          "shipping_total",
+          "discount_total",
+          "tax_total",
+          "refunded_total",
+          "total",
+          "subtotal",
+          "paid_total",
+          "refundable_amount",
+          "gift_card_total",
+          "gift_card_tax_total",
+          "item_tax_total",
+          "shipping_tax_total",
+        ].sort()
+      )
     })
 
     it("looks up order", async () => {


packages/medusa/src/api/routes/admin/price-lists/modules-queries/get-price-list.ts
```diff
@@ -12,9 +12,7 @@ export async function getPriceListPricingModule(
   }
 ): Promise<PriceList> {
   const [priceLists, count] = await listAndCountPriceListPricingModule({
-    filters: {
-      id: [id],
-    },
+    filters: { id: [id] },
```
sradevski, line 15: Doesn't this work if it is `filters: { id }`?

packages/medusa/src/api-v2/admin/price-lists/query-config.ts
```diff
@@ -0,0 +1,63 @@
+import { prefixArrayItems } from "@medusajs/utils"
+import { defaultAdminProductsVariantFields } from "../products/query-config"
+
+export const defaultAdminPriceListFields = [
```
sradevski, line 40: Does this match what we had in v1? If we can match it than that'd be great, if not then we'll have to live with the breaking changes :)

riqwan: Yes, the v2 response matches v1 response. The transformation of v2 to v1 happens in the utils function. This is because the data model of pricing between v1 and v2 is vastly different. 

This variable and the ones beneath this is however wrong though. Its meant to showcase a list of attributes that should be present in the response. But in reality, this is a list of remote query fields to query through the price_list entry point. 

I'll look into it a bit more to see how we can clearly depict the 2.

packages/medusa/src/api/routes/admin/price-lists/modules-queries/get-price-list.ts
riqwan: It will, this file will be deleted soon though 😄 its legacy stuff. Just waiting on one more endpoint before I nuke this.

packages/medusa/src/api-v2/admin/price-lists/query-config.ts
sradevski: Ok cool, there is a lot going on here so I won't pretend I can judge what's a good approach, I was just wondering if before as well we expanded so much by default

packages/medusa/src/api-v2/admin/price-lists/query-config.ts
```diff
@@ -0,0 +1,63 @@
...
+  "price_set_money_amounts.money_amount.max_quantity",
+  "price_set_money_amounts.money_amount.created_at",
+  "price_set_money_amounts.money_amount.deleted_at",
+  "price_set_money_amounts.money_amount.updated_at",
+  "price_set_money_amounts.price_rules.value",
+  "price_set_money_amounts.price_rules.rule_type.rule_attribute",
+  ...prefixArrayItems(
+    defaultAdminProductsVariantFields,
+    "price_set_money_amounts.price_set.variant_link.variant."
+  ),
```
olivermrbl: thought: Should we add a `fieldAlias` to the link definition, so we don't have to query this `variant_link` directly?

Something like:
```
{
      serviceName: Modules.PRICING,
      relationship: {
        serviceName: LINKS.ProductVariantPriceSet,
        primaryKey: "price_set_id",
        foreignKey: "id",
        alias: "variant_link",
      },
      fieldAlias: {
        variant: "variant_link.variant"
      },
    },
```

This would reduce the field to `price_set_money_amounts.price_set.variant`

olivermrbl: Also, maybe we should review the UI to see if the variants on money amount are still required. It seems more appropriate to do a separate query in case we need them.

integration-tests/modules/__tests__/price-lists/admin/price-lists.spec.ts
```diff
@@ -0,0 +1,310 @@
...
+import { createVariantPriceSet } from "../../../helpers/create-variant-price-set"
+
+jest.setTimeout(50000)
+
+const env = { MEDUSA_FF_MEDUSA_V2: true }
+const adminHeaders = {
+  headers: { "x-medusa-access-token": "test_token" },
+}
+
+describe("Price Lists", () => {
```
olivermrbl: todo: Should be updated to use the new `medusaIntegrationTestRunner`

packages/medusa/src/api-v2/admin/price-lists/query-config.ts
olivermrbl: question: Actually, do we need all of this stuff to be default on the list endpoint? I can see why it could make sense on retrieval, but we don't need all prices and their variants to display the price list table, do we?. 

And if we do, I am almost leaning toward changing the UI rather than having this colossal query

packages/medusa/src/api-v2/admin/price-lists/utils/list-price-lists.ts
```diff
@@ -0,0 +1,90 @@
...
+
+      return {
+        ...priceSetMoneyAmount.money_amount,
+        price_list_id: priceList.id,
+        variant_id: productVariant?.id ?? null,
+        variant: productVariant ?? null,
+        region_id: rules["region_id"] ?? null,
+        rules,
+      }
+    })
```
olivermrbl: See comments on `query-config.ts`

olivermrbl (approved):
LGTM, but let's make sure to align on the query config before we merge

packages/medusa/src/api-v2/admin/price-lists/query-config.ts
riqwan: We added it to make it compatible with the existing API response, back when enabling modules was still backwards compatible. 

We can definitely remove most of it and just keep the ID and query the variants API on the FE.

riqwan: No, we don't. I've made some changes here that aligns with this, will ping you when I've pushed.

olivermrbl: Nice, sounds good!

riqwan: @olivermrbl take a look at the latest commit, this is the prettiest I could make for now. 😄

The `fields` and `expands` is not friendly to the remote query approach, since the api has to know how the public interface is mapped to both the modules.

packages/medusa/src/utils/omit-deep.ts
```diff
@@ -32,3 +32,50 @@ export function omitDeep<T extends object = object>(
     return nextInput
   }, {} as T)
 }
+
+export function pickDeep<T extends object = object>(
```
olivermrbl: nit: Move to separate file `pick-deep.ts`

olivermrbl (commented):
Think this looks better. We just need tests to pass, and I would be fine with merging 💪

PR #6594, opened 2024-03-06
feat: Refresh payment collection + delete session

olivermrbl:
### What
Add workflow for refreshing a payment collection. 

The idea is that on all cart updates, we want two things to happen (in the context of payments) 1. the currently active payment sessions should be destroyed, and 2. the payment collection should be updated with the new cart total.

We do this to ensure that we always collect the correct payment amount. 

From a customer perspective, this would mean that every time something on the cart is updated, the customer would need to enter their payment details anew. 

To me, this is a good tradeoff to avoid inconsistencies with payment collection.

Additionally, I updated the Payment Module interface with `upsert` and `updated` following our established convention.

### Note
This PR depends on a fix to the `remoteJoiner` that @carlos-r-l-rodrigues is working on.

Update: Fix merged in #6602

### DIFF
--- a/integration-tests/api/__tests__/admin/payment.js
+++ b/integration-tests/api/__tests__/admin/payment.js
@@ -1,17 +1,5 @@
-const path = require("path")
-
-const startServerWithEnvironment =
-  require("../../../environment-helpers/start-server-with-environment").default
-const { useApi } = require("../../../environment-helpers/use-api")
-const { useDb } = require("../../../environment-helpers/use-db")
-const adminSeeder = require("../../../helpers/admin-seeder")
-
-const {
-  simplePaymentCollectionFactory,
-} = require("../../../factories/simple-payment-collection-factory")
-const {
-  simpleCustomerFactory,
-} = require("../../../factories/simple-customer-factory")
+const { medusaIntegrationTestRunner } = require("medusa-test-utils")
+const { createAdminUser } = require("../../../helpers/create-admin-user")
 
 jest.setTimeout(30000)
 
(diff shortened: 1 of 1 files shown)


packages/core-flows/src/definition/payment-collection/steps/delete-payment-session.ts
```diff
@@ -0,0 +1,46 @@
...
+  async (input, { container }) => {
+    const service = container.resolve<IPaymentModuleService>(
+      ModuleRegistrationName.PAYMENT
+    )
+
+    if (!input || !input.payment_collection) {
+      return
+    }
+
+    await service.createPaymentSession(input.payment_collection.id, {
```
olivermrbl, line 38: comment: Intentionally hard-deleting and creating a new payment session, as we cannot restore the session at the third party provider

packages/types/src/payment/mutations.ts
```diff
@@ -35,21 +34,24 @@ export interface UpdatePaymentCollectionDTO
...
+}
 
-  /**
-   * The status of the payment collection.
-   */
-  status?: PaymentCollectionStatus
+/**
+ * The attributes to update in the payment collection.
+ */
+export interface PaymentCollectionUpdatableFields {
```
sradevski, line 50: I think we can call these `UpdatePaymentCollectionDTO` to be consistent with create and upsert, and really the mean the same thing.

packages/core-flows/src/definition/payment-collection/steps/delete-payment-session.ts
sradevski: Makes sense 👍

olivermrbl:
Converting to draft while we are debugging an issue with the Workflows SDK

packages/core-flows/src/common/steps/use-remote-query.ts
```diff
@@ -0,0 +1,29 @@
...
+    const queryObject = remoteQueryObjectFromString({
+      entryPoint: data.entry_point,
+      fields: data.fields,
+      variables: data.variables,
+    })
+
+    const result = await query(queryObject)
+
+    console.log("Remote query object: ", queryObject)
+    console.log("Remote query result: ", result)
```
adrien2p: todo: rm

integration-tests/modules/__tests__/cart/store/cart.workflows.spec.ts
```diff
@@ -664,18 +666,151 @@ medusaIntegrationTestRunner({
...
+                  payment_collection: undefined,
+                }),
+              ])
+              expect(payCols.length).toEqual(0)
+            })
           })
+        })
+      })
 
+      describe.only("refreshPaymentCollectionForCart", () => {
```
adrien2p: todo: rm only

packages/core-flows/src/definition/cart/workflows/refresh-payment-collection.ts
```diff
@@ -0,0 +1,75 @@
...
+  cart_id: string
+}
+
+// We export a step running the workflow too, so that we can use it as a subworkflow e.g. in the update cart workflows
+export const refreshPaymentCollectionForCartStepId =
+  "refresh-payment-collection-for-cart"
+export const refreshPaymentCollectionForCartStep = createStep(
+  refreshPaymentCollectionForCartStepId,
+  async (data: StepInput, { container }) => {
+    await refreshPaymentCollectionForCartWorkflow(container).run({
```
adrien2p, line 27: discussion: As I was explaining, using workflows in workflows can be useful, but there is no easy way to no that a workflow is being used in another workflow and that can bring some downside. As usual, responsibility principle and domain concerns must be taken into consideration and updating a workflow should necessarely impact the workflows that are using another workflow as the changes might not be related but still will impact the other workflows. So I just want to put a warning on those usage to be more cautious in those context

packages/core-flows/src/definition/cart/workflows/refresh-payment-collection.ts
```diff
@@ -0,0 +1,75 @@
...
+)
+
+export const refreshPaymentCollectionForCartWorkflowId =
+  "refresh-payment-collection-for-cart"
+export const refreshPaymentCollectionForCartWorkflow = createWorkflow(
+  refreshPaymentCollectionForCartWorkflowId,
+  (input: WorkflowData<WorklowInput>): WorkflowData<void> => {
+    const test = transform({ input }, (data) => {
+      console.log("Workflow >>>: ", data.input)
+    })
```
adrien2p: todo:" rm

packages/core-flows/src/definition/cart/workflows/refresh-payment-collection.ts
```diff
@@ -0,0 +1,75 @@
...
+        "total",
+        "currency_code",
+        "payment_collection.id",
+        "payment_collection.payment_sessions.id",
+      ],
+      variables: { id: input.cart_id, test },
+    })
+
+    deletePaymentSessionStep({
+      payment_session_id: carts[0].payment_collection.payment_sessions?.[0].id,
```
adrien2p, line 55: 🙏 it is super cool to get this works knowing that at that time the carts does not even exists in itself but it is resolved in the future :)

packages/core-flows/src/definition/payment-collection/steps/update-payment-collection.ts
```diff
@@ -0,0 +1,61 @@
...
+  },
+  async (prevData, { container }) => {
+    if (!prevData) {
+      return
+    }
+    const paymentModuleService = container.resolve<IPaymentModuleService>(
+      ModuleRegistrationName.PAYMENT
+    )
+
+    console.log("PREV DATA: ", prevData)
```
adrien2p: todo: rm

adrien2p: I might miss other logs like that so maybe a look around would be good :)

packages/payment/src/migrations/Migration20240225134525.ts
```diff
@@ -28,6 +28,7 @@ export class Migration20240225134525 extends Migration {
         ALTER TABLE IF EXISTS "payment_session" ADD COLUMN IF NOT EXISTS "payment_authorized_at" TIMESTAMPTZ NULL;
         ALTER TABLE IF EXISTS "payment_session" ADD COLUMN IF NOT EXISTS "raw_amount" JSONB NOT NULL;
         ALTER TABLE IF EXISTS "payment_session" ADD COLUMN IF NOT EXISTS "deleted_at" TIMESTAMPTZ NULL;
+        ALTER TABLE IF EXISTS "payment_session" ADD COLUMN IF NOT EXISTS "context" JSONB NULL;
```
adrien2p, line 31: suggestion: I believe the payment module does not have the appropriate config for the dev mikro config and is outdated in terms of loaders etc, should we take care of that, not necessecarly here

packages/payment/src/models/payment-session.ts
```diff
@@ -43,6 +43,9 @@ export default class PaymentSession {
   @Property({ columnType: "jsonb" })
   data: Record<string, unknown> = {}
 
+  @Property({ columnType: "jsonb", nullable: true })
+  context?: Record<string, unknown> | null
```
adrien2p: todo: rm ?

packages/types/src/payment/common.ts
```diff
@@ -417,6 +417,11 @@ export interface PaymentSessionDTO {
    */
   data: Record<string, unknown>
 
+  /**
+   * Payment session context
+   */
+  context?: Record<string, unknown>
```
adrien2p, line 423: discussion: in the DTO's that reflect a public model API I don't think we should have `?` but we can take care of that at a later point

packages/workflows-sdk/src/helper/workflow-export.ts
```diff
@@ -183,6 +183,7 @@ function createContextualWorkflowRunner<
       throw new Error(errorMessage)
     }
 
+    console.log(resultFrom, transaction.getContext())
```
adrien2p: todo: rm

adrien2p (approved):
LGTM few comments 🔥

packages/payment/src/models/payment-session.ts
olivermrbl: I added the context, so we have the data for when we need to create a new payment session in the compensation flow of `deletePaymentSession`. The alternative would be to not have a compensation step, because the context is a required property for the creation at the third-party provider.

packages/core-flows/src/definition/cart/workflows/refresh-payment-collection.ts
olivermrbl: Yes, indeed - super cool!

packages/core-flows/src/definition/cart/workflows/refresh-payment-collection.ts
olivermrbl: We will likely need to address this more generally for 2.0 to have a pattern in place.

packages/payment/src/models/payment-session.ts
olivermrbl: Let me know what you think.

adrien2p: I meant remove the `?` symbol 😂

olivermrbl: Ahh haha, I thought you were asking a question because of the `?` 😆

PR #6595, opened 2024-03-06
fix: integration-tests/modules

adrien2p:


integration-tests/modules/package.json
```diff
@@ -5,7 +5,7 @@
   "license": "MIT",
   "private": true,
   "scripts": {
-    "test:integration": "node --expose-gc ./../../node_modules/.bin/jest --ci --silent=false -i --detectOpenHandles --logHeapUsage --forceExit",
+    "test:integration": "node --expose-gc ./../../node_modules/.bin/jest --ci --silent=true --concurrency=50% --detectOpenHandles --logHeapUsage --forceExit -- __tests__/workflow-engine/api.ts",
```
riqwan:
```suggestion
    "test:integration": "node --expose-gc ./../../node_modules/.bin/jest --ci --silent=true --concurrency=50% --detectOpenHandles --logHeapUsage --forceExit",
```

olivermrbl:
@adrien2p, looks like one suite is still failing 🤔

integration-tests/helpers/admin-seeder.js
```diff
@@ -1,5 +1,5 @@
 const Scrypt = require("scrypt-kdf")
-const { User } = require("@medusajs/medusa")
+const { User } = require("@medusajs/medusa/dist/models/user")
```
sradevski, line 2: Why was this change needed?

sradevski (approved):

adrien2p:
@olivermrbl I was looking at it but from some reason the bot still merge it :/

integration-tests/helpers/admin-seeder.js
adrien2p: Because importing from the entry file is causing issues since all import will get evaluated, meaning all models to and therefore some decorators relying on the feature flags gets applied before they are loaded

adrien2p: this is some trouble I ve seen multiple times already and they are annoying, same when a file import from an index file, the other import gets evaluated but not necessarely at the right moment, we can re discuss that later

PR #6596, opened 2024-03-06
feat(inventory-next, types): inventory module conversion

pKorsholm:
**What**
- update inventory package to use the new utilities and mikro-orm
- introduce this new package as inventory-next

### DIFF
--- a/packages/medusa-test-utils/src/module-test-runner.ts
+++ b/packages/medusa-test-utils/src/module-test-runner.ts
@@ -22,7 +22,6 @@ export function moduleIntegrationTestRunner({
   joinerConfig = [],
   schema = "public",
   debug = false,
-  resolve,
   testSuite,
   resolve,
   injectedDependencies = {},
@@ -33,7 +32,6 @@ export function moduleIntegrationTestRunner({
   joinerConfig?: any[]
   schema?: string
   dbName?: string
-  resolve?: string
   injectedDependencies?: Record<string, any>
   resolve?: string
   debug?: boolean


olivermrbl:
@pKorsholm is this ready for review?

pKorsholm:
@olivermrbl yes, just had to review and cleanup

olivermrbl:
Epic 💪

packages/inventory-next/package.json
```diff
@@ -0,0 +1,58 @@
+{
+  "name": "@medusajs/inventory-next",
+  "version": "0.0.1",
+  "description": "Inventory Module for Medusa",
+  "main": "dist/index.js",
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/medusajs/medusa",
+    "directory": "packages/inventor-new"
```
olivermrbl: nit
```suggestion
    "directory": "packages/inventory-next"
```

packages/inventory-next/src/services/index.ts
```diff
@@ -0,0 +1,4 @@
+// export { default as InventoryItemService } from "./inventory-item"
```
olivermrbl: question: Is this a TODO?

packages/inventory-next/src/services/inventory-level.ts
```diff
@@ -0,0 +1,111 @@
...
+   * @param quantity - The quantity to adjust from the reserved quantity.
+   * @param context
+   */
+  async adjustReservedQuantity(
+    inventoryItemId: string,
+    locationId: string,
+    quantity: number,
+    context: Context = {}
+  ): Promise<void> {
+    // const manager = context.transactionManager!
```
olivermrbl: question: Is this a TODO?

packages/inventory-next/src/models/inventory-level.ts
```diff
@@ -0,0 +1,103 @@
...
+  metadata: Record<string, unknown> | null
+
+  @ManyToOne(() => InventoryItem, {
+    persist: false,
+  })
+  inventory_item: InventoryItem
+
+  @BeforeCreate()
+  private beforeCreate(): void {
+    this.id = generateEntityId(this.id, "ilev")
```
adrien2p, line 96: q: should we add `this.inventory_item_id ??= this.inventory_item?.id` just as safe guard? I am still unsure about it

packages/utils/src/common/partition-array.ts
```diff
@@ -0,0 +1,17 @@
+export const partitionArray = <T>(
```
adrien2p, line 14: suggestion: tsdoc + example + tests :)

packages/inventory-next/src/services/inventory.ts
```diff
@@ -0,0 +1,1053 @@
...
+  inventoryLevelService: InventoryLevelService<any>
+  reservationItemService: ModulesSdkTypes.InternalModuleService<any>
+}
+
+const generateMethodForModels = [
+  { model: InventoryItem, singular: "InventoryItem", plural: "InventoryItems" },
+  {
+    model: InventoryLevel,
+    singular: "InventoryLevel",
+    plural: "InventoryLevels",
```
adrien2p: suggestion: rm

packages/inventory-next/src/services/inventory.ts
```diff
@@ -0,0 +1,1053 @@
...
+  },
+  {
+    model: ReservationItem,
+    singular: "ReservationItem",
+    plural: "ReservationItems",
```
adrien2p: suggestion: rm

packages/inventory-next/src/services/inventory.ts
```diff
@@ -0,0 +1,1053 @@
...
+
+type InjectedDependencies = {
+  baseRepository: DAL.RepositoryService
+  inventoryItemService: ModulesSdkTypes.InternalModuleService<any>
+  inventoryLevelService: InventoryLevelService<any>
+  reservationItemService: ModulesSdkTypes.InternalModuleService<any>
+}
+
+const generateMethodForModels = [
+  { model: InventoryItem, singular: "InventoryItem", plural: "InventoryItems" },
```
adrien2p: suggestion: same singular/plural should be auto infered here too

packages/inventory-next/src/services/inventory.ts
```diff
@@ -0,0 +1,1053 @@
...
+        dto: InventoryNext.InventoryLevelDTO
+        singular: "InventoryLevel"
+        plural: "InventoryLevels"
+      }
+      ReservationItem: {
+        dto: InventoryNext.ReservationItemDTO
+        singular: "ReservationItem"
+        plural: "ReservationItems"
+      }
+    }
```
adrien2p: suggestion: same the inference should be enough no?

packages/inventory-next/src/utils/build-query.ts
```diff
@@ -0,0 +1,166 @@
+import { ExtendedFindConfig, FindConfig } from "@medusajs/types"
```
olivermrbl: question: Leftover from copy/paste? Doesn't seem to be relevant any more

packages/inventory-next/src/services/inventory.ts
```diff
@@ -0,0 +1,1053 @@
...
+    this.inventoryItemService_ = inventoryItemService
+    this.inventoryLevelService_ = inventoryLevelService
+    this.reservationItemService_ = reservationItemService
+  }
+
+  __joinerConfig(): ModuleJoinerConfig {
+    return joinerConfig
+  }
+
+  @InjectTransactionManager("baseRepository_")
```
adrien2p: q: do we need that on private method?

packages/inventory-next/src/utils/query.ts
```diff
@@ -0,0 +1,93 @@
+import { Brackets, EntityManager, FindOptionsWhere } from "typeorm"
```
olivermrbl: comment: A lot of stuff commented out here. Please add todos or remove.

Also, why are we mixing typeorm and MikroORM types? This module should not have a dep on typeorm

packages/types/src/inventory/index.ts
```diff
@@ -1,2 +1,7 @@
+// import * as InventoryCommon from "./common"
+// import * as InventoryMutations from "./mutations"
+// export const InventoryNext = { ...InventoryCommon, ...InventoryMutations }
```
olivermrbl: question: Is this a TODO?

packages/inventory-next/src/services/inventory.ts
```diff
@@ -0,0 +1,1053 @@
...
+    model: InventoryLevel,
+    singular: "InventoryLevel",
+    plural: "InventoryLevels",
+  },
+  {
+    model: ReservationItem,
+    singular: "ReservationItem",
+    plural: "ReservationItems",
+  },
+]
```
olivermrbl: todo: `singular` and `plural` are not needed, you can flatten to string[]

packages/inventory-next/src/services/inventory.ts
```diff
@@ -0,0 +1,1053 @@
...
+      InventoryLevel: {
+        dto: InventoryNext.InventoryLevelDTO
+        singular: "InventoryLevel"
+        plural: "InventoryLevels"
+      }
+      ReservationItem: {
+        dto: InventoryNext.ReservationItemDTO
+        singular: "ReservationItem"
+        plural: "ReservationItems"
+      }
```
olivermrbl: Same here

packages/inventory-next/src/utils/build-query.ts
adrien2p: question: is this necessary now? shouldn't the build query utils used internally be enough?

adrien2p: also it seems more typeorm related right?

olivermrbl (commented):
I think there is still some cleaning up to do

adrien2p:
Strong work 💪 I ve got few questions

packages/inventory-next/src/services/inventory.ts
pKorsholm: probably not as this should always have a transaction if it's called 😅 I'll remove it

packages/inventory-next/src/models/inventory-item.ts
```diff
@@ -0,0 +1,120 @@
...
+    defaultRaw: "now()",
+  })
+  updated_at: Date
+
+  @InventoryItemDeletedAtIndex.MikroORMIndex()
+  @Property({ columnType: "timestamptz", nullable: true })
+  deleted_at: Date | null = null
+
+  @InventoryItemSkuIndex.MikroORMIndex()
+  @Property({ columnType: "text", nullable: true })
```
sradevski, line 63: A very general question: We have a lot of these fields on the product model as well, which causes for a lot of duplication and chance of these two getting out of sync. Which one is the source of truth? I don't know much about the logic behind the inventory module, but I see the benefit of having this information on the inventory item itself, however we have to ensure the two are in sync.

Or should the inventory item just reference the product?

packages/inventory-next/src/models/reservation-item.ts
```diff
@@ -0,0 +1,95 @@
...
+  @Property({ columnType: "timestamptz", nullable: true })
+  deleted_at: Date | null = null
+
+  @ReservationItemLineItemIdIndex.MikroORMIndex()
+  @Property({ type: "text", nullable: true })
+  line_item_id: string | null = null
+
+  @ReservationItemInventoryItemIdIndex.MikroORMIndex()
+  @Property({ type: "text" })
+  inventory_item_id: string
```
sradevski: Should this and the location_id be a foreign key?

packages/inventory-next/src/models/inventory-level.ts
```diff
@@ -0,0 +1,104 @@
...
+    type: "text",
+    mapToPk: true,
+    onDelete: "cascade",
+  })
+  @InventoryLevelInventoryItemIdIndex.MikroORMIndex()
+  inventory_item_id: string
+
+  @InventoryLevelLocationIdIndex.MikroORMIndex()
+  @Property({ type: "text" })
+  location_id: string
```
sradevski, line 75: Should the location_id be a foreign key?

packages/inventory-next/src/services/inventory.ts
```diff
@@ -0,0 +1,1034 @@
...
+    this.inventoryItemService_ = inventoryItemService
+    this.inventoryLevelService_ = inventoryLevelService
+    this.reservationItemService_ = reservationItemService
+  }
+
+  __joinerConfig(): ModuleJoinerConfig {
+    return joinerConfig
+  }
+
+  private async ensureInventoryLevels(
```
sradevski, line 88: The naming is a bit confusing, shouldn't it just be `listInventoryLevels`?

packages/inventory-next/src/models/reservation-item.ts
pKorsholm: we can't foreign key location_id as the locations are stored in a separate module

packages/inventory-next/src/models/inventory-level.ts
pKorsholm: we can't foreign key location_id as the locations are stored in a separate module

packages/inventory-next/src/services/inventory.ts
pKorsholm: it's not exactly `listInventoryLevels` as the purpose is to validate that the inventory levels being requested for creating reservations/updating levels and the arguments are distinct from what would be expected for `listInventoryLevels` based on what we usually provide as arguments. It's also not in the expected behavior of our listEndpoints to throw if all records are not found based on the provided criteria.

 We could rename this as `ensureInventoryLevelsExist`, `listAndValidateInventoryLevels` or `listInventoryLevelsOrThrow` instead, but I don't this just calling this `listInventoryLevels` is an appropriate solution. wdyt?

packages/inventory-next/src/models/inventory-item.ts
pKorsholm: there is a relation between product-variants and inventory-items `productVariantInventoryItem` (described in the link-modules package`). Given that both packages are installed and running this should be the source of truth rather than then product-variant fields. 

If we expect to always be running both packages in v2 it might make sense to remove the fields from the product-module @adrien2p ?

adrien2p: Yeah the idea was mostly to be backward compatible since the product module was also used in v1, but more generally the inventory on module should probably not even exists and creates some nasty side effects when customers are buying the same variants. Probably in v2 we might have either inventory management on or off, maybe we should validate that more globally @olivermrbl

olivermrbl (approved):
Nice work @pKorsholm 💪 

Imo. fine to merge when tests are passing.

packages/inventory-next/src/models/reservation-item.ts
sradevski: Ah I assumed it's part of the inventory module 😄

packages/inventory-next/src/services/inventory.ts
sradevski: listAndValidateInventoryLevels sounds good to me, as we actually do both here 👍

packages/inventory-next/src/models/inventory-item.ts
srindom: A product variant can be made up from x number of inventory items. An inventory item is the physical thing being shipped and a product variant is the commercial representation of that thing. For example, a customer will buy an "oval table" which is the product variant, but the variant consists of 4 x "table legs" and 1 x "table top". 

The duplication of these fields is, as Adrien mentions, because we introduced the inventory module as a way to remain backward compatible with what was there in the past (where inventory could also be tracked on the `inventory_quantity` field on product variants). In v2, we can enforce that anyone who needs inventory tracking should use the inventory module, so cleaning up these fields would be sensible.

I would keep SKU on product variants since that could still make sense but, hs_code, origin_country, measurements, etc. would all make sense to remove.

srindom (approved):
LGTM

PR #6601, opened 2024-03-06
feat: Capture payment

olivermrbl:
**What**

Add capture payment flow to v2 Rest API

**Discussion**

For internal module calculations, I would like to use `bignumber.js` library as with the totals calculations. However, it seems misplaced to introduce utility functions in `@medusajs/utils` for something so specific as calculating the fully captured amount. This is very internal to the payment module.

In this PR, I've added `bignumber.js` as a dependency in the payment module. I am curious to hear what others think about this. 

With all that said, I am open to introducing these utilities, as I also see why it would be good not to have this dependency live in the payment module. Perhaps, this could be decorators, similar to how we imagine the cart totals enrichment will be built.


packages/core-flows/src/payment/steps/capture-payment.ts
```diff
@@ -0,0 +1,22 @@
+import { ModuleRegistrationName } from "@medusajs/modules-sdk"
+import { IPaymentModuleService } from "@medusajs/types"
+import { StepResponse, createStep } from "@medusajs/workflows-sdk"
+
+type StepInput = {
+  payment_id: string
+  captured_by?: string
```
fPolic, line 7: q: should we be able to pass an amount as an argument for partial capture?

packages/payment/integration-tests/__tests__/services/payment-module/index.spec.ts
```diff
@@ -515,61 +514,104 @@ moduleIntegrationTestRunner({
...
                 ],
+                captured_at: expect.any(Date),
+              })
+            )
+          })
 
-                // TODO: uncomment when totals calculations are implemented
-                // captured_amount: 100,
-                // captured_at: expect.any(Date),
+          it("should split a payment in two captures a payment successfully", async () => {
```
fPolic, line 522: nit: typo

packages/payment/src/services/payment-module.ts
```diff
@@ -456,13 +485,13 @@ export default class PaymentModuleService<
...
     )
 
-    // TODO: revisit when https://github.com/medusajs/medusa/pull/6253 is merged
-    // if (payment.captured_amount + data.amount === payment.amount) {
-    //   await this.paymentService_.update(
-    //     { id: payment.id, captured_at: new Date() },
-    //     sharedContext
-    //   )
-    // }
+    // When the entire authorized amount has been captured, we mark it fully capture by setting the captured_at field
```
fPolic, line 488: q: should we mark the collection as `completed_at` as well here?

fPolic (approved):
LGTM!

packages/payment/src/services/payment-module.ts
```diff
@@ -409,33 +416,55 @@ export default class PaymentModuleService<
   ): Promise<PaymentDTO> {
     const payment = await this.paymentService_.retrieve(
```
sradevski, line 417: Maybe we don't need to handle it right away, but esp. when it comes to payments we need to ensure there are no race conditions (and currently there are few scenarios where they can occur). Just something to keep in mind

sradevski (approved):

packages/payment/src/services/payment-module.ts
olivermrbl: For now, I think we should leave this decision to the module's consumer. It can be done automatically via a subscriber or manually through a call to complete the payment collection.

packages/core-flows/src/payment/steps/capture-payment.ts
olivermrbl: es, nice catch. Will add in 👍

olivermrbl:
@carlos-r-l-rodrigues, I'll merge this now. Let's discuss the totals element as part of building the utilities.

PR #6602, opened 2024-03-06
Fix(orchestrator): Joiner field alias

carlos-r-l-rodrigues:
What:
  Fix when requesting nested fields of a fieldAlias.


olivermrbl (approved):
Can't say much about the changes here, but I say let's get it in, so we can start to play around with it 😄  

Wdyt?

carlos-r-l-rodrigues:
> Can't say much about the changes here, but I would argue we should get it in so we can start to play around with it 😄
> 
> Wdyt?

that works for me!

PR #6608, opened 2024-03-07
feat: include weight from product or product variant in webshipper order line

madsleejensen:
- will include weight (as grams) when sending the fulfillment order to webshipper

for reference the relevant Webshipper api documentation can be found here: https://docs.webshipper.io/#order_lines

### DIFF
--- a/packages/medusa-fulfillment-webshipper/src/services/webshipper-fulfillment.js
+++ b/packages/medusa-fulfillment-webshipper/src/services/webshipper-fulfillment.js
@@ -1,7 +1,7 @@
 import { humanizeAmount } from "medusa-core-utils"
 import Webshipper from "../utils/webshipper"
 import { AbstractFulfillmentService } from "@medusajs/medusa"
-import { promiseAll } from "@medusajs/utils"
+import { promiseAll, isDefined } from "@medusajs/utils"
 
 class WebshipperFulfillmentService extends AbstractFulfillmentService {
   static identifier = "webshipper"
@@ -701,7 +701,7 @@ class WebshipperFulfillmentService extends AbstractFulfillmentService {
 
     const weight = item?.variant?.weight ?? item?.variant?.product?.weight
 
-    if (weight !== undefined) {
+    if (isDefined(weight)) {
       webShipperItem.weight = weight
       webShipperItem.weight_unit = "g"
     }


packages/medusa-fulfillment-webshipper/src/services/__tests__/webshipper-fulfillment.js
```diff
@@ -323,5 +323,101 @@ describe("WebshipperFulfillmentService", () => {
...
+          totalsService,
+        },
+        {}
+      )
+
+      medusaItem.variant = {
+        weight: 1234
+      }
+
+      medusaItem.variant.product = {}
```
sradevski:
```suggestion
      medusaItem.variant = {
        weight: 1234,
        product: {}
      }
```

packages/medusa-fulfillment-webshipper/src/services/__tests__/webshipper-fulfillment.js
```diff
@@ -323,5 +323,101 @@ describe("WebshipperFulfillmentService", () => {
...
+
+      let item
+      try {
```
sradevski: I think we can omit the `try/catch` in the test if it's not expected

packages/medusa-fulfillment-webshipper/src/services/__tests__/webshipper-fulfillment.js
```diff
@@ -323,5 +323,101 @@ describe("WebshipperFulfillmentService", () => {
...
+
+    it("builds a webshipper item with weight from product", async () => {
+      const webshipper = new WebshipperFulfillmentService(
+        {
+          totalsService,
+        },
+        {}
+      )
+
+      medusaItem.variant = {}
```
sradevski: Same as above

packages/medusa-fulfillment-webshipper/src/services/__tests__/webshipper-fulfillment.js
```diff
@@ -323,5 +323,101 @@ describe("WebshipperFulfillmentService", () => {
...
+
+    it("builds a webshipper item with weight where product variant overrides product", async () => {
+      const webshipper = new WebshipperFulfillmentService(
+        {
+          totalsService,
+        },
+        {}
+      )
+
+      medusaItem.variant = {
```
sradevski, line 388: Same as above, let's combine the setters

packages/medusa-fulfillment-webshipper/src/services/__tests__/webshipper-fulfillment.js
```diff
@@ -323,5 +323,101 @@ describe("WebshipperFulfillmentService", () => {
...
+        weight: 1
+      }
+      medusaItem.variant.product = {
+        weight: 2,
+      }
+
+      let item
+      try {
```
sradevski: Same as above

packages/medusa-fulfillment-webshipper/src/services/__tests__/webshipper-fulfillment.js
```diff
@@ -323,5 +323,101 @@ describe("WebshipperFulfillmentService", () => {
...
+        {}
+      )
+
+      medusaItem.variant = {}
+      medusaItem.variant.product = {
+        weight: 4321,
+      }
+
+      let item
+      try {
```
sradevski: Same as above

packages/medusa-fulfillment-webshipper/src/services/webshipper-fulfillment.js
```diff
@@ -699,6 +699,13 @@ class WebshipperFulfillmentService extends AbstractFulfillmentService {
       webShipperItem.tarif_number = tarifNumber
     }
 
+    const weight = item?.variant?.weight || item?.variant?.product?.weight
+
+    if (weight) {
```
sradevski: I guess it doesn't make a lot of sense for something to have 0 weight and to be shippable, but it's better of we check undefined here. We just need to make sure the weights are not set to 0 by default, so undefined will never happen

packages/medusa-fulfillment-webshipper/src/services/webshipper-fulfillment.js
```diff
@@ -699,6 +699,13 @@ class WebshipperFulfillmentService extends AbstractFulfillmentService {
       webShipperItem.tarif_number = tarifNumber
     }
 
+    const weight = item?.variant?.weight || item?.variant?.product?.weight
```
sradevski:
```suggestion
    const weight = item?.variant?.weight ?? item?.variant?.product?.weight
```

But if the weight is set to 0 by default somewhere, this might cause a bug. Let's check that.

packages/medusa-fulfillment-webshipper/src/services/__tests__/webshipper-fulfillment.js
madsleejensen: good point, just copy pasted the boilerplate from a similar test case.

packages/medusa-fulfillment-webshipper/src/services/webshipper-fulfillment.js
madsleejensen: yeah maybe its ok to allow 0 values to be sent.

packages/medusa-fulfillment-webshipper/src/services/webshipper-fulfillment.js
```diff
@@ -699,6 +699,13 @@ class WebshipperFulfillmentService extends AbstractFulfillmentService {
       webShipperItem.tarif_number = tarifNumber
     }
 
+    const weight = item?.variant?.weight ?? item?.variant?.product?.weight
+
+    if (weight !== undefined) {
```
adrien2p: todo: !isDefined(weight)

madsleejensen:
@sradevski @adrien2p let me know if you have further feedback :)

riqwan:

PR #6609, opened 2024-03-07
chore: Start using medusa test runner on product API tests

sradevski:
I want to get this merged where we only switch to the runner, so actual test changes are more obvious when reviewing.


olivermrbl (approved):
Very nice!

PR #6610, opened 2024-03-07
feat: Refund payment

olivermrbl:
Essentially the same as #6601 but for refunds

### DIFF
--- a/packages/payment/src/services/payment-module.ts
+++ b/packages/payment/src/services/payment-module.ts
@@ -450,9 +450,9 @@ export default class PaymentModuleService<
       )
     }
 
-    const capturedAmount = payment.captures.reduce((acc, next) => {
-      const bn = new BigNumber(next.raw_amount.value)
-      return acc.plus(bn)
+    const capturedAmount = payment.captures.reduce((captureAmount, next) => {
+      const amountAsBigNumber = new BigNumber(next.raw_amount.value)
+      return captureAmount.plus(amountAsBigNumber)
     }, BigNumber(0))
 
     const authorizedAmount = BigNumber(payment.raw_amount.value)
@@ -518,9 +518,9 @@ export default class PaymentModuleService<
       data.amount = payment.amount as number
     }
 
-    const capturedAmount = payment.captures.reduce((acc, next) => {
-      const bn = new BigNumber(next.raw_amount.value)
-      return acc.plus(bn)
+    const capturedAmount = payment.captures.reduce((captureAmount, next) => {
+      const amountAsBigNumber = new BigNumber(next.raw_amount.value)
+      return captureAmount.plus(amountAsBigNumber)
     }, BigNumber(0))
     const refundAmount = BigNumber(data.amount)
 


packages/core-flows/src/payment/steps/refund-payment.ts
```diff
@@ -0,0 +1,23 @@
...
+  async (input: StepInput, { container }) => {
+    const paymentModule = container.resolve<IPaymentModuleService>(
+      ModuleRegistrationName.PAYMENT
+    )
+
+    const payment = await paymentModule.refundPayment(input)
+
+    return new StepResponse(payment)
+  }
+)
```
adrien2p, line 23: q: should we add a compensation in case this step is used elsewhere and the workflow fails?

packages/core-flows/src/payment/workflows/refund-payment.ts
```diff
@@ -0,0 +1,17 @@
...
+export const refundPaymentWorkflow = createWorkflow(
+  refundPaymentWorkflowId,
+  (
+    input: WorkflowData<{
+      payment_id: string
+      created_by?: string
+      amount?: number
+    }>
+  ) => {
+    const payment = refundPaymentStep(input)
```
adrien2p, line 14: nit: return refundPaymentStep(input)

packages/payment/src/services/payment-module.ts
```diff
@@ -507,17 +507,29 @@ export default class PaymentModuleService<
...
-    //     `Refund amount for payment: ${payment.id} cannot be greater than the amount captured on the payment.`
-    //   )
-    // }
+    if (!data.amount) {
+      data.amount = payment.amount as number
+    }
+
+    const capturedAmount = payment.captures.reduce((acc, next) => {
+      const bn = new BigNumber(next.raw_amount.value)
+      return acc.plus(bn)
```
adrien2p: todo: acc -> captureAmount, bn -> amountBn

packages/payment/src/services/payment-module.ts
```diff
@@ -507,17 +507,29 @@ export default class PaymentModuleService<
...
+    }, BigNumber(0))
+    const refundAmount = BigNumber(data.amount)
+
+    if (capturedAmount.lt(refundAmount)) {
+      throw new MedusaError(
+        MedusaError.Types.INVALID_DATA,
+        `You cannot refund more than what is captured on the payment.`
```
adrien2p, line 530:
```suggestion
        `You cannot refund more than what have been captured on the payment.`
```

wdyt?

adrien2p (approved):
LGTM

packages/core-flows/src/payment/steps/refund-payment.ts
olivermrbl: I don't see how we can compensate anything in this case. We can't revert a refund at the third-party provider.

One idea I had was to update the payment status to be `requires_action` or something similar to indicate that the current state of things is incorrect.

However, I'd like to punt this for a future PR. It will be more clear what to do, when we start playing around with it.

packages/core-flows/src/payment/workflows/refund-payment.ts
olivermrbl: question: How does this work in case someone appends a step to `refundPaymentStep`? This is why I chose to separate the two.

PR #6614, opened 2024-03-07
Feat(order): order changes

carlos-r-l-rodrigues:
This is a PR to keep them relatively small. Very likely changes, validations and other features will be added.

What:
  Basic methods to cancel, confirm or decline order changes
  Apply order changes to modify and create a new version of an order

Things related to calculation, Order and Item totals are not covered in this PR. Properties won't match with definition, etc.


srindom:
Is there anything specific you would like us to pay attention to? Overall, I think this looks good, but it's a bit tricky to follow given the size of the PR.

carlos-r-l-rodrigues:
> Is there anything specific you would like us to pay attention to? Overall, I think this looks good, but it's a bit tricky to follow given the size of the PR.

Nothing specific. When we have endpoints and flow I will ask for a more careful review of a few points.

olivermrbl (approved):
Looks good @carlos-r-l-rodrigues, don't have anything to comment on right now. Excited to start seeing the APIs shape up 💪 

I guess the wrapper-methods for RMA flows could end up looking something like this:
```
const return = await service.createReturn("order_1234", {
  created_by: "cx_1234",
  return_items: [{
    item_id: "li_1234",
    quantity: 1
  }],
})
```

And then under the hood, we call `createOrderChange`:
```
const orderChange = await service.createOrderChange({
  order_id: "order_1234",
  created_by: "cx_1234",
  actions: [
    {
      action: ChangeActionType.RETURN_ITEM,
      reference: "return",
      reference_id: "return_123",
      details: {
        reference_id: createdOrder.items[2].id,
        quantity: 1,
      },
    },
    },
  ],
})
```

carlos-r-l-rodrigues:
> I guess the wrapper-methods for RMA flows could end up looking something like this:

Yes, exactly this.
We will have multiple "bundled actions" for returns, swaps, etc and we just transform the method input to become actions.

PR #6616, opened 2024-03-07
feat(stock-location): Convert module to use mikro-orm

pKorsholm:
**What**
- Add new stock-location service ready for rc.2


packages/stock-location-next/src/migrations/Migration20240307161216.ts
```diff
@@ -0,0 +1,23 @@
...
+
+    this.addSql('alter table if exists "stock_location" add constraint "stock_location_address_id_foreign" foreign key ("address_id") references "stock_location_address" ("id") on update cascade on delete set null;');
+  }
+
+  async down(): Promise<void> {
+    this.addSql('alter table if exists "stock_location" drop constraint if exists "stock_location_address_id_foreign";');
+
+    this.addSql('drop table if exists "stock_location_address" cascade;');
+
+    this.addSql('drop table if exists "stock_location" cascade;');
```
adrien2p, line 20: todo: shouldn't we need to include the handling of the case where the table already exists now that we are migrating and the old one was used in v1?

packages/stock-location-next/src/services/__tests__/noop.ts
```diff
@@ -0,0 +1,5 @@
+describe("noop", function () {
```
adrien2p, line 1: todo: rm file

packages/stock-location-next/src/services/stock-location.ts
```diff
@@ -0,0 +1,194 @@
...
+  @InjectTransactionManager("baseRepository_")
+  async create_(
+    data: CreateStockLocationInput[],
+    @MedusaContext() context: Context = {}
+  ): Promise<TEntity[]> {
+    return await this.stockLocationService_.create(data, context)
+  }
+
+  update(
+    data: UpdateStockLocationInput & { id: string },
```
adrien2p: q: the id shouldn't be part of the DTO? also shouldn't we create proper DTO type like for the other modules?

packages/stock-location-next/src/services/stock-location.ts
```diff
@@ -0,0 +1,194 @@
...
+  @InjectTransactionManager("baseRepository_")
+  async update_(
+    data: (UpdateStockLocationInput & { id: string })[],
+    @MedusaContext() context: Context = {}
+  ): Promise<TEntity[]> {
+    return await this.stockLocationService_.update(data, context)
+  }
+
+  updateStockLocationAddress(
+    data: StockLocationAddressInput & { id: string },
```
adrien2p, line 163: q: same question for all those types

adrien2p (approved):
LGTM, few Question though

packages/stock-location-next/src/migrations/Migration20240307161216.ts
olivermrbl: @adrien2p, I think we are missing this in a lot of modules, so this could be tackled as part of cleaning up. Wdyt? 

I imagine we will have some discussing to do around the migration.

adrien2p: Yes sure 🙏 i know i ll tackle mine once i ve finished everything touching the entities ahah

packages/stock-location-next/src/services/__tests__/noop.ts
pKorsholm: can't remove this as the unit tests will break and the pipeline won't pass, we have the same test in the other modules to make unit-tests pass (fulfillment-module included)

packages/stock-location-next/src/migrations/Migration20240307161216.ts
pKorsholm: yes, but the schema here hasn't changed so this should be alright no?

packages/stock-location-next/src/services/__tests__/noop.ts
adrien2p: you can probably just add that `--passWithNoTests` in your command in the package.json, I think it is better than an empty test no?

carlos-r-l-rodrigues (approved):

olivermrbl:
Should we merge this one? 🙏

PR #6617, opened 2024-03-07
feat(medusa): added list price list products endpoint

riqwan:
what:

- adds an endpoint to list price list products

### DIFF
--- a/packages/medusa/src/subscribers/product.ts
+++ b/packages/medusa/src/subscribers/product.ts
@@ -1,25 +1,29 @@
 import { IEventBusService, ISearchService } from "@medusajs/types"
-import { defaultSearchIndexingProductRelations } from "@medusajs/utils"
+import { defaultSearchIndexingProductRelations, FlagRouter } from "@medusajs/utils"
 import { indexTypes } from "medusa-core-utils"
 import { isSearchEngineInstalledResolutionKey } from "../loaders/plugins"
 import ProductService from "../services/product"
 import ProductVariantService from "../services/product-variant"
+import ProductCategoryFeatureFlag from "../loaders/feature-flags/product-categories";
 
 type InjectedDependencies = {
   eventBusService: IEventBusService
   searchService: ISearchService
   productService: ProductService
+  featureFlagRouter: FlagRouter
 }
 
 class ProductSearchSubscriber {
   private readonly eventBusService_: IEventBusService
   private readonly searchService_: ISearchService
   private readonly productService_: ProductService
+  private readonly featureFlagRouter_: FlagRouter
 
   constructor(container: InjectedDependencies) {
     this.eventBusService_ = container.eventBusService
     this.searchService_ = container.searchService
     this.productService_ = container.productService
+    this.featureFlagRouter_ = container.featureFlagRouter
 
     /**
      * Do not subscribe to any event in case no search engine have been installed.
@@ -51,9 +55,17 @@ class ProductSearchSubscriber {
   }
 
   handleProductCreation = async (data) => {
+    const relations = [...defaultSearchIndexingProductRelations]
+    if (
+      this.featureFlagRouter_.isFeatureEnabled(ProductCategoryFeatureFlag.key)
+    ) {
+      relations.push("categories")
+    }
+
     const product = await this.productService_.retrieve(data.id, {
-      relations: defaultSearchIndexingProductRelations,
+      relations,
     })
+
     await this.searchService_.addDocuments(
       ProductService.IndexName,
       [product],
@@ -62,9 +74,17 @@ class ProductSearchSubscriber {
   }
 
   handleProductUpdate = async (data) => {
+    const relations = [...defaultSearchIndexingProductRelations]
+    if (
+      this.featureFlagRouter_.isFeatureEnabled(ProductCategoryFeatureFlag.key)
+    ) {
+      relations.push("categories")
+    }
+
     const product = await this.productService_.retrieve(data.id, {
-      relations: defaultSearchIndexingProductRelations,
+      relations,
     })
+
     await this.searchService_.addDocuments(
       ProductService.IndexName,
       [product],
@@ -77,9 +97,17 @@ class ProductSearchSubscriber {
   }
 
   handleProductVariantChange = async (data) => {
+    const relations = [...defaultSearchIndexingProductRelations]
+    if (
+      this.featureFlagRouter_.isFeatureEnabled(ProductCategoryFeatureFlag.key)
+    ) {
+      relations.push("categories")
+    }
+
     const product = await this.productService_.retrieve(data.product_id, {
-      relations: defaultSearchIndexingProductRelations,
+      relations,
     })
+
     await this.searchService_.addDocuments(
       ProductService.IndexName,
       [product],


sradevski (approved):

packages/medusa/src/api-v2/admin/price-lists/[id]/products/route.ts
```diff
@@ -0,0 +1,38 @@
...
+    priceList.prices?.map((price) => price.variant_id).filter(Boolean) || []
+
+  req.filterableFields = {
+    ...req.filterableFields,
+    variants: {
+      id: variantIds,
+    },
+  }
+
+  AdminGetProductsRoute(req, res)
```
Comment (adrien2p) [inline packages/medusa/src/api-v2/admin/price-lists/[id]/products/route.ts]: thought: I am a bit unsure of this approach, if any changes are applied to the get product route do we really want them to be applied here too? I think a simple duplication would be better in that scenario

packages/medusa/src/api-v2/admin/price-lists/types.ts
```diff
@@ -0,0 +1,25 @@
+import { PriceListStatus, PriceListType } from "@medusajs/types"
+
+export type AdminPriceListEndpointDTO = {
```
adrien2p: question: I believe that this type doesn't make real sense without consuming listPriceLists, if that the case then I would move that type in the same file as the function and maybe not calling it a EndPointDTO but maybe just `AdminSanitizedPriceListDTO`?

packages/medusa/src/api-v2/admin/products/route.ts
```diff
@@ -1,17 +1,17 @@
...
-import { CreateProductDTO } from "@medusajs/types"
-import { createProductsWorkflow } from "@medusajs/core-flows"
-import { remoteQueryObjectFromString } from "@medusajs/utils"
-
 export const GET = async (
   req: AuthenticatedMedusaRequest,
   res: MedusaResponse
 ) => {
-  const remoteQuery = req.scope.resolve("remoteQuery")
+  const remoteQuery = req.scope.resolve(LinkModuleUtils.REMOTE_QUERY)
```
adrien2p: question: Why don't we use `ContainerRegistrationKeys`, since when do we have that utils and since when the module SDK is aware of how the remote query is registered into an external container it does not know anything about? :) I believe this entire util can be removed entirely

adrien2p: I would recommend to remove this enum everywhere it is used in favor of the container registration keys which was created for that purpose.

It is also what we are using in the loader as example

```ts
container.register(
    ContainerRegistrationKeys.REMOTE_LINK,
    asValue(medusaApp.link)
  )
  container.register(
    ContainerRegistrationKeys.REMOTE_QUERY,
    asValue(medusaApp.query)
  )
  ```

File: packages/medusa/src/api-v2/admin/price-lists/[id]/products/route.ts (hunk 1 above)
Comment (adrien2p) [reply packages/medusa/src/api-v2/admin/price-lists/[id]/products/route.ts]: Worst case this become a middleware and we apply the product route on the end point then 😋

File: packages/medusa/src/api-v2/admin/price-lists/types.ts [hunk 4]
```diff
@@ -0,0 +1,25 @@
...
+    id: string
+    variant_id: string
+    currency_code?: string
+    amount?: number
+    min_quantity?: number
+    max_quantity?: number
+  }[]
+  customer_groups?: {
+    id: string
+  }
```
srindom: thought: the price list rule's conditions are open-ended right? would suggest that we don't parse it to anything special but just spit back the conditions as they are stored. This was necessary for v1 compat, but in v2 we can change the structure

packages/medusa/src/api-v2/admin/price-lists/[id]/products/route.ts
Comment (riqwan) [reply packages/medusa/src/api-v2/admin/price-lists/[id]/products/route.ts]: Yes, This one is basically a product endpoint 😄 this should really be a filter on the product endpoint but we also have this endpoint that does the same thing. 

If you feel strongly about it, I can extract the product remote query to a util and reuse it here. Wdyt?

packages/medusa/src/api-v2/admin/price-lists/types.ts
riqwan: This type is for the remote query collection result, which is also the result of the endpoint, which is why I named it like that. We can definitely name this something else and move it somewhere else. But my thought was that it would be useful for the frontend as well since it's the resultant type of the endpoint. 

Wdyt?

packages/medusa/src/api-v2/admin/products/route.ts
riqwan: Ah, didn't know there was one already, will change em

packages/medusa/src/api-v2/admin/price-lists/types.ts
riqwan: Sounds good, I'll remove this. That would have to mean that we can't expand on the customer groups anymore on this endpoint. The FE would have to make a separate query to fetch that info.

packages/medusa/src/api-v2/admin/price-lists/middlewares.ts
```diff
@@ -27,4 +35,14 @@ export const adminPriceListsRoutesMiddlewares: MiddlewareRoute[] = [
       ),
     ],
   },
+  {
+    method: ["GET"],
+    matcher: "/admin/price-lists/:id/products",
```
sradevski: Looking at this with a fresh mind now 😄 

Shouldn't this work a bit differently, where you call `/admin/products` with a filter on the price lists instead? It feels strange to me to ask for the products under a price list, as if the price lists own the product entity. I think the relationship is the other way around.

packages/medusa/src/api-v2/admin/price-lists/[id]/products/route.ts
Comment (sradevski) [reply packages/medusa/src/api-v2/admin/price-lists/[id]/products/route.ts]: And this somewhat confirms my comment on the route, if we need to do something like this it means there could be an issue with the idea of fetching products for a price list the way we are

sradevski (commented):
I approved yesterday but I missed some important aspects, leaving some comments now 😄

packages/medusa/src/api-v2/admin/price-lists/[id]/products/route.ts
Comment (riqwan) [reply packages/medusa/src/api-v2/admin/price-lists/[id]/products/route.ts]: In total agreement on that. Shall we remove this API route and stick to the filter on products then? We can make a note on the breaking changes that this route will no longer be supported on v2

packages/medusa/src/api-v2/admin/price-lists/types.ts
srindom: Yep that's all good 👍

packages/medusa/src/api-v2/admin/price-lists/[id]/products/route.ts
Comment (sradevski) [reply packages/medusa/src/api-v2/admin/price-lists/[id]/products/route.ts]: @riqwan yeah I think that would be much more intuitive, and as long as we document the breaking change it shouldn't be a big deal.

integration-tests/modules/__tests__/product/admin/products.spec.ts
```diff
@@ -0,0 +1,200 @@
...
+        appContainer = getContainer()
+        pricingModule = appContainer.resolve(ModuleRegistrationName.PRICING)
+        productModule = appContainer.resolve(ModuleRegistrationName.PRODUCT)
+      })
+
+      beforeEach(async () => {
+        await createAdminUser(dbConnection, adminHeaders, appContainer)
+      })
+
+      describe("GET /admin/products/:id", () => {
```
sradevski: I think having these tests in the module is fine, but I'll probably move them (or remove them if they are duplicate) once I finish with enabling the `api` tests, just FYI

packages/medusa/src/api-v2/admin/price-lists/queries/list-price-lists.ts
```diff
@@ -0,0 +1,84 @@
...
+  variables,
+}: {
+  container: MedusaContainer
+  fields: string[]
+  variables: Record<string, any>
+}): Promise<[AdminPriceListRemoteQueryDTO[], number]> {
+  const remoteQuery = container.resolve(ContainerRegistrationKeys.REMOTE_QUERY)
+  const queryObject = remoteQueryObjectFromString({
+    entryPoint: "price_list",
+    fields: adminPriceListRemoteQueryFields,
```
sradevski: Shouldn't we use the `fields` variable that is passed to the method here?

packages/medusa/src/api-v2/admin/price-lists/queries/list-price-lists.ts
```diff
@@ -0,0 +1,84 @@
...
+
+    return acc
+  }, {})
+}
+
+function buildPriceSetPrices(
+  priceSetMoneyAmounts: PriceSetMoneyAmountDTO[]
+): Record<string, any>[] {
+  return priceSetMoneyAmounts.map((priceSetMoneyAmount) => {
+    const productVariant = (priceSetMoneyAmount.price_set as any).variant
```
sradevski: Why do we need the cast to any here, and can it be avoided?

packages/medusa/src/api-v2/admin/price-lists/query-config.ts
```diff
@@ -1,53 +1,57 @@
...
-    "price_set_money_amounts.price_rules.rule_type.rule_attribute",
-  ],
-  customerGroupsFields: [
-    "price_list_rules.price_list_rule_values.value",
-    "price_list_rules.rule_type.rule_attribute",
-    "price_set_money_amounts.price_rules.value",
-    "price_set_money_amounts.price_rules.rule_type.rule_attribute",
-  ],
-}
+export const adminPriceListRemoteQueryFields = [
```
sradevski, line 5: What is the difference between this one and the `defaultAdminPriceListFields` (I mean conceptually)? Maybe this should be moved inside the `listPriceLists` file instead?

packages/medusa/src/api-v2/admin/products/route.ts
```diff
@@ -1,22 +1,58 @@
...
 
 export const GET = async (
   req: AuthenticatedMedusaRequest,
   res: MedusaResponse
 ) => {
-  const remoteQuery = req.scope.resolve("remoteQuery")
+  const remoteQuery = req.scope.resolve(ContainerRegistrationKeys.REMOTE_QUERY)
+  const filterableFields = { ...req.filterableFields }
+  const filterByPriceListIds =
+    (filterableFields.price_list_id as string[]) || []
```
sradevski: Are we sure this will be an array? Should we instead of `Array.isArray(...) ? pricelist : [pricelist]`?

packages/medusa/src/api-v2/admin/products/route.ts
```diff
@@ -1,22 +1,58 @@
...
+  const priceListVariantIds: string[] = []
+
+  // When filtering by price_list_id, we need use the remote query to get
+  // the variant IDs through the price list price sets.
+  if (filterByPriceListIds.length) {
+    const [priceLists] = await listPriceLists({
+      container: req.scope,
+      fields: ["prices.variant_id"],
+      variables: { filters: { id: filterByPriceListIds }, skip: 0, take: null },
```
sradevski, line 31: Not doing pagination might explode here, we need to think how cases like this should be handled. Can you just add a TODO comment maybe?

packages/medusa/src/api-v2/admin/products/route.ts
sradevski: I think this can get messy if for the filter we have something like `$and: [{price_list_id: [1]}, {...}]` (this will be a general problem if we want to operate on the filters in any way).  Let's add a TODO at the very least, and maybe start a discussion in eng-core

adrien2p: from an API point of view we do not allow such object shape with $and or $or, I believe it should be fine as it is defined by the validator

packages/medusa/src/api-v2/admin/price-lists/queries/list-price-lists.ts
riqwan: ah good catch 💪🏻

integration-tests/modules/__tests__/product/admin/products.spec.ts
riqwan: Are both v1 and v2 test suites ready to be run within api? I couldn't see V2 run in the pipeline, so i opted to still keep it here. But all good to move when its ready 👍🏻

packages/medusa/src/api-v2/admin/price-lists/query-config.ts
riqwan: `defaultAdminPriceListFields` is the contractual fields between the consumer and the API, whereas `adminPriceListRemoteQueryFields` is the contractual fields between the API and the module. 

With the pricing module, the internal models have a lot more complicated logic in there compared to what we had in v1. I'd like to keep this here because its still a query configuration between the api and the module, until we have a better solution and a more unified approach on how to tackle this.

packages/medusa/src/api-v2/admin/products/route.ts
riqwan: Yes, its defined in the query validator to always be an array inside validators.ts

packages/medusa/src/api-v2/admin/products/route.ts
riqwan: can you elaborate on why this might explode?

packages/medusa/src/api-v2/admin/price-lists/queries/list-price-lists.ts
riqwan: This is a deep remote link, we don't have any readily available types for this yet. 😄  We also need to figure out how to handle remote query types. I've temporarily added a partial type on the function and removed the cast. Will come back to this in the later PRs.

packages/medusa/src/api-v2/admin/products/route.ts
riqwan: I've linked the `filterableFields` type to the validator type and removed the cast. If anything in the API validators change in the future, this here will explode.

riqwan:
@adrien2p @sradevski can y'all take a final look?

integration-tests/modules/__tests__/product/admin/products.spec.ts
sradevski: @riqwan they are not ran in the pipeline yet, but they can be run manually while you are doing development. You can see my message in Slack from few days ago on how to do it 😄

packages/medusa/src/api-v2/admin/products/route.ts
sradevski: @riqwan 
I was just thinking the response could have a lot of entries, making this a very slow request. It is really more of a concern in general, maybe here it wouldn't be so bad.

sradevski (approved):

integration-tests/modules/__tests__/product/admin/products.spec.ts
riqwan: Yes, got that. I def want this running on the pipeline, thats why I kept it in modules.

packages/medusa/src/api-v2/admin/products/route.ts
riqwan: Agree on that. We def shouldn't be removing pagination on unscoped entities. This will only return as many price_lists as there are filter of price_list_ids here. The take:null here is more to not return a false result in case someone ever tried to scope price_list_id greater than the default pagination limit (which in itself is an edge case scenario).

PR #6618, opened 2024-03-07
chore: Adjusting the v2 product module to follow the v1 specs

sradevski:
In this PR:
1. I added upsert support for the product
2. I updated the create and update signatures to match the latest interface standards
3. Small changes to make the v1 and v2 APIs compatible (WIP)

### DIFF
--- a/integration-tests/api/__tests__/admin/product.js
+++ b/integration-tests/api/__tests__/admin/product.js
@@ -56,8 +56,8 @@ medusaIntegrationTestRunner({
       await createAdminUser(dbConnection, adminHeaders, container)
     })
 
-    describe("/admin/products", () => {
-      describe.only("GET /admin/products", () => {
+    describe.skip("/admin/products", () => {
+      describe("GET /admin/products", () => {
         beforeEach(async () => {
           await productSeeder(dbConnection)
           await simpleSalesChannelFactory(dbConnection, {
@@ -344,7 +344,10 @@ medusaIntegrationTestRunner({
         it("returns a list of deleted products with free text query", async () => {
           const response = await api
             .get(
-              "/admin/products?deleted_at[gt]=01-26-1990&q=test",
+              `/admin/products?deleted_at[${breaking(
+                () => "gt",
+                () => "$gt"
+              )}]=01-26-1990&q=test`,
               adminHeaders
             )
             .catch((err) => {
@@ -411,7 +414,13 @@ medusaIntegrationTestRunner({
 
         it("returns a list of deleted products", async () => {
           const response = await api
-            .get("/admin/products?deleted_at[gt]=01-26-1990", adminHeaders)
+            .get(
+              `/admin/products?deleted_at[${breaking(
+                () => "gt",
+                () => "$gt"
+              )}]=01-26-1990`,
+              adminHeaders
+            )
             .catch((err) => {
               console.log(err)
             })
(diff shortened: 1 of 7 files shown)


packages/medusa/src/types/common.ts
```diff
@@ -545,14 +545,24 @@ export class FindPaginationParams {
   @IsOptional()
   @Type(() => Number)
   limit?: number = 20
+
+  /**
+   * {@inheritDoc RequestQueryFields.order}
+   */
+  @IsString()
```
sradevski, line 552: Any concerns with adding this to the FindPaginationParams (and to the mixin below)?

packages/product/src/repositories/product.ts
```diff
@@ -136,7 +137,7 @@ export class ProductRepository extends DALUtils.mikroOrmBaseRepositoryFactory<Pr
         productData?.categories?.map((c) => c.id) || []
       )
 
-      tagIds = tagIds.concat(productData?.tags?.map((c) => c.id) || [])
+      tagIds = tagIds.concat(productData?.tags?.map((c: any) => c.id) || [])
```
sradevski, line 140: I'll be cleaning up the module itself later, so didn't want to go too deep here on fixing the typings.

integration-tests/api/__tests__/admin/product.js
```diff
@@ -56,7 +56,7 @@ medusaIntegrationTestRunner({
       await createAdminUser(dbConnection, adminHeaders, container)
     })
 
-    describe("/admin/products", () => {
+    describe.skip("/admin/products", () => {
```
sradevski, line 59: I wanted to have a relatively small PR so we can merge progressively.

Since it's WIP there are few tests that are failing due to me commenting out some test code. Once both APIs are compatible I'll remove this. I think I'll be able to do this tomorrow, so if we're ok with not having these tests running for a couple of days we can leave it as it is, or I can just wait to merge the PR once all of the changes are in place.

packages/product/src/services/product-module-service.ts
```diff
@@ -563,15 +578,100 @@ export default class ProductModuleService<
...
+    let updated: Product[] = []
+
+    if (forCreate.length) {
+      created = await this.create_(forCreate, sharedContext)
+    }
+    if (forUpdate.length) {
+      updated = await this.update_(forUpdate, sharedContext)
+    }
+
+    const result = [...created, ...updated]
```
srindom, line 615: thought(non-blocking): could these be parallelized?
```typescript
    if (forCreate.length) {
      created = this.create_(forCreate, sharedContext)
    }
    if (forUpdate.length) {
      updated = this.update_(forUpdate, sharedContext)
    }

    const result = await Promise.all([created, updated]).then(res => res.flat())
```

srindom (approved):
LGTM!

packages/product/src/services/product-module-service.ts
sradevski: @srindom I didn't do it with flat as I want to distingush between what updated and what was created for the events. It can be done without flattening though, but I figured in reality it is rare to do both create and update at the same time, so I figured no need to add that for now.

srindom: I see - all good 👍

srindom:
feel free to add automerge

PR #6620, opened 2024-03-07
feat(dashboard): manage discounts conditions

fPolic:
**What**
- forms for managing discount conditions

---

https://github.com/medusajs/medusa/assets/16856471/bb0b3fe8-fa60-479b-bc42-6db9d422bcdf


olivermrbl (approved):
This feels good 💪 

Will let @kasperkristensen have a look too, as I am not sure where we ended up regarding the edit conditions view.

kasperkristensen:
Looks good to me overall. Sorry for the late review, to make sure we can get this in asap,  I have opened a PR with some fixes to designs, types, and naming conventions: https://github.com/medusajs/medusa/pull/6707. The only other thing that seems to be missing is conditions in the create form. 

I tried to make the edit conditions modal a bit more responsive, but its still not quite there, but not sure how we can make it responsive with the current design 😅 Let's just keep it for now, and maybe its something we can have Ludvig take a look at after the release.

PR #6621, opened 2024-03-07
chore(integration-tests): Introduce V2 tests in Admin Payments API test suite

olivermrbl:


integration-tests/api/__tests__/admin/payment.js
```diff
@@ -83,69 +127,134 @@ medusaIntegrationTestRunner({
...
           {
-            amount: 5000,
-            reason: "return",
-            note: "Do not like it",
+            amount: 500,
+            ...breaking(
+              () => ({
+                // TODO: We should probably introduce this in V2 too
+                reason: "return",
+                note: "Do not like it",
```
olivermrbl, line 206: comment: Think we should add this to the payment module

olivermrbl:
@sradevski, I like the `breaking` util. Simple to use. Let me know if I followed the intended pattern.

integration-tests/api/__tests__/admin/payment.js
```diff
@@ -11,9 +13,47 @@ const adminHeaders = {
...
+    adminHeaders
+  )
+
+  // return payment
+  return response.data.payment_collection.payments[0]
+}
+
 medusaIntegrationTestRunner({
-  env: { MEDUSA_FF_PRODUCT_CATEGORIES: true },
+  // env: { MEDUSA_FF_MEDUSA_V2: true },
```
sradevski, line 52: When I run the tests I just do `export MEDUSA_FF_MEDUSA_V2=true` so it's a bit easier to switch between the two in the terminal where I run the tests, but this works as well 😄

sradevski (approved):
Nice one! I guess this is one of the few APIs that will have some breaking changes

olivermrbl:
Yup, this one will have quite a few breaking changes, because we've introduced the notion of a `Capture` as standalone entity.

integration-tests/api/__tests__/admin/payment.js
olivermrbl: I see, just thought this would be an easy switch to flip, but I like your approach too.
 I'll keep it as is for now.

PR #6630, opened 2024-03-08
feat(fulfillment): Soft deletes

adrien2p:
**What**
- Ensure soft delete works properly according to the soft delete configuration and validate all relation of the entire data model
- Add is_enabled to the providers in order to manage new providers to enabled or disabled
- include joiner config update

FIXES CORE-1853
FIXES CORE-1830
FIXES CORE-1719

### DIFF
--- a/packages/fulfillment/src/models/fulfillment-provider.ts
+++ b/packages/fulfillment/src/models/fulfillment-provider.ts
@@ -1,14 +1,11 @@
 import { generateEntityId } from "@medusajs/utils"
 import {
   BeforeCreate,
-  Collection,
   Entity,
-  OneToMany,
   OnInit,
   PrimaryKey,
   Property,
 } from "@mikro-orm/core"
-import ShippingOption from "./shipping-option"
 
 @Entity()
 export default class FulfillmentProvider {
@@ -17,12 +14,6 @@ export default class FulfillmentProvider {
 
   @Property({ columnType: "boolean", defaultRaw: "true" })
   is_enabled: boolean = true
-
-  @OneToMany(
-    () => ShippingOption,
-    (shippingOption) => shippingOption.fulfillment_provider
-  )
-  shipping_options = new Collection<ShippingOption>(this)
 
   @BeforeCreate()
   onCreate() {
--- a/packages/fulfillment/integration-tests/__tests__/fulfillment-module-service/index.spec.ts
+++ b/packages/fulfillment/integration-tests/__tests__/fulfillment-module-service/index.spec.ts
@@ -229,7 +229,7 @@ moduleIntegrationTestRunner({
           expect(provider.is_enabled).toEqual(isEnabled)
         }
 
-        await shutdown()
+        await shutdown().catch(() => void 0)
       })
 
       it("should soft delete and restore the data respecting the configured cascade", async () => {


packages/fulfillment/src/loaders/providers.ts
```diff
@@ -65,22 +66,34 @@ async function syncDatabaseProviders({ container }) {
...
+    )
+    const providersToEnabled = providerIdentifiers.filter((id) =>
+      loadedProvidersMap.has(id)
+    )
+    const providersToDisable = providers.filter(
+      (p) => !providerIdentifiers.includes(p.id)
+    )
 
-    await providerService.create(providersToCreate)
+    await promiseAll([
```
olivermrbl: suggestion(non-blocking): Separate in variable assignments for readability

packages/fulfillment/src/migrations/Migration20240311134415_InitialSetupMigration.ts
```diff
@@ -1,12 +1,12 @@
...
     this.addSql('CREATE INDEX IF NOT EXISTS "IDX_fulfillment_address_deleted_at" ON "fulfillment_address" (deleted_at) WHERE deleted_at IS NOT NULL;');
 
-    this.addSql('create table if not exists "fulfillment_provider" ("id" text not null, "metadata" jsonb null, "created_at" timestamptz not null default now(), "updated_at" timestamptz not null default now(), "deleted_at" timestamptz null, constraint "fulfillment_provider_pkey" primary key ("id"));');
+    this.addSql('create table if not exists "fulfillment_provider" ("id" text not null, "is_enabled" boolean not null default true, "metadata" jsonb null, "created_at" timestamptz not null default now(), "updated_at" timestamptz not null default now(), "deleted_at" timestamptz null, constraint "fulfillment_provider_pkey" primary key ("id"));');
```
olivermrbl: question: Should we remove the other fields while we're at it?

Fulfillment providers should only have `id` and `is_enabled`, similar to payment providers.

olivermrbl: If so, you'll need to revisit indexes too

packages/fulfillment/src/joiner-config.ts
```diff
@@ -17,9 +32,24 @@ Object.entries(LinkableKeys).forEach(([key, value]) => {
 
 export const entityNameToLinkableKeysMap: MapToConfig = entityLinkableKeysMap
 
+const aliases = [
```
olivermrbl: todo: Replace with inline alias definition for readability (and consistency with other modules). This is an unnecessary "optimization"

olivermrbl (approved):
LGTM, couple of comments

packages/fulfillment/src/joiner-config.ts
olivermrbl: Alternatively, we apply it to all other modules.

adrien2p: Yeah I was trying to propose something before applying everywhere but I ve deleted it to do it like the other modules. The main idea behind was that since the format is known and can be done automatically, it was easier to read for bigger modules by just provided the necessary information and build the rest based on it. but let see that another time

olivermrbl: It would be amazing if this was done automatically, as in we don't need either of these approaches.

olivermrbl: I haven't faced a different format yet, so it looks like something that can be generalized

adrien2p: I think it can, I can do a test

adrien2p: but I believe it can also wait later, but I have an idea to do it based on the entities just like for the rest. lets get this merged and think about this improvement later wdyt?

packages/fulfillment/src/migrations/Migration20240311134415_InitialSetupMigration.ts
adrien2p: what do you mean re visit indexes?

packages/fulfillment/src/models/fulfillment-provider.ts
```diff
@@ -1,67 +1,29 @@
...
-  @Property({ columnType: "jsonb", nullable: true })
-  metadata: Record<string, unknown> | null = null
+  @Property({ columnType: "boolean", defaultRaw: "true" })
+  is_enabled: boolean = true
 
   @OneToMany(
     () => ShippingOption,
     (shippingOption) => shippingOption.fulfillment_provider
   )
   shipping_options = new Collection<ShippingOption>(this)
```
olivermrbl: question: Should we remove this one too?

adrien2p: 👍 done

adrien2p:
should we slap an auto merge @olivermrbl if you have no more things to adjust :)

olivermrbl:
Let's do it

PR #6631, opened 2024-03-08
bug: fix constraint on customer table to allow soft-deletes

madsleejensen:
Currently there is a constraint on the `customer` table to ensure uniqueness based on `email` `has_account` columns. I think the intention is to allow soft deletion of customers (because it derives from that `SoftDeletableEntity`), and the current constraint prevents us from adding a customer after it has been soft-deleted. 

This PR will replace the current constraint with a compound unique partial index that fixes this. 

A few notes regarding the implementation:
- I updated the typeorm decorators to match the new index, even though i dont think it is used for anything.
- If the migration would ever need to be reverted it will attempt to remove rows that to satisfy the previous constraint (but cases where the deleted customer has active foreign keys wont work and the "revert" will fail, i looked in  similar cases in the current migrations and looks like it is not really handled, also seems quite hard to handle because of some tables has circular foreign keys)

### DIFF
--- a/.changeset/eighty-eagles-reply.md
+++ b/.changeset/eighty-eagles-reply.md
@@ -1,0 +1,5 @@
+---
+"@medusajs/medusa": patch
+---
+
+bug: fix constraint on customer table to allow soft-deletes


sradevski (commented):
Looks good to me, but I'll let others more familiar with this part of the codebase to approve

carlos-r-l-rodrigues (approved):
LGTM

adrien2p (approved):

olivermrbl:
@madsleejensen, thanks. Will be part of a release next week :)

PR #6633, opened 2024-03-08
feat(medusa,dahsboard): Initial work on pricing domain

kasperkristensen:
**What**
- Initial work on pricing domain.
- List page
- Details page
- Partial edit form
- Fixes admin/price-lists/:id/products endpoint to allow filtering products by multiple ids.

**Note**
Pushing this now as the current design need a bit of tweaking so we display all relevant data to users. Future PR will include completed Edit moda, Create modal, Edit prices modal and Add prices modal.

### DIFF
--- a/packages/admin-next/dashboard/public/locales/en-US/translation.json
+++ b/packages/admin-next/dashboard/public/locales/en-US/translation.json
@@ -288,9 +288,9 @@
     "editConditions": "Edit conditions",
     "conditionsHint": "Create conditions to apply on the discount",
     "isTemplateDiscount": "Is this a template discount?",
-    "percentageDescription" : "Discount applied in %",
-    "fixedDescription" : "Amount discount",
-    "shippingDescription" : "Override delivery amount",
+    "percentageDescription": "Discount applied in %",
+    "fixedDescription": "Amount discount",
+    "shippingDescription": "Override delivery amount",
     "selectRegionFirst": "Select region first",
     "templateHint": "Template discounts allow you to define a set of rules that can be used across a group of discounts. This is useful in campaigns that should generate unique codes for each user, but where the rules for all unique codes should be the same.",
     "conditions": {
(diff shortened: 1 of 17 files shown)


packages/admin-next/dashboard/src/routes/pricing/pricing-list/components/pricing-list-table/pricing-table-actions.tsx
```diff
@@ -0,0 +1,12 @@
...
+import { ActionMenu } from "../../../../../components/common/action-menu"
+
+type PricingTableActionsProps = {
+  priceList: PriceList
+}
+
+export const PricingTableActions = ({
+  priceList,
+}: PricingTableActionsProps) => {
+  return <ActionMenu groups={[]} />
```
olivermrbl: question: I assume this will be added in a follow-up PR?

packages/medusa/src/services/price-list.ts
```diff
@@ -378,6 +378,9 @@ class PriceListService extends TransactionBaseService {
       const productVariantRepo = manager.withRepository(
         this.productVariantRepo_
       )
+
+      console.log(selector, config, priceListId, requiresPriceList)
```
olivermrbl: todo: rm
```suggestion
```

olivermrbl (approved):
Nice work @kasperkristensen 💪 

I noticed Customer Groups management was not listed in the PR description. This will also need to be added. And the configuration card too. You are likely already on top of this but thought I'd highlight

Also, it seems the table view is not 1:1 with the design. Columns "Description" and "Type" are missing. Is this intentional?

kasperkristensen:
@olivermrbl Type missing was not intentional I've added it 👍 I have omitted description on purpose, as I don't believe it's useful in a table view, most cases you are just going to see `Used for our...`.

And yes everything will be added in a follow up PR. The configuration card was one of the things I made Ludvig add (the design tweaks mentioned in the PR description), he just got around to it before this was reviewed.

fPolic (approved):
LGTM

PR #6635, opened 2024-03-08
feat: Confirm inventory in create cart workflow

olivermrbl:
Add confirm variant inventory step to:
- Create cart workflow
- Add to cart workflow
- Update line item in cart workflow

### DIFF
--- a/integration-tests/modules/__tests__/cart/store/cart.workflows.spec.ts
+++ b/integration-tests/modules/__tests__/cart/store/cart.workflows.spec.ts
@@ -155,7 +155,7 @@ medusaIntegrationTestRunner({
                 sales_channel_id: salesChannel.id,
               },
               [Modules.STOCK_LOCATION]: {
-                location_id: location.id,
+                stock_location_id: location.id,
               },
             },
             {
@@ -280,7 +280,7 @@ medusaIntegrationTestRunner({
                 sales_channel_id: salesChannel.id,
               },
               [Modules.STOCK_LOCATION]: {
-                location_id: location.id,
+                stock_location_id: location.id,
               },
             },
             {
@@ -476,7 +476,7 @@ medusaIntegrationTestRunner({
                 sales_channel_id: salesChannel.id,
               },
               [Modules.STOCK_LOCATION]: {
-                location_id: location.id,
+                stock_location_id: location.id,
               },
             },
             {
@@ -568,7 +568,7 @@ medusaIntegrationTestRunner({
                 sales_channel_id: salesChannel.id,
               },
               [Modules.STOCK_LOCATION]: {
-                location_id: location.id,
+                stock_location_id: location.id,
               },
             },
             {
@@ -690,7 +690,7 @@ medusaIntegrationTestRunner({
                 sales_channel_id: salesChannel.id,
               },
               [Modules.STOCK_LOCATION]: {
-                location_id: location.id,
+                stock_location_id: location.id,
               },
             },
             {
@@ -817,7 +817,7 @@ medusaIntegrationTestRunner({
                   sales_channel_id: salesChannel.id,
                 },
                 [Modules.STOCK_LOCATION]: {
-                  location_id: location.id,
+                  stock_location_id: location.id,
                 },
               },
               {


olivermrbl:
Update: Converting to draft. Will add it to the rest of the cart workflows right away. The PR will be kept small still.

packages/core-flows/src/definition/cart/workflows/update-line-item-in-cart.ts
```diff
@@ -32,6 +32,58 @@ export const updateLineItemInCartWorkflow = createWorkflow(
...
+      variables: { id: input.cart.sales_channel_id },
+    })
+
+    const productVariantInventoryItems = useRemoteQueryStep({
+      entry_point: "product_variant_inventory_items",
+      fields: ["variant_id", "inventory_item_id", "required_quantity"],
+      variables: { variant_id: variantIds },
+    }).config({ name: "inventory-items" })
+
+    const confirmInventoryInput = transform(
```
sradevski, line 57: It seems this confirm inventory transformation is repeated in multiple steps, so we can probably extract it. Now I don't know what convention we want to follow with workflows and where we put this sort of "helper" code, but for the time being it can be part of the cart workflow folder?

olivermrbl: Agree, I am amid refactoring it. Will convert to draft in the mean time.

adrien2p: it could be a common step that we reuse in different places

packages/core-flows/src/definition/cart/utils/prepare-confirm-inventory-input.ts
```diff
@@ -0,0 +1,61 @@
...
+}: ConfirmInventoryPreparationInput) => {
+  if (!productVariantInventoryItems.length) {
+    return []
+  }
+
+  const variantsMap = new Map<string, ProductVariantDTO>(
+    variants.map((v) => [v.id, v])
+  )
+
+  const itemsToConfirm: {
```
sradevski: NIT: Define the type outside the function (eg. above it)

packages/core-flows/src/definition/cart/utils/prepare-confirm-inventory-input.ts
```diff
@@ -0,0 +1,61 @@
...
+    quantity: number
+    location_ids: string[]
+  }[] = []
+
+  items.forEach((item) => {
+    const variantInventoryItem = productVariantInventoryItems.find(
+      (i) => i.variant_id === item.variant_id
+    )
+
+    const variant = variantsMap.get(item.variant_id!)
```
sradevski: Let's move the check of (!variantInventoryItem) before we get the variant from the map

sradevski (approved):

packages/core-flows/src/definition/cart/utils/prepare-confirm-inventory-input.ts
```diff
@@ -0,0 +1,61 @@
...
+import { MedusaError } from "medusa-core-utils"
+
+interface ConfirmInventoryPreparationInput {
+  productVariantInventoryItems: {
+    variant_id: string
+    inventory_item_id: string
+    required_quantity: number
+  }[]
+  items: { variant_id?: string; quantity: number }[]
+  variants: ProductVariantDTO[]
```
adrien2p: todo: it seems that we dont need the DTO but only the id and manage inventory, if thats correct it would be better to only requires those instead of requiring the entire dto. This should always be the approach to take IMO

packages/core-flows/src/definition/cart/utils/prepare-confirm-inventory-input.ts
```diff
@@ -0,0 +1,61 @@
...
+  variants,
+}: ConfirmInventoryPreparationInput) => {
+  if (!productVariantInventoryItems.length) {
+    return []
+  }
+
+  const variantsMap = new Map<string, ProductVariantDTO>(
+    variants.map((v) => [v.id, v])
+  )
+
```
adrien2p, line 35: suggestion: you could create a map of productVariantInventoryItems in order to not have to do a find for each item later. following what you did with the variant map

PR #6636, opened 2024-03-08
fix: make v2 with modules run

srindom:


packages/medusa/src/loaders/index.ts
```diff
@@ -91,12 +93,10 @@ async function loadMedusaV2({ directory, expressApp }) {
     next()
   })
 
-  const featureFlagRouter = featureFlagsLoader(configModule, Logger)
-
   const pgConnection = await pgConnectionLoader({ container, configModule })
 
   container.register({
-    [ContainerRegistrationKeys.MANAGER]: asValue(dataSource.manager),
+    // [ContainerRegistrationKeys.MANAGER]: asValue(dataSource.manager),
```
srindom: Is this safe to do? cc: @adrien2p

packages/medusa/src/commands/migrate.js
```diff
@@ -70,15 +73,32 @@ const main = async function ({ directory }) {
...
-    await runIsolatedModulesMigration(configModule)
-
-    await runLinkMigrations(directory)
+    if (featureFlagRouter.isFeatureEnabled(MedusaV2Flag.key)) {
+      const container = createMedusaContainer()
+      const pgConnection = await pgConnectionLoader({ configModule, container })
+      container.register({
+        [ContainerRegistrationKeys.CONFIG_MODULE]: asValue(configModule),
+        [ContainerRegistrationKeys.LOGGER]: asValue(Logger),
+        [ContainerRegistrationKeys.MANAGER]: asValue(dataSource.manager),
```
srindom: can we get rid of this? cc: @adrien2p?

packages/medusa/src/loaders/index.ts
adrien2p: Yes i believe so, modules does not rely on the manager

packages/medusa/src/commands/migrate.js
adrien2p: You can just keep the logger

packages/medusa/package.json
```diff
@@ -70,7 +70,7 @@
     "core-js": "^3.6.5",
     "cors": "^2.8.5",
     "cross-spawn": "^7.0.3",
-    "dotenv": "^16.0.3",
+    "dotenv": "^16.4.5",
```
adrien2p, line 73: In the future with node 21.7 it will not be needed anymore

packages/utils/src/common/deep-copy.ts
```diff
@@ -10,10 +10,6 @@ import { isObject } from "./is-object"
 export function deepCopy<T extends Record<any, any> = Record<any, any>>(
   obj: T | T[]
 ): T | T[] {
-  if (typeof structuredClone != "undefined") {
```
adrien2p, line 13: Nit could you update the tsdoc please

adrien2p (approved):

adrien2p (approved):

carlos-r-l-rodrigues (approved):

PR #6639, opened 2024-03-09
fix: Medusa V2 loader

srindom:
**What**
- Runs loadMedusaV2 when v2 flag is enabled.
- V2 loader loads only project-level framework features. E.g., project APIs, subscribers, jobs, and workflows.
- No plugin support yet.


packages/modules-sdk/src/loaders/register-modules.ts
```diff
@@ -25,7 +26,14 @@ export const registerMedusaModule = (
   const modDefinition = definition ?? ModulesDefinition[moduleKey]
 
   if (modDefinition === undefined) {
-    throw new Error(`Module: ${moduleKey} is not defined.`)
+    console.log(moduleKey, " is a custom module will register it as such")
+    console.log("moduleDeclaration", moduleDeclaration)
```
carlos-r-l-rodrigues: Maybe we should warn here that no definition was found, and the custom module is being loaded as Internal/Shared.
At least it doesn't hide the fact that you can setup a custom module with all the available options, like `isQueryable`, `dependencies`, etc...

carlos-r-l-rodrigues (approved):

carlos-r-l-rodrigues:
Not related to this PR, but just an idea about plugins, like payment/shipping providers. We can probably categorize the plugin and auto register them into the modules they belong. 
Basically "merging" the plugins config with the modules' option.

packages/modules-sdk/src/loaders/register-modules.ts
srindom: Will actually remove the log here for now and then widen the loading capabilities and how they are communicated through the logs later.

PR #6640, opened 2024-03-09
feat(dashboard): Return Reasons domain

kasperkristensen:
**What**
- Adds Return Reason domain.

### DIFF
--- a/.changeset/loud-adults-report.md
+++ b/.changeset/loud-adults-report.md
@@ -1,0 +1,6 @@
+---
+"@medusajs/client-types": patch
+"@medusajs/medusa": patch
+---
+
+fix(medusa): Adds updated_at query param to list-reservations. Fixes OAS for list-inventory-items.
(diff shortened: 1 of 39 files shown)


olivermrbl (approved):
LGTM!

One thought: I wonder if we should add a tooltip to Value? I don't think it is obvious what to use it for and how.

Wdyt?

CleanShot 2024-03-10 at 11 06 55@2x

olivermrbl:
And/or a link to our docs? But this is maybe a separate discussion to have in the future; that is to generally include documentation in our admin.

kasperkristensen:
@olivermrbl I'll throw a tooltip on it, not sure if it makes sense to link to docs here, as the description for the value is in our developer docs, which I don't think we should link to in admin (other than in the developer domains like API keys and Workflows). 

Might be worth considering if the `value` field should be removed in 2.0, as it's a bit weird to make admins fill out a identifier, when we could just let that be the ID. I guess it might be in relation to having nested return reasons, but we have never really supported this (at least not from the admin dashboard).

PR #6647, opened 2024-03-11
feat(): Update transformer middleware and API

adrien2p:
**What**
Update all transform middleware to support the new API
- deprecate `defaultRelations`
- deprecate `allowedRelations`
- Add `defaults` and `allowed` in replacement for `defaultFields` and `allowedFields` respectively
- in the `defaults` it is possible to specify a field such as `*variants` in order to be recognized as a relation only without specifying any property
- add support for `remoteQueryConfig` assigned to req like we have for `listConfig` and `retrieveConfig`
- add support to override `allowed|allowedFields` if a previous middleware have set it up on the req.allowed
- The api now accepts `fields` as the only accepted fields to manage the requested props and relations, the `expand` property have been deprecated. New supported symbols have been added in complement of the fields
  - `+` (e.g `/store/products?fields=+description`) to specify that description should be added as part of the returned data among the other defined fields
  - `-` (e.g `/store/products?fields=-description`) to specify that description should be removed as part of the returned data
  - `*` (e.g `/store/products?fields=*variants`) to specify that the variants relations should be added as part of the returned data among the other defined fields without having to specify which property of the variants should be returned. In the `defaults` config of the transform middleware it is also possible to use this symbol
  - In the case no symbol is provided, it will replace the default fields and mean that only the specified fields must be returned

About the allowed validation, all fields in the `defaults` configuration must be present in the `allowed` configuration. 
In case the `defaults` contains full relation selection (e.g `*product.variants`) it should be present in the `allowed` as `product.variants`. In case in the `defaults` you add `product.variants.id`, it will be allowed if the `allowed` configuration includes either `product.variants.id` as full match or `product.variants` as it means that we allow all properties from `product.variants`

Also, support for `*` selection on the remote query/joiner have been added

**Note**
All v2 end points refactoring can be done separately


integration-tests/api/__tests__/admin/order/order.js
```diff
@@ -1613,7 +1613,7 @@ describe("/admin/orders", () => {
       )
     })
 
-    it("lists orders with specific fields and relations", async () => {
+    it.only("lists orders with specific fields and relations", async () => {
```
carlos-r-l-rodrigues: rm `.only`

integration-tests/modules/__tests__/cart/store/carts.spec.ts
```diff
@@ -738,7 +738,7 @@ medusaIntegrationTestRunner({
       })
 
       describe("GET /store/carts", () => {
-        it("should get cart", async () => {
+        it.only("should get cart", async () => {
```
carlos-r-l-rodrigues: rm 'only'

adrien2p:
I am currently trying to rework the allowedProperties used for the cleanResponseData util

carlos-r-l-rodrigues (approved):

packages/medusa/src/utils/get-query-config.ts
```diff
@@ -19,93 +20,116 @@ export function pickByConfig<TModel extends BaseEntity>(
...
+      customFields.some(
+        (field) => !(field.startsWith("-") || field.startsWith("+"))
+      )
+    if (shouldReplaceDefaultFields) {
+      allFields = new Set(customFields.map((f) => f.replace(/^[+-]/, "")))
+    } else {
+      customFields.forEach((field) => {
+        if (field.startsWith("+")) {
+          allFields.add(field.replace(/^\+/, ""))
+        } else if (field.startsWith("-")) {
```
sradevski, line 66: I guess replace will only replace the first occurrence, but we need to be careful not to mess up a variable name called `some-variable` for example. Just something that can easily happen 😄

sradevski (approved):
Just removing the `.only` on the tests, the rest looks good 👍 🥳

packages/medusa/src/utils/get-query-config.ts
carlos-r-l-rodrigues: I guess here is the case to just replace one, and the one after the beginning of the string `ˆ`. Looks good.
Not even the regexp is required, but this isn't important 😆

sradevski: yeah exactly, but I can easily see someone coming in later on and changing the code thereby breaking the logic :D Just having a test for it would be nice.

adrien2p: @carlos-r-l-rodrigues @sradevski you ll have to re review, I ve done some changes in the mean time and the test includes this check

packages/medusa/src/api-v2/admin/products/route.ts
```diff
@@ -54,11 +54,9 @@ export const GET = async (
     entryPoint: "product",
     variables: {
       filters: filterableFields,
-      order: req.listConfig.order,
-      skip: req.listConfig.skip,
-      take: req.listConfig.take,
+      ...req.remoteQueryConfig.variables,
```
sradevski: Reading this again, `variables` is quite ambiguous I'd say, can we maybe make it more specific, eg. `pagination`?

packages/medusa/src/api/middlewares/transform-query.ts
```diff
@@ -78,12 +92,22 @@ export function transformStoreQuery<
...
-        req.includes ?? {},
-        queryConfig
-      )
       attachListOrRetrieveConfig<TEntity>(req, queryConfig)
+      /**
+       * TODO: shouldn't this correspond to returnable fields instead of allowed fields? also it is used by the cleanResponseData util and is supposed to
+       * only return the fields that are allowed to be queried either from the default of from what the user requested.
+       * The validation occurred above and the fields are already validated and should be used to filter the response.
+       */
+      const queryConfigRes = req.retrieveConfig ?? req.listConfig
```
sradevski: Should we exctract this logic into a function? I see it occuring twice

sradevski (approved):

packages/medusa/src/api-v2/admin/products/route.ts
adrien2p: yes we could, it includes, skip, take but also ordering, I ve named it variables to match the object API and recognize it. also, I am fine with pagination but order would be out of context in that case no?

packages/medusa/src/api/middlewares/transform-query.ts
adrien2p: I ve done a larger cleanup as now both middlewares are the same, only the API change

packages/medusa/src/api-v2/admin/products/route.ts
sradevski: @adrien2p well ordering is related to pagination IMO, so it would be fine as such. Anyway that's my opinion, happy to hear what others think, but variables is too generic

adrien2p: I am fine with pagination, I ve renamed and updated the types. you ll get notified when I ll push

sradevski (approved):

olivermrbl:
@adrien2p, nice work!

Feel free to merge whenever you find fitting.

adrien2p:
@olivermrbl I missed the changeset here, sorry

PR #6650, opened 2024-03-11
feat: Implement missing methods in product module and make more tests pass

sradevski:
The 2 bigger remaining tasks are:
1. handling prices for variants
2. Handling options (breaking change)

After that all tests should pass on both v1 and v2


packages/product/src/joiner-config.ts
```diff
@@ -49,7 +49,7 @@ export const joinerConfig: ModuleJoinerConfig = {
       },
     },
     {
-      name: ["variant", "variants"],
+      name: ["producyt_variant", "product_variants", "variant", "variants"],
```
olivermrbl:
```suggestion
      name: ["product_variant", "product_variants", "variant", "variants"],
```

packages/medusa/src/api-v2/admin/products/[id]/variants/route.ts
```diff
@@ -56,5 +58,5 @@ export const POST = async (
     throw errors[0].error
   }
 
-  res.status(200).json({ product_variant: result[0] })
+  res.status(200).json({ variant: result[0] })
```
Comment (olivermrbl) [inline packages/medusa/src/api-v2/admin/products/[id]/variants/route.ts:61]: question: As per our recent discussion, should we return product here?

olivermrbl (approved):
Solid work 💪

packages/medusa/src/api-v2/admin/products/[id]/variants/route.ts
Comment (sradevski) [reply packages/medusa/src/api-v2/admin/products/[id]/variants/route.ts:61]: @olivermrbl yes there is a test for it that is skipped, once enabled I'll update this, I'll review all routes at that time to make sure we are consistent

packages/product/src/joiner-config.ts
sradevski: gosh, nice catch 😄

PR #6651, opened 2024-03-11
chore(order): big number calculations

carlos-r-l-rodrigues:


packages/utils/src/totals/transform-properties-to-bignumber.ts
```diff
@@ -0,0 +1,53 @@
+import { BigNumber } from "./big-number"
```
olivermrbl, line 1: suggestion: Few tests would be neat

packages/order/src/utils/calculate-order-change.ts
```diff
@@ -46,9 +50,9 @@ export class OrderChangeProcessing {
     this.transactions = JSON.parse(JSON.stringify(transactions ?? []))
     this.actions = JSON.parse(JSON.stringify(actions ?? []))
 
-    const transactionTotal = transactions.reduce((acc, transaction) => {
-      return acc + transaction.amount
-    }, 0)
+    const transactionTotal = MathBN.add(...transactions.map((tr) => tr.amount))
```
olivermrbl, line 53: suggestion: Add `sum`to math lib.

packages/order/src/services/__tests__/util/actions/exchanges.ts
```diff
@@ -96,18 +96,20 @@ describe("Order Exchange - Actions", function () {
...
+      currentOrderTotal: "312.5",
+      temporaryDifference: "62.5",
       futureDifference: 0,
-      futureTemporaryDifference: 0,
-      pendingDifference: 312.5,
-      differenceSum: 42.5,
+      futureTemporaryDifference: "0",
+      pendingDifference: "312.5",
+      differenceSum: "42.5",
     })
```
olivermrbl, line 109: question: This is only for assertion purposes right? Users won't receive these as strings

adrien2p (approved):
LGTM 🔥 just questions from oli to look at

packages/utils/src/totals/math.ts
```diff
@@ -0,0 +1,104 @@
...
+  static add(...nums: BNInput[]): BigNumberJS {
+    let sum = new BigNumberJS(0)
+    for (const num of nums) {
+      const n = MathBN.convert(num)
+      sum = sum.plus(n)
+    }
+    return sum
+  }
+
+  static minus(...nums: BNInput[]): BigNumberJS {
```
riqwan: nit: subtract or sub?

```suggestion
  static subtract(...nums: BNInput[]): BigNumberJS {
```

packages/order/src/services/__tests__/util/actions/exchanges.ts
```diff
@@ -96,18 +96,20 @@ describe("Order Exchange - Actions", function () {
       actions: actions,
     })
 
-    expect(changes.summary).toEqual({
-      transactionTotal: 0,
+    const sumToJSON = JSON.parse(JSON.stringify(changes.summary))
+    expect(sumToJSON).toEqual({
+      transactionTotal: "0",
```
riqwan, line 101: q: is the non stringified version a big number? should the modules always return a bignumber?

riqwan (approved):
lgtm!

packages/order/src/utils/calculate-order-change.ts
carlos-r-l-rodrigues: Can it be just an alias to `add`?

packages/order/src/services/__tests__/util/actions/exchanges.ts
carlos-r-l-rodrigues: yeah, this is only used internally, so everything that does math operations will handle bignumbers which ends up being "string".
The module responses will always be numeric for these fields.

packages/order/src/utils/calculate-order-change.ts
olivermrbl: Yup

packages/order/src/services/__tests__/util/actions/exchanges.ts
carlos-r-l-rodrigues: It's because that wasn't affected by any calculation internally. So it has the type BigNumber which transforms it to numeric when toJSON is called..
When things are affected by calculations, they become a BigNumberJS, that will be a string when toJSON is called.
But these details are not relevant to the module's consumer.

packages/utils/src/totals/transform-properties-to-bignumber.ts
carlos-r-l-rodrigues: sure. will add some.

packages/utils/src/totals/math.ts
carlos-r-l-rodrigues: will go with `sub`

PR #6661, opened 2024-03-11
feat: Add shipping method to cart workflow

olivermrbl:
Add `addShippingMethodtoCart` workflow

~~PR is awaiting fulfillment module migration update. In its current state it is not backward compatible, which is causing tests to fail.~~


packages/core-flows/src/definition/cart/workflows/add-shipping-method-to-cart.ts
```diff
@@ -0,0 +1,75 @@
...
+  createWorkflow,
+  transform,
+} from "@medusajs/workflows-sdk"
+import { useRemoteQueryStep } from "../../../common/steps/use-remote-query"
+import { addShippingMethodToCartStep } from "../steps"
+import { getShippingOptionPriceSetsStep } from "../steps/get-shipping-option-price-sets"
+import { refreshCartPromotionsStep } from "../steps/refresh-cart-promotions"
+
+interface AddShippingMethodToCartWorkflowInput {
+  cart: CartDTO
```
adrien2p: suggestion: I believe that the input should only expect what they need to work, here it looks like the id and currency code would be enough so I would say that it is not necessary to enforce the rest of the data to be provided wdyt? It should be always the approach for all workflows, always asking what they need not all the data.

packages/core-flows/src/definition/cart/steps/get-shipping-option-price-sets.ts
```diff
@@ -0,0 +1,82 @@
...
+    if (notFound.length) {
+      throw new MedusaError(
+        MedusaError.Types.INVALID_DATA,
+        `Shipping options with IDs ${notFound.join(", ")} do not have a price`
+      )
+    }
+
+    const calculatedPriceSets = await pricingModuleService.calculatePrices(
+      { id: priceSetIds },
+      { context: data.context as Record<string, string | number> }
```
riqwan: suggestion: you can use the `PricingContext` type here

packages/core-flows/src/definition/cart/workflows/add-shipping-method-to-cart.ts
```diff
@@ -0,0 +1,75 @@
...
+            cart_id: data.input.cart.id,
+          }
+        })
+
+        return options
+      }
+    )
+
+    addShippingMethodToCartStep({ shipping_methods: shippingMethodInput })
+    refreshCartPromotionsStep({ id: input.cart.id })
```
riqwan: q: this would also require shipping method tax lines, no? There is a step for that already `updateTaxLinesStep`

sradevski (approved):

PR #6667, opened 2024-03-11
feat(core-flows,medusa,types): add automatic-taxes to region + generate tax lines endpoint

riqwan:
what:

- endpoint to generate tax lines
- update workflows to force calculate tax lines with a flag
- added automatic_taxes to region

### DIFF
--- a/integration-tests/modules/__tests__/fixtures/fulfillment/index.ts
+++ b/integration-tests/modules/__tests__/fixtures/fulfillment/index.ts
@@ -1,0 +1,138 @@
+import {
+  CreateFulfillmentDTO,
+  CreateShippingOptionDTO,
+  IFulfillmentModuleService,
+} from "@medusajs/types"
+
+export function generateCreateFulfillmentData(
+  data: Partial<CreateFulfillmentDTO> & {
+    provider_id: string
+    shipping_option_id: string
+  }
+) {
+  const randomString = Math.random().toString(36).substring(7)
+
+  return {
+    location_id: "test-location",
+    packed_at: null,
+    shipped_at: null,
+    delivered_at: null,
+    canceled_at: null,
+    data: null,
+    provider_id: data.provider_id,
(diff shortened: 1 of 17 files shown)


packages/core-flows/src/definition/cart/steps/get-item-tax-lines.ts
```diff
@@ -9,21 +9,36 @@ import {
...
 ): TaxCalculationContext | null {
   const address = cart.shipping_address
+  const shouldCalculateTax = forceTaxCalculation || cart.region?.automatic_taxes
+
+  if (!shouldCalculateTax) {
+    return null
+  }
+
+  if (!address || !address?.country_code) {
+    if (forceTaxCalculation) {
```
carlos-r-l-rodrigues:
```suggestion
  if (forceTaxCalculation && !address?.country_code) {
```

olivermrbl (approved):
Code-wise, LGTM!

packages/core-flows/src/definition/cart/steps/get-item-tax-lines.ts
riqwan: @carlos-r-l-rodrigues we want this to skip if there is no country code and only throw if its forcing a calculation

olivermrbl: ~~I think carlos is just suggesting a simplification of the check no?~~

My bad, I see what you mean.

riqwan: doing this breaks the flow of `forceTaxCalculation=false` to return with null if address.country_code is not present. We could remove the `!address` or break it into 2 if conditions

This is the expected flow:
```
if (forceTaxCalculation && !countryCode) {
  throw error
}

if (!countryCode) {
  return null
}
```

sradevski (approved):

PR #6672, opened 2024-03-12
feat(fulfillment): Migration backward compatibility

adrien2p:
**What**
- Update migration for backward compatibility. This does not take into account data migration and table cleanup (@olivermrbl do we have a general tasks for the modules on that subject)?
- Rename fulfillment provider id to provider id
- add integration tests to ensure the migration backward compatibility
- add new module type for the options to be used in the medusa config for example

FIXES CORE-1724

### DIFF
--- a/integration-tests/modules/__tests__/fulfillment/index.spec.ts
+++ b/integration-tests/modules/__tests__/fulfillment/index.spec.ts
@@ -34,7 +34,7 @@ medusaIntegrationTestRunner({
             relations: [
               "service_zones.geo_zones",
               "service_zones.shipping_options.shipping_profile",
-              "service_zones.shipping_options.fulfillment_provider",
+              "service_zones.shipping_options.provider",
               "service_zones.shipping_options.type",
               "service_zones.shipping_options.rules",
               "service_zones.shipping_options.fulfillments.labels",


adrien2p:
WIP until I add some integration tests

olivermrbl (approved):

adrien2p:
@olivermrbl you ll have to re review once I push my changes 👍

integration-tests/modules/package.json
```diff
@@ -5,7 +5,7 @@
   "license": "MIT",
   "private": true,
   "scripts": {
-    "test:integration": "node --expose-gc ./../../node_modules/.bin/jest --ci --silent=false -i --detectOpenHandles --logHeapUsage --forceExit",
+    "test:integration": "node --expose-gc ./../../node_modules/.bin/jest --ci --silent=false -i --detectOpenHandles --logHeapUsage --forceExit -- __tests__/fulfillment/index.spec.ts",
```
olivermrbl: todo: remove file path

packages/fulfillment/src/types/index.ts
```diff
@@ -7,3 +12,22 @@ export type InitializeModuleInjectableDependencies = {
 
 export const FulfillmentIdentifiersRegistrationName =
   "fulfillment_providers_identifier"
+
+export type FulfillmentModuleServiceInitializeOptions =
```
olivermrbl: suggestion:
```suggestion
export type FulfillmentModuleOptions =
```

olivermrbl (approved):

integration-tests/modules/package.json
adrien2p: You probably have an old version 👍

PR #6674, opened 2024-03-12
feat(fulfillment): Separate list and context rules validation

adrien2p:
**What**

- Add method to validate fulfillment option from the provider
- Separate list/list and count from context rules validation and add listShippingOptionsForContext

FIXES CORE-1861


packages/fulfillment/src/services/fulfillment-module-service.ts
```diff
@@ -1271,6 +1273,47 @@ export default class FulfillmentModuleService<
...
+      data
+    )
+  }
+
+  @InjectManager('baseRepository_')
+  async isShippingOptionValidForContext(
+    shippingOptionId: string,
+    context: Record<string, unknown> = {},
+    @MedusaContext() sharedContext: Context = {}
+  ) {
```
olivermrbl: question: Would it make sense to call it `listShippingOptionsForContext` and have the same behavior that `listShippingOptions` had before?

olivermrbl: Or `listValidShippingOptions`

adrien2p: I prefer the first one, i can do that

packages/fulfillment/src/services/fulfillment-module-service.ts
```diff
@@ -189,42 +186,55 @@ export default class FulfillmentModuleService<
...
   }
 
+  @InjectManager("baseRepository_")
+  // @ts-ignore
+  async listAndCountShippingOptions(
+    filters: FilterableShippingOptionProps = {},
+    config: FindConfig<ShippingOptionDTO> = {},
+    @MedusaContext() sharedContext: Context = {}
+  ): Promise<[ShippingOptionDTO[], number]> {
+    const { filters: normalizedFilters, config: normalizedConfig } =
```
sradevski: Why do we need the custom implementation here? I know there is a good reason, but what is it? 😄

sradevski (approved):

packages/fulfillment/src/services/fulfillment-module-service.ts
olivermrbl: With the separate method `listShippingOptionsForContext` maybe we don't need this?

I would expect `listShippingOptions` to handle filters the same way other list methods do.

adrien2p: Ahah there is no good reason, it is just because of the shortcut fulfillment_set_id and so on which means that we need to re map them.
But i ll wait until we settle on the approach and then cleanup.

packages/fulfillment/src/services/fulfillment-module-service.ts
adrien2p: I am wondering if it makes sense to do that in truth. Why two methods one to list and one to list plus filter on context when we had one method to do this and if you dont pass the context it just performs a classic list. What do you prefer? But i think that if we have a separate method for rules validation then it shouldn't list otherwise it comes back to what we had before no?

olivermrbl:
I might have caused some confusion. After giving this some thought, here's what I think we should do.

I would suggest we have the following methods:

`listShippingOptions`

`listAndCountShippingOptions`

`validateShippingOption`

None of the methods should expose a different filter object from what we usually do i.e. I think we should remove the normalization.

The first two methods would behave like any other `list*` method, with the exception of the context parameter. If the context is passed, it performs an additional validation to ensure only the shipping options applicable for the given context are returned. As you suggest @adrien2p. 

The latter would be a convenient method, wrapping the `listShippingOptions`. 

```ts
async validateShippingOption(option, context) {
  const [shippingOption] = await this.listShippingOptions({ id: option.id }, context)

  if (shippingOptions) {
    return true
  }

  return false
}
```

Whether or not we add the third method I don't have a strong opinion about. Just think it's nice to call a dedicated method for the purpose of validating the applicableness of options.

For example, I will need to add a `refreshShipping` step in all cart workflows to ensure the currently created shipping methods are still applicable after the update. If not, they should be deleted.

adrien2p:
@olivermrbl I ve done the changes as per the discussion, let me know what you think

packages/fulfillment/src/services/fulfillment-provider.ts
```diff
@@ -78,8 +78,8 @@ export default class FulfillmentProviderService extends ModulesSdkUtils.internal
     return await provider.validateFulfillmentData(optionData, data, context)
   }
 
-  async validateOption(data: any) {
-    const provider = this.retrieveProviderRegistration(data.provider_id)
+  async validateOption(providerId: string, data: Record<string, unknown>) {
```
sradevski, line 81: Do we even need this in the custom service, or should we just put this logic in `fulfillment-module-service` instead?

sradevski (approved):

packages/fulfillment/src/services/fulfillment-provider.ts
adrien2p: This is a service that communicate with the third party provider installed on the module, everything link to third party should be separated from the module service itself IMO, wdyt?

srindom:
Sorry for chiming in late - I have one additional method that I'd like the module to have:

```
getShippingOptionsForContext(fulfillmentSetIds, fulfillmentContext)
```

This is a convenience method on top of `listShippingOptions` + it verifies that options are valid for the context. This will be the primary function used when figuring out shipping options applicable to a cart. 

Also want to encourage that `listShippingOptions` doesn't know the context but instead behaves exactly like all other `list*` methods throughout our modules. 

The motivation for this is:

1. It simplifies the work the module consumer has to do to get shipping options. (e.g., you don't need to remember the exact data structure of service_zones -> geo_zones).
2. It allows us to validate and throw if the context passed is insufficient for finding shipping options.
3. It provides a single API that we can innovate behind (e.g., we can add new features in the context or similar without requiring people to update their `list` params.
4. We keep behavior of all `list*` calls the same whic

adrien2p:
@srindom yes it makes perfect sense. What do you think of the current changes? Also, I am waiting for oli to be available to discuss something around the failing test. then it should be good to be merged if you approve the changes

packages/fulfillment/src/services/fulfillment-module-service.ts
```diff
@@ -1375,4 +1380,181 @@ export default class FulfillmentModuleService<
...
+   *        province_code: "AB"
+   *      },
+   *      {
+   *        type: "country",
+   *        country_code: "SE"
+   *      }
+   *    ]
+   *  }
+   */
+  private static buildGeoZoneConstraintsFromAddress(
```
olivermrbl, line 1501: love it!

olivermrbl (approved):
great work @adrien2p , this lgtm

adrien2p:
@srindom ill wait for your approval before putting the automerge label

PR #6675, opened 2024-03-12
feat(core-flows,medusa,utils): added price list prices upsert and delete workflows + endpoints

riqwan:
what:

- adds price list prices upsert endpoint
- adds price list prices remove endpoint

TODO:

This will have to change to the new batch endpoints, but will come in a separate PR.


packages/core-flows/src/price-list/steps/remove-price-list-prices.ts
```diff
@@ -0,0 +1,38 @@
...
+    const { removedPriceListPrices = [] } = data || {}
+
+    const pricingModule = container.resolve<IPricingModuleService>(
+      ModuleRegistrationName.PRICING
+    )
+
+    if (removedPriceListPrices.length) {
+      // TODO: This needs to be implemented
+      // pricingModule.restorePrices(removedPriceListPrices.map((p) => p.id))
+    }
```
olivermrbl: nit(non-blocking):
```suggestion
    return new StepResponse(null, psmas.map(psma => psma.id))
  },
  async (data, { container }) => {
    if (!data) {
      return
    }

    const pricingModule = container.resolve<IPricingModuleService>(
      ModuleRegistrationName.PRICING
    )

    // TODO: This needs to be implemented
    // pricingModule.restorePrices(data)
```

packages/core-flows/src/price-list/workflows/update-price-lists.ts
```diff
@@ -1,13 +1,38 @@
...
+
+    updatePriceListsStep(
+      transform({ input }, (data) => {
+        return data.input.price_lists_data.map((priceListsData) => {
+          const { prices, ...rest } = priceListsData
+
+          return rest
+        })
+      })
+    )
```
olivermrbl: comment: I am not too fond of this way of using the transforms, makes it hard to read. Please separate in "variable assignments"

packages/core-flows/src/price-list/steps/upsert-price-list-prices.ts
```diff
@@ -0,0 +1,195 @@
...
+} from "@medusajs/utils"
+import { StepResponse, createStep } from "@medusajs/workflows-sdk"
+
+type WorkflowStepInput = {
+  price_list_id: string
+  prices?: (UpdatePriceListPriceWorkflowDTO | CreatePriceListPriceWorkflowDTO)[]
+}[]
+
+export const upsertPriceListPricesStepId = "upsert-price-list-prices"
+export const upsertPriceListPricesStep = createStep(
```
olivermrbl, line 22: question: Can we break this into separate steps for the workflow? 

We are not only upserting price list prices here, so the name is misleading. Also, it makes it less reusable. 

Wdyt/

packages/core-flows/src/price-list/workflows/update-price-lists.ts
olivermrbl: Pretty sure you don't need them here actually. You are passing workflow inputs. Should not need a transformer.

packages/core-flows/src/price-list/steps/upsert-price-list-prices.ts
riqwan: I did that initially 😄, it felt like an overkill since you have to run the same set of steps for both create and update. I'll give it another whirl. Maybe if i extract the variant links and price list checks I would be able to reuse for both.

riqwan: @olivermrbl made some changes, let me know wyt?

packages/core-flows/src/price-list/steps/upsert-price-list-prices.ts
```diff
@@ -0,0 +1,129 @@
...
+      dataBeforePriceUpdate.push({
+        price_list_id: priceListId,
+        prices: buildPriceSetPricesForModule(psmas),
+      })
+    }
+
+    // TODO: `addPriceListPrices` will return a list of price lists
+    // This should be reworked to return prices instead, as we need to
+    // do a revert incase this step fails
+    const [createdPriceListPrices, _] = await Promise.all([
```
olivermrbl:
```suggestion
    const [createdPriceListPrices, _] = await promiseAll([
```

packages/core-flows/src/price-list/steps/upsert-price-list-prices.ts
olivermrbl: Nice!

packages/core-flows/src/price-list/steps/upsert-price-list-prices.ts
```diff
@@ -0,0 +1,129 @@
...
+      ModuleRegistrationName.PRICING
+    )
+
+    if (createdPriceListPrices.length) {
+      pricingModule.removePrices(createdPriceListPrices.map((p) => p.id))
+    }
+
+    if (updatedPriceListPrices.length) {
+      pricingModule.updatePriceListPrices(updatedPriceListPrices)
+    }
```
olivermrbl:
```suggestion
    if (createdPriceListPrices.length) {
      await pricingModule.removePrices(createdPriceListPrices.map((p) => p.id))
    }

    if (updatedPriceListPrices.length) {
      await pricingModule.updatePriceListPrices(updatedPriceListPrices)
    }
```

packages/core-flows/src/price-list/workflows/upsert-price-list-prices.ts
```diff
@@ -0,0 +1,27 @@
...
+}
+
+export const upsertPriceListPricesWorkflowId = "upsert-price-list-prices"
+export const upsertPriceListPricesWorkflow = createWorkflow(
+  upsertPriceListPricesWorkflowId,
+  (input: WorkflowData<WorkflowInput>): WorkflowData<void> => {
+    validatePriceListsStep(input.price_lists_data)
+    const variantPriceMap = validateVariantPriceLinksStep(
+      input.price_lists_data
+    )
```
olivermrbl: suggestion: `parallelize` the validation

packages/pricing/src/migrations/Migration20240312163523.ts
```diff
@@ -0,0 +1,18 @@
+import { Migration } from "@mikro-orm/migrations"
```
olivermrbl: question: Can we move this to the master migration?

olivermrbl (commented):
Nice, left a couple of comments

packages/pricing/src/migrations/Migration20240312163523.ts
riqwan: yeah, this was supposed to be removed. my bad!

PR #6678, opened 2024-03-12
feat(medusa,pricing,types): added get endpoints for pricing rule types

riqwan:
what:

- adds list endpoint for rule types
- adds get endpoint for rule types


packages/pricing/src/migrations/Migration20240312163523.ts
```diff
@@ -0,0 +1,18 @@
+import { Migration } from "@mikro-orm/migrations"
```
olivermrbl: todo: Include in "master" migration

packages/pricing/src/migrations/Migration20240312163523.ts
```diff
@@ -0,0 +1,18 @@
...
+
+export class Migration20240312163523 extends Migration {
+  async up(): Promise<void> {
```
olivermrbl: question: Is the new index missing here?

olivermrbl (approved):
Nice work, minor comments 💪

packages/pricing/src/migrations/Migration20240312163523.ts
riqwan: no, its already present in the first migration. It just got missed in the model.

packages/pricing/src/migrations/Migration20240312163523.ts
riqwan: done

PR #6685, opened 2024-03-13
feat: Update the product options model and refactor the product module

sradevski:
The changes in this PR are:
1. Change how product options are created and stored. The relationship changed from
`options --> option values <-- variants`
to
`options --> option values --> variant options <-- variants`

Now we can enforce non-duplicate option values, easier creation and updates of options, and more.

2. Refactors the product module. The product module did a lot of things in a non-ideal approach, and this is a step towards a more consistent usage of the base repository and methods exposed by a module. There is still work left to improve the module, but a large chunk of the changes are included in this PR

Things to do as a follow-up:
1. Remove many-to-many relationships  if an empty list is passed in the base repository.
2. Improve the typings of the module
3. Further cleanup and improvements (there are few questions that I need answered before I can improve the API)


packages/product/src/migrations/InitialSetup20240312114652.ts
```diff
@@ -0,0 +1,184 @@
+import { Migration } from "@mikro-orm/migrations"
+
+export class InitialSetup20240312114652 extends Migration {
```
sradevski: The migration does the bare minimum to make it compatible in the tests with the existing tables, there is more work left to be done here

packages/core-flows/src/handlers/product/update-product-variants.ts
```diff
@@ -14,16 +14,13 @@ export async function updateProductVariants({
...
     container.resolve(ModulesDefinition[Modules.PRODUCT].registrationName)
 
-  for (const [
-    product_id,
-    variantsUpdateData = [],
-  ] of productVariantsMap) {
+  for (const [product_id, variantsUpdateData = []] of productVariantsMap) {
     updateVariantsData.push(
-      ...(variantsUpdateData as unknown as UpdateProductVariantOnlyDTO[]).map(
+      ...(variantsUpdateData as unknown as UpdateProductVariantDTO[]).map(
```
riqwan: sugg: you can remove the cast as productVariantsMap is already typed correctly.

packages/product/package.json
```diff
@@ -29,7 +29,7 @@
     "prepublishOnly": "cross-env NODE_ENV=production tsc --build && tsc-alias -p tsconfig.json",
     "build": "rimraf dist && tsc --build && tsc-alias -p tsconfig.json",
     "test": "jest --runInBand --bail --forceExit -- src/**/__tests__/**/*.ts",
-    "test:integration": "jest --forceExit -- integration-tests/**/__tests__/**/*.ts",
+    "test:integration": "jest --forceExit --runInBand -- integration-tests/**/__tests__/**/products.spec.ts",
```
riqwan:
```suggestion
    "test:integration": "jest --forceExit --runInBand -- integration-tests/**/__tests__/**/*.ts",
```

packages/product/src/models/product-variant-option.ts
```diff
@@ -0,0 +1,82 @@
...
+    nullable: true,
+  })
+  option_value!: ProductOptionValue
+
+  @Property({ columnType: "text", nullable: true })
+  variant_id!: string
+
+  @ManyToOne(() => ProductVariant, {
+    fieldName: "variant_id",
+    nullable: true,
```
riqwan, line 48: q: should these (variant_id and option_value_id) be nullable? For a pivot entity, these should be required right?

packages/product/src/models/product-variant.ts
```diff
@@ -122,6 +122,14 @@ class ProductVariant {
   @Property({ columnType: "text", nullable: true })
   product_id!: string
 
+  @ManyToOne(() => Product, {
+    onDelete: "cascade",
+    index: "IDX_product_variant_product_id",
+    fieldName: "product_id",
+    nullable: true,
```
riqwan, line 129: q: variants won't exist without a product, should this be flipped to false?

Product module was our first mikroorm experiment, we might've made it nullable because we were still trying to figure out how this all works. 😄

packages/product/src/models/product.ts
```diff
@@ -129,6 +130,7 @@ class Product {
     pivotTable: "product_tags",
     index: "IDX_product_tag_id",
     cascade: ["soft-remove"] as any,
+    // TODO: Do we really want to remove tags if the product is deleted?
```
riqwan, line 133: 👍🏻 only the pivot entities, not the tag itself.

packages/product/src/services/product-category.ts
```diff
@@ -125,7 +124,7 @@ export default class ProductCategoryService<
   @InjectTransactionManager("productCategoryRepository_")
   async update(
     id: string,
-    data: ProductCategoryServiceTypes.UpdateProductCategoryDTO,
+    data: ProductTypes.UpdateProductCategoryDTO,
```
riqwan, line 127: nothing to do here for now, but just fyi you can remove most of this and rely on the new service base class. I think on the list and retrieve methods are extended to include the "descendant tree"

packages/product/src/services/product-module-service.ts
```diff
@@ -194,61 +200,31 @@ export default class ProductModuleService<
...
-    for (const productVariantData of data) {
-      productVariantData.options = productVariantData.options?.map((option) => {
-        const productOption = productOptionsMap.get(option.option_id!)
-
-        return {
-          option: productOption?.id,
-          value: option.value,
+    const productVariantsWithOptions = await promiseAll(
+      data.map(async (variant) => {
+        if (!Object.keys(variant.options ?? {}).length) {
```
riqwan: sugg: there is a util you could use here `isPresent(variant.options)`

riqwan (commented):
💪🏻 will do the rest in a bit.

packages/product/src/models/product-variant-option.ts
sradevski: @riqwan There is annoying issue where you might only pass one or the other and MikroORM will complain about it. I think there is a quite a bit of work left to do the modeling properly, for now I just do whatever works to get the module passing tests :)

packages/product/src/models/product-variant.ts
sradevski: @riqwan same as above, I'll tackle this separately (I have a todo for myself already)

packages/product/src/models/product.ts
sradevski: Yeah that's what I would assume as well, but I'll revisit this separately

packages/product/src/services/product-category.ts
sradevski: Oh definitely, but I wouldn't dare touching this just yet, there are already too many changes 😄

kasperkristensen (commented):
LGTM, but since I am not in the know with a lot module work I'll leave the approve to someone else. 

thought: Not needed in this PR, but I think it would be great to introduce batch endpoints for the product options - batch create, batch update, batch delete.

packages/product/src/services/product-module-service.ts
```diff
@@ -816,378 +946,273 @@ export default class ProductModuleService<
...
+    options: ProductOption[]
+  ):
+    | ProductTypes.CreateProductVariantDTO[]
+    | ProductTypes.UpdateProductVariantDTO[] {
+    const variantsWithOptions = variants.map((variant: any) => {
+      const variantOptions = Object.entries(variant.options ?? {}).map(
+        ([key, val]) => {
+          const option = options.find((o) => o.title === key)
+          const optionValue = option?.values?.find(
+            (v: any) => (v.value.value ?? v.value) === val
```
riqwan, line 1185: should this just be:
```suggestion
            (v: any) => v.value === val
```

packages/types/src/product/common.ts
```diff
@@ -1101,9 +1100,9 @@ export interface CreateProductVariantDTO {
    */
   width?: number
   /**
-   * The product variant options to create and associate with the product variant.
+   * The product variant options to associate with the product variant.
    */
-  options?: CreateProductVariantOptionDTO[]
+  options?: Record<string, string>
```
riqwan, line 1105: options can have multiple values if i remember correctly.
```suggestion
  options?: Record<string, string[]>
```

packages/product/src/services/product-module-service.ts
```diff
@@ -816,378 +946,273 @@ export default class ProductModuleService<
...
+          productData.categories = await this.productCategoryService_.list(
+            { id: productData.categories.map((c) => c.id) },
+            { take: null },
+            sharedContext
+          )
+        }
 
-        if (productData.is_giftcard) {
-          productData.discountable = false
+        // TODO: Maybe we also want to delete the options and variants that are not in the list?
```
riqwan, line 1064: Yes, I think an options update on product update should ideally be replacing what we currently have and then have atomic operations like addProductOptions and removeProductOptions to append or remove

riqwan (approved):
There is a lot going on in this PR and its hard to follow. 😄 Slapping a ✅ for you to move forward.

sradevski:
> There is a lot going on in this PR and its hard to follow. 😄 Slapping a ✅ for you to move forward.

@riqwan definitely, thanks for the review so far though!

Unfortunately there was no easy way to break it down as everything was a bit intertwined, but basically as long as the tests pass and the interface is right we should be fine. There are a lot of improvements left but we'll do those afterwards

packages/types/src/product/common.ts
sradevski: @riqwan you can only have one option of a type, for example `color: green`. You can't have it both green and blue

riqwan: Ah sorry, I thought this was on the Product.

packages/product/src/services/product-module-service.ts
```diff
@@ -271,107 +248,127 @@ export default class ProductModuleService<
 
   @InjectTransactionManager("baseRepository_")
   protected async updateVariants_(
-    data: ProductTypes.UpdateProductVariantOnlyDTO[],
+    data: ProductTypes.UpdateProductVariantDTO[],
     @MedusaContext() sharedContext: Context = {}
   ): Promise<TProductVariant[]> {
+    // Validation step
```
olivermrbl, line 254: suggestion: Look into improving performance of this method by replacing arrays with maps. We are looping over the same data a couple of times, I believe – improvements could cascade to `diffVariants`.

Note, I usually don't care much about this because often times we would be optimizing for an ultra-edge case, but variants is actually one of those places where it makes sense to be meticulous about achieving good performance, as we've seen many cases with dozens and even hundreds of variants on products.

This can be punted for a follow-up itereation, so this doesn't hang any longer.

packages/product/src/services/product-module-service.ts
```diff
@@ -450,8 +456,28 @@ export default class ProductModuleService<
...
+          : {}),
+      } as ProductTypes.UpdateProductOptionDTO
+    })
+
+    if (normalizedInput.some((option) => !option.id)) {
+      throw new MedusaError(
+        MedusaError.Types.INVALID_DATA,
+        "Tried to update options without specifying an ID"
+      )
+    }
```
olivermrbl, line 477: question: Could this be moved to before the normalization?

packages/product/src/services/product-module-service.ts
```diff
@@ -460,6 +486,92 @@ export default class ProductModuleService<
...
+    if (toUpdate.length) {
+      const existingOptions = await this.productOptionService_.list(
+        { id: toUpdate.map((o) => o.id) },
+        { take: null },
+        sharedContext
+      )
+
+      const updateOptions = await promiseAll(
+        toUpdate.map(async (optionToUpdate) => {
+          const dbOption = existingOptions.find(
```
olivermrbl, line 516: suggestion(non-blocking): Convert to map for lookups in constant time

packages/product/src/services/product-module-service.ts
```diff
@@ -815,378 +946,273 @@ export default class ProductModuleService<
...
-    @MedusaContext() sharedContext: Context = {}
-  ): Promise<void> {
-    await this.productCategoryService_.delete(categoryId, sharedContext)
-
-    await this.eventBusModuleService_?.emit<ProductCategoryEventData>(
-      ProductCategoryEvents.CATEGORY_DELETED,
-      { id: categoryId }
-    )
-  }
+const uniqBy = <T>(arr: T[], key: keyof T) => {
```
olivermrbl, line 1212: suggestion: Move to utils

olivermrbl (approved):
Strong work @sradevski. Lot of stuff in this PR, I think it will make sense to get this in now and incorporate improvements/changes as we build out the APIs

packages/product/src/services/product-module-service.ts
sradevski: I think there are a lot of opportunities for performance improvements, but most of them would be at the DB level as that's where most of the time is spent. Using maps is also nice of course, but usually not where bottlenecks are. 

Anyway this module needs a lot more work, for now I'll focus on functionality first.

packages/product/src/services/product-module-service.ts
sradevski: No actually the option value can either be an object or a primitive, depending if it was normalized before passing it to the function. Something to improve though

packages/product/src/services/product-module-service.ts
sradevski: Yeah we probably can, I'll do more cleanup when we have more focus on the quality of the modules

PR #6687, opened 2024-03-13
Feat(core-flows, inventory-next): List and create inventory items

pKorsholm:
**What**
- Add inventory endpoints
  - list-inventory
  - create inventory-level
  - create-inventory item
- Fix integration tests to run with "modern" modules marked as legacy


olivermrbl:
As discussed, can we split this into three separate PRs? 🙏

PR #6692, opened 2024-03-13
feat(core-flows,medusa,types,utils): add rules to promotion endpoints + workflow

riqwan:
what:

- adds endpoint + workflow to add rules to a promotion
  - promotion rules
  - application target rules
  - application buy rules
- updates promotion module:
  - add*Rules to return created PromotionRules instead of Promotion
  - remove*Rules return void similar to other remove methods

### DIFF
--- a/packages/core-flows/src/promotion/steps/add-rules-to-promotions.ts
+++ b/packages/core-flows/src/promotion/steps/add-rules-to-promotions.ts
@@ -1,24 +1,15 @@
 import { ModuleRegistrationName } from "@medusajs/modules-sdk"
 import {
-  CreatePromotionRuleDTO,
+  AddPromotionRulesWorkflowDTO,
   IPromotionModuleService,
-  PromotionRuleTypes,
 } from "@medusajs/types"
 import { RuleType } from "@medusajs/utils"
 import { StepResponse, createStep } from "@medusajs/workflows-sdk"
 
-type WorkflowStepInput = {
-  rule_type: PromotionRuleTypes
-  data: {
-    id: string
-    rules: CreatePromotionRuleDTO[]
-  }
-}
-
 export const addRulesToPromotionsStepId = "add-rules-to-promotions"
 export const addRulesToPromotionsStep = createStep(
   addRulesToPromotionsStepId,
-  async (input: WorkflowStepInput, { container }) => {
+  async (input: AddPromotionRulesWorkflowDTO, { container }) => {
     const { data, rule_type: ruleType } = input
 
     const promotionModule = container.resolve<IPromotionModuleService>(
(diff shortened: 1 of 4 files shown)


packages/core-flows/src/promotion/workflows/add-rules-to-promotions.ts
```diff
@@ -0,0 +1,19 @@
+import { CreatePromotionRuleDTO, PromotionRuleTypes } from "@medusajs/types"
+import { WorkflowData, createWorkflow } from "@medusajs/workflows-sdk"
+import { addRulesToPromotionsStep } from "../steps"
+
+type WorkflowInput = {
```
olivermrbl: todo: Move to `@medusajs/types`

packages/core-flows/src/promotion/steps/add-rules-to-promotions.ts
```diff
@@ -0,0 +1,85 @@
...
+      (await promotionModule.removePromotionBuyRules(
+        id,
+        buyRuleIds.map((id) => ({ id }))
+      ))
+
+    targetRuleIds.length &&
+      (await promotionModule.removePromotionBuyRules(
+        id,
+        targetRuleIds.map((id) => ({ id }))
+      ))
```
olivermrbl, line 74: question: Can we parallelize?

packages/core-flows/src/promotion/steps/add-rules-to-promotions.ts
```diff
@@ -0,0 +1,85 @@
...
+
+    const createdPromotionBuyRules =
+      ruleType === RuleType.BUY_RULES
+        ? await promotionModule.addPromotionBuyRules(data.id, data.rules)
+        : []
+
+    const createdPromotionTargetRules =
+      ruleType === RuleType.TARGET_RULES
+        ? await promotionModule.addPromotionTargetRules(data.id, data.rules)
+        : []
```
olivermrbl, line 32: question: Can we parallelize?

olivermrbl (approved):
LGTM, great work. Minor comments. 

Can't figure out if it makes sense to squash the endpoints, but I think we'll find out when building out the admin UI for it.

riqwan:
@olivermrbl Same, I had an urge to do `promotions/[id]/[rule_type]` as well 😄

packages/core-flows/src/promotion/steps/add-rules-to-promotions.ts
riqwan: Only one of these would ever be run on a workflow run, parallelising doesn't help much here.

olivermrbl: Ah yes, don't know how I missed that. Discard the comment.

PR #6693, opened 2024-03-13
Feat(core-flows, inventory-next, medusa, types): Add create inventory item endpoint

pKorsholm:
**What**
- add `create-inventory-item` endpoint
- fix integration-tests w/ plugins and legacy modules updated to be "modern"


packages/core-flows/src/inventory/steps/attach-inventory-items.ts
```diff
@@ -0,0 +1,55 @@
...
+import { InventoryItemDTO } from "@medusajs/types"
+
+export const attachInventoryItemToVariantsStepId =
+  "attach-inventory-items-to-variants-step"
+export const attachInventoryItemToVariants = createStep(
+  attachInventoryItemToVariantsStepId,
+  async (
+    input: {
+      inventoryItem: InventoryItemDTO
+      tag?: string
```
olivermrbl, line 13: question: If this step is specifically attaching inventory items to variants, why do we call this `tag`?

packages/core-flows/src/inventory/steps/prepare-create-inventory-items.ts
```diff
@@ -0,0 +1,51 @@
...
+    )
+    const remoteQuery = container.resolve(
+      ContainerRegistrationKeys.REMOTE_QUERY
+    )
+
+    if (
+      featureFlagRouter.isFeatureEnabled(ManyToManyInventoryFeatureFlag.key)
+    ) {
+      return
+    }
```
olivermrbl: thought: core flows are for 2.0 only, so can we delete this?

packages/core-flows/src/inventory/steps/prepare-create-inventory-items.ts
```diff
@@ -0,0 +1,51 @@
...
+  MedusaError,
+  remoteQueryObjectFromString,
+} from "@medusajs/utils"
+import { StepResponse, createStep } from "@medusajs/workflows-sdk"
+
+import { InventoryNext } from "@medusajs/types"
+
+// import { TaggedInventoryItem } from "@medusajs/types"
+
+export const prepareCreateInventoryItemsStepId =
```
olivermrbl: suggestion: Rename to something specific to validation

packages/core-flows/src/inventory/steps/create-inventory-items.ts
```diff
@@ -0,0 +1,59 @@
...
+  async (data: string[] | undefined, { container }) => {
+    if (!data?.length) {
+      return
+    }
+
+    const inventoryService = container.resolve(ModuleRegistrationName.INVENTORY)
+
+    await promiseAll(
+      data.map(async (item) => {
+        await inventoryService!.deleteInventoryItem(item)
```
olivermrbl: question: Can we use bulk?

packages/medusa/src/api-v2/admin/inventory-items/middlewares.ts
```diff
@@ -0,0 +1,40 @@
...
+import { transformBody, transformQuery } from "../../../api/middlewares"
+
+import { MiddlewareRoute } from "../../../types/middlewares"
+import { authenticate } from "../../../utils/authenticate-middleware"
+
+export const adminInventoryRoutesMiddlewares: MiddlewareRoute[] = [
+  {
+    method: "ALL",
+    matcher: "/admin/inventory-items*",
+    middlewares: [authenticate("admin", ["session", "bearer"])],
```
olivermrbl: todo: Add `api_key`

packages/core-flows/src/inventory/workflows/create-inventory-items.ts
```diff
@@ -0,0 +1,32 @@
...
+import {
+  attachInventoryItemToVariants,
+  createInventoryItemsStep,
+  prepareCreateInventoryItems,
+} from "../steps"
+
+import { CreateInventoryItemInput } from "@medusajs/types"
+
+interface WorkflowInput {
+  items: (CreateInventoryItemInput & { _associationTag?: string })[]
```
olivermrbl: question: I like the idea of the `_associationTag`, but doesn't the `attachInventoryItemToVariants` step defeat its purpose?

packages/medusa/src/api-v2/admin/inventory-items/route.ts
```diff
@@ -0,0 +1,91 @@
...
+
+  if (
+    !variant_id &&
+    !featureFlagRouter.isFeatureEnabled(ManyToManyInventoryFeatureFlag.key)
+  ) {
+    throw new MedusaError(
+      MedusaError.Types.NOT_ALLOWED,
+      "variant_id is required"
+    )
+  }
```
olivermrbl: thought: Can we remove this, since this is for v2?

packages/medusa/src/api-v2/admin/inventory-items/route.ts
```diff
@@ -0,0 +1,91 @@
...
+  })
+
+  const inventory_items = items.map((i) => {
+    const [stocked_quantity, reserved_quantity] = i.location_levels.reduce(
+      ([stocked, reserved], level) => [
+        stocked + level.stocked_quantity ?? 0,
+        reserved + level.reserved_quantity ?? 0,
+      ],
+      [0, 0]
+    )
```
olivermrbl: thought: Wondering if this should be computed in the module. Would that be possible?

olivermrbl: Lol, I guess we had the same thought. Saw this in your follow-up PR 😄

packages/medusa/src/api-v2/admin/inventory-items/route.ts
srindom: We should def remove!

packages/medusa/src/api-v2/admin/inventory-items/route.ts
pKorsholm: yes, I'll add this in the list pr and remote the endpoint from the route-file 😄

packages/core-flows/src/inventory/workflows/create-inventory-items.ts
pKorsholm: I'm not sure I understand what you mean.

the `attach` method requires knowing which variant to attach each inventoryItem to. I guess keeping it abstract would be nice if you wanna override the step and attach to something different, but we can also just rename `_associationTag` to `variant_id`

olivermrbl: Okay got it, so the idea is that with the `associationTag` you only need to override a single step to attach them to something else? 

I am fine to keep the `_associationTag`, but I just find it a bit off to expose this in the workflow input. To me, the `_` prefix signals, that this is a property handled internally. Not a strong opinion though, so if you guys think we should keep it, I am fine with that.

packages/core-flows/src/inventory/steps/attach-inventory-items.ts
pKorsholm: we can call it variant_id, as mentioned below keeping it abstract might be easier if somebody want's to override the attach step?

olivermrbl: But the step is called `attachInventoryItemToVariants` – wouldn't that need to be generic too for this to make sense? Disclaimer, I don't fully understand the purpose of this tag being passed around, so at this point I am just asking questions to get the idea behind it.

packages/core-flows/src/inventory/workflows/create-inventory-items.ts
pKorsholm: I've renamed it to be `tag` from end-to-end just to not rename it halfway through

integration-tests/plugins/medusa-config.js
srindom, line 1: suggestion: Maybe revert these changes as they seem unrelated?

packages/core-flows/src/inventory/steps/validate-inventory-items-for-create.ts
```diff
@@ -0,0 +1,49 @@
...
+    )
+    const remoteQuery = container.resolve(
+      ContainerRegistrationKeys.REMOTE_QUERY
+    )
+
+    if (
+      featureFlagRouter.isFeatureEnabled(ManyToManyInventoryFeatureFlag.key)
+    ) {
+      return
+    }
```
olivermrbl: todo:
```suggestion
```

packages/medusa/src/api-v2/admin/inventory-items/validators.ts
```diff
@@ -0,0 +1,268 @@
...
+ *     type: object
+ *     externalDocs:
+ *       description: "Learn about the metadata attribute, and how to delete and update it."
+ *       url: "https://docs.medusajs.com/development/entities/overview#metadata-attribute"
+ */
+export class AdminPostInventoryItemsReq {
+  @IsOptional()
+  @IsString()
+  variant_id?: string
+
```
srindom: Variant ID shouldn't exist here - the only times we associate a variant with an inventory item as part of creation will be in `POST /admin/products` and `POST /admin/products/:id/variants`. On the inventory creation side we will not allow associations with variants.

packages/medusa/src/loaders/medusa-app.ts
srindom, line 1: What's happening here?

packages/core-flows/src/inventory/workflows/create-inventory-items.ts
```diff
@@ -0,0 +1,33 @@
...
+
+export const createInventoryItemsWorkflowId = "create-inventory-items-workflow"
+export const createInventoryItemsWorkflow = createWorkflow(
+  createInventoryItemsWorkflowId,
+  (input: WorkflowData<WorkflowInput>) => {
+    validateInventoryItemsForCreate(input.items)
+
+    const items = createInventoryItemsStep(input.items)
+
+    attachInventoryItemToVariants(items)
```
srindom: should not be in this workflow

packages/core-flows/src/inventory/steps/validate-inventory-items-for-create.ts
```diff
@@ -0,0 +1,49 @@
...
+
```
srindom: this feature flag should not exist anywhere in v2

packages/core-flows/src/inventory/steps/attach-inventory-items.ts
srindom, line 1: This is likely still relevant for `createProductVariant`/`createProduct` workflows, but just want to stress again - it should not be used in any workflows called by `/admin/inventory**`

packages/medusa/src/loaders/medusa-app.ts
pKorsholm: The updated modules require that the database url is passed along as `clientUrl` rather than `url` in the legacy modules. The way we mark entire module definitions as `legacy` cause us to load the updated modules for inventory and stock location using the input for the old modules essentially causing the next modules to run against a different database breaking the schema when running integration tests afterwards.

packages/medusa/src/api-v2/admin/inventory-items/route.ts
```diff
@@ -7,6 +7,23 @@ import {
...
+// Create inventory-item
+export const POST = async (
+  req: AuthenticatedMedusaRequest<AdminPostInventoryItemsReq>,
+  res: MedusaResponse
+) => {
+  const { result } = await createInventoryItemsWorkflow(req.scope).run({
+    input: { items: [req.validatedBody] },
+  })
+
+  const inventory_item = result[0]
```
olivermrbl: todo: Use `remoteQuery`

packages/core-flows/src/inventory/steps/validate-inventory-items-for-create.ts
```diff
@@ -0,0 +1,45 @@
...
+  MedusaError,
+  remoteQueryObjectFromString,
+} from "@medusajs/utils"
+import { StepResponse, createStep } from "@medusajs/workflows-sdk"
+
+import { InventoryNext } from "@medusajs/types"
+
+export const validateInventoryItemsForCreateStepId =
+  "validate-inventory-items-for-create-step"
+export const validateInventoryItemsForCreate = createStep(
```
olivermrbl: thought: This step seems to specifically validate if product variant inventory items already exist and I think the name should reflect that.

question: Where is this step used?

todo: The input `inventoryItem` is not used, so we can reduce the type to variants IDs

packages/core-flows/src/inventory/steps/attach-inventory-items.ts
```diff
@@ -0,0 +1,43 @@
...
+import { ContainerRegistrationKeys } from "@medusajs/utils"
+import { InventoryItemDTO } from "@medusajs/types"
+
+export const attachInventoryItemToVariantsStepId =
+  "attach-inventory-items-to-variants-step"
+export const attachInventoryItemToVariants = createStep(
+  attachInventoryItemToVariantsStepId,
+  async (
+    input: {
+      inventoryItem: InventoryItemDTO
```
adrien2p: todo: here we are asking for the entire DTO but we are only using th id, can we update the type to not be required to pass everything and more constraint to what we are expecting instead to reduce coupling.

packages/core-flows/src/inventory/steps/validate-inventory-items-for-create.ts
```diff
@@ -0,0 +1,45 @@
...
+      tag?: string
+    }[],
+    { container }
+  ) => {
+    const remoteQuery = container.resolve(
+      ContainerRegistrationKeys.REMOTE_QUERY
+    )
+
+    const query = remoteQueryObjectFromString({
+      entryPoint: "product_variant_inventory_item",
```
adrien2p: todo: this string if someone is not aware of the link seems to be getting from nowhere, should we use `LINKS.ProductVariantInventoryItem`

packages/core-flows/src/inventory/steps/validate-inventory-items-for-create.ts
```diff
@@ -0,0 +1,45 @@
...
+
+import { InventoryNext } from "@medusajs/types"
+
+export const validateInventoryItemsForCreateStepId =
+  "validate-inventory-items-for-create-step"
+export const validateInventoryItemsForCreate = createStep(
+  validateInventoryItemsForCreateStepId,
+  async (
+    input: {
+      inventoryItem: InventoryNext.CreateInventoryItemInput
```
adrien2p: todo: rm

packages/core-flows/src/inventory/steps/validate-inventory-items-for-create.ts
```diff
@@ -0,0 +1,45 @@
...
+      tag?: string
```
adrien2p: suggestion: tag seems weird no? What is the tag when you consume the step, shouldn't we ask for variant_id?

packages/medusa/src/api-v2/admin/inventory-items/route.ts
adrien2p: remote query with the requested fields

adrien2p (approved):
🔥 few comments

packages/core-flows/src/inventory/steps/validate-inventory-items-for-create.ts
pKorsholm: well, we should not use the remote query here anyways right, we should just use the link module service to list the links? 😉

packages/core-flows/src/inventory/steps/validate-inventory-items-for-create.ts
pKorsholm: it's not used anywhere, I kept it because of Sebs comment here: https://github.com/medusajs/medusa/pull/6693#discussion_r1525308767

packages/core-flows/src/inventory/steps/validate-inventory-items-for-create.ts
pKorsholm: we could, I've kept to the terminology from api-v1, figured we kept it as tag instead of variant=_id to keep it abstract in case of overrides: https://github.com/medusajs/medusa/pull/6693#discussion_r1523708129

PR #6694, opened 2024-03-13
Feat(inventory-next, medusa): Add List inventory items endpoint

pKorsholm:
**What**
- add list-inventory-items endpoint
- fix integration-tests w/ plugins and legacy modules updated to be "modern"

### DIFF
--- a/packages/core-flows/src/index.ts
+++ b/packages/core-flows/src/index.ts
@@ -1,3 +1,4 @@
+export * from "./auth"
 export * from "./api-key"
 export * from "./customer"
 export * from "./customer-group"
--- a/packages/medusa/src/loaders/index.ts
+++ b/packages/medusa/src/loaders/index.ts
@@ -35,6 +35,7 @@ import searchIndexLoader from "./search-index"
 import servicesLoader from "./services"
 import strategiesLoader from "./strategies"
 import subscribersLoader from "./subscribers"
+import medusaProjectApisLoader from "./load-medusa-project-apis"
 
 type Options = {
   directory: string
@@ -81,7 +82,12 @@ async function loadLegacyModulesEntities(configModules, container) {
   }
 }
 
-async function loadMedusaV2({ configModule, featureFlagRouter, expressApp }) {
+async function loadMedusaV2({
+  rootDirectory,
+  configModule,
+  featureFlagRouter,
+  expressApp,
+}) {
   const container = createMedusaContainer()
 
   // Add additional information to context of request
@@ -124,6 +130,14 @@ async function loadMedusaV2({ configModule, featureFlagRouter, expressApp }) {
     featureFlagRouter,
   })
 
+  medusaProjectApisLoader({
+    rootDirectory,
+    container,
+    app: expressApp,
+    configModule,
+    activityId: "medusa-project-apis",
+  })
+
   return {
     container,
     app: expressApp,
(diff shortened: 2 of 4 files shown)


packages/medusa/src/api-v2/admin/inventory-items/route.ts
```diff
@@ -0,0 +1,54 @@
...
+    },
+    fields: [...(req.listConfig.select as string[])],
+  })
+
+  const { rows: items, metadata } = await remoteQuery({
+    ...query,
+  })
+
+  const inventory_items = items.map((i) => {
+    const [stocked_quantity, reserved_quantity] = i.location_levels.reduce(
```
riqwan: sugg: might be good to check the existence of location_levels or safely traverse here since you can break it here if you send "expand=" without it. 

Additionally, this could also be done on location level load within the module?

packages/medusa/src/api-v2/admin/inventory-items/validators.ts
```diff
@@ -0,0 +1,268 @@
...
+ *     description: An optional set of key-value pairs with additional information.
+ *     type: object
+ *     externalDocs:
+ *       description: "Learn about the metadata attribute, and how to delete and update it."
+ *       url: "https://docs.medusajs.com/development/entities/overview#metadata-attribute"
+ */
+export class AdminPostInventoryItemsReq {
+  @IsOptional()
+  @IsString()
+  variant_id: string
```
riqwan, line 215:
```suggestion
  variant_id?: string
```

packages/medusa/src/api-v2/admin/inventory-items/validators.ts
```diff
@@ -0,0 +1,268 @@
...
+
+  @IsString()
+  @IsOptional()
```
riqwan, line 218: q: does the order of the decorator matter here?

integration-tests/plugins/medusa-config.js
```diff
@@ -38,41 +40,95 @@ module.exports = {
...
-      options: {
-        providers: [
-          {
-            name: "emailpass",
-            scopes: {
-              admin: {},
-              store: {},
+    workflows: true,
+    ...(enableMedusaV2
+      ? {
```
riqwan: q: whats going on here?

packages/medusa/src/api-v2/admin/inventory-items/query-config.ts
```diff
@@ -0,0 +1,55 @@
+import { InventoryItemDTO, InventoryLevelDTO } from "@medusajs/types"
+
+export const defaultAdminInventoryItemRelations = []
+export const allowedAdminInventoryItemRelations = []
+
+export const defaultAdminLocationLevelFields: (
+  | keyof InventoryLevelDTO
+  | "available_quantity"
```
riqwan: q: this can safely go in the InventoryLevelDTO right?

riqwan (approved):
added some suggestions, but lgtm!

packages/medusa/src/api-v2/admin/inventory-items/validators.ts
srindom: should not be here for v2 since inventory items and products are completely separate from each other

Note: We can still support creating an inventory item when creating a product variant for convenience, but inventory item creation with a variant should not be supported at the API level

integration-tests/plugins/__tests__/inventory-next/inventory-items/index.spec.ts
srindom, line 1: q: move this to modules test suite?

packages/medusa/src/api-v2/admin/inventory-items/validators.ts
pKorsholm: remedied this in the create pr where this dto is used 😄

integration-tests/plugins/medusa-config.js
pKorsholm: removed it, forgot about the modules test suite 😅

packages/medusa/src/api-v2/admin/inventory-items/route.ts
pKorsholm: changed it up to calculate in the database instead

PR #6696, opened 2024-03-13
feat(core-flows,medusa,types): remove rules from promotion endpoints + workflows

riqwan:
what:

- adds endpoints to remove rules from a promotion:
  - promotion rules
  - target rules
  - buy rules

- updates promotion remove*Rules method to simply accept rule_ids to delete instead of array of objects with just id


integration-tests/modules/__tests__/promotion/admin/promotion-rules.spec.ts
```diff
@@ -342,6 +342,194 @@ medusaIntegrationTestRunner({
           )
```
olivermrbl, line 342: comment(this file): At some point, this file should be moved to `integration-tests/api`

olivermrbl (approved):
Nice work @riqwan

integration-tests/modules/__tests__/promotion/admin/promotion-rules.spec.ts
riqwan: Shouldn't we wait for the CI to run v2 api before doing that?

olivermrbl: Yeah, that's true

PR #6697, opened 2024-03-14
feat(medusa,types): GET admin promotion endpoint to fetch by code

riqwan:
what:

- GET admin promotion endpoint will search for a promotion either through ID or code

### DIFF
--- a/packages/medusa/src/api-v2/admin/promotions/[id]/route.ts
+++ b/packages/medusa/src/api-v2/admin/promotions/[id]/route.ts
@@ -29,7 +29,6 @@ export const GET = async (
       select: req.retrieveConfig.select,
       relations: req.retrieveConfig.relations,
       take: 1,
-      skip: 0,
     }
   )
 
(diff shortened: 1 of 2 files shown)


packages/types/src/promotion/common/promotion.ts
```diff
@@ -42,8 +42,8 @@ export interface UpdatePromotionDTO {
 
 export interface FilterablePromotionProps
   extends BaseFilterable<FilterablePromotionProps> {
-  id?: string[]
-  code?: string[]
+  id?: string | string[]
+  code?: string | string[]
```
adrien2p: nit: should we add OperatorMap?

packages/medusa/src/api-v2/admin/promotions/[id]/route.ts
```diff
@@ -2,27 +2,43 @@ import {
...
-    select: req.retrieveConfig.select,
-    relations: req.retrieveConfig.relations,
-  })
+  const [promotion] = await promotionModuleService.list(
+    { $or: [{ id: idOrCode }, { code: idOrCode }] },
+    {
+      select: req.retrieveConfig.select,
+      relations: req.retrieveConfig.relations,
+      take: 1,
+      skip: 0,
```
Comment (adrien2p) [inline packages/medusa/src/api-v2/admin/promotions/[id]/route.ts]: suggestion: rm, I believe that we only expect one result here right?

adrien2p (approved):

packages/types/src/promotion/common/promotion.ts
riqwan: why not!

PR #6698, opened 2024-03-14
feat(core-flows,medusa,types,utils): add/remove fulfillment shipping option rules

riqwan:
what:

- adds fulfillment shipping option rules batch endpoint
- remove fulfillment shipping option rules batch endpoint
- breaks my British based education to not call it fulfilment with a single "l"

### DIFF
--- a/packages/admin-next/dashboard/public/locales/en-US/translation.json
+++ b/packages/admin-next/dashboard/public/locales/en-US/translation.json
@@ -446,6 +446,7 @@
         "adminOnly": "Admin only",
         "adminOnlyHint": "When enabled the shipping option will only be available in the admin dashboard, and not in the storefront."
       },
+      "taxInclusiveHint": "When enabled, the shipping option's price will be tax inclusive.",
       "requirements": {
         "label": "Requirements",
         "hint": "Specify the requirements for the shipping option."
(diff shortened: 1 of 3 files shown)


packages/medusa/src/api-v2/admin/fulfillment/shipping-options/[id]/rules/batch/remove/route.ts
```diff
@@ -0,0 +1,40 @@
...
+  const fulfillmentService: IFulfillmentModuleService = req.scope.resolve(
+    ModuleRegistrationName.FULFILLMENT
+  )
+
+  const shippingOption = await fulfillmentService.retrieveShippingOption(id, {
+    select: defaultAdminShippingOptionFields,
+    relations: defaultAdminShippingOptionRelations,
+  })
+
+  res.status(200).json({ shipping_option: shippingOption })
```
Comment (riqwan) [inline packages/medusa/src/api-v2/admin/fulfillment/shipping-options/[id]/rules/batch/remove/route.ts:39]: question here for the POST batch remove:

Do we still follow the same response as a delete endpoint? 
```
{
  id: req.validatedBody.rule_ids,
  object: "shipping-option-rules",
  deleted: true
}
```

or do we stick to what we do with regular POST endpoints?

packages/utils/src/common/rules.ts
```diff
@@ -0,0 +1,10 @@
+export enum RuleOperator {
```
adrien2p, line 1: suggestion: should we use that and the type above in the fulfillment module now? wdyt?

riqwan: yes agree, just pushed a commit to that.

PR #6701, opened 2024-03-14
chore(medusa): use batch/add and batch/remove endpoints for promotion rules

riqwan:
what:

- use standardized batch endpoint routes for adding and removing rules

### DIFF
--- a/integration-tests/modules/__tests__/inventory/index.spec.ts
+++ b/integration-tests/modules/__tests__/inventory/index.spec.ts
@@ -1,0 +1,726 @@
+import { IInventoryServiceNext, IStockLocationService } from "@medusajs/types"
+
+import { ContainerRegistrationKeys } from "@medusajs/utils"
+import { ModuleRegistrationName } from "@medusajs/modules-sdk"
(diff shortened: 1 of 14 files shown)


integration-tests/modules/__tests__/promotion/admin/promotion-rules.spec.ts
```diff
@@ -376,14 +378,12 @@ medusaIntegrationTestRunner({
...
-              ...adminHeaders,
-              data: { rule_ids: [standardPromotion.rules[0].id] },
-            }
+          const response = await api.post(
+            `/admin/promotions/${standardPromotion.id}/rules/batch/remove`,
+            { rule_ids: [standardPromotion.rules[0].id] },
+            adminHeaders
           )
-
+          console.log("response.data ---- ", response.data)
```
riqwan:
```suggestion

```

olivermrbl:
Same comment as previously, but I think we should consider squashing the endpoints. Let's not deal with this now. Think it will be more clear as we build the UI in admin

PR #6702, opened 2024-03-14
feat(medusa,core-flows,types): adds update promotion rule endpoint + workflow

riqwan:
what:

- adds endpoint + workflow to update promotion rule
- adds method in promotion to update promotion rules

### DIFF
--- a/packages/promotion/src/services/promotion-module.ts
+++ b/packages/promotion/src/services/promotion-module.ts
@@ -823,8 +823,11 @@ export default class PromotionModuleService<
     }
 
     const [updatedRules] = await Promise.all([
-      this.promotionRuleService_.update(rulesToUpdate),
-      this.promotionRuleValueService_.delete(ruleValueIdsToDelete),
+      this.promotionRuleService_.update(rulesToUpdate, sharedContext),
+      this.promotionRuleValueService_.delete(
+        ruleValueIdsToDelete,
+        sharedContext
+      ),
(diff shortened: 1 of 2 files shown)


packages/promotion/src/services/promotion-module.ts
```diff
@@ -750,6 +754,83 @@ export default class PromotionModuleService<
...
+            value,
+            promotion_rule: promotionRule,
+          }))
+        )
+      }
+    }
+
+    const [updatedRules] = await Promise.all([
+      this.promotionRuleService_.update(rulesToUpdate),
+      this.promotionRuleValueService_.delete(ruleValueIdsToDelete),
```
olivermrbl: question: Do these need the context?

olivermrbl (commented):
LGTM, just a single question :)

olivermrbl (approved):

olivermrbl:
Also, let's get the other PRs merged first

olivermrbl:
Needs a rebasing

packages/promotion/src/services/promotion-module.ts
riqwan: These don't since you're working with the rule object directly. Same for remove as well. Don't think the deliniation is necessary. I'd say makes a case to squash them. We can do that at a later point since changing that is easy now.

riqwan: edit: yes, these do. I thought you were talking about something else entirely 😄

PR #6703, opened 2024-03-14
feat(dashboard,medusa): Draft order detail

kasperkristensen:
**What**
- Adds draft order details page
- Adds Shipping and Billing address forms to both draft order and order pages
- Adds Email form to both draft order and order pages
- Adds transfer ownership form to draft order, order and customer pages
- Update Combobox component allowing it to work with async data (`useInfiniteQuery`)

**@medusajs/medusa**
- Include country as a default relation of draft order addresses

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -8166,7 +8166,7 @@ __metadata:
     i18next-browser-languagedetector: 7.2.0
     i18next-http-backend: 2.4.2
     match-sorter: ^6.3.4
-    medusa-react: latest
+    medusa-react: "workspace:^"
     postcss: ^8.4.33
     prettier: ^3.1.1
     react: 18.2.0
@@ -38261,7 +38261,7 @@ __metadata:
   languageName: unknown
   linkType: soft
 
-"medusa-react@^9.0.13, medusa-react@workspace:packages/medusa-react":
+"medusa-react@^9.0.13, medusa-react@workspace:^, medusa-react@workspace:packages/medusa-react":
   version: 0.0.0-use.local
   resolution: "medusa-react@workspace:packages/medusa-react"
   dependencies:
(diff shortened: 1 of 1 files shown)


packages/admin-next/dashboard/src/components/common/combobox/combobox.tsx
```diff
@@ -12,217 +12,318 @@ import * as Popover from "@radix-ui/react-popover"
...
+    }
+
+    if (!isValueControlled) {
+      setUncontrolledValue(newValues || emptyState)
+    }
+    if (onChange) {
+      onChange(newValues)
+    }
+
+    setUncontrolledSearchValue("")
```
riqwan, line 98: shouldn't this be guarded against `isArrayValue`? On line 88 its being set to the label and here we're setting it back to ""

olivermrbl:
UI-wise LGTM!

Not sure if the following error is directly related to this PR or something more general, but upon opening "Transfer Ownership", the following error is thrown client-side:

CleanShot 2024-03-15 at 08 36 50

olivermrbl (approved):

kasperkristensen:
@olivermrbl The reason the transfer orders modal doesn't work is because we are building the dashboard inside of our monorepo with local versions of our internal dependencies. 

This means that yarn will symlink `medusa-react` into `dashboard` including its node_modules. We then end up with two copies of `@tanstack/react-query` in `/dashboard/node_modules` and that causes this error. In the modal I am using `useInfiniteQuery` that gets imported from `node_modules/@tanstack/react-query` where every other query is importing it from `/node_modules/medusa-react/node_modules/@tanstack/...`. It won't occur in "real-life", and to get around it we can install `medusa-react@latest` instead of `workspace:^`, but that is currently missing a couple of new hooks, so will have to be after the next release. You can test it out in local now though, as it's not an issue in development, only production 👍

riqwan (approved):
lgtm!

PR #6707, opened 2024-03-15
fix(dashboard): Discount condition fixes

kasperkristensen:


packages/admin-next/dashboard/src/components/common/list-summary/list-summary.tsx
```diff
@@ -61,6 +61,7 @@ export const ListSummary = ({
             className="cursor-default whitespace-nowrap"
             title={title}
```
fPolic: nit: I would remove this - it's a bit annoying since we already have a tooltip on hover

fPolic (commented):
LGTM - looks better!

Only thing I noticed with this changes is that ListSummary text is cut i.e. longer entity names are not truncated in the middle:
Screenshot 2024-03-15 at 10 15 44

kasperkristensen:
> LGTM - looks better!
> 
> Only thing I noticed with this changes is that ListSummary text is cut i.e. longer entity names are not truncated in the middle: Screenshot 2024-03-15 at 10 15 44

Hmm, I am not really sure to do about this, because before it was truncating to some point, and then everything would just overflow off the screen. Again there is just something fundamentally off with this component, but to not spend more time on it for now, should we maybe just change it to say "7 products" which you can then hover to get the list instead of trying to show 1-2 names, in any case we aren't really giving users a lot of useful information.

PR #6709, opened 2024-03-15
Chore/upgrade mikroorm 6

sradevski:
Leaving it open here as a draft until we pick it up again.
There are changes I did during the development that need to be removed, so please ignore the PR until it's marked as ready


riqwan:
@sradevski would it be worth to keep this open? We could always bring back once we migrate to DML, which will make this much more easier

sradevski:
Definitely one for closing, thanks :)

PR #6710, opened 2024-03-15
docs: prep for v2 documentation

shahednasser:
This PR includes documentation that preps for v2 docs (but doesn't introduce new docs).

_Note: The number of file changes in the PR is due to find-and-replace within the `references` which is unavoidable. Let me know if I should move it to another PR._

## Changes

- Change Medusa version in base OAS used for v2.
- Fix to docblock generator related to not catching all path parameters.
- Added typedoc plugin that generates ER Diagrams, which will be used specifically for data model references in commerce modules.
- Changed OAS tool to output references in `www/apps/api-reference/specs-v2` directory when the `--v2` option is used.
- Added a version switcher to the API reference to switch between V1 and V2. This switcher is enabled by an environment variable, so it won't be visible/usable at the moment.
- Upgraded docusaurus to v3.0.1
- Added new Vale rules to ensure correct spelling of Medusa Admin and module names.
- Added new components to the `docs-ui` package that will be used in future documentation changes.

### DIFF
--- a/www/apps/api-reference/components/VersionNote/index.tsx
+++ b/www/apps/api-reference/components/VersionNote/index.tsx
@@ -10,7 +10,9 @@ const VersionNote = () => {
     <>
       {version === "2" && (
         <Note type="warning" title="Production Warning">
-Medusa v2.0 is in development and not suitable for production environments. As such, the API reference is incomplete and subject to change, so please use it with caution. 
+          Medusa v2.0 is in development and not suitable for production
+          environments. As such, the API reference is incomplete and subject to
+          change, so please use it with caution.
         </Note>
       )}
     </>


www/apps/api-reference/components/VersionNote/index.tsx
```diff
@@ -0,0 +1,23 @@
...
+
+  return (
+    <>
+      {version === "2" && (
+        <Note type="warning" title="Production Warning">
+          Medusa V2 is still in development, so, the API Routes are highly
+          experimental and are subject to change. The reference is also
+          incomplete. We strongly advise against using it for production, at the
+          moment.
+        </Note>
```
shahednasser: @olivermrbl let me know what you think of this note's text, it'll be shown in the API reference when switching from V1 to V2

www/apps/api-reference/components/VersionNote/index.tsx
```diff
@@ -0,0 +1,23 @@
...
+  const { version } = useVersion()
+
+  return (
+    <>
+      {version === "2" && (
+        <Note type="warning" title="Production Warning">
+          Medusa V2 is still in development, so, the API Routes are highly
+          experimental and are subject to change. The reference is also
+          incomplete. We strongly advise against using it for production, at the
+          moment.
```
olivermrbl: nit: Minor rewording
```suggestion
Medusa v2.0 is in development and not suitable for production environments. As such, the API reference is incomplete and subject to change, so please use it with caution. 
```

olivermrbl (approved):

PR #6711, opened 2024-03-15
feat(core-flows,medusa,types): create/update workflows to create and update PriceList  prices

riqwan:
what:

- create and update workflows for price list prices
- update price list endpoint does a "set"

### DIFF
--- a/packages/medusa/src/api-v2/admin/price-lists/[id]/prices/batch/add/route.ts
+++ b/packages/medusa/src/api-v2/admin/price-lists/[id]/prices/batch/add/route.ts
@@ -3,7 +3,7 @@ import {
   AuthenticatedMedusaRequest,
   MedusaResponse,
 } from "../../../../../../../types/routing"
-import { listPriceLists } from "../../../../queries"
+import { getPriceList } from "../../../../queries"
 import {
   adminPriceListRemoteQueryFields,
   defaultAdminPriceListFields,
@@ -27,11 +27,11 @@ export const POST = async (
     throw errors[0].error
   }
 
-  const [[priceList]] = await listPriceLists({
+  const priceList = await getPriceList({
+    id,
     container: req.scope,
     remoteQueryFields: adminPriceListRemoteQueryFields,
     apiFields: defaultAdminPriceListFields,
-    variables: { filters: { id }, skip: 0, take: 1 },
   })
 
   res.status(200).json({ price_list: priceList })
--- a/packages/medusa/src/api-v2/admin/price-lists/[id]/prices/batch/remove/route.ts
+++ b/packages/medusa/src/api-v2/admin/price-lists/[id]/prices/batch/remove/route.ts
@@ -3,7 +3,7 @@ import {
   AuthenticatedMedusaRequest,
   MedusaResponse,
 } from "../../../../../../../types/routing"
-import { listPriceLists } from "../../../../queries"
+import { getPriceList } from "../../../../queries"
 import {
   adminPriceListRemoteQueryFields,
   defaultAdminPriceListFields,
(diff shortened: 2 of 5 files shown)


packages/core-flows/src/price-list/steps/update-price-list-prices.ts
```diff
@@ -0,0 +1,90 @@
...
+      priceListPricesToUpdate.push({
+        price_list_id: id,
+        prices: pricesToUpdate,
+      })
+    }
+
+    const priceIds = priceListPricesToUpdate
+      .map((priceListData) => priceListData.prices.map((price) => price.id))
+      .flat(1)
+      .filter(Boolean)
```
olivermrbl: suggestion: Move both the mapping and filtering to the loop just before:

```ts
let priceIds: string[] = []
for (const priceListData of data) {
  const pricesToUpdate: UpdatePriceListPriceDTO[] = []
  const { prices = [], id } = priceListData

  for (const price of prices) {
    pricesToUpdate.push({
      ...price,
      price_set_id: variantPriceSetMap[price.variant_id!],
    })

    // This should take care of both no?
    if (price.id) {
      priceIds.push(price.id)
    }
  }

  priceListPricesToUpdate.push({
    price_list_id: id,
    prices: pricesToUpdate,
  })
}
```

olivermrbl (approved):
LGTM

packages/core-flows/src/price-list/steps/update-price-list-prices.ts
```diff
@@ -0,0 +1,90 @@
...
+  IPricingModuleService,
+  PriceSetMoneyAmountDTO,
+  UpdatePriceListPriceDTO,
+  UpdatePriceListPricesDTO,
+  UpdatePriceListPricesWorkflowDTO,
+} from "@medusajs/types"
+import { buildPriceSetPricesForModule } from "@medusajs/utils"
+import { StepResponse, createStep } from "@medusajs/workflows-sdk"
+
+type WorkflowStepInput = {
```
adrien2p: too: rename to something close to the step ?

adrien2p (approved):
🔥 just few comments to handle

packages/core-flows/src/price-list/steps/update-price-list-prices.ts
riqwan: nice!

packages/core-flows/src/price-list/steps/create-price-list-prices.ts
```diff
@@ -0,0 +1,61 @@
...
+    if (!createdIds?.length) {
+      return
+    }
+
+    const pricingModule = container.resolve<IPricingModuleService>(
+      ModuleRegistrationName.PRICING
+    )
+
+    if (createdIds.length) {
+      await pricingModule.removePrices(createdIds)
```
sradevski, line 58: Would this be `removePriceListPrices`?

packages/core-flows/src/price-list/steps/create-price-list-prices.ts
```diff
@@ -0,0 +1,61 @@
...
+  CreatePriceListPricesWorkflowStepDTO,
+  IPricingModuleService,
+} from "@medusajs/types"
+import { StepResponse, createStep } from "@medusajs/workflows-sdk"
+
+export const createPriceListPricesStepId = "create-price-list-prices"
+export const createPriceListPricesStep = createStep(
+  createPriceListPricesStepId,
+  async (stepInput: CreatePriceListPricesWorkflowStepDTO, { container }) => {
+    const { data, variant_price_map: variantPriceSetMap } = stepInput
```
sradevski, line 14: A general question: Should the diffing of what to create and what to remove be done in the module itself?

packages/core-flows/src/price-list/steps/create-price-list-prices.ts
riqwan: We don't have a scoped method for that yet. Both price set and price list prices are the same entity, so its ok to do this for now. Will revisit this when that becomes available.

packages/core-flows/src/price-list/steps/create-price-list-prices.ts
riqwan: My personal pref would be for the module to keep things atomic, but I don't have a strong opinion on this. Feel free to put it up on the eng channel and check with the others

PR #6712, opened 2024-03-15
feat(medusa): rework prices add/remove to batch spec

riqwan:
what:

- reworks add and remove apis to batch spec


olivermrbl (approved):

sradevski (approved):

packages/medusa/src/api-v2/admin/price-lists/[id]/prices/batch/add/route.ts
```diff
@@ -0,0 +1,38 @@
...
+
+  if (Array.isArray(errors) && errors[0]) {
+    throw errors[0].error
+  }
+
+  const [[priceList]] = await listPriceLists({
+    container: req.scope,
+    remoteQueryFields: adminPriceListRemoteQueryFields,
+    apiFields: defaultAdminPriceListFields,
+    variables: { filters: { id }, skip: 0, take: 1 },
```
Comment (adrien2p) [inline packages/medusa/src/api-v2/admin/price-lists/[id]/prices/batch/add/route.ts:34]: todo: why do we have pagination here?

adrien2p (commented):
🔥

adrien2p (approved):

packages/medusa/src/api-v2/admin/price-lists/[id]/prices/batch/add/route.ts
Comment (riqwan) [reply packages/medusa/src/api-v2/admin/price-lists/[id]/prices/batch/add/route.ts:34]: meaning? I'm just doing a take 1

Comment (riqwan) [reply packages/medusa/src/api-v2/admin/price-lists/[id]/prices/batch/add/route.ts:34]: Will keep an eye on this comment to see if there is anything to change. 🤞🏻

Comment (adrien2p) [reply packages/medusa/src/api-v2/admin/price-lists/[id]/prices/batch/add/route.ts:34]: I mean, there should not be a need to do take one no?

Comment (adrien2p) [reply packages/medusa/src/api-v2/admin/price-lists/[id]/prices/batch/add/route.ts:34]: I just don't see the need to do pagination to force a take 1 here ahah, since you are filtering by id it can just do a retrieve no?

Comment (riqwan) [reply packages/medusa/src/api-v2/admin/price-lists/[id]/prices/batch/add/route.ts:34]: Ah, I just noticed that the remote query does a retrieve in that case. Didn't know that 😄 thanks! Will be added in the base PR

Comment (adrien2p) [reply packages/medusa/src/api-v2/admin/price-lists/[id]/prices/batch/add/route.ts:34]: ahah no worries, now you know :)

PR #6715, opened 2024-03-16
feat: Init. v2 implementation in admin

olivermrbl:
**What**

Add first couple of v2 domains in admin.

Much of what is in this PR is subject to change, but right now, we need to have a playground where we can start stitching UI together with our backend to see what's missing, what needs improving, and how things are generally coming together.

`@medusajs/medusa`

- Add (temp) V2 user support in `medusa user ...` command
- Add defaults workflow in V2 loader

`@medusajs/link-modules`

- Add store default currency link

`@medusajs/core-flows`
- Add `createDefaultsWorkflow`
  - Creates store (if not exists)
  - Creates default sales channel (if not exists)

`@medusajs/admin-next`

- Add login (currently fixed to use `emailpass`)
- Add settings foundation
  - User management
  - Store management

Note: @kasperkristensen, I've generally stayed away from adding conditional logic in admin except for the `shell.tsx`. I wanted to avoid having to copy over much of the layout. However, I am fine with doing that, if you think it is better.

### DIFF
--- a/.changeset/cold-mice-deliver.md
+++ b/.changeset/cold-mice-deliver.md
@@ -1,0 +1,6 @@
+---
+"@medusajs/modules-sdk": patch
+"@medusajs/orchestration": patch
+---
+
+fix(orchestration): Properly handle select all
--- a/packages/modules-sdk/src/__tests__/remote-quer.spec.ts
+++ b/packages/modules-sdk/src/__tests__/remote-quer.spec.ts
@@ -1,0 +1,31 @@
+import { RemoteQuery } from "../remote-query"
+
+describe("Remote query", () => {
+  it("should properly handle fields and relations transformation", () => {
+    const expand = {
+      fields: ["name", "age"],
+      expands: {
+        friend: {
+          fields: ["name"],
+          expands: {
+            ball: {
+              fields: ["*"],
+            },
+          },
+        },
+      },
+    }
+
+    const result = RemoteQuery.getAllFieldsAndRelations(expand)
+
+    expect(result).toEqual({
+      select: ["name", "age", "friend.name"],
+      relations: ["friend", "friend.ball"],
+      args: {
+        "": undefined,
+        friend: undefined,
+        "friend.ball": undefined,
+      },
+    })
+  })
+})
--- a/packages/modules-sdk/src/remote-query.ts
+++ b/packages/modules-sdk/src/remote-query.ts
@@ -4,12 +4,14 @@ import {
   toRemoteJoinerQuery,
 } from "@medusajs/orchestration"
 import {
+  JoinerArgument,
   JoinerRelationship,
   JoinerServiceConfig,
   LoadedModule,
   ModuleJoinerConfig,
   RemoteExpandProperty,
   RemoteJoinerQuery,
+  RemoteNestedExpands,
 } from "@medusajs/types"
 import { isString, toPascalCase } from "@medusajs/utils"
 
@@ -78,42 +80,43 @@ export class RemoteQuery {
   }
 
   public static getAllFieldsAndRelations(
-    data: any,
+    expand: RemoteExpandProperty | RemoteNestedExpands[number],
     prefix = "",
-    args: Record<string, unknown[]> = {}
+    args: JoinerArgument = {} as JoinerArgument
   ): {
     select: string[]
     relations: string[]
-    args: Record<string, unknown[]>
+    args: JoinerArgument
   } {
+    expand = JSON.parse(JSON.stringify(expand))
+
     let fields: Set<string> = new Set()
     let relations: string[] = []
 
-    data.fields?.forEach((field: string) => {
+    for (const field of expand.fields ?? []) {
       if (field === "*") {
-        // Select all, so we don't specify any field and rely on relation only
-        return
+        expand.fields = []
+        break
       }
       fields.add(prefix ? `${prefix}.${field}` : field)
-    })
-    args[prefix] = data.args
+    }
 
-    if (data.expands) {
-      for (const property in data.expands) {
-        const newPrefix = prefix ? `${prefix}.${property}` : property
+    args[prefix] = expand.args
 
-        relations.push(newPrefix)
-        fields.delete(newPrefix)
+    for (const property in expand.expands ?? {}) {
+      const newPrefix = prefix ? `${prefix}.${property}` : property
 
-        const result = RemoteQuery.getAllFieldsAndRelations(
-          data.expands[property],
-          newPrefix,
-          args
-        )
+      relations.push(newPrefix)
+      fields.delete(newPrefix)
 
-        result.select.forEach(fields.add, fields)
-        relations = relations.concat(result.relations)
-      }
+      const result = RemoteQuery.getAllFieldsAndRelations(
+        expand.expands![property],
+        newPrefix,
+        args
+      )
+
+      result.select.forEach(fields.add, fields)
+      relations = relations.concat(result.relations)
     }
 
     return { select: [...fields], relations, args }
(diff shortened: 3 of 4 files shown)


packages/core-flows/src/defaults/steps/create-default-store.ts
```diff
@@ -0,0 +1,49 @@
...
+
+    let shouldDelete = false
+    let [store] = await storeService.list()
+
+    if (!store) {
+      store = await createStoresWorkflow(container).run({
+        input: {
+          stores: [
+            {
+              // Should we be more sophisticated about the defaults here?
```
sradevski: It would be a nice improvement for sure, so we can just tag the comment with a TODO or FUTURE

kasperkristensen (approved):
Admin changes LGTM 👍

packages/admin-next/dashboard/src/v2-routes/profile/profile-detail/profile-detail.tsx
```diff
@@ -0,0 +1,26 @@
...
+  if (isLoading) {
+    return <div>Loading...</div>
+  }
+
+  if (isError || !user) {
+    if (error) {
+      throw error
+    }
+
+    throw json("An unknown error has occured", 500)
```
adrien2p, line 17: Q:if there is no use and no error, what could be the cause?

packages/admin-next/dashboard/src/v2-routes/profile/profile-edit/profile-edit.tsx
```diff
@@ -0,0 +1,26 @@
...
+import { useV2Session } from "../../../lib/api-v2"
+
+export const ProfileEdit = () => {
+  const { user, isLoading, isError, error } = useV2Session()
+
+  const { t } = useTranslation()
+
+  if (isError) {
+    throw error
+  }
```
adrien2p, line 14: Q: previously we also checked if there is no error and no user, shouldn't we do the same here? otherewise the heading will be displayed but without a form

packages/admin-next/dashboard/src/lib/api-v2/store.ts
```diff
@@ -0,0 +1,14 @@
+import { adminStoreKeys, useAdminCustomQuery } from "medusa-react"
+
+export const useV2Store = ({ initialData }: { initialData?: any }) => {
```
adrien2p, line 3:
```suggestion
export const useV2Store = ({ initialData }: { initialData?: any } = {}) => {
```

packages/admin-next/dashboard/src/v2-routes/store/store-add-currencies/store-add-currencies.tsx
```diff
@@ -0,0 +1,17 @@
+import { RouteFocusModal } from "../../../components/route-modal"
+import { AddCurrenciesForm } from "./components/add-currencies-form/add-currencies-form"
+import { useV2Store } from "../../../lib/api-v2"
+
+export const StoreAddCurrencies = () => {
+  const { store, isLoading, isError, error } = useV2Store({})
```
adrien2p, line 6:
```suggestion
  const { store, isLoading, isError, error } = useV2Store()
```

packages/admin-next/dashboard/src/v2-routes/store/store-detail/loader.ts
```diff
@@ -0,0 +1,38 @@
...
+
+const fetchQuery = async (
+  query: FetchQueryOptions<Response<{ stores: any[] }>>
+) => {
+  try {
+    const res = await queryClient.fetchQuery(query)
+    // TODO: Reconsider store retrieval
+    return res
+  } catch (error) {
+    const err = error ? JSON.parse(JSON.stringify(error)) : null
```
adrien2p, line 22:
```suggestion
    const err = (error ? JSON.parse(JSON.stringify(error)) : null) as Error & { status: number } | null
```

packages/admin-next/dashboard/src/v2-routes/store/store-detail/loader.ts
```diff
@@ -0,0 +1,38 @@
...
+
+    if ((err as Error & { status: number })?.status === 401) {
```
adrien2p, line 24:
```suggestion
    if (err?.status === 401) {
```

packages/admin-next/dashboard/src/v2-routes/store/store-detail/store-detail.tsx
```diff
@@ -0,0 +1,36 @@
...
+
+  const { store, isLoading, isError, error } = useV2Store({
+    initialData: initialData,
+  })
+
+  if (isLoading) {
+    return <div>Loading...</div>
+  }
+
+  if (isError || !store) {
```
adrien2p, line 20: Q: There is some places where we use the `usev2Store` but we only check if there is isError but not if store is present, maybe worth a check in the other places above

packages/core-flows/src/defaults/steps/create-default-store.ts
```diff
@@ -0,0 +1,49 @@
...
+}
+
+export const createDefaultStoreStepId = "create-default-store"
+export const createDefaultStoreStep = createStep(
+  createDefaultStoreStepId,
+  async (data: CreateDefaultStoreStepInput, { container }) => {
+    const storeService = container.resolve(ModuleRegistrationName.STORE)
+
+    let shouldDelete = false
+    let [store] = await storeService.list()
```
adrien2p:
```suggestion
    let [store] = await storeService.list({}, { take: 1 })
```

packages/core-flows/src/sales-channel/steps/create-default-sales-channel.ts
```diff
@@ -0,0 +1,42 @@
...
+export const createDefaultSalesChannelStepId = "create-default-sales-channel"
+export const createDefaultSalesChannelStep = createStep(
+  createDefaultSalesChannelStepId,
+  async (input: StepInput, { container }) => {
+    const salesChannelService = container.resolve<ISalesChannelModuleService>(
+      ModuleRegistrationName.SALES_CHANNEL
+    )
+
+    let shouldDelete = false
+    let [salesChannel] = await salesChannelService.list()
```
adrien2p:
```suggestion
    let [salesChannel] = await salesChannelService.list({}, { take: 1 })
```

packages/medusa/src/api-v2/admin/stores/query-config.ts
```diff
@@ -5,6 +5,10 @@ export const defaultAdminStoreFields = [
   "name",
   "supported_currency_codes",
   "default_currency_code",
+  "default_currency.code",
```
adrien2p: todo: duplicated

packages/medusa/src/api-v2/admin/stores/query-config.ts
```diff
@@ -5,6 +5,10 @@ export const defaultAdminStoreFields = [
...
+  "default_currency.name",
+  "default_currency.symbol",
+  "default_currency.symbol_native",
```
adrien2p, line 8: suggestion: maybe worth cleaning the config below using the new api?

olivermrbl:
@adrien2p, a general comment: I am not going to make any changes to the UI, as this is copied over from the components for V1. I think it is fine to add the comments here, so we have them for the future, but I won't address them in this PR.

adrien2p:
> @adrien2p, a general comment: I am not going to make any changes to the UI, as this is copied over from the components for V1. I think it is fine to add the comments here, so we have them for the future, but I won't address them in this PR.

@olivermrbl I understand, from experience, the comments will never be addressed, so maybe just discard them, though, there is comments on workflows and v2 hooks/functions which is probably not coming from v1 copy pasting, if you could also answer some of my question it would be nice to understand the why. and if something really need to be done, then we should create linea tickets. But again, after a certain time the linear tickets are closed automatically and we will probably forget about it, especially about issues

adrien2p (approved):

PR #6722, opened 2024-03-17
feat: Sales Channels API routes + workflows

olivermrbl:
**What**
- Admin sales channels API routes
- Workflows for creating, updating, and deleting sales channels
- Align `ISalesChannelModuleService` interface with update + upsert conventions
- Integrate v2 tests into v1 sales channels integration test suite
- Add metadata to sales channel entity

Sales channel <> Product management will come in a follow-up PR.

On the integration tests, creating, updating, and deleting are passing with the V2 flag enabled. You'll have to take my word for it (or run them yourself), as they are not included in the CI.

### DIFF
--- a/.changeset/silly-weeks-melt.md
+++ b/.changeset/silly-weeks-melt.md
@@ -1,0 +1,5 @@
+---
+"@medusajs/workflows-sdk": patch
+---
+
+fix(workflows-sdk): Value resolver should resolve non StepResponse
--- a/packages/workflow-engine-inmemory/integration-tests/__fixtures__/workflow_1.ts
+++ b/packages/workflow-engine-inmemory/integration-tests/__fixtures__/workflow_1.ts
@@ -1,7 +1,7 @@
 import {
-  StepResponse,
   createStep,
   createWorkflow,
+  StepResponse,
 } from "@medusajs/workflows-sdk"
 
 const step_1 = createStep(
@@ -15,7 +15,6 @@ const step_1 = createStep(
       return
     }
 
-    console.log("reverted", compensateInput.compensate)
     return new StepResponse({
       reverted: true,
     })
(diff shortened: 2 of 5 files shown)


packages/medusa/src/api-v2/admin/sales-channels/route.ts
```diff
@@ -0,0 +1,68 @@
...
+    filters: req.filterableFields,
+    order: req.listConfig.order,
+    skip: req.listConfig.skip,
+    take: req.listConfig.take,
+  }
+
+  const queryObject = remoteQueryObjectFromString({
+    entryPoint: "sales_channels",
+    variables,
+    fields: defaultAdminSalesChannelFields,
```
sradevski: This should come from the `req` object instead.

packages/medusa/src/api-v2/admin/sales-channels/route.ts
```diff
@@ -0,0 +1,68 @@
...
+  if (Array.isArray(errors) && errors[0]) {
+    throw errors[0].error
+  }
+
+  const remoteQuery = req.scope.resolve(ContainerRegistrationKeys.REMOTE_QUERY)
+
+  const queryObject = remoteQueryObjectFromString({
+    entryPoint: "sales_channels",
+    variables: { id: req.params.id },
+    fields: defaultAdminSalesChannelFields,
```
sradevski: Ideally we also support selecting `fields` on POST, so the user can select whatever they want after creating a resource. Nothing to action on this PR, just something we need to probably tackle

sradevski (approved):

packages/medusa/src/api-v2/admin/sales-channels/route.ts
olivermrbl: Adding the `transformQuery` middleware should do the job

PR #6723, opened 2024-03-18
migrate to @tanstack/react-query v5

tboutron:
## **What** - what changes are in this PR
Migration of `@tanstack/react-query` from v4 to v5

## **Why** - why are these changes relevant
Medusajs is futur, we don't want past library

## **How** - how have the changes been implemented
- Upgraded libraries
- Ran some regex to upgrade to new react-query syntax
- While yarn build return error
  => Fix errors
- While test using local project
  => Fix erros
- Youpi, let's make a PR

### DIFF
--- a/.changeset/old-hotels-fetch.md
+++ b/.changeset/old-hotels-fetch.md
@@ -1,0 +1,7 @@
+---
+"medusa-payment-stripe": major
+"medusa-react": major
+"@medusajs/admin-ui": major
+---
+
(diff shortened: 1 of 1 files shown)


olivermrbl:
@tboutron, appreciate the contribution. However, we would like to manage a major library upgrade internally, as there are other elements than just the upgrade that need addressing. We should discuss when the time is right for the react-query upgrade, if you are still interested in contributing.

PR #6725, opened 2024-03-18
feat: Add products to sales channel

olivermrbl:
Depends on #6722


fPolic (approved):

packages/core-flows/src/sales-channel/steps/associate-products-with-channels.ts
```diff
@@ -0,0 +1,47 @@
...
+}
+
+export const associateProductsWithSalesChannelsStepId =
+  "associate-products-with-channels"
+export const associateProductsWithSalesChannelsStep = createStep(
+  associateProductsWithSalesChannelsStepId,
+  async (input: StepInput, { container }) => {
+    const remoteLink = container.resolve(ContainerRegistrationKeys.REMOTE_LINK)
+
+    const links = input.links
```
adrien2p, line 19: discussion: I believe that we are missing a validation step somewhere, unless we decide to trust the data and if an id does not exists it is the user problem. But currently, you could insert any id and it would work without issues. Do you think we should do something about that?

packages/medusa/src/api-v2/admin/sales-channels/[id]/products/batch/add/route.ts
```diff
@@ -0,0 +1,49 @@
...
+  if (Array.isArray(errors) && errors[0]) {
+    throw errors[0].error
+  }
+
+  const remoteQuery = req.scope.resolve(ContainerRegistrationKeys.REMOTE_QUERY)
+
+  const queryObject = remoteQueryObjectFromString({
+    entryPoint: "sales_channels",
+    variables: { id: req.params.id },
+    fields: defaultAdminSalesChannelFields,
```
Comment (adrien2p) [inline packages/medusa/src/api-v2/admin/sales-channels/[id]/products/batch/add/route.ts:43]: todo: should we use the new req.remoteQueryConfig.fields?

packages/medusa/src/api-v2/admin/sales-channels/middlewares.ts
```diff
@@ -56,4 +57,9 @@ export const adminSalesChannelRoutesMiddlewares: MiddlewareRoute[] = [
     matcher: "/admin/sales-channels/:id",
     middlewares: [],
   },
+  {
+    method: ["POST"],
+    matcher: "/admin/sales-channels/:id/products/batch/add",
+    middlewares: [transformBody(AdminPostSalesChannelsChannelProductsBatchReq)],
```
adrien2p, line 63: discussion: Same question as another pr, should we support transformQuery here wdyt?

packages/medusa/src/api-v2/admin/sales-channels/[id]/products/batch/add/route.ts
Comment (adrien2p) [reply packages/medusa/src/api-v2/admin/sales-channels/[id]/products/batch/add/route.ts:43]: in fact, depends on the next comment sorry

adrien2p (approved):
🔥 few comments

PR #6726, opened 2024-03-18
fix(workflows-sdk): Value resolver should resolve non StepResponse

adrien2p:
**What**
When marking a step as success, the response should be returned by the async step to be usable in other step. It was only working if the provided response was a `StepResponse`. It can now be just an object as well


packages/workflow-engine-inmemory/integration-tests/__fixtures__/workflow_2.ts
```diff
@@ -22,15 +22,16 @@ const step_1 = createStep(
   })
 )
 
+export const workflow2Step2Invoke = jest.fn((input, context) => {
+  console.log("triggered async request", context.metadata.idempotency_key)
```
sradevski: I guess we can remove this

sradevski (approved):

carlos-r-l-rodrigues (approved):

PR #6729, opened 2024-03-18
feat(create-medusa-app): Added --v2 option

shahednasser:
## What

Added a `--v2` option to the `create-medusa-app` command that clones the [feat/v2](https://github.com/medusajs/medusa-starter-default/pull/150) branch of the starter default and makes minor changes to the setup process.

## Why

This option is supposed to make it easier to test out internally a Medusa setup with all the commerce modules. The starter's branch installs and configures the modules + the V2 feature flag.

## Testing

To test it out, run the snapshot with the `--v2` option.

## Notes

- I couldn't install the new admin dashboard in the starter branch as it's not on NPM yet (at the time of writing this), so at the moment, this opens the current admin dashboard instead.
- When trying to create an invite using the CLI tool I get the error `Unable to resolve inviteService`. Not sure if we should also make changes to the CLI tool to allow creating an invite when V2 is enabled (I know now invites are creating within the User Module, so maybe when the V2 feature flag is enabled that should be resolved instead), but for now I just disabled running the create invite command as this is mainly for internal testing. Let me know what the suggested approach is here.


olivermrbl (approved):

olivermrbl:
Nice one @shahednasser 💪

PR #6730, opened 2024-03-18
feat(fulfillment): Events management

adrien2p:
**What**
This pr includes some cleanup and refactoring around the abstract event emitter method for the module service to not rely on a named class property but instead on resolution naming.

Includes fulfillment set events on creation. 

The idea is that if that pr allows us to align and agreed on the approach as well as including the cleanup for other pr to use, If this gets merged I ll continue with another pr to do the rest of the event management

partially fix CORE-1735

### DIFF
--- a/.changeset/cold-mice-deliver.md
+++ b/.changeset/cold-mice-deliver.md
@@ -1,0 +1,6 @@
+---
+"@medusajs/modules-sdk": patch
+"@medusajs/orchestration": patch
+---
+
+fix(orchestration): Properly handle select all
--- a/.changeset/tidy-gifts-talk.md
+++ b/.changeset/tidy-gifts-talk.md
@@ -1,0 +1,7 @@
+---
+"@medusajs/core-flows": minor
+"@medusajs/product": minor
+"@medusajs/types": minor
+---
+
+Updated the signature of the product module variant and options endpoints to follow our conventions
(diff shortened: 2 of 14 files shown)


packages/fulfillment/src/services/fulfillment-module-service.ts
```diff
@@ -1557,4 +1566,66 @@ export default class FulfillmentModuleService<
...
+    for (const fulfillmentSet of createdFulfillmentSets) {
+      sharedContext.messageAggregator!.saveRawMessageData(
+        buildMessage({
+          eventName: FulfillmentUtils.FulfillmentEvents.created,
+          id: fulfillmentSet.id,
+          object: "fulfillment_set",
+        })
+      )
+
+      if (fulfillmentSet.service_zones?.length) {
```
sradevski: We can have an early return here, and below as well

sradevski (approved):
LGTM

packages/utils/src/event-bus/utils.ts
```diff
@@ -0,0 +1,86 @@
+import { camelToSnakeCase, kebabCase, lowerCaseFirst } from "../common"
```
olivermrbl, line 1: praise: nice!

packages/fulfillment/src/services/fulfillment-module-service.ts
adrien2p: simplified a bit more

adrien2p: wdyt?

olivermrbl (approved):

PR #6731, opened 2024-03-18
feat: make it possible to use subscribers in v2

srindom:


adrien2p (approved):
LGTM 🔥  does upgrading bullmq to 2 majors have any impacts?

PR #6737, opened 2024-03-19
feat(core-flows): Use remote link methods to dismiss all links related to inventory item

pKorsholm:
**What**
- use `remoteLink.delete` to dismiss all connections to `inventory_item_id`


carlos-r-l-rodrigues:
I think we can rename that file and the step to refer to deletion as well, because in theory you can still have a place where you just want to call a detach and that would be your current step.

adrien2p:
Should we also check that the link configurations include the cascade option?

Cc @carlos-r-l-rodrigues

adrien2p (commented):
Do we have tests to validate the deletions of the links with multiple levels or links?

packages/core-flows/src/inventory/steps/deatach-inventory-items.ts
```diff
@@ -13,47 +13,21 @@ export const deatachInventoryItemStep = createStep(
   async (ids: string[], { container }) => {
     const remoteLink = container.resolve(ContainerRegistrationKeys.REMOTE_LINK)
 
-    const linkModule: ILinkModule = remoteLink.getLinkModule(
-      Modules.PRODUCT,
-      "variant_id",
-      Modules.INVENTORY,
-      "inventory_item_id"
-    )
+    await remoteLink.delete({
```
riqwan: oh cool, didn't know this does a soft delete. Would be good to have a test that checks this.

riqwan: suggestion: don't have to do this now, but I wonder if we can do a generic `deleteLinkStep()` that accepts an object. Something like this:

```
deleteLinkStep({
  [Modules.INVENTORY]: { inventory_item_id: data.ids },
})
```

riqwan (approved):
one suggestion, but lgtm!

packages/core-flows/src/inventory/steps/deatach-inventory-items.ts
carlos-r-l-rodrigues: https://github.com/medusajs/medusa/blob/94208a979bb14a6bbbb4ea5ba8f712a7068917df/packages/core-flows/src/common/steps/remove-remote-links.ts

if needed you can bring this one to the PR

carlos-r-l-rodrigues:
> Should we also check that the link configurations include the cascade option?
> 
> Cc @carlos-r-l-rodrigues

Yeah, I think so. But I'm not sure if we have one yet. Maybe the link between product variant and inventory item?

adrien2p:
> > Should we also check that the link configurations include the cascade option?
> > Cc @carlos-r-l-rodrigues
> 
> Yeah, I think so. But I'm not sure if we have one yet. Maybe the link between product variant and inventory item?

Yeah something like that, or cart -> sales_channel -> fulfillment too

packages/core-flows/src/inventory/steps/deatach-inventory-items.ts
riqwan: nice!

olivermrbl: @carlos-r-l-rodrigues very nice!

olivermrbl (approved):

carlos-r-l-rodrigues (approved):

PR #6739, opened 2024-03-19
feat: v2 - add worker mode

srindom:
**What**
- Adds support for starting a Medusa process with a worker mode.
- The worker modes supported are "shared", "worker", "server"
- In "worker" mode, API routes are not registered and modules that need to run workers (e.g., event bus redis) can use the flag to conditionally start workers.
- In "server" mode, API routes are registered and workers are not started.
- In "shared" mode, API routes are registered and workers are started. This is great for development.


srindom:
@carlos-r-l-rodrigues, @adrien2p - I will need some help with conditionally starting workflow-engine-redis workers too.

adrien2p (commented):
Finally !! 🎉 we can start subscribers only 🔥

adrien2p:
It can be an option of the module right? And you can access the options as the second argument of the constructor too

adrien2p:
It will free the node event loop a lot when handling subscribers separately from the api

srindom:
> It can be an option of the module right? And you can access the options as the second argument of the constructor too

My thinking on this is that it makes more sense to define it at the MedusaApp-level and not the module-level. This gives a more standardized way for modules to do conditional stuff depending on the worker mode.

That said we could have a priority stack for this:

1. Worker mode in module confg
2. Worker mode in project config
3. Worker mode in environment variables
4. Default "shared"

--

Think it would be great if the mode always gets passed to the module as moduleDeclaration instead of an option. Again to standardize usage.

We could potentially do it like this in medusa-config.js
```javascript
// An eventbus only worker
{
  resolve: "@medusajs/event-bus-redis",
  options: { redisUrl: REDIS_URL },
  worker_mode: "worker"
},
{
  resolve: "@medusajs/workflow-engine-redis",
  options: { redis: { url: REDIS_URL } },
  worker_mode: "server"
},

// A workflow engine only worker
{
  resolve: "@medusajs/event-bus-redis",
  options: { redisUrl: REDIS_URL },
  worker_mode: "server"
},
{
  resolve: "@medusajs/workflow-engine-redis",
  options: { redis: { url: REDIS_URL } },
  worker_mode: "worker"
},
```

adrien2p:
Yes definitely, i like the priority order, but the modules does not have access to the config as it is not injected and they do not require it as a dependency. But it can be changed and when using medusa app it can be added to the dependencies and getting injected in the module, so we can apply the ordering as you suggested

srindom:
@olivermrbl, @adrien2p - can we get this in as is for now and then come back to improvements later on? :)

adrien2p (approved):
yes we can @srindom if you want you can create a linear ticket to come back to it later with non automatic closing :)

olivermrbl (approved):

PR #6740, opened 2024-03-19
feat: Add product and pricing link on create and delete operations

sradevski:
Things that remain to be done:
1. Handle product and variant updates
2. Add tests for the workflows independently
3. Align the endpoints to the new code conventions we defined
4. Finish up the update/upsert endpoints for variants

All of those can be done in a separate PR, as this is quite large already.

### DIFF
--- a/packages/core-flows/src/product/steps/create-variant-pricing-link.ts
+++ b/packages/core-flows/src/product/steps/create-variant-pricing-link.ts
@@ -1,4 +1,5 @@
 import { Modules } from "@medusajs/modules-sdk"
+import { ContainerRegistrationKeys } from "@medusajs/utils"
 import { StepResponse, createStep } from "@medusajs/workflows-sdk"
 
 type StepInput = {
@@ -12,7 +13,7 @@ export const createVariantPricingLinkStepId = "create-variant-pricing-link"
 export const createVariantPricingLinkStep = createStep(
   createVariantPricingLinkStepId,
   async (data: StepInput, { container }) => {
-    const remoteLink = container.resolve("remoteLink")
+    const remoteLink = container.resolve(ContainerRegistrationKeys.REMOTE_LINK)
     await remoteLink.create(
       data.links.map((entry) => ({
         [Modules.PRODUCT]: {
@@ -27,11 +28,11 @@ export const createVariantPricingLinkStep = createStep(
     return new StepResponse(void 0, data)
   },
   async (data, { container }) => {
-    if (!data) {
+    if (!data?.links?.length) {
       return
     }
 
-    const remoteLink = container.resolve("remoteLink")
+    const remoteLink = container.resolve(ContainerRegistrationKeys.REMOTE_LINK)
     const links = data.links.map((entry) => ({
       [Modules.PRODUCT]: {
         variant_id: entry.variant_id,
--- a/packages/core-flows/src/product/workflows/create-product-variants.ts
+++ b/packages/core-flows/src/product/workflows/create-product-variants.ts
@@ -1,4 +1,4 @@
-import { ProductTypes } from "@medusajs/types"
+import { ProductTypes, PricingTypes } from "@medusajs/types"
 import {
   WorkflowData,
   createWorkflow,
@@ -10,8 +10,11 @@ import {
 } from "../steps"
 import { createPriceSetsStep } from "../../pricing"
 
+// TODO: Create separate typings for the workflow input
 type WorkflowInput = {
-  product_variants: ProductTypes.CreateProductVariantDTO[]
+  product_variants: (ProductTypes.CreateProductVariantDTO & {
+    prices?: PricingTypes.CreateMoneyAmountDTO[]
+  })[]
 }
 
 export const createProductVariantsWorkflowId = "create-product-variants"
@@ -38,11 +41,11 @@ export const createProductVariantsWorkflow = createWorkflow(
           .map((variant, i) => {
             return {
               id: variant.id,
-              prices: (data.input.product_variants[i] as any).prices,
+              prices: data.input.product_variants[i]?.prices,
             }
           })
           .flat()
-          .filter((v) => v.prices?.length > 0)
+          .filter((v) => !!v.prices?.length)
     )
 
     // TODO: From here until the final transform the code is the same as when creating a product, we can probably refactor
--- a/packages/core-flows/src/product/workflows/create-products.ts
+++ b/packages/core-flows/src/product/workflows/create-products.ts
@@ -1,4 +1,4 @@
-import { ProductTypes } from "@medusajs/types"
+import { ProductTypes, PricingTypes } from "@medusajs/types"
 import {
   WorkflowData,
   createWorkflow,
@@ -9,7 +9,13 @@ import { createPriceSetsStep } from "../../pricing"
 
 // TODO: We should have separate types here as input, not the module DTO. Eg. the HTTP request that we are handling
 // has different data than the DTO, so that needs to be represented differently.
-type WorkflowInput = { products: ProductTypes.CreateProductDTO[] }
+type WorkflowInput = {
+  products: (Omit<ProductTypes.CreateProductDTO, "variants"> & {
+    variants?: (ProductTypes.CreateProductVariantDTO & {
+      prices?: PricingTypes.CreateMoneyAmountDTO[]
+    })[]
+  })[]
+}
 
 export const createProductsWorkflowId = "create-products"
 export const createProductsWorkflow = createWorkflow(
@@ -39,11 +45,11 @@ export const createProductsWorkflow = createWorkflow(
             const inputProduct = data.input.products[i]
             return p.variants?.map((v, j) => ({
               id: v.id,
-              prices: (inputProduct?.variants?.[j] as any).prices,
+              prices: inputProduct?.variants?.[j]?.prices,
             }))
           })
           .flat()
-          .filter((v) => v.prices?.length > 0)
+          .filter((v) => !!v.prices?.length)
       }
     )
 
@@ -72,6 +78,7 @@ export const createProductsWorkflow = createWorkflow(
 
     createVariantPricingLinkStep(variantAndPriceSetLinks)
 
+    // TODO: Should we just refetch the products here?
     return transform(
       {
         createdProducts,
--- a/packages/medusa/src/api-v2/admin/products/validators.ts
+++ b/packages/medusa/src/api-v2/admin/products/validators.ts
@@ -679,6 +679,7 @@ export class ProductTypeReq {
   value: string
 }
 
+// TODO: Add support for rules
 export class ProductVariantPricesCreateReq {
   @IsString()
   currency_code: string
@@ -700,12 +701,8 @@ export class ProductVariantPricesUpdateReq {
   @IsOptional()
   id?: string
 
-  @ValidateIf((o) => !o.id)
-  @Validate(XorConstraint, ["currency_code"])
-  region_id?: string
-
-  @ValidateIf((o) => !o.id)
-  @Validate(XorConstraint, ["region_id"])
+  @IsString()
+  @IsOptional()
   currency_code?: string
 
   @IsInt()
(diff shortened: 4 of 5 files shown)


packages/core-flows/src/product/steps/create-variant-pricing-link.ts
```diff
@@ -0,0 +1,46 @@
...
+        [Modules.PRICING]: {
+          price_set_id: entry.price_set_id,
+        },
+      }))
+    )
+
+    return new StepResponse(void 0, data)
+  },
+  async (data, { container }) => {
+    if (!data) {
```
riqwan: To guard against an empty object scenario

```suggestion
    if (!data?.links?.length) {
```

packages/core-flows/src/product/steps/create-variant-pricing-link.ts
```diff
@@ -0,0 +1,46 @@
...
+    variant_id: string
+    price_set_id: string
+  }[]
+}
+
+export const createVariantPricingLinkStepId = "create-variant-pricing-link"
+export const createVariantPricingLinkStep = createStep(
+  createVariantPricingLinkStepId,
+  async (data: StepInput, { container }) => {
+    const remoteLink = container.resolve("remoteLink")
```
riqwan: You can use the `ContainerRegistrationKeys.REMOTE_LINK` here and in the other files

packages/core-flows/src/product/steps/remove-variant-pricing-link.ts
```diff
@@ -0,0 +1,49 @@
...
+    )
+
+    const links = (await linkModule.list(
+      {
+        variant_id: data.variant_ids,
+      },
+      { select: ["id", "variant_id", "price_set_id"] }
+    )) as { id: string; variant_id: string; price_set_id: string }[]
+
+    await remoteLink.delete(links.map((link) => link.id))
```
riqwan, line 29: does this work? Correct me if i'm wrong, but I think you might have to do something like this:

```
// or remoteLink.dismiss?

await remoteLink.delete([{
  [Modules.PRODUCT]: { variant_id: "" },
  [Modules.PRICING]: { price_set_id: "" },
}])
```

packages/core-flows/src/product/workflows/create-product-variants.ts
```diff
@@ -1,6 +1,14 @@
...
+  transform,
+} from "@medusajs/workflows-sdk"
+import {
+  createProductVariantsStep,
+  createVariantPricingLinkStep,
+} from "../steps"
+import { createPriceSetsStep } from "../../pricing"
 
 type WorkflowInput = {
   product_variants: ProductTypes.CreateProductVariantDTO[]
```
riqwan: suggestion: would be good to add a specific workflow type here since we're accepting prices here along with `CreateProductVariantDTO`

packages/core-flows/src/product/workflows/create-products.ts
```diff
@@ -10,6 +17,77 @@ export const createProductsWorkflow = createWorkflow(
...
+      {
+        createdProducts,
+        variantAndPriceSets,
+      },
+      (data) => {
+        return data.createdProducts.map((product) => ({
+          ...product,
+          variants: product.variants?.map((variant) => ({
+            ...variant,
+            price_set: data.variantAndPriceSets.find(
```
riqwan, line 92: suggestion: we could also just refetch in the create API using the remote query and skip this dance.

packages/core-flows/src/product/workflows/create-products.ts
```diff
@@ -1,7 +1,14 @@
...
-import { createProductsStep } from "../steps"
+import {
+  WorkflowData,
+  createWorkflow,
+  transform,
+} from "@medusajs/workflows-sdk"
+import { createProductsStep, createVariantPricingLinkStep } from "../steps"
+import { createPriceSetsStep } from "../../pricing"
 
+// TODO: We should have separate types here as input, not the module DTO. Eg. the HTTP request that we are handling
```
riqwan, line 10: This is what I did in my PR and found the experience to be much better working with modules.

packages/core-flows/src/product/steps/remove-variant-pricing-link.ts
```diff
@@ -0,0 +1,49 @@
...
+    return new StepResponse(void 0, links)
+  },
+  async (prevData, { container }) => {
+    if (!prevData?.length) {
+      return
+    }
+
+    const remoteLink = container.resolve(ContainerRegistrationKeys.REMOTE_LINK)
+    await remoteLink.create(
```
riqwan, line 38: You might have to do a restore here in case any of the steps fail. You still want to have the prices reverted upon compensation.

packages/core-flows/src/product/workflows/delete-product-variants.ts
```diff
@@ -1,12 +1,17 @@
...
+  removeVariantPricingLinkStep,
+} from "../steps"
 
 type WorkflowInput = { ids: string[] }
 
 export const deleteProductVariantsWorkflowId = "delete-product-variants"
 export const deleteProductVariantsWorkflow = createWorkflow(
   deleteProductVariantsWorkflowId,
   (input: WorkflowData<WorkflowInput>): WorkflowData<void> => {
+    // Question: Should we also remove the price set manually, or would that be cascaded?
```
riqwan, line 13: I think when using `delete` through the remoteLink, it soft deletes and cascades.

packages/core-flows/src/product/workflows/delete-products.ts
```diff
@@ -1,12 +1,31 @@
...
   (input: WorkflowData<WorkflowInput>): WorkflowData<void> => {
+    const productsToDelete = getProductsStep({ ids: input.ids })
+    const variantsToBeDeleted = transform({ productsToDelete }, (data) => {
+      return data.productsToDelete
+        .flatMap((product) => product.variants)
+        .map((variant) => variant.id)
+    })
+
+    // Question: Should we also remove the price set manually, or would that be cascaded?
+    // Question: Since we soft-delete the product, how do we restore the product with the prices and the links?
```
riqwan, line 26: restoring the links will cascade restore the prices as well

packages/medusa/src/api-v2/admin/products/helpers.ts
```diff
@@ -0,0 +1,46 @@
...
+
+// The variant had prices before, but that is not part of the price_set money amounts. Do we remap the request and response or not?
+export const remapKeysForProduct = (selectFields: string[]) => {
+  const productFields = selectFields.filter(
+    (fieldName: string) => !fieldName.startsWith("variants.prices")
+  )
+  const pricingFields = selectFields
+    .filter((fieldName: string) => fieldName.startsWith("variants.prices"))
+    .map((fieldName: string) =>
+      fieldName.replace("variants.prices.", "variants.price_set.money_amounts.")
```
riqwan, line 11: suggestion: once we do the model changes in pricing, we should just return the exact shape we would get from the remote query, price_set included. I don't see any reason to hide it from the consumers.

packages/medusa/src/api-v2/admin/products/validators.ts
```diff
@@ -679,3 +678,44 @@ export class ProductTypeReq {
...
+  @IsInt()
+  amount: number
+
+  @IsOptional()
+  @IsInt()
+  min_quantity?: number
+
+  @IsOptional()
+  @IsInt()
+  max_quantity?: number
```
riqwan, line 696: this is missing `rules` for a price. This can be region_id or sales_channel_id or pretty much anything else.

packages/medusa/src/api-v2/admin/products/validators.ts
```diff
@@ -679,3 +678,44 @@ export class ProductTypeReq {
...
+}
+
+export class ProductVariantPricesUpdateReq {
+  @IsString()
+  @IsOptional()
+  id?: string
+
+  @ValidateIf((o) => !o.id)
+  @Validate(XorConstraint, ["currency_code"])
+  region_id?: string
```
riqwan: this should now be under "rules" object

packages/medusa/src/api-v2/admin/products/query-config.ts
```diff
@@ -22,6 +22,11 @@ export const defaultAdminProductsVariantFields = [
   "ean",
   "upc",
   "barcode",
+  "prices.id",
+  "prices.currency_code",
+  "prices.amount",
+  "prices.created_at",
+  "prices.updated_at",
```
riqwan, line 29: we need to append rules here as well and in the response, normalize it to a flexible object like:
```
{
  region_id: "region-id"
  sales_channel_id: "sales-channel-id"
}
```

riqwan (approved):
clean PR! 💪🏻 🔥 

Question about the tests: can we enable the v2 API tests on the pipeline? It feels like we won't know whats breaking unless we run it locally.

sradevski:
> clean PR! 💪🏻 🔥
> 
> Question about the tests: can we enable the v2 API tests on the pipeline? It feels like we won't know whats breaking unless we run it locally.

I will enable it once I have all the tests passing, this is still WIP and there are few tests that will fail on V2

packages/core-flows/src/product/steps/remove-variant-pricing-link.ts
sradevski: @riqwan There is a test for this that is failing, I just decided to not focus on it for now, and I'll get back to it later, since I had some questions about how this should be done properly (including cascades, soft deletes, etc.)

packages/core-flows/src/product/workflows/create-products.ts
sradevski: @riqwan true, I guess it would also be good to decide what the workflow should respond with. If we do a refetch in the HTTP handler, I'd rather not do it both in the workflow and there. Let me raise this in eng-code, I'll add a in the meantime.

packages/core-flows/src/product/workflows/create-products.ts
sradevski: I'll just add some TODOs for now, there is a lot of cleanup to be done anyway

packages/core-flows/src/product/steps/remove-variant-pricing-link.ts
sradevski: @riqwan this is the part I am not sure how exactly it works, let me leave it as-is for a second.

packages/medusa/src/api-v2/admin/products/helpers.ts
sradevski: @riqwan we talked with Oli that we want to keep the model simple, as it is more intuitive to return just a prices entry. This is the whole point of decoupling the HTTP from module endpoints, so we can map to whatever works best for the API in mind.

packages/medusa/src/api-v2/admin/products/validators.ts
sradevski: @riqwan Should a price set have rules though? Or should we reserve that only for price lists? I don't want this to be full-blown pricelist model IMO. WDYT?

packages/medusa/src/api-v2/admin/products/validators.ts
sradevski: @riqwan removing it for now, see my comment above.

packages/medusa/src/api-v2/admin/products/query-config.ts
sradevski: See my comment above, let's see how we should handle these first.

packages/medusa/src/api-v2/admin/products/validators.ts
riqwan: True, not the price set, but the prices under the price set.

Ex.

EUR price is 400 where rules = {} // default price for EUR
FR price is 300 where rules = { region_id: "france" }

sradevski: @riqwan sorry I wasn't clear, I meant should the money amounts under a price set have rules, or should we reserve that only for price lists' money amounts, and have the price set money amounts just be defaults per currency?

riqwan: It would deviate from how we use price lists currently.

For the bulk variants editor in product admin you can set default currency price and region prices for a variant. How would that look like in this API with the delineation?

sradevski: @riqwan I haven't seen the new designs so I am not sure, but I guess we would need to support it. Let me add a TODO and I'll tackle it separately, maybe we would change this to have a different shape than it has now, adding a rule option.

PR #6741, opened 2024-03-19
Feat(medusa): add list location levels endpoint in api v2

pKorsholm:


packages/medusa/src/api-v2/admin/inventory-items/middlewares.ts
```diff
@@ -55,6 +51,21 @@ export const adminInventoryRoutesMiddlewares: MiddlewareRoute[] = [
...
+      transformQuery(
+        AdminGetInventoryItemsItemLocationLevelsParams,
+        QueryConfig.listLocationLevelsTransformQueryConfig
+      ),
+    ],
+  },
+  {
+    method: ["POST"],
+    matcher: "/admin/inventory-items/:id/location-levels",
+    middlewares: [transformBody(AdminPostInventoryItemsItemLocationLevelsReq)],
```
sradevski, line 69: As we've discussed in eng-core, We want to return the entire parent object, so we need to add the `transformQuery` middleware here as well (and then in the handler refetch the item and return it)

sradevski (approved):
One comment, but feel free to do that in a separate PR

packages/medusa/src/api-v2/admin/inventory-items/middlewares.ts
pKorsholm: I think returning the inventory-item invalidates the point of this endpoint (listing a sub-entity directly) as a retrieve for the inventory-item id with expand parameters would accomplish the same thing. 

I'm fine with closing this pr instead and just relying on that endpoint instead

olivermrbl: ~~We only return the parent object on POST requests (and DELETE, but that's slightly different), so this is the only one that needs to be updated here. We will want to keep the `GET /admin/inventory-items/:id/location-levels`~~

~~The current state of the PR would also introduce a breaking change, as we are returning the parent entity in this endpoint in V1, so incorporating this change would also avoid that.~~

Forget what I said. Current state is as expected. 

But we could still add the transformQuery on the POST as suggested above.

olivermrbl:
Created a ticket for it in our clean-up project, so we can merge this now.

PR #6742, opened 2024-03-19
fis(orchestration): Properly handle select all

adrien2p:
**What**
Fix select `*` without breaking select none `[]`

### DIFF
--- a/packages/modules-sdk/src/__tests__/remote-quer.spec.ts
+++ b/packages/modules-sdk/src/__tests__/remote-quer.spec.ts
@@ -1,0 +1,31 @@
+import { RemoteQuery } from "../remote-query"
+
+describe("Remote query", () => {
+  it("should properly handle fields and relations transformation", () => {
+    const expand = {
+      fields: ["name", "age"],
+      expands: {
+        friend: {
+          fields: ["name"],
+          expands: {
+            ball: {
+              fields: ["*"],
+            },
+          },
+        },
+      },
+    }
+
+    const result = RemoteQuery.getAllFieldsAndRelations(expand)
+
+    expect(result).toEqual({
+      select: ["name", "age", "friend.name"],
+      relations: ["friend", "friend.ball"],
+      args: {
+        "": undefined,
+        friend: undefined,
+        "friend.ball": undefined,
+      },
+    })
+  })
+})


packages/modules-sdk/src/remote-query.ts
```diff
@@ -86,34 +86,40 @@ export class RemoteQuery {
     relations: string[]
     args: Record<string, unknown[]>
   } {
+    const dataCopy = JSON.parse(JSON.stringify(data))
```
sradevski: Maybe we can name this better? queryConfig or something like that?

packages/modules-sdk/src/remote-query.ts
```diff
@@ -86,34 +86,40 @@ export class RemoteQuery {
...
     let fields: Set<string> = new Set()
     let relations: string[] = []
+    let selectAll = false
 
-    data.fields?.forEach((field: string) => {
-      if (field === "*") {
-        // Select all, so we don't specify any field and rely on relation only
+    dataCopy.fields?.forEach((field: string) => {
+      if (selectAll || field === "*") {
+        selectAll = true
```
sradevski: Do we need `selectAll` or should we just do `dataCopy.fields = []` here?

adrien2p: Screenshot 2024-03-19 at 17 14 39

the only issue is that the loop still continue if you do that, but I can improve still

packages/modules-sdk/src/remote-query.ts
```diff
@@ -86,34 +86,40 @@ export class RemoteQuery {
...
         return
```
riqwan: can you return from a forEach loop?

packages/modules-sdk/src/remote-query.ts
adrien2p: Improved a bit further than that 👍

packages/modules-sdk/src/remote-query.ts
adrien2p: Screenshot 2024-03-19 at 17 21 28
Sorry I ve removed the comment without purpose. I was saying that you can't because you still go through the loop. But I can still improve though.

I ve put an example of why it is not possible above

.changeset/cold-mice-deliver.md
```diff
@@ -0,0 +1,6 @@
+---
+"@medusajs/modules-sdk": patch
+"@medusajs/orchestration": patch
+---
+
+fis(orchestration): Properly handle select all
```
riqwan: Love me a fis 😄 

```suggestion
fix(orchestration): Properly handle select all
```

riqwan (approved):

packages/modules-sdk/src/remote-query.ts
```diff
@@ -78,42 +80,43 @@ export class RemoteQuery {
...
     prefix = "",
-    args: Record<string, unknown[]> = {}
+    args: JoinerArgument = {} as JoinerArgument
   ): {
     select: string[]
     relations: string[]
-    args: Record<string, unknown[]>
+    args: JoinerArgument
   } {
+    expand = JSON.parse(JSON.stringify(expand))
```
riqwan, line 91: Just curious, why not this? 

```suggestion
    expand = { ...expand }
```

adrien2p: Because it is a shallow copy

packages/modules-sdk/src/remote-query.ts
adrien2p: I ve already answered, see stev comment

.changeset/cold-mice-deliver.md
adrien2p: Ahah thanks nice catch

packages/modules-sdk/src/remote-query.ts
riqwan: wut! TIL 😄

olivermrbl (commented):
If this fixes something, would it make sense to add a test to ensure it doesn't break again?

sradevski (approved):

olivermrbl (approved):

PR #6743, opened 2024-03-19
feat(core-flows, types, medusa): Add Update location level endpoint for api-v2

pKorsholm:


riqwan (approved):
nice! 🔥

packages/medusa/src/api-v2/admin/inventory-items/[id]/location-levels/[location_id]/route.ts
```diff
@@ -45,3 +47,37 @@ export const DELETE = async (req: MedusaRequest, res: MedusaResponse) => {
...
+      variables: {
+        inventory_item_id,
+        location_id,
+      },
+      fields: req.retrieveConfig.select as string[],
+    })
+  )
+
+  res.status(200).json({
+    inventory_level,
```
Comment (olivermrbl) [inline packages/medusa/src/api-v2/admin/inventory-items/[id]/location-levels/[location_id]/route.ts]: nit: Respond with parent entity inventory item to comply with our convention. I also believe responding with the inventory_level would be a breaking change, and we might as well avoid that

olivermrbl (approved):
LGTM, just a minor comment!

integration-tests/modules/__tests__/inventory/index.spec.ts
```diff
@@ -284,6 +284,97 @@ medusaIntegrationTestRunner({
...
+        beforeEach(async () => {
+          const invItemReps = await api.post(
+            `/admin/inventory-items`,
+            { sku: "test-sku" },
+            adminHeaders
+          )
+
+          inventoryItemId = invItemReps.data.inventory_item.id
+
+          const sl = await appContainer
```
adrien2p: todo: as per our convention (old one) we shouldn't use one or two letter variable name, only in inline function

adrien2p: I know I can just read the resolve to understand that it is the stock location module and therefore it is probably a stock location, but better getting use to it now to stop doing that :)

integration-tests/modules/__tests__/inventory/index.spec.ts
```diff
@@ -284,6 +284,97 @@ medusaIntegrationTestRunner({
...
+            `/admin/inventory-items/${inventoryItemId}/location-levels/${locationId}`,
+            {
+              stocked_quantity: 15,
+              incoming_quantity: 5,
+            },
+            adminHeaders
+          )
+
+          expect(result.status).toEqual(200)
+          expect(result.data.inventory_level).toEqual(
```
adrien2p: todo: as per the convention I believe it should return the inventory_item, @olivermrbl is that correct?

integration-tests/modules/__tests__/inventory/index.spec.ts
```diff
@@ -284,6 +284,97 @@ medusaIntegrationTestRunner({
...
+                incoming_quantity: 5,
+              },
+              adminHeaders
+            )
+            .catch((e) => e)
+
+          expect(error.response.status).toEqual(404)
+          expect(error.response.data).toEqual({
+            type: "not_found",
+            message: `Item does-not-exist is not stocked at location ${locationId}`,
```
adrien2p, line 387: suggestion: The message here let you think that the item exists but is not stocked there, but in fact the item does not exists, should we change the message?

packages/medusa/src/api-v2/admin/inventory-items/[id]/location-levels/[location_id]/route.ts
```diff
@@ -45,3 +47,37 @@ export const DELETE = async (req: MedusaRequest, res: MedusaResponse) => {
...
+  }
+
+  const [inventory_level] = await remoteQuery(
+    remoteQueryObjectFromString({
+      entryPoint: "inventory_level",
+      variables: {
+        inventory_item_id,
+        location_id,
+      },
+      fields: req.retrieveConfig.select as string[],
```
Comment (adrien2p) [inline packages/medusa/src/api-v2/admin/inventory-items/[id]/location-levels/[location_id]/route.ts]: todo: req.remoteQueryConfig.fields, it is important to do the distinction as the value are not quite the same for select all

packages/types/src/inventory/common/inventory-level.ts
```diff
@@ -53,7 +54,8 @@ export interface InventoryLevelDTO {
   deleted_at: string | Date | null
 }
 
-export interface FilterableInventoryLevelProps {
+export interface FilterableInventoryLevelProps
+  extends BaseFilterable<FilterableInventoryLevelProps> {
```
adrien2p, line 59: Q: Should we allow OperatorMap on the props? Also, why do we use NumericalComparisonOperator which is using another type of operator and why don't we rely on OperatorMap instead for those specific operator. It seems weird to have two differet type of possible operators for our modules e.g lt vs $lt

integration-tests/modules/__tests__/inventory/index.spec.ts
olivermrbl: I believe I pointed that out above. Is this something else? 🤔

adrien2p: Unfortunately, your comment was far below and therefore i put this one before reaching yours

integration-tests/modules/__tests__/inventory/index.spec.ts
pKorsholm: the item can definitely exist, just without being stocked at the location, i.e. no inventory level with that combination of properties exist. I can run another query just to validate the inventory-items passed along or change the message a bit? 

I think partly this is just due to the id I chose 😅

adrien2p: but is this id existing for the item?

olivermrbl:
@adrien2p, OK for you to merge this?

PR #6748, opened 2024-03-19
feat: Update the signatures for variants and options to follow convention

sradevski:


packages/core-flows/src/handlers/product/update-product-variants.ts
```diff
@@ -27,7 +27,7 @@ export async function updateProductVariants({
   }
```
adrien2p, line 27: todo: general, missing change set

packages/core-flows/src/product/steps/update-product-options.ts
```diff
@@ -39,11 +40,10 @@ export const updateProductOptionsStep = createStep(
...
-    // TODO: We need to update the module's signature
-    // await service.upsertOptions(
-    //   prevData.map((r) => ({
-    //     ...r,
-    //   }))
-    // )
+    await service.upsertOptions(
+      prevData.map((r) => ({
+        ...r,
+      }))
```
adrien2p: Q: do we need the shallow copy here?

packages/core-flows/src/product/steps/update-product-variants.ts
```diff
@@ -39,11 +40,10 @@ export const updateProductVariantsStep = createStep(
...
-    // TODO: We need to update the module's signature
-    // await service.upsertVariants(
-    //   prevData.map((r) => ({
-    //     ...r,
-    //   }))
-    // )
+    await service.upsertVariants(
+      prevData.map((r) => ({
+        ...r,
+      }))
```
adrien2p: same question

packages/product/src/services/product-module-service.ts
```diff
@@ -197,58 +198,158 @@ export default class ProductModuleService<
...
-        )[0]
-      })
+    const createdVariants = await this.baseRepository_.serialize<
+      ProductTypes.ProductVariantDTO[]
+    >(variants, { populate: true })
+
+    return Array.isArray(data) ? createdVariants : createdVariants[0]
+  }
+
+  async createVariants_(
```
adrien2p: todo: missing InjectTransactionManager

packages/product/src/services/product-module-service.ts
```diff
@@ -197,58 +198,158 @@ export default class ProductModuleService<
...
+    if (data.some((v) => !v.product_id)) {
+      throw new MedusaError(
+        MedusaError.Types.INVALID_DATA,
+        "Tried to create variants without specifying a product_id"
+      )
+    }
+
+    const productOptions = await this.productOptionService_.list(
+      {
+        product_id: uniqBy(data, "product_id").map(
```
adrien2p:
```suggestion
        product_id: [...new Set<string>(data.map(d => d.product_id))]
```

adrien2p: not sure this util uniqBy is required, also, a single loop is necessary, which also simplify the readability and cyclomatic complexity, even if we dont really care here

packages/product/src/services/product-module-service.ts
```diff
@@ -197,58 +198,158 @@ export default class ProductModuleService<
...
+    data:
+      | ProductTypes.UpsertProductVariantDTO[]
+      | ProductTypes.UpsertProductVariantDTO,
+    @MedusaContext() sharedContext: Context = {}
+  ): Promise<
+    ProductTypes.ProductVariantDTO[] | ProductTypes.ProductVariantDTO
+  > {
+    const input = Array.isArray(data) ? data : [data]
+    const forUpdate = input.filter(
+      (variant): variant is UpdateProductVariantInput => !!variant.id
```
adrien2p, line 281: info: can't wait the new incoming version, which will remove the need to do that ahah

packages/product/src/services/product-module-service.ts
```diff
@@ -197,58 +198,158 @@ export default class ProductModuleService<
...
+    )
+    const forCreate = input.filter(
+      (variant): variant is ProductTypes.CreateProductVariantDTO => !variant.id
+    )
```
adrien2p, line 285:
```suggestion
const forUpdate: UpdateProductVariantInput[] = []
const forCreate: CreateProductVariantDTO[] = []
input.forEach(variant => {
  if (!!variant.id) {
    forUpdate.push(variant as UpdateProductVariantInput)
    return
  }
  
  forCreate.push(variant as CreateProductVariantDTO)
})
```

packages/product/src/services/product-module-service.ts
```diff
@@ -197,58 +198,158 @@ export default class ProductModuleService<
...
-  ): Promise<ProductTypes.ProductVariantDTO[]> {
-    const productVariants = await this.updateVariants_(data, sharedContext)
+  ): Promise<
+    ProductTypes.ProductVariantDTO[] | ProductTypes.ProductVariantDTO
+  > {
+    let normalizedInput: UpdateProductVariantInput[] = []
+    if (isString(idOrSelector)) {
+      normalizedInput = [{ id: idOrSelector, ...data }]
+    } else {
+      const variants = await this.productVariantService_.list(
```
adrien2p, line 328: q: is it really necessary to list in order to get the id only? knowing that the internal update method accept an array of selector/data? We can improve later, but in theory we have many queries to fetch the same data here and there.

packages/product/src/services/product-module-service.ts
```diff
@@ -197,58 +198,158 @@ export default class ProductModuleService<
...
-    >(productVariants)
+  async upsertVariants(
+    data: ProductTypes.UpsertProductVariantDTO[],
+    sharedContext?: Context
+  ): Promise<ProductTypes.ProductVariantDTO[]>
+  async upsertVariants(
+    data: ProductTypes.UpsertProductVariantDTO,
+    sharedContext?: Context
+  ): Promise<ProductTypes.ProductVariantDTO>
+  @InjectTransactionManager("baseRepository_")
```
adrien2p, line 270: nit: add some spec

packages/product/src/services/product-module-service.ts
```diff
@@ -427,11 +528,44 @@ export default class ProductModuleService<
...
+    const options = await this.createOptions_(input, sharedContext)
+
+    const createdOptions = await this.baseRepository_.serialize<
+      ProductTypes.ProductOptionDTO[]
+    >(options, { populate: true })
+
+    return Array.isArray(data) ? createdOptions : createdOptions[0]
+  }
+
+  async createOptions_(
```
adrien2p: todo: InjectTransactionManager

adrien2p: todo: protected?

packages/product/src/services/product-module-service.ts
adrien2p: todo: protected

packages/product/src/services/product-module-service.ts
```diff
@@ -441,21 +575,101 @@ export default class ProductModuleService<
...
+
+  async upsertOptions(
+    data: ProductTypes.UpsertProductOptionDTO[],
+    sharedContext?: Context
+  ): Promise<ProductTypes.ProductOptionDTO[]>
+  async upsertOptions(
+    data: ProductTypes.UpsertProductOptionDTO,
+    sharedContext?: Context
+  ): Promise<ProductTypes.ProductOptionDTO>
+  @InjectTransactionManager("baseRepository_")
```
adrien2p, line 593: nit: some space

packages/product/src/services/product-module-service.ts
```diff
@@ -441,21 +575,101 @@ export default class ProductModuleService<
...
+      | ProductTypes.UpsertProductOptionDTO,
+    @MedusaContext() sharedContext: Context = {}
+  ): Promise<ProductTypes.ProductOptionDTO[] | ProductTypes.ProductOptionDTO> {
+    const input = Array.isArray(data) ? data : [data]
+    const forUpdate = input.filter(
+      (option): option is UpdateProductOptionInput => !!option.id
+    )
+    const forCreate = input.filter(
+      (option): option is ProductTypes.CreateProductOptionDTO => !option.id
+    )
```
adrien2p, line 606: same suggestion as the previous one, feel free to not do it if you want

packages/core-flows/src/handlers/product/update-product-variants.ts
sradevski: Ah right, we decided to start adding changesets even for incomplete v2 work.

packages/core-flows/src/product/steps/update-product-options.ts
sradevski: I don't think so no 👍

packages/product/src/services/product-module-service.ts
sradevski: To be honest, I prefer the readability of the current solution, even if it loops twice (it should be insignificant, and if we ever see that's where the performance issue lies, we can optimize)

packages/product/src/services/product-module-service.ts
sradevski: @adrien2p this is the pattern we followed elsewhere, and it's about having the `update_` methods operating on top of an update object with IDs, and not relying on the selector. I am fine either way to be honest, but for now I won't touch it and we can revisit all modules later

packages/core-flows/src/handlers/product/update-product-variants.ts
adrien2p: The product module is already released and accessible, thats why but maybe not? @olivermrbl

packages/product/src/services/product-module-service.ts
sradevski: Same comment, I'd leave it for now 👍

packages/product/src/services/product-module-service.ts
adrien2p: I believe it was a mistake in the other modules as people didn't know that the internal method accepted selectors, fine to keep it as it is for now and we can come back to it at a later point. I remember pointing this to seb pr's too

packages/product/src/services/product-module-service.ts
```diff
@@ -295,7 +395,7 @@ export default class ProductModuleService<
   }
 
   @InjectManager("baseRepository_")
-  async diffVariants_(
+  protected async diffVariants_(
```
adrien2p, line 398: todo: InjectTransactionManager here

adrien2p (approved):
LGTM 🔥

adrien2p (commented):
LGTM, one last comment 🔥

PR #6750, opened 2024-03-20
experiment: Use JSON schema for rule engine

sradevski:


packages/fulfillment/src/services/fulfillment-module-service.ts
```diff
@@ -152,14 +157,24 @@ export default class FulfillmentModuleService<
...
-        if (!shippingOption.rules?.length) {
+        if (!isPresent(shippingOption.conditions)) {
           return true
         }
 
-        return isContextValid(
-          context,
-          shippingOption.rules.map((r) => r)
-        )
+        const validate = ajv.compile({
```
sradevski, line 164: Of course we can create a tiny util that takes care of compiling and validating, so we end up with a one-liner. We could also memoize the compilation for improved performance, but I don't think it will have significant impact anyway.

riqwan:
@sradevski is this still relevant?

sradevski:
@riqwan It was just a POC to demonstrate how it can look like, closing

PR #6751, opened 2024-03-20
feat(utils): autogenerates create and update methods when dto is provided

riqwan:
what:

when you provide an optional create_dto or update_dto to the model factory types, the abstract service will generate a simple service method for the public services of the modules.

```
MoneyAmount: {
  dto: PricingTypes.MoneyAmountDTO
  create_dto: PricingTypes.CreateMoneyAmountDTO
}
```

In the above example, only a create method will be generated, but the update will be skipped.


packages/utils/src/modules-sdk/abstract-module-service-factory.ts
```diff
@@ -303,6 +336,40 @@ export function abstractModuleServiceFactory<
...
+        ): Promise<T[]> {
+          const service = this.__container__[serviceRegistrationName]
+          const entities = await service.create(data, sharedContext)
+
+          return await this.baseRepository_.serialize<T[]>(entities, {
+            populate: true,
+          })
+        }
+
+        applyMethod(methodImplementation, 2)
```
adrien2p:
```suggestion
        applyMethod(methodImplementation, 1)
```

packages/utils/src/modules-sdk/abstract-module-service-factory.ts
```diff
@@ -303,6 +336,40 @@ export function abstractModuleServiceFactory<
...
+        ): Promise<T[]> {
+          const service = this.__container__[serviceRegistrationName]
+          const entities = await service.update(data, sharedContext)
+
+          return await this.baseRepository_.serialize<T[]>(entities, {
+            populate: true,
+          })
+        }
+
+        applyMethod(methodImplementation, 2)
```
adrien2p:
```suggestion
        applyMethod(methodImplementation, 1)
```

packages/utils/src/modules-sdk/abstract-module-service-factory.ts
```diff
@@ -155,17 +171,34 @@ export type AbstractModuleService<
...
+  [ModelName in keyof ModelsConfig as UpdateMethodName<
+    ModelsConfig,
+    ModelName
+  >]: {
+    (
+      data: ModelsConfig[ModelName]["update_dto"][],
+      sharedContext?: Context
+    ): Promise<ModelsConfig[ModelName]["dto"][]>
+  }
 }
```
adrien2p: todo:we should add the overload as per our convention

adrien2p (commented):
Nice work 🔥

sradevski (commented):
LGTM other than Adrien's comments

sradevski (approved):

packages/utils/src/modules-sdk/abstract-module-service-factory.ts
adrien2p: Will you add the single data along side the array data?

riqwan: I did already, should be in the new commit

packages/pricing/src/services/pricing-module.ts
```diff
@@ -94,13 +94,21 @@ export default class PricingModuleService<
     InjectedDependencies,
     PricingTypes.PriceSetDTO,
     {
-      MoneyAmount: { dto: PricingTypes.MoneyAmountDTO }
+      MoneyAmount: {
+        dto: PricingTypes.MoneyAmountDTO
```
olivermrbl, line 98: very-nit and non-blocking: I am not super fond of the naming here. 

I would prefer something like:
```
{
  common: PricingTypes.MoneyAmountDTO,
  create:  ...,
  update: ...,
}
```

olivermrbl (approved):
Very nice @riqwan!

packages/pricing/src/services/pricing-module.ts
riqwan: I can do the create and update change as a part of this pr, but the dto change would be huge 😄 , can do that as a different PR, i'm currently in merge hell rn.

adrien2p: I believe that the suffix dto was more self describing than create which is an action verb. Also, we are missing the update overload no?

I mean as a consumer, what does create and update represent it is unknown

riqwan: done

packages/utils/src/modules-sdk/abstract-module-service-factory.ts
```diff
@@ -155,17 +171,52 @@ export type AbstractModuleService<
...
+} & {
+  [ModelName in keyof ModelsConfig as UpdateMethodName<
+    ModelsConfig,
+    ModelName
+  >]: {
+    (data: ModelsConfig[ModelName]["update"], sharedContext?: Context): Promise<
+      ModelsConfig[ModelName]["dto"]
+    >
+  }
 }
```
riqwan, line 220: @adrien2p both the update overloads are here

packages/pricing/src/services/pricing-module.ts
olivermrbl: If you are configuring this type argument, I would presume users are aware that they are working with DTOs for the abstract service factory, which is why I think the `_dto` is redundant.

adrien2p: As you prefer, i just think an action verb might be more confusing for the configuration. Maybe adding more documentation to it will help

adrien2p: Maybe something to help would be to rename the template arg to TModelConfigDtos instead of model config. I see that the concept of dto was part of the template args name and it have been removed, Riqwan could you re add the term dto as part of those template args please

adrien2p: Should we add the same shape for the main model dto config then? @riqwan or do we generate them only for the other models for now?

riqwan: I see most modules override the create and update, so i'm leaning towards keep that out for now.

PR #6758, opened 2024-03-20
Feat(user): Invite expiration fix

pKorsholm:
**What** 
- Fix expiration duration for invites

### DIFF
--- a/packages/user/src/services/invite.ts
+++ b/packages/user/src/services/invite.ts
@@ -6,6 +6,7 @@ import {
   MedusaError,
   ModulesSdkUtils,
   arrayDifference,
+  isString,
 } from "@medusajs/utils"
 import jwt, { JwtPayload } from "jsonwebtoken"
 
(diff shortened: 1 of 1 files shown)


sradevski (approved):

packages/user/src/services/invite.ts
```diff
@@ -151,9 +151,6 @@ export default class InviteService<
 
   private generateToken(data: any): string {
     const jwtSecret: string = this.getOption("jwt_secret")
-    const expiresIn: number =
```
sradevski, line 154: Question: Do we not want to have support for configuring the duration?

olivermrbl (approved):
Other than @sradevski's comment, LGTM!

packages/user/src/services/invite.ts
pKorsholm: I initially removed this since we're validating the invite expiration anyways, but I'll re-add it 😄

PR #6760, opened 2024-03-20
Feat(workflows-sdk): additional properties to context

carlos-r-l-rodrigues:
What:
Additional properties added to step context:

workflowId
stepName
action
idempotencyKey
attempt


adrien2p (approved):

srindom:
Nice!

srindom:
Should we maybe move these to the context instead, so we can do this:

```
createStep("myStep", (input, { container, context }) => {
  // context.idempotencyKey
})
```

Right now we have:

```
createStep("myStep", (input, { container, context, idempotencyKey }) => {
})
```

Does that make sense or is there a different purpose for `context`?

PR #6768, opened 2024-03-21
chore: merge money amounts and price set money amounts

riqwan:
what: 

- merges price set money amounts and money amount

### DIFF
--- a/.changeset/ninety-months-allow.md
+++ b/.changeset/ninety-months-allow.md
@@ -1,0 +1,8 @@
+---
+"@medusajs/medusa": patch
+"@medusajs/event-bus-redis": patch
+"@medusajs/modules-sdk": patch
+"@medusajs/types": patch
+---
+
+feat: v2 - add worker mode
--- a/.changeset/small-guests-whisper.md
+++ b/.changeset/small-guests-whisper.md
@@ -1,0 +1,5 @@
+---
+"@medusajs/medusa": patch
+---
+
+fix(medusa): ensure feature flags are loaded properly in migrate command
--- a/.changeset/violet-lizards-help.md
+++ b/.changeset/violet-lizards-help.md
@@ -1,0 +1,8 @@
+---
+"@medusajs/medusa": patch
+"medusa-test-utils": patch
+"@medusajs/types": patch
+"@medusajs/utils": patch
+---
+
+chore: v2 - deprecate extra in favor of driver options
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,5 +1,6 @@
 # All files not owned by other teams must be reviewed by the core team
 *                                               @medusajs/core 
+/.changeset/                                    @medusajs/engineering
 /packages/                                      @medusajs/engineering
 /integration-tests/                             @medusajs/engineering
 /docs-util/                                     @medusajs/docs
--- a/packages/event-bus-redis/src/services/event-bus-redis.ts
+++ b/packages/event-bus-redis/src/services/event-bus-redis.ts
@@ -42,11 +42,14 @@ export default class RedisEventBusService extends AbstractEventBusModuleService 
     })
 
     // Register our worker to handle emit calls
-    new Worker(moduleOptions.queueName ?? "events-queue", this.worker_, {
-      prefix: `${this.constructor.name}`,
-      ...(moduleOptions.workerOptions ?? {}),
-      connection: eventBusRedisConnection,
-    })
+    const shouldStartWorker = moduleDeclaration.worker_mode !== "server"
+    if (shouldStartWorker) {
+      new Worker(moduleOptions.queueName ?? "events-queue", this.worker_, {
+        prefix: `${this.constructor.name}`,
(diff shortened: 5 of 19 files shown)


adrien2p:
There is a lot of simplification and deletion happening ahah

integration-tests/modules/helpers/create-variant-price-set.ts
```diff
@@ -36,15 +36,11 @@ export const createVariantPriceSet = async ({
...
 
   await remoteLink.create({
-    productService: {
-      variant_id: variantId,
-    },
-    pricingService: {
-      price_set_id: priceSet.id,
-    },
+    productService: { variant_id: variantId },
+    pricingService: { price_set_id: priceSet.id },
```
adrien2p: todo: use Modules.*

packages/pricing/src/models/price-set-money-amount.ts
```diff
@@ -103,10 +108,10 @@ export default class PriceSetMoneyAmount {
     fieldName: "price_list_id",
     onDelete: "cascade",
   })
-  price_list_id: string
+  price_list_id: string | null = null
```
adrien2p, line 111: todo: nullable + on delete cascade does not make much sense I believe, the only way would be that you update the item and set the price_list_id to null manually, but I believe that we never do that right otherwise you would end up with orphan data. plus, the on delete cascade means that we remove that tuple if the price list is removed. So I believe we can remove nullable wdyt?

packages/pricing/src/models/price-set-money-amount.ts
```diff
@@ -103,10 +108,10 @@ export default class PriceSetMoneyAmount {
...
 
   @ManyToOne(() => PriceList, { persist: false, nullable: true })
-  price_list?: PriceList
+  price_list: PriceList | null = null
```
adrien2p, line 114: depends on the previous one

adrien2p (commented):
🔥

adrien2p (approved):

packages/pricing/src/models/price-set-money-amount.ts
riqwan: A price set money amount can exist without being attached to a price list. If we remove the nullable, this will become a required property, which is not what we want.

The goal with the cascade here is to delete all the price set money amounts associated to the price list when a price list is deleted.

Unless I misunderstood the suggestion, it won't be possible to make this not nullable.

sradevski (approved):
Nice one!

packages/pricing/src/models/price-set-money-amount.ts
adrien2p: Do you have an example of scenario? Do we have a way to later attach a psma to a price list? For that i suppose you need first to retrieve a psma and then attach it to a price list

packages/pricing/src/models/price-set-money-amount.ts
```diff
@@ -61,6 +65,19 @@ export default class PriceSetMoneyAmount {
   @Property({ columnType: "text", nullable: true })
   title: string | null = null
 
+  @PriceSetMoneyAmountCurrencyCodeIndex.MikroORMIndex()
+  @Property({ columnType: "text" })
+  currency_code: string
+
+  @Property({ columnType: "numeric", serializer: Number })
+  amount: number
```
olivermrbl, line 73: suggestion: use BigNumber

olivermrbl (approved):
loving these PRs!

packages/pricing/src/models/price-set-money-amount.ts
adrien2p: Already suggested and it will be part a separate pr

packages/pricing/src/models/price-set-money-amount.ts
riqwan: [solved over chat] result: nothing to do

packages/pricing/src/models/price-set-money-amount.ts
olivermrbl: Got it

packages/pricing/src/models/price-set-money-amount.ts
olivermrbl: All regular product variant prices fall in this category no? That is, they exist without a price list id.

packages/pricing/src/models/price-set-money-amount.ts
riqwan: This one is coming 2 PRs down, got one more biggie before it. 😄

olivermrbl: lfg! 😄

packages/pricing/src/models/price-set-money-amount.ts
riqwan: Yes exactly 💯

adrien2p: what got me is that it is only doable by hand in the db to re attach to a price list. if i correctly undestood

PR #6769, opened 2024-03-21
feat: List products middleware

olivermrbl:
Add cross-module filter middlewares for list-products
- sales channel
- price lists

Both are required for admin v2 flows.


integration-tests/api/__tests__/admin/product.js
```diff
@@ -1104,6 +1120,128 @@ medusaIntegrationTestRunner({
...
+                }
+              )
+
+              return { salesChannel, product }
+            },
+            async () => {
+              const salesChannel = await scService.create({
+                name: "Test channel",
+                description: "Lorem Ipsum",
+              })
```
riqwan, line 1222: sugg: might be good to add a second product and sales channel to scrape the false positive chance of the test.

packages/medusa/src/api-v2/admin/products/middlewares.ts
```diff
@@ -31,6 +35,8 @@ export const adminProductRoutesMiddlewares: MiddlewareRoute[] = [
         AdminGetProductsParams,
         QueryConfig.listProductQueryConfig
       ),
+      maybeApplySalesChannelsFilter(),
+      maybeApplyPriceListsFilter(),
```
riqwan, line 39: praise: 🥇

riqwan (approved):
🥇

integration-tests/api/__tests__/admin/product.js
```diff
@@ -1104,6 +1120,128 @@ medusaIntegrationTestRunner({
...
+
+              await remoteLink.create({
```
sradevski, line 1224: It's fine to keep this for now since we are still working on the product module, but ideally we try to achieve everything in the API tests through the public interfaces. For example, you'd do an update on the `product` to set the sales channel, rather than creating the link within the test.

sradevski (approved):
Looks good, and

integration-tests/api/__tests__/admin/product.js
```diff
@@ -27,52 +32,16 @@ let {
 
 jest.setTimeout(50000)
 
-const productFixture = {
```
olivermrbl, line 30: @sradevski, had to move the fixture inside the `moduleIntegrationTestRunner`. When passing the environment variable directly to the runner instead of setting it in `.env`, it is not globally available at this point in time.

sradevski: Sounds good 👍

PR #6770, opened 2024-03-21
docs-util: added AI generator

shahednasser:
## What

Adds an AI generator to the docblock tool that uses OpenAI.

The generator at the moment only generates examples for functions when the `--generate-examples` option is provided.

## Note

I've included the generated examples of the `IOrderModuleService` as a reference of the type of result provided by the AI generator, with minor tweeks I've made. I haven't made any changes to descriptions in that file.


olivermrbl (approved):
🚀

PR #6771, opened 2024-03-21
fix(medusa): migrate with feature flag

srindom:


packages/medusa/src/commands/migrate.js
```diff
@@ -91,6 +90,7 @@ const main = async function ({ directory }) {
         { registerInContainer: false }
       )
     } else {
+      const dataSource = await getDataSource(directory)
```
riqwan, line 93: todo: this is needed on the `revert` and `show` args as well

riqwan (approved):

PR #6772, opened 2024-03-21
chore: Deprecate extra in favor of driver options

adrien2p:
**What**
Deprecate databse_extra in favor of database_driver_options

### DIFF
--- a/packages/types/src/common/config-module.ts
+++ b/packages/types/src/common/config-module.ts
@@ -520,6 +520,25 @@ export type ProjectConfigOptions = {
    * ```
    */
   jobs_batch_size?: number
+
+  /**
+   * Configure the application's worker mode. The default value is `shared`.
+   * - Use `shared` if you want to run the application in a single process.
+   * - Use `worker` if you want to run the a worker process only.
+   * - Use `server` if you want to run the application server only.
+   *
+   * @example
+   * ```js title="medusa-config.js"
+   * module.exports = {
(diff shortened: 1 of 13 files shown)


packages/types/src/common/config-module.ts
```diff
@@ -342,6 +342,8 @@ export type ProjectConfigOptions = {
    *   // ...
    * }
    * ```
+   *
+   * @deprecated use `database_driver_options` instead
```
olivermrbl: question: For Typeorm connections, should this be backward compatible?

adrien2p: The typeorm datasource is not loaded if the v2 flag is enabled, so in theory this works only for v1 right?

olivermrbl: My thinking was that if we release this tomorrow, `database_extra` will be denoted as deprecated, and v1 users will attempt to use `database_driver_options` instead, which is incompatible with the Typeorm datasource. I might be overthinking here, let me know if it makes sense.

srindom: think it's fine to leave out the deprecation warning for now - when we get to v2 we should probably remove it altogether

adrien2p: ahah you are right sorry, I ll remove that flag

adrien2p:
@srindom should we slap an automerge on this one?

srindom (approved):

PR #6773, opened 2024-03-21
Update example-logged-in-user.mdx

Deveosys:
Adds the ability to find the current user by API Token


Deveosys:
Wait, 

My edit is not needed, instead we should add the need for authenticate middleware in existing doc. : 

```typescript
import { authenticate, type MiddlewaresConfig, type User, type UserService } from "@medusajs/medusa";

// ...

export const config: MiddlewaresConfig = {
    routes: [
       {
          matcher: /^\/admin\/(?!auth).*/,
          middlewares: [authenticate(), registerLoggedInUser],
       },
    ],
};
```

PR #6774, opened 2024-03-21
docs-util: fix clean command removing used schemas

shahednasser:
This PR fixes a bug in the `clean:oas` command of the docblock-generator tool that deletes used schemas. The fix is to check all nested schema types and not just `properties` (`oneOf`, `allOf`, etc...) recursively.

I regenerated the V2 OAS as there was an error in them due to missing schemas that were removed previously by the clean command.


shahednasser:

olivermrbl (approved):

PR #6775, opened 2024-03-21
feat: shard modules and API integration tests

riqwan:
what:

- runs module and api integration tests in matrix strategy
  - v28 of jest comes with sharding support, which we can utilize when we upgrade
- splits the tests by number of matrix jobs and run them in parallel
  - This introduced some flakiness in some specs, but couldn't reproduce locally. Those have been skipped for now.
- uses swc/jest for added performance
  - Locally, a chunk took 90 seconds with babel and 30 seconds with swc. 
  - This translated to 2 mins saved per shard in CI, but haven't tested this enough.

### DIFF
--- a/.github/workflows/action.yml
+++ b/.github/workflows/action.yml
@@ -225,7 +225,6 @@ jobs:
   integration-tests-api:
     if: ${{ always() }}
     runs-on: ubuntu-latest
-    name: Check status of matrix jobs
     needs: integration-tests-api-matrix
     steps:
       - run: exit 1
@@ -353,7 +352,6 @@ jobs:
   integration-tests-modules:
     if: ${{ always() }}
     runs-on: ubuntu-latest
-    name: Check status of matrix jobs
     needs: integration-tests-modules-matrix
     steps:
       - run: exit 1


adrien2p:
The flakiness can be due to the fact that for example all api tests running on tue same worker are using the same db, migrating to medusaTestRunner should help fixing that in the long term. If it happens in the module then we need to investigate cause if they are using our runner there is one db per suite

riqwan:
@adrien2p thanks that makes sense, I’ll try migrating the flaky ones to see if that fixes it.

olivermrbl:
> If it happens in the module then we need to investigate cause if they are using our runner there is one db per suite

@adrien2p, one of the skipped tests (cart.js) is using the `moduleIntegrationTestRunner`, so we might need to dig into this one.

adrien2p:
Yes ill have a look

sradevski (approved):
Amazing stuff, such an improvement 👍

integration-tests/modules/__tests__/cart/store/carts.spec.ts
```diff
@@ -65,7 +65,7 @@ medusaIntegrationTestRunner({
       })
 
       describe("POST /store/carts", () => {
-        it("should create a cart", async () => {
+        it.skip("should create a cart", async () => {
```
sradevski, line 68: Let's just follow-up and re-enable these, but it shouldn't be a blocker IMO

adrien2p:
@riqwan I am trying to investigate the cart just so you know

integration-tests/modules/__tests__/cart/store/carts.spec.ts
adrien2p: yeah it is not for that pr, I am investigating it but if there is an issue it is separate

adrien2p (approved):

olivermrbl (approved):
🏆

integration-tests/modules/__tests__/cart/store/carts.spec.ts
olivermrbl: Agree, would also just get this in now, and resolve (future) flaky tests in follow-up PRs

riqwan:
Wasn't expecting an approval yet. 😄  I had some more things to do here, but it might be a good idea to take this out for a spin. 

We will still be limited by the unit tests, it takes about 20 mins. I wanted to split that into 3 buckets as well, but I need to brush up on my bash skills and configure turbo to filter them. But that can come as a follow up! 🤞🏻

pKorsholm (approved):
Amazing 😍

PR #6778, opened 2024-03-21
Feat(orchestration,workflows-sdk,core-flows): workflow cancel

carlos-r-l-rodrigues:


packages/workflows-sdk/src/utils/composer/create-step.ts
```diff
@@ -167,8 +167,7 @@ function applyStep<
             const invokeResult =
               stepOutput?.__type ===
               OrchestrationUtils.SymbolWorkflowStepResponse
-                ? stepOutput.compensateInput &&
-                  JSON.parse(JSON.stringify(stepOutput.compensateInput))
+                ? stepOutput.compensateInput && stepOutput.compensateInput
```
adrien2p: Suggestion : maybe do a deepCopy to be sure we dont have references issues unless it is expected ?

adrien2p (approved):

packages/workflows-sdk/src/utils/composer/create-step.ts
carlos-r-l-rodrigues: yeah, good call. Will change both to use a deepCopy

PR #6713, opened 2024-03-15
feat(dashboard) admin 3.0 return creation

fPolic:
**What**
- request return flow

PR #6782, opened 2024-03-21
feat(dashboard) admin 3.0 claims

fPolic:
**What**
- create a claim flow

---

**Depends on**
- [x] #6713


olivermrbl:
Shipping for replacement items should not be return shipping:

image

olivermrbl:
The subtitle is incorrectly positioned for "Shipping for replacement items":

image

olivermrbl:
Question: What do we plan to add here and when?

image

fPolic:
> Question: What do we plan to add here and when?

I've added a placeholder UI, the other issues should be resolved as well.

PR #6785, opened 2024-03-22
chore: add TSDocs to the API Key Module

shahednasser:
Add TSDocs to the API Key Module's resources

### DIFF
--- a/packages/types/src/api-key/common/api-key.ts
+++ b/packages/types/src/api-key/common/api-key.ts
@@ -53,7 +53,8 @@ export interface ApiKeyDTO {
   created_at: Date
 
   /**
-   * Who revoked the API key.
+   * Who revoked the API key. For example,
+   * the ID of the user that revoked it.
    */
   revoked_by: string | null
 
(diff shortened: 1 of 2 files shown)


packages/types/src/api-key/common/api-key.ts
```diff
@@ -1,23 +1,90 @@
...
+   */
   created_by: string
+
+  /**
+   * The date the API key was created.
+   */
   created_at: Date
+
+  /**
+   * Who revoked the API key.
```
sradevski: NIT: I think it's nice to specify what exactly this points to (eg. "the ID of the user that revoked the API key")

packages/types/src/api-key/mutations/api-key.ts
```diff
@@ -1,24 +1,74 @@
...
   id?: string
+
+  /**
+   * The title of the API key.
+   */
   title?: string
+
+  /**
+   * The type of the API key. It's only
+   * usable and required when creating an API key.
```
sradevski: Can we rephrase this to `Required only when creating an API key`? I think the usable part is implied

sradevski (approved):

packages/types/src/api-key/common/api-key.ts
shahednasser: I was thinking of that too, but also felt like it's somewhat limiting. i.e. if someone is using the module on their own, they can use something other than user ID such as email. However, it makes sense to add an example so I'll do that 👍🏻

PR #6791, opened 2024-03-22
feat(medusa, core-flows): add retrieve stock location endpoint to api-v2

pKorsholm:


carlos-r-l-rodrigues (approved):

integration-tests/modules/__tests__/stock-location/index.spec.ts
```diff
@@ -52,5 +52,46 @@ medusaIntegrationTestRunner({
...
+
+        expect(response.status).toEqual(200)
+        expect(response.data.stock_location).toEqual(
+          expect.objectContaining({ id: locationId, ...location })
+        )
+      })
+
+      it("should get a stock location", async () => {
+        let error
+        await api
```
olivermrbl: comment: Same comment as seen on other PR. Rest API tests should live in `integration-tests/api`. 

Let's move forward with this PR and migrate later.

packages/medusa/src/api-v2/admin/stock-locations/[id]/route.ts
```diff
@@ -0,0 +1,31 @@
...
+} from "@medusajs/utils"
+import { MedusaRequest, MedusaResponse } from "../../../../types/routing"
+
+import { MedusaError } from "@medusajs/utils"
+
+export const GET = async (req: MedusaRequest, res: MedusaResponse) => {
+  const { id } = req.params
+  const remoteQuery = req.scope.resolve(ContainerRegistrationKeys.REMOTE_QUERY)
+
+  const [stock_location] = await remoteQuery(
```
Comment (olivermrbl) [inline packages/medusa/src/api-v2/admin/stock-locations/[id]/route.ts:14]: comment: @carlos-r-l-rodrigues, your idea of adding a "throw-if-not-found" mechanism to remote query could make sense. I think in most endpoints, we are not checking for the existence of the retrieved resource when doing a get by ID.

olivermrbl (approved):

packages/medusa/src/api-v2/admin/stock-locations/[id]/route.ts
Comment (olivermrbl) [reply packages/medusa/src/api-v2/admin/stock-locations/[id]/route.ts:14]: Or alternatively, we add the `if (!result) {` to all get one endpoints, but I think having this baked in would be nice.

olivermrbl (approved):

olivermrbl:
@pKorsholm, can you rebase and change base branch to develop before merging?

PR #6792, opened 2024-03-22
feat: Add `successRedirectUrl` to auth options

olivermrbl:
**What**

Add `successRedirectUrl` to provider options. The option can be used to configure a URL for the callback to redirect to.

For example, you could have the following flow:
- `POST /auth/admin/google`
- Redirect to Google auth
- `POST /auth/admin/google/callback`
  - Validate callback
- Redirect to admin with token in URL e.g. `http://localhost:5173/login?access_token=...`
- Decode token client-side
- `POST /admin/users` with auth user from token

If no `successRedirectUrl` is configured, we respond with the token from the callback.

### DIFF
--- a/packages/medusa/src/api-v2/auth/[scope]/[auth_provider]/callback/route.ts
+++ b/packages/medusa/src/api-v2/auth/[scope]/[auth_provider]/callback/route.ts
@@ -27,8 +27,7 @@ export const GET = async (req: MedusaRequest, res: MedusaResponse) => {
   if (success) {
     const { jwt_secret } = req.scope.resolve("configModule").projectConfig
 
-    const user = { ...authUser }
-    const token = jwt.sign(user, jwt_secret)
+    const token = jwt.sign(authUser, jwt_secret)
 
     if (successRedirectUrl) {
       const url = new URL(successRedirectUrl!)
--- a/packages/auth/src/services/auth-user.ts
+++ b/packages/auth/src/services/auth-user.ts
@@ -14,6 +14,7 @@ import {
 import { AuthUser } from "@models"
 
 type InjectedDependencies = {
+  baseRepository: DAL.RepositoryService
   authUserRepository: DAL.RepositoryService
 }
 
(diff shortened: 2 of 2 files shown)


packages/medusa/src/api-v2/auth/[scope]/[auth_provider]/callback/route.ts
```diff
@@ -20,18 +20,26 @@ export const GET = async (req: MedusaRequest, res: MedusaResponse) => {
...
+    const user = { ...authUser }
+    const token = jwt.sign(user, jwt_secret)
+
+    if (successRedirectUrl) {
+      const url = new URL(successRedirectUrl!)
+      url.searchParams.append("auth_token", token)
+
+      const redirectUrl = `${url}`
+
+      return res.redirect(redirectUrl.toString())
```
Comment (carlos-r-l-rodrigues) [inline packages/medusa/src/api-v2/auth/[scope]/[auth_provider]/callback/route.ts]:
```suggestion
      return res.redirect(redirectUrl)
```
nit

carlos-r-l-rodrigues (approved):

packages/medusa/src/api-v2/auth/[scope]/[auth_provider]/callback/route.ts
```diff
@@ -20,18 +20,26 @@ export const GET = async (req: MedusaRequest, res: MedusaResponse) => {
...
-    return
-  }
+  const { success, error, authUser, successRedirectUrl } = authResult
 
   if (success) {
     const { jwt_secret } = req.scope.resolve("configModule").projectConfig
-    const token = jwt.sign(authUser, jwt_secret)
-    return res.status(200).json({ token })
+
+    const user = { ...authUser }
```
Comment (sradevski) [inline packages/medusa/src/api-v2/auth/[scope]/[auth_provider]/callback/route.ts]: Why do we need to destructure the authUser here?

packages/medusa/src/api-v2/auth/[scope]/[auth_provider]/callback/route.ts
Comment (sradevski) [reply packages/medusa/src/api-v2/auth/[scope]/[auth_provider]/callback/route.ts]: Or just do `url.toString()` which I think is better.

sradevski (approved):

packages/medusa/src/api-v2/auth/[scope]/[auth_provider]/callback/route.ts
Comment (olivermrbl) [reply packages/medusa/src/api-v2/auth/[scope]/[auth_provider]/callback/route.ts]: The payload to `jwt.sign` needs to be a plain object and it was being returned as an entity object here. 

I revisited the approach and decided to serialize the auth user response at the service level in 8e8681a50be0d70a9dfdcddafd0238c0a1cc1ac3, so we don't need this dirty destructuring.

Comment (sradevski) [reply packages/medusa/src/api-v2/auth/[scope]/[auth_provider]/callback/route.ts]: @olivermrbl ah I see, yeah that sound better indeed

sradevski (approved):

PR #6797, opened 2024-03-22
Feat: draft order api

carlos-r-l-rodrigues:
- Create draft-order workflows + API endpoint
- Worfklows SDK - prevent workflows using the same name


packages/medusa/src/api-v2/admin/draft-orders/validators.ts
```diff
@@ -0,0 +1,181 @@
...
+  items?: Item[]
+
+  @IsString()
+  region_id: string
+
+  @IsArray()
+  @IsOptional()
+  @Type(() => Discount)
+  @ValidateNested({ each: true })
+  discounts?: Discount[]
```
olivermrbl: suggestion: `promo_codes?: string[]`

packages/medusa/src/api-v2/admin/draft-orders/validators.ts
```diff
@@ -0,0 +1,181 @@
...
+
+  @IsObject()
+  @IsOptional()
+  data?: Record<string, unknown> = {}
+
+  @IsNumber()
+  amount: BigNumberInput
+}
+
+class Discount {
```
olivermrbl: sugggestion: rm

packages/medusa/src/api-v2/admin/draft-orders/validators.ts
```diff
@@ -0,0 +1,181 @@
...
+
+  @IsOptional()
+  @ValidateNested({ each: true })
+  @Type(() => AdminGetOrdersParams)
+  $or?: AdminGetOrdersParams[]
+}
+
+enum Status {
+  completed = "completed",
+}
```
olivermrbl, line 59: question: Does this make sense to keep?

packages/medusa/src/api-v2/admin/draft-orders/route.ts
```diff
@@ -0,0 +1,76 @@
...
+  }
+
+  const { result, errors } = await createOrdersWorkflow(req.scope).run({
+    input: workflowInput,
+    throwOnError: false,
+  })
+
+  if (Array.isArray(errors) && errors[0]) {
+    throw errors[0].error
+  }
```
olivermrbl, line 71: suggestion: Refetch draft order with `remoteQuery`

packages/medusa/src/api-v2/admin/draft-orders/route.ts
```diff
@@ -0,0 +1,76 @@
...
+      skip: req.listConfig.skip,
+      take: req.listConfig.take,
+    },
+    fields: defaultAdminOrderFields,
+  })
+
+  const { rows: orders, metadata } = await remoteQuery(queryObject)
+
+  res.json({
+    orders,
```
olivermrbl: thought: Would it make sense to still return these as `draft_orders`? If not, this is breaking, so should be documented.

packages/medusa/src/api-v2/admin/draft-orders/route.ts
```diff
@@ -0,0 +1,76 @@
...
+  const { result, errors } = await createOrdersWorkflow(req.scope).run({
+    input: workflowInput,
+    throwOnError: false,
+  })
+
+  if (Array.isArray(errors) && errors[0]) {
+    throw errors[0].error
+  }
+
+  res.status(200).json({ order: result })
```
olivermrbl: thought: Would it make sense to still return this as `draft_order`? If not, this is breaking, so should be documented.

packages/medusa/src/api-v2/admin/draft-orders/middlewares.ts
```diff
@@ -0,0 +1,42 @@
...
+      transformQuery(
+        AdminGetOrdersOrderParams,
+        QueryConfig.retrieveTransformQueryConfig
+      ),
+    ],
+  },
+  {
+    method: ["POST"],
+    matcher: "/admin/draft-orders",
+    middlewares: [transformBody(AdminPostDraftOrdersReq)],
```
olivermrbl: suggestion: Add `transformQuery(...)`

packages/medusa/src/api-v2/admin/draft-orders/[id]/route.ts
```diff
@@ -0,0 +1,21 @@
...
+  const variables = { id: req.params.id }
+
+  const queryObject = remoteQueryObjectFromString({
+    entryPoint: "order",
+    variables,
+    fields: defaultAdminOrderFields,
+  })
+
+  const [order] = await remoteQuery(queryObject)
+  res.status(200).json({ order })
```
Comment (olivermrbl) [inline packages/medusa/src/api-v2/admin/draft-orders/[id]/route.ts]: thought: Same as other comments. Would consider `draft_order` as the response.

integration-tests/modules/__tests__/order/draft-order.spec.ts
```diff
@@ -0,0 +1,253 @@
...
+          ],
+        }
+
+        const response = await api.post(
+          "/admin/draft-orders",
+          payload,
+          adminHeaders
+        )
+
+        expect(response.status).toEqual(200)
```
adrien2p, line 371: q: don't we expect anything here to check the data?

packages/core-flows/src/definition/cart/steps/confirm-inventory.ts
```diff
@@ -22,10 +22,10 @@ export const confirmInventoryStep = createStep(
     )
 
     // TODO: Should be bulk
-    const promises = data.items.map((item) => {
+    const promises = data.items.map(async (item) => {
       const itemQuantity = item.required_quantity * item.quantity
 
-      return inventoryService.confirmInventory(
+      return await inventoryService.confirmInventory(
```
adrien2p, line 28: 🙏 praise

packages/core-flows/src/order/steps/get-item-tax-lines.ts
```diff
@@ -0,0 +1,141 @@
...
+  TaxCalculationContext,
+  TaxableItemDTO,
+  TaxableShippingDTO,
+} from "@medusajs/types"
+import { MedusaError } from "@medusajs/utils"
+import { StepResponse, createStep } from "@medusajs/workflows-sdk"
+
+interface StepInput {
+  order: OrderWorkflowDTO
+  items: OrderLineItemDTO[]
```
adrien2p, line 18: suggestion: We are expecting an entire item here, but only use a few props from it, should we have a separate DTO for this

adrien2p: Same for the shipping methods

packages/core-flows/src/order/steps/get-item-tax-lines.ts
```diff
@@ -0,0 +1,141 @@
...
+    return null
+  }
+
+  const customer = order.customer
+    ? {
+        id: order.customer.id,
+        email: order.customer.email,
+        customer_groups: order.customer.groups?.map((g) => g.id) || [],
+      }
+    : undefined
```
adrien2p:
```suggestion
  const customer = order.customer && {
        id: order.customer.id,
        email: order.customer.email,
        customer_groups: order.customer.groups?.map((g) => g.id) || [],
      }
```

packages/core-flows/src/order/steps/get-item-tax-lines.ts
```diff
@@ -0,0 +1,141 @@
...
+        id: item.id,
+        product_id: item.product_id!,
+        product_name: item.variant_title,
+        product_sku: item.variant_sku,
+        product_type: item.product_type,
+        product_type_id: item.product_type,
+        quantity: item.quantity,
+        unit_price: item.unit_price,
+        currency_code: order.currency_code,
+      } as any)
```
adrien2p: Q: shouldnt it be `TaxableItemDTO`

packages/core-flows/src/order/steps/get-item-tax-lines.ts
```diff
@@ -0,0 +1,141 @@
...
+  shippingMethods: OrderShippingMethodDTO[]
+): TaxableShippingDTO[] {
+  return shippingMethods.map(
+    (shippingMethod) =>
+      ({
+        id: shippingMethod.id,
+        shipping_option_id: shippingMethod.shipping_option_id!,
+        unit_price: shippingMethod.amount,
+        currency_code: order.currency_code,
+      } as any)
```
adrien2p: same question but for `TaxableShippingDTO` :)

packages/core-flows/src/order/steps/get-item-tax-lines.ts
```diff
@@ -0,0 +1,141 @@
...
+
+    const shippingMethodsTaxLines = (await taxService.getTaxLines(
+      normalizeLineItemsForShipping(order, shippingMethods),
+      taxContext
+    )) as ShippingTaxLineDTO[]
+
+    return new StepResponse({
+      lineItemTaxLines,
+      shippingMethodsTaxLines,
+    })
```
adrien2p: nit:

```suggestion
    const taxService = container.resolve<ITaxModuleService>(
      ModuleRegistrationName.TAX
    )

    const taxContext = normalizeTaxModuleContext(order, forceTaxCalculation)
    
    const stepResponseData = {
      lineItemTaxLines: [] as ItemTaxLineDTO[],
      shippingMethodsTaxLines: [] as ShippingTaxLineDTO[]
    }

    if (!taxContext) {
      return new StepResponse(stepResponseData)
    }

    stepResponseData.lineItemTaxLines = (await taxService.getTaxLines(
      normalizeLineItemsForTax(order, items),
      taxContext
    )) as ItemTaxLineDTO[]

    stepResponseData.shippingMethodsTaxLines = (await taxService.getTaxLines(
      normalizeLineItemsForShipping(order, shippingMethods),
      taxContext
    )) as ShippingTaxLineDTO[]

    return new StepResponse(stepResponseData)
```

packages/core-flows/src/order/steps/set-tax-lines-for-items.ts
```diff
@@ -0,0 +1,128 @@
...
+  CreateShippingMethodTaxLineDTO,
+  IOrderModuleService,
+  ItemTaxLineDTO,
+  OrderDTO,
+  ShippingTaxLineDTO,
+} from "@medusajs/types"
+import { StepResponse, createStep } from "@medusajs/workflows-sdk"
+
+interface StepInput {
+  order: OrderDTO
```
adrien2p, line 14: suggestion: We are only using the id, should we update the type to reflect that and not force to pass an entire order dto?

adrien2p: Same suggestion in general for all inputs, the idea being that we should ask for what we need and not more to prevent forcing the consumer to retrieve all the data that we do not need

packages/core-flows/src/order/steps/set-tax-lines-for-items.ts
```diff
@@ -0,0 +1,128 @@
...
+  async (data: StepInput, { container }) => {
+    const { order, item_tax_lines, shipping_tax_lines } = data
+    const orderService = container.resolve<IOrderModuleService>(
+      ModuleRegistrationName.ORDER
+    )
+
+    const getShippingTaxLinesPromise =
+      await orderService.listShippingMethodTaxLines({
+        shipping_method_id: shipping_tax_lines.map((t) => t.shipping_line_id),
+      })
```
adrien2p, line 31: discussion: what result do we receive if we provide an empty array? I would expect nothing, but just want to double check with you

packages/core-flows/src/order/steps/set-tax-lines-for-items.ts
```diff
@@ -0,0 +1,128 @@
...
+
+    const getItemTaxLinesPromise = await orderService.listLineItemTaxLines({
+      item_id: item_tax_lines.map((t) => t.line_item_id),
+    })
+
+    const itemsTaxLinesData = normalizeItemTaxLinesForOrder(item_tax_lines)
+    const setItemTaxLinesPromise = itemsTaxLinesData.length
+      ? orderService.setLineItemTaxLines(order.id, itemsTaxLinesData)
+      : 0
```
adrien2p: suggestion: shouldn't it be `Promise.resolve([])`?

packages/core-flows/src/order/steps/set-tax-lines-for-items.ts
```diff
@@ -0,0 +1,128 @@
...
+
+    const shippingTaxLinesData =
+      normalizeShippingTaxLinesForOrder(shipping_tax_lines)
+    const setShippingTaxLinesPromise = shippingTaxLinesData.length
+      ? await orderService.setShippingMethodTaxLines(
+          order.id,
+          shippingTaxLinesData
+        )
+      : 0
```
adrien2p: same question as above

packages/core-flows/src/order/steps/set-tax-lines-for-items.ts
adrien2p: suggestion: maybe checking item_tax_lines.length?

packages/core-flows/src/order/steps/set-tax-lines-for-items.ts
adrien2p: suggestion: maybe checking shipping_tax_lines.length?

packages/core-flows/src/order/steps/set-tax-lines-for-items.ts
```diff
@@ -0,0 +1,128 @@
...
+
+    const [existingShippingMethodTaxLines, existingLineItemTaxLines] =
+      await Promise.all([
```
adrien2p: suggestion: should we use promiseAll?

packages/core-flows/src/order/steps/set-tax-lines-for-items.ts
```diff
@@ -0,0 +1,128 @@
...
+        getShippingTaxLinesPromise,
+        getItemTaxLinesPromise,
+        setItemTaxLinesPromise,
+        setShippingTaxLinesPromise,
+      ])
+
+    return new StepResponse(null, {
```
adrien2p: suggestion: for those steps where we return null, the intention I believe is to return nothing, should we use void 0 instead?

packages/core-flows/src/order/utils/prepare-custom-line-item-data.ts
```diff
@@ -0,0 +1,29 @@
...
+  metadata?: Record<string, any>
+  unitPrice: BigNumberInput
+  variant: {
+    title: string
+    sku?: string
+    barcode?: string
+  }
+}
+
+export function prepareCustomLineItemData(data: Input) {
```
adrien2p: todo: adding type to the output

packages/core-flows/src/order/workflows/create-orders.ts
```diff
@@ -0,0 +1,203 @@
...
+} from "../../definition/cart"
+import { prepareConfirmInventoryInput } from "../../definition/cart/utils/prepare-confirm-inventory-input"
+import { prepareLineItemData } from "../../definition/cart/utils/prepare-line-item-data"
+import { createOrdersStep, updateOrderTaxLinesStep } from "../steps"
+import { prepareCustomLineItemData } from "../utils/prepare-custom-line-item-data"
+
+export const createOrdersWorkflowId = "create-orders"
+export const createOrdersWorkflow = createWorkflow(
+  createOrdersWorkflowId,
+  (input: WorkflowData<CreateOrderDTO>): WorkflowData<OrderDTO> => {
```
adrien2p, line 27: discussion: I believe that we usually export an interface from the types package for the input, my thinking is that if in the future people add steps, they might ask for more data, and they should be able to use the current interface and expand it. wdyt? not necessarely now but just throwing it here

packages/core-flows/src/order/workflows/create-orders.ts
```diff
@@ -0,0 +1,203 @@
...
+            quantity: item.quantity as number,
+            metadata: item?.metadata ?? {},
+          })
+        }
+
+        return prepareLineItemData({
+          variant: variant,
+          unitPrice: MathBN.max(
+            0,
+            item.unit_price ?? data.priceSets[item.variant_id!]?.amount
```
adrien2p: Q: don't we want the min between those two?

packages/link-modules/src/definitions/order-promotion.ts
```diff
@@ -0,0 +1,55 @@
...
+    },
+    {
+      serviceName: Modules.PROMOTION,
+      relationship: {
+        serviceName: LINKS.OrderPromotion,
+        primaryKey: "promotion_id",
+        foreignKey: "id",
+        alias: "promotion_link",
+      },
+    },
```
adrien2p, line 53: suggestion: should we add some fieldAliases?

packages/medusa/src/api-v2/admin/draft-orders/[id]/route.ts
```diff
@@ -0,0 +1,21 @@
...
+
+export const GET = async (req: MedusaRequest, res: MedusaResponse) => {
+  const remoteQuery = req.scope.resolve(ContainerRegistrationKeys.REMOTE_QUERY)
+
+  const variables = { id: req.params.id }
+
+  const queryObject = remoteQueryObjectFromString({
+    entryPoint: "order",
+    variables,
+    fields: defaultAdminOrderFields,
```
Comment (adrien2p) [inline packages/medusa/src/api-v2/admin/draft-orders/[id]/route.ts:16]: suggestion: req.remoteQueryConfig.fields

packages/medusa/src/api-v2/admin/draft-orders/middlewares.ts
adrien2p: link to my previous suggestion too

packages/medusa/src/api-v2/admin/draft-orders/query-config.ts
```diff
@@ -0,0 +1,23 @@
+export const defaultAdminOrderRelations = []
+export const allowedAdminOrderRelations = []
+export const defaultAdminOrderFields = [
+  "id",
+  "version",
+  "items",
+  "items.detail",
+  "metadata",
+  "created_at",
+  "updated_at",
```
adrien2p:
```suggestion
  "id",
  "version",
  "*items",
  "*items.detail",
  "metadata",
  "created_at",
  "updated_at",
```

packages/medusa/src/api-v2/admin/draft-orders/query-config.ts
```diff
@@ -0,0 +1,23 @@
...
+]
+
+export const retrieveTransformQueryConfig = {
+  defaultFields: defaultAdminOrderFields,
+  defaultRelations: defaultAdminOrderRelations,
+  allowedRelations: allowedAdminOrderRelations,
+  isList: false,
```
adrien2p, line 50:
```suggestion
export const retrieveTransformQueryConfig = {
  defaults: defaultAdminOrderFields,
  isList: false,
```

packages/medusa/src/api-v2/admin/draft-orders/route.ts
```diff
@@ -0,0 +1,76 @@
...
+      filters: {
+        ...req.filterableFields,
+        status: OrderStatus.DRAFT,
+      },
+      order: req.listConfig.order,
+      skip: req.listConfig.skip,
+      take: req.listConfig.take,
+    },
+    fields: defaultAdminOrderFields,
+  })
```
adrien2p:
```suggestion
  const queryObject = remoteQueryObjectFromString({
    entryPoint: "order",
    variables: {
      filters: {
        ...req.filterableFields,
        status: OrderStatus.DRAFT,
      },
      ...req.remoteQueryConfig.pagination
    },
    fields: req.remoteQueryConfig.fields,
  })
```

packages/medusa/src/api-v2/admin/draft-orders/validators.ts
```diff
@@ -0,0 +1,181 @@
...
+  // Additional filters from BaseFilterable
+  @IsOptional()
+  @ValidateNested({ each: true })
+  @Type(() => AdminGetOrdersParams)
+  $and?: AdminGetOrdersParams[]
+
+  @IsOptional()
+  @ValidateNested({ each: true })
+  @Type(() => AdminGetOrdersParams)
+  $or?: AdminGetOrdersParams[]
```
adrien2p, line 54: Q: do we really want that as part of the http filters? eventually OperatorMap on id and name?
If we really want to provide that, we need to rework it a bit I believe, at the moment each item of $and and $or would include the pagination

packages/order/src/joiner-config.ts
```diff
@@ -22,5 +22,12 @@ export const joinerConfig: ModuleJoinerConfig = {
   serviceName: Modules.ORDER,
   primaryKeys: ["id"],
   linkableKeys: LinkableKeys,
-  alias: [],
+  alias: [
+    {
+      name: ["order", "orders"],
+      args: {
+        entity: "Order",
```
adrien2p: suggestion: Order.name?

packages/order/src/models/line-item-adjustment.ts
```diff
@@ -28,7 +22,7 @@ export default class LineItemAdjustment extends AdjustmentLine {
     entity: () => LineItem,
     columnType: "text",
     fieldName: "item_id",
-    cascade: [Cascade.REMOVE],
+    onDelete: "cascade",
```
adrien2p, line 25: todo: I believe I haven't seen that change in the migration, could you double check :)

packages/order/src/models/order-change-action.ts
```diff
@@ -50,7 +49,7 @@ export default class OrderChangeAction {
     entity: () => Order,
     columnType: "text",
     fieldName: "order_id",
-    cascade: [Cascade.REMOVE],
+    onDelete: "cascade",
```
adrien2p, line 52: could you check for all those cascade that they appear in the migration 👍

packages/order/src/services/order-module-service.ts
```diff
@@ -694,20 +694,25 @@ export default class OrderModuleService<
   ): Promise<void>
   async removeLineItems(itemIds: string, sharedContext?: Context): Promise<void>
   async removeLineItems(
-    selector: Partial<OrderTypes.OrderLineItemDTO>,
+    selector: Partial<OrderTypes.FilterableOrderLineItemProps>,
     sharedContext?: Context
   ): Promise<void>
 
   @InjectTransactionManager("baseRepository_")
   async removeLineItems(
```
adrien2p: suggestion: should we name it `deleteLineItems` to keep consistency with what is generated under the hood, otherwise both method will be available and sometimes it will be remove or delete

packages/types/src/order/common.ts
```diff
@@ -443,6 +452,11 @@ export interface OrderLineItemDTO extends OrderLineItemTotalsDTO {
    * The date when the order line item was last updated.
    */
   updated_at: Date
+
+  /**
+   * Holds custom data in key-value pairs.
+   */
+  metadata?: Record<string, unknown> | null
```
adrien2p, line 459: nit: you can remove `?`

adrien2p: on the DTO itself that represent the object, we dont need to use `?`

packages/types/src/order/common.ts
```diff
@@ -902,3 +916,11 @@ export interface FilterableOrderTransactionProps
   reference?: string | string[]
   reference_id?: string | string[]
 }
+
+export interface FilterableOrderItemProps
+  extends BaseFilterable<FilterableOrderItemProps> {
+  id?: string | string[]
+  order_id?: string | string[]
+  version?: string | string[]
+  item_id?: string | string[]
```
adrien2p: suggestion: maybe add OperatorMap?

packages/types/src/order/service.ts
```diff
@@ -152,12 +165,12 @@ export interface IOrderModuleService extends IModuleService {
   removeLineItems(itemIds: string[], sharedContext?: Context): Promise<void>
   removeLineItems(itemIds: string, sharedContext?: Context): Promise<void>
   removeLineItems(
-    selector: Partial<OrderLineItemDTO>,
+    selector: Partial<FilterableOrderLineItemProps>,
     sharedContext?: Context
   ): Promise<void>
```
adrien2p: suggestion: use delete instead of remove for delete method?

packages/types/src/order/workflows.ts
```diff
@@ -0,0 +1,8 @@
+import { CustomerDTO } from "../customer"
+import { RegionDTO } from "../region"
+import { OrderDTO } from "./common"
+
+export interface OrderWorkflowDTO extends OrderDTO {
+  customer?: CustomerDTO
+  region?: RegionDTO
```
adrien2p, line 7: Q: goes with some other about typings, do we expect the entire region and customer? I believe that worflows are in a separate layer, so we can have differences in the DTO to specify what is needed by the workflows. 

It is a more global discussion no need to change it for nw 
cc @olivermrbl

adrien2p (commented):
nice workf man 🔥

packages/medusa/src/api-v2/admin/draft-orders/validators.ts
carlos-r-l-rodrigues: I kept what we have in v1. I assume this is to create a manual order that is already complete with all the info needed.

packages/order/src/services/order-module-service.ts
carlos-r-l-rodrigues: I don't know if it is ok to create this breaking change. @olivermrbl ?
Even though Orders v2 will differ a bit from v1 as we incorporate the "Cart" in it.

adrien2p: Maybe i missed something, but remove could mean remove from something and delete would mean delete something. But here either way it will delete it from the order right? As we delete the order item. I dont have a strong opinion but if both methods does the same then i think we should keep the consistency wdyt?

carlos-r-l-rodrigues: I agree. And maybe we should stick to keep the endpoints compatible to v1 as possible, but not necessarily the services/modules.

adrien2p: Agreed 👍

packages/core-flows/src/order/steps/set-tax-lines-for-items.ts
carlos-r-l-rodrigues: changed it to void 0.
there is no need to be a Promise.

packages/order/src/services/order-module-service.ts
olivermrbl: The `remove[Something]` and `add[Something]` was from before we had the autogenerated methods. I think it makes sense to remove these in favor of delete and create. 

> And maybe we should stick to keep the endpoints compatible to v1 as possible, but not necessarily the services/modules.

Yes, absolutely. The services do not need to be compatible with V1.

packages/order/src/models/line-item-adjustment.ts
carlos-r-l-rodrigues: the migration was correct already

packages/core-flows/src/order/workflows/create-orders.ts
carlos-r-l-rodrigues: here if unit_price is provided it has to be used and replace the variant price, and the .max is to prevent negative values (this is being validated in v1)

packages/core-flows/src/order/utils/prepare-custom-line-item-data.ts
```diff
@@ -0,0 +1,39 @@
...
+  variant_barcode?: string
+  variant_title?: string
+  unit_price: BigNumberInput
+  metadata?: Record<string, any>
+}
+
+export function prepareCustomLineItemData(data: Input): Output {
+  const { variant, unitPrice, quantity, metadata } = data
+
+  const lineItem: any = {
```
adrien2p, line 27: suggestion: shoult it be Output instead of any?

packages/medusa/src/api-v2/admin/draft-orders/[id]/route.ts
Comment (adrien2p) [reply packages/medusa/src/api-v2/admin/draft-orders/[id]/route.ts:16]: Should we update it?

packages/medusa/src/api-v2/admin/draft-orders/query-config.ts
```diff
@@ -0,0 +1,45 @@
+export const defaultAdminOrderRelations = []
+export const allowedAdminOrderRelations = []
+
+export const defaultAdminListOrderFields = [
+  "id",
+  "status",
+  "version",
+  "items.*",
```
adrien2p: todo: *items, under the hood, it will do the necessary, here, if someone decide to use the listConfig or retrieveConfig, they will have a * in the select which does not work in most orm

adrien2p: same for all other `.*`

packages/medusa/src/api-v2/admin/draft-orders/query-config.ts
```diff
@@ -0,0 +1,45 @@
...
+  "shipping_methods.tax_lines.*",
+  "shipping_methods.adjustments.*",
+  "summary",
+  "metadata",
+  "created_at",
+  "updated_at",
+]
+
+export const retrieveTransformQueryConfig = {
+  defaultFields: defaultAdminOrderFields,
```
adrien2p, line 47: todo: defaults

packages/medusa/src/api-v2/admin/draft-orders/query-config.ts
```diff
@@ -0,0 +1,45 @@
...
+  isList: false,
+}
+
+export const listTransformQueryConfig = {
+  defaultFields: defaultAdminListOrderFields,
```
adrien2p, line 54: todo: defaults

adrien2p (approved):

carlos-r-l-rodrigues:
@olivermrbl I will merge this one. I have a next PR touching multiple modules to type them to use Big numbers branched from this.

PR #6799, opened 2024-03-22
chore: added tsdocs to Region Module resources

shahednasser:
Add/update TSDocs to the Region Module's resources


packages/types/src/region/service.ts
```diff
@@ -218,28 +294,32 @@ export interface IRegionModuleService extends IModuleService {
   ): Promise<[RegionDTO[], number]>
 
   /**
-   * This method retrieves a country by its ID.
+   * This method retrieves a country by its 2 character ISO code.
    *
-   * @param {string} countryId - The country's ID.
-   * @param {FindConfig<RegionCountryDTO>} config - The configurations determining how the region country is retrieved. Its properties, such as `select` or `relations`, accept the
-   * attributes or relations associated with a region country.
+   * @param {string} countryId - The country's ID, which is the 2 character ISO code.
```
shahednasser, line 299: @olivermrbl can you confirm whether this is true (that the 2 character ISO code is used instead of the ID)? Looking at the module I found that the iso code is the primary key.

olivermrbl: Yes, that's correct 👍

olivermrbl (approved):

olivermrbl:
@shahednasser, this one will need to have conflicts resolved

shahednasser:
@olivermrbl resolved

PR #6800, opened 2024-03-22
feat(core-flows, medusa): add update stock location endpoint to api-v2

pKorsholm:
**What**
- add update stock location endpoint to api-v2


integration-tests/modules/__tests__/stock-location/index.spec.ts
```diff
@@ -52,5 +52,33 @@ medusaIntegrationTestRunner({
...
+            name: "test location",
+          },
+          adminHeaders
+        )
+
+        stockLocationId = createResponse.data.stock_location.id
+      })
+      it("should update stock location name", async () => {
+        const response = await api.post(
+          `/admin/stock-locations/${stockLocationId}`,
```
olivermrbl: todo: I realize we have 3-4 PRs containing API tests in `modules`, so I think now would be a good time to migrate these to `integration-tests/api` to not accumulate too much debt.

packages/core-flows/src/stock-location/steps/update-stock-locations.ts
```diff
@@ -0,0 +1,52 @@
...
+import { StepResponse, createStep } from "@medusajs/workflows-sdk"
+import {
+  convertItemResponseToUpdateRequest,
+  getSelectsAndRelationsFromObjectArray,
+} from "@medusajs/utils"
+
+import { ModuleRegistrationName } from "@medusajs/modules-sdk"
+
+export const updateStockLocationsStepId = "update-stock-locations-step"
+export const updateStockLocationsStep = createStep(
```
olivermrbl, line 21: todo: As per our convention, this should use the input type `{ selector, update }` instead of relying on the ID for the updates. You can reference the `update-regions` or `update-users` steps.

packages/stock-location-next/src/services/stock-location.ts
```diff
@@ -117,13 +121,53 @@ export default class StockLocationModuleService<
...
 
+  upsert(
+    data: UpsertStockLocationInput,
+    context?: Context
+  ): Promise<StockLocationTypes.StockLocationDTO>
+  upsert(
+    data: UpsertStockLocationInput[],
+    context?: Context
+  ): Promise<StockLocationTypes.StockLocationDTO[]>
+  async upsert(
```
olivermrbl, line 134: todo: missing decorator

packages/stock-location-next/src/services/stock-location.ts
```diff
@@ -117,13 +121,53 @@ export default class StockLocationModuleService<
...
   }
 
+  upsert(
+    data: UpsertStockLocationInput,
+    context?: Context
+  ): Promise<StockLocationTypes.StockLocationDTO>
+  upsert(
+    data: UpsertStockLocationInput[],
+    context?: Context
+  ): Promise<StockLocationTypes.StockLocationDTO[]>
```
olivermrbl:
```suggestion
  async upsert(
    data: UpsertStockLocationInput,
    context?: Context
  ): Promise<StockLocationTypes.StockLocationDTO>
  async upsert(
    data: UpsertStockLocationInput[],
    context?: Context
  ): Promise<StockLocationTypes.StockLocationDTO[]>
```

packages/stock-location-next/src/services/stock-location.ts
```diff
@@ -117,13 +121,53 @@ export default class StockLocationModuleService<
...
+  async upsert(
+    data: UpsertStockLocationInput | UpsertStockLocationInput[],
+    context: Context = {}
```
olivermrbl: todo: Missing `@MedusaContext` argument decorator

adrien2p:
@pKorsholm could you add a little description please 👍

.changeset/silent-files-kiss.md
```diff
@@ -0,0 +1,6 @@
+---
+"@medusajs/core-flows": patch
+"@medusajs/medusa": patch
```
adrien2p, line 3: todo: missing types package

packages/core-flows/src/stock-location/steps/update-stock-locations.ts
```diff
@@ -0,0 +1,59 @@
...
+import {
+  convertItemResponseToUpdateRequest,
+  getSelectsAndRelationsFromObjectArray,
+} from "@medusajs/utils"
+
+import { ModuleRegistrationName } from "@medusajs/modules-sdk"
+import { UpdateStockLocationInput } from "@medusajs/types"
+
+interface StepInput {
+  selector: FilterableStockLocationProps
```
adrien2p, line 16: Q: shouldn't `FilterableStockLocationProps` requires some cleanup? At the moment, the filterable props are completely different from the other modules, and the comparisons operators differ from the rest of the modules

packages/core-flows/src/stock-location/steps/update-stock-locations.ts
```diff
@@ -0,0 +1,59 @@
...
+      return
+    }
+
+    const stockLocationService = container.resolve<IStockLocationServiceNext>(
+      ModuleRegistrationName.STOCK_LOCATION
+    )
+
+    await stockLocationService.upsert(
+      revertInput.map((item) => ({
+        id: item.id,
```
adrien2p, line 54: Q: shouldn't the address/metadata be there too?

packages/medusa/src/api-v2/admin/stock-locations/[id]/route.ts
```diff
@@ -4,8 +4,40 @@ import {
...
+
+  const remoteQuery = req.scope.resolve(ContainerRegistrationKeys.REMOTE_QUERY)
+
+  const [stock_location] = await remoteQuery(
+    remoteQueryObjectFromString({
+      entryPoint: "stock_locations",
+      variables: {
+        id,
+      },
+      fields: req.retrieveConfig.select as string[],
```
Comment (adrien2p) [inline packages/medusa/src/api-v2/admin/stock-locations/[id]/route.ts]: todo: req.remoteQueryConfig.fields

packages/stock-location-next/src/services/stock-location.ts
```diff
@@ -117,13 +121,55 @@ export default class StockLocationModuleService<
...
+    ) as CreateStockLocationInput[]
+
+    const operations: Promise<StockLocation[]>[] = []
+
+    if (toCreate.length) {
+      operations.push(this.create_(toCreate, context))
+    }
+    if (toUpdate.length) {
+      operations.push(this.update_(toUpdate, context))
+    }
```
adrien2p, line 169: todo: those two operations happen in different transactions, should we create an upsert_ method that is wrapped in a single transaction to prevent any unexpected side effects?

packages/stock-location-next/src/services/stock-location.ts
```diff
@@ -134,14 +180,24 @@ export default class StockLocationModuleService<
...
+      normalizedInput = [{ id: idOrSelector, ...data }]
+    } else {
+      const locations = await this.list(idOrSelector, {}, context)
+
+      normalizedInput = locations.map((region) => ({
+        id: region.id,
+        ...data,
+      }))
+    }
+    const updated = await this.update_(normalizedInput, context)
```
adrien2p, line 207: suggestion: if instead of calling update_ you call `this.stockLocationService_.update` you wouldn't have to do the if else before as the internal update accepts a selector too.

cc @olivermrbl

packages/stock-location-next/src/services/stock-location.ts
```diff
@@ -219,9 +216,11 @@ export default class StockLocationModuleService<
 
   @InjectTransactionManager("baseRepository_")
   async update_(
-    data: (UpdateStockLocationInput & { id: string })[],
+    data:
+      | (UpdateStockLocationInput & { id: string })[]
+      | { data: any; selector: FilterableStockLocationProps },
```
adrien2p, line 222: suggestion: data: UpdateStockLocationInput | UpdateStockLocationInput[] ?

packages/core-flows/src/stock-location/steps/update-stock-locations.ts
```diff
@@ -0,0 +1,61 @@
...
+        name: item.name,
+        ...(item.metadata ? { metadata: item.metadata } : {}),
+        ...(item.address ? { address: item.address } : {}),
```
adrien2p, line 57: todo: if item.metadata was null, it will be replaced with an empty object, not sure this is what we are expecting. here is my suggestion
```suggestion
        metadata: isDefined(item.metadata) ? item.metadata : null
        address: isDefined(item.address) ? item.address : null
```

wdyt?

adrien2p (approved):
🔥 one more thing

packages/core-flows/src/stock-location/steps/update-stock-locations.ts
pKorsholm: it wont override the fields unless they are part of the item, since the operator is spreading an empty object and not assigning an empty object thus the key `metadata` wont exist, same goes for address. 

On the other hand the code you wrote seems to nullify the metadata if it wasn't updated since it then wont be included.

adrien2p: Oh yes you are right 😂 nevermind, i thought we were assigning and not spreading, my bad

PR #6801, opened 2024-03-22
feat(types, medusa, core-flows): add delete-stock-location endpoint to api-v2

pKorsholm:


integration-tests/modules/__tests__/stock-location/index.spec.ts
```diff
@@ -52,5 +54,64 @@ medusaIntegrationTestRunner({
         )
       })
     })
+
+    describe("Delete stock location", () => {
```
olivermrbl: todo: Same as other comments.

olivermrbl (approved):
LGTM

PR #6802, opened 2024-03-22
Fix/pricing create missmatch

adrien2p:
**What**
remove indexes usage and refactor create implemetation

### DIFF
--- a/packages/pricing/src/models/price-set.ts
+++ b/packages/pricing/src/models/price-set.ts
@@ -46,7 +46,7 @@ export default class PriceSet {
   @ManyToMany({
     entity: () => RuleType,
     pivotEntity: () => PriceSetRuleType,
-    cascade: [Cascade.PERSIST, "soft-remove" as Cascade],
+    cascade: ["soft-remove" as Cascade],
   })
   rule_types = new Collection<RuleType>(this)
 
--- a/packages/pricing/src/services/pricing-module.ts
+++ b/packages/pricing/src/services/pricing-module.ts
@@ -2,6 +2,7 @@ import {
   AddPricesDTO,
   Context,
   CreatePriceListRuleDTO,
+  CreatePricesDTO,
   DAL,
   InternalModuleDeclaration,
   ModuleJoinerConfig,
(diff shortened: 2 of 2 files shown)


carlos-r-l-rodrigues (approved):

packages/pricing/integration-tests/__tests__/services/pricing-module/price-set.spec.ts
```diff
@@ -5,9 +5,10 @@ import {
   IPricingModuleService,
 } from "@medusajs/types"
 import { SqlEntityManager } from "@mikro-orm/postgresql"
-import { SuiteOptions, moduleIntegrationTestRunner } from "medusa-test-utils"
+import { moduleIntegrationTestRunner, SuiteOptions } from "medusa-test-utils"
 import { PriceSetRuleType } from "../../../../src"
 import { seedPriceData } from "../../../__fixtures__/seed-price-data"
+import * as console from "console"
```
carlos-r-l-rodrigues: rm

sradevski (approved):

adrien2p:
There is one more thing to fix, when we create some items, i ve seen that we then re query them before returning them, there, we can also have a different order than what was the input.

packages/pricing/src/models/price.ts
```diff
@@ -129,10 +129,14 @@ export default class Price {
   @BeforeCreate()
   onCreate() {
     this.id = generateEntityId(this.id, "price")
+    this.price_set_id ??= this.price_set?.id!
```
olivermrbl, line 132: thought(non-blocking): When we add these assignments to the hooks, I get a slight feeling that we should be doing something different. With our approach, this becomes a very important part of the model definition, yet easily forgotten. And it is not the easiest to debug in case you accidentally left them out. Maybe there is a better way.

Have you had the same thought, or is this just me? Also, this is not something we need to act on, but I was just curious.

adrien2p: For now i think it is the easiest approach. In terms of debug, I didn't find it particularly difficult on my side, usually you have something telling you the foreign key is missing. Do you have something in mind? I could think about it and come back to you if i find better

PR #6806, opened 2024-03-23
feat(medusa-payment-paypal,medusa-payment-stripe): speed up unit tests in CI

riqwan:
what:

- shards unit tests into 2
- reduce sharding from 3 to 2 in api and modules
- migrate paypal and stripe plugins to use swc jest
- removes dependent build command from turbo
- move build step to cached steps 

improvements:

- This reduces ~10 mins of CI time from unit tests
- Cancels previous workflows correctly
- reduces about a minute each on each step due to steps cleanup

coming up: ([PR](https://github.com/medusajs/medusa/pull/6820#issuecomment-2018024984))

- migrating medusa to swc shaves another 7 minutes in unit tests. We can then remove the sharding in unit tests and bring them under 10 mins in total.


olivermrbl (approved):
💈 !!

PR #6813, opened 2024-03-25
feat: Modify the abstract repository upsert to handle subresources as per convention

sradevski:


packages/utils/src/dal/mikro-orm/mikro-orm-repository.ts
```diff
@@ -333,91 +345,186 @@ export function mikroOrmBaseRepositoryFactory<T extends object = object>(
...
+          relation.targetMeta?.properties[lowerCaseFirst(entity.name)]
+            ?.joinColumns
+        const joinColumnsConstraints = {}
+        joinColumns?.forEach((joinColumn, index) => {
+          const referencedColumnName = relation.referencedColumnNames[index]
+          joinColumnsConstraints[joinColumn] = data[referencedColumnName]
+        })
+
+        // TODO: We are assuming that the table uses an ID as a primary key, adjust this to support composite keys
+        await manager.nativeDelete(relation.type, {
```
carlos-r-l-rodrigues, line 625: Here we should soft delete and maybe only execute this with a explicit flag. Like having another method `upsertAndDelete`.
The reason is that the upsert (+ delete) method will have to return all the keys affected so that we can emit events + invoke the remote link about deleted entities/ids.

cc: @adrien2p

adrien2p: yes we discussed that, I suggested upsertAndDelete but oli prefer `set` or something like that instead, we discussed on friday. Also, we ve talked about it with stev, and yes we will return everything impacted and how it was impacted when the feature is done. it should be farly simple because we are limiting to one level deep for relation

olivermrbl: I am also perfectly fine with `upsertAndDelete`. My idea was just to have a clear distinction from `upsert` when doing deletions as part of the action, because that is not expected from a conventional upsert.

adrien2p: yes exactly, that was my argument too, we are expecting upsert to only do create/update, so havaing upsertAndDelete would be ideal IMO

packages/utils/src/dal/mikro-orm/mikro-orm-repository.ts
```diff
@@ -420,6 +435,329 @@ export function mikroOrmBaseRepositoryFactory<T extends object = object>(
...
+                  { ...originalEntry, id: (entityEntry as any).id },
+                  relation
+                )
+              return
+            })
+          )
+        })
+      )
+
+      // // We want to populate the identity map with the data that was written to the DB, and return an entity object
```
sradevski, line 545: Leaving this for a follow-up PR as it isn't as simple as uncommenting it and it isn't required per-se.

packages/types/src/dal/repository-service.ts
```diff
@@ -70,6 +71,12 @@ export interface RepositoryService<T = any> extends BaseRepositoryService<T> {
   ): Promise<[T[], Record<string, unknown[]>]>
 
   upsert(data: any[], context?: Context): Promise<T[]>
+
+  upsertWithReplace(
+    data: any[],
+    upsertConfig?: UpsertWithReplaceConfig<T>,
```
riqwan: suggestion: Would recommend scoping the variable names with `upsertWithReplace` as well since its not doing just an upsert or change to config if there is a context of the function

```suggestion
    config?: UpsertWithReplaceConfig<T>,
```

packages/utils/src/dal/mikro-orm/mikro-orm-repository.ts
```diff
@@ -420,6 +435,329 @@ export function mikroOrmBaseRepositoryFactory<T extends object = object>(
...
+      ) {
+        // If it is undefined (not passed) then just return
+        if (entryCopy[relation.name] === undefined) {
+          return
+        }
+
+        // If null is passed (to be unset) then we need to set that on the join column
+        if (entryCopy[relation.name] === null) {
+          delete entryCopy[relation.name]
+          entryCopy[relation.joinColumns[0]] = null
```
riqwan, line 650: question: will there be multiple join columns here? If so, how do we know if its the first one?

packages/utils/src/dal/mikro-orm/mikro-orm-repository.ts
```diff
@@ -420,6 +435,329 @@ export function mikroOrmBaseRepositoryFactory<T extends object = object>(
...
+            allRelationsToHandle,
+            relation,
+            entryCopy,
+            upsertsPerType
+          )
+        })
+
+        const mainEntity = this.getEntityWithId(manager, entity.name, entryCopy)
+        reconstructedResponse.push({ ...mainEntity, ...reconstructedEntry })
+        originalDataMap.set(mainEntity.id, entry)
```
riqwan, line 501: question: `entry` here would be with relations right? should we pass the entry without any relations here to prevent any side effects?

packages/utils/src/dal/mikro-orm/mikro-orm-repository.ts
```diff
@@ -420,6 +435,329 @@ export function mikroOrmBaseRepositoryFactory<T extends object = object>(
...
+        const newEntity = this.getEntityWithId(
+          manager,
+          relation.type,
+          entryCopy[relation.name]
+        )
+
+        delete entryCopy[relation.name]
+        entryCopy[relation.joinColumns[0]] = newEntity.id
+
+        if (!upsertsPerType[relation.type]) {
```
riqwan: question: would `relation.type` here be ManyToOne? if so whats the difference between this and `relation.reference`? Can you maybe add a comment on what this is here?

packages/utils/src/dal/mikro-orm/mikro-orm-repository.ts
```diff
@@ -420,6 +435,329 @@ export function mikroOrmBaseRepositoryFactory<T extends object = object>(
...
+      await promiseAll(
+        entries.map(async (data) => {
+          orderedEntities.push(data)
+          const existingEntity = existingEntitiesMap.get(data.id)
+          if (existingEntity) {
+            await manager.nativeUpdate(entityName, { id: data.id }, data)
+            updatedEntities.push(data)
+          } else {
+            await manager.insert(entityName, data)
+            createdEntities.push(data)
```
riqwan: suggestion: we don't seem to be doing anything with these, should we remove?

packages/utils/src/dal/mikro-orm/mikro-orm-repository.ts
```diff
@@ -420,6 +435,329 @@ export function mikroOrmBaseRepositoryFactory<T extends object = object>(
...
+      const typesToUpsert = Object.entries(upsertsPerType)
+      if (!typesToUpsert.length) {
+        return []
+      }
+
+      return (
+        await promiseAll(
+          typesToUpsert.map(async ([type, data]) => {
+            return await this.upsertMany_(manager, type, data)
+          })
```
riqwan:
```suggestion
          typesToUpsert.map(([type, data]) => {
            return this.upsertMany_(manager, type, data)
          })
```

packages/utils/src/dal/mikro-orm/mikro-orm-repository.ts
```diff
@@ -420,6 +435,329 @@ export function mikroOrmBaseRepositoryFactory<T extends object = object>(
       // TODO return the all, created, updated entities
       return upsertedEntities
     }
+
+    // UpsertWithReplace does several things to simplify module implementation.
+    // For each entry of your base entity, it will go through all relations, and it will do a diff between what is passed and what is in the database.
+    // For each relation, it create new entries (without an ID), it will associate existing entries (with only an ID), and it will update existing entries (with an ID and other fields).
+    // Finally, it will delete the relation entries that were omitted in the new data.
+    // The response is a POJO of the data that was written to the DB, including all new IDs. The order is preserved with the input.
+    // Limitations: We expect that IDs are used as primary keys, and we don't support composite keys. Also, we only support 1-level depth of upserts
```
riqwan: q: would a many to many with a pivot entity more than the join keys be supported here? Might be worth to add that in as a limitation as we have a couple of models like that

packages/utils/src/dal/mikro-orm/integration-tests/__tests__/mikro-orm-repository.spec.ts
```diff
@@ -0,0 +1,844 @@
...
+        relationsToUpsert: ["entity1"],
+        relationsToSkip: [],
+      })
+      const listedEntities = await manager2().find({
+        where: { id: "2" },
+        options: { populate: ["entity1"] },
+      })
+
+      expect(listedEntities).toHaveLength(1)
+      expect(listedEntities[0]).toEqual(
```
riqwan, line 314: sugg: an assertion on what happens to entity1 (if it still exists or is deleted)

packages/utils/src/dal/mikro-orm/integration-tests/__tests__/mikro-orm-repository.spec.ts
```diff
@@ -0,0 +1,844 @@
...
+          id: "2",
+          title: "newen2",
+          entity1: null,
+        })
+      )
+    })
+
+    it("should only create the parent entity of a one-to-many if relation is not included", async () => {
+      const entity1 = {
+        id: "1",
```
riqwan, line 328: q: if an ID is passed and is not present in the db, we seem to be creating it. Should we guard against this case and throw in the event of an incorrect ID being passed?

packages/utils/src/dal/mikro-orm/integration-tests/__tests__/mikro-orm-repository.spec.ts
```diff
@@ -0,0 +1,844 @@
...
+      )
+    })
+
+    it("should successfully update, create, and delete subentities an entity with a one-to-many relation", async () => {
+      const entity1 = {
+        id: "1",
+        title: "en1",
+        entity2: [
+          { id: "2", title: "en2-1" },
+          { id: "3", title: "en2-2" },
```
riqwan, line 432: based on the above guard case comment, I would expect here for an update to happen, but it is indeed creating it. 

Example, if an ID is passed of a resource that is already deleted, it would recreate it instead of throwing an error.

riqwan (commented):
uff, nice work man! 🔥 I'm still going through the tests, i'll drop what i have so far.

packages/utils/src/dal/mikro-orm/integration-tests/__tests__/mikro-orm-repository.spec.ts
```diff
@@ -0,0 +1,844 @@
...
+      })
+
+      const listedEntities = await manager1().find({
+        where: { id: "1" },
+        options: { populate: ["entity3"] },
+      })
+
+      expect(listedEntities).toHaveLength(1)
+      expect(listedEntities[0].title).toEqual("newen1")
+      expect(listedEntities[0].entity3.getItems()).toHaveLength(2)
```
riqwan, line 618: sugg: assert that only the pivot tables are being deleted and keeping the inverse entities alive.

packages/utils/src/dal/mikro-orm/integration-tests/__tests__/mikro-orm-repository.spec.ts
```diff
@@ -0,0 +1,844 @@
...
+      )
+    })
+
+    it("should successfully update, create, and delete subentities an entity with a many-to-many relation", async () => {
+      const entity1 = {
+        id: "1",
+        title: "en1",
+        entity3: [
+          { id: "4", title: "en3-1" },
+          { id: "5", title: "en3-2" },
```
riqwan, line 594: q: should updating actually be allowed in a many to many? 

The entity3 might be a part of several entities, updating through this entity will also affect all the other entities that its associated with.

packages/utils/src/dal/mikro-orm/mikro-orm-repository.ts
```diff
@@ -116,6 +120,17 @@ export class MikroOrmBaseRepository<T extends object = object>
     throw new Error("Method not implemented.")
   }
 
+  upsertWithReplace(
+    data: unknown[],
+    upsertConfig: UpsertWithReplaceConfig<T> = {
+      relationsToUpsert: [],
+      relationsToSkip: [],
```
riqwan: do we need the `relationsToSkip` if `relationsToUpsert` is essentially a whitelist?

riqwan (approved):
I've raised some questions. If on agreement, nothing that can't come as separate iterations, overall lgtm! 🔥

packages/utils/src/dal/mikro-orm/mikro-orm-repository.ts
adrien2p: for this iteration, handling multiple primary/foreign keys have been punt to later

packages/utils/src/dal/mikro-orm/mikro-orm-repository.ts
adrien2p: this is the original data, that is why we are keeping the original entry

packages/utils/src/dal/mikro-orm/mikro-orm-repository.ts
adrien2p: type is the Class name, reference is 1:m for example

packages/utils/src/dal/mikro-orm/mikro-orm-repository.ts
adrien2p: we need to keep it as we will need to return what have been created/deleted/updated later for the event management

packages/utils/src/dal/mikro-orm/mikro-orm-repository.ts
adrien2p: in theory mikro orm would still resolve the join columns internally, we can eventually add a test to validate that but it should work

packages/utils/src/dal/mikro-orm/integration-tests/__tests__/mikro-orm-repository.spec.ts
adrien2p: it should create it, we are by design allowing to have an id provided when we go through the generateEntityId, I would say that it should be allowed if for example the user does not want to use the same id generation pattern wdyt?

packages/utils/src/dal/mikro-orm/integration-tests/__tests__/mikro-orm-repository.spec.ts
adrien2p: that is intended I believe, as we usually check for unique if deleted at is null, otherwise you would call the restore method no?

packages/utils/src/dal/mikro-orm/integration-tests/__tests__/mikro-orm-repository.spec.ts
adrien2p: that is true, we now only allow to create, attach an existing, delete

packages/utils/src/dal/mikro-orm/mikro-orm-repository.ts
sradevski: @riqwan for now we only support ID-based keys in the function, so there will only be 1 (or it will fail otherwise). Adding support for composite key references will be a lot more work with very little benefit for now IMO

packages/utils/src/dal/mikro-orm/mikro-orm-repository.ts
olivermrbl: +1

If not, I would change it to `include: []`

packages/utils/src/dal/mikro-orm/mikro-orm-repository.ts
sradevski: @riqwan that gets handled below, we add it here just so we have a reference to the input data, and then we iterate over each relation and handle it below

packages/utils/src/dal/mikro-orm/mikro-orm-repository.ts
sradevski: `relation.type` would be the name of the entity class (eg. `Variant`).

What we do here is per type/relation we do upsert of the data, so if you have many to one on Variant and Tag, you'd do an upsert per type.

I can add a comment though

packages/utils/src/dal/mikro-orm/mikro-orm-repository.ts
sradevski: @riqwan we could yeah, it was left there for when we fire off events, but I can remove it for now

packages/utils/src/dal/mikro-orm/mikro-orm-repository.ts
sradevski: @riqwan good point, it's not supported now and I can add it as a limitation

packages/utils/src/dal/mikro-orm/integration-tests/__tests__/mikro-orm-repository.spec.ts
sradevski: @riqwan In the workflows there are scenarios where you want to recreate a deleted entry, which is why we allowed passing an ID in the modules (but it shouldn't be exposed in the HTTP endpoints). I think this is a bigger discussion that I'll leave for another time

packages/utils/src/dal/mikro-orm/integration-tests/__tests__/mikro-orm-repository.spec.ts
sradevski: That's a good point, let me think about it.

packages/utils/src/dal/mikro-orm/mikro-orm-repository.ts
sradevski: @riqwan you might want to let the ORM take care of a relation (eg. if you pass `{collection: {id: "1234"}}` the ORM will link that to the product, so you don't want to either update nor skip it).

I think we can rethink what is being configured, but this works for now so I'll leave it as-is for the PR (and leave a comment in the code)

packages/utils/src/dal/mikro-orm/mikro-orm-repository.ts
```diff
@@ -420,6 +435,329 @@ export function mikroOrmBaseRepositoryFactory<T extends object = object>(
...
+                await this.assignCollectionRelation_(
+                  manager,
+                  { ...originalEntry, id: (entityEntry as any).id },
+                  relation
+                )
+              return
+            })
+          )
+        })
+      )
```
adrien2p:
```suggestion
await promiseAll(
  upsertedTopLevelEntities.map(async (entityEntry, i) => {
    const originalEntry = originalDataMap.get((entityEntry as any).id)!
    const reconstructedEntry = reconstructedResponse[i]

    return allRelations?.map(async (relation) => {
      const relationName = relation.name as keyof T
      if (!upsertConfig.relationsToUpsert?.includes(relationName)) {
        return
      }

      // TODO: Handle ONE_TO_ONE
      // One to one and Many to one are handled outside of the assignment as they need to happen before the main entity is created
      if (
        relation.reference === ReferenceType.ONE_TO_ONE ||
        relation.reference === ReferenceType.MANY_TO_ONE
      ) {
        return
      }

      reconstructedEntry[relationName] =
        await this.assignCollectionRelation_(
          manager,
          { ...originalEntry, id: (entityEntry as any).id },
          relation
        )
      return
    })
  }).flat()
)
```

packages/utils/src/dal/mikro-orm/mikro-orm-repository.ts
```diff
@@ -420,6 +435,329 @@ export function mikroOrmBaseRepositoryFactory<T extends object = object>(
...
+
+        const pivotData = normalizedData.map((currModel) => {
+          return {
+            [parentPivotColumn]: (data as any).id,
+            [currentPivotColumn]: currModel.id,
+          }
+        })
+
+        const qb = manager.qb(relation.pivotEntity)
+        await qb.insert(pivotData).onConflict().ignore().execute()
```
olivermrbl, line 593: question: Do we want to always ignore conflicts on insertion?

adrien2p (approved):
🔥

packages/utils/src/dal/mikro-orm/integration-tests/__tests__/mikro-orm-repository.spec.ts
sradevski: Same as above, for now we allow it, if we don't we can also potentially get read of a DB SELECT operation in the upsert

packages/utils/src/dal/mikro-orm/mikro-orm-repository.ts
olivermrbl: suggestion: We could remove the `ordering` from the upsert method and let that be someone, you'd have to do separately. Then, `upsertMany` could have a return type of `{ created, updated }`.

Alternatively, just do `{ created, updated, ordered }`.

packages/utils/src/dal/mikro-orm/mikro-orm-repository.ts
olivermrbl: @sradevski, can I get you show this with an example? I am not sure I understand.

packages/utils/src/dal/mikro-orm/integration-tests/__tests__/mikro-orm-repository.spec.ts
```diff
@@ -0,0 +1,844 @@
...
+      expect(wrap(listedEntities[0]).toPOJO()).toEqual(
+        expect.objectContaining({
+          id: "1",
+          title: "en1",
+        })
+      )
+    })
+
+    // TODO: Should we support this
+    it("should successfully create an entity with a sub-entity many-to-one relation", async () => {
```
olivermrbl: In my opinion, we shouldn't allow this. But this is because I can't see a use case for it - I find it off that a "child" can create its "parent", which is how I think of this. 

If anyone has a good one, I am OK with keeping it in.

adrien2p: no this is true, a child should not create its parent, here it would mean a fulfillment can create a shipping option, or a variant can create a product. I don't believe it should be feasable

adrien2p: only the following type of relation should create
- many to many
- one to one
- one to many

packages/utils/src/dal/mikro-orm/mikro-orm-repository.ts
sradevski: Sure, For example for a product:
```
{
    id: "prod1",
    tags: [{id: "tag1", value: "1"}, {id: "tag2", value: "2"}],
    collection: {id: "col1", title: "Col" },
}
```

And then we do an update to the tags

```
product.upsertWithReplace(
    [{id: "prod1", tags: [{id: "tag3", value: "3"}], collection: {id: "col2"}}],
    {relationsToUpsert: ["tags"], relationsToSkip: []}
)
```

What we expect here is that we upsert the tags, but we only reassign a new collection to the product ("col2"). We don't want to ignore that relation, nor upsert it, we simply want to let the ORM associate it to the product

@olivermrbl @riqwan does that make sense?

packages/utils/src/dal/mikro-orm/mikro-orm-repository.ts
sradevski: @olivermrbl since the primary key in the pivot table will be the IDs of the two sides, a conflict means that the relation is already there so we can ignore the conflict. 

Of course, if we support composite keys, or other custom properties this might need to change.

packages/utils/src/dal/mikro-orm/integration-tests/__tests__/mikro-orm-repository.spec.ts
sradevski: @olivermrbl I was on the fence about this one, and I initially implemented without supporting it. We have a couple of places where this is currently possible in the product module, and since what to upsert needs to be explicitly stated, I thought it won't harm us if we support it. I can remove it if you think that's better, of course.

packages/utils/src/dal/mikro-orm/mikro-orm-repository.ts
sradevski: @olivermrbl I would probably return {created, updated, ordered} rather than just created and updated as there are places where the ordering does matter. But maybe we can leave it for another iteration (and I'll add a comment for now)? There are another 2-3 improvements that are better suited for a follow-up

packages/utils/src/dal/mikro-orm/integration-tests/__tests__/mikro-orm-repository.spec.ts
adrien2p: I believe we should not support that, furthermore, with the new conventions around relations, all many to one are marked as persist false as we are only supposed to attach the foreign key but the parent is expected to already exists and the child being either created separately or through the one to many cascade on the collection

olivermrbl: Yeah, I would also remove the support for it

packages/utils/src/dal/mikro-orm/mikro-orm-repository.ts
olivermrbl: Got it thanks

packages/utils/src/dal/mikro-orm/mikro-orm-repository.ts
olivermrbl: But can't we handle this with just the `include`? 

Upsert tags on product and associate collection with product:
```
product.upsertWithReplace(
    [{id: "prod1", tags: [{id: "tag3", value: "3"}], collection: {id: "col2"}}],
    {include: ["tags"]}
)
```

Upsert tags on product and upsert collection:
```
product.upsertWithReplace(
    [{id: "prod1", tags: [{id: "tag3", value: "3"}], collection: {title: "new one"}}],
    {include: ["tags", "collection"]}
)
```

When is the skip needed?

sradevski: @olivermrbl maybe my example was a bit incomplete 😄 , Now imagine you have `variants` list in the data, but you don't want the function to do anything about it, so you want to tell `upsertWithReplace` to ignore the variants (and with just `include`, you can't do that.

packages/utils/src/dal/mikro-orm/integration-tests/__tests__/mikro-orm-repository.spec.ts
sradevski: Ok sounds good 👍. In that case I suggest we also remove the support in our HTTP API, as it is equally strange to support it there as well. Instead, we should always only rely on IDs

packages/utils/src/dal/mikro-orm/mikro-orm-repository.ts
sradevski: @olivermrbl now that I think about it, if we don't support `many-to-one`, then just having `include` should suffice. I would call it `relations` just to make it clear that it doesn't refer to all fields, but only to the relations

packages/utils/src/dal/mikro-orm/integration-tests/__tests__/mikro-orm-repository.spec.ts
sradevski: @riqwan added a comment but I won't address this for now, as ultimately the control is in the caller if they want to pass additional data or not.

sradevski:
@olivermrbl @riqwan @adrien2p addressed all the changes. The biggest one is removing support for many-to-one creations (you can still associate the relation) and changing the config definition. This resulted in removing some more code so it also became more concise.

riqwan (approved):

sradevski:
Merging, @olivermrbl if you have any additional comments I'll address them in a separate PR

PR #6818, opened 2024-03-25
Fix/pricing create price lists

adrien2p:
**What**
Pricing module createPriceLists should not rely on indexes and simplify creation

### DIFF
--- a/.changeset/perfect-pigs-report.md
+++ b/.changeset/perfect-pigs-report.md
@@ -1,0 +1,6 @@
+---
+"@medusajs/medusa": patch
+"@medusajs/core-flows": patch
+---
+
+feat: Create user account
--- a/integration-tests/api/__tests__/admin/user.js
+++ b/integration-tests/api/__tests__/admin/user.js
@@ -1,232 +1,411 @@
 const jwt = require("jsonwebtoken")
-const path = require("path")
 
-const setupServer = require("../../../environment-helpers/setup-server")
-const { useApi } = require("../../../environment-helpers/use-api")
-const { initDb, useDb } = require("../../../environment-helpers/use-db")
-
-const userSeeder = require("../../../helpers/user-seeder")
-const adminSeeder = require("../../../helpers/admin-seeder")
+const { medusaIntegrationTestRunner } = require("medusa-test-utils")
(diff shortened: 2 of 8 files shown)


adrien2p:
@riqwan should we merge that?

packages/pricing/src/services/pricing-module.ts
```diff
@@ -713,79 +713,62 @@ export default class PricingModuleService<
...
-    }
-
-    const priceLists = await this.priceListService_.create(
-      priceListsToCreate,
-      sharedContext
-    )
+    const priceListsToCreate: PricingTypes.CreatePriceListDTO[] = data.map(
+      (priceListData) => {
+        const id = generateEntityId(
+          (priceListData as unknown as TPriceList).id,
```
riqwan, line 721: does the create dto come with an id?

packages/pricing/src/services/pricing-module.ts
```diff
@@ -713,79 +713,62 @@ export default class PricingModuleService<
...
+          "plist"
```
riqwan: sugg: this should really be coming from a constant in the model.

olivermrbl: +1

olivermrbl:
Can I get you to rebase off of `develop` and prepare for merge?

packages/pricing/src/services/pricing-module.ts
adrien2p: No it doesn't, but js doesn't prevent your from passing it and the types are just for ease of use so just in case i use it if present 👍

packages/pricing/src/services/pricing-module.ts
adrien2p: Yes true, ill change that in a followup pr the one for the tules if thats fine with you guys

adrien2p:
> Can I get you to rebase off of `develop` and prepare for merge?

Oh, didn't know the other pr have been merged, i ve been merging in it when it was available ahah. I ll do that

adrien2p:
I have @riqwan commit in the history but only my changes, so should be good, i ll tackle the feedback

PR #6819, opened 2024-03-25
feat: Create user account

olivermrbl:
Refactor of existing create user flow:
- Add `createUserAccount` workflow, creating a user and associating with an auth user

The following scenarios can take place:
1. Create a user with newly created auth user -> returns user + token
2. Create a user with an actor ID in auth request session -> throws if actor ID has different email than what is passed in request

- Migrate `integration-tests/api/**/user.js` to use new test runner


packages/medusa/src/api-v2/admin/users/route.ts
```diff
@@ -43,16 +50,49 @@ export const POST = async (
...
+  )
+
+  const { jwt_secret } = req.scope.resolve("configModule").projectConfig
+
+  let authUser = await authModuleService.retrieve(req.auth.auth_user_id)
+  let token = jwt.sign(authUser, jwt_secret)
+
+  let user: UserDTO | null = null
+
+  if (req.auth.actor_id) {
```
olivermrbl, line 53: I am a bit on the fence about how we should handle this scenario.

If an `actor_id` is assigned to the `req.auth` object, it means the request carries authentication in the form of either a bearer or a session token, AND that this authenticated session already has an associated user. 

There are two sub-scenarios we can choose from here:
1. Throw error if the email provided doesn't match the email of the authenticated user (current implementation)
2. Respond with the already authenticated user, even though the email might be different

I chose 1. as I think it would be unintuitive to respond with a user with a different email from what was passed in a request to create a new user. 

I am curios to hear what others think.

packages/medusa/src/api-v2/admin/users/route.ts
```diff
@@ -43,16 +50,49 @@ export const POST = async (
...
+
+  user = result
+
+  // Refresh token with updated auth user app_metadata
+  authUser = await authModuleService.retrieve(req.auth.auth_user_id)
+  token = jwt.sign(authUser, jwt_secret)
 
-  const [user] = result
-  res.status(200).json({ user })
+  res.status(200).json({ user, token })
```
olivermrbl: We respond with the user and token so the subsequent session token conversion is a simpler flow. This is purely convenience, so I am happy to remove it in favor of just the user if we want to prioritize consistency over a minor DX improvement.

packages/medusa/src/api-v2/admin/users/route.ts
olivermrbl: If we proceed with 1., I will need to update a few modules tests, that are currently failing e.g. `create-user.spec.js` (this test illustrates my point well, as it is failing right now because of an email mismatch).

integration-tests/api/__tests__/admin/user.js
```diff
@@ -1,232 +1,433 @@
...
+        expect(response.status).toEqual(200)
+
+        expect(response.data.users).toEqual(
+          expect.arrayContaining([
+            expect.objectContaining({
+              id: "admin_user",
+              email: "admin@medusa.js",
+              created_at: expect.any(String),
+              updated_at: expect.any(String),
+              ...breaking(
```
sradevski: NIT: I'd do this above as well (where you have v1response, v2response) just so it's more obvious what is the breaking change

packages/medusa/src/api-v2/admin/users/route.ts
sradevski: A question: Why do we need to pass the `email` in the GET endpoint? The token is the ultimate decision-maker on who to fetch, so it feels redundant to require the email (so I guess we'd end up closer to your 2nd scenario)

olivermrbl: This is the `POST` endpoint to create a new user. Could you have missed this, or does your question still stand in this context?

sradevski: @olivermrbl  Agh the diff showed GET above your changes, and I didn't check the line numbers, my bad.

So the POST is only to get a new token, but you cannot update info on the user itself (eg. first name)? If it is just for renewing a token, then my comment still stands, there is no need to pass an email IMO as it doesn't provide any added security.

olivermrbl: You can indeed pass first name and last name in the request.

The flow to create a new authenticated user is:
1. Create an AuthUser (`@medusajs/auth`)
``` 
POST /auth/admin/emailpass
{ email: "test@email.com", password: "supersecret" }
```

This will return a token signed with the AuthUser. The idea is that this token carries the authentication for the subsequent request to create an actual User.

2. Create a User (`@medusajs/auth`)
```
POST /admin/users
{ email: "test@email.com", first_name: "John", last_name: "Doe" }
{ Authorization: "Bearer [token-from-first-request]" }
```

On the server, we then perform the following actions as part of 2.:
1. In the `authenticate-middleware.ts`, we extract the AuthUser from the bearer token. This is passed along as `req.auth`.
2. In the API route creating a user, we then extract the `auth_user_id` from `req.auth` and retrieve the AuthUser
3. Then, we extract the `actor_id` from `req.auth`. 
4. If `actor_id` is present, we retrieve the User
5. else, we create a new User and associate it with the AuthUser using the `auth_user_id` extracted from `req.auth`
6. Finally, we respond with the created User and a fresh token signed with the AuthUser

My question relates to bullet 4. If `actor_id` is present in `req.auth`, it means the AuthUser attached to the request has been used previously to create a User. For example, this could happen if you cURL `POST /admin/users` with the bearer token as authorization header multiple times.

Now, in case you pass a different email, do we:
1. Respond with the User, even though the email of that User does not match the one passed in the request
2. Throw

sradevski: Discussed in slack, and I understood the background behind this better. We also concluded it's better to throw if you try to associate an already associated AuthUser.

adrien2p (approved):

PR #6820, opened 2024-03-25
feat(medusa): migrate medusa unit tests / plugins integration tests to swc jest

riqwan:
what:

- migrates medusa package to swc/jest
- migrates plugins integration tests to swc/jest
- parallelises a few of the heavy test packages
- fixes typeorm circular dependencies

**Unit Tests**

Before: 1 job => ~30 mins
After: 2 jobs => ~5 mins

Screenshot 2024-03-25 at 15 11 49

**Plugins**

Before: 1 job => ~10 mins
After: 1 job => ~5 mins

Screenshot 2024-03-25 at 15 26 03


packages/medusa/src/loaders/helpers/subscribers/__tests__/index.spec.ts
```diff
@@ -57,7 +57,7 @@ describe("SubscriberLoader", () => {
     expect(eventBusServiceMock.subscribe).toHaveBeenCalledTimes(5)
   })
 
-  it("should have registered subscribers with the correct props", async () => {
+  it.skip("should have registered subscribers with the correct props", async () => {
```
riqwan, line 60: The errors on these are very confusing, seems to come from a previous mock. Will take a look at this later. 

cc @adrien2p

olivermrbl:
This looks super nice @riqwan. However, why does the PR contain so many unrelated changes? For example, the `Relation` and changes to payload types.

riqwan:
@olivermrbl swc follows the ESM spec very strictly unlike tsc and babel, which is what we use to build. During transpiling tests with swc, we build with swc unconvering these issues. swc builds are really fast! It would be very beneficial for us if we could move our builds to swc, but thats a much larger task. 

The changes I made were to address 2 things:

- accessing properties before they're initialized (the payload changes)
- Circular dependencies (typeorm relations)

carlos-r-l-rodrigues:
> * Circular dependencies (typeorm relations)

I haven't checked the full PR yet, but if you solved circular dependencies, does that mean we can finally use ES6 (Es2020) everywhere, including /medusa?

adrien2p:
@olivermrbl the issues was the circular dependencies that we postponed long time ago as well as using var that are not yet declared. This pr tackle it.

@carlos-r-l-rodrigues yes in theory we should be able to if the package doesn't have circular or use of variables that are declared later


Edit: just saw riqwan message

integration-tests/plugins/package.json
```diff
@@ -5,7 +5,7 @@
   "license": "MIT",
   "private": true,
   "scripts": {
-    "test:integration": "node --expose-gc ./../../node_modules/.bin/jest --silent=false --runInBand --bail --detectOpenHandles --logHeapUsage --clearMocks --forceExit",
+    "test:integration": "jest --silent=false  --maxWorkers=50% --bail --detectOpenHandles --clearMocks --forceExit",
```
adrien2p, line 8: comment: can we keep the expose gc to follow memory usage?

packages/medusa-payment-paypal/package.json
```diff
@@ -18,7 +18,7 @@
   "license": "MIT",
   "scripts": {
     "prepublishOnly": "cross-env NODE_ENV=production tsc --build",
-    "test": "jest --runInBand --bail --forceExit",
+    "test": "jest --silent --bail --maxWorkers=50% --forceExit",
```
adrien2p, line 21: suggest: if you can take the time to add --passWithNoTests everywhere for the unit tests to not have to create noop tests in all packages 👍

packages/medusa/src/models/payment.ts
```diff
@@ -29,23 +30,23 @@ export class Payment extends BaseEntity {
...
+  cart: Relation<Cart>
 
   @Index()
   @Column({ nullable: true })
   order_id: string
 
   @ManyToOne(() => Order, (order) => order.payments)
   @JoinColumn({ name: "order_id" })
-  order: Order
+  order: Relation<Order>
```
sradevski, line 49: Were these changes intended (since the title says just moving tests to `swc jest`?

sradevski: Ah nevermind saw the comments above just now 👍

packages/medusa/src/api/routes/admin/price-lists/update-price-list.ts
```diff
@@ -267,4 +268,12 @@ export class AdminPostPriceListsPriceListPriceListReq {
     IsBoolean(),
   ])
   includes_tax?: boolean
+
+  constructor() {
+    if (
+      !featureFlagRouter.isFeatureEnabled(TaxInclusivePricingFeatureFlag.key)
+    ) {
+      delete this.includes_tax
+    }
```
adrien2p: as seen together, we can now remove that everywhere

adrien2p (approved):

packages/medusa/src/models/payment.ts
adrien2p: does that have any impact on manipulating a relation or its type?

adrien2p: ahah nevermind `export type Relation<T> = T;`

integration-tests/plugins/package.json
riqwan: It takes up additional time to run the tests, do we need it all the time? If it breaks due to memory, we will know and we can add it in

packages/medusa-payment-paypal/package.json
riqwan: Good idea, can do that in a follow up PR, this one is already a biggie.

packages/medusa/src/api/routes/admin/price-lists/update-price-list.ts
riqwan: done ✅

PR #6821, opened 2024-03-25
feat(admin-next): Email password invite flow in admin 2.0

olivermrbl:
**Add invite flow for email-password auth provider**

https://github.com/medusajs/medusa/assets/59018053/553adf5d-e122-4115-b4a8-288eaad84859

To test it out, you need the latest changes from `develop`, in which invite support has been added to our CLI.

Then do: 
1. Run admin dashboard: `cd packages/admin-next/dashboard && yarn run dev`
2. Start your Medusa project (that has latest changes) 
3. Run the following command in your Medusa project: `medusa user -e some@email.com --invite`
4. Copy token from terminal output
5. Access the invite URL in admin: `http://localhost:5173/invite?token=[token]`
6. Accept the invite


packages/admin-next/dashboard/src/v2-routes/invite/invite.tsx
```diff
@@ -0,0 +1,393 @@
...
+  const invite: DecodedInvite | null = token ? decodeToken(token) : null
+  const isValidInvite = invite && validateDecodedInvite(invite)
+
+  return (
+    <div className="min-h-dvh w-dvw bg-ui-bg-base relative flex items-center justify-center p-4">
+      <div className="flex w-full max-w-[300px] flex-col items-center">
+        <LogoBox
+          className="mb-4"
+          checked={success}
+          containerTransition={{
```
sradevski, line 57: Probably not something for now, but a general comment: It's much better to encapsulate and standardize transitions and animations, both for consistency in appearance, as well as simplfying the code. As you can see now, half of the JSX is animation config. cc @kasperkristensen WDYT?

sradevski (approved):

packages/admin-next/dashboard/src/v2-routes/invite/invite.tsx
```diff
@@ -0,0 +1,393 @@
...
+    useV2AcceptInvite(token)
+
+  const handleSubmit = form.handleSubmit(async (data) => {
+    await createAuthUser(
+      {
+        email: data.email,
+        password: data.password,
+      },
+      {
+        onSuccess: async ({ token: authToken }) => {
```
fPolic: nit/suggestion: use try/catch with async await instead of calbacks

packages/admin-next/dashboard/src/v2-routes/invite/invite.tsx
```diff
@@ -0,0 +1,393 @@
...
+    first_name: z.string().min(1),
+    last_name: z.string().min(1),
+    password: z.string().min(1),
+    repeat_password: z.string().min(1),
+  })
+  .superRefine(({ password, repeat_password }, ctx) => {
+    if (password !== repeat_password) {
+      ctx.addIssue({
+        code: z.ZodIssueCode.custom,
+        message: "Passwords do not match",
```
fPolic: todo: should translate

packages/admin-next/dashboard/src/v2-routes/invite/invite.tsx
```diff
@@ -0,0 +1,393 @@
...
+  exp: string
+  iat: number
+}
+
+export const Invite = () => {
+  const [searchParams] = useSearchParams()
+  const [success, setSuccess] = useState(false)
+
+  const token = searchParams.get("token")
+  const invite: DecodedInvite | null = token ? decodeToken(token) : null
```
fPolic, line 48: question: could token be expired here? should we check for that?

packages/admin-next/dashboard/src/v2-routes/invite/invite.tsx
fPolic: +1 think that we could have something as `<Fade direction= "up" />` or at least easing functions in the UI lib

fPolic (approved):
NIce 💪 
Left a few comments.

packages/admin-next/dashboard/src/v2-routes/invite/invite.tsx
olivermrbl: Do we want to address such things in UI for v2, or should we wait till after? 

And if so, do we know where we want to place these?

fPolic: I would define these in `medusajs/ui` but probably not necessary for v2 atm

kasperkristensen: If the animation is ever re-used I think it would make sense to move it to a file in `@medusajs/dashboard`, but it's a one off animation, so don't see a reason to move it into a separate file atm. Feel free to move it to a const outside of the JSX for better readability though.

And I would prefer to not move any framer logic into `@medusajs/ui`. We don't use Framer in the project, only a small number of CSS animations. If people want to add animations on top of that they can then use framer or their preferred animation component, but I don't think it should be something that is built-in, or force people to use a specific animation library ✌️

sradevski: @kasperkristensen it doesn't necessarily need to be part of `medusajs/ui`, it can be a separate `animations` folder in the admin, but my point is that it would be nice to see what is extractable. And definitely not urgent for v2, just a general comment after looking at the code.

packages/admin-next/dashboard/src/v2-routes/invite/invite.tsx
olivermrbl: This is validated on the backend, think that is fine for now.

olivermrbl:
@fPolic, could I get you to give this another review?

fPolic:
On it @olivermrbl 👀

fPolic:
The main flow works nicely!

However, I've noticed 2 issues:
1. If you try to reuse the same invite link twice -> it won't create an account but the flow will look like everything is OK (should we check if the invite token is used already?)

2. The other odd issue, probably not directly related to your PR is that when you try to log in with email/pass that doesn't exist....it will create a new record in `auth_user` table

olivermrbl:
> If you try to reuse the same invite link twice -> it won't create an account but the flow will look like everything is OK (should we check if the invite token is used already?)

Did you accept it with a different email + password or the same as the first time?

> The other odd issue, probably not directly related to your PR is that when you try to log in with email/pass that doesn't exist....it will create a new record in auth_user table

This is actually intended behaviour and part of our new auth flow. Funny enough, I just had a discussed with Seb on how to address this in our docs. Probably could have a cron job cleaning up unused auth users on a fixed schedule, but we'll let that be up to users to decide

fPolic:
> Did you accept it with a different email + password or the same as the first time?

I used a different email.

> This is actually intended behaviour and part of our new auth flow.

Ahh alright 😅 I thought that might be intentional but wanted to check.

olivermrbl:
"Issue" with double-accepting invites resolved internally. We failed to reproduce, so will keep it as is.

PR #6823, opened 2024-03-26
chore(dashboard): render main layout as home page for v2

riqwan:
Trying to keep these dashboard PRs short.

what:

- renders an empty order page as home page to showcase the main layout

Currently it renders settings page as the default home page, this PR moves that over to the orders page, which will be currently empty, but it renders the main layout where most of the v2 work will be done

Screenshot 2024-03-26 at 09 43 23

### DIFF
--- a/.changeset/angry-bottles-happen.md
+++ b/.changeset/angry-bottles-happen.md
@@ -1,0 +1,6 @@
+---
+"@medusajs/admin-ui": patch
+"@medusajs/admin": patch
+---
+
+fix(admin-ui, admin): Add missing regions relation
--- a/.changeset/calm-spoons-look.md
+++ b/.changeset/calm-spoons-look.md
@@ -1,6 +1,6 @@
 ---
-"@medusajs/core-flows": minor
-"@medusajs/medusa": minor
+"@medusajs/core-flows": patch
+"@medusajs/medusa": patch
 ---
 
 Align the v2 product HTTP endpoints to follow conventions
--- a/.changeset/perfect-pigs-report.md
+++ b/.changeset/perfect-pigs-report.md
@@ -1,0 +1,6 @@
+---
+"@medusajs/medusa": patch
+"@medusajs/core-flows": patch
+---
+
+feat: Create user account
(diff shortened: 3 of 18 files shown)


packages/admin-next/dashboard/src/components/layout-v2/main-layout/main-layout.tsx
```diff
@@ -0,0 +1,205 @@
...
+        </div>
+        <CoreRouteSection />
+        <ExtensionRouteSection />
+      </div>
+    </aside>
+  )
+}
+
+const Header = () => {
+  const { store, isError, error } = useV2Store({})
```
riqwan, line 49: only thing that changes here is the v2Store call. Lmk if this is the desired approach.

packages/admin-next/dashboard/src/lib/api-v2/store.ts
```diff
@@ -8,7 +8,7 @@ export const useV2Store = ({ initialData }: { initialData?: any }) => {
     { initialData }
   )
 
-  const store = data.stores[0]
+  const store = data?.stores[0]
```
kasperkristensen, line 11: nit: not really related to your PR, but make sense to manually set the error in the case where isLoading is false and the stores array is empty, to have the same effect as if the current useAdminStore endpoint fails. Otherwise, you might end up with a false positive, where loading is completed and no error is thrown, but at the same time store is undefined, causing the admin to break.

kasperkristensen (approved):
LGTM, added a comment on the useV2Store hook, but feel free to ignore it as its not really related to your PR.

packages/admin-next/dashboard/src/lib/api-v2/store.ts
```diff
@@ -8,7 +8,11 @@ export const useV2Store = ({ initialData }: { initialData?: any }) => {
...
+
+  if (!isLoading && !isError && typeof store === "undefined") {
+    throw new Error("Store does not exist")
+  }
```
riqwan, line 15: @kasperkristensen is this what you mean?

PR #6836, opened 2024-03-27
feat(utils): custom serialization that allows for non self ref

adrien2p:
**What**
Allow to specify a new option for the serialize utils which is `preventCircularRef`, by default this options is set to true and you have to opt out if you want. The result of the serialized data will be the following:
- service_zone
  - geo_zones[]
 
instead of

- service_zone
  - geo_zones[]
    - service_zone

I suggest that we live with it and if everything is fine, and we are ready to upgrade, I can open a pr on mikro repo

### DIFF
--- a/.changeset/silent-files-kiss.md
+++ b/.changeset/silent-files-kiss.md
@@ -1,0 +1,7 @@
+---
+"@medusajs/core-flows": patch
+"@medusajs/medusa": patch
+"@medusajs/types": patch
+---
+
+feat(core-flows, medusa): add update stock location endpoint to api-v2
--- a/integration-tests/api/__tests__/admin/stock-location/index.spec.ts
+++ b/integration-tests/api/__tests__/admin/stock-location/index.spec.ts
@@ -56,6 +56,35 @@ medusaIntegrationTestRunner({
       })
     })
 
+    describe("Update stock locations", () => {
+      let stockLocationId
+
+      beforeEach(async () => {
+        const createResponse = await api.post(
+          `/admin/stock-locations`,
+          {
+            name: "test location",
+          },
+          adminHeaders
+        )
+
+        stockLocationId = createResponse.data.stock_location.id
+      })
+
+      it("should update stock location name", async () => {
+        const response = await api.post(
+          `/admin/stock-locations/${stockLocationId}`,
+          {
+            name: "new name",
+          },
+          adminHeaders
+        )
+        expect(response.status).toEqual(200)
+
+        expect(response.data.stock_location.name).toEqual("new name")
+      })
+    })
+
     describe("Get stock location", () => {
       let locationId
       const location = {
--- a/packages/core-flows/src/stock-location/steps/index.ts
+++ b/packages/core-flows/src/stock-location/steps/index.ts
@@ -1,2 +1,3 @@
 export * from "./create-stock-locations"
+export * from "./update-stock-locations"
 export * from "./delete-stock-locations"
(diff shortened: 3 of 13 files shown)


packages/utils/src/dal/mikro-orm/utils.ts
```diff
@@ -143,9 +158,341 @@ export const mikroOrmUpdateDeletedAtRecursively = async <
...
+    const ret = {} as EntityDTO<Loaded<T, P>>
+    const keys = new Set<string>(meta.primaryKeys)
+    Object.keys(entity).forEach((prop) => keys.add(prop))
+    const visited = root.visited.has(entity)
+
+    if (!visited) {
+      root.visited.add(entity)
+    }
+
+    ;[...keys]
```
sradevski: Just a NIT: We can maybe break down this loop and the one below using helper functions, so it's easier to understand the flow of data

sradevski (commented):
I won't lie, there is a lot going on here and it's quite difficult to review it properly. Can we maybe at least add few tests for the more common scenarios?

packages/utils/src/dal/mikro-orm/utils.ts
adrien2p: The idea was to keep it as close as possible to what mikro orm has, the idea behind to not divert too much and be able in to read both side by side to see the diff at any time, i would prefer not break it down to smaller function for that reason in that particular case. I can add some tests though

sradevski: @adrien2p ah so most of the code is taken from mikroORM? Let's maybe add a comment in case someone else needs to modify it? I am ok with that then 👍

adrien2p: Yes exactly, i ve taken it and added the necessary logic for our needs

olivermrbl: Would it make sense to encapsulate our custom logic to make it very explicit what was modified? 

Either with comments, or in a separate method - you know best what would be fitting.

adrien2p: Yes i ll add comments and tests

packages/utils/src/dal/mikro-orm/__tests__/mikro-orm-serializer.spec.ts
```diff
@@ -0,0 +1,102 @@
...
+    const entity1 = new Entity1({ id: "1", deleted_at: null })
+    const entity2 = new Entity2({
+      id: "2",
+      deleted_at: null,
+      entity1: entity1,
+    })
+    entity1.entity2.add(entity2)
+
+    const serialized = await mikroOrmSerializer(entity1, {
+      preventCircularRef: false,
```
sradevski: Shouldn't this test be `true`, and the expected result below not containing entity1 inside entity2?

adrien2p: I was on it, I am ion a call

sradevski: Ah ok sorry for rushing to review again 😄

adrien2p: no problem, let me just get back at it

sradevski (approved):

adrien2p:
@olivermrbl can I merge this one?

olivermrbl (approved):

PR #6838, opened 2024-03-27
feat(workflow-engine-redis): use worker_mode

srindom:


riqwan:
@srindom is this still relevant?

PR #6840, opened 2024-03-27
feat(dashboard,medusa-react): Add Categories domain

kasperkristensen:
**What**
- Adds Categories domain.
  - Adds as much of the new design as we currently support, remainder will be left for 2.0.
- Fixes an issue with medusa-react where creating a new Category would not invalidate the list query.

### DIFF
--- a/packages/admin-next/dashboard/src/routes/categories/common/components/category-tree/category-tree.tsx
+++ b/packages/admin-next/dashboard/src/routes/categories/common/components/category-tree/category-tree.tsx
@@ -76,7 +76,7 @@ export const CategoryTree = ({
     return true
   }
 
-  const showOverlay = isDisabled || isLoading
+  const showOverlay = isLoading || isDisabled
 
   if (isLoading) {
     return (
@@ -140,7 +140,7 @@ export const CategoryTree = ({
         threshold={10}
       />
       {showOverlay && (
-        <div className="bg-ui-bg-disabled/40 absolute inset-0 animate-pulse" />
+        <div className="bg-ui-bg-disabled absolute inset-0 size-full animate-pulse opacity-80" />
       )}
     </div>
   )


packages/admin-next/dashboard/src/routes/categories/common/components/category-tree/category-tree.tsx
```diff
@@ -0,0 +1,135 @@
+import { ProductCategory } from "@medusajs/medusa"
```
fPolic, line 1: A few things I noticed with the tree:

1. Sometimes when an item is moved around, and then you click on Edit in the Action menu, the tree just rerenders and the edit modal is not open

2. toggling one item sometimes toggle another item

https://github.com/medusajs/medusa/assets/16856471/d54c6419-b638-4c13-8d7d-74408e73a71f

packages/admin-next/dashboard/src/routes/categories/category-detail/components/category-general-section/category-general-section.tsx
```diff
@@ -0,0 +1,97 @@
...
+                    to: `edit`,
+                  },
+                ],
+              },
+              {
+                actions: [
+                  {
+                    label: t("actions.delete"),
+                    icon: <Trash />,
+                    onClick: handleDelete,
```
fPolic, line 71: suggestion: disable this if the category has children since delete will fail then

kasperkristensen: IMO the best UX is to not disable buttons, but let the error occur and display it to the user. There is a TODO in the `onError` handler to show a toast with a translated error message for this case.

packages/admin-next/dashboard/src/routes/categories/common/components/category-tree/category-tree.tsx
kasperkristensen: I can't reproduce this locally, but I can see it on the preview. I think this stems from the staging server being very slow, and you can end up performing actions before the ordering has propagated correctly. I'll add some logic to disable the nodes while the server is still loading 👍

PR #6842, opened 2024-03-27
fix(utils): Mikro orm prop filtering should check existence

adrien2p:
**What**
- Should return non decorated but visible properties.
- Should only prevent circular on non scalar fields


adrien2p:
@olivermrbl for some reason, the previous pr was containing a bug, though, all pipeline were green.

cc @riqwan, I was surprised that they ran in a few sec when you look in details, something link to the cache maybe?

adrien2p:
I ll try to update the test to reflect that

carlos-r-l-rodrigues (approved):

adrien2p:
@riqwan , If i run the test locally, they are failing, though the pipeline is green

riqwan:
@adrien2p yeah, the cache doesn't seem to be busting. Let me take a look

PR #6844, opened 2024-03-27
feat: Admin V2 users domain

olivermrbl:
Add V2 users domain

Very few changes needed:
- Removed use of `role` property from all user-related component
- Tweaked the send invite request payload from `user: { ... }` to `email`

- Moved `user-edit-form.tsx` to be a shared component


packages/admin-next/dashboard/src/v2-routes/users/user-detail/components/user-general-section/user-general-section.tsx
```diff
@@ -0,0 +1,88 @@
...
+          {t("fields.name")}
+        </Text>
+        <Text
+          size="small"
+          leading="compact"
+          className={clx({
+            "text-ui-fg-subtle": !name,
+          })}
+        >
+          {name ?? "-"}
```
olivermrbl, line 83: Change from V1: Removed `role`

packages/admin-next/dashboard/src/v2-routes/users/user-invite/components/invite-user-form/invite-user-form.tsx
```diff
@@ -0,0 +1,381 @@
...
+import { ActionMenu } from "../../../../../components/common/action-menu"
+import { NoRecords } from "../../../../../components/common/empty-table-content"
+import { Form } from "../../../../../components/common/form"
+import { LocalizedTablePagination } from "../../../../../components/localization/localized-table-pagination"
+import { RouteFocusModal } from "../../../../../components/route-modal"
+import { useV2Store } from "../../../../../lib/api-v2"
+import { InviteDTO } from "@medusajs/types"
+
+const InviteUserSchema = zod.object({
+  email: zod.string().email(),
```
olivermrbl, line 44: Change from V1: Removed `role`

packages/admin-next/dashboard/src/v2-routes/users/user-list/components/user-list-table/user-list-table.tsx
```diff
@@ -0,0 +1,232 @@
...
+  const { t } = useTranslation()
+
+  return useMemo(
+    () => [
+      columnHelper.accessor("email", {
+        header: t("fields.email"),
+        cell: ({ row }) => {
+          return row.original.email
+        },
+      }),
```
olivermrbl, line 209: Change from V1: Removed `role`

packages/admin-next/dashboard/src/routes/users/user-edit/index.ts
```diff
@@ -1 +1 @@
-export { UserEdit as Component } from "./user-edit"
+export { UserEdit as Component } from "../../../modules/user/user-edit"
```
riqwan, line 1: q: whats this change for?

packages/admin-next/dashboard/src/v2-routes/users/user-invite/components/invite-user-form/invite-user-form.tsx
```diff
@@ -0,0 +1,381 @@
...
+  const { t } = useTranslation()
+
+  const form = useForm<zod.infer<typeof InviteUserSchema>>({
+    defaultValues: {
+      email: "",
+    },
+    resolver: zodResolver(InviteUserSchema),
+  })
+
+  const { invites, isLoading, isError, error } = useAdminInvites()
```
riqwan, line 59: shouldn't this come from the new lib/v2/* hooks?

riqwan (approved):

packages/admin-next/dashboard/src/routes/users/user-edit/index.ts
olivermrbl: We should strive to move all reusable (between v1 and v2) components to `modules` to reduce the maintenance burden. This change is here because I was able to share the user-edit between the two modes.

packages/admin-next/dashboard/src/v2-routes/users/user-invite/components/invite-user-form/invite-user-form.tsx
olivermrbl: If we can reuse hooks from v1, I don't think we should add new ones. Hopefully, it will be like this for many hooks, as we will not be introducing that many breaking changes. 

However, coming from the promotions domain (which is new), I understand the concern.

packages/admin-next/dashboard/src/routes/users/user-edit/index.ts
riqwan: I see, thats a good idea!

PR #6845, opened 2024-03-27
Fix/pricing data manipulation

adrien2p:
**What**
- Fix returned data order
- improve implentation
- rm unnecessary promises and/or promise all
- replace promise.all for promiseAll

### DIFF
--- a/www/apps/api-reference/specs/admin/openapi.full.yaml
+++ b/www/apps/api-reference/specs/admin/openapi.full.yaml
@@ -34,12 +34,6 @@ tags:
     externalDocs:
       description: How to manage currencies
       url: https://docs.medusajs.com/modules/regions-and-currencies/admin/manage-currencies
-  - name: Customers
-    description: |
-      Customers can either be created when they register through the Store APIs, or created by the admin using the Admin APIs.
-    externalDocs:
-      description: How to manage customers
-      url: https://docs.medusajs.com/modules/customers/admin/manage-customers
   - name: Customer Groups
     description: |
       Customer Groups can be used to organize customers that share similar data or attributes into dedicated groups.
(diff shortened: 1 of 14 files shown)


packages/pricing/src/services/pricing-module.ts
```diff
@@ -247,6 +249,236 @@ export default class PricingModuleService<
     return Array.isArray(data) ? results : results[0]
   }
 
+  async addRules(
```
sradevski, line 252: A general question - why is this called `addRules` and not `createRules` so we are consistent?

packages/pricing/src/services/pricing-module.ts
```diff
@@ -247,6 +249,236 @@ export default class PricingModuleService<
...
+  @InjectManager("baseRepository_")
+  async removePrices(
+    ids: string[],
+    @MedusaContext() sharedContext: Context = {}
+  ): Promise<void> {
+    await this.removePrices_(ids, sharedContext)
+  }
+
+  @InjectManager("baseRepository_")
+  async addPriceListPrices(
```
sradevski, line 451: Same here, why not `create*`?

sradevski (approved):
Quite difficult to follow the changes due to the reordering of the code, but overall it seems fine to me. There is still left to improve probably

packages/pricing/src/services/pricing-module.ts
adrien2p: I must say that i have no idea, there is plenty of improvements left to do. I haven't focused on naming here but indeed consistency is missing here, probably because it was done before the auto generated methods

olivermrbl (approved):

packages/pricing/src/services/pricing-module.ts
olivermrbl: Can maybe be punted to a more appropriate time

riqwan (approved):
🏅

PR #6850, opened 2024-03-27
feat(dashboard): admin 3.0 order refund form

fPolic:
**What**
- added order refund form

### DIFF
--- a/packages/admin-next/dashboard/public/locales/en-US/translation.json
+++ b/packages/admin-next/dashboard/public/locales/en-US/translation.json
@@ -750,9 +750,10 @@
     "invalidInvite": "The invite is invalid or has expired.",
     "successTitle": "Your account has been created",
     "successHint": "Get started with Medusa Admin right away.",
-    "successAction": "Start using Medusa",
+    "successAction": "Sign in to start using Medusa",
     "invalidTokenTitle": "Your invite token is invalid",
-    "invalidTokenHint": "Try requesting a new invite link."
+    "invalidTokenHint": "Try requesting a new invite link.",
+    "passwordMismatch": "Passwords do not match"
   },
   "resetPassword": {
     "title": "Reset password",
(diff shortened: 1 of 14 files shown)


olivermrbl:
What is this dropdown on the actual payment supposed to hold? Something that will come in a follow up?

CleanShot 2024-03-28 at 13 16 12

olivermrbl:
Any way we can have the form render immediately upon opening the refund menu?

See the attached video. It looks like there is a slight delay.

https://github.com/medusajs/medusa/assets/59018053/a950586c-2872-4a1b-be18-4ea75f281733

olivermrbl:
The error message for missing "Reason" should be more sensible from a merchant perspective:

CleanShot 2024-03-28 at 13 23 01

fPolic:
> What is this dropdown on the actual payment supposed to hold? Something that will come in a follow up?

Not sure what was intended but do we add a "Refund" action here as well that will call ` /admin/payments/{id}/refund` for that specific payment? Wdyt?

fPolic:
> Any way we can have the form render immediately upon opening the refund menu?

Fixed. If the modal which is a nested route is fetching the same data with same params as the details page, `react-query` will have this immediately. For cases where different data is needed, we will have to show loader UI.

olivermrbl:
> Not sure what was intended but do we add a "Refund" action here as well that will call  /admin/payments/{id}/refund for that specific payment? Wdyt?

Yes, I actually think this is exactly what we want to have. A "Refund" action on each payment.

packages/medusa/src/services/payment-provider.ts
```diff
@@ -809,6 +809,7 @@ export default class PaymentProviderService extends TransactionBaseService {
 
       const toCreate = {
         payment_id: payment.id,
+        order_id: payment.order_id,
```
fPolic, line 812: note: associate created refund with the order otherwise creating refund with `/admin/payments/:id/refund` will create "orphan" refunds that are not calculated in order totals

olivermrbl: Can I get you to create a ticket for this, so we won't forget?

fPolic: this line should be a fix for the issue, I just added a comment to explain why I made this change

olivermrbl: Ah sorry, I misunderstood that. Thought you meant we need the link.

olivermrbl (approved):
LGTM

PR #6854, opened 2024-03-28
Add discountable-form translations (es and en).

danielrodcaball:
### What

English and Spanish translations to `packages/admin-ui/ui/src/components/forms/product/discountable-form/index.tsx`

### Why

Labels and texts are fixed in English.

### How

Using react-i18next for translations.


danielrodcaball (commented):
-

PR #6855, opened 2024-03-28
chore: remove noop tests

riqwan:
what:

- removes all noop tests and use --passwithnotest flag from jest


sradevski (approved):
Nice cleanup 👍

PR #6860, opened 2024-03-28
chore(workflows-sdk): add idempotencyKey to shared context

carlos-r-l-rodrigues:


adrien2p (approved):
🔥

packages/workflows-sdk/src/utils/composer/create-step.ts
```diff
@@ -121,6 +121,9 @@ function applyStep<
 
     const handler = {
       invoke: async (transactionContext) => {
+        transactionContext.context.idempotencyKey =
+          transactionContext.idempotency_key
```
srindom: Shouldn't it be?
```
transactionContext.metadata.idempotency_key
```

adrien2p: It is in too many places, but it is also in the transaction context i believe 😅

carlos-r-l-rodrigues: I added the correct type to the handler.

PR #6864, opened 2024-03-28
feat: region payment providers management workflows/api

adrien2p:
**What**
- Allow to update a region to provide the payment providers to enable for the given region, act as a upsert and replace at the link level
- Add more integration tests for the API update
- Add workflow integration tests
- fix update region workflow rollback
- fix missing check in workflows
- Allow to create a region and enable payment providers for the given region?
- bulk handling 
- Add integration test for region creation with payment provider through the API or the workflow

**boy scoot**

- Allow to use the service name (especially for link access) directly as an entry point of the remote query


adrien2p:
It is marked as ready to review but I am still working on this pr :D so no need to review for now

packages/core-flows/src/region/steps/upsert-and-replace-region-payment-providers.ts
```diff
@@ -0,0 +1,284 @@
...
+        MedusaError.Types.INVALID_DATA,
+        "Invalid input, must provide either regions or region_selector"
+      )
+    }
+  }
+
+  return regionPaymentProvidersMap
+}
+
+export const upsertAndReplaceRegionPaymentProvidersStepId =
```
olivermrbl: todo: We are not really upserting anything related to providers or region afaict, so let's call this step something else e.g. `setRegionPaymentProvidersStep`
```suggestion
export const setRegionPaymentProvidersStep =
```

packages/core-flows/src/region/steps/upsert-and-replace-region-payment-providers.ts
```diff
@@ -0,0 +1,284 @@
...
+import {
+  arrayDifference,
+  ContainerRegistrationKeys,
+  MedusaError,
+  promiseAll,
+  remoteQueryObjectFromString,
+} from "@medusajs/utils"
+
+export interface UpsertAndReplaceRegionPaymentProvidersStepInput {
+  input:
```
olivermrbl: suggestion: ~~`input` doesn't need to be an array~~ resolved

packages/core-flows/src/region/steps/upsert-and-replace-region-payment-providers.ts
```diff
@@ -0,0 +1,284 @@
...
+    | {
+        region_selector: FilterableRegionProps
```
olivermrbl: suggestion:
```suggestion
        selector: FilterableRegionProps
```

packages/core-flows/src/region/steps/upsert-and-replace-region-payment-providers.ts
```diff
@@ -0,0 +1,284 @@
...
+        payment_providers: string[]
+      }
+    | {
+        regions: { id: string }[]
```
olivermrbl: suggestion:
```suggestion
        ids: string[]
```

packages/core-flows/src/region/steps/upsert-and-replace-region-payment-providers.ts
```diff
@@ -0,0 +1,284 @@
...
+        payment_providers: string[]
+      }[]
+}
+
+type LinkItems = {
+  [Modules.REGION]: { region_id: string }
+  [Modules.PAYMENT]: { payment_provider_id: string }
+}[]
+
+async function validatePaymentProvidersExists(
```
olivermrbl: question: Should we put some of these functions in a `utils` folder?

packages/core-flows/src/region/workflows/create-regions.ts
```diff
@@ -1,13 +1,76 @@
...
+
+      return {
+        regions: regionsData,
+        regions_providers_map: regionsProvidersMap,
+      }
+    })
+
+    const regions = createRegionsStep(data.regions)
+
+    const normalizedRegionProviderData = transform(
```
olivermrbl, line 34: todo: Can we make this easier to read? I think the change of `input` would simplify it a bit.

Can we get rid of the map? You would never update an amount of regions that could cause a bottleneck on performance.

olivermrbl (commented):
Nice work, I left a couple of comments and thoughts.

packages/core-flows/src/region/steps/upsert-and-replace-region-payment-providers.ts
olivermrbl: todo: Create separate types for `input`. This is hard to read

packages/core-flows/src/region/steps/upsert-and-replace-region-payment-providers.ts
```diff
@@ -0,0 +1,284 @@
...
+  regionIds: string[],
+  { remoteQuery }: { remoteQuery: RemoteQueryFunction }
+): Promise<
+  {
+    [Modules.REGION]: { region_id: string }
+    [Modules.PAYMENT]: { payment_provider_id: string }
+  }[]
+> {
+  const query = remoteQueryObjectFromString({
+    // TODO: use service: LINKS.RegionPaymentProvider, for that, the links.ts must be moved to the utils package, so that for link querying we can use link const instead of plain strings
```
olivermrbl: suggestion: Should we go ahead and do that now? Seems like a minor change, so I think it's OK to include it

packages/core-flows/src/region/steps/upsert-and-replace-region-payment-providers.ts
```diff
@@ -0,0 +1,284 @@
...
+import { createStep, StepResponse } from "@medusajs/workflows-sdk"
+import {
+  arrayDifference,
+  ContainerRegistrationKeys,
+  MedusaError,
+  promiseAll,
+  remoteQueryObjectFromString,
+} from "@medusajs/utils"
+
+export interface UpsertAndReplaceRegionPaymentProvidersStepInput {
```
olivermrbl: suggestion: Would it make sense to simplify the input to:
```ts
interface RegionPPInput {
  id: string
  payment_providers: string[]
}

interface StepInput {
  input: RegionPPInput[]
}
```

Can we achieve what we want to?

adrien2p: Yes we can definitely, in the case the step becomes again just a setter kind of and the region validation/querying falls back in the hands of the consumer, which is not bad i believe

riqwan:
@adrien2p is this ready for review?

adrien2p:
> @adrien2p is this ready for review?

Not yet i have some changes to make tomorrow

packages/core-flows/src/region/workflows/create-regions.ts
adrien2p: I don't think we can simplify, the API allows us to create many regions, each of those regions can have a set of providers, so I am creating a hash for each region before creation which is associated to the set of providers for that region. After the creation, I need to re associate the right region id based on the input hash of the corresponding region to grab the corresponding set of providers. we can discuss it if you have another idea

packages/core-flows/src/region/steps/upsert-and-replace-region-payment-providers.ts
adrien2p: I think to not increase the PR size, I ll move the links and modules to the utils, I ll update the link usage but not the Modules enum to not impact all modules, I ve put a todo to say that we should now use the Modules that comes from the utils package instead of the one from the modules-sdk

packages/core-flows/src/region/workflows/create-regions.ts
sradevski: We have quite a few scenarios like this, and we'll have a lot more, in some cases where the hash approach doesn't work.

For the products and prices, I rely on the indices, and I know that is fine because the API guarantees same input/output ordering. What if we set that as a requirement on modules? Then relying on indices would be safe to do, and have the code a lot simpler

sradevski (approved):

packages/core-flows/src/region/workflows/create-regions.ts
adrien2p: yes we can go with that approach indeed 👍

adrien2p: done @sradevski

adrien2p:
@riqwan you can now 👍

adrien2p:
@olivermrbl would you have time to have another look at this one 👍

packages/types/src/payment/common.ts
```diff
@@ -1,5 +1,5 @@
 import { BaseFilterable } from "../dal"
-import { OperatorMap } from "../dal/utils"
+import { OperatorMap } from "../dal/utils" /* ********** PAYMENT COLLECTION ********** */
```
olivermrbl: What happened here? 😄

olivermrbl (approved):
LGTM, nice work @adrien2p

packages/types/src/payment/common.ts
adrien2p: Oh 😂 didn't see that ahah, ill fix that

packages/core-flows/src/region/workflows/create-regions.ts
```diff
@@ -1,13 +1,57 @@
...
+      {
+        regionIndexToPaymentProviders: data.regionIndexToPaymentProviders,
+        regions,
+      },
+      (data) => {
+        return data.regionIndexToPaymentProviders.map(
+          ({ region_index, payment_providers }) => {
+            return {
+              id: data.regions[region_index].id,
+              payment_providers: payment_providers ?? [],
```
adrien2p: Todo: no default empty array here as it would remove all links

packages/core-flows/src/region/steps/set-regions-payment-providers.ts
```diff
@@ -0,0 +1,205 @@
...
+      regionIds,
+      { remoteQuery }
+    )
+
+    const linksToRemove = currentExistingLinks
+      .filter((existingLink) => {
+        return !data.input.some((input) => {
+          return (
+            input.id === existingLink[Modules.REGION].region_id &&
+            input.payment_providers.includes(
```
adrien2p, line 140: Todo: optional chaining

packages/core-flows/src/region/steps/set-regions-payment-providers.ts
```diff
@@ -0,0 +1,205 @@
...
+  MedusaError,
+  Modules,
+  promiseAll,
+  remoteQueryObjectFromString,
+} from "@medusajs/utils"
+
+export interface SetRegionsPaymentProvidersStepInput {
+  input: {
+    id: string
+    payment_providers: string[]
```
adrien2p: Todo: accept undefined, meaning dont do anything

carlos-r-l-rodrigues (approved):
🔥

PR #6865, opened 2024-03-28
chore: medusa shutdown

adrien2p:
[WIP]

**What**
- include a new shutdown function returned by medusa loader that should clean up internal resources if needed.
- include onApplicationShutdown hook for the module to have medusa app handling shutdown properly and allow any modules to hook themself to clean up potential resources if needed

**NOTE**
We can rework the naming if necessary

**NOTE**
Now, when running the tests with `medusaTestRunner` there is no more open handles after the tests finishes, meaning that we shouldn't have any open tcp left and no need for forceExit.

Same for graceful shutdown, and if any users consume our shutdown, the result should be the same


adrien2p:
I believe it is already a very good start

carlos-r-l-rodrigues (approved):

fPolic (approved):

riqwan (approved):

PR #6866, opened 2024-03-28
feat: Refactor the product module definitions and implementation

sradevski:
There are several things done in this PR, namely:

Unify the service endpoints API to always work with a model rather than allowing to pass both ID and model (eg. both type_id and type being available in the request to create).
Start using upsertWithReplace to simplify the code and fix some deassociation bugs
Apply some changes to tests to deal with the pricing breaking changes
Correctly define the model relationships (with both ID and entity fields available)
All tests for the product are passing, which should bring us back to a great baseline.


integration-tests/api/__tests__/admin/product.js
```diff
@@ -32,6 +32,50 @@ let {
 
 jest.setTimeout(50000)
 
+const getProductFixture = () => ({
```
olivermrbl, line 35: question: The reason I moved this into the `medusaIntegrationTestRunner` a few PRs ago was that the `breaking` helper only works outside of the runner, if you use `process.env`. Passing the V2 FF in `env` will not work as expected (I am quite sure).

sradevski: @olivermrbl I now changed it to a function, so it will be evaluated when called (after everything is loaded). In that case I believe it should work even if you pass the var in `env`, (since at the end it gets set on process.env).

olivermrbl: Ah sorry, I missed this part of the change!

packages/product/src/services/product.ts
```diff
@@ -1,11 +1,23 @@
-import { Context, DAL, FindConfig, ProductTypes } from "@medusajs/types"
+import {
```
olivermrbl, line 1: Hard to review this file, because of all the rearrangement. I think we should get this to land on `develop` now, and then we can always clean up (if needed), as we use the module.

olivermrbl (approved):
Very nice clean-up of the module service. I don't have any comments right now. Let's get this in and address any issues in follow-up PRs?

packages/product/src/services/product.ts
sradevski: Yeah it was a patch I decided to pull now, there was no better way around it. I'll get this merged ASAP 👍

PR #6869, opened 2024-03-28
fix(workflow-engine-*): subscribe response and error

carlos-r-l-rodrigues:


packages/workflow-engine-inmemory/src/services/workflow-orchestrator.ts
```diff
@@ -463,29 +463,33 @@ export class WorkflowOrchestratorService {
...
       },
       onStepFailure: ({ step, transaction }) => {
+        const stepName = step.definition.action!
         const errors = transaction.getErrors(TransactionHandlerType.INVOKE)[
-          step.id
+          stepName
         ]
         customEventHandlers?.onStepFailure?.({ step, transaction, errors })
 
         notify({ eventType: "onStepFailure", step, errors })
```
srindom: still not seeing errors in the events

packages/workflow-engine-inmemory/src/services/workflow-orchestrator.ts
```diff
@@ -463,29 +463,33 @@ export class WorkflowOrchestratorService {
         notify({ eventType: "onStepBegin", step })
       },
       onStepSuccess: ({ step, transaction }) => {
-        const response = transaction.getContext().invoke[step.id]
+        const stepName = step.definition.action!
+        const response = transaction.getContext().invoke[stepName]
```
srindom: if this is WorkflowData we should clean it up so we only send back the output.

srindom (approved):
LGTM

PR #6870, opened 2024-03-28
feat(dashboard): Variant edit form

kasperkristensen:
**What**
- Adds form to edit variant details.
- If stock and inventory modules are installed we omit inputs for stock and inventory as they should be managed in a separate form in that case. (Still a bit unsure about this, but its what we have in the current admin)


riqwan (approved):
🏅

packages/admin-next/dashboard/src/routes/products/product-edit-variant/components/product-edit-variant-form/product-edit-variant-form.tsx
```diff
@@ -0,0 +1,454 @@
...
+  ean: z.string().optional(),
+  upc: z.string().optional(),
+  barcode: z.string().optional(),
+  inventory_quantity: optionalInt,
+  manage_inventory: z.boolean(),
+  allow_backorder: z.boolean(),
+  weight: optionalInt,
+  height: optionalInt,
+  width: optionalInt,
+  length: optionalInt,
```
fPolic, line 39: nit: we would need to validate that it's a positive int as well

packages/admin-next/dashboard/src/routes/products/product-edit-variant/components/product-edit-variant-form/product-edit-variant-form.tsx
```diff
@@ -0,0 +1,454 @@
...
+
+    await mutateAsync({
+      variant_id: variant.id,
+      weight: parseNumber(weight),
+      height: parseNumber(height),
+      width: parseNumber(width),
+      length: parseNumber(length),
+      ...conditionalPayload,
+      ...rest,
+    })
```
fPolic: todo: missing `handleSuccess` to close the modal

packages/admin-next/dashboard/src/routes/products/product-edit-variant/product-edit-variant.tsx
```diff
@@ -0,0 +1,57 @@
...
+  )
+
+  if (!isLoading && !variant) {
+    throw json({
+      status: 404,
+      message: `Variant with ID ${variant_id} was not found.`,
+    })
+  }
+
+  const ready = !isLoading && !!product && !!variant
```
fPolic, line 37:
```suggestion
  const ready = !isLoading && !!variant
```

packages/admin-next/dashboard/src/routes/products/product-edit-variant/loader.ts
```diff
@@ -0,0 +1,40 @@
...
+import { LoaderFunctionArgs } from "react-router-dom"
+
+import { medusa, queryClient } from "../../../lib/medusa"
+
+const queryKey = (id: string) => {
+  return [adminProductKeys.detail(id), adminStoreKeys.details()]
+}
+
+const queryFn = async (id: string) => {
+  const productRes = await medusa.admin.products.retrieve(id)
```
fPolic, line 11: suggestion: do we need to fetch the product here...seems that fetching only the variant we are editing is sufficient

packages/admin-next/dashboard/src/routes/products/product-edit-variant/components/product-edit-variant-form/product-edit-variant-form.tsx
```diff
@@ -0,0 +1,454 @@
...
+      barcode,
+      ...rest
+    } = data
+
+    /**
+     * If stock and inventory is not enabled, we need to send the inventory and
+     * stock related fields to the API. If it is enabled, it should be handled
+     * in the separate stock and inventory form.
+     */
+    const conditionalPayload = !isStockAndInventoryEnabled
```
fPolic, line 110: q: Didn't we discuss that we assume that inventory/stock location modules are enabled for this version? Asking because I have similar situations with fulfilment and I'm not sure about the approach.

fPolic (approved):
LGTM!
I've added a couple of minor comments.

packages/admin-next/dashboard/src/routes/products/product-edit-variant/components/product-edit-variant-form/product-edit-variant-form.tsx
olivermrbl: I believe this was for 2.0. At least I think it should be. I don't think we can enforce this, as that would be a significant upgrade for users to go through to get a new version of admin. And we will communicate that the admin redesign is fully compatible with setups that run on 1.*, which includes simple inventory management, so having to switch to modules would be a big caveat to that 😄

packages/admin-next/dashboard/src/routes/products/product-edit-variant/product-edit-variant.tsx
kasperkristensen: TS is not smart enough to extract that the product is defined if variant also is, so we need both here :D

packages/admin-next/dashboard/src/routes/products/product-edit-variant/loader.ts
kasperkristensen: Yes, we need it for the options editing (not implemented here yet, as it requires adding a "Create <new option value>" feature in the select component). The variant does not hold the name, or the other options in product.

packages/admin-next/dashboard/src/routes/products/product-edit-variant/components/product-edit-variant-form/product-edit-variant-form.tsx
kasperkristensen: I understood it as it not being needed for the alpha/beta version. But we should have support for both cases for the final release, and I added it here as the check is rather simple.

PR #6872, opened 2024-03-28
feat: Implemented price set update with prices and aligned pricing API

sradevski:


packages/pricing/integration-tests/__tests__/services/pricing-module/price-set.spec.ts
```diff
@@ -262,24 +262,67 @@ moduleIntegrationTestRunner({
         })
       })
 
-      describe("update", () => {
+      describe.only("update", () => {
```
olivermrbl:
```suggestion
      describe("update", () => {
```

packages/pricing/src/services/pricing-module.ts
```diff
@@ -246,7 +251,104 @@ export default class PricingModuleService<
...
+
+    return isString(idOrSelector) ? priceSets[0] : priceSets
+  }
+
+  @InjectTransactionManager("baseRepository_")
+  protected async update_(
+    data: UpdatePriceSetInput[],
+    @MedusaContext() sharedContext: Context = {}
+  ): Promise<PriceSet[]> {
+    // TODO: We are not handling rule types, rules, etc. here, add support after data models are finalized
```
olivermrbl, line 344: question: What more is missing about the data models?

packages/types/src/pricing/service.ts
```diff
@@ -602,18 +603,120 @@ export interface IPricingModuleService extends IModuleService {
...
+   *     {
+   *       prices: [{amount: 100, currency_code: "USD"}]
+   *     }
+   *   )
+   *
+   *   // do something with the price set or return it
+   * }
+   */
+  upsert(
+    data: UpsertPriceSetDTO[],
```
olivermrbl:
```suggestion
    data: UpsertPriceSetDTO,
```

packages/types/src/pricing/service.ts
```diff
@@ -602,18 +603,120 @@ export interface IPricingModuleService extends IModuleService {
...
+    sharedContext?: Context
+  ): Promise<PriceSetDTO[]>
```
olivermrbl:
```suggestion
  ): Promise<PriceSetDTO>
```

packages/pricing/src/services/pricing-module.ts
sradevski: @olivermrbl rules are not handled at all for price sets currently, among other things. I think it's easier to see the `create` method to see what else is missing, there is a lot going on there that I wouldn't do before Recap

packages/types/src/pricing/service.ts
sradevski: Ah nice catch 👍

olivermrbl (approved):
Mergeable after comments have been addressed 👍

PR #6880, opened 2024-03-29
feat: Add v2 product types endpoints

sradevski:
Also adjusts the product type module APIs to follow the conventions

### DIFF
--- a/packages/medusa/src/api-v2/admin/product-types/[id]/route.ts
+++ b/packages/medusa/src/api-v2/admin/product-types/[id]/route.ts
@@ -9,6 +9,7 @@ import {
 
 import { UpdateProductTypeDTO } from "@medusajs/types"
 import { remoteQueryObjectFromString } from "@medusajs/utils"
+import { refetchProductType } from "../helpers"
 
 export const GET = async (
   req: AuthenticatedMedusaRequest,
@@ -45,7 +46,13 @@ export const POST = async (
     throw errors[0].error
   }
 
-  res.status(200).json({ product_type: result[0] })
+  const productType = await refetchProductType(
+    result[0].id,
+    req.scope,
+    req.remoteQueryConfig.fields
+  )
+
+  res.status(200).json({ product_type: productType })
 }
 
 export const DELETE = async (
--- a/packages/medusa/src/api-v2/admin/product-types/route.ts
+++ b/packages/medusa/src/api-v2/admin/product-types/route.ts
@@ -12,6 +12,7 @@ import {
   AdminPostProductTypesReq,
 } from "./validators"
 import { createProductTypesWorkflow } from "@medusajs/core-flows"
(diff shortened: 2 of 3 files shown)


packages/medusa/src/api-v2/admin/product-types/query-config.ts
```diff
@@ -0,0 +1,22 @@
...
+  "created_at",
+  "updated_at",
+]
+
+export const retrieveProductTypeTransformQueryConfig = {
+  defaultFields: defaultAdminProductTypeFields,
+  defaultRelations: defaultAdminProductTypeRelations,
+  allowedRelations: allowedAdminProductTypeRelations,
+  isList: false,
+}
```
olivermrbl: Should we use the new `defaults`?

packages/medusa/src/api-v2/admin/product-types/route.ts
```diff
@@ -0,0 +1,56 @@
...
+  const { result, errors } = await createProductTypesWorkflow(req.scope).run({
+    input: { product_types: input },
+    throwOnError: false,
+  })
+
+  if (Array.isArray(errors) && errors[0]) {
+    throw errors[0].error
+  }
+
+  res.status(200).json({ product_type: result[0] })
```
olivermrbl: suggestion: Use remote query

packages/medusa/src/api-v2/admin/product-types/[id]/route.ts
```diff
@@ -0,0 +1,71 @@
...
+      update: req.validatedBody,
+    },
+    throwOnError: false,
+  })
+
+  if (Array.isArray(errors) && errors[0]) {
+    throw errors[0].error
+  }
+
+  res.status(200).json({ product_type: result[0] })
```
Comment (olivermrbl) [inline packages/medusa/src/api-v2/admin/product-types/[id]/route.ts]: suggestion: Use remote query

packages/medusa/src/api-v2/admin/product-types/route.ts
sradevski: @olivermrbl do you mean to refetch the data? there is no need in this case I believe, as the `type` model doesn't have any relations and it only has 2 fields to modify, but I am happy to add it if you think we should

olivermrbl: It would be for the sake of consistency. If someone links to a `product_type_id`, they won't be able to query the relations on create, as in most other endpoints. This is an edge case, so fine to keep it out. I just like the idea of having the same query API in all APIs, so we won't ever be in a situation where someone asks for it and we need to add it back in.

sradevski: I was being thoughtful about the amount of requests, but you won't be creating types as much anyway, so  I can add it 👍

olivermrbl (approved):
LGTM!

PR #6886, opened 2024-03-29
feat(workflows-sdk,orchestration): async step as background task

carlos-r-l-rodrigues:
**What:**

When a step flagged as `async: true` returns a `new StepResponse` itp will be marked as success automatically after its execution.

If it doesn't return a `StepResponse` instance, it is considered a regular async step that will require `setStepSuccess` or `setStepFailure` to called later.

```typescript
createStep(
  {
    name: "step_background",
    async: true,
  },
  jest.fn(async (input) => {
    await setTimeout(200)

    return new StepResponse(input)
  })
)
```

Regular:
```typescript
createStep(
  {
    name: "step_background",
    async: true,
  },
  jest.fn(async (input) => {
    await setTimeout(200)
  })
)
```


adrien2p (approved):
Looks good to me, so if we return a step response, after the execution we mark it as success and we set the returned response, otherwise we expect a manual action as usual right?

srindom:
> Looks good to me, so if we return a step response, after the execution we mark it as success and we set the returned response, otherwise we expect a manual action as usual right?

Exactly 👍

srindom:
I am seeing this error after trying the snapshot:

CleanShot 2024-04-01 at 9  39 58

carlos-r-l-rodrigues:
> I am seeing this error after trying the snapshot:
> 
> CleanShot 2024-04-01 at 9  39 58

Apparently your orchestration package is an old version comming from a node_modules inside modules-sdk

carlos-r-l-rodrigues:
@srindom can we merge this?

srindom:
Yep! Lgtm

PR #6890, opened 2024-03-31
Update deploying-on-railway.md

brody192:
This removes the nixpacks.toml that has broken many medusa builds when deploying to Railway, and provides extra clarity on what `railway run` does.

The nixpacks.toml that was presented in these docs fixes node and yarn at old versions and breaks builds in many strange ways, not to mention that it uses incorrect syntax.

The project that `create-medusa-app` creates comes with only a v1 yarn lock file, nixpacks will detect this automatically and use yarn for the build.

### DIFF
--- a/www/apps/docs/content/deployments/server/deploying-on-railway.md
+++ b/www/apps/docs/content/deployments/server/deploying-on-railway.md
@@ -114,23 +114,7 @@ Refer to the [admin deployment guides on how to deploy the admin separately](../
 
 ---
 
-## (Optional) Step 1: Add Nixpacks Configurations
-
-If you've created your project using `create-medusa-app`, you might receive errors during the deployment process as Railway uses NPM by default. To avoid that, you need to configure Nixpacks to either use `yarn` or add the `--legacy-peer-deps` option to `npm install`.
-
-In the root of your Medusa project, add the `nixpacks.toml` file with the following content:
-
-```toml
-[phases.setup]
-nixPkgs = ['nodejs', 'yarn']
-
-[phases.install]
-cmds=['yarn install']
-```
-
----
-
-## Step 2: Add Worker Mode Configuration
+## Step 1: Add Worker Mode Configuration
 
 :::note
 
@@ -152,7 +136,7 @@ This allows you to switch between modes for different deployed Medusa instances 
 
 ---
 
-## Step 3: Create GitHub Repository
+## Step 2: Create GitHub Repository
 
 Before you deploy your Medusa backend you need to create a GitHub repository and push the code base to it.
 
@@ -187,7 +171,7 @@ After pushing the changes, you can find the files in your GitHub repository.
 
 ---
 
-## Step 4: Deploy to Railway
+## Step 3: Deploy to Railway
 
 In this section, you’ll create the PostgreSQL and Redis databases first, then deploy two instances of the Medusa backend: one having a `server` runtime mode, and another having a `worker` runtime mode.
 
@@ -337,7 +321,7 @@ The last step is to add a domain name to your Medusa backend. To do that:
 
 ---
 
-## Step 5: Test the Backend
+## Step 4: Test the Backend
 
 Every change you make to the settings redeploys the backend. You can check the Deployments of the backend by clicking on the GitHub repository’s card and choosing the Deployments tab.
 
@@ -373,11 +357,11 @@ This error may be thrown by a module that uses Redis. If you see it in your buil
 
 ---
 
-## Run Commands on the Backend
+## Run Commands locally with the service variables available
 
-To run commands on your backend, you can use [Railway’s CLI tool to run a local shell and execute commands](https://docs.railway.app/develop/cli#local-shell).
+To run commands with your service variables available, you can use [Railway’s CLI tool to run a local shell and execute commands](https://docs.railway.app/develop/cli#local-shell).
 
-For example, you can run commands on the backend to seed the database or create a new user using [Medusa’s CLI tool](../../cli/reference.mdx).
+For example, you can run commands to seed the database or create a new user using [Medusa’s CLI tool](../../cli/reference.mdx).
 
 ### Create Admin User
 


www/apps/docs/content/deployments/server/deploying-on-railway.md
```diff
@@ -373,11 +357,11 @@ This error may be thrown by a module that uses Redis. If you see it in your buil
 
 ---
 
-## Run Commands on the Backend
+## Run Commands locally with the service variables available
```
shahednasser, line 360: I don't think this title makes things clearer. Maybe:

```suggestion
## Run Commands Locally on the Deployed Backend
```

And then in the following text we mention that to run commands on your deployed backend, you run the command locally using Railway's CLI tool

shahednasser (changes requested):
Thank you for submitting this PR. However, I don't support removing the nixpacks configuration step. It's highly recommended to use `yarn` for installation, or as mentioned in the docs at least pass `--legacy-peer-deps` to `npm install`. So, I'd prefer we keep that.

On the other hand, while going through the deployment guide and while testing out certain things, I noticed the following:

1. When installing with yarn v1, which is what nixpacks chooses by default, there's a typescript error on build. Seems like yarn v1 installs the incorrect version of `@types/mime`. The fix is to install `@types/mime@1.3.5`. While I pushed this to the starter repository (which medusa installations are cloned from), we have to add a troubleshooting guide for devs with existing projects that follow this guide.
2. In the section about the admin, the second option "Deploying Separately" doesn't mention removing the `build:admin` command from the `build` command, which is necessary to ensure the admin build doesn't hog the resources.

Would you be interested in making these changes? Point 1 requires adding a new section under the Troubleshooting section, and point 2 requires mentioning under the "Deploying separately" section removing `build:admin` from the `build` script (with a code example showing it)

brody192:
>I don't support removing the nixpacks configuration step.

At this point, I think it would be beneficial to both of us if I mention that I'm a community moderator for Railway, I have helped many people deploy their medusa apps to Railway simply by asking them to remove the nixpacks.toml file provided in the docs.

>It's highly recommended to use yarn for installation

And yarn will be used, the default repository for `create-medusa-app` has a [yarn lock file that is version 1](https://github.com/medusajs/medusa-starter-default/blob/master/yarn.lock#L2C3-L2C19) thus yarn will automatically be used when deploying to Railway.

>or as mentioned in the docs at least pass --legacy-peer-deps to npm install. So, I'd prefer we keep that.

As long as yarn is being used (it would be) is there now a need for this? if there is a need it's my opinion that such issues should be fixed in the starter repository.

>Would you be interested in making these changes?

I'm sorry but I don't have the time, I only wanted to remove the nixpacks.toml file that uses incorrect syntax and breaks builds, along with removing misleading information on `railway run`

www/apps/docs/content/deployments/server/deploying-on-railway.md
brody192: "Run Commands Locally on the Deployed Backend" is an incredibly misleading sentence, nothing is ever "ran on the backend" and Railway provides no such way to do that, it needs to be made known that these commands are only ever run locally.

devonsydney:
> I don't support removing the nixpacks configuration step. 

I will raise my hand to say that I was one of the people Brody (graciously) helped get my application back up and running in production by removing the misleading nixpacks.toml that was supplied in the Medusa guide. Lost many hours of productivity and had a production deploy I couldn't modify until that simple suggestion was made. Everything quickly fell into place after that. So I would agree with Brody that it should be removed, or at least language on the dangers of using the Medusa recommended .toml.

If yo remove the .toml, your nixpacks build settings will look something like this (I set node engine to 20 in my `package.json`):

```
╔════════════════ Nixpacks v1.21.2 ═══════════════╗
║ setup      │ nodejs_20, yarn-1_x                ║
║─────────────────────────────────────────────────║
║ install    │ yarn install --frozen-lockfile     ║
║─────────────────────────────────────────────────║
║ build      │ yarn run build                     ║
║─────────────────────────────────────────────────║
║ start      │ yarn run start                     ║
╚═════════════════════════════════════════════════╝
```

> 1. When installing with yarn v1, which is what nixpacks chooses by default, there's a typescript error on build. Seems like yarn v1 installs the incorrect version of `@types/mime`. The fix is to install `@types/mime@1.3.5`. While I pushed this to the starter repository (which medusa installations are cloned from), we have to add a troubleshooting guide for devs with existing projects that follow this guide.

I also bashed my head against this error for a bit. The fix for me was NOT to install @types/mime@1.3.5, but rather to simply add the `NIXPACKS_NO_CACHE=1` service variable to my deployed application and redeploy. Then I could remove the `NIXPACKS_NO_CACHE` service variable again and it continues to deploy fine. Node magic.

shahednasser:
Thanks for both of your inputs. There's one thing to point out though: by default, `yarn.lock` and `package.json` are both ignored. So, if I push the changes as-is after `create-medusa-app` (which I did to test out the steps mentioned by @brody192 ) it leads to `npm` being used. Now, we can say here in the docs to remove `yarn.lock` from `.gitignore`, so I'm fine with this option.

However, @brody192 can you confirm what's wrong with using nixpacks? The only reason i could get the deployment working (without committing `yarn.lock`) was using nixpacks 😅 otherwise, I have problems with the deployment. I think the only reason people had previously problems with it due to the `mime` error, and @devonsydney while I understand it worked for you, the solution of setting `NIXPACKS_NO_CACHE` isn't enough because I discovered that the issue was happening locally too. So, we definitely need `@types/mime@1.3.5` installed to support yarn v1 (locally or in deployments)

I respect both of your experiences in this 😄 but before making any change to the deployment guide I would love to know all the available options, as we have to consider that some people my have my case, and some people may have yours. The documentation is supposed to provide solutions that can work for everyone.

To sum it up @brody192 if you can please explain to me that exact problem with using nixpacks (other than the `mime` issue) then that would be great. I think we can change the step to remove `yarn.lock` from `.gitignore`, but we still need to add a troubleshooting guide as this may still cause errors for anyone using yarn v1. I can take this one on (or both on really if you're not interested in adjusting the contribution).

www/apps/docs/content/deployments/server/deploying-on-railway.md
shahednasser: Then I must've misunderstood the use of the Railway CLI. The point of this section is to point users to how they can run command on their deployed backend, such as to create a user. What in your opinion is the alternative here?

brody192: The help text for `railway run` is as follows "Run a local command using variables from the active environment"

So to say "Run Commands on the Deployed Backend" is misleading, but I'm not saying it won't work, just that the wording needs to change.

brody192:
>Now, we can say here in the docs to remove yarn.lock from .gitignore, so I'm fine with this option.

Yep it would be best if that was mentioned in the docs and also removed from the starter repo's .gitignore

>can you confirm what's wrong with using nixpacks

it's not wrong to use a nixpacks.toml file, but the file you provided used incorrect syntax and pinned node and yarn at old versions and doesn't allow for adding on extra nix packages onto what's installed by default, I don't know how it ever worked, to be honest.

>So, we definitely need @types/mime@1.3.5 installed to support yarn v1

It would be good to mention that in your docs so that I can direct users of Railway to the medusa docs to lessen my support load.

If yarn should be used then the proper solution would be to instruct users to generate a yarn lock file, not add an uneeded nixpacks.toml file because that's just a work around.

shahednasser:
> it's not wrong to use a nixpacks.toml file, but the file you provided used incorrect syntax and pinned node and yarn at old versions and doesn't allow for adding on extra nix packages onto what's installed by default, I don't know how it ever worked, to be honest.

It worked fine for me 😄 it also uses the syntax from [nixpacks documentation](https://nixpacks.com/docs/configuration/file). Can you clarify why the syntax is wrong?

EDIT: should we instead use `railway.toml` as instructed [here](https://docs.railway.app/reference/config-as-code)? I think this documentation wasn't around when I first wrote the deployment documentation.

My main concern with asking users to commit their `yarn.lock` is that if they don't use `yarn`, they'll have to install it locally and start using it. While it's more recommended to use yarn, we still don't force users to use it. It's also possible to use `npm` with railway, but they need to pass the `--legacy-peer-deps` option to the installation command as sometimes it causes issues otherwise.

So if we can fix the `nixpacks` file then that would be a perfect solution in my opinion. Otherwise we can look at the alternative.

> It would be good to mention that in your docs so that I can direct users of Railway to the medusa docs to lessen my support load.

Yup, I'm planning on adding a general troubleshooting guide. The only reason we didn't have it is that i wasn't aware of it.

In all cases, if you'd rather not make any changes to this PR, feel free to close it. I can work on another PR with a good solution, but would be good to provide your input on the problem with the nixpacks file (as I personally can't find/reproduce any issues in it).

PR #6892, opened 2024-04-01
feat: Pricing update and refactor product API tests to not rely on internals

sradevski:
This PR achieves the following
- Remove dependency on internals for seeding the tests (more work left, but major work done)
- Adds the workflow for updating variant's price

I will do a follow-up PR to further clean up the tests and remove all internal dependencies

### DIFF
--- a/packages/core-flows/src/product/steps/get-variant-pricing-link.ts
+++ b/packages/core-flows/src/product/steps/get-variant-pricing-link.ts
@@ -1,0 +1,50 @@
+import { ModuleRegistrationName, Modules } from "@medusajs/modules-sdk"
+import { IProductModuleService } from "@medusajs/types"
+import {
+  ContainerRegistrationKeys,
+  MedusaError,
+  arrayDifference,
+} from "@medusajs/utils"
+import { StepResponse, createStep } from "@medusajs/workflows-sdk"
+
+type StepInput = {
+  ids: string[]
+}
+
+export const getVariantPricingLinkStepId = "get-variant-pricing-link"
+export const getVariantPricingLinkStep = createStep(
+  getVariantPricingLinkStepId,
+  async (data: StepInput, { container }) => {
+    if (!data.ids.length) {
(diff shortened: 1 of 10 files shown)


packages/core-flows/src/product/steps/get-variant-pricing-link.ts
```diff
@@ -0,0 +1,41 @@
...
+
+    const existingItems = await linkService.list(
+      { variant_id: data.ids },
+      { select: ["variant_id", "price_set_id"] }
+    )
+
+    if (existingItems.length !== data.ids.length) {
+      throw new MedusaError(
+        MedusaError.Types.NOT_FOUND,
+        "Some variant pricing links were not found"
```
olivermrbl: suggestion: `Variants with IDs ${...} do not have prices`

packages/core-flows/src/product/workflows/update-product-variants.ts
```diff
@@ -15,6 +23,72 @@ export const updateProductVariantsWorkflow = createWorkflow(
...
+      },
+      (data) => {
+        return data.updatedVariants.map((variant, i) => {
+          const linkForVariant = data.variantPriceSetLinks.find(
+            (link) => link.variant_id === variant.id
+          )
+
+          const priceSetForVariant = data.updatedPriceSets.find(
+            (priceSet) => priceSet.id === linkForVariant?.price_set_id
+          )
```
olivermrbl, line 87: nit: convert to maps, this is one of those potentially heavy operations

olivermrbl (approved):
Nicr work – overall LGTM

packages/core-flows/src/product/workflows/update-product-variants.ts
sradevski: @olivermrbl I'd like to do some benchmarking and optimization as a separate task if you agree (I'm taking notes on this)

PR #6893, opened 2024-04-01
fix: ensure that shutdown works in loaderOnly mode

srindom:


packages/modules-sdk/src/medusa-module.ts
```diff
@@ -262,14 +264,13 @@ export class MedusaModule {
     let finishLoading: any
     let errorLoading: any
 
+    const loadingPromise = new Promise((resolve, reject) => {
```
sradevski, line 267: Since we are in an async function, can't we just `return` or `throw` instead of this pattern? I guess we want to allow others to be able to wait on a module to get loaded, it just feels like there is a better way to achieve it 😄

sradevski (approved):
LGTM!

packages/modules-sdk/src/medusa-module.ts
carlos-r-l-rodrigues: This prevents the same module being initialized multiple times while loading, like if you load it in a lambda handler and that is called multiple times in concurrency.
There isn't any other way to achieve this in a single process other than sharing a single Promise.

adrien2p (approved):

PR #6895, opened 2024-04-01
chore: added TSDocs to auth_cors configuration

shahednasser:
Added description of `auth_cors` configuration as TSDoc comment


sradevski (approved):
I guess we also need to update create-medusa-app to set the AUTH_CORS, but the docs look good to me

shahednasser:
@sradevski I've actually added the following configuration to the v2 starter branch:

```ts
auth_cors: process.env.AUTH_CORS || ADMIN_CORS
```

I think for internal use cases (which is what the v2 flag is for), the auth CORS would be the admin CORs, but you'd still be able to set it in the environment variable. Let me know if you think otherwise though 😄

sradevski:
> @sradevski I've actually added the following configuration to the v2 starter branch:
> 
> ```ts
> auth_cors: process.env.AUTH_CORS || ADMIN_CORS
> ```
> 
> I think for internal use cases (which is what the v2 flag is for), the auth CORS would be the admin CORs, but you'd still be able to set it in the environment variable. Let me know if you think otherwise though 😄

That sounds perfect 👍

PR #6898, opened 2024-04-01
fix: Fixes to product module and improving tests

sradevski:


integration-tests/api/__tests__/admin/product.js
```diff
@@ -2585,102 +2546,6 @@ medusaIntegrationTestRunner({
           expect(variantPost).not.toBeTruthy()
         })
 
-        // TODO: This one is a bit more complex, leaving for later
-        it.skip("successfully deletes a product variant and its associated option values", async () => {
```
sradevski, line 2589: This is testing implementation detail of the module, and as such I moved the tests to the module (there already was a test for the product, I just added one for the variant)

integration-tests/api/__tests__/admin/product.js
```diff
@@ -2008,7 +2006,8 @@ medusaIntegrationTestRunner({
           })
         })
 
-        it("successfully updates a variant's price by changing an existing price (given a region_id)", async () => {
+        // TODO: Do we want to add support for region prices through the product APIs?
```
olivermrbl, line 2009: Yes, I think we should keep this. On the product details page, we allow you to update region and currency prices for variants in the bulk editor. 

That said, I would be fine starting with the currency prices, so this doesn't block the end-to-end flows.

packages/core-flows/src/product/workflows/update-products.ts
```diff
@@ -15,6 +15,7 @@ export const updateProductsWorkflow = createWorkflow(
   (
     input: WorkflowData<WorkflowInput>
   ): WorkflowData<ProductTypes.ProductDTO[]> => {
+    // TODO: Delete price sets for removed variants
```
olivermrbl, line 18: question: Wouldn't the cascade on the link definition take care of this? At least I would expect it to.

olivermrbl (approved):

packages/core-flows/src/product/workflows/update-products.ts
adrien2p: Theoretically, when calling remoteLink.delete, it will call the module service softDelete method, to cascade the deletion, the soft delete will return the deleted entries deeply mapped with the linkable keys and allow to perform the cascade soft delete to all the module services.

The same should apply for restoring. 

That way, if soft deleting a variant is cascading to other entities that have a linkable key, we can also perform the cascade for those entities that was link to something else.

To see more, you can look at the executeCascade on the remote link.

sradevski: I think the only issue is that we currently don't do a soft-delete in the upsertWithReplace, and at this point we don't handle the deleted variants in any way. In the `deleteVariants` workflow this is already handled. Since this doesn't block end-to-end flow for now I decided not to do it for now.

integration-tests/api/__tests__/admin/product.js
sradevski: Sounds good, I'll tackle it once we have a full end-to-end setup in place 👍

packages/types/src/product/common.ts
```diff
@@ -579,6 +579,18 @@ export interface ProductVariantOptionDTO {
...
+   */
+  created_at: string | Date
+  /**
+   * When the product variant option was updated.
+   */
+  updated_at: string | Date
+  /**
+   * When the product variant option was deleted.
+   */
+  deleted_at?: string | Date
```
adrien2p, line 593:
```suggestion
  deleted_at: string | Date | null
```

adrien2p (approved):

PR #6901, opened 2024-04-02
chore(medusa-test-utils):Handle errors gracefully

adrien2p:
**What**
- Better error handling and error message
- update deps management and dynamic import/require
- Pass a new flag to the modules loaders for the module loaders to be able to act depending on it. In that case, the module can determine what should be run or not. e.g in the workflow engine redis, when we are only partially loading the module, we do not want to set the Distributed transaction storage


packages/medusa-test-utils/src/medusa-test-runner.ts
```diff
@@ -147,37 +149,61 @@ export function medusaIntegrationTestRunner({
...
+        force_modules_migration,
+        database_extra: {},
+        dbUrl: dbConfig.clientUrl,
+        dbSchema: dbConfig.schema,
+      })
+
+      dataSourceRes = dbDataSource
+      pgConnectionRes = pgConnection
+    } catch (error) {
+      console.error("Error initializing database", error)
```
srindom: thought: shouldn't we still throw here?

packages/medusa-test-utils/package.json
```diff
@@ -42,6 +43,9 @@
     "@medusajs/medusa": {
       "optional": true
     },
+    "@medusajs/modules-sdk": {
+      "optional": true
+    },
```
srindom: thought: looking at the code this doesn't seem optional. E.g., 

```
const moduleSdkImports = require("@medusajs/modules-sdk")
```

Will fail if there is no @medusajs/modules-sdk installed so isn't it more correct to leave it as a required peer dep?

srindom (approved):
LGTM - just two thoughts

packages/medusa-test-utils/src/medusa-test-runner.ts
adrien2p: Yes definitely, my mistake

packages/medusa-test-utils/package.json
adrien2p: Yes my bad, i wanted to put it as a peerDependencies, i was playing with it to see if one of my issue was coming from the dep or something else and then forgot to move it back

adrien2p:
@srindom just waiting the CI to be green and I believe that we can merge

sradevski (approved):
LGTM

PR #6906, opened 2024-04-02
feat: Correctly define all product module models and SQL migration script

sradevski:
The only thing remaining is modifying `upsertWithReplace` to handle many-to-many constraints correctly and not cascade updates to them. This is something that we should do separately though.


.changeset/brave-beds-join.md
```diff
@@ -0,0 +1,5 @@
+---
+"@medusajs/product": minor
```
olivermrbl: todo: Doesn't have to be a minor - can be a patch

olivermrbl (approved):

packages/product/src/models/product-category.ts
```diff
@@ -88,6 +101,10 @@ class ProductCategory {
   })
   updated_at?: Date
 
+  @Index({ name: "IDX_product_category_deleted_at" })
+  @Property({ columnType: "timestamptz", nullable: true })
+  deleted_at?: Date
```
adrien2p, line 106: todo: `deleted_at: Date | null`

packages/product/src/models/product-collection.ts
```diff
@@ -63,16 +66,8 @@ class ProductCollection {
...
-    this.id = generateEntityId(this.id, "pcol")
-
-    if (!this.handle && this.title) {
-      this.handle = kebabCase(this.title)
-    }
-  }
-
   @BeforeCreate()
-  onCreate() {
+  onInit() {
```
adrien2p, line 70: suggestion: line 66 `deletes_at: Date | null`

packages/product/src/models/product-image.ts
```diff
@@ -6,27 +6,33 @@ import {
...
 import Product from "./product"
 
-type OptionalRelations = "products"
-type OptionalFields = DAL.SoftDeletableEntityDateColumns
+const imageUrlIndexName = "IDX_product_image_url"
+const imageUrlIndexStatement = createPsqlIndexStatementHelper({
+  name: imageUrlIndexName,
+  tableName: "image",
+  columns: ["url"],
+  unique: false,
```
adrien2p, line 25: suggestion: rm

packages/product/src/models/product-image.ts
```diff
@@ -6,27 +6,33 @@ import {
...
+  where: "deleted_at IS NULL",
+})
 
+imageUrlIndexStatement.MikroORMIndex()
```
adrien2p, line 29: comment: In general I prefer having the index on the property itself, unless it is a multi columns, but I don't have a strong opinion but since this is what we are doing everywhere I d like to know why we changed that in this pr :)

packages/product/src/models/product-collection.ts
adrien2p: same for all models

adrien2p (commented):
LGTM few comments a bit after the merge 😂 I suppose the rest will be handle later right?

packages/product/src/models/product-image.ts
sradevski: @adrien2p I was aiming for consistency on where we define the index, regardless if it is multi-column or not. Like that, it's easier to find what indices have been defined for a model (just by looking at the top). That is also in-line with how postgres presents the indices if you do `/d` for a table, for example.

packages/product/src/models/product-category.ts
sradevski: Ah true, I'll take a note on this

PR #6907, opened 2024-04-02
feat(dashboard): Add forms for edit and create product options

kasperkristensen:
**What**
- Adds forms to create and edit product options.


olivermrbl (approved):
I guess we can merge these and reuse for the 2.0 implementation?

kasperkristensen:
@olivermrbl I think how options are edited and created is quite different, but still think its fine to merge, so Stevche can use the forms as shells for the new version.

PR #6909, opened 2024-04-02
fix(dashboard): Combobox should not disable body after being closed

kasperkristensen:
**What**
- Fixes an issue where opening a Combobox inside a modal, and then closing the modal would result in the body preserving the `pointer-events: none;` style.


olivermrbl (approved):

riqwan (approved):
🏅

PR #6911, opened 2024-04-02
feat(medusa): added endpoints for rule attribute/operator/values options

riqwan:
what:

adds endpoints that returns attribute options, operator options and value options for a particular rule.

### DIFF
--- a/www/apps/docs/content/references/fulfillment/classes/fulfillment.AbstractFulfillmentService.mdx
+++ b/www/apps/docs/content/references/fulfillment/classes/fulfillment.AbstractFulfillmentService.mdx
@@ -158,7 +158,7 @@ class MyFulfillmentService extends AbstractFulfillmentService {
   async validateOption(
     data: Record<string, unknown>
   ): Promise<boolean> {
-    return data.id == "my-fulfillment"
+    return data.id === "my-fulfillment"
   }
 }
 ```


riqwan:
@olivermrbl i've removed the optional cases for create promotion apis. I'll keep the change from string to number (`application_method.value`) as is and do it as a part of migrating to bignumber.

packages/medusa/src/api-v2/admin/promotions/query-config.ts
```diff
@@ -51,3 +51,10 @@ export const listTransformQueryConfig = {
   ...retrieveTransformQueryConfig,
   isList: true,
 }
+
+export const listRuleValueTransformQueryConfig = {
+  defaultFields: [],
+  defaultRelations: [],
+  allowedRelations: [],
```
adrien2p: todo: defaults, allowed

packages/medusa/src/api-v2/admin/promotions/rule-attribute-options/route.ts
```diff
@@ -0,0 +1,15 @@
...
+export const GET = async (
+  req: AuthenticatedMedusaRequest,
+  res: MedusaResponse
+) => {
+  const attributes = []
+
+  res.json({
+    attributes,
+  })
+}
```
adrien2p: Q: are we missing something ? :D

packages/medusa/src/api-v2/admin/promotions/rule-operator-options/route.ts
```diff
@@ -0,0 +1,32 @@
...
+import {
+  AuthenticatedMedusaRequest,
+  MedusaResponse,
+} from "../../../../types/routing"
+
+export const GET = async (
+  req: AuthenticatedMedusaRequest,
+  res: MedusaResponse
+) => {
+  const operators = [
```
adrien2p: discussion: any reason why we are not using the RuleOperato instead? looks like the promotion rule operator is the same and could be removed to ensure that we are using all the same operators everywhere

packages/medusa/src/api-v2/admin/promotions/rule-operator-options/route.ts
```diff
@@ -0,0 +1,32 @@
...
+    {
+      id: "in",
```
adrien2p: suggestion: RuleOperator.IN

adrien2p: if you agree, then same for the other places

packages/medusa/src/api-v2/admin/promotions/rule-operator-options/route.ts
adrien2p: suggestion: since this is a const, maybe we can extract it from the function

packages/medusa/src/api-v2/admin/promotions/rule-attribute-options/route.ts
riqwan: lol, to be removed! 😄

packages/medusa/src/api-v2/admin/promotions/rule-operator-options/route.ts
riqwan: you're right. Changing it!

riqwan: @adrien2p addressed all your comments. 💪🏻

packages/medusa/src/api-v2/admin/promotions/rule-value-options/[rule_type]/[rule_attribute_id]/route.ts
```diff
@@ -0,0 +1,95 @@
...
+
+  validateRuleType(ruleType)
+  validateRuleAttribute(ruleType, ruleAttributeId)
+
+  const { rows } = await remoteQuery(
+    remoteQueryObjectFromString({
+      entryPoint: queryConfig.entryPoint,
+      variables: {
+        filters: req.filterableFields,
+        skip: req.listConfig.skip,
```
Comment (adrien2p) [inline packages/medusa/src/api-v2/admin/promotions/rule-value-options/[rule_type]/[rule_attribute_id]/route.ts]: Suggestion : you can spread req.remoteQueryConfig.pagination

packages/medusa/src/api-v2/admin/promotions/utils/validate-rule-attribute.ts
```diff
@@ -0,0 +1,19 @@
...
+  ruleType: string,
+  ruleAttributeId: string
+) {
+  const ruleAttributes = ruleAttributesMap[ruleType] || []
+  const ruleAttribute = ruleAttributes.find((obj) => obj.id === ruleAttributeId)
+
+  if (!ruleAttribute) {
+    throw new MedusaError(
+      MedusaError.Types.INVALID_DATA,
+      `Invalid rule attribute`
```
adrien2p: Suggestion : maybe providing the info back ruleAttributeId?

adrien2p (approved):
LGTM, few suggestions

packages/medusa/src/api-v2/admin/promotions/rule-value-options/[rule_type]/[rule_attribute_id]/route.ts
Comment (riqwan) [reply packages/medusa/src/api-v2/admin/promotions/rule-value-options/[rule_type]/[rule_attribute_id]/route.ts]: nice!

PR #6914, opened 2024-04-02
fix(fulfillment): Wrong type usage

adrien2p:
**What**
Fix the types used in the fulfillment module service interface


shahednasser (approved):
LGTM

PR #6918, opened 2024-04-03
feat: Add some of the missing pieces in the create product form

sradevski:
The changes are still partial, there is more work to be done to have everything function properly.
Also, refactored the file upload from the product media form to a separate component

### DIFF
--- a/packages/admin-next/dashboard/public/locales/en-US/translation.json
+++ b/packages/admin-next/dashboard/public/locales/en-US/translation.json
@@ -55,6 +55,7 @@
   },
   "actions": {
     "save": "Save",
+    "saveAsDraft": "Save as draft",
     "create": "Create",
     "delete": "Delete",
     "remove": "Remove",
@@ -151,7 +152,7 @@
     "attributes": "Attributes",
     "editProduct": "Edit Product",
     "editAttributes": "Edit Attributes",
-    "organization": "Organization",
+    "organization": "Organize",
     "editOrganization": "Edit Organization",
     "options": "Options",
     "editOptions": "Edit Options",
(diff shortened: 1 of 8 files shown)


kasperkristensen:
@sradevski just something to discuss: I have omitted placeholders on purpose for the following reasons:

1. They provide little to no value.
2. The increase the number of translations keys by a lot, which increases the workload of people looking to translate the UI for very little benefit.

So I would prefer to skip them entirely, except for the very few cases where they make sense, eg. search fields, native selects where we need a "dummy" placeholder value, etc. But I am open to including them if we think they provide something useful for the users.

packages/admin-next/dashboard/public/locales/en-US/translation.json
```diff
@@ -151,7 +152,7 @@
     "attributes": "Attributes",
     "editProduct": "Edit Product",
     "editAttributes": "Edit Attributes",
-    "organization": "Organization",
+    "organization": "Organize",
```
kasperkristensen, line 155: nit(non-blocking): update the key to match

kasperkristensen (approved):
LGTM, only thing is the discussion on placeholders, that I think we need to align on 👍

sradevski:
> @sradevski just something to discuss: I have omitted placeholders on purpose for the following reasons:
> 
> 1. They provide little to no value.
> 2. The increase the number of translations keys by a lot, which increases the workload of people looking to translate the UI for very little benefit.
> 
> So I would prefer to skip them entirely, except for the very few cases where they make sense, eg. search fields, native selects where we need a "dummy" placeholder value, etc. But I am open to including them if we think they provide something useful for the users.

@kasperkristensen got it on the placeholders, I do agree that other than the UI looking a bit "more full" they don't provide a lot of value. I simply followed the designs, but I'm happy to remove them if we are happy with it design-wise 👍

kasperkristensen:
> > @sradevski just something to discuss: I have omitted placeholders on purpose for the following reasons:
> > 
> > 1. They provide little to no value.
> > 2. The increase the number of translations keys by a lot, which increases the workload of people looking to translate the UI for very little benefit.
> > 
> > So I would prefer to skip them entirely, except for the very few cases where they make sense, eg. search fields, native selects where we need a "dummy" placeholder value, etc. But I am open to including them if we think they provide something useful for the users.
> 
> @kasperkristensen got it on the placeholders, I do agree that other than the UI looking a bit "more full" they don't provide a lot of value. I simply followed the designs, but I'm happy to remove them if we are happy with it design-wise 👍

Alright if you agree, I think we should remove them, and Ludvig hasn't commented on it when he's reviewed what has been done so far, so don't think he will mind 😅

PR #6920, opened 2024-04-03
feat(dashboard): Edit variant options + create option callback for combobox

kasperkristensen:
**What**
- Adds the ability to edit options on a variant
- Adds the ability to create a new option within a combobox.

**Note**
- The combobox is no longer a proper modal, as the Radix and Ariakit modal implementations being used in the same context were causing issues, as they both apply styles to the body to prevent scrolling and interactions outside the "modal". This means that the if you open a Combobox inside a Radix dialog, you will be able to scroll the dialog. This is not ideal, but something we will have to revisit later.


olivermrbl:
@kasperkristensen, how much of the variant-related changes in this PR can we reuse for 2.0? Do we need to clean up the PR wrt now focusing on 2.0

kasperkristensen:
@olivermrbl I think again this is also quite different, but still think it makes sense to merge and use as a starting point

sradevski (approved):
The combobox has gotten quite complex, so I didn't really do a full review, but let's just get this merged and out of the way

PR #6922, opened 2024-04-03
feat: Normalize known DB errors to a MedusaError when possible

sradevski:
Before we would swallow the error and return a generic error to the user. This will provide more information to the caller if it is one of the known errors.

### DIFF
--- a/packages/utils/src/dal/mikro-orm/db-error-mapper.ts
+++ b/packages/utils/src/dal/mikro-orm/db-error-mapper.ts
@@ -1,0 +1,83 @@
+import {
+  ForeignKeyConstraintViolationException,
+  InvalidFieldNameException,
+  NotFoundError,
+  NotNullConstraintViolationException,
+  UniqueConstraintViolationException,
+} from "@mikro-orm/core"
+import { MedusaError, upperCaseFirst } from "../../common"
+
+export const dbErrorMapper = (err: Error) => {
+  if (err instanceof NotFoundError) {
+    console.log(err)
+    throw new MedusaError(MedusaError.Types.NOT_FOUND, err.message)
+  }
+
+  if (err instanceof UniqueConstraintViolationException) {
+    const info = getConstraintInfo(err)
+    if (!info) {
+      throw err
+    }
+
+    throw new MedusaError(
+      MedusaError.Types.INVALID_DATA,
+      `${upperCaseFirst(info.table)} with ${info.keys
+        .map((key, i) => `${key}: ${info.values[i]}`)
+        .join(", ")} already exists.`
+    )
+  }
+
+  if (err instanceof NotNullConstraintViolationException) {
+    throw new MedusaError(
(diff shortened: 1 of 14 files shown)


packages/utils/src/dal/mikro-orm/integration-tests/__tests__/mikro-orm-repository.spec.ts
```diff
@@ -774,5 +784,73 @@ describe("mikroOrmRepository", () => {
...
+    //   })
+
+    //   expect(listedEntities).toHaveLength(2)
+    //   expect(listedEntities[0].entity3.getItems()).toEqual(
+    //     listedEntities[1].entity3.getItems()
+    //   )
+    // })
+  })
+
+  describe.only("error mapping", () => {
```
adrien2p: todo: should we remove the only?

adrien2p (approved):

packages/utils/src/dal/mikro-orm/integration-tests/__tests__/mikro-orm-repository.spec.ts
sradevski: Definitely 😄

olivermrbl (approved):
great addition @sradevski!

packages/utils/src/dal/mikro-orm/db-error-mapper.ts
```diff
@@ -0,0 +1,82 @@
...
+    return null
+  }
+
+  const [keys, values] = detail.match(/\([^\(.]*\)/g) || []
+
+  if (!keys || !values) {
+    return null
+  }
+
+  return {
```
carlos-r-l-rodrigues, line 72: Can we also return the name of the constraint (err.constraint)? I believe adding it to the message in parentheses would be helpful.

carlos-r-l-rodrigues (approved):

packages/utils/src/dal/mikro-orm/db-error-mapper.ts
sradevski: @carlos-r-l-rodrigues I think that might be too much of an implementation detail that we leak to the FE. What would be nice though is to have it as part of MedusaError in some form and then we plug that into some error logging/tracing before responding to the end user, but I think we can tackle that later when we integrate better with some tracing library

carlos-r-l-rodrigues: Here is the DAL utils, so I expect the backend to have access to these properties.
it is alright to throw from here directly to the front end, but we need to have access to the details on the backend, even to handle errors differently based on the constraint name.

sradevski: @carlos-r-l-rodrigues yeah definitely, whichi s why I think we should add the constraint on MedusaError as a separate field, but not as part of the message. I suggest we just handle that separately though

PR #6927, opened 2024-04-03
feat(dashboard): Update v2 store domain to follow conventions, and abstract some logic

kasperkristensen:
**What**
- Updates the V2 store domain.
- Abstracts some of the API requests logic into re-usable hooks

**Note**
- Partial PR as we need to add support for setting a default currency, region and location. Currently the `q` param is missing on all V2 endpoints, so we can't use our combobox component to list these options. Will be added later when that has been fixed in core.
- The PR includes a generic hook for fetching combobox data, that I added before realising that `q` is not currently supported. But keeping it as it will be useful once support is added.

### DIFF
--- a/.changeset/breezy-months-change.md
+++ b/.changeset/breezy-months-change.md
@@ -1,0 +1,5 @@
+---
+"@medusajs/medusa": patch
+---
+
+fix(medusa): Fix get-query-config backward compatiblity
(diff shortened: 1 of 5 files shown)


packages/admin-next/dashboard/src/v2-routes/store/store-detail/loader.ts
```diff
@@ -1,21 +1,22 @@
...
-  queryFn: async () => medusa.client.request("GET", "/admin/stores"),
+  queryFn: async () => medusa.admin.custom.get("/stores"),
 })
 
 const fetchQuery = async (
   query: FetchQueryOptions<Response<{ stores: any[] }>>
 ) => {
   try {
     const res = await queryClient.fetchQuery(query)
+    console.log("bib", res)
```
olivermrbl: todo: rm

olivermrbl (approved):
Great!

PR #6928, opened 2024-04-03
fix(dashboard): Promotions loader

kasperkristensen:
**What**
- Fixes the loader for promotions which was using a hook inside of a function.
- Add types to the promotions query hook.


riqwan (approved):
🏅

PR #6932, opened 2024-04-04
Feat(customer, medusa, admin-next): Customer and Customer group end to end flows

pKorsholm:
**What**
- add admin-next routes for customers and customer-groups
- fix filtering and add q parameter for querying customers and customer-groups


packages/admin-next/dashboard/src/v2-routes/customer-groups/customer-group-add-customers/components/add-customers-form/add-customers-form.tsx
```diff
@@ -0,0 +1,338 @@
...
+    },
+    getRowId: (row) => row.id,
+    onPaginationChange: setPagination,
+    onRowSelectionChange: setRowSelection,
+    getCoreRowModel: getCoreRowModel(),
+    manualPagination: true,
+    meta: {
+      customerGroupId,
+    },
+  })
```
kasperkristensen, line 112: Would be great to update this to use useDataTable instead

PR #6942, opened 2024-04-04
feat(utils): Support free text configuration

adrien2p:
**What**
- Add support for searchable entity properties to configure default free text search when providing a `q` filter
- `@Searchable` decorator to mark a property or relation to be searchable form the current entity

FIXES CORE-1910

### DIFF
--- a/packages/utils/src/dal/mikro-orm/integration-tests/__tests__/mikro-orm-repository.spec.ts
+++ b/packages/utils/src/dal/mikro-orm/integration-tests/__tests__/mikro-orm-repository.spec.ts
@@ -385,6 +385,38 @@ describe("mikroOrmRepository", () => {
       )
     })
 
+    it("should clear the parent entity from the one-to-many relation", async () => {
+      const entity1 = {
+        id: "1",
+        title: "en1",
+        entity2: [{ title: "en2-1", entity1: null }],
+      }
+
+      await manager1().upsertWithReplace([entity1], {
+        relations: ["entity2"],
+      })
+      const listedEntities = await manager1().find({
+        where: { id: "1" },
+        options: { populate: ["entity2"] },
+      })
+
+      expect(listedEntities).toHaveLength(1)
+      expect(listedEntities[0]).toEqual(
+        expect.objectContaining({
+          id: "1",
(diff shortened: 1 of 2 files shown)


packages/utils/src/dal/mikro-orm/mikro-orm-free-text-search-filter.ts
```diff
@@ -0,0 +1,163 @@
+import "@mikro-orm/core"
```
adrien2p: suggestion: @adrien2p you could cleanup this code a bit to make it more readable please 🙏

riqwan (approved):
clean!

riqwan (approved):
Will add some reservations about API level control for scoping here that we discussed, but we can get to that when we have use-cases for that!

packages/inventory-next/src/models/inventory-item.ts
```diff
@@ -64,6 +63,7 @@ export class InventoryItem {
   deleted_at: Date | null = null
 
   @InventoryItemSkuIndex.MikroORMIndex()
+  @Searchable()
```
sradevski, line 66: Another alternative is to have an entity-level `Searchable` where you pass `[fieldnames]`

packages/inventory-next/src/repositories/inventory-item.ts
```diff
@@ -1,63 +0,0 @@
-import { Context, DAL } from "@medusajs/types"
```
sradevski, line 1: Let's try to remove similar files from other modules now? We will forget to clean up otherwise 😄

packages/inventory-next/src/models/inventory-item.ts
adrien2p: Thats a nice approach too, lets keep it like it for now, and maybe decide on this later, that way the API can start support the search wdyt?

packages/inventory-next/src/repositories/inventory-item.ts
adrien2p: I ve removed it everywhere we can :D there is no other places for now

PR #6944, opened 2024-04-04
feat(product): return parents tree

carlos-r-l-rodrigues:
*What*

include option `include_ancestors_tree` to list products including the parent tree.


packages/product/src/repositories/product-category.ts
```diff
@@ -56,65 +59,118 @@ export class ProductCategoryRepository extends DALUtils.MikroOrmBaseTreeReposito
...
-      )
+    mpaths.push({ mpath: Array.from(parentMpaths) })
 
-      const descendantsByParentId = groupBy(
-        descendantsForCategory,
-        (pc) => pc.parent_category_id
-      )
+    const whereOptions = {
+      ...findOptions.where,
+      $or: mpaths,
```
riqwan, line 116: we'll be overriding existing or conditions here, might be good to merge mpaths with existing.

packages/product/src/repositories/product-category.ts
```diff
@@ -56,65 +59,118 @@ export class ProductCategoryRepository extends DALUtils.MikroOrmBaseTreeReposito
...
-    if (!includeDescendantsTree) {
+    if (!includeDescendantsTree && !includeParentsTree) {
       return productCategories
     }
 
-    return this.buildProductCategoriesWithDescendants(
+    return this.buildProductCategoriesWithTree(
+      {
+        descendants: includeDescendantsTree,
+        parents: includeParentsTree,
```
riqwan, line 69: nit: feel free to ignore - calling this ancestors, to align with descendants.

packages/product/integration-tests/__tests__/services/product-category/index.ts
```diff
@@ -173,6 +170,365 @@ moduleIntegrationTestRunner({
...
+                  parent_category_id: null,
+                  parent_category: null,
+                },
+              },
+              category_children: [
+                {
+                  id: "category-1-b-1",
+                  handle: "category-1-b-1",
+                  mpath: "category-0.category-1.category-1-b.category-1-b-1.",
+                  parent_category_id: "category-1-b",
```
riqwan, line 525: shouldn't there be an empty category_children here?
```suggestion
                  parent_category_id: "category-1-b",
                  category_children: [],
```

riqwan (approved):
nice! 🥇

packages/product/integration-tests/__tests__/services/product-category/index.ts
carlos-r-l-rodrigues: When both flags are `true` we decided to remove the parent and children properties from categories that are not in the root level. This is tricky to support without dealing with circular references.

carlos-r-l-rodrigues:
@olivermrbl @kasperkristensen @adrien2p can we merge this?

olivermrbl (commented):
LGTM

not important but there's a discrepancy in naming ancestors vs. parents

also, can we select fields on ancestors? e.g. if we only want `id` and `name`

carlos-r-l-rodrigues:
It was named parents, beu Riqwan suggested to keep it similar to descendants. I believe everywhere it was named like ancestors, which makes more sense as all of them are returned.

At the moment the fields returned are the same requested to the categories.

PR #6945, opened 2024-04-04
feat: Add create location fulfillment set flow + API

olivermrbl:
Init shipping management

Applied new folder structure


riqwan (approved):

olivermrbl:
@adrien2p, started working on the fulfillment API here

sradevski (approved):

packages/core-flows/src/stock-location/workflows/create-location-fulfillment-set.ts
```diff
@@ -0,0 +1,33 @@
...
+  createLocationFulfillmentSetWorkflowId,
+  (input: WorkflowData<CreateLocationFulfillmentSetWorkflowInputDTO>) => {
+    const fulfillmentSet = createFulfillmentSets([
+      {
+        name: input.fulfillment_set_data.name,
+        type: input.fulfillment_set_data.type,
+      },
+    ])
+
+    const links = transform({ input, fulfillmentSet }, (data) => [
```
adrien2p: suggestion: I would do that in the `associateFulfillmentSetsWithLocationStep` instead of here to have it wrapped in the step

adrien2p: So that someone can reuse the step. the step could accept the same thing as the set region payment providers step

adrien2p (approved):
LGTM, lets wait to merge the other pr to not have more conflicts around the types

packages/core-flows/src/stock-location/workflows/create-location-fulfillment-set.ts
olivermrbl: Hmm, can I get you to explain what you would like to see changed? The input type wouldn't change, as far as I can see. 

And in the create region workflow, we are also using a transformer to normalize the data for the step 🤔

adrien2p: Yes, but we don't create the links shape outside of the step, the step takes care of creating the links so you dont have to worry about how to build link

olivermrbl:
Replaced validation with new zod approach - don't think it needs a re-review, but thought I'd mention it

PR #6951, opened 2024-04-05
fix(workflows-sdk): Handle result from a trasnformer after a step fail

adrien2p:
**What**
Currently, when a step fail and a workflow is marked as throw on error false, the composer will still try to resolve the result object if it is a transformer and fail depending on the transformer implementation. This pr will prevent that from happening and handle the error properly as well as pushing it back into the errors result

### DIFF
--- a/.changeset/plenty-ravens-lay.md
+++ b/.changeset/plenty-ravens-lay.md
@@ -1,0 +1,5 @@
+---
+"@medusajs/workflows-sdk": patch
+---
+
+fix(workflows-sdk): Handle result from a trasnformer after a step fail


carlos-r-l-rodrigues (approved):

packages/workflows-sdk/src/helper/workflow-export.ts
```diff
@@ -211,7 +211,13 @@ function createContextualWorkflowRunner<
 
     let result
     if (options?.wrappedInput) {
-      result = await resolveValue(resultFrom, transaction.getContext())
+      result = resolveValue(resultFrom, transaction.getContext())
+      if (result instanceof Promise) {
+        result = await result.catch((e) => {
```
sradevski, line 216: I believe there is no need to await here (but of course is fine to do so)

sradevski (approved):
Nice

packages/workflows-sdk/src/helper/workflow-export.ts
adrien2p: Yeah I ll keep it in case there is asynchronous function running

PR #6953, opened 2024-04-05
chore: remove team from generate docblock actions

shahednasser:
The "Create Pull Request" step always shown as "failed" because of the `team-reviewers` option, even when it's actually successful. This PR removes the team assignment since it's done automatically anyway


shahednasser:

PR #6956, opened 2024-04-05
feat: Add boilerplate for the file module

sradevski:
There is a lot more work left to be done on the file module, but at least this will allow us to wire things up


packages/core-flows/src/file/steps/delete-files.ts
```diff
@@ -0,0 +1,17 @@
+import { ModuleRegistrationName } from "@medusajs/modules-sdk"
+import { IFileModuleService } from "@medusajs/types"
+import { StepResponse, createStep } from "@medusajs/workflows-sdk"
+
+export const deleteFilesStepId = "delete-files"
+export const deleteFilesStep = createStep(
+  { name: deleteFilesStepId, noCompensation: true },
```
carlos-r-l-rodrigues, line 7: you can remove `noCompensation` and by not defining the compensation function as the second argument, this flag will be true.

carlos-r-l-rodrigues (approved):

PR #6962, opened 2024-04-05
feat: Create shipping options workflow

adrien2p:
**What**

- Create main workflow
- Update create pricing rule types step to be idempotent
- update remote joiner to support granular isList confguration on field aliases
- Add full workflow integration tests


adrien2p:
@olivermrbl what do you think if we get this PR merged and I ll do the end point and tests in a separate pr just to not make this one bigger?

olivermrbl:
@adrien2p , I think that makes sense

integration-tests/modules/__tests__/fulfillment/workflows/create-shipping-options.ts
```diff
@@ -0,0 +1,270 @@
...
+        expect(prices).toContainEqual(
+          expect.objectContaining({
+            currency_code: "eur",
+            amount: 100,
+            rules_count: 1,
+          })
+        )
+      })
+
+      it.only("should revert the shipping options and prices", async () => {
```
olivermrbl:
```suggestion
      it("should revert the shipping options and prices", async () => {
```

olivermrbl (approved):
Very nice, LGTM!

PR #6965, opened 2024-04-05
fix: Add count to stock-locations response

olivermrbl:


packages/medusa/src/api-v2/admin/stock-locations/route.ts
```diff
@@ -49,6 +49,8 @@ export const GET = async (req: MedusaRequest, res: MedusaResponse) => {
 
   res.status(200).json({
     stock_locations,
-    ...req.remoteQueryConfig.pagination,
+    count: metadata.count,
```
adrien2p, line 50: todo: variables could spread the remoteQueryConfig.pagination right?

adrien2p (approved):

PR #6967, opened 2024-04-05
chore(orchestration): serialize error

carlos-r-l-rodrigues:
What:

Serialize Error instances to object.
This will have the correct behavior when JSON.stringify(error) is used to store it.


.changeset/clever-countries-guess.md
```diff
@@ -0,0 +1,5 @@
+---
+"@medusajs/orchestration": patch
```
adrien2p, line 2: todo: we should add the workflow sdk

adrien2p (approved):

.changeset/clever-countries-guess.md
carlos-r-l-rodrigues: just the test has changed there.

PR #6973, opened 2024-04-06
Chores: Do not disable foreign keys by default

adrien2p:
**What**
By default, the migrations are run with the option as true (default value), which can create issue when working in a shared env and not having the root user rights


sradevski (approved):

srindom:
This works! 👍

PR #6975, opened 2024-04-06
docs: Add reminder to swap client-id in PayPalScriptProvider

### LINKED ISSUE #6974: Include extra reminder to swap out "client-id" in the PayPalScriptProcessor when going live with PayPay payments
### Preliminary Checks

- [X] This issue is not a duplicate. Before opening a new issue, please search existing issues: https://github.com/medusajs/medusa/issues

### Issue Summary

Include extra reminder to swap out "client-id" in the PayPalScriptProcessor when going live with PayPay payments when using `nextjs-starter-medusa`. The Sandbox mode works with the default "test" string, but if it isn't change when going live, attempting to pay with PayPal makes the pop-up window fail.

_No response_

- [X] I would like to fix this issue

ajz003:
This PR Includes extra reminder in the `medusa-payment-paypal` docs to swap out "client-id" in the PayPalScriptProcessor when going live with PayPay payments when using `nextjs-starter-medusa`. The Sandbox mode works with the [default "test" string](https://github.com/medusajs/nextjs-starter-medusa/blob/2be5564f383e154e8595eefbd5ec0e62b56c3e7a/src/modules/checkout/components/payment-wrapper/index.tsx#L49), but if it isn't change when going live, attempting to pay with PayPal makes the pop-up window fail. It's not immediately obvious that this is the issue when this happens.

This PR also renames the component `PayPalScriptProcessor` to `PayPalScriptProvider` to match the [component](https://github.com/paypal/react-paypal-js/blob/main/src/components/PayPalScriptProvider.tsx) from `@paypal/react-paypal-js`

Resolves #6974


shahednasser:
Thank you for opening this PR @ajz003 ! Seems like this is a bug in the Next.js storefront rather than the docs, as it should be an environment variable rather than `test`. I've spoken to our team member responsible for the Next.js storefront and he'll take care of fixing it there, so I'll close this PR and its associated issue. Thanks again!

PR #6976, opened 2024-04-06
feat: Bring back API key sales channel management

olivermrbl:
- Add API key sales channel management
- Add HTTP responses for API keys and sales channels
- Use HTTP responses in `dashboard` and remove now redundant types


packages/types/src/http/sales-channel/admin/sales-channel.ts
```diff
@@ -0,0 +1,26 @@
...
+  name: string
+  description: string | null
+  is_disabled: boolean
+  metadata: Record<string, unknown> | null
+}
+
+/**
+ * @experimental
+ */
+export interface AdminSalesChannelListResponse extends PaginatedResponse {
```
olivermrbl, line 17: @adrien2p, think we should structure the responses this way instead. Minor difference from what we have right now

adrien2p: yeah that sounds good 👍

adrien2p (approved):

sradevski (approved):

PR #6978, opened 2024-04-06
feat: Switch to zod for the product API and implement missing primitives

sradevski:


integration-tests/api/__tests__/admin/product.js
```diff
@@ -2764,10 +2769,10 @@ medusaIntegrationTestRunner({
 
           const payload = {
             title: "Second variant",
-            sku: variant.sku,
```
sradevski, line 2767: These were all null before which is not a valid value

packages/medusa/src/api-v2/admin/products/helpers.ts
```diff
@@ -39,14 +41,27 @@ export const remapKeysForVariant = (selectFields: string[]) => {
...
+  TReq extends AdminCreateProductType | AdminUpdateProductType,
+  TResp extends CreateProductDTO | UpdateProductDTO
+>(
+  p: TReq
+): TResp => {
   return {
     ...p,
-    variants: p.variants?.map(remapVariant),
+    // TODO: Map to enum
+    status: p.status as ProductStatus,
```
sradevski: I will tackle this now

packages/medusa/src/api-v2/utils/validate-query.ts
```diff
@@ -0,0 +1,67 @@
...
+/**
+ * Omit the non filterable config from the validated object
+ * @param obj
+ */
+const getFilterableFields = <T extends RequestQueryFields>(obj: T): T => {
+  const result = omit(obj, ["limit", "offset", "fields", "order"]) as T
+  return removeUndefinedProperties(result)
+}
+
+export function validateAndTransformQuery<TEntity extends BaseEntity>(
```
adrien2p, line 37: q: shouldn't this be in the. middlewares?

packages/medusa/src/api-v2/admin/products/query-config.ts
```diff
@@ -28,8 +28,6 @@ export const defaultAdminProductsVariantFields = [
 
 export const retrieveVariantConfig = {
   defaultFields: defaultAdminProductsVariantFields,
```
adrien2p: todo: defaults

packages/medusa/src/api-v2/utils/validators.ts
```diff
@@ -0,0 +1,70 @@
+import { z } from "zod"
+
+export const createSelectParams = () => {
```
adrien2p, line 3: waiting for those for my pr :D

adrien2p (approved):

packages/medusa/src/api-v2/admin/products/helpers.ts
adrien2p: we can use nativeEnum for that

packages/medusa/src/api-v2/admin/products/helpers.ts
```diff
@@ -39,14 +41,27 @@ export const remapKeysForVariant = (selectFields: string[]) => {
...
   }
 }
 
-export const remapVariant = (v: ProductVariantDTO) => {
+export const remapProductResponse = (p: ProductDTO) => {
+  return {
+    ...p,
```
adrien2p: todo: if we could remove all one letter please

PR #6980, opened 2024-04-06
feat: Totals calculation

carlos-r-l-rodrigues:


carlos-r-l-rodrigues:
@srindom If possible, I would like a double check on [this test](https://github.com/medusajs/medusa/blob/938e9b5e044467e9bad9d24b6012d2040e8c6ea0/packages/utils/src/totals/__tests__/totals.ts) to make sure the calculations are correct.

carlos-r-l-rodrigues:
@olivermrbl I will have another PR to add the calculation on Cart. cc: @adrien2p

olivermrbl:
@carlos-r-l-rodrigues, sounds good

packages/order/integration-tests/__tests__/create-order.ts
```diff
@@ -194,7 +194,8 @@ moduleIntegrationTestRunner({
       it("should create an order, shipping method and items. Including taxes and adjustments associated with them", async function () {
         const createdOrder = await service.create(input)
 
-        expect(createdOrder).toEqual(expectation)
+        const serializedOrder = JSON.parse(JSON.stringify(createdOrder))
```
adrien2p, line 197: Q: shouldnt it be already serialized when the service return the order?

packages/payment/src/services/payment-module.ts
```diff
@@ -567,7 +567,10 @@ export default class PaymentModuleService<
     )
 
     // When the entire authorized amount has been captured, we mark it fully capture by setting the captured_at field
-    if (capturedAmount.plus(newCaptureAmount).eq(authorizedAmount)) {
+    const totalCaptured = MathBN.convert(
+      MathBN.add(capturedAmount, newCaptureAmount)
+    )
+    if (MathBN.eq(totalCaptured, authorizedAmount)) {
```
adrien2p: Q: just to be sure, even if that is not suppose the happen, would gte be ok to be used here?

packages/types/src/order/mutations.ts
```diff
@@ -286,9 +288,9 @@ export interface ConfirmOrderChangeDTO {
 /** ORDER CHANGE ACTION START */
 
 export interface CreateOrderChangeActionDTO {
-  order_change_id: string
-  reference: string
-  reference_id: string
+  order_change_id?: string
+  reference?: string
+  reference_id?: string
```
adrien2p, line 293: Q: we can potentially create change actions that are not attached?

packages/utils/src/totals/cart/index.ts
```diff
@@ -0,0 +1,222 @@
...
+  region?: {
+    automatic_taxes?: boolean
+  }
+}
+
+export function decorateCartTotals(
+  cartLike: DecorateCartLikeInputDTO,
+  config: TotalsConfig = {}
+): CartLikeWithTotals {
+  transformPropertiesToBigNumber(cartLike)
```
adrien2p, line 44: as discussed, we might have lost precision when serializing, I am mentionning it here just so that we remember about it until we find the right approach :D

adrien2p (approved):

packages/payment/src/services/payment-module.ts
carlos-r-l-rodrigues: fair enough... will add it.

packages/types/src/order/mutations.ts
carlos-r-l-rodrigues: yes, they can be generic and open to any modification.

packages/order/integration-tests/__tests__/create-order.ts
carlos-r-l-rodrigues: this "serialization" is just to transform the BigNumbers to numbers. the serialize of the Mikro-ORM entity already happened inside the module

packages/utils/src/totals/cart/index.ts
carlos-r-l-rodrigues: will add tests to that in another PR where more modules will be using BigNumber

packages/types/src/cart/common.ts
```diff
@@ -654,67 +675,72 @@ export interface CartDTO {
...
   /**
    * The subtotal of the cart.
    */
-  subtotal: number
+  subtotal: BigNumberValue
+
+  /**
+   * The subtotal of the cart excluding taxes.
+   */
+  subtotal_without_taxes: BigNumberValue
```
olivermrbl: @carlos-r-l-rodrigues, I would still say `subtotal` should always be without taxes, so I wouldn't add a dedicated property for this – what made you reconsider and add this?

carlos-r-l-rodrigues: I talked to Seb and I will revert it. Subtotal will be without taxes even if taxes are included

PR #6984, opened 2024-04-06
feat: Delete service zone

olivermrbl:


carlos-r-l-rodrigues (approved):

adrien2p (commented):
@olivermrbl Same as for the rules we discussed, there is no place where we do check the service zones does belong to the fs no?

PR #6986, opened 2024-04-07
feat: Add support for setting sales channel when creating a product

sradevski:


packages/core-flows/src/common/steps/remove-remote-links.ts
```diff
@@ -35,7 +35,7 @@ export const removeRemoteLinkStep = createStep(
     )
     await link.delete(grouped)
 
-    return new StepResponse(void 0, grouped)
```
sradevski, line 38: There is a typing issue where if you return `void 0` you cannot do `.config` on the step to change the name

packages/core-flows/src/product/workflows/create-products.ts
```diff
@@ -83,23 +102,6 @@ export const createProductsWorkflow = createWorkflow(
...
-          variants: product.variants?.map((variant) => ({
-            ...variant,
-            price_set: data.variantAndPriceSets.find(
-              (v) => v.variant.id === variant.id
-            )?.price_set,
-          })),
-        }))
-      }
-    )
+    return createdProducts
```
sradevski, line 105: Since we refetch after a workflow, I think it's better to not bother and reconstruct the data for now

packages/core-flows/src/common/steps/remove-remote-links.ts
```diff
@@ -35,7 +35,7 @@ export const removeRemoteLinkStep = createStep(
...
+    return new StepResponse(grouped, grouped)
```
carlos-r-l-rodrigues, line 38: nit:
```suggestion
    return new StepResponse(grouped)
```

carlos-r-l-rodrigues (approved):

packages/medusa/src/api-v2/admin/products/validators.ts
```diff
@@ -202,6 +202,7 @@ export const AdminCreateProduct = z
     tags: z.array(AdminUpdateProductTag).optional(),
     options: z.array(AdminCreateProductOption).optional(),
     variants: z.array(AdminCreateProductVariant).optional(),
+    sales_channels: z.array(z.object({ id: z.string() })).optional(),
```
olivermrbl, line 205: question: Should we allow string IDs too? I know this would be diverging from 1.0, but I don't see the benefit of constructing objects in the client.

olivermrbl (approved):
Other than Carlos' comment, LGTM

packages/medusa/src/api-v2/admin/products/validators.ts
adrien2p: I do agree, this is due to the legacy usage of providing ids, the naming was always referring to the target object. In other places we do allow the ids too instead of objects

sradevski: @olivermrbl I tried not to introduce too many breaking changes in the API, but we can revisit once we have things working

PR #6989, opened 2024-04-07
fix(modules-sdk): Fix remote query selection

adrien2p:
Fix remote query selection for different relations levels as part of a single processed query

### DIFF
--- a/.changeset/tender-timers-compete.md
+++ b/.changeset/tender-timers-compete.md
@@ -1,0 +1,5 @@
+---
+"@medusajs/modules-sdk": patch
+---
+
+fix(modules-sdk): Fix remote query selection


packages/modules-sdk/src/remote-query.ts
```diff
@@ -93,9 +93,11 @@ export class RemoteQuery {
     let fields: Set<string> = new Set()
     let relations: string[] = []
 
+    let hasStar = false
+
     for (const field of expand.fields ?? []) {
       if (field === "*") {
-        expand.fields = undefined
+        hasStar = true
```
sradevski: Let's name this something else, like `shouldSelectAll`

adrien2p: nice one,m changed 👍

sradevski (approved):

carlos-r-l-rodrigues (approved):

PR #6998, opened 2024-04-07
feat: Admin V2 Customers

olivermrbl:
Extracted almost all of it from #6932 and then cleaned it up and added the new hooks

### DIFF
--- a/packages/medusa/src/api-v2/admin/customers/validators.ts
+++ b/packages/medusa/src/api-v2/admin/customers/validators.ts
@@ -9,8 +9,8 @@ export const AdminCustomerParams = createSelectParams()
 export const AdminCustomerGroupParams = createSelectParams()
 
 export const AdminCustomerGroupInCustomerParams = z.object({
-  id: z.union([z.string(), z.array(z.string())]),
-  name: z.union([z.string(), z.array(z.string())]),
+  id: z.union([z.string(), z.array(z.string())]).optional(),
+  name: z.union([z.string(), z.array(z.string())]).optional(),
   created_at: createOperatorMap().optional(),
   updated_at: createOperatorMap().optional(),
   deleted_at: createOperatorMap().optional(),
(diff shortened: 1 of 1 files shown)


packages/admin-next/dashboard/src/v2-routes/customers/customer-detail/components/customer-group-section/customer-group-section.tsx
```diff
@@ -0,0 +1,54 @@
+// TODO: Should be added with Customer Groups UI
```
olivermrbl, line 1: Should be added back when adding in Customer Groups UI

packages/admin-next/dashboard/src/v2-routes/customers/customer-create/components/create-customer-form/create-customer-form.tsx
```diff
@@ -1,61 +1,44 @@
...
-        path: ["password_confirmation"],
-      })
-    }
-  })
+const CreateCustomerSchema = zod.object({
+  email: zod.string().email(),
+  first_name: zod.string().min(1),
+  last_name: zod.string().min(1),
+  phone: zod.string().min(1).optional(),
+})
```
sradevski, line 18: How would passwords be handled now? Send a reset password link?

packages/types/src/http/customer/admin/customer.ts
```diff
@@ -0,0 +1,83 @@
...
+
+/**
+ * @experimental
+ */
+export interface AdminCustomerResponse {
+  customer: CustomerResponse
+}
+
+/**
+ * @experimental
```
sradevski, line 65: What's the purpose of `experimental`?

sradevski (approved):
Looking good!

packages/types/src/http/customer/admin/customer.ts
olivermrbl: 1.0 and 2.0 types are exported from the same package, so the @experimental flag was added so external developers don't confuse the types.

packages/admin-next/dashboard/src/v2-routes/customers/customer-create/components/create-customer-form/create-customer-form.tsx
olivermrbl: Yes, we should not be able to set the password on customer accounts. That is a separate flow after creating the customer.

PR #7006, opened 2024-04-08
docs: fix typedoc errors and warnings

shahednasser:
Fix typedoc errors and warning shown during reference generation


packages/types/src/customer/mutations.ts
```diff
@@ -283,26 +283,6 @@ export interface CustomerUpdatableFields {
...
-   * Holds custom data in key-value pairs.
-   */
-  metadata?: Record<string, unknown> | null
-
-  /**
-   * Who the customer group.
-   */
-  created_by?: string
-}
-
```
shahednasser, line 305: clarification: this interface is defined twice in the same file. So, I removed one of the definitions

olivermrbl (approved):

PR #7009, opened 2024-04-08
feat: Categories retrieve + list API

olivermrbl:


packages/product/src/services/product-category.ts
```diff
@@ -76,6 +77,17 @@ export default class ProductCategoryService<
     delete filters.include_descendants_tree
     delete filters.include_ancestors_tree
 
+    // Apply free text search filter
+    if (filters?.q) {
```
olivermrbl, line 81: These are temporary until we figure out what we want to do with the ProductCategoryService. Right now, it is a custom service not using the abstract service factory, so the free text search is not supported out of the box. 

cc @adrien2p

sradevski (approved):
Nice 👍 

I haven't put much focus on categories as I believe might need some thinking through, but this is nice that we get at least the basic things working

PR #7017, opened 2024-04-08
fix(medusa, core-flows): Delete service zone

olivermrbl:


packages/core-flows/src/fulfillment/steps/delete-service-zones.ts
```diff
@@ -10,7 +10,7 @@ export const deleteServiceZonesStep = createStep(
       ModuleRegistrationName.FULFILLMENT
     )
 
-    await service.softDelete(ids)
+    await service.softDeleteServiceZones(ids)
```
sradevski, line 13: Ive seen this in few places :)

sradevski (approved):

packages/core-flows/src/fulfillment/steps/delete-service-zones.ts
olivermrbl: False positive test because of idempotent operation. We need to be more deliberate about testing for existence separately from the deletion 😄

fPolic (approved):

PR #7018, opened 2024-04-08
Feat(medusa, core-flows, types): add reservation endpoints

pKorsholm:
**What**
- Add CRUD for reservations


packages/medusa/src/api-v2/admin/reservations/validators.ts
```diff
@@ -0,0 +1,129 @@
+import { FindParams, extendedFindParamsMixin } from "../../../types/common"
```
olivermrbl, line 1: suggestion(non-blocking): Use `zod`

olivermrbl (approved):

olivermrbl:
Once the conflict is resolved, we can merge

PR #7022, opened 2024-04-09
Feat(modules-sdk): define link

carlos-r-l-rodrigues:
```typescript
defineLink(
  `${Modules.CART}.cart_id`,
  `${Modules.PAYMENT}.payment_collection_id`
)

defineLink(
  `${Modules.CART}.cart_id`,
  `${Modules.PAYMENT}.payment_collection_id[]` // isList: true
)

defineLink(
  {
     module: Modules.CART,
     key: "cart_id",
     alias: "cart_alias"
  },
  Modules.PAYMENT + ".payment_collection_id"
)

defineLink(
  `${Modules.CART}.cart_id`,
  `${Modules.PAYMENT}.payment_collection_id`,
  {
    database: {
      table: "cart_payment_collection",
      idPrefix: "capaycol",
    },
  }
)
```


olivermrbl:
Would it make sense to simplify the API of `defineLink` to only support one way? My idea is to try to reduce the amount of decisions developers need to make in the context of link definitions. We make it easy for them and they don't have to worry about what is most appropriate.

What I have in mind was something like this:
```ts
// /src/links/index.ts

const someLink = {
  modules: [
    {
      name: Modules.CART,
      key: "cart_id",
    },
    {
      name: Modules.PAYMENT,
      key: "payment_collection_id",
      isList: true,
    },
  ],
  options: { idPrefix: "capaycol" }
}

const someOtherLink = {
  modules: [
    {
      name: Modules.CART,
      key: "cart_id",
    },
    {
      name: Modules.SALES_CHANNEL,
      key: "sales_channel_id",
    },
  ],
}

defineLink([ someLink, someOtherLink ])
```

Curious to hear what you think about this

cc @srindom

carlos-r-l-rodrigues:
I like this approach too because at the end we have multiple extra properties that we'll need. 
So having a proper object with all the available options, like the `fieldAlias`, is good.

olivermrbl:
@carlos-r-l-rodrigues, should we reopen this and get it merged?

I think regardless of where we will load links created by external devs, we will need the utility to perform it.

carlos-r-l-rodrigues:
I haven't applied this last format proposed. I will fix conflicts and open another PR..

PR #7034, opened 2024-04-09
feat: Create product category flow

olivermrbl:
What
- Add create product category flow
- Convert API integration test suite to V2 test runner

### DIFF
--- a/.changeset/proud-crabs-marry.md
+++ b/.changeset/proud-crabs-marry.md
@@ -1,0 +1,5 @@
+---
+"medusa-core-utils": patch
+---
+
+fix(medusa-core-utils): add missing awilix dependency
--- a/integration-tests/api/__tests__/admin/product-category.ts
+++ b/integration-tests/api/__tests__/admin/product-category.ts
@@ -210,148 +210,153 @@ medusaIntegrationTestRunner({
         expect(response.data.offset).toEqual(0)
         expect(response.data.limit).toEqual(7)
 
-        expect(response.data.product_categories).toEqual(expect.arrayContaining([
-          expect.objectContaining({
-            id: productCategoryChild.id,
-            ...breaking(
-              () => ({
-                parent_category: expect.objectContaining({
+        expect(response.data.product_categories).toEqual(
+          expect.arrayContaining([
+            expect.objectContaining({
+              id: productCategoryChild.id,
+              ...breaking(
+                () => ({
+                  parent_category: expect.objectContaining({
+                    id: productCategory.id,
+                    handle: productCategory.handle,
+                    rank: 0,
+                  }),
+                }),
+                () => ({
+                  parent_category_id: productCategory.id,
+                })
+              ),
+              category_children: [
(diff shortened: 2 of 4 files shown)


packages/core-flows/src/product-category/steps/create-product-category.ts
```diff
@@ -0,0 +1,35 @@
+import { ModuleRegistrationName } from "@medusajs/modules-sdk"
+import {
+  CreateProductCategoryDTO,
+  IProductModuleService,
+} from "@medusajs/types"
+import { StepResponse, createStep } from "@medusajs/workflows-sdk"
+
+type CreateProductCategoryStepInput = {
+  product_category: CreateProductCategoryDTO
```
sradevski, line 9: Typically we accept an array for the input, should we do the same here? Or is there a reason we can't?

packages/medusa/src/api-v2/admin/product-categories/query-config.ts
```diff
@@ -11,10 +11,39 @@ export const defaults = [
...
+  "is_internal",
+  "rank",
+  "parent_category_id",
+  "created_at",
+  "updated_at",
+  "metadata",
+
   "parent_category.id",
   "parent_category.name",
+  "parent_category.handle",
```
sradevski: Should we just do `*parent_category`, or is there an issue with circular dependencies? Same for the children

packages/medusa/src/api-v2/admin/product-categories/route.ts
```diff
@@ -33,3 +40,33 @@ export const GET = async (
...
   })
 }
+
+export const POST = async (
+  req: AuthenticatedMedusaRequest<AdminCreateProductCategoryType>,
+  res: MedusaResponse<AdminProductCategoryResponse>
+) => {
+  const { result, errors } = await createProductCategoryWorkflow(req.scope).run(
+    {
+      input: { product_category: { ...req.validatedBody } },
```
sradevski:
```suggestion
      input: { product_category: req.validatedBody  },
```

sradevski (approved):
LGTM, but I didn't review the tests since the diff is impossible :)

packages/core-flows/src/product-category/steps/create-product-category.ts
olivermrbl: The product category service + repository will need a makeover to allow for bulk operations. This is a pretty comprehensive change, so I decided to open this PR now and then we can address the bulk support in a follow-up PR.

packages/medusa/src/api-v2/admin/product-categories/query-config.ts
olivermrbl: Think we can do that. Will update it 👍

olivermrbl:
> but I didn't review the tests since the diff is impossible

Yeah, I am not sure we can do it any differently when converting the test suites. As long as the V1 tests are still passing without any changes, I think we should be fine.

PR #7037, opened 2024-04-09
feat: Add support for price setting and updates for products

sradevski:
We still need to add a `batch` method for variants, but I'll do that in a separate PR


packages/admin-next/dashboard/src/v2-routes/products/product-create/schema.ts
```diff
@@ -0,0 +1,82 @@
...
+export const normalizeVariants = (
+  variants: (Partial<CreateProductVariantDTO> & {
+    prices?: Record<string, string>
+  })[]
+) => {
+  return variants.map((variant) => ({
+    ...variant,
+    prices: Object.entries(variant.prices || {}).map(([key, value]: any) => ({
+      currency_code: key,
+      amount: value ? parseFloat(value) : 0,
```
fPolic, line 77: this needs to be converted into the databse format before sending the request (see `getDbAmunt` example)

packages/admin-next/dashboard/src/v2-routes/products/product-prices/pricing-edit.tsx
```diff
@@ -0,0 +1,85 @@
...
+>
+
+export const PricingEdit = ({ product }: { product: ExtendedProductDTO }) => {
+  const { t } = useTranslation()
+  const { handleSuccess } = useRouteModal()
+  const form = useForm<UpdateVariantPricesSchemaType>({
+    defaultValues: {
+      variants: product.variants.map((variant: any) => ({
+        prices: variant.prices.reduce((acc: any, price: any) => {
+          acc[price.currency_code] = price.amount
```
fPolic, line 31: the format also needs to be converted (see `getPresentationalAmount`)

packages/admin-next/dashboard/src/v2-routes/products/product-create/components/create-product.tsx
```diff
@@ -0,0 +1,175 @@
...
+export const CreateProductPage = () => {
+  const { t } = useTranslation()
+  const [tab, setTab] = useState<Tab>(Tab.PRODUCT)
+  const { handleSuccess } = useRouteModal()
+  const form = useForm<CreateProductSchemaType>({
+    defaultValues: defaults,
+    resolver: zodResolver(CreateProductSchema),
+  })
+
+  const { mutateAsync, isLoading } = useCreateProduct()
```
kasperkristensen, line 35:
```suggestion
  const { mutateAsync, isPending } = useCreateProduct()
```

packages/admin-next/dashboard/src/v2-routes/products/product-prices/pricing-edit.tsx
kasperkristensen: ^^ We need to do these conversions. 

I guess it sort of works in theory when you don't do the conversions at either end, but only in the context of the admin dashboard, to all other third party consumers the value will be wrong. If the user enters 100 for a USD cell, and we simply sent that off to the backend, then it will be read as 1 USD by a third party such as Stripe, so we need to send off the DB "safe" value otherwise people will be selling the products at 1/10 of the price

kasperkristensen (commented):
Over all it looks good, just need to do the presentational/dbAmount conversions, and update the usage of mutation hooks to use `isPending` as `isLoading` is not returned from those anymore.

On another note, I am not able to create variants, I get the following error each time both in the create product and create variant flows:

```
error:   Cannot set field 'id' of Product variant option to null
{
  message: "Cannot set field 'id' of Product variant option to null",
  name: 'Error',
  stack: "Error: Cannot set field 'id' of Product variant option to null\n" +
    '    at dbErrorMapper (/Users/user/work/v2-server/node_modules/@medusajs/utils/dist/dal/mikro-orm/db-error-mapper.js:40:15)\n' +
    '    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)',
  type: 'invalid_data',
  code: undefined,
  date: 2024-04-10T09:49:43.853Z
}
```

kasperkristensen (approved):
Merging this as is and then we will resolve the price transforms later

PR #7038, opened 2024-04-09
feat: Add support for batch method for products and product variants

sradevski:
There is still work to be done on the typings, but I'll do that in one go after we align the typings for the HTTP layer

### DIFF
--- a/packages/core-flows/src/product/steps/update-product-variants.ts
+++ b/packages/core-flows/src/product/steps/update-product-variants.ts
@@ -27,7 +27,7 @@ export const updateProductVariantsStep = createStep(
       if (data.product_variants.some((p) => !p.id)) {
         throw new MedusaError(
           MedusaError.Types.INVALID_DATA,
-          "Product variant id is required when doing a batch update"
+          "Product variant ID is required when doing a batch update of product variants"
         )
       }
 
--- a/packages/core-flows/src/product/steps/update-products.ts
+++ b/packages/core-flows/src/product/steps/update-products.ts
@@ -27,7 +27,7 @@ export const updateProductsStep = createStep(
       if (data.products.some((p) => !p.id)) {
         throw new MedusaError(
           MedusaError.Types.INVALID_DATA,
-          "Product id is required when doing a batch update"
+          "Product ID is required when doing a batch update of products"
         )
       }
 
--- a/packages/medusa/src/api-v2/admin/products/[id]/variants/op/batch/route.ts
+++ b/packages/medusa/src/api-v2/admin/products/[id]/variants/op/batch/route.ts
@@ -8,7 +8,7 @@ import {
   AdminCreateProductType,
 } from "../../../../validators"
 import { BatchMethodRequest } from "@medusajs/types"
-import { refetchProduct, remapProductResponse } from "../../../../helpers"
+import { refetchBatchVariants, remapVariantResponse } from "../../../../helpers"
 
 export const POST = async (
   req: AuthenticatedMedusaRequest<
@@ -43,10 +43,15 @@ export const POST = async (
     throw errors[0].error
   }
 
-  const product = await refetchProduct(
-    productId,
+  const batchResults = await refetchBatchVariants(
+    result,
     req.scope,
     req.remoteQueryConfig.fields
   )
-  res.status(200).json({ product: remapProductResponse(product) })
+
+  res.status(200).json({
+    created: batchResults.created.map(remapVariantResponse),
+    updated: batchResults.updated.map(remapVariantResponse),
+    deleted: batchResults.deleted,
+  })
 }
--- a/packages/medusa/src/api-v2/admin/products/middlewares.ts
+++ b/packages/medusa/src/api-v2/admin/products/middlewares.ts
@@ -3,7 +3,7 @@ import { authenticate } from "../../../utils/authenticate-middleware"
 import { maybeApplyLinkFilter } from "../../utils/maybe-apply-link-filter"
 import { validateAndTransformBody } from "../../utils/validate-body"
 import { validateAndTransformQuery } from "../../utils/validate-query"
-import { createBatchParams } from "../../utils/validators"
+import { createBatchBody } from "../../utils/validators"
 import * as QueryConfig from "./query-config"
 import { maybeApplyPriceListsFilter } from "./utils"
 import {
@@ -71,7 +71,7 @@ export const adminProductRoutesMiddlewares: MiddlewareRoute[] = [
     matcher: "/admin/products/op/batch",
     middlewares: [
       validateAndTransformBody(
-        createBatchParams(AdminCreateProduct, AdminBatchUpdateProduct)
+        createBatchBody(AdminCreateProduct, AdminBatchUpdateProduct)
       ),
       validateAndTransformQuery(
         AdminGetProductParams,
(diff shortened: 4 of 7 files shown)


packages/medusa/src/api-v2/admin/products/middlewares.ts
```diff
@@ -64,6 +66,19 @@ export const adminProductRoutesMiddlewares: MiddlewareRoute[] = [
       ),
     ],
   },
+  {
+    method: ["POST"],
+    matcher: "/admin/products/op/batch",
```
sradevski, line 71: Currently there is an issue with the routing, where both `/:id` and `/batch` middlewares get executed, but the route itself isn't. I think how middlewares and routes are loaded will need a revamp (my guess is that we first register all middlewares, and then all routes, but we should it route by route instead)

For now just added an `op/` prefix so we can start using it, but I created a ticket to revisit this

packages/medusa/src/api-v2/utils/validators.ts
```diff
@@ -1,5 +1,16 @@
 import { z } from "zod"
 
+export const createBatchParams = (
```
olivermrbl: suggestion: 
```suggestion
export const createBatchBody = (
```

packages/medusa/src/api-v2/admin/products/[id]/variants/op/batch/route.ts
```diff
@@ -0,0 +1,52 @@
...
+  if (Array.isArray(errors) && errors[0]) {
+    throw errors[0].error
+  }
+
+  const product = await refetchProduct(
+    productId,
+    req.scope,
+    req.remoteQueryConfig.fields
+  )
+  res.status(200).json({ product: remapProductResponse(product) })
```
Comment (olivermrbl) [inline packages/medusa/src/api-v2/admin/products/[id]/variants/op/batch/route.ts]: thought: Would it make sense to establish a new convention specifically for batch operations, that is, to always respond with the result of each operation?

packages/core-flows/src/product/workflows/batch-products.ts
```diff
@@ -0,0 +1,34 @@
+import { WorkflowData, createWorkflow } from "@medusajs/workflows-sdk"
+import { batchProductsStep } from "../steps/batch-products"
+import { PricingTypes, ProductTypes } from "@medusajs/types"
+
+type WorkflowInput = {
```
olivermrbl, line 5: todo: Move to `@medusajs/types`, but I guess this will come in a follow up PR

packages/core-flows/src/product/workflows/batch-product-variants.ts
```diff
@@ -0,0 +1,33 @@
+import { WorkflowData, createWorkflow } from "@medusajs/workflows-sdk"
+import { batchProductVariantsStep } from "../steps/batch-product-variants"
+import { PricingTypes, ProductTypes } from "@medusajs/types"
+
+type BatchProductVariantsInput = {
```
olivermrbl, line 5: Same here

packages/core-flows/src/product/steps/update-products.ts
```diff
@@ -16,6 +23,22 @@ export const updateProductsStep = createStep(
       ModuleRegistrationName.PRODUCT
     )
 
+    if ("products" in data) {
+      if (data.products.some((p) => !p.id)) {
+        throw new MedusaError(
+          MedusaError.Types.INVALID_DATA,
+          "Product id is required when doing a batch update"
```
olivermrbl: suggestion, very-nitpicky(non-blocking): The step is not aware of its consuming operation, so I would rename the error message to reflect that:
```suggestion
          "Product ID is required in input 'products'"
```

packages/core-flows/src/product/steps/update-product-variants.ts
```diff
@@ -16,6 +23,24 @@ export const updateProductVariantsStep = createStep(
       ModuleRegistrationName.PRODUCT
     )
 
+    if ("product_variants" in data) {
+      if (data.product_variants.some((p) => !p.id)) {
+        throw new MedusaError(
+          MedusaError.Types.INVALID_DATA,
+          "Product variant id is required when doing a batch update"
```
olivermrbl: Same as other comment.

packages/core-flows/src/product/steps/update-product-variants.ts
```diff
@@ -1,12 +1,19 @@
...
+  MedusaError,
+  getSelectsAndRelationsFromObjectArray,
+} from "@medusajs/utils"
 import { StepResponse, createStep } from "@medusajs/workflows-sdk"
 
-type UpdateProductVariantsStepInput = {
-  selector: ProductTypes.FilterableProductVariantProps
-  update: ProductTypes.UpdateProductVariantDTO
-}
+type UpdateProductVariantsStepInput =
```
olivermrbl, line 9: thought(non-blocking): Should we separate the two operations in two steps, similar to how we have two methods in the module interface? Same comment applies to the products step.

packages/core-flows/src/product/steps/batch-products.ts
```diff
@@ -0,0 +1,75 @@
...
+      { createTransaction, updateTransaction, deleteTransaction }
+    )
+  },
+
+  async (flow, { container }) => {
+    if (!flow) {
+      return
+    }
+
+    if (flow.createTransaction) {
```
olivermrbl, line 57: question: Do we know if this is the correct approach to handle such a scenario? If so, I think we should add this in our conventions doc.

packages/core-flows/src/pricing/steps/update-price-sets.ts
```diff
@@ -1,27 +1,48 @@
...
+  MedusaError,
   getSelectsAndRelationsFromObjectArray,
 } from "@medusajs/utils"
 import { StepResponse, createStep } from "@medusajs/workflows-sdk"
 
-type UpdatePriceSetsStepInput = {
-  selector?: PricingTypes.FilterablePriceSetProps
-  update?: PricingTypes.UpdatePriceSetDTO
-}
+type UpdatePriceSetsStepInput =
```
olivermrbl, line 9: Same as other comment on separation of concerns.

olivermrbl (commented):
This is great, have added a few comments and questions

packages/medusa/src/api-v2/admin/products/[id]/variants/op/batch/route.ts
Comment (sradevski) [reply packages/medusa/src/api-v2/admin/products/[id]/variants/op/batch/route.ts]: Yeah I think that might be better, let me put that in the conventions and update the code 👍

packages/core-flows/src/product/workflows/batch-products.ts
sradevski: Yup, I need to move a bunch of typings

packages/core-flows/src/product/steps/update-products.ts
sradevski: Sounds good, and I agree

packages/core-flows/src/product/steps/update-product-variants.ts
sradevski: @olivermrbl I am leaning towards not separating them because ultimately the step says "This step is for updating a variant", and it doesn't really matter if it is through a selector or you pass data. The fact that we use `upsert` on the module shouldn't matter for the workflow. 

I don't feel strongly about this though, so if you think it's better to separate it then that's fine with me, but I am not sure what we would name the other step 😄

packages/core-flows/src/product/steps/batch-products.ts
sradevski: @olivermrbl as far as Adrien and Carlos were concerned, yes this is how it works. I will add more testing once this is merged just to see if everything works as expected, but I wanted to at least have these methods available for the admin to be able to do pricing updates.

I'll also add it to the conventions, I'll add a ticket for myself.

olivermrbl: Cool, sounds good

packages/core-flows/src/product/steps/update-product-variants.ts
olivermrbl: I am fine keeping the step as is, just afraid it will end up getting messy in the future. But let's address it then, instead of doing anything premature now

olivermrbl (approved):

PR #7042, opened 2024-04-09
fix(dashboard): Minor fixes to region domain and api

kasperkristensen:
**What**
- Minor fixes to styling of domain, and ensures that display_name is returned for countries.

### DIFF
--- a/packages/admin-next/dashboard/src/v2-routes/regions/region-create/components/create-region-form/create-region-form.tsx
+++ b/packages/admin-next/dashboard/src/v2-routes/regions/region-create/components/create-region-form/create-region-form.tsx
@@ -16,24 +16,24 @@ import { useForm, useWatch } from "react-hook-form"
 import { useTranslation } from "react-i18next"
 import * as zod from "zod"
 
-import { RegionCountryDTO, PaymentProviderDTO } from "@medusajs/types"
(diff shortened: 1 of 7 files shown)


packages/admin-next/dashboard/src/v2-routes/regions/region-create/components/create-region-form/create-region-form.tsx
```diff
@@ -78,7 +78,7 @@ export const CreateRegionForm = ({
 
   const { t } = useTranslation()
 
-  const { mutateAsync, isLoading } = useCreateRegion()
+  const { mutateAsync, isPending } = useCreateRegion()
```
sradevski, line 81: Why is it `isPending` here? Should it be isPending everywhere or this hook in particular is defined like that?

sradevski, line 81: I guess we need to switch to isPending everywhere?

sradevski (approved):

packages/admin-next/dashboard/src/v2-routes/regions/region-create/components/create-region-form/create-region-form.tsx
kasperkristensen: Yes, isLoading was renamed to isPending in v5.

PR #7051, opened 2024-04-10
feat: Wait bullmq to finish processing jobs before being killed

sradevski:


packages/medusa/src/loaders/index.ts
```diff
@@ -149,12 +151,12 @@ async function loadMedusaV2({
   await createDefaultsWorkflow(container).run()
 
   const shutdown = async () => {
-    await promiseAll([
-      container.dispose(),
-      pgConnection?.context?.destroy(),
-      expressShutdown(),
-      medusaAppOnApplicationShutdown()
-    ])
+    await medusaAppOnApplicationShutdown(),
```
adrien2p: Todo: rm coma at the end

packages/medusa/src/loaders/index.ts
```diff
@@ -149,12 +151,12 @@ async function loadMedusaV2({
...
+      await promiseAll([
+        container.dispose(),
+        pgConnection?.context?.destroy(),
+        expressShutdown(),
```
adrien2p: Q: should we move that one before the medusa app shutdown to ensure no more requests can be handled?

packages/medusa/src/loaders/index.ts
```diff
@@ -339,13 +349,14 @@ export default async ({
...
 
   async function shutdown() {
+    await medusaAppOnApplicationShutdown()
+
     await promiseAll([
       container.dispose(),
       dbConnection?.destroy(),
       pgConnection?.context?.destroy(),
       redisShutdown(),
       expressShutdown(),
```
adrien2p, line 351: Same question here depending on the previous one 🙃

adrien2p (approved):

srindom:
Think we need to address this in `event-bus-redis` too?

packages/medusa/src/loaders/index.ts
sradevski: Yeah I think that makes sense, cc @edast will you take over the PR?

fPolic (approved):

packages/medusa/src/loaders/index.ts
edast: done

packages/workflow-engine-redis/src/services/workflow-orchestrator.ts
```diff
@@ -129,6 +145,11 @@ export class WorkflowOrchestratorService {
       container,
     } = options ?? {}
 
+    if (this.isShuttingDown) {
+      // don't accpet any more events
+      return
+    }
```
olivermrbl, line 151: question: ~~Will the workflow run get retried if we hit this case? Would be great if we could somehow test this. It's quite a crucial flow~~

packages/workflow-engine-redis/src/services/workflow-orchestrator.ts
```diff
@@ -110,7 +117,16 @@ export class WorkflowOrchestratorService {
...
+    this.logger.info("Workflow orchestrator gradeful shutdown.")
+    // set flag to shuting down
+    this.isShuttingDown = true
+    while (this.activeStepsCount > 0) {
+      this.logger.info(`Active steps: ${this.activeStepsCount}`)
+      // sleep for 1 second
+      await new Promise((resolve) => setTimeout(resolve, 1000))
+    }
     await this.redisDistributedTransactionStorage_.onApplicationShutdown()
+    this.logger.info("Workflow orchestrator finished!")
```
olivermrbl, line 129: thought: Not sure we should log all this stuff, great for debugging, but not super useful for other users
```suggestion
    this.isShuttingDown = true
    while (this.activeStepsCount > 0) {
      await new Promise((resolve) => setTimeout(resolve, 1000))
    }
    await this.redisDistributedTransactionStorage_.onApplicationShutdown()
```

olivermrbl:
Redis event bus should also be updated, such that the worker finishes ongoing jobs:
https://github.com/medusajs/medusa/blob/122b3ea76b790eff1e7ec64810662505370c26e9/packages/event-bus-redis/src/services/event-bus-redis.ts#L64-L68

PR #7053, opened 2024-04-10
[feat] add setToken auth resource in js sdk

markus-gx:
Resolves #7052 

This introduces a method for the `.auth`-Resource which allows to set an access token to the SDK. 

Reason: 

- Different Frameworks need to handle sessions/tokens differently - this makes it easier if one wants to handle the token on their own, but needs to inject it for the requests to work.


riqwan:

If you would still like to open this PR, please rebase with the v1 branch, point to the v1 branch and re-request a review & we'll get back on it quickly.

PR #7073, opened 2024-04-15
fix-docs: TS error when importing a function as a type only (example logged in user middleware)

adevinwild:


shahednasser (approved):
LGTM thanks!

PR #7077, opened 2024-04-15
feat(medusa,core-flows,types): adds batch operations to price list prices

riqwan:
what:

- adds batch operations to price list prices

RESOLVES CORE-1969
RESOLVES CORE-1970

### DIFF
--- a/.github/actions/cache-deps/action.yml
+++ b/.github/actions/cache-deps/action.yml
@@ -11,9 +11,9 @@ runs:
       with:
         path: |
           .yarn/cache
-        key: ${{ runner.os }}-yarn-${{inputs.extension}}-v8-${{ github.runner_id }}-${{ hashFiles('**/yarn.lock') }}
+        key: ${{ runner.os }}-yarn-${{inputs.extension}}-v8-${{ hashFiles('**/yarn.lock') }}
         restore-keys: |
-          ${{ runner.os }}-yarn-${{inputs.extension}}-v8-${{ github.runner_id }}
+          ${{ runner.os }}-yarn-${{inputs.extension}}-v8
       # We want to only bootstrap and install if no cache is found.
     - run: yarn install --immutable
       shell: bash
--- a/.github/workflows/action.yml
+++ b/.github/workflows/action.yml
@@ -2,10 +2,10 @@ name: Medusa Pipeline
 on:
   push:
     branches:
-      - develop
+    - develop
   pull_request:
     branches:
-      - develop
+    - develop
 
 jobs:
   setup:


packages/medusa/src/api-v2/admin/price-lists/validators.ts
```diff
@@ -104,48 +105,27 @@ export class AdminPostPriceListsPriceListReq {
...
-
-  @IsOptional()
-  @IsInt()
-  max_quantity?: number
-
-  @IsOptional()
-  @IsObject()
-  rules?: Record<string, string>
-}
+export const AdminCreatePriceListPrice = z.object({
```
sradevski, line 16: Also define the type of it like this above it:
```suggestion
export type AdminCreatePriceListPriceType = z.infer<
  typeof AdminCreatePriceListPrice
>
export const AdminCreatePriceListPrice = z.object({
```

And then pass that to `AuthenticatedMedusaRequest<AdminCreatePriceListPriceType>`

This applies for pretty much all validators

packages/medusa/src/api-v2/admin/price-lists/middlewares.ts
```diff
@@ -50,19 +49,7 @@ export const adminPriceListsRoutesMiddlewares: MiddlewareRoute[] = [
...
-    middlewares: [
-      transformBody(AdminPostPriceListsPriceListPricesBatchRemoveReq),
-    ],
-  },
-  {
-    method: ["POST"],
-    matcher: "/admin/price-lists/:id/prices/batch/update",
-    middlewares: [transformBody(AdminPostPriceListPriceBatchUpdate)],
+    matcher: "/admin/price-lists/:id/prices/batch",
+    middlewares: [validateAndTransformBody(AdminBatchPriceListPrices)],
```
sradevski: Also add `validateAndTransformQuery` after the body transformation. Applies to all routes basically, since we always want to return the complete entity

packages/medusa/src/api-v2/admin/price-lists/[id]/prices/batch/route.ts
```diff
@@ -0,0 +1,38 @@
...
+      },
+    },
+    throwOnError: false,
+  })
+
+  if (Array.isArray(errors) && errors[0]) {
+    throw errors[0].error
+  }
+
+  res.status(200).json(result)
```
Comment (sradevski) [inline packages/medusa/src/api-v2/admin/price-lists/[id]/prices/batch/route.ts]: We would want to iterate over the result and refetch the data (see the product module). I know it's a bit wasteful, but that's what we've agreed on as a convention, so we always have a consistent response and the users can choose which fields should be returned

packages/medusa/src/api-v2/admin/price-lists/middlewares.ts
```diff
@@ -50,19 +49,7 @@ export const adminPriceListsRoutesMiddlewares: MiddlewareRoute[] = [
...
-    matcher: "/admin/price-lists/:id/prices/batch/remove",
-    middlewares: [
-      transformBody(AdminPostPriceListsPriceListPricesBatchRemoveReq),
-    ],
-  },
-  {
-    method: ["POST"],
-    matcher: "/admin/price-lists/:id/prices/batch/update",
-    middlewares: [transformBody(AdminPostPriceListPriceBatchUpdate)],
+    matcher: "/admin/price-lists/:id/prices/batch",
```
sradevski, line 65: I guess we don't have a `:id/prices/:price_id` route here so it works as expected, but I wasn't so lucky in the product module as there is some route conflict happening 😄

packages/core-flows/src/price-list/workflows/batch-price-list-prices.ts
```diff
@@ -0,0 +1,59 @@
...
+
+    const updatePriceListPricesInput = transform(
+      { data: input.data, map: variantPriceUpdateMap },
+      ({ data, map }) => ({
+        data: [{ id: data.id, prices: data.update }],
+        variant_price_map: map,
+      })
+    )
+
+    const [created, updated, deleted] = parallelize(
```
sradevski, line 31: I think it's better to have a `batch-price-list-prices` step, and inside it call the `*-price-list-prices` workflows instead, since there might be more to a workflow of creating a price than just calling this one step, so it's more future-proof like that.

Check the product workflows to see how I've done it after some discussions with Adrien and Carlos.

packages/core-flows/src/price-list/workflows/batch-price-list-prices.ts
```diff
@@ -0,0 +1,59 @@
...
+      validateVariantPriceLinksStep([{ prices: input.data.create }]),
+      validateVariantPriceLinksStep([{ prices: input.data.update }]).config({
+        name: "variant-price-link-map-update",
+      })
+    )
+
+    const createPriceListPricesInput = transform(
+      { data: input.data, map: variantPriceCreateMap },
+      ({ data, map }) => ({
+        data: [{ id: data.id, prices: data.create }],
```
sradevski: Maybe this normalization can be done at the HTTP layer instead, and we always expect that the price set ID is passed in the workflow?

riqwan: The id here is the price list id, I think its ok to transform a bit here to make it work for other workflows.

riqwan:
@sradevski - this one works with compensation, each workflow being in its own step.

packages/core-flows/src/price-list/workflows/batch-price-list-prices.ts
sradevski: @riqwan yeah either way works, I was just thinking what we'd expect from the input, if you prefer it here then that's fine with me 👍

sradevski (approved):
Nice 👍 I'll switch the product module as well to have each workflow as a step. We did talk about allowing you to do something like `xWorkflow.asStep()` rather than creating new steps manually, so maybe we can do that sooner than later (once Carlos and Adrien are back). This is great until then 👍

PR #7084, opened 2024-04-16
fix: Support bigNumber in upsertWithReplace

sradevski:
It's not the most elegant of solutions due to how we've implemented our BigNumber representation, but I believe it's good-enough

### DIFF
--- a/packages/utils/src/dal/mikro-orm/mikro-orm-repository.ts
+++ b/packages/utils/src/dal/mikro-orm/mikro-orm-repository.ts
@@ -639,16 +639,6 @@ export function mikroOrmBaseRepositoryFactory<T extends object = object>(
             Object.assign(normalizedDataItem, {
               ...joinColumnsConstraints,
             })
-            // Non-persist relation columns should be removed before we do the upsert.
-            Object.entries(relation.targetMeta?.properties ?? {})
-              .filter(
-                ([_, propDef]) =>
-                  propDef.persist === false &&
-                  propDef.reference === ReferenceType.MANY_TO_ONE
-              )
-              .forEach(([key]) => {
-                delete normalizedDataItem[key]
-              })
           })
 
           await this.upsertMany_(manager, relation.type, normalizedData)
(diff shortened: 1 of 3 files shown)


adrien2p:
Just jumping real quick, maybe it is worth adding a metadata that you can use to identify big number fields representations and their property names (the way we do for searchable properties, which is also the way custom types allows us to do)

packages/utils/src/dal/mikro-orm/mikro-orm-repository.ts
```diff
@@ -717,7 +717,28 @@ export function mikroOrmBaseRepositoryFactory<T extends object = object>(
...
+            [key, value],
+            [nonRawFieldKey, nonRawField],
+          ]
+        })
+        .reduce((acc, [key, value]) => {
+          acc[key] = value
+          return acc
+        }, {})
+
+      return { id: (created as any).id, ...bigNumberFields, ...data }
```
riqwan: we could bring this closer to the mikroORM implementation that prevents us from hard coding `raw_`, here is a quick concept - [commit](https://github.com/medusajs/medusa/compare/fix/test?expand=1#diff-03588856044247b1b2d450be24ae9f199407d1dfd1089cdb17d89fcae6ca3f5eR718-R725)

MikroORM does a whole song and dance around diff comparators, which i think is very neat, but not necessary for our case for now. Here is a ref - [MikroORM code](https://github.com/mikro-orm/mikro-orm/blob/master/packages/core/src/entity/EntityFactory.ts#L179-L180)

olivermrbl: Looks pretty neat to me – anything stopping us from moving forward with this?

sradevski: I think longer-term, we really need to revisit our implementation of BigNumber, and implement it using a custom type rather than this sort of wrapper around `Property`. But your suggestion is better than what I did @riqwan, so let me do that for now 👍

sradevski: @riqwan seems to be working well, great suggestion, much nicer!

packages/utils/src/dal/mikro-orm/mikro-orm-repository.ts
```diff
@@ -716,8 +716,12 @@ export function mikroOrmBaseRepositoryFactory<T extends object = object>(
         managed: false,
         persist: false,
       })
+      const loadedData: object = {}
+      for (const property of (created as any).__helper.__loadedProperties) {
+        loadedData[property] = created[property]
+      }
```
riqwan: nit: we could return the loaded data directly.

```suggestion
      const loadedData: object = { id: (created as any).id }
      
      for (const property of (created as any).__helper.__loadedProperties) {
        loadedData[property] = created[property]
      }
```

riqwan (approved):

packages/utils/src/dal/mikro-orm/mikro-orm-repository.ts
sradevski: @riqwan found another solution, and this fixes the issue with non-persist columns not being removed (and resulted in a bit of cleanup).

PR #7087, opened 2024-04-17
refactor(dashboard): refresh domains

fPolic:
**What**
Refresh following domains: 
- sales channels
  - fixes, toast 
- inventory
  - fixes, toast 
- invite
  - fixes, toast 
- login
  - properly display login error 
- store
  - toast, add support for `q` param on currencies, enable editing default currency and region
- profile
  - fixes, toast 
- locations
  - fixes, toast  
  - _NOTE:_ delete seems as not working because RQ returns soft deleted location on the list endpoint?


olivermrbl (approved):
LGTM, nice work.

PR #7089, opened 2024-04-17
fix: correct calculations of subtotal for shipping option requirements

kevinfurmanski:
When using min/max subtotal for shipping options, there is a case when the `listCartOptions` method in medusa-js will return the wrong option. If I have a "Paid shipping" option between 0-999 SEK costing 29 SEK and a "Free shipping" option of orders above 999 SEK and the line items amount to for example 980 SEK, then when you select the "Paid shipping" option for 29 SEK the order total goes over 999 SEK and the "Paid shipping" option is hidden and instead the "Free shipping" option is shown. When you then press the "Free shipping" option the order total goes under 999 SEK and the "Free shipping" option is hidden and the "Paid shipping" option is shown again. It's an endless cycle that you can't get out of due to the fact that the shipping cost is included in the total. This should be modified to not include the shipping price.

### DIFF
--- a/.changeset/lemon-forks-collect.md
+++ b/.changeset/lemon-forks-collect.md
@@ -1,0 +1,5 @@
+---
+"@medusajs/medusa": patch
+---
+
+fix: correct calculations of subtotal for shipping option requirements


sradevski (approved):
LGTM!

kevinfurmanski:
@sradevski I added a changeset and rebased from develop

sradevski:
@olivermrbl Just a heads up, maybe we should also include this in today's release.

PR #7092, opened 2024-04-17
feat: Update product category

olivermrbl:
**What**
- Update product category flow
- Drop composite unique index on `parent_category_id` and `rank`*

*It seems MikroORM has an issue with batch updating rows with a composite unique index. For the time being, we drop the index. We can add it back when the issue has been resolved.


riqwan (approved):
lgtm!

packages/product/src/repositories/product-category.ts
```diff
@@ -394,6 +394,10 @@ export class ProductCategoryRepository extends DALUtils.MikroOrmBaseTreeReposito
       id: { $ne: targetCategoryId },
     })
 
+    const allCategories = await manager.find(ProductCategory, {
+      parent_category_id: targetParentId,
+    })
+
```
riqwan: is this added here for a reason?

olivermrbl: Damn, leftover from debugging. Nice catch

integration-tests/api/__tests__/admin/product-category.ts
```diff
@@ -774,73 +779,138 @@ medusaIntegrationTestRunner({
...
-          dbConnection,
-          {
-            name: "category child 1",
-            parent_category: productCategoryChild,
-            rank: 1,
-          }
-        )
+        await breaking(
+          async () => {
+            productCategoryParent = await simpleProductCategoryFactory(
```
sradevski, line 784: Not for now, but a general comment - It's better to do the seeding through the HTTP API as well so we don't access internals (in this case, modules) if not absolutely necessary

packages/product/src/migrations/InitialSetup20240401153642.ts
```diff
@@ -63,7 +63,9 @@ export class InitialSetup20240315083440 extends Migration {
     this.addSql('create unique index if not exists "IDX_category_handle_unique" on "product_category" (handle) where deleted_at is null;')
     this.addSql('create index if not exists "IDX_product_category_path" on "product_category" ("mpath") where deleted_at is null;');
     this.addSql('create index if not exists "IDX_product_category_deleted_at" on "product_collection" ("deleted_at");');
-    
+    // TODO: Batch updating composite unique index in MikroORM is faulty. Should be added back when issue has been resolved.
```
sradevski, line 66: What is the behavior exactly? Since it is not a partial index, would a `UNIQUE` constraint work?

sradevski (approved):

packages/product/src/migrations/InitialSetup20240401153642.ts
olivermrbl: Given two categories with a rank and same parent:
```
{ id: "1234", rank: 1, parent_id: "4242" }
{ id: "4321", rank: 2, parent_id: "4242" }
```

and we attempt to update the rank:
```
{ id: "1234", rank: 2, parent_id: "4242" }
{ id: "4321", rank: 1, parent_id: "4242" }
```

MikroORM will throw an error saying we are violating the unique constraints because it doesn't properly handle a composite unique index.

>  would a UNIQUE constraint work?
Wouldn't this lead to the same outcome? I can give it a shot.

sradevski: Ah ok got it, I guess since the index check happens as each update is performed, it temporarily violates the unique constraint. It looks like a Postgres limitation, see: https://stackoverflow.com/questions/43574582/postgresql-bulk-update-having-unique-index @olivermrbl

riqwan: Constraints can't work in our case because we need to include "null" values for parent_category_id in our constraint, which constraints don't consider. Additionally, we also need partial unique composite indexes now that categories are soft deletable, which constraints also don't allow.

PR #7094, opened 2024-04-18
fix: Switch to zod for customer endpoints, fix inconsistencies

sradevski:


integration-tests/modules/__tests__/customer-group/admin/list-customer-group-customers.ts
```diff
@@ -54,7 +54,7 @@ medusaIntegrationTestRunner({
         )
 
         const response = await api.get(
-          `/admin/customer-groups/${group.id}/customers`,
```
sradevski, line 57: This is an anti-pattern and we should just use the `customers` endpoint with a filter (as done for eg. for products and price lists)

packages/medusa/src/api-v2/admin/customer-groups/[id]/customers/batch/route.ts
```diff
@@ -0,0 +1,66 @@
+import { AdminCustomerGroupResponse, BatchMethodRequest } from "@medusajs/types"
```
Comment (olivermrbl) [inline packages/medusa/src/api-v2/admin/customer-groups/[id]/customers/batch/route.ts:1]: A general thought: Are we sure we want to remove existing batch endpoints? If so, I think we keep adding todos to the admin clean-up. Maybe keeping these in and removing them in our polishing phase is better?

olivermrbl (approved):

fPolic (approved):

packages/medusa/src/api-v2/admin/customer-groups/[id]/customers/batch/route.ts
Comment (sradevski) [reply packages/medusa/src/api-v2/admin/customer-groups/[id]/customers/batch/route.ts:1]: @olivermrbl it's a pretty small change in 1 place, so I'd rather do it sooner than later. I am actually planning to just unify all of the batch endpoints in 1 go, both on FE and BE as it will be faster

PR #7095, opened 2024-04-18
feat(core-flows,medusa,types): cancel fulfillments API

riqwan:
what:

- adds POST cancel fulfillments endpoint

RESOLVES CORE-1972


packages/medusa/src/api-v2/admin/fulfillments/[id]/route.ts
```diff
@@ -0,0 +1,38 @@
...
+  AdminFulfillmentDeleteResponse,
+  IFulfillmentModuleService,
+} from "@medusajs/types"
+
+import {
+  AuthenticatedMedusaRequest,
+  MedusaResponse,
+} from "../../../../types/routing"
+
+export const DELETE = async (
```
Comment (riqwan) [inline packages/medusa/src/api-v2/admin/fulfillments/[id]/route.ts]: @olivermrbl since we're not exposing delete for fulfillment in the module, am i right in assuming that this is a cancel endpoint? If so, wouldn't it make more sense to call this `POST /admin/fulfillments/:id/cancel`?

Comment (sradevski) [reply packages/medusa/src/api-v2/admin/fulfillments/[id]/route.ts]: I think that makes sense, since we are not deleting anything here.

packages/medusa/src/api-v2/admin/fulfillments/[id]/route.ts
```diff
@@ -0,0 +1,38 @@
...
+  req: AuthenticatedMedusaRequest,
+  res: MedusaResponse<AdminFulfillmentDeleteResponse>
+) => {
+  const { id } = req.params
+  const fulfillmentModuleService = req.scope.resolve<IFulfillmentModuleService>(
+    ModuleRegistrationName.FULFILLMENT
+  )
+
+  await fulfillmentModuleService.retrieveFulfillment(id)
```
Comment (sradevski) [inline packages/medusa/src/api-v2/admin/fulfillments/[id]/route.ts]: Is this only for validation? If so, I'd suggest we do it in the workflow instead

packages/medusa/src/api-v2/admin/fulfillments/[id]/route.ts
Comment (olivermrbl) [reply packages/medusa/src/api-v2/admin/fulfillments/[id]/route.ts]: Yup, makes sense 👍

packages/medusa/src/api-v2/admin/fulfillments/[id]/route.ts
Comment (olivermrbl) [reply packages/medusa/src/api-v2/admin/fulfillments/[id]/route.ts]: I have come across a similar case a few times, and I can't figure out what the right thing to do is, so maybe we should establish a convention.

Typically, our workflows expect a specific shape of data for bulk mutating a resource and therefore do not have validation built-in – as they would have in the case of a single mutation through `retrieve`. Given that our HTTP layer almost always operates on a single (root) resource, one could argue the validation is well-placed in the controller.

E.g. right now, sending a request to update a product with an ID that does not exist will perform a selector-based update. There is not a product to update, so the return value of the workflow in an empty array and the subsequent call to fetch the product fails.

Comment (sradevski) [reply packages/medusa/src/api-v2/admin/fulfillments/[id]/route.ts]: @olivermrbl you have a point due to the fact that we do a `select`-based mutation. My thought-process is that workflows should validate that what they are operating on is correct, since they can be executed from a non-http flow, but in this specific scenario I can see why it might make sense to do the validation here, cc @riqwan WDYT?

Comment (riqwan) [reply packages/medusa/src/api-v2/admin/fulfillments/[id]/route.ts]: I think its a moot point for this endpoint since the module internally calls `retrieveFulfillment` to validate its existence.

I agree that the validation generally being on the workflow makes total sense as its built to run independently, but running validations on the endpoint is also totally valid in cases where the endpoint needs to have agency over its job of rendering the response graciously and consistently.

packages/core-flows/src/fulfillment/steps/cancel-fulfillments.ts
```diff
@@ -0,0 +1,18 @@
...
+
+export const cancelFulfillmentsStepId = "cancel-fulfillments"
+export const cancelFulfillmentsStep = createStep(
+  cancelFulfillmentsStepId,
+  async (ids: string[], { container }) => {
+    const service = container.resolve<IFulfillmentModuleService>(
+      ModuleRegistrationName.FULFILLMENT
+    )
+
+    await promiseAll(ids.map((id) => service.cancelFulfillment(id)))
```
riqwan: @olivermrbl this looks like a non reversible step. Should we remove bulk capabilities here if that is the case?

packages/medusa/src/api-v2/admin/fulfillments/[id]/route.ts
Comment (sradevski) [reply packages/medusa/src/api-v2/admin/fulfillments/[id]/route.ts]: The only aspect to consider here is to not fire 3-4 DB requests for just checking if something exists, which is why I suggested moving the check elsewhere. Ideally the HTTP endpoint can "clean up" the response in a purely synchronous manner, but I agree that there might be times where an async call to some endpoint is needed for a better response.

Anyhow, it's not too important for now, but these are considerations we'll need to make as we make v2.0 more prod-ready

packages/core-flows/src/fulfillment/steps/cancel-fulfillments.ts
olivermrbl: I think that makes sense, nice catch. We should treat all steps that interface with 3rd party providers this way – as we did with payments already. So bulk operations would have to be handled "manually" by the consumer

riqwan: nice, changed em!

olivermrbl (approved):
LGTM

PR #7096, opened 2024-04-18
chore: Move most of the remaining endpoints to zod

sradevski:


packages/medusa/src/api-v2/admin/api-keys/[id]/sales-channels/batch/route.ts
```diff
@@ -0,0 +1,83 @@
...
+  )
+
+  if (apiKey.type !== ApiKeyType.PUBLISHABLE) {
+    throw new MedusaError(
+      MedusaError.Types.INVALID_DATA,
+      "Sales channels can only be associated with publishable API keys"
+    )
+  }
+
+  if (create && create.length) {
```
Comment (fPolic) [inline packages/medusa/src/api-v2/admin/api-keys/[id]/sales-channels/batch/route.ts:34]:
```suggestion
  if ( create?.length) {
```

packages/medusa/src/api-v2/admin/api-keys/[id]/sales-channels/batch/route.ts
```diff
@@ -0,0 +1,83 @@
...
+      input: workflowInput,
+      throwOnError: false,
+    })
+
+    if (Array.isArray(errors) && errors[0]) {
+      throw errors[0].error
+    }
+  }
+
+  if (toDelete && toDelete.length) {
```
Comment (fPolic) [inline packages/medusa/src/api-v2/admin/api-keys/[id]/sales-channels/batch/route.ts:54]:
```suggestion
  if (toDelete?.length) {
```

fPolic (approved):
LGTM

packages/medusa/src/api-v2/admin/api-keys/[id]/sales-channels/batch/route.ts
```diff
@@ -0,0 +1,83 @@
...
+    const workflowInput = {
+      data: [
+        {
+          api_key_id: req.params.id,
+          sales_channel_ids: toDelete,
+        },
+      ],
+    }
+
+    const { errors } = await removeSalesChannelsFromApiKeyWorkflow(
```
Comment (riqwan) [inline packages/medusa/src/api-v2/admin/api-keys/[id]/sales-channels/batch/route.ts:64]: q: any reason why these aren't in one workflow?

Context - if the delete fails, you can't revert the creates. If the case is that only one is running at a time, maybe we can do an XOR validation with zod

riqwan (approved):
💪🏻

packages/medusa/src/api-v2/admin/api-keys/[id]/sales-channels/batch/route.ts
Comment (sradevski) [reply packages/medusa/src/api-v2/admin/api-keys/[id]/sales-channels/batch/route.ts:64]: @riqwan I'll move them to a single workflow, it was just too many changes to do at once so I figured I'll just make the user-facing HTTP interface to what we want it to be first

PR #7098, opened 2024-04-18
Chore/finish migration zod

sradevski:


integration-tests/modules/__tests__/pricing/admin/rule-types.spec.ts
```diff
@@ -122,9 +122,9 @@ medusaIntegrationTestRunner({
             .catch((e) => e)
 
           expect(response.status).toEqual(400)
-          expect(response.data.message).toEqual(
-            "name must be a string, name should not be empty"
-          )
+          // expect(response.data.message).toEqual(
```
sradevski, line 125: The zod errors are rubbish now, I created a ticket for us to fix it

packages/medusa/src/api-v2/admin/draft-orders/helpers.ts
```diff
@@ -0,0 +1,23 @@
...
+export const refetchOrder = async (
+  orderId: string,
+  scope: MedusaContainer,
+  fields: string[]
+) => {
+  const remoteQuery = scope.resolve(ContainerRegistrationKeys.REMOTE_QUERY)
+  const queryObject = remoteQueryObjectFromString({
+    entryPoint: "order",
+    variables: {
+      filters: { id: orderId },
```
olivermrbl, line 16: question: Shouldn't this have a permanent filter on `status: "draft"`?

packages/medusa/src/api-v2/utils/validate-body.ts
```diff
@@ -4,17 +4,12 @@ import { z, ZodError } from "zod"
...
+  body: T
 ): Promise<z.ZodRawShape> {
   try {
-    let schema = zodSchema
-    if (config.strict) {
-      schema = schema.strict()
-    }
-
-    return await schema.parseAsync(body)
+    zodSchema
```
olivermrbl, line 11: question: What happened here? Is this intentional?

olivermrbl (approved):
LGTM, just a few questions

packages/medusa/src/api-v2/utils/validate-body.ts
sradevski: @olivermrbl the schema can return one of two types - ZodType and ZodEffect. You can only apply `strict` on ZodType. ZodEffect is returned if you do a `.refine` on the schema, which is a must for any XOR logic or cross-data checks. Therefore we cannot have this config, as `strict()` would have to be applied before calling `refine`.

I also think we can always have strict schemas, and if not we can just do the nonStrict definition as part of the validator, than as a config. Let me know if you think we can handle this in some other way

PR #7101, opened 2024-04-19
feat(core-flows,medusa,types): fulfillment API: create

riqwan:
what:

- adds fulfillment create API

RESOLVES CORE-1962

### DIFF
--- a/.github/workflows/generate-references.yml
+++ b/.github/workflows/generate-references.yml
@@ -60,6 +60,56 @@ jobs:
           add-paths: www/apps/api-reference/specs
           branch: "docs/generate-api-ref"
           branch-suffix: "timestamp"
+  api-v2:
+    runs-on: ubuntu-latest
+    if: ${{ github.event_name == 'release' || github.event.inputs.referenceName == 'all' || github.event.inputs.referenceName == 'api' }}
+    steps:
+      - name: Cancel Previous Runs
+        uses: styfle/cancel-workflow-action@0.11.0
+        with:
+          access_token: ${{ github.token }}
+
+      - name: Checkout
+        uses: actions/checkout@v3
+        with:
+          token: ${{ secrets.REFERENCE_PAT }}
+          fetch-depth: 0
+
+      - name: Setup Node.js environment
+        uses: actions/setup-node@v3
+        with:
+          node-version: "16.10.0"
+          cache: "yarn"
+
+      - name: Install dependencies
+        uses: ./.github/actions/cache-deps
+        with:
+          extension: reference
+
+      - name: Build Packages
+        run: yarn build
+
+      - name: Generate API Reference (v2)
+        run: yarn openapi:generate --v2
+
+      - name: Install and Build docs-util dependencies
+        run: yarn && yarn build
+        working-directory: docs-util
+
+      - name: Generate Changeset
+        run: "yarn generate:changeset"
+        working-directory: docs-util/packages/scripts
+
+      - name: Create Pull Request
+        uses: peter-evans/create-pull-request@v4
+        with:
+          commit-message: "chore(docs): Generated API Reference (v2)"
+          base: "develop"
+          title: "chore(docs): Updated API Reference (v2)"
+          labels: "type: chore"
+          add-paths: www/apps/api-reference/specs
+          branch: "docs/generate-api-ref"
+          branch-suffix: "timestamp"
   ui:
     runs-on: ubuntu-latest
     if: ${{ github.event_name == 'release' || github.event.inputs.referenceName == 'all' || github.event.inputs.referenceName == 'ui' }}
--- a/www/apps/docs/content/development/api-routes/example-logged-in-user.mdx
+++ b/www/apps/docs/content/development/api-routes/example-logged-in-user.mdx
@@ -23,11 +23,11 @@ Learn more about [middlewares in its guide](./add-middleware.mdx).
 Create the file `src/api/middlewares.ts` with the following content:
 
 ```ts title="src/api/middlewares.ts"
-import type { 
+import { 
   authenticate,
-  MiddlewaresConfig, 
-  User, 
-  UserService,
+  type MiddlewaresConfig, 
+  type User, 
+  type UserService,
 } from "@medusajs/medusa"
 import type { 
   MedusaNextFunction, 


packages/core-flows/src/fulfillment/steps/create-fulfillment.ts
```diff
@@ -0,0 +1,28 @@
...
+  async (ids, { container }) => {
+    if (!ids?.length) {
+      return
+    }
+
+    const service = container.resolve<IFulfillmentModuleService>(
+      ModuleRegistrationName.FULFILLMENT
+    )
+
+    await service.delete(ids)
```
olivermrbl: todo: 
```suggestion
    await service.cancelFulfillment(ids)
```

packages/core-flows/src/fulfillment/steps/create-fulfillment.ts
```diff
@@ -0,0 +1,28 @@
...
+    )
+
+    const fulfillment = await service.createFulfillment(data)
+
+    return new StepResponse(fulfillment, fulfillment.id)
+  },
+  async (ids, { container }) => {
+    if (!ids?.length) {
+      return
+    }
```
olivermrbl: todo: We are only working with a single fulfillment at a time, so this is not an array. Perhaps, we should add a test for the compensation flow?
```suggestion
    if (!id) {
      return
    }
```

riqwan: good catch, added it in

packages/core-flows/src/fulfillment/steps/create-fulfillment.ts
```diff
@@ -0,0 +1,28 @@
...
+  async (id, { container }) => {
+    if (!id) {
+      return
+    }
+
+    const service = container.resolve<IFulfillmentModuleService>(
+      ModuleRegistrationName.FULFILLMENT
+    )
+
+    await service.cancelFulfillment(id)
```
sradevski, line 26: Wouldn't make sense to fully delete the fulfillment here, considering it's a compensating operation?

packages/medusa/src/api-v2/admin/fulfillments/route.ts
```diff
@@ -0,0 +1,31 @@
...
+  MedusaResponse,
+} from "../../../types/routing"
+import { refetchFulfillment } from "./helpers"
+import { AdminCreateFulfillmentType } from "./validators"
+
+export const POST = async (
+  req: AuthenticatedMedusaRequest<AdminCreateFulfillmentType>,
+  res: MedusaResponse
+) => {
+  const { result: fullfillment, errors } = await createFulfillmentWorkflow(
```
sradevski, line 13: Typically the workflows take an array, should we do the same for a fulfillment?

packages/medusa/src/api-v2/admin/fulfillments/validators.ts
```diff
@@ -3,5 +3,42 @@ import { createSelectParams } from "../../utils/validators"
...
 
-export const AdminCancelFulfillment = z.object({})
 export type AdminCancelFulfillmentType = z.infer<typeof AdminCancelFulfillment>
+export const AdminCancelFulfillment = z.object({})
+
+export type AdminCreateFulfillmentType = z.infer<typeof AdminCreateFulfillment>
+export const AdminCreateFulfillment = z.object({
+  location_id: z.string(),
+  provider_id: z.string(),
+  delivery_address: z.object({
```
sradevski: There is an `AddressPayload` in `common-validators` (it was used in the orders), should we standardize the model and use that instead?

packages/medusa/src/api-v2/admin/fulfillments/validators.ts
```diff
@@ -3,5 +3,42 @@ import { createSelectParams } from "../../utils/validators"
...
+    address_2: z.string().optional(),
+    city: z.string().optional(),
+    country_code: z.string().optional(),
+    province: z.string().optional(),
+    postal_code: z.string().optional(),
+    phone: z.string().optional(),
+    metadata: z.object({}).optional(),
+  }),
+  items: z.array(
+    z.object({
```
sradevski: NIT: Extract this object into a separate variable eg. `FulfillmentItem`

packages/medusa/src/api-v2/admin/fulfillments/validators.ts
```diff
@@ -3,5 +3,42 @@ import { createSelectParams } from "../../utils/validators"
...
+      title: z.string(),
+      sku: z.string(),
```
sradevski: Are both an SKU and barcode required?

packages/core-flows/src/fulfillment/steps/create-fulfillment.ts
riqwan: From what I can understand, create can potentially call an external provider and cancel does the same. I think it would be beneficial to keep those as logged and not delete it.

packages/medusa/src/api-v2/admin/fulfillments/route.ts
riqwan: same reason as above, since this calls an external provider, we're keeping them non-bulk.

packages/core-flows/src/fulfillment/steps/create-fulfillment.ts
sradevski: Ok make sense 👍

packages/medusa/src/api-v2/admin/fulfillments/route.ts
sradevski: Ok make sense 👍

packages/medusa/src/api-v2/admin/fulfillments/validators.ts
riqwan: According to the module, both are required. Any reason why it wouldn't be?

packages/medusa/src/api-v2/admin/fulfillments/validators.ts
riqwan: nice! I've addressed this [here](https://github.com/medusajs/medusa/pull/7118/files#diff-3717666eaae8ef3178a183f8f85232d20d40e56d75d1edf6f534859c6147d96e)

packages/medusa/src/api-v2/admin/fulfillments/validators.ts
riqwan: I've addressed this [here](https://github.com/medusajs/medusa/pull/7118/files#diff-3717666eaae8ef3178a183f8f85232d20d40e56d75d1edf6f534859c6147d96e)

packages/medusa/src/api-v2/admin/fulfillments/validators.ts
sradevski: @riqwan both are unique identifiers, so I was wondering if it's enough to have one or the other

sradevski (approved):

PR #7102, opened 2024-04-19
feat: Add batch method to collection and clean up some batch implementations

sradevski:


packages/medusa/src/api-v2/admin/collections/[id]/products/route.ts
```diff
@@ -0,0 +1,33 @@
...
+  })
+
+  if (Array.isArray(errors) && errors[0]) {
+    throw errors[0].error
+  }
+
+  res.status(200).json({
+    added: add,
+    removed: remove,
+  })
```
Comment (olivermrbl) [inline packages/medusa/src/api-v2/admin/collections/[id]/products/route.ts]: todo: As discussed, batch linking endpoints should return the root resource

packages/medusa/src/api-v2/admin/price-lists/[id]/products/route.ts
```diff
@@ -0,0 +1,57 @@
...
+import { LinkMethodRequest } from "@medusajs/types/src"
+import { fetchPriceListPriceIdsForProduct } from "../../helpers"
+
+export const POST = async (
+  req: AuthenticatedMedusaRequest<LinkMethodRequest>,
+  res: MedusaResponse
+) => {
+  const id = req.params.id
+  const { add, remove = [] } = req.validatedBody
+  if (add?.length) {
```
Comment (olivermrbl) [inline packages/medusa/src/api-v2/admin/price-lists/[id]/products/route.ts:16]: question: If we already have the batch endpoint, should we leave this out?

comment: Should `/admin/price-lists/:id/prices/batch` not be renamed to not include `/batch`?

olivermrbl (approved):
The response of the batch linking endpoints should be updated according to our discussion, but approving this, so we can merged after

packages/medusa/src/api-v2/admin/price-lists/[id]/products/route.ts
Comment (sradevski) [reply packages/medusa/src/api-v2/admin/price-lists/[id]/products/route.ts:16]: @olivermrbl I discussed this with Riqwan, and the `/prices` endpoint is a CRUD endpoint to create price entities, and has no relation with a product. This endpoint is about linking/unlinking products from a price list, so it is much closer in nature to the linking/unlinking batch methods than the CRUD ones. Let me know if you think differently.

Comment (olivermrbl) [reply packages/medusa/src/api-v2/admin/price-lists/[id]/products/route.ts:16]: Ah yes, this makes sense. I missed that.

PR #7116, opened 2024-04-22
feat: Convert several batch methods to the standardized format

sradevski:


packages/core-flows/src/api-key/steps/link-sales-channels-to-publishable-key.ts
```diff
@@ -0,0 +1,62 @@
+import { Modules } from "@medusajs/modules-sdk"
```
riqwan, line 1: sugg: This looks like it could be a generic link and unlink step that could be used across all these add/remove workflows

packages/medusa/src/api-v2/admin/api-keys/[id]/sales-channels/route.ts
```diff
@@ -0,0 +1,44 @@
...
+) => {
+  const { add, remove } = req.validatedBody
+  const apiKey = await refetchApiKey(req.params.id, req.scope, ["id", "type"])
+
+  if (apiKey.type !== ApiKeyType.PUBLISHABLE) {
+    throw new MedusaError(
+      MedusaError.Types.INVALID_DATA,
+      "Sales channels can only be associated with publishable API keys"
+    )
+  }
```
Comment (riqwan) [inline packages/medusa/src/api-v2/admin/api-keys/[id]/sales-channels/route.ts:22]: q: shouldn't this be part of the workflow?

riqwan (approved):
delicious cleanup!

packages/core-flows/src/api-key/steps/link-sales-channels-to-publishable-key.ts
sradevski: @riqwan agreed, we should revisit quite a few of the workflows that we can unify, but it's something for another time 😄

packages/medusa/src/api-v2/admin/api-keys/[id]/sales-channels/route.ts
Comment (sradevski) [reply packages/medusa/src/api-v2/admin/api-keys/[id]/sales-channels/route.ts:22]: @riqwan I just kept the existing logic, but I agree, it might be better if it's part of the workflow

PR #7117, opened 2024-04-22
feat(dashboard): metadata component

fPolic:
**What**
- add new metadata component

**Note**
- _example of usage on customer edit form_
- we are not handling update metadata case in the internal module service so for now delete case doesn't work properly

---

https://github.com/medusajs/medusa/assets/16856471/b588752d-9cf5-4d96-9cf8-760a764ab03e


kasperkristensen (approved):
LGTM

Maybe confirm with @ludvig18, but I think the intended behaviour is the same as what we have in the current admin. So when a user hovers a row, it should show a menu button, where from the user can delete the row, insert an empty row above and below.

Re inferring the type based on the input: I think its a cool idea, but also a bit dangerous as someone might write "true" and want the string and not boolean value. If we added the menu as we have in the existing component, we could potentially have a submenu called "Type", which has options based on what the value could be cast as. So if the user writes "true" the submenu would have the options "String" and "Boolean". In that case we would properly also need some kind of visual indication what the field is currently cast to.

fPolic:
Think we can merge this now since the API of the component won't change so we can start adding this to UIs. I discussed with Ludvig and we will add a context menu with (Delete, Add row below/above) instead of the current delete action but this isn't blocking.

PR #7118, opened 2024-04-22
feat(core-flows,medusa,types): add create shipment api for fulfillments

riqwan:
what:

- adds create shipment API for shipments - /admin/fulfillments/:id/shipment

A shipment is essentially a fulfillment that has "labels" and a "shipped_at" set.

RESOLVES CORE-1973

### DIFF
--- a/packages/core-flows/src/fulfillment/steps/update-fulfillment-workflow.ts
+++ b/packages/core-flows/src/fulfillment/steps/update-fulfillment-workflow.ts
@@ -9,9 +9,18 @@ export const updateFulfillmentWorkflowStep = createStep(
     data: FulfillmentWorkflow.UpdateFulfillmentWorkflowInput,
     { container }
   ) => {
-    const { transaction, result: updated } = await updateFulfillmentWorkflow(
-      container
-    ).run({ input: data })
+    const {
+      transaction,
+      result: updated,
+      errors,
+    } = await updateFulfillmentWorkflow(container).run({
+      input: data,
+      throwOnError: false,
+    })
+
+    if (Array.isArray(errors) && errors[0]) {
+      throw errors[0].error
+    }
 
     return new StepResponse(updated, transaction)
   },


packages/core-flows/src/fulfillment/steps/update-fulfillment-workflow.ts
```diff
@@ -0,0 +1,26 @@
...
+export const updateFulfillmentWorkflowStepId = "update-fulfillment-workflow"
+export const updateFulfillmentWorkflowStep = createStep(
+  updateFulfillmentWorkflowStepId,
+  async (
+    data: FulfillmentWorkflow.UpdateFulfillmentWorkflowInput,
+    { container }
+  ) => {
+    const { transaction, result: updated } = await updateFulfillmentWorkflow(
+      container
+    ).run({ input: data })
```
sradevski: throwOnError: false, and then throw errors[0].error if there are errors 😄

sradevski (commented):
Just that one comment on handling the workflow errors correctly, otherwise LGTM 👍

packages/medusa/src/api-v2/admin/fulfillments/validators.ts
```diff
@@ -1,44 +1,39 @@
...
+  tracking_number: z.string(),
+  tracking_url: z.string(),
+  label_url: z.string(),
+})
+
 export type AdminCancelFulfillmentType = z.infer<typeof AdminCancelFulfillment>
 export const AdminCancelFulfillment = z.object({})
 
 export type AdminCreateFulfillmentType = z.infer<typeof AdminCreateFulfillment>
 export const AdminCreateFulfillment = z.object({
```
olivermrbl, line 26: comment: This will need to accept `order_id` and then `order` should probably be changed to context or similar. Both are things we can add in follow up PRs, so only adding these to have them on record.

olivermrbl (approved):
Lgtm

packages/medusa/src/api-v2/admin/fulfillments/validators.ts
riqwan: created a ticket for it - [linear](https://linear.app/medusajs/issue/CORE-2010/fulfillment-module-api-accept-order-id-context)

PR #7122, opened 2024-04-22
fix: Standardize remaining batch methods and update FE calls

sradevski:


packages/core-flows/src/fulfillment/workflows/batch-shipping-option-rules.ts
```diff
@@ -0,0 +1,51 @@
...
+    }))
+
+    const deleteInput = transform({ input }, (data) => ({
+      ids: data.input.delete ?? [],
+    }))
+
+    // TODO: Currently we don't support edits, add support for this.
+    // We just call the steps directly here since there are no independent workflows, switch to CRUD workflows if they get added.
+    const [created, deleted] = parallelize(
+      createShippingOptionRulesStep(createInput),
```
riqwan, line 45: doesn't it work with just this?

```suggestion
      createShippingOptionRulesStep({ data: input.create ?? [] }),
```

riqwan (approved):
lgtm!

packages/core-flows/src/fulfillment/workflows/batch-shipping-option-rules.ts
sradevski: @riqwan I just kept the pattern the same with other places, but it might, I didn't try TBH 😄 let me try.

PR #7125, opened 2024-04-22
feat: V2 batch update products in categories

olivermrbl:


packages/types/src/workflow/product-category/index.ts
```diff
@@ -14,3 +15,8 @@ export interface UpdateProductCategoryWorkflowInput {
   id: string
   data: UpdateProductCategoryDTO
 }
+
+export interface BatchUpdateProductsOnCategoryWorkflowInput
+  extends LinkMethodRequest {
+  id: string
+}
```
riqwan: There's also `LinkWorkflowInput`

```suggestion
export interface BatchUpdateProductsOnCategoryWorkflowInput
  extends LinkWorkflowInput {}
```

packages/types/src/workflow/product-category/index.ts
```diff
@@ -1,3 +1,4 @@
+import { LinkMethodRequest } from "../../common"
```
riqwan:
```suggestion
import { LinkWorkflowInput } from "../../common"
```

packages/core-flows/src/product/steps/batch-link-products-in-category.ts
```diff
@@ -0,0 +1,91 @@
...
+      }
+    )
+
+    const productsWithUpdatedCategories = dbProducts.map((p) => {
+      if (toRemoveSet.has(p.id)) {
+        return {
+          id: p.id,
+          category_ids: (p.categories ?? [])
+            .filter((c) => c.id !== data.id)
+            .map((c) => c.id),
```
riqwan, line 36: q: should we just pass categories and skip the song and dance between module and workflows?

```suggestion
          categories: (p.categories ?? [])
            .filter((c) => c.id !== data.id)
```

riqwan (approved):
lgtm!

packages/types/src/workflow/product-category/index.ts
olivermrbl: Nice, will use that

packages/core-flows/src/product/steps/batch-link-products-in-category.ts
olivermrbl: Could make sense. We can even go as far as typing the `categories` input as `categories: { id: string }[]`, and then we wouldn't have to normalize anything at either level.

riqwan: I think this is how we do for some other modules as well

olivermrbl: Keeping it as is for now and creating a ticket

carlos-r-l-rodrigues (approved):

sradevski (approved):

PR #5984, opened 2024-01-02
feat(): sendgrid plugin typescript rewrite

luluhoc:
rewrite SendGrid plugin to typescript; bump SendGrid dep and rewrite to new medusa API

adds some functionality also as custom subject to orders

and some docs corrections `order_shipped_template` should be `order_shipment_created_template`

it has breaking changes in the plugin configuration i updated the docs too

PR #7127, opened 2024-04-23
docs: Fix order.shipment_created option name in sendgrid plugin

shahednasser:
Due to a bug in the SendGrid plugin, the `order.shipment_created` event uses different option names for localized/not localized options.

This PR adds a temporary fix in the docs until the bug is addressed.


olivermrbl:
Would it make sense to fix the bug instead?

shahednasser:
@olivermrbl I opened an issue for it on Linear, just didn't know how high priority it would be so opened this PR for a temporary solution

luluhoc:
I rewrote the plugin with this issue fixed no response whatsoever https://github.com/medusajs/medusa/pull/5984

PR #7130, opened 2024-04-23
feat(medusa): migrate store cart endpoints to zod

riqwan:
what:

- migrates all store cart endpoints to zod

RESOLVES CORE-2014


packages/medusa/src/api-v2/store/carts/helpers.ts
```diff
@@ -0,0 +1,22 @@
+import { MedusaContainer } from "@medusajs/types"
+import {
+  ContainerRegistrationKeys,
+  remoteQueryObjectFromString,
+} from "@medusajs/utils"
+
+export const refetchCart = async (
```
olivermrbl, line 7: comment: I have noticed these refetch-helpers do not accept filters. I think we should add that going forward, so that you can also pass params on a POST request. 

cc @sradevski

olivermrbl (approved):
LGTM

packages/medusa/src/api-v2/store/carts/helpers.ts
riqwan: This could also be a single generic entity:

```
export const refetchEntity = async (
  entryPoint: string,
  idOrFilter: string | object,
  scope: MedusaContainer,
  fields: string[]
) => {
  const remoteQuery = scope.resolve(ContainerRegistrationKeys.REMOTE_QUERY)
  const filters = isString(idOrFilter) ? { id: idOrFilter } : idOrFilter
    entryPoint,
    variables: { filters },
    fields,
  })

  const [entity] = await remoteQuery(queryObject)

  return entity
}
```

sradevski: @riqwan I completely agree, I was thinking about it but I didn't want to do it while doing the cleanup, as it will be pretty easy to refactor it. I think we can still keep eg. `refetchProduct` as a method, which instead calls `refetchEntity("product", ...args)`.

PR #7132, opened 2024-04-23
Fixed logged-in-user.mdx example

danvayn:
Previous documentation stuffed type imports with regular imports. This was fine, but either remove the unnecessary 'type' in the first import statement or import like this PR.


shahednasser (approved):
LGTM

PR #7133, opened 2024-04-23
feat(order): bundled actions

carlos-r-l-rodrigues:
what:
- fix order summary versioning
- calculate totals to calculate summary
- add method to revert last version
- bundled actions (fulfillment and shipment) more will come in other PR
- shipping method versioning like items

not included:
- summary fields will be refactored/normalized in other PR

Fixes CORE-2013 CORE-2015

### DIFF
--- a/integration-tests/modules/__tests__/order/order.spec.ts
+++ b/integration-tests/modules/__tests__/order/order.spec.ts
@@ -226,7 +226,7 @@ medusaIntegrationTestRunner({
                   promotion_id: expect.any(String),
                   code: "VIP_25 ETH",
                   raw_amount: {
-                    value: "5.0000000000000000000e-18",
+                    value: "5e-18",
                     precision: 20,
                   },
                   provider_id: expect.any(String),
@@ -237,11 +237,11 @@ medusaIntegrationTestRunner({
                   subtotal: 5e-18,
                   total: 5e-18,
                   raw_subtotal: {
-                    value: "5.0000000000000000000e-18",
+                    value: "5e-18",
                     precision: 20,
                   },
                   raw_total: {
-                    value: "5.0000000000000000000e-18",
+                    value: "5e-18",
                     precision: 20,
                   },
                 },
@@ -289,6 +289,7 @@ medusaIntegrationTestRunner({
                 metadata: null,
                 created_at: expect.any(String),
                 updated_at: expect.any(String),
+                deleted_at: null,
                 quantity: 1,
                 fulfilled_quantity: 0,
                 shipped_quantity: 0,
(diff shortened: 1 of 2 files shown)


packages/types/src/order/service.ts
```diff
@@ -1371,4 +1386,24 @@ export interface IOrderModuleService extends IModuleService {
     config?: RestoreReturn<TReturnableLinkableKeys>,
     sharedContext?: Context
   ): Promise<Record<TReturnableLinkableKeys, string[]> | void>
+
+  revertLastVersion(orderId: string, sharedContext?: Context): Promise<void>
+
+  restoreNextVersion(orderId: string, sharedContext?: Context): Promise<void>
```
olivermrbl: question: Do we need these to be part of the public interface?

I would have imagined that the methods we expose would be more action-specific and use these under the hood.

olivermrbl (approved):

packages/types/src/order/service.ts
carlos-r-l-rodrigues: I pushed a version fixing this. To me the revertLastVersion can be public but it doesn't make sense to restore that.
this will act as a soft delete of the latest changes applied

olivermrbl: Got it, perhaps we just need to think about how we present the method. `revertLastVersion` is not tied to anything commerce related, which is why I feel it's slightly off.

I think the shipping method versioning was included in that. Was that intentional?

carlos-r-l-rodrigues: it wasn't intentional, but I guess I fixed a couple of things related to this PR as well. Like adding `deleted_at` in places where they weren't present.
This module will need a cleanup after most of the functionalities are in place.

olivermrbl: Cool, I am fine to keep those changes in

packages/order/src/utils/transform-order.ts
```diff
@@ -29,6 +29,19 @@ export function formatOrder(
       }
     })
 
+    order.shipping_methods = order.shipping_methods?.map((shippingMethod) => {
+      const sm = { ...shippingMethod.shipping_method }
```
olivermrbl, line 33: question: ~~There is likely a good answer to this question, but I might have missed this part previously: why does the shipping method have a `shipping_method`~~?

olivermrbl: Nevermind, I realise it was introduced in this pr 😄

carlos-r-l-rodrigues: "shipping_methods" is now a relation table linking order_id + shipping_method + version.

this util is just to reshape the final response of the module, where you receive the shipping_method entity, not the "link" table.

PR #7134, opened 2024-04-23
feat: Add local file provider and wire everything up in the file module

sradevski:


packages/medusa/src/api-v2/admin/uploads/middlewares.ts
```diff
@@ -1,25 +1,36 @@
...
 import { authenticate } from "../../../utils/authenticate-middleware"
+import { validateAndTransformQuery } from "../../utils/validate-query"
+import { retrieveUploadConfig } from "./query-config"
+import { AdminGetUploadParams } from "./validators"
 
-const upload = multer({ dest: "uploads/" })
+// TODO: For now we keep the files in memory, as that's how they get passed to the workflows
+// This will need revisiting once we are closer to prod-ready v2, since with workflows and potentially
+// services on other machines using streams is not as simple as it used to be.
+const upload = multer({ storage: multer.memoryStorage() })
```
olivermrbl, line 11: question/thought: I guess we will need to somehow integrate this part of the flow with the file provider so that we can store the uploads on a remote server e.g. S3 if you are using the S3 plugin. Do you already have thoughts on this?

packages/file/src/module-definition.ts
```diff
@@ -1,6 +1,8 @@
 import { ModuleExports } from "@medusajs/types"
```
olivermrbl, line 1: comment: Think we should try to simplify this loader-file for v2 – a lot of boilerplate even for simple modules 😄

packages/utils/src/file/abstract-file-provider.ts
```diff
@@ -0,0 +1,24 @@
+import { FileTypes, IFileProvider } from "@medusajs/types"
+
+export class AbstractFileProviderService implements IFileProvider {
```
olivermrbl, line 3: question: Are we not missing the protected/private upload methods?

packages/file-local/src/services/local-file.ts
```diff
@@ -0,0 +1,100 @@
+import { FileTypes } from "@medusajs/types"
+import { AbstractFileProviderService, MedusaError } from "@medusajs/utils"
+import fs from "fs/promises"
+import path from "path"
+
+export class LocalFileService extends AbstractFileProviderService {
```
olivermrbl, line 6: See comment on abstract class.

olivermrbl (approved):

packages/utils/src/file/abstract-file-provider.ts
sradevski: @olivermrbl We are (I have a comment about it in the http routes), but I want to first implement the providers and see how the protected endpoint should look like. I feel like it should just be a config, not a separate endpoint,but I'll open that discussion once we have all the core pieces in place.

packages/file/src/module-definition.ts
sradevski: @olivermrbl agreed, I think there is quite a bit work left to do when it comes to improving the DX, but we can do that in one go for all modules

packages/medusa/src/api-v2/admin/uploads/middlewares.ts
sradevski: @olivermrbl So what happens currently is, the entire data/file the user sent will be kept in memory, passed through the workflow, to the file provider, and then it gets converted from in-memory binary content to a buffer and stored as a file.

If we decide that, for uploads, we want to avoid using workflows and call a module directly then we can probably support streaming more easily, as you wouldn't need any sort of serialization. That might be the way to go, but honestly for now it doesn't matter as much, as supporting that won't be a major change.

packages/file-local/src/services/local-file.ts
sradevski: Answered there 😄

packages/utils/src/file/abstract-file-provider.ts
olivermrbl: Got it, SGTM.

PR #7143, opened 2024-04-24
feat: Add s3-compatible plugin for the file module

sradevski:
I have run the tests against Minio, DO, and AWS and it's looking fine.


packages/file-s3/src/services/s3-file.ts
```diff
@@ -0,0 +1,154 @@
...
+
+    // TODO: Allow passing a full path for storage per request, not as a global config.
+    const fileKey = `${this.config_.prefix}${parsedFilename.name}-${ulid()}${
+      parsedFilename.ext
+    }`
+
+    const content = Buffer.from(file.content, "binary")
+    const command = new PutObjectCommand({
+      // TODO: Add support for private files
+      // We probably also want to support a separate bucket altogether for private files
```
olivermrbl, line 96: Yeah I think that sounds reasonable

olivermrbl (approved):
Very nice 💪

PR #7144, opened 2024-04-24
feat(medusa,types): added store apis for products

riqwan:
what:

- adds products list api
- adds products retrieve api

RESOLVES CORE-2016

### DIFF
--- a/.github/workflows/action.yml
+++ b/.github/workflows/action.yml
@@ -2,10 +2,10 @@ name: Medusa Pipeline
 on:
   push:
     branches:
-      - develop
+    - develop
   pull_request:
     branches:
-      - develop
+    - develop
 
 jobs:
   setup:
@@ -320,9 +320,6 @@ jobs:
         uses: ./.github/actions/cache-deps
         with:
           extension: pipeline
-
-      - name: Build
-        run: yarn build
 
       - name: Run Modules integration tests
         run: yarn test:integration:modules


integration-tests/modules/__tests__/product/store/index.spec.ts
```diff
@@ -0,0 +1,332 @@
+import { ModuleRegistrationName } from "@medusajs/modules-sdk"
```
sradevski, line 1: @riqwan I know you wrote a lot of tests here, but does it make sense to actually make the `integration-tests/api` tests work instead? If you feel like it will take a lot of time I'm fine with doing it later on, but we really need to try and clean up the tests

integration-tests/modules/__tests__/product/store/index.spec.ts
```diff
@@ -0,0 +1,332 @@
...
+const env = { MEDUSA_FF_MEDUSA_V2: true }
+const adminHeaders = { headers: { "x-medusa-access-token": "test_token" } }
+
+async function createProductsWithVariants(
+  productModule: IProductModuleService,
+  productsData: CreateProductDTO
+): Promise<[ProductDTO, ProductVariantDTO[]]> {
+  const { variants: variantsData, ...productData } = productsData
+
+  const [product] = await productModule.create([productData])
```
sradevski: Creating the product like this is problematic, as there are things that run in workflows that won't get executed. We should try and call the `admin` HTTP endpoint instead. There you will also just pass the `variants` as part of the product, no need for 2 separate calls

integration-tests/modules/__tests__/product/store/index.spec.ts
```diff
@@ -0,0 +1,332 @@
...
+          expect(response.data.count).toEqual(1)
+          expect(response.data.products).toEqual([
+            expect.objectContaining({
+              id: product2.id,
+            }),
+          ])
+        })
+
+        it("should list all products for a sales channel", async () => {
+          const salesChannel = await salesChannelModule.create({
```
sradevski: If you use the HTTP admin endpoints you won't need to do any of the remote linking, you'll simply create the salesChannel and then pass the sales_channel_id to the createProduct call

integration-tests/modules/__tests__/product/store/index.spec.ts
```diff
@@ -0,0 +1,332 @@
...
+          expect(error.response.status).toEqual(400)
+          expect(error.response.data).toEqual({
+            message:
+              "Pricing parameters (currency_code or region_id) are required to calculate prices",
+            type: "invalid_data",
+          })
+        })
+
+        it("should get product with prices when context is present", async () => {
+          await createVariantPriceSet({
```
sradevski: Same with these calls, if you use the HTTP endpoints you won't need to do this manually

packages/medusa/src/api-v2/store/products/helpers.ts
```diff
@@ -0,0 +1,97 @@
...
+  const priceSetCalculatedPriceMap: Record<
+    string,
+    PricingTypes.CalculatedPriceSet[]
+  > = {}
+
+  for (const link of links) {
+    variantPriceSetMap[link.variant_id] = link.price_set_id
+  }
+
+  const calculatedPrices = await pricingModule.calculatePrices(
```
sradevski: I think this is fine, but ideally we can plug this into remote query somehow, so we don't access modules directly in the HTTP layer. Alternatively, should we have this as a workflow? I think for now just leaving a todo comment to reevaluate is fine though

packages/medusa/src/api-v2/store/products/query-config.ts
```diff
@@ -0,0 +1,42 @@
...
+  "updated_at",
+  "*type",
+  "*collection",
+  "*options",
+  "*options.values",
+  "*tags",
+  "*images",
+  "*variants",
+  "*variants.options",
+  "*sales_channels",
```
sradevski: Is this useful for the storefront?

packages/medusa/src/api-v2/store/products/validators.ts
```diff
@@ -0,0 +1,49 @@
...
+
+export type StoreGetProductsParamsType = z.infer<typeof StoreGetProductsParams>
+export const StoreGetProductsParams = createFindParams({
+  offset: 0,
+  limit: 50,
+}).merge(
+  z
+    .object({
+      region_id: z.string().optional(),
+      currency_code: z.string().optional(),
```
sradevski, line 41: Why do we need this?

packages/medusa/src/api-v2/store/products/validators.ts
```diff
@@ -0,0 +1,49 @@
...
+      customer_id: z.string().optional(),
```
sradevski: What exactly does it mean to filter a product by customer id?

packages/medusa/src/api-v2/utils/common-validators/common.ts
```diff
@@ -24,3 +25,8 @@ export const BigNumberInput = z.union([
     precision: z.number(),
   }),
 ])
+
+export const OptionalBooleanValidator = z.preprocess(
```
sradevski, line 29: Nice 👍 Let's maybe update all references to it already?

packages/medusa/src/api-v2/utils/middlewares/pricing/set-pricing-context.ts
```diff
@@ -0,0 +1,88 @@
...
+import { NextFunction } from "express"
+import { MedusaRequest } from "../../../../types/routing"
+import { refetchEntities, refetchEntity } from "../../refetch-entity"
+
+export function setPricingContext() {
+  return async (req: MedusaRequest, _, next: NextFunction) => {
+    // If the endpoint doesn't request prices, we can exit early
+    if (
+      !req.remoteQueryConfig.fields.some((field) =>
+        field.startsWith("variants.prices")
```
sradevski: This check won't work if we fetch a variant directly, for example. Do we even need the check? Should we leave it to the one applying the middleware to decide on the check (eg. by passing a func)?

packages/medusa/src/api-v2/utils/middlewares/pricing/set-pricing-context.ts
```diff
@@ -0,0 +1,88 @@
+import { MedusaPricingContext } from "@medusajs/types"
+import { isPresent, MedusaError } from "@medusajs/utils"
+import { NextFunction } from "express"
+import { MedusaRequest } from "../../../../types/routing"
+import { refetchEntities, refetchEntity } from "../../refetch-entity"
+
+export function setPricingContext() {
```
sradevski: If this is specific to products, maybe we should move it there? There are a lot of assumptions on the req data made here that it doesn't seem like we can have it as a generic middleware?

riqwan: fair point, will move it there

packages/medusa/src/api-v2/utils/middlewares/pricing/set-pricing-context.ts
riqwan: Yup, this is specific to products, will change the name of the middleware

packages/medusa/src/api-v2/store/products/validators.ts
riqwan: these are not filters, they're parameters for pricing context to calculate prices

packages/medusa/src/api-v2/store/products/validators.ts
riqwan: They're parameters for pricing context to calculate prices

packages/medusa/src/api-v2/store/products/query-config.ts
riqwan: Will double check, but I don't think its necessary. Good catch!

integration-tests/modules/__tests__/product/store/index.spec.ts
riqwan: will give it a shot 💪🏻

integration-tests/modules/__tests__/product/store/index.spec.ts
riqwan: 2 points:

- I'd like these tests to be run on the CI, which it won't with the api tests.
- I don't get why we do that. Wouldn't it just be better to move over the test cases to modules and make it work for v2? We're not really writing more tests for v1, so we wouldn't be missing out much. We do the same everywhere else - v2 routes, v2 dashboard, why not the same here?

packages/medusa/src/api-v2/store/products/validators.ts
sradevski: @riqwan It feels a bit strange for the user to have to provide all of this as part of the query params, but I don't know how else to do it. 

I think for the region_id I get it, but eg. the customer_id should be retrieved from the request token/context, otherwise I can get prices for another customer if I know their ID.

Another point, should we also check the publishable key for this info?

packages/medusa/src/api-v2/store/products/validators.ts
sradevski: Let's discuss in the other comment thread

integration-tests/modules/__tests__/product/store/index.spec.ts
sradevski: @riqwan The idea is that, unless we intended a breaking change, the test scenarios in `api` are still valid and should work regardless if it is v1 or v2. Eg. for the product endpoints, we have very few breaking changes, and that gives us some confidence that we are compliant with v1, except for the intentional breaking changes.

But if you feel like that is not a valid point, then we should discuss and settle on a single approach, as currently we have a split between the two

integration-tests/modules/__tests__/product/store/index.spec.ts
```diff
@@ -0,0 +1,332 @@
...
+      let appContainer
+      let product
+      let product2
+      let product3
+      let product4
+      let variant
+      let variant2
+      let variant3
+      let variant4
+      let pricingModule: IPricingModuleService
```
sradevski: Ideally we should not directly access the modules from the e2e tests, we should treat them as a black box and rely on the HTTP endpoints. The only exception is if there is something that isn't accessible through the HTTP endpoints and we really need to validate it (eg. due to some background task), but that should be rare

packages/medusa/src/api-v2/store/products/validators.ts
riqwan: Same. Thats a good point about customer_id, I'll use a middleware to pick it up from the token. 

I've added a todo comment there for publishable keys, I don't have the answer for it right now. Do you know if the concept of a default sales channel is being carried over to v2?

olivermrbl: > Do you know if the concept of a default sales channel is being carried over to v2?

It will be carried over. The store holds a default sales channel ID, similar to v1

packages/medusa/src/api-v2/store/products/helpers.ts
carlos-r-l-rodrigues: I also believe we should have another property to use the remote query.
It can't be `prices` because the object format of calculated prices is different, but we could have a property `calculated_prices` that receives `{context: pricingContext}` as a variable
We could completely eliminate this helper `wrapProductsWithPrices`

riqwan: That sounds good! is it possible to pass deep filters to pricing if product is the entry point to the query?

carlos-r-l-rodrigues: It is, every property that is an entity can have arguments/variables.
however the function `remoteQueryObjectFromString` will have to be adapted to add `__args` in these expansions, not only at the entry point level as it is currently doing.

riqwan: Nice, will take a look. I had this need in a few other places too, so this will be great

olivermrbl: Should we merge @carlos-r-l-rodrigues's PR and integrate the changes in this PR before we proceed?

riqwan: definitely, can kill a lot of the boiler plate in price list and products

packages/medusa/src/api-v2/utils/common-validators/common.ts
riqwan: This file already has too many changes, will do it as a follow up 👍🏻

packages/medusa/src/api-v2/store/products/[id]/route.ts
```diff
@@ -0,0 +1,26 @@
+import { isPresent } from "@medusajs/utils"
+import { MedusaRequest, MedusaResponse } from "../../../../types/routing"
+import { refetchProduct } from "../helpers"
+import { StoreGetProductsParamsType } from "../validators"
+
+export const GET = async (
+  req: MedusaRequest<StoreGetProductsParamsType>,
+  res: MedusaResponse
+) => {
+  const context = isPresent(req.pricingContext)
```
Comment (sradevski) [inline packages/medusa/src/api-v2/store/products/[id]/route.ts:10]: Nothing to do now, I am just thinking if we can make the context-passing and calculations setup a bit nicer. Just talking out loud here 😄

packages/medusa/src/api-v2/store/products/middlewares.ts
```diff
@@ -0,0 +1,67 @@
...
+  StoreGetProductsParams,
+  StoreGetProductsParamsType,
+} from "./validators"
+
+export const storeProductRoutesMiddlewares: MiddlewareRoute[] = [
+  {
+    method: "ALL",
+    matcher: "/store/products*",
+    middlewares: [
+      authenticate("store", ["session", "bearer"], {
```
sradevski, line 22: Again, for another time - I think having a global middleware that just populates the context will be much better than relying on `authenticate`, as this middleware ideally should do actual authentication (and the `allowUnauthenticated` flag should not exist).

sradevski (approved):
Looks good!

packages/medusa/src/api-v2/store/products/[id]/route.ts
```diff
@@ -0,0 +1,26 @@
...
+  const product = await refetchProduct(
+    {
+      id: req.params.id,
+      context,
+    },
+    req.scope,
+    req.remoteQueryConfig.fields
+  )
+
+  res.json({ product })
```
Comment (adrien2p) [inline packages/medusa/src/api-v2/store/products/[id]/route.ts:25]:
```suggestion
  const refetchProductParams: object = {
    id: req.params.id,
  }
  
  if (isPresent(req.pricingContext)) {
    config.context = {
      "variants.calculated_price": { context: req.pricingContext },
    }
  }
   
  const product = await refetchProduct(
    refetchProductParams,
    req.scope,
    req.remoteQueryConfig.fields
  )

  res.json({ product })
```

packages/medusa/src/api-v2/store/products/helpers.ts
```diff
@@ -0,0 +1,34 @@
...
+import { StoreGetProductsParamsType } from "./validators"
+
+// For category filters, we only allow showcasing public and active categories
+// TODO: This should ideally be done in the middleware, write a generic filter to conditionally
+// map these values or normalize the filters to the ones expected by remote query
+export function wrapWithCategoryFilters(filters: StoreGetProductsParamsType) {
+  const categoriesFilter = isPresent(filters.category_id)
+    ? {
+        categories: {
+          ...filters.category_id,
```
adrien2p, line 13: suggestion: I believe the spread here will create an issue, it will assign the index as the key and the id as the value as it seems to be an array spreaded into an object

packages/medusa/src/api-v2/store/products/query-config.ts
```diff
@@ -0,0 +1,41 @@
...
+  "*options",
+  "*options.values",
+  "*tags",
+  "*images",
+  "*variants",
+  "*variants.options",
+]
+
+export const retrieveProductQueryConfig = {
+  defaults: defaultStoreProductFields,
```
adrien2p, line 33: q: do we need to limit what is accessible by the store, I am thinking if someone add more link or properties or something, the store could request for some admin data? This is more of a global question :)

packages/medusa/src/api-v2/store/products/[id]/route.ts
Comment (adrien2p) [reply packages/medusa/src/api-v2/store/products/[id]/route.ts:25]: wdyt? It is only a suggestion

packages/medusa/src/api-v2/utils/middlewares/products/set-pricing-context.ts
```diff
@@ -0,0 +1,89 @@
...
+    if (
+      !req.remoteQueryConfig.fields.some((field) =>
+        field.startsWith("variants.calculated_price")
+      )
+    ) {
+      delete req.filterableFields.region_id
+      delete req.filterableFields.currency_code
+
+      return next()
+    }
```
adrien2p, line 19: Q: here, If I pass the region id and/or currency but I forget about the relation, I will still get a result but without the context filtering, it can make it hard to debug, what if we throw an invalid data here instead to say that it is not possible to filter the with the context without this relation, or alternatively we infer the context filtering if region id or currency code is present since they are only used in that context.

wdyt?

packages/medusa/src/api-v2/utils/middlewares/products/set-pricing-context.ts
```diff
@@ -0,0 +1,89 @@
...
+
+    // If a currency code is explicitly passed, we should be using that instead of the
+    // regions currency code
+    if (query.currency_code) {
+      const currency = await refetchEntity(
+        "currency",
+        { code: query.currency_code },
+        req.scope,
+        ["code"]
```
adrien2p, line 49: Q: isn't the code here the same as the one provided? Are we doing that to check if the entity exists?

packages/medusa/src/api-v2/utils/middlewares/products/set-pricing-context.ts
```diff
@@ -0,0 +1,89 @@
...
+        ["id"]
+      )
+
+      pricingContext.customer_group_id = customerGroups.map((cg) => cg.id)
+
+      delete req.filterableFields.customer_id
+    }
+
+    // If a region or currency_code is not present in the context, we will not be able to calculate prices
+    if (!isPresent(pricingContext)) {
```
adrien2p, line 74: suggestion: shouldn't we check for the currency and or region? Could it be possible that none of them are present but another key is? I think a check on region or currency would be more future proof. Wdyt?

PR #7149, opened 2024-04-25
feat: Add support for search to all endpoints

sradevski:


olivermrbl (approved):
Nice, love how simple it was to add this 💪 

cc @adrien2p

PR #7150, opened 2024-04-25
feat(core-flows, medusa): add shipping methods to cart API

riqwan:
what:

- adds API to add shipping methods to cart
- refreshes shipping methods on cart mutations

RESOLVES CORE-2007

### DIFF
--- a/integration-tests/modules/__tests__/cart/store/cart.workflows.spec.ts
+++ b/integration-tests/modules/__tests__/cart/store/cart.workflows.spec.ts
@@ -1529,8 +1529,7 @@ medusaIntegrationTestRunner({
           expect(errors).toEqual([
             expect.objectContaining({
               error: expect.objectContaining({
-                message:
-                  "Shipping Options (Test shipping option) are invalid for cart.",
+                message: `Shipping Options are invalid for cart.`,
                 type: "invalid_data",
               }),
             }),
@@ -1552,7 +1551,7 @@ medusaIntegrationTestRunner({
           expect(errors).toEqual([
             expect.objectContaining({
               error: expect.objectContaining({
-                message: "Shipping Options (does-not-exist) not found",
+                message: "Shipping Options are invalid for cart.",
                 type: "invalid_data",
               }),
             }),
--- a/packages/core-flows/src/definition/cart/steps/refresh-cart-shipping-methods.ts
+++ b/packages/core-flows/src/definition/cart/steps/refresh-cart-shipping-methods.ts
@@ -16,7 +16,7 @@ export const refreshCartShippingMethodsStep = createStep(
     const { shipping_methods: shippingMethods = [] } = cart
 
     if (!shippingMethods?.length) {
-      return new StepResponse(void 0)
+      return new StepResponse(void 0, [])
     }
 
     const fulfillmentModule = container.resolve<IFulfillmentModuleService>(
@@ -27,25 +27,9 @@ export const refreshCartShippingMethodsStep = createStep(
       ModuleRegistrationName.CART
     )
 
-    const shippingMethodsToDelete: string[] = []
     const shippingOptionIds: string[] = shippingMethods.map(
       (sm) => sm.shipping_option_id!
     )
-
-    const shippingOptions = await fulfillmentModule.listShippingOptions(
-      { id: shippingOptionIds },
-      { select: ["id", "name"], take: null }
-    )
-
-    const diff = arrayDifference<string>(
-      shippingOptionIds,
-      shippingOptions.map((o) => o.id)
-    )
-
-    // If any shipping methods are found without shipping_option_id, delete them
-    if (diff.length) {
-      shippingMethodsToDelete.push(...diff)
-    }
 
     const validShippingOptions =
       await fulfillmentModule.listShippingOptionsForContext(
@@ -62,18 +46,17 @@ export const refreshCartShippingMethodsStep = createStep(
         { relations: ["rules"] }
       )
 
-    const invalidOptionIds = arrayDifference(
+    const validShippingOptionIds = validShippingOptions.map((o) => o.id)
+    const invalidShippingOptionIds = arrayDifference(
       shippingOptionIds,
-      validShippingOptions.map((o) => o.id)
+      validShippingOptionIds
     )
 
-    if (invalidOptionIds.length) {
-      shippingMethodsToDelete.push(...invalidOptionIds)
-    }
+    const shippingMethodsToDelete = shippingMethods
+      .filter((sm) => invalidShippingOptionIds.includes(sm.shipping_option_id!))
+      .map((sm) => sm.id)
 
-    if (shippingMethodsToDelete.length) {
-      await cartModule.softDeleteShippingMethods(shippingMethodsToDelete)
-    }
+    await cartModule.softDeleteShippingMethods(shippingMethodsToDelete)
 
     return new StepResponse(void 0, shippingMethodsToDelete)
   },
--- a/packages/core-flows/src/definition/cart/steps/validate-cart-shipping-options.ts
+++ b/packages/core-flows/src/definition/cart/steps/validate-cart-shipping-options.ts
@@ -24,23 +24,6 @@ export const validateCartShippingOptionsStep = createStep(
       ModuleRegistrationName.FULFILLMENT
     )
 
-    const shippingOptions = await fulfillmentModule.listShippingOptions(
-      { id: optionIds },
-      { select: ["id", "name"], take: null }
-    )
-
-    const diff = arrayDifference(
-      optionIds,
-      shippingOptions.map((o) => o.id)
-    )
-
-    if (diff.length) {
-      throw new MedusaError(
-        MedusaError.Types.INVALID_DATA,
-        `Shipping Options (${diff.join(", ")}) not found`
-      )
-    }
-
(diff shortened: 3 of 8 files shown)


integration-tests/modules/__tests__/cart/store/cart.workflows.spec.ts
```diff
@@ -1397,34 +1397,46 @@ medusaIntegrationTestRunner({
...
+          shippingProfile = await fulfillmentModule.createShippingProfiles({
+            name: "Test",
+            type: "default",
+          })
 
-          const fulfillmentSet = await fulfillmentModule.create({
+          // TODO: Setting up the geozones here as US, but testing with european countries
+          // in the tests doesn't throw an validations in the module. The context needs to have
+          // some required attributes to test against service zones
+          fulfillmentSet = await fulfillmentModule.create({
```
riqwan, line 1421: @olivermrbl wdyt?

olivermrbl: I am not sure I understand. Can I get you to elaborate?

riqwan: nvm, dug a little deeper and its already built in 🎉  - [commit](https://github.com/medusajs/medusa/pull/7150/commits/0f925a76793943fbb8a6cf44d336aeca7ffbd8ed)

packages/core-flows/src/definition/cart/workflows/add-shipping-method-to-cart.ts
```diff
@@ -24,22 +26,38 @@ export const addShippingMethodToWorkflow = createWorkflow(
...
+      ],
+      variables: { id: input.cart_id },
+      list: false,
+    })
+
     const optionIds = transform({ input }, (data) => {
       return (data.input.options ?? []).map((i) => i.id)
     })
 
+    validateCartShippingOptionsStep({
```
olivermrbl, line 40: question: I wonder if we should just remove the method, if it's no longer applicable? We've talked about doing that previously. 

That would be a slightly better DX than telling users to handle this themselves in the client.

olivermrbl: Maybe it could be a config passed to the step?

olivermrbl, line 40: We need to add this validation/removal to all other cart updates to ensure that you cannot checkout with an invalid option. E.g. a customer used to have free shipping, but no longer has because they removed items from their cart.

olivermrbl (approved):
Nice work! A couple of comments.

packages/core-flows/src/definition/cart/workflows/add-shipping-method-to-cart.ts
riqwan: yes, this is coming in the next PR. I wanted to get this in separately.

riqwan: I created a new step called `refreshCartShippingMethodsStep` to do this, feels cleaner that way as the inputs are slightly different:

```
refreshCartShippingMethodsStep({ cart }) - removes shipping methods if no longer applicable
validateCartShippingOptionsStep({ cart, option_ids }) - throws error
```

should we merge them?

olivermrbl: I am fine to keep the two separated. In that case, I would probably unify the validation logic in a helper we can use in both methods, as this is almost identical right?

Update: I looked at the commit, and it looks like there are slight differences, so I am not sure if this is possible. I'll let you decide on how to proceed.

packages/core-flows/src/definition/cart/steps/refresh-cart-shipping-methods.ts
```diff
@@ -0,0 +1,89 @@
...
+    const shippingOptionIds: string[] = shippingMethods.map(
+      (sm) => sm.shipping_option_id!
+    )
+
+    const shippingOptions = await fulfillmentModule.listShippingOptions(
+      { id: shippingOptionIds },
+      { select: ["id", "name"], take: null }
+    )
+
+    const diff = arrayDifference<string>(
```
sradevski: Let's just make sure the difference is the methods that were in the cart but not in the DB, not the way around (it might be correct, just wanted to mention it as something that can easily fall through the cracks)

packages/core-flows/src/definition/cart/steps/refresh-cart-shipping-methods.ts
```diff
@@ -0,0 +1,89 @@
...
+      shippingOptionIds,
+      shippingOptions.map((o) => o.id)
+    )
+
+    // If any shipping methods are found without shipping_option_id, delete them
+    if (diff.length) {
+      shippingMethodsToDelete.push(...diff)
+    }
+
+    const validShippingOptions =
```
sradevski, line 34: I am a bit confused by this workflow as it lists shipping options twice and deletes, maybe we can make the code more obvious why we are doing this?

packages/core-flows/src/definition/cart/workflows/add-to-cart.ts
```diff
@@ -22,6 +23,19 @@ import { prepareConfirmInventoryInput } from "../utils/prepare-confirm-inventory
 import { prepareLineItemData } from "../utils/prepare-line-item-data"
 import { refreshPaymentCollectionForCartStep } from "./refresh-payment-collection"
 
+const cartFields = [
```
sradevski: This is a lot of data, do we really need everything? I'm afraid this might have a toll on performance.

packages/core-flows/src/definition/cart/steps/validate-cart-shipping-options.ts
```diff
@@ -0,0 +1,78 @@
...
+import { arrayDifference, MedusaError } from "@medusajs/utils"
+import { createStep, StepResponse } from "@medusajs/workflows-sdk"
+import { IFulfillmentModuleService } from "../../../../../types/dist/fulfillment/service"
+
+interface StepInput {
+  cart: CartDTO
+  option_ids: string[]
+}
+
+export const validateCartShippingOptionsStepId =
```
sradevski, line 12: Validating and refetching seems like it has a very similar logic, any chance we can somehow refactor the two?

packages/core-flows/src/definition/cart/workflows/add-shipping-method-to-cart.ts
sradevski: I had a similar concern, maybe `refresh` can just return if a refresh was performed or not, and that can serve for the cases when we do want to validate? Just something that seemed confusing, but if you can't think of anything better than this is fine I guess

riqwan: The inputs are different and the outcome is different. They serve 2 different purposes even though they're doing similar things. If you both feel strongly enough, I'll merge them in.

packages/core-flows/src/definition/cart/steps/refresh-cart-shipping-methods.ts
riqwan: I think this might be outdated, take a look again 🤞🏻

packages/core-flows/src/definition/cart/workflows/add-to-cart.ts
riqwan: These actions are performed on a cart context, depending on what the rules are, they could be any of these. We can optimize this by whitelisting what we allow as "rules", but that can come later.

packages/core-flows/src/definition/cart/steps/refresh-cart-shipping-methods.ts
```diff
@@ -0,0 +1,72 @@
...
+    const invalidShippingOptionIds = arrayDifference(
+      shippingOptionIds,
+      validShippingOptionIds
+    )
+
+    const shippingMethodsToDelete = shippingMethods
+      .filter((sm) => invalidShippingOptionIds.includes(sm.shipping_option_id!))
+      .map((sm) => sm.id)
+
+    await cartModule.softDeleteShippingMethods(shippingMethodsToDelete)
```
sradevski, line 59: Question: Do we want to do a soft or hard delete in this case, if a shipping method doesn't exist anymore?

sradevski (approved):

packages/core-flows/src/definition/cart/steps/refresh-cart-shipping-methods.ts
sradevski: @riqwan yup it was outdated, looks good 👍

packages/core-flows/src/definition/cart/steps/refresh-cart-shipping-methods.ts
riqwan: We can only soft delete here as we have to revert this if any other steps fail

packages/core-flows/src/definition/cart/workflows/add-shipping-method-to-cart.ts
olivermrbl: I am fine keeping it as is, so we can get it merged.

PR #7152, opened 2024-04-25
fix: Corrections to the store region endpoint for v2

sradevski:


packages/medusa/src/api-v2/store/regions/query-config.ts
```diff
@@ -6,22 +6,15 @@ export const defaultStoreRegionFields = [
   "updated_at",
   "deleted_at",
   "metadata",
-  "countries.id",
```
olivermrbl, line 9: thought: Think we still need the countries in the storefront

olivermrbl (approved):
Otherwise, LGTM

packages/medusa/src/api-v2/store/regions/query-config.ts
sradevski: @olivermrbl I was just thinking it might be better to not fetch it by default, WDYT? I can add it back for now and we can review what is and isn't fetched by default for all endpoints later on I guess

PR #7153, opened 2024-04-25
feat(modules-sdk): remote query context filter

carlos-r-l-rodrigues:
What:

- Remote query now handles `context` keywork in the arguments to forward it as a filter to the method `list`
- Pricing module `list` method returning `calculated_price` if requested as a field

### DIFF
--- a/packages/pricing/src/services/pricing-module.ts
+++ b/packages/pricing/src/services/pricing-module.ts
@@ -183,14 +183,22 @@ export default class PricingModuleService<
     const priceSets = await super.list(filters, config, sharedContext)
 
     if (pricingContext && priceSets.length) {
+      const priceSetIds: string[] = []
+      const priceSetMap = new Map()
+      for (const priceSet of priceSets) {
+        priceSetIds.push(priceSet.id)
+        priceSetMap.set(priceSet.id, priceSet)
+      }
+
       const calculatedPrices = await this.calculatePrices(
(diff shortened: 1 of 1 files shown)


olivermrbl (approved):
I think this is very nice

packages/pricing/src/services/pricing-module.ts
```diff
@@ -152,6 +153,79 @@ export default class PricingModuleService<
...
+    const priceSets = await super.list(filters, config, sharedContext)
+
+    if (pricingContext && priceSets.length) {
+      const calculatedPrices = await this.calculatePrices(
+        { id: priceSets.map((p) => p.id) },
+        { context: pricingContext },
+        sharedContext
+      )
+
+      for (let idx = 0; idx < priceSets.length; idx++) {
```
riqwan: we stopped using indexes to set because it was causing flakiness in the CI, might be better to create a map before and assign from that.

riqwan (approved):
Carlos, this is beautiful! 🔥

packages/pricing/src/services/pricing-module.ts
carlos-r-l-rodrigues: In this case it is safe to do that because the method `calculatePrices` is returning the a `.map` of the input.
But in any case, I changed it to the map to keep the same pattern we use in other places.

PR #7160, opened 2024-04-26
feat: Add missing endpoints and normalize customer and currency endpoints for storefront

sradevski:


packages/medusa/src/api-v2/store/currencies/validators.ts
```diff
@@ -1,30 +1,19 @@
...
-  @Type(() => StoreGetCurrenciesParams)
-  $and?: StoreGetCurrenciesParams[]
+export const StoreGetCurrencyParams = createSelectParams()
 
-  @IsOptional()
-  @ValidateNested({ each: true })
-  @Type(() => StoreGetCurrenciesParams)
-  $or?: StoreGetCurrenciesParams[]
-}
+export type StoreGetCurrenciesParamsType = z.infer<
```
riqwan, line 6: q: is this type getting inferred correctly for you? I'm getting any

packages/medusa/src/api-v2/store/customers/me/addresses/route.ts
```diff
@@ -59,5 +63,11 @@ export const POST = async (
...
   }
 
-  res.status(200).json({ address: result[0] })
+  const customer = await refetchCustomer(
+    customerId,
+    req.scope,
+    req.remoteQueryConfig.fields
+  )
+
+  res.status(200).json({ customer })
```
riqwan, line 72: shouldn't this return address? Feels weird to be returning the customer here

packages/medusa/src/api-v2/store/customers/middlewares.ts
```diff
@@ -1,62 +1,108 @@
...
+  {
+    method: ["GET"],
+    matcher: "/store/customers/me/addresses/:address_id",
+    middlewares: [
+      validateAndTransformQuery(
+        StoreGetCustomerAddressParams,
+        QueryConfig.retrieveAddressTransformQueryConfig
+      ),
+    ],
   },
```
riqwan, line 87: the returns of these 2 should be consistent imo

packages/medusa/src/api-v2/store/customers/validators.ts
```diff
@@ -1,244 +1,60 @@
...
+>
+export type StoreGetCustomerAddressesParamsType = z.infer<
+  typeof StoreCreateCustomerAddress
+>
+export type StoreCreateCustomerAddressType = z.infer<
+  typeof StoreCreateCustomerAddress
+>
+export type StoreUpdateCustomerAddressType = z.infer<
+  typeof StoreUpdateCustomerAddress
+>
```
riqwan, line 60: q: didn't we agree on setting these close to their values? Did that change?

riqwan (approved):
some comments to be addressed, but overall lgtm!

packages/medusa/src/api-v2/store/currencies/validators.ts
sradevski: @riqwan it seems it doesn't, nice catch. It is a general issue though it seems, not specific to the currencies.

packages/medusa/src/api-v2/store/customers/me/addresses/route.ts
sradevski: @riqwan based on our conventions we respond with the parent object, this just follows the same pattern (you can get the addresses from the `customer`). If you feel like it's strange can you raise it in `eng-core` and let's discuss it there?

packages/medusa/src/api-v2/store/customers/middlewares.ts
sradevski: @riqwan same comment as above, but this is what we decided when you do a POST on a child entity

packages/medusa/src/api-v2/store/customers/validators.ts
sradevski: @riqwan since you raised the point of the order in which things are defined I put them at the end here, and we can probably do it in other places as necessary. I don't think it matters that much TBH

sradevski:
@riqwan I'm gonna merge as for the time being the comments are not tied to this PR, but let's please discuss the pattern of returning the parent object on POST

PR #7169, opened 2024-04-29
feat(fulfillment): shipping options context field

carlos-r-l-rodrigues:


packages/fulfillment/src/services/fulfillment-module-service.ts
```diff
@@ -137,6 +137,49 @@ export default class FulfillmentModuleService<
...
+      filters.context = optionsContext
+
+      return await this.listShippingOptionsForContext(
+        filters,
+        config,
+        sharedContext
+      )
+    }
+
+    return await super.listShippingOptions(filters, config, sharedContext)
```
adrien2p, line 182: q: if the `shipping_options_context` is not present, but the context is present in the filters, would it break the list here?

packages/fulfillment/src/services/fulfillment-module-service.ts
```diff
@@ -137,6 +137,49 @@ export default class FulfillmentModuleService<
...
+      return
+    }
+
+    const shippingOptionsContext = filters.context ?? {}
+
+    // cleanup virtual field "shipping_options_context"
+    config.relations?.splice(fieldIdx, 1)
+    delete filters.context
+
+    return shippingOptionsContext
```
adrien2p:
```suggestion
    const fieldIdx = config.relations?.indexOf("shipping_options_context")
    const shouldListWithContext = fieldIdx > -1
    const shippingOptionsContext = filters.context ?? {}

    delete filters.context

    if (!shouldListWithContext) {
      return
    }

    // cleanup virtual field "shipping_options_context"
    config.relations?.splice(fieldIdx, 1)

    return shippingOptionsContext
```

adrien2p: I moved the delete above just to be sure that we do not end up calling the simple list with a context if the virtual relation is not present, let me know wyt

riqwan (approved):
nice!

PR #7175, opened 2024-04-30
feat(medusa, core-flows, types): Allow to update the rules from a shipping options

adrien2p:
**What**
Add support for the following operations
- update rules from the update shipping options end point
- update rules from the batch update end point

Also added some improvements, that can be revisited later
- Add a rule value normalizer, jsonb will transform the input value to a [primitive](https://www.postgresql.org/docs/current/datatype-json.html#JSON-TYPE-MAPPING-TABLE) when possible meaning that passing `"true"` will result in storing `true` and not the string. The normalizer takes care of that

### DIFF
--- a/.changeset/nervous-glasses-exercise.md
+++ b/.changeset/nervous-glasses-exercise.md
@@ -1,0 +1,5 @@
+---
+"@medusajs/medusa": patch
+---
+
+feat(medusa): add middleware filters + scope products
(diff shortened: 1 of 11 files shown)


fPolic (approved):
LGTM 🙏

adrien2p:
@fPolic should I merge?

sradevski (approved):

fPolic:
> @fPolic should I merge?

Yes please 🚢

PR #7178, opened 2024-04-30
feat(medusa): add middleware filters + scope products

riqwan:


integration-tests/modules/__tests__/product/store/index.spec.ts
```diff
@@ -349,6 +394,34 @@ medusaIntegrationTestRunner({
           )
         })
 
+        // TODO: There are 2 problems that need to be solved to enable this test
+        // 1. When adding product to another category, the product is being removed from earlier assigned categories
+        // 2. MikroORM seems to be doing a join strategy to load relationships, we need to send a separate query to fetch relationships
```
adrien2p, line 399: info: for the second point you can pass the strategy as part of the options. But it would require to add it as part of the FindConfig type, the underlying service and repository will then consume it

packages/medusa/src/api-v2/utils/middlewares/products/set-pricing-context.ts
```diff
@@ -70,8 +86,8 @@ export function setPricingContext() {
       delete req.filterableFields.customer_id
     }
 
-    // If a region or currency_code is not present in the context, we will not be able to calculate prices
-    if (!isPresent(pricingContext)) {
+    // If a currency_code is not present in the context, we will not be able to calculate prices
+    if (!isPresent(pricingContext.currency_code)) {
```
adrien2p, line 90: todo: based on the error message it seems that we should add a check on the region id, probably something along the line of `!pricingContext.currency_code && !pricingContext.region_id` then we throw

adrien2p (approved):
LGTM 💪 couple of comments

integration-tests/modules/__tests__/product/store/index.spec.ts
```diff
@@ -349,6 +394,34 @@ medusaIntegrationTestRunner({
           )
         })
 
+        // TODO: There are 2 problems that need to be solved to enable this test
+        // 1. When adding product to another category, the product is being removed from earlier assigned categories
```
riqwan, line 398: @sradevski do you know anything about this?

integration-tests/modules/__tests__/product/store/index.spec.ts
riqwan: yup, will fix this as a follow up.

packages/medusa/src/api-v2/utils/middlewares/products/set-pricing-context.ts
riqwan: the currency code will be set as a part of the region, so we only need to check currency_code here.

PR #7181, opened 2024-04-30
chore(orchestration): replace workflow definition in v1

carlos-r-l-rodrigues:


sradevski (approved):

packages/orchestration/src/workflow/workflow-manager.ts
```diff
@@ -101,9 +101,11 @@ export class WorkflowManager {
         : true
 
       if (!areStepsEqual) {
-        throw new Error(
-          `Workflow with id "${workflowId}" and step definition already exists.`
-        )
+        if (process.env.MEDUSA_FF_MEDUSA_V2 == "true") {
```
olivermrbl, line 104: nit: I believe this does not work as expected if the feature flag is set in the project config instead of the environment variables:
```
featureFlags: {
  medusa_v2: true
}
```

adrien2p: The workflow manager doesn't have access to the feature flag. Maybe when loading them we could set the env at the same time since it is define in the config? If the env is provided then we keep it as it is, if the property is provided then we set the env wdyt?

olivermrbl: Yeah, I think that make sense. So essentially making sure that whatever is set in the `featureFlags` config is also set in environment variables.

PR #7182, opened 2024-04-30
fix: Ensure passed Id filters get applied on top of maybe link filter

sradevski:


packages/utils/src/common/array-intersection.ts
```diff
@@ -0,0 +1,17 @@
...
+  const firstArraySet = new Set(firstArray)
+  const res = new Set<TElement>()
+
+  secondArray.forEach((element) => {
+    if (firstArraySet.has(element)) {
+      res.add(element)
+    }
+  })
+
+  return Array.from(res)
```
riqwan, line 16: I think you can get a union through spreading 2 sets

```suggestion
  const firstArraySet = new Set(firstArray)
  const secondArraySet = new Set(secondArray)
  
  return Array.from(new Set([...firstArraySet, ...secondArraySet]););
```

integration-tests/modules/__tests__/product/store/index.spec.ts
```diff
@@ -247,6 +247,18 @@ medusaIntegrationTestRunner({
             ])
           })
 
+          it("should list products by id", async () => {
```
riqwan, line 250: actually, now that I think about it, the union extension you added might not work.

If you have `/store/products?id[]=product-not-in-sales-channel&sales_channel_id[]=sc-id`, this should return an empty products response. I'm not sure that will be the case with a union

sradevski: @riqwan what I implemented is an intersection, not a union, so it should behave exactly as you said in the example, the ID has to be both in the id filters and in the sales channels.

sradevski: @riqwan merging as from what I can tell it works as expected, let me know if you found some other issue with the current approach.

PR #7183, opened 2024-04-30
feat(workflows-sdk): Execute workflows as step in other workflows

adrien2p:
**What**
- introduce `runAsStep` to the workflows so that they can be executed smoothly into other workflows
- unify the workflows run API such that it is not necessary to initiate a workflow before calling the run (backward compatible)

for the unified API:

**Run the workflow with the workflow engine**
```
const { acknowledgement, errors, result } = await wfe.run("workflow-id", {
    container: req.scope,
    context: { requestId: req.requestId },
    throwOnError: false,
  })
```
**Run the workflow directly**
```
  const { acknowledgement, errors, result } = await buildProjectEnvironmentWorkflow.run({
    container: req.scope,
    input: {},
    context: { requestId: req.requestId },
    throwOnError: false,
  })
```
**Run workflow within a workflow**
```
const workflow = createWorkflow("", function () {
    const subworkflowResult = subWorkflow.runAsStep({ ... })
  })
```


olivermrbl (commented):
question: Would it be possible to avoid the `asStep`?

olivermrbl:
btw. love this idea

adrien2p:
> question: Would it be possible to avoid the `asStep`?

No we can't really by pass it currently, without that we can't properly identify the typings. But if we dont care about typings then we can

carlos-r-l-rodrigues:
> question: Would it be possible to avoid the `asStep`?

at runtime it is 100% possible, but we lose types.

carlos-r-l-rodrigues (approved):

olivermrbl:
Hmm, I am actually leaning toward trading off types for a simpler API, but I think I would have to play around with it a bit. Can we hold off merging for a day or two, or do we need it now?

adrien2p:
Without the typings you will have a lot of cast to do in exchange. 
The run is returning a promise, so you would have to cast it as not being a promise. Also, it returns an object with result, errors etc, you would have to cast it to a workflow result of the type you want etc etc. I am not sure the trade off is worth it considering the complexity it would bring to the usage no?

Also, how to explain the difference of behaviour between calling run in a workflow or elsewhere ? I don't believe mixing multiple behaviour like this one under a single function would be good as it could create a lot of confusion

adrien2p:
@olivermrbl @carlos-r-l-rodrigues what do you think about [this](https://github.com/medusajs/medusa/pull/7183/commits/3c4075e37a7f2a179a2bd1639e553363e22187ee) basically we hide the logic under the same name as the run we used through the exported workflow.

The diff between the two is that outside of the composer we do `workflow(container).run` and inside the workflow composer we do `workflow.run`

olivermrbl:
> The run is returning a promise, so you would have to cast it as not being a promise. Also, it returns an object with result, errors etc, you would have to cast it to a workflow result of the type you want etc etc. I am not sure the trade off is worth it considering the complexity it would bring to the usage no?
> Also, how to explain the difference of behaviour between calling run in a workflow or elsewhere ? I don't believe mixing multiple behaviour like this one under a single function would be good as it could create a lot of confusion

Yeah, you might be right. That's why I wanted to give it a spin to see what it feels like.

> what do you think about [this](https://github.com/medusajs/medusa/pull/7183/commits/3c4075e37a7f2a179a2bd1639e553363e22187ee) basically we hide the logic under the same name as the run we used through the exported workflow

I really like this! That would ~~effectively~~ almost make it the same API.

adrien2p:
@olivermrbl let us know what you think after giving it a spin 👍 

Also, notice that this run in the composer return a step as usual, so you have the ability to call config on it as well and manipulate the result as any other steps

sradevski:
> @olivermrbl @carlos-r-l-rodrigues what do you think about [this](https://github.com/medusajs/medusa/pull/7183/commits/3c4075e37a7f2a179a2bd1639e553363e22187ee) basically we hide the logic under the same name as the run we used through the exported workflow.
> 
> The diff between the two is that outside of the composer we do `workflow(container).run` and inside the workflow composer we do `workflow.run`

My 2-cents on this is that the fact that they are ALMOST but not exactly the same will cause more confusion than having a different `asStep` call, rather than run. I think if we can't make it exactly the same it's better to make it very different, so it is very clear how the usages differ.

sradevski (commented):
This is great to see, let's just settle on the API and get the current nested workflows cleaned up.

olivermrbl:
> My 2-cents on this is that the fact that they are ALMOST but not exactly the same will cause more confusion than having a different asStep call, rather than run. 

Yeah, after giving this some thought, I think this makes sense. 

My concern is not so much with whether we use `run` or `runAsStep`. The part that is a sub-optimal DX is the fact that the way to initiate AND run the workflow is different for the two use cases, which is what you also allude to @sradevski.

Update: Will open a discussion internally, as I have a few thoughts on this.

adrien2p:
> > My 2-cents on this is that the fact that they are ALMOST but not exactly the same will cause more confusion than having a different asStep call, rather than run. 
> 
> 
> 
> Yeah, after giving this some thought, I think this makes sense. 
> 
> 
> 
> My concern is not so much with whether we use `run` or `runAsStep`. The part that is a sub-optimal DX is the fact that the way to initiate AND run the workflow is different for the two use cases, which is what you also allude to @sradevski.
> 
> 
> 
> Update: Will open a discussion internally, as I have a few thoughts on this.

It is different indeed but they are not doing the same thing. 

In one case it initiates and executes the workflow, in the other one it is a static wrapper on the workflow that creates a step to include in a workflow composer but does not initiate and execute the workflow, the later part happens in the step that is being returned. 

The comparison would be like a static method factory on a class for example which would create the class under the hood with some param, but does not execute the instance method.

That it why i originally call it as step to give a sense that we are wrapping the future execution into a step. Let's discuss that in the conv

olivermrbl:
I have a suggestion for an alternative approach. Please consider this a braindump, as I haven't thought it through:

We unify the API for running a workflow by forcing the `container` to be passed in the config. This would mean that there are two ways to run a workflow:

**Run the workflow with the workflow engine**
```
const { acknowledgement, errors, result } = await wfe.run("workflow-id", {
    container: req.scope,
    context: { requestId: req.requestId },
    throwOnError: false,
  })
```
**Run the workflow directly**
```
  const { acknowledgement, errors, result } = await buildProjectEnvironmentWorkflow.run({
    container: req.scope,
    input: {},
    context: { requestId: req.requestId },
    throwOnError: false,
  })
```

That would unify the API between these two approaches. But more importantly, it will enable us to bring the same DX to the "workflow-as-steps" API:
```
const workflow = createWorkflow("", function () {
    const subworkflowResult = subWorkflow.run({ ... })
  })
```



Now, this spawns a few immediate questions: 
1. Do we lose anything by removing the `workflow(container).run` approach? 
2. And if so, can we move this elsewhere in favor of a unified API?

adrien2p:
@olivermrbl , i believe i already did that before 😀 and i think it would be cool to not have to pass the container separately as today but part of the run config. I can take some times tomorrow to go through that and identify if there will be any issues, but theoretically i dont think so

packages/workflows-sdk/src/utils/composer/create-workflow.ts
```diff
@@ -256,5 +263,36 @@ export function createWorkflow<
...
 
+  mainFlow.run = ({
+    input,
+  }: {
+    input: TData
+  }): ReturnType<StepFunction<TData, TResult>> => {
+    // TODO: Async sub workflow is not supported yet
+    // Info: Once the export workflow can fire the execution through the engine if loaded, the async workflow can be executed,
+    // the step would inherit the async configuration and subscribe to the onFinish event of the sub worklow and mark itself as success or failure
+    return createStep(
```
riqwan, line 197: ayy very cool! 🎉 

Just out of curiosity, do you think it would be possible export this directly?

Something like:

```
export const workflow = createWorkflow(
  workflowId,
  (input => {
    SomeStep(input)
    
    return WorkflowStep(differentWorkflow, input)
  })
```

adrien2p: I ve slightly change the API to unify them all 👍 wdyt?

sradevski (approved):
I think this looks good to me, we basically have `run` and `runAsStep` and the container is passed as an arg right?

adrien2p:
@sradevski yes thats correct, we still maintain the backward compatibility so that we support users that already uses it and i dont have to migrate all our workflows now 👍

riqwan (approved):

olivermrbl:
LGTM

PR #7196, opened 2024-05-01
fix: Correctly fetch cart when updating a line item

sradevski:


packages/medusa/src/api-v2/store/carts/[id]/line-items/[line_id]/route.ts
```diff
@@ -2,28 +2,35 @@ import {
...
 import { StoreUpdateCartLineItemType } from "../../../validators"
+import { prepareListQuery } from "../../../../../../utils/get-query-config"
 
 export const POST = async (
   req: MedusaRequest<StoreUpdateCartLineItemType>,
   res: MedusaResponse
 ) => {
-  const cartModuleService = req.scope.resolve<ICartModuleService>(
-    ModuleRegistrationName.CART
+  const cart = await refetchCart(
```
Comment (sradevski) [inline packages/medusa/src/api-v2/store/carts/[id]/line-items/[line_id]/route.ts:15]: We are using prepareListQuery to have support for *

olivermrbl (approved):

PR #7198, opened 2024-05-01
fix: Accept filters in `softDelete` + fulfillment location clean-up

olivermrbl:


olivermrbl:
@adrien2p, I might need you to take over on this one. We need to support filter-based soft deletes similar to what we support in regular deletes. Otherwise, cascading soft-deletes will not work as expected. We can huddle about this.

carlos-r-l-rodrigues (approved):

packages/core-flows/src/stock-location/steps/associate-locations-with-fulfillment-sets.ts
```diff
@@ -24,12 +24,12 @@ export const associateFulfillmentSetsWithLocationStep = createStep(
...
           return {
-            [Modules.FULFILLMENT]: {
-              fulfillment_set_id: id,
-            },
             [Modules.STOCK_LOCATION]: {
               stock_location_id: link.location_id,
             },
+            [Modules.FULFILLMENT]: {
+              fulfillment_set_id: id,
+            },
```
adrien2p: @olivermrbl @carlos-r-l-rodrigues somehow we can't retrieve the link module that correspond to this tuple 🤔

integration-tests/api/__tests__/admin/stock-location/index.spec.ts
```diff
@@ -390,7 +398,7 @@ medusaIntegrationTestRunner({
         stockLocationId = createResponse.data.stock_location.id
       })
 
-      it("should create a fulfillment set for the location", async () => {
+      it.only("should create a fulfillment set for the location and then delete it and its associations", async () => {
```
adrien2p: todo: rm .only :D

PR #7201, opened 2024-05-02
feat(medusa,core-flows): complete cart [part-1]

riqwan:
what:

- adds a very basic complete cart endpoint that creates an order
- the complete cart workflow currently does the following:
  - create tax lines
  - create order

### DIFF
--- a/packages/core-flows/src/definition/cart/steps/create-order-from-cart.ts
+++ b/packages/core-flows/src/definition/cart/steps/create-order-from-cart.ts
@@ -7,9 +7,9 @@ interface StepInput {
   cart: CartWorkflowDTO
 }
 
-export const createOrderFromCartWorkflowId = "create-order-from-cart"
-export const createOrderFromCartWorkflow = createStep(
-  createOrderFromCartWorkflowId,
+export const createOrderFromCartStepId = "create-order-from-cart"
+export const createOrderFromCartStep = createStep(
+  createOrderFromCartStepId,
   async (input: StepInput, { container }) => {
     const { cart } = input
 
--- a/packages/core-flows/src/definition/cart/workflows/complete-cart.ts
+++ b/packages/core-flows/src/definition/cart/workflows/complete-cart.ts
@@ -1,7 +1,7 @@
 import { CompleteCartWorkflowInputDTO, OrderDTO } from "@medusajs/types"
 import { WorkflowData, createWorkflow } from "@medusajs/workflows-sdk"
 import { useRemoteQueryStep } from "../../../common"
-import { createOrderFromCartWorkflow } from "../steps"
+import { createOrderFromCartStep } from "../steps"
 import { updateTaxLinesStep } from "../steps/update-tax-lines"
 import { completeCartFields } from "../utils/fields"
 
@@ -38,6 +38,6 @@ export const completeCartWorkflow = createWorkflow(
       list: false,
     }).config({ name: "final-cart" })
 
-    return createOrderFromCartWorkflow({ cart: finalCart })
+    return createOrderFromCartStep({ cart: finalCart })
   }
 )


packages/core-flows/src/definition/cart/workflows/complete-cart.ts
```diff
@@ -0,0 +1,19 @@
+import { CompleteCartWorkflowInputDTO } from "@medusajs/types"
+import { WorkflowData, createWorkflow } from "@medusajs/workflows-sdk"
+
+/*
+  - [] Create Tax Lines
+  - [] Authorize Payment
+    - fail:
+      - [] Delete Tax lines
+  - [] Reserve Item from inventory (if enabled)
```
sradevski, line 13: Should we reserve the items first, since that doesn't require interacting with an external service, and it serves as some sort of a lock, avoiding the possibility of us authorizing a payment for an item that is no longer available.

packages/core-flows/src/definition/cart/workflows/complete-cart.ts
```diff
@@ -0,0 +1,19 @@
+import { CompleteCartWorkflowInputDTO } from "@medusajs/types"
+import { WorkflowData, createWorkflow } from "@medusajs/workflows-sdk"
+
+/*
+  - [] Create Tax Lines
```
sradevski, line 9: I guess we should do some validation beforehand, that the prices of the products/cart haven't changed, the shipping method is still available, etc.

sradevski (approved):
LGTM for the boilerplate, just a couple of comments for the follow-up PRs

packages/core-flows/src/definition/cart/workflows/complete-cart.ts
riqwan: Can definitely see some value is changing a few things here, but for now I just want to get this to a point where its similar to v1 endpoint. 

PS: the PR is still in draft. Want to get a basic test working before opening up the boilerplate code for merge.

sradevski: @riqwan sounds good, it's fine to get the boilerplate in place anyway 👍

riqwan:
@carlos-r-l-rodrigues added a basic cart complete endpoint here.

riqwan:
@sradevski this should now be a workable endpoint to use in the starter to create an order. @carlos-r-l-rodrigues and I will add steps in the workflow to complete this endpoint.

packages/medusa/src/api-v2/store/carts/[id]/complete/route.ts
```diff
@@ -0,0 +1,34 @@
...
+    throw errors[0].error
+  }
+
+  const order = await refetchOrder(
+    result.id,
+    req.scope,
+    req.remoteQueryConfig.fields
+  )
+
+  res.status(200).json({ order })
```
Comment (olivermrbl) [inline packages/medusa/src/api-v2/store/carts/[id]/complete/route.ts:33]: thought(non-blocking): There will likely be failure modes where we have to return early with the cart instead of the order e.g. if the payment authorization falls through

packages/core-flows/src/definition/cart/steps/create-order-from-cart.ts
```diff
@@ -0,0 +1,57 @@
...
+import { OrderStatus } from "@medusajs/utils"
+import { StepResponse, createStep } from "@medusajs/workflows-sdk"
+import { createOrdersWorkflow } from "../../../order/workflows/create-orders"
+
+interface StepInput {
+  cart: CartWorkflowDTO
+}
+
+export const createOrderFromCartWorkflowId = "create-order-from-cart"
+export const createOrderFromCartWorkflow = createStep(
```
olivermrbl: question: Should this be suffixed with `Step` instead of workflow?

If this is intended to be a workflow, then we would use `createWorkflow` and `runAsStep` in cart-completion

olivermrbl (approved):
excited to see this being pieced together!

packages/medusa/src/api-v2/store/carts/[id]/complete/route.ts
Comment (riqwan) [reply packages/medusa/src/api-v2/store/carts/[id]/complete/route.ts:33]: Shouldn't we throw with special codes in those cases?

packages/core-flows/src/definition/cart/steps/create-order-from-cart.ts
riqwan: good catch, will move it around.

packages/medusa/src/api-v2/store/carts/[id]/complete/route.ts
Comment (adrien2p) [reply packages/medusa/src/api-v2/store/carts/[id]/complete/route.ts:33]: we do not have that feature currently to re run a workflow from a failed step, instead maybe we should have smaller workflow that takes care of the different blocks of action and we run them sequentially here with the existing idempotency key mechanism? what is the plan at the moment?

cc @carlos-r-l-rodrigues

Comment (riqwan) [reply packages/medusa/src/api-v2/store/carts/[id]/complete/route.ts:33]: If we have time, I think we should go ahead and try getting this working with one workflow. We already have all the data in the workflow engine, we just have to point the orchestrator to run the flow to the right step. 

If not, like you said, we'll have to fall back to route level handling.

Comment (adrien2p) [reply packages/medusa/src/api-v2/store/carts/[id]/complete/route.ts:33]: yes that means reverting only to a certaint point and restart from there cc @carlos-r-l-rodrigues looks like the goto we discussed no?

PR #7204, opened 2024-05-02
fix(fulfillment): Update shipping options rules

adrien2p:
**What**
Fix the update shipping options to apply the rules update properly and remove unused code

### DIFF
--- a/.changeset/shiny-apples-serve.md
+++ b/.changeset/shiny-apples-serve.md
@@ -1,0 +1,5 @@
+---
+"@medusajs/fulfillment": patch
+---
+
+fix(fulfillment): Update shipping options rules


packages/fulfillment/src/services/fulfillment-module-service.ts
```diff
@@ -1148,14 +1148,33 @@ export default class FulfillmentModuleService<
         shippingOption
       )
 
-      const existingRulesMap = new Map(
-        existingRules.map((rule) => [rule.id, rule])
-      )
+      const existingRulesMap: Map<
```
sradevski, line 1151: Do you think it makes sense to use upsertWithReplace here?

sradevski (approved):

packages/fulfillment/src/services/fulfillment-module-service.ts
adrien2p: I would like to not spread that upsertWithReplace too much for now, because I believe we need to rework it when the DAL will be made properly replaceable and I believe for now this helper needs to be rethinked at a later point since at the moment if you replace the DAL you loose that feature and the end user need to re write it completely which can be hard for a lot of users. since we are not in a place where there is many upsert with replace and it is pretty contained, I believe it is fine to keep it like it is. It was a bit different for the product where many of those actions was performed. is that ok for you? we can discuss more about it later if you want

fPolic (approved):

PR #7211, opened 2024-05-02
feat(medusa-cli): added v2 flag

shahednasser:
Added a `--v2` flag to the `medusa new` command that clones the V2 branch of the starter


olivermrbl (approved):
Noice!

PR #7215, opened 2024-05-02
chore(): start moving some packages to the core directory

adrien2p:
**What**
Move to core the following packages, more might still move depending on future decision
- core|tooling
  - workflows-sdk
  - modules-sdk
  - orchestration
  - types
  - utils
  - test-utils
- cli
  - create-medusa-app
  - medusa-cli
  - medusa-dev-cli
  - oas

**NOTE**
The medusa package will not be moved as part of that pr, I would do that once I have finalized the refactoring to not have to many conflicts with the current work on the api

will wait to remove the generated dir until everything else is moved and react/js packages are refactored somehow


sradevski:
@adrien2p the only one I am not so sure of is the `core-flows` package. I think `core` should not contain any business logic, and as such, `core-flows` might be better of staying at the top level, similar to `medusa`. Let's see what others think though.

adrien2p:
> @adrien2p the only one I am not so sure of is the `core-flows` package. I think `core` should not contain any business logic, and as such, `core-flows` might be better of staying at the top level, similar to `medusa`. Let's see what others think though.

I agree and I wasn't sure as well, this one needs some discussion I agree

adrien2p:
My thinking was more that the core-flows is our core opinionated business logic and maybe in that case it can go to the core dir.

sradevski:
> My thinking was more that the core-flows is our core opinionated business logic and maybe in that case it can go to the core dir.

Yeah let's see, maybe we can split them in 2 different folders, one that is business-agnostic, and one that has business logic. I guess `types` would also fall into the business logic bucket rather than the core

PR #7217, opened 2024-05-02
chore(): rm payment plugins

adrien2p:
**What**
rm legacy payment plugins
the following will require a migration to providers
- paypal 
- klarna


riqwan (approved):
delish! 🔥

PR #7227, opened 2024-05-03
chore: complete cart part 2

carlos-r-l-rodrigues:
- Confirm inventory as workflow

### DIFF
--- a/packages/modules/link-modules/src/definitions/readonly/cart-product.ts
+++ b/packages/modules/link-modules/src/definitions/readonly/cart-product.ts
@@ -26,6 +26,18 @@ export const CartProduct: ModuleJoinerConfig = {
         },
       },
     },
-    // TODO: Add 'carts' to product modules
+    {
+      serviceName: Modules.PRODUCT,
+      relationship: {
+        serviceName: Modules.CART,
+        primaryKey: "variant_id",
+        foreignKey: "id",
+        alias: "cart_items",
+        isList: true,
+        args: {
+          methodSuffix: "LineItems",
+        },
+      },
+    },
   ],
 }


packages/core/core-flows/src/definition/cart/workflows/confirm-variant-inventory.ts
```diff
@@ -0,0 +1,96 @@
...
+      }
+
+      if (salesChannelId && !hasSalesChannelStockLocation) {
+        throw new MedusaError(
+          MedusaError.Types.INVALID_DATA,
+          `Sales channel ${salesChannelId} is not associated with any stock locations.`
+        )
+      }
+
+      if (priceNotFound.length && !data.input.ignore_price_check) {
```
olivermrbl, line 75: question: It feels a bit weird to perform a price check in this workflow, no?

packages/modules/link-modules/src/definitions/readonly/cart-product.ts
```diff
@@ -0,0 +1,43 @@
+import { Modules } from "@medusajs/modules-sdk"
+import { ModuleJoinerConfig } from "@medusajs/types"
+
+export const CartProduct: ModuleJoinerConfig = {
```
olivermrbl, line 4: question: Is this used anywhere, I can't seem to find it?

carlos-r-l-rodrigues: at `packages/core/core-flows/src/definition/cart/utils/fields.ts`
`items.variant` is the link

packages/core/core-flows/src/definition/cart/workflows/confirm-variant-inventory.ts
carlos-r-l-rodrigues: yes, I just transposed the full logic that was in place.
I can move that to another place in the next PR.
Is the cart flow the right place to check that though?

olivermrbl: I think it was added as an extra safety measure to ensure you are not adding a variant without a price. We can remove it for now and assume that variants have prices, but we might add it back in at some point.

packages/modules/link-modules/src/definitions/readonly/cart-product.ts
olivermrbl: Got it, that's interesting. It was not immediately obvious to me, that this was extending items.variant. How is that possible without any mention of items in the joiner config?

packages/core/core-flows/src/definition/cart/workflows/confirm-variant-inventory.ts
carlos-r-l-rodrigues: I can create an extra step to check that. no problem. and remove this check from here

packages/modules/link-modules/src/definitions/readonly/cart-product.ts
carlos-r-l-rodrigues: at the moment, without the GraphQL definition a link is module to module, so it will try to get the defined property alias at any "level".
this is why we have to also add the "args.methodSuffix" to it.
When we have the GQL definition, we gonna link at the Entity level directly. then it would be from the entity CartItem to ProductVariant.

carlos-r-l-rodrigues: btw `foreignKey: "items.variant_id"` also works
I will keep it like this so it is more explicity while we don't have the graph of all modules.

packages/core/core-flows/src/definition/cart/workflows/confirm-variant-inventory.ts
olivermrbl: Do you want to leave this for a separate PR, or include this now?

packages/modules/link-modules/src/definitions/readonly/cart-product.ts
olivermrbl: This is easier to read imo 👍

olivermrbl (approved):
LGTM, just a comment

packages/core/core-flows/src/definition/cart/workflows/confirm-variant-inventory.ts
carlos-r-l-rodrigues: Yeah, I'll add it in a next PR.

PR #7230, opened 2024-05-03
feat(fulfillment, core-flows): Apply correct address and context + cleanup

adrien2p:
The tests will be added in subsequent requests where the worklows are being worked on


packages/core/core-flows/src/definition/cart/workflows/list-shipping-options-for-cart.ts
```diff
@@ -1,11 +1,7 @@
-import { ListShippingOptionsForCartWorkflowInputDTO } from "@medusajs/types"
-import { deepFlatMap, MedusaError } from "@medusajs/utils"
-import {
-  createWorkflow,
-  transform,
-  WorkflowData,
-} from "@medusajs/workflows-sdk"
-import { useRemoteQueryStep } from "../../../common/steps/use-remote-query"
+import {ListShippingOptionsForCartWorkflowInputDTO} from "@medusajs/types"
```
olivermrbl: looks like we need some linting

sradevski (approved):
Nice 👍

packages/core/core-flows/src/definition/cart/workflows/list-shipping-options-for-cart.ts
adrien2p: nice catch 💪

PR #7231, opened 2024-05-03
feat: Consolidate payment functions and handle sessions for storefront

sradevski:


packages/core/core-flows/src/definition/cart/steps/validate-cart-shipping-options.ts
```diff
@@ -5,6 +5,10 @@ import { createStep, StepResponse } from "@medusajs/workflows-sdk"
 
 interface StepInput {
   cart: CartDTO
+  shippingOptionsContext: {
```
sradevski, line 8: For now these are the only things we can set through the UI, we can revisit this check once we get closer to prod readiness

adrien2p:
[this](https://github.com/medusajs/medusa/pull/7230) pr should be merged soon, you will be able to use the changes

packages/medusa/src/api-v2/store/payments/payment-providers/route.ts
```diff
@@ -0,0 +1,35 @@
...
+  AuthenticatedMedusaRequest,
+  MedusaResponse,
+} from "../../../../types/routing"
+import {
+  ContainerRegistrationKeys,
+  remoteQueryObjectFromString,
+} from "@medusajs/utils"
+import { StoreGetPaymentProvidersParamsType } from "../validators"
+
+// TODO: The provider fetching need to be context-based, similar to shipping methods.
```
olivermrbl: Payment providers are configured per Region, so we could consider moving this endpoint:
```
GET /store/regions/:region_id/payment-providers
```

Alternatively, we would need to pass the `region_id` ID here. There's no other context to take into account.

packages/medusa/src/api-v2/store/payments/payment-sessions/route.ts
```diff
@@ -0,0 +1,84 @@
+import {
```
olivermrbl: thought(non-blocking): Not convinced that this mismatch between endpoint path and response is the right approach. It would break with the general convention of our Rest API.

sradevski: @olivermrbl I kept it as it was originally, but I completely agree that we shouldn't have this mismatch. I'll look into changing it.

On a more general note what do you think our payment-related APIs should look like?

I added them under `payments`, but honestly I am not convinced if that's the best way?

packages/medusa/src/api-v2/store/payment-collections/validators.ts
```diff
@@ -0,0 +1,30 @@
...
+
+export type StoreCreatePaymentCollectionType = z.infer<
+  typeof StoreCreatePaymentCollection
+>
+export const StoreCreatePaymentCollection = z
+  .object({
+    cart_id: z.string(),
+    region_id: z.string(),
+    currency_code: z.string(),
+    amount: z.number(),
```
olivermrbl, line 28: thought: Think we should remove `amount` - you shouldn't be able to pass your own amount from the storefront.

packages/medusa/src/api-v2/store/payment-collections/route.ts
```diff
@@ -0,0 +1,50 @@
...
+  ContainerRegistrationKeys,
+  remoteQueryObjectFromString,
+} from "@medusajs/utils"
+import {
+  AuthenticatedMedusaRequest,
+  MedusaResponse,
+} from "../../../types/routing"
+import { StoreCreatePaymentCollectionType } from "./validators"
+
+export const POST = async (
```
olivermrbl, line 12: thought: We should think about how this endpoint can be made more generic, i.e. remove the strict dependency on a `cart_id`. Ideally, you can initiate a payment collection for any type of resource, but this is not important in the first iteration.

olivermrbl: This was the reason for having the initial payment collection creation go through the `/store/carts` resource.

olivermrbl (approved):
LGTM, just a few comments

packages/core/core-flows/src/definition/cart/workflows/add-shipping-method-to-cart.ts
```diff
@@ -40,6 +40,7 @@ export const addShippingMethodToWorkflow = createWorkflow(
     validateCartShippingOptionsStep({
       option_ids: optionIds,
       cart,
+      shippingOptionsContext: { is_return: "false", enabled_in_store: "true" },
```
riqwan, line 43: q: why are these hard-coded here? Shouldn't they come from some db data?

sradevski: @riqwan they are specific to the storefront, whenever you want to list shipping methods for a cart/storefront these two have to be set basically. It's not great, but that's how it works currently.

packages/medusa/src/api-v2/store/payment-collections/route.ts
sradevski: @olivermrbl I agree, but it's better if we have it as a separate endpoint, so if in the future we loosen the requirements (eg. you no longer need to pass a cart_id, but you can pass something else) it won't be a breaking change.

packages/medusa/src/api-v2/store/payment-collections/validators.ts
sradevski: Good point, I just took how it was before, but I completely agree 👍

PR #7248, opened 2024-05-06
feat(core-flows,typers,utils,medusa): add payment auth step to complete cart workflow - [complete cart part 3]

riqwan:
what:

- adds step to authorize payment sessions
- workflow throws relevant errors
- http returns cart/order or throw depending on workflow errors

### DIFF
--- a/packages/core/core-flows/src/definition/cart/steps/validate-cart-payments.ts
+++ b/packages/core/core-flows/src/definition/cart/steps/validate-cart-payments.ts
@@ -39,7 +39,6 @@ export const validateCartPaymentsStep = createStep(
       )
     }
 
-    // TODO: Should we be able to authorize multiple payment sessions for a cart?
-    return new StepResponse(paymentsToProcess[0])
+    return new StepResponse(paymentsToProcess)
   }
 )
--- a/packages/core/core-flows/src/definition/cart/workflows/complete-cart.ts
+++ b/packages/core/core-flows/src/definition/cart/workflows/complete-cart.ts
@@ -8,7 +8,6 @@ import { useRemoteQueryStep } from "../../../common"
 import { authorizePaymentSessionStep } from "../../../payment/steps/authorize-payment-session"
 import { createOrderFromCartStep, validateCartPaymentsStep } from "../steps"
 import { reserveInventoryStep } from "../steps/reserve-inventory"
-import { updateTaxLinesStep } from "../steps/update-tax-lines"
 import { completeCartFields } from "../utils/fields"
 import { confirmVariantInventoryWorkflow } from "./confirm-variant-inventory"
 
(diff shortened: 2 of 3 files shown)


packages/core/core-flows/src/definition/cart/steps/validate-cart-payments.ts
```diff
@@ -0,0 +1,45 @@
...
+    )
+
+    if (!paymentsToProcess?.length) {
+      throw new MedusaError(
+        MedusaError.Types.INVALID_DATA,
+        `Payment sessions are required to complete cart`
+      )
+    }
+
+    // TODO: Should we be able to authorize multiple payment sessions for a cart?
```
olivermrbl: Don't think we should, but since this is more a generic validation step, maybe this should return all the payments to process, and then in our specific workflow, we select `[0]`, because we know there will only ever be one (for the time being).

packages/core/core-flows/src/definition/cart/workflows/complete-cart.ts
```diff
@@ -32,17 +22,25 @@ export const completeCartWorkflow = createWorkflow(
       list: false,
     })
 
+    const paymentSession = validateCartPaymentsStep({ cart })
+
+    // Question: While running cart complete, we seem to assume that the payment collection is already created.
```
olivermrbl: Hmm, I actually think this step was just translated from the existing cart completion strategy. I don't think we need it anymore. At least I can't think of a scenario right now.

packages/core/core-flows/src/definition/cart/workflows/complete-cart.ts
```diff
@@ -32,17 +22,25 @@ export const completeCartWorkflow = createWorkflow(
...
+    // The payment collection amount should already have the tax amount, why are we updating the tax lines here?
+    updateTaxLinesStep({ cart_or_cart_id: cart, force_tax_calculation: true })
+
+    authorizePaymentSessionStep({
+      id: paymentSession.id,
```
olivermrbl: Depends on how you treat my other comment:
```suggestion
      id: paymentSessions[0].id,
```

olivermrbl (approved):
Overall LGTM, nice work!

We might need to put some more thought into the error handling as we start to test the flow, but I think this works for now

packages/medusa/src/api-v2/store/carts/[id]/complete/route.ts
```diff
@@ -33,5 +64,8 @@ export const POST = async (
     req.remoteQueryConfig.fields
   )
 
-  res.status(200).json({ order })
+  res.status(200).json({
```
Comment (sradevski) [inline packages/medusa/src/api-v2/store/carts/[id]/complete/route.ts:67]: The fact that we return different types depending on the success is a bit strange to me. Can't we have a cart-order link, and we just return a cart here with the `order` expanded instead?

carlos-r-l-rodrigues (approved):

packages/medusa/src/api-v2/store/carts/[id]/complete/route.ts
Comment (riqwan) [reply packages/medusa/src/api-v2/store/carts/[id]/complete/route.ts:67]: I like the idea, we could bring this up in a future iteration. For now, its keeping in line with what we did with v1.

PR #7249, opened 2024-05-06
fix: package.json run test integration in packages

adrien2p:
**What**
- fix product category repositories
- align tests
- fix categories routes
- fix package.json scripts

### DIFF
--- a/integration-tests/api/__tests__/admin/product-category.ts
+++ b/integration-tests/api/__tests__/admin/product-category.ts
@@ -565,7 +565,7 @@ medusaIntegrationTestRunner({
         }
 
         const response = await api.post(
-          `/admin/product-categories?fields=*category_children,*parent_category`,
+          `/admin/product-categories`,
           payload,
           adminHeaders
         )
@@ -600,7 +600,7 @@ medusaIntegrationTestRunner({
         }
 
         const response = await api.post(
-          `/admin/product-categories?fields=*category_children`,
+          `/admin/product-categories`,
           payload,
           adminHeaders
         )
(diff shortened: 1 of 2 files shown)


packages/modules/product/src/repositories/product-category.ts
```diff
@@ -51,19 +51,27 @@ export class ProductCategoryRepository extends DALUtils.MikroOrmBaseTreeReposito
...
+      populate.includes("parent_category") ||
+      fields.some((field) => field.startsWith("parent_category"))
 
     if (shouldExpandParent) {
-      populate.indexOf("parent_category") === -1 &&
-        populate.push("parent_category")
+      // Since the tree is manually built, we dont want to get the relations
+      const relationIndex = populate.indexOf("parent_category")
+      relationIndex !== -1 && populate.splice(relationIndex, 1)
+      familyOptions.includeAncestorsTree ??= true
```
olivermrbl: We cannot assume that wanting to expand `parent_category` is the same as wanting to include ancestors. 

`include_ancestors` will expand all parent categories, while as only passing `parent_category` should expand the immediate parent.

olivermrbl: Same goes for descendants

adrien2p: ok, got it, I ll change that 👍

carlos-r-l-rodrigues (approved):

carlos-r-l-rodrigues:
@olivermrbl if you can have a look
this one is needed to unblock a couple other PRs.

olivermrbl (approved):

PR #7250, opened 2024-05-06
feat(medusa): complete cart create reservation

carlos-r-l-rodrigues:


packages/core/core-flows/src/definition/cart/steps/reserve-inventory.ts
```diff
@@ -0,0 +1,51 @@
...
+    const inventoryService = container.resolve<IInventoryServiceNext>(
+      ModuleRegistrationName.INVENTORY
+    )
+
+    const items = data.items.map((item) => ({
+      line_item_id: item.id,
+      inventory_item_id: item.inventory_item_id,
+      quantity: item.required_quantity * item.quantity,
+      allow_backorder: item.allow_backorder,
+      location_id: item.location_ids[0],
```
olivermrbl, line 29: Why does the step accept multiple location IDs, if we always use only one?

packages/core/core-flows/src/order/workflows/create-orders.ts
```diff
@@ -39,8 +38,8 @@ export const createOrdersWorkflow = createWorkflow(
       findOrCreateCustomerStep({
         customerId: input.customer_id,
         email: input.email,
-      }),
-      validateVariantsExistStep({ variantIds })
+      })
+      // validateVariantsExistStep({ variantIds })
```
olivermrbl, line 42: should this be removed?

olivermrbl (approved):
LGTM

packages/core/core-flows/src/order/workflows/create-orders.ts
carlos-r-l-rodrigues: I will refactor the order workflows to match all the changes done in carts. If you don't mind I will keep this commented out for now.

The RemoteQuery will validate if the ids exist or not, but it should be possible to customize the message thrown by the RemoteQuery.

packages/core/core-flows/src/definition/cart/steps/reserve-inventory.ts
carlos-r-l-rodrigues: Currently this is the behavior, [to get first location](https://github.com/medusajs/medusa/blob/884428a1b573e499d7659aefed639bf797147428/packages/medusa/src/services/product-variant-inventory.ts#L539).
But I believe this should be reviewed to allocate the stock in all available locations until it matches the quantity.

packages/core/core-flows/src/order/workflows/create-orders.ts
olivermrbl: Sure, let's keep it in

packages/core/core-flows/src/definition/cart/steps/reserve-inventory.ts
olivermrbl: Yeah I agree. We might even need to look into stock allocation strategies, so merchants can configure exactly how an order is fulfilled and from where

PR #7260, opened 2024-05-07
feat: Add missing typings for the http layer

sradevski:
Before I proceed with adding more types, I want us to align on the typings pattern we want to follow.

Few things to discuss:
1. Create base typings that are shared across admin and store. Do we want to follow the pattern?
2. How do we want to name the typings between Admin and Store? Do we export different packages for each? Or do we use a prefix as in this PR?
3. Some of the typings are named as `<Entity>Response`, while others are not. I think removing/not having the `Response` suffix is more readable and we don't lose any information, but we need to decide what pattern we want to follow.

Here is an example of how the typings are used in the SDK:
```
  list: async (
      queryParams?: FindParams & StoreRegionFilters,
      headers?: Record<string, any>
    ): Promise<{ regions: StoreRegion[] } & PaginatedResponse> => {...}
```

And the same typings would be used through the frontend on components


packages/core/types/src/http/collection/common.ts
```diff
@@ -0,0 +1,21 @@
...
+
+export interface BaseProductCollection {
+  id?: string
+  title?: string
+  handle?: string
+  created_at?: string
+  updated_at?: string
+  deleted_at?: string
+  products?: BaseProduct[]
+  metadata?: Record<string, unknown>
```
adrien2p: todo: we are missing the nullability, I think you can review the different types to ensure that nullable fields are configured here as well

packages/core/types/src/http/product/common.ts
```diff
@@ -0,0 +1,192 @@
+import { BaseFilterable, OperatorMap } from "../../dal"
+import { BaseProductCollection } from "../collection/common"
+
+export type ProductStatus = "draft" | "proposed" | "published" | "rejected"
```
adrien2p, line 4: todo: unrelated, but I ve seen that we have an enum for that in the types package and it shouldnt be there if we want to clean it up :) instead it should a type and you can maybe move this one in the place the enum is located in the product type. wdyt?

packages/core/types/src/http/product/common.ts
```diff
@@ -0,0 +1,192 @@
...
+
+export interface BaseProduct {
```
adrien2p, line 6: Q: those the base interface takes into concideration the not allowed fields from the store end point? just to not leak something that is not allowed but is present in the types

packages/core/types/src/http/product/common.ts
```diff
@@ -0,0 +1,192 @@
...
+  tags?: BaseProductTag[]
+  variants?: BaseProductVariant[]
+  options?: BaseProductOption[]
+  images?: BaseProductImage[]
+  discountable?: boolean
+  external_id?: string
+  created_at?: string
+  updated_at?: string
+  deleted_at?: string
+  metadata?: Record<string, unknown>
```
adrien2p: todo: re work the nullability everywhere

packages/core/types/src/http/product/common.ts
```diff
@@ -0,0 +1,192 @@
...
+  tags?: {
+    value?: string[]
+  }
+  type_id?: string | string[]
+  category_id?: string | string[] | OperatorMap<string>
+  categories?: { id: OperatorMap<string> } | { id: OperatorMap<string[]> }
+  collection_id?: string | string[] | OperatorMap<string>
+  created_at?: OperatorMap<string>
+  updated_at?: OperatorMap<string>
+  deleted_at?: OperatorMap<string>
```
adrien2p, line 146: Q: have we added the support for the operators in the http validators? for the other props as well

adrien2p (commented):
Is it something that we should spend time on right now? I mean, I agree that we need the typings but wouldn't our time be better spend on other topics for now so that we dont had more places where we have to change the types if we change something 5 levels up? 

Also, the base typing can be good but it needs to take into consideration the non allowed properties in the store API which differ from the Admin API in theory.

In terms of naming, as long as the user import them from the buddles it is fine, but if he tries to look for e.g Product, he might have some questions in regards to why there is many Product types exported?

packages/core/types/src/http/product/common.ts
sradevski: @adrien2p yes we have it there 👍 We will probably need to review that they are applied everywhere though.

packages/core/types/src/http/product/common.ts
sradevski: @adrien2p I don't think we need to have explicit `null` values as typings in the `http` response, what is the benefit of it? These models won't be used for write operations, we'll have separate typings for that. I am happy to add them, it's just that they clutter the typings a bit, but if there is a use-case we should definitely add them.

packages/core/types/src/http/product/common.ts
sradevski: @adrien2p the `Base<Entity>` typings will only have the types that are shared across both, and then each of `admin` and `store` will additionally extend what is specific to them. Does that make sense?

packages/core/types/src/http/product/common.ts
sradevski: @adrien2p yeah I will move things around once you're done with the cleanup

packages/core/types/src/http/collection/common.ts
sradevski: Commented below, I'll resolve this

sradevski:
> Is it something that we should spend time on right now? I mean, I agree that we need the typings but wouldn't our time be better spend on other topics for now so that we dont had more places where we have to change the types if we change something 5 levels up? I know we need them, just wondering if we wont have plenty of changes which will require us to change the types over and over again until we are stable? Maybe if we keep them minimal and easy to change thats fine 🤔

@adrien2p yes we need them now, both the admin and starter are currently with tons of `any` and it makes it really difficult to work on them. It is not really a ton of effort to create the typings, even if they are not complete for now, and I don't see most of them changing much anyway.

> Also, the base typing can be good but it needs to take into consideration the non allowed properties in the store API which differ from the Admin API in theory.

Commented below, the `base` typings will only be the intersection of admin and store.

> In terms of naming, as long as the user import them from the buddles it is fine, but if he tries to look for e.g Product, he might have some questions in regards to why there is many Product types exported?

How do you imagine this looking (I mean the bundle, what names will they have, etc.)? If you can provide an example that'd be great.

adrien2p:
We already export each thing as part of bundles, e.g you can access the http types from `HttpTypes`, same for the product modules, we have `ProductTypes`, they are exported from the `bundles` file in the types packages. 

Everything exported from the index is for lazyness, if you don't want to import the bundle and then from the bundle access the type, also to make discoverability easier. So, we just need to be careful with clash naming, so i would say that it is not a problem in it self to call something `Product` but we need to be aware that all the types might be accessible from the index and if there is clash naming then thats a problem. But personnaly I prefer to import from the bundle and from that bundle use the types so that it makes it clear where it is coming from and for the reader (consumer) they know where to find it easily

olivermrbl:
1. Create base typings that are shared across admin and store. Do we want to follow the pattern?

I think this pattern makes sense. We need to find a way to dedupe some of the typings, and this seems like an appropriate way of doing so.

2. How do we want to name the typings between Admin and Store? Do we export different packages for each? Or do we use a prefix as in this PR?

Prefixes as in this PR SGTM.


3. Some of the typings are named as <Entity>Response, while others are not. I think removing/not having the Response suffix is more readable and we don't lose any information, but we need to decide what pattern we want to follow.

I don't have any opinion about this, but I do like the explicit mention of response. Perhaps, there is an argument to take this even further and use `[Entity]HttpResponse`, but I don't know. I am fine with whatever approach we end up taking.

packages/core/types/src/http/product/common.ts
adrien2p: 👍 It is in the same package, but in the product directory

packages/core/types/src/http/product/common.ts
adrien2p: Yes that make sense, we will have to identify what is not allowed in the store 👍

packages/core/types/src/http/product/common.ts
adrien2p: I think they are important in the sense that they means different thing, undefined probably mean that the property is not there, null means it has been returned but does not have any values, so you can set some props to null and when you get them back they might be present but null, and it is not because they are not undefined that they are not null if that make sense

olivermrbl:
@sradevski, @adrien2p – let's try to align on the last few things tomorrow, so we can move forward with this one?

sradevski:
> @sradevski, @adrien2p – let's try to align on the last few things tomorrow, so we can move forward with this one?

@olivermrbl sounds good, I just focused on getting few more things out for reviewing so I can give people enough time for it. Let's align today 👍

olivermrbl:
I suggest we start with the following conventions:
- Sharing base types across domains
- Use prefixes for Admin and Store exports
- Export a bundle `HttpTypes` without the suffixes, so you do: `HttpTypes.Product`

Let me know what you think?

sradevski:
@olivermrbl sounds good to me, let's do that 👍

sradevski:
@olivermrbl I'll add the typings to the SDK in a separate PR, for this one let's focus on the typings.

I just realized that we can't do HttpTypes.Product, since we have both Admin and Store Http types.

What we could have is StoreTypes.Product and AdminTypes.Product instead.

adrien2p:
> @olivermrbl I'll add the typings to the SDK in a separate PR, for this one let's focus on the typings.
> 
> 
> 
> I just realized that we can't do HttpTypes.Product, since we have both Admin and Store Http types.
> 
> 
> 
> What we could have is StoreTypes.Product and AdminTypes.Product instead.

Yes that what was i was referring too about the clashes 😅 so lets have the bundled types as you suggested for admin and store, such that you can have AdminHttpTypes and StoreHttpTypes or eventually HttpTypes.Admin.Product and same for store wdyt?

olivermrbl:
Sounds good – as Adrien hints, it would be good to sneak in `Http` some how. How we do it, I don't have a strong opinion about, so I'm fine that you decide that

sradevski:
@adrien2p @olivermrbl at that point, does it make sense to keep it simple and have what we already have, eg.
`HttpTypes.StoreProduct` and `HttpTypes.AdminProduct`? This will reduce the amount of remapping we need to do as well?

olivermrbl:
That sounds good to me, @adrien2p what do you think?

adrien2p:
I am good with that too 💪

sradevski:
@adrien2p @olivermrbl sounds good, please review the PR then as that's how it already is 😄

packages/core/types/src/http/product/common.ts
sradevski: Oops missed addressing this comment. I think it makes sense, just one issue currently is that pretty much anything can be null, do we just set it on all fields? Not sure how we approach this..

adrien2p: Not sure everything can be null when it is expected to be present. But props like metadata for example are nullable while they are present as well as can be optional. Same for deleted at for example. I think we can start with the obvious ones like the two mentioned here and then we can refine along the water if the other ones are not obvious wdyt ?

sradevski: Yeah sure, let's go with that

sradevski: Added some already, we can refine as we go 👍

adrien2p: If we can add that to the limit too, if limit is null explicitly then we return everything otherwise we apply the value and if not provided we apply the default limit wdyt?

olivermrbl (approved):

PR #7266, opened 2024-05-07
feat: Add order endpoints for store

sradevski:


packages/medusa/src/api-v2/store/orders/[id]/route.ts
```diff
@@ -0,0 +1,13 @@
+import { MedusaRequest, MedusaResponse } from "../../../../types/routing"
+import { refetchOrder } from "../helpers"
+
+// TODO: Do we want to apply some sort of authentication here? My suggestion is that we do
```
Comment (olivermrbl) [inline packages/medusa/src/api-v2/store/orders/[id]/route.ts:4]: We haven't really had much previously, but it might make sense to add. So far, all you really needed was the order ID.

olivermrbl (approved):
LGTM

PR #7271, opened 2024-05-08
fix(medusa): Fix request type in upload route

shahednasser:
The `/admin/uploads` route uses an incorrect input type `CreateProductDTO`. This PR fixes it to the type defined in `validators.ts`.


sradevski (approved):
Nice catch 👍

PR #7272, opened 2024-05-08
fix(core-flows): set SalesChannels on Product update

fPolic:
**WHAT**
- set sales channels when updating products


packages/core/core-flows/src/sales-channel/steps/set-products-sales-channels.ts
```diff
@@ -0,0 +1,70 @@
...
+            },
+            [Modules.SALES_CHANNEL]: {
+              sales_channel_id: sc,
+            },
+          }
+        })
+      )
+      .flat()
+
+    const createdLinks = await remoteLink.create(linksToCreate)
```
carlos-r-l-rodrigues: `dismiss` and `create` have to be in different steps, and a compensation for each.
We can only have one operation per step, as we don't have DB transactions to rollback everything if something fails.
Or, we implement an upsert on the remote link that does both.

packages/core/core-flows/src/product/workflows/update-products.ts
```diff
@@ -20,7 +33,60 @@ export const updateProductsWorkflow = createWorkflow(
     input: WorkflowData<WorkflowInput>
   ): WorkflowData<ProductTypes.ProductDTO[]> => {
```
sradevski, line 138: Let's add at least 1 test in `product.js` that ensures this change is working as expected.

packages/core/core-flows/src/product/workflows/update-products.ts
```diff
@@ -20,7 +33,60 @@ export const updateProductsWorkflow = createWorkflow(
...
+        if (
+          !(data.input as UpdateProductsStepInputSelector).update.sales_channels
+        ) {
+          return []
+        }
+        return data.updatedProducts.map((p) => {
+          return {
+            product_id: p.id,
+            sales_channel_ids: (
+              (data.input as UpdateProductsStepInputSelector).update
```
sradevski: NIT: you can do `data.input.update.sales_channels?.map(c => c.id) ?? []` instead, it's a bit easier to read.

packages/core/core-flows/src/sales-channel/steps/set-products-sales-channels.ts
fPolic: ahh make sense, I will change the implementation

re upsert: think long term, it would be nice if the remote link could handle this since it would simplify steps implementation in a few places

fPolic: @carlos-r-l-rodrigues wdyt about [this](https://github.com/medusajs/medusa/pull/7272/commits/59ef5b5c0b12807f6bfe9cc33ed70fe356d6f940) approach?

packages/core/core-flows/src/product/workflows/update-products.ts
```diff
@@ -20,7 +37,87 @@ export const updateProductsWorkflow = createWorkflow(
...
+          }
+        }
+      )
+
+      return productIds
+    })
+
+    const currentLinks = getProductSalesChannelLinkStep({
+      ids: updatedProductIds,
+    })
```
carlos-r-l-rodrigues: no need to create `getProductSalesChannelLinkStep` function. you can just reuse the common step like this:

```
const currentLinks = useRemoteQueryStep({
  entry_point: "product_sales_channel",
  fields: ["product_id", "sales_channel_id"],
  variables: { product_id: updatedProductIds },
});
```

packages/core/core-flows/src/product/steps/get-product-sales-channel-link.ts
```diff
@@ -0,0 +1,33 @@
+import { Modules } from "@medusajs/modules-sdk"
+import { ContainerRegistrationKeys } from "@medusajs/utils"
+import { StepResponse, createStep } from "@medusajs/workflows-sdk"
+
+type StepInput = {
+  ids: string[]
+}
+
+export const getProductSalesChannelLinkStepId = "get-product-sales-channel-link"
+export const getProductSalesChannelLinkStep = createStep(
```
olivermrbl: question: Can we use the `remoteQueryStep` instead of having a separate step for this?

packages/core/core-flows/src/product/workflows/update-products.ts
```diff
@@ -20,7 +37,87 @@ export const updateProductsWorkflow = createWorkflow(
...
+        return data.updatedProducts
+          .map((p) => {
+            return (
+              data.input as UpdateProductsStepInputSelector
+            ).update.sales_channels!.map((channel) => ({
+              sales_channel_id: channel.id,
+              product_id: p.id,
+            }))
+          })
+          .flat()
```
carlos-r-l-rodrigues: take a look at the util `deepFlatMap` that we have.
it might help you reduce the complexity of these .map, .flat.

fPolic: Not sure if it's applicable here since we are merging data from 2 objects

packages/core/core-flows/src/product/workflows/update-products.ts
```diff
@@ -1,15 +1,32 @@
...
-      selector: ProductTypes.FilterableProductProps
-      update: ProductTypes.UpdateProductDTO
-    }
-  | {
-      products: ProductTypes.UpsertProductDTO[]
-    }
+  | UpdateProductsStepInputSelector
+  | UpdateProductsStepInputProducts
 
 type WorkflowInput = UpdateProductsStepInput
```
adrien2p, line 34: Q: We have a mix of workflow types in the types package and in the core-flows package, it will require a cleanup at some point, not now but just raising it. we also have different places in the types package where the workflow types are define, workflows, workflow, and in some dir. maybe we can create a ticket

adrien2p:
@fPolic I have done a commit proposal, can I get you to look at it and let me know what you think

packages/core/core-flows/src/product/workflows/update-products.ts
```diff
@@ -1,26 +1,139 @@
...
+    })
+
+    detachProductsFromSalesChannelsStep({ links: currentLinks })
+
+    const salesChannelLinks = transform(
+      { input, updatedProducts },
+      prepareSalesChannelLinks
+    )
+
+    associateProductsWithSalesChannelsStep({ links: salesChannelLinks })
```
adrien2p: suggestion: could we use the createLinksStep instead and in the preparation we return the propoer conf

packages/core/core-flows/src/product/workflows/update-products.ts
```diff
@@ -1,26 +1,139 @@
...
+      updateProductIds
+    )
+
+    const currentLinks = useRemoteQueryStep({
+      entry_point: "product_sales_channel",
+      fields: ["product_id", "sales_channel_id"],
+      variables: { product_id: updatedProductIds },
+    })
+
+    detachProductsFromSalesChannelsStep({ links: currentLinks })
```
adrien2p: suggestion: should we use removeRemoteLinkStep?

packages/core/core-flows/src/product/workflows/update-products.ts
```diff
@@ -1,26 +1,139 @@
...
+
+    const toUpdateInput = transform({ input }, prepareUpdateProductInput)
+    const updatedProducts = updateProductsStep(toUpdateInput)
+    const updatedProductIds = transform(
+      { updatedProducts, input },
+      updateProductIds
+    )
+
+    const currentLinks = useRemoteQueryStep({
+      entry_point: "product_sales_channel",
```
adrien2p, line 149: suggestion: service: Links.productSalesChannel

packages/core/core-flows/src/product/workflows/update-products.ts
```diff
@@ -1,26 +1,139 @@
...
+}) {
+  if ("products" in input) {
+    let productIds = updatedProducts.map((p) => p.id)
+    const discardedProductIds: string[] = input.products
+      .filter((p) => !p.sales_channels)
+      .map((p) => p.id as string)
+    return arrayDifference(productIds, discardedProductIds)
+  }
+
+  return !input.update.sales_channels ? [] : null
```
adrien2p: Q: should we return undefined instead of null?

packages/core/core-flows/src/product/workflows/update-products.ts
adrien2p: just saw that it requires a little adaptation of that step but it should be good to have so that we don't rely on the string for the link themselves wdyt?

fPolic:
> @fPolic I have done a commit proposal, can I get you to look at it and let me know what you think

Looks good @adrien2p, I've added generic link steps as you suggested

olivermrbl (approved):

packages/core/core-flows/src/product/workflows/update-products.ts
```diff
@@ -1,26 +1,167 @@
...
 export const updateProductsWorkflow = createWorkflow(
   updateProductsWorkflowId,
   (
     input: WorkflowData<WorkflowInput>
   ): WorkflowData<ProductTypes.ProductDTO[]> => {
     // TODO: Delete price sets for removed variants
-    // TODO Update sales channel links
-    return updateProductsStep(input)
+
+    const toUpdateInput = transform({ input }, prepareUpdateProductInput)
```
olivermrbl, line 141: thought: I don't know if I like, that the steps that were added here are so specific to sales channel management. E.g. as soon as we add another link to products, we will need to revisit all of these. Or create separate steps.

adrien2p: I need to relook into it but your comment is on the transformer, are you referring to the transformer of the next step or the step itself just to be sure

packages/core/core-flows/src/product/workflows/update-products.ts
olivermrbl: `prepareUpdateProductInput`, `updateProductIds`, and `prepareToDeleteLinks` are all named as if they were generic transformers for updating products, but they are coupled with sales channel management as far as I can tell. So what I mean is, if we add another link to product, we will need to refactor these to account for that.

adrien2p: Oh sorry maybe i ve named them wrongly, but we can rework that, i just based the name on the workflow step they were used for. I didnt have a better idea at the moment but the main thing was to extract them (not necessarily generic) to have a better readability of the workflow and transformers themselves

adrien2p: Also, those transformers are not exported they are only local so very specific here, i am fine with finding another name or rework the implementation a bit

olivermrbl:
Should we merge this and revisit later?

adrien2p:
It is fine to me, @fPolic wdyt?

fPolic:
Sounds good!

PR #7276, opened 2024-05-08
feat: Create the Medusa API SDK as js-sdk package

sradevski:
> The format of the SDK is something I temporarily did to use in the starter, so it is open for discussion.
> 
> Few points that I would like us to decide are:
> 1. Should `class Medusa` export the `store` endpoints as a flat list (how it was originally) or under a `store` field (how it > is now in this PR)? There are pros and cons with both approaches, so we need to decide what we prefer. Alternatively, we can export 2 separate `MedusaAdmin` and `MedusaStore` classes, and they will only contain either admin or store endpoints but not both.
> 2. How should our wrapper of `fetch` behave? Should we convert to JSON by default? Should we throw exceptions on non-200 statuses (similar to Axios)? For now I kept the API as the standard `fetch`, but I do think we can provide a bit better experience than that.
> 3. Anything else we would like to support as part of the SDK?

> Creating as draft until we have a bit better definition of how we want our SDK to look like.

The implementation focuses on the client, and less on the Store and Admin endpoints. Those will be implemented in follow-up PRs

### DIFF
--- a/packages/core/js-sdk/package.json
+++ b/packages/core/js-sdk/package.json
@@ -1,0 +1,40 @@
+{
+  "name": "@medusajs/js-sdk",
+  "version": "0.0.1",
+  "description": "SDK for the Medusa API",
+  "main": "dist/index.js",
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/medusajs/medusa",
+    "directory": "packages/core/js-sdk"
+  },
+  "engines": {
+    "node": ">=18"
+  },
+  "publishConfig": {
+    "access": "public"
+  },
+  "files": [
+    "dist"
+  ],
+  "author": "Medusa",
+  "license": "MIT",
+  "devDependencies": {
+    "cross-env": "^5.2.1",
+    "jest": "^29.6.3",
+    "msw": "^2.3.0",
+    "rimraf": "^5.0.1",
+    "ts-jest": "^29.1.1",
+    "typescript": "^5.1.6"
+  },
+  "dependencies": {
+    "@medusajs/types": "^1.11.16",
+    "qs": "^6.12.1"
+  },
+  "scripts": {
+    "prepublishOnly": "cross-env NODE_ENV=production tsc --build",
+    "build": "rimraf dist && tsc --build",
+    "test": "jest --passWithNoTests --runInBand --bail --forceExit",
+    "watch": "tsc --build --watch"
+  }
+}
--- a/packages/core/js-sdk/src/client.ts
+++ b/packages/core/js-sdk/src/client.ts
@@ -1,0 +1,183 @@
+import qs from "qs"
+import { ClientFetch, Config, FetchArgs, FetchInput, Logger } from "./types"
+
+const isBrowser = () => typeof window !== "undefined"
+
+const toBase64 = (str: string) => {
+  if (typeof window !== "undefined") {
+    return window.btoa(str)
+  }
+
+  return Buffer.from(str).toString("base64")
+}
+
+const sanitizeHeaders = (headers: Headers) => {
+  return {
+    ...Object.fromEntries(headers.entries()),
+    Authorization: "<REDACTED>",
+  }
+}
+
+const normalizeRequest = (
+  init: FetchArgs | undefined,
+  headers: Headers
+): RequestInit | undefined => {
+  let body = init?.body
+  if (body && headers.get("content-type")?.includes("application/json")) {
+    body = JSON.stringify(body)
+  }
+
+  return {
+    ...init,
+    headers,
+    ...(body ? { body: body as RequestInit["body"] } : {}),
+  } as RequestInit
+}
+
+const normalizeResponse = async (resp: Response, reqHeaders: Headers) => {
+  if (resp.status >= 300) {
+    const error = new FetchError(resp.statusText, resp.status)
+    throw error
+  }
+
+  // If we both requested JSON, we try to parse. Otherwise, we return the raw response.
+  const isJsonRequest = reqHeaders.get("accept")?.includes("application/json")
+  return isJsonRequest ? await resp.json() : resp
+}
+
+export class FetchError extends Error {
+  status: number | undefined
+
+  constructor(message: string, status?: number) {
+    super(message)
+    this.status = status
+  }
+}
+
+export class Client {
+  public fetch_: ClientFetch
+  private logger: Logger
+
+  private DEFAULT_JWT_STORAGE_KEY = "medusa_auth_token"
+  private token = ""
+
+  constructor(config: Config) {
+    const logger = config.logger || {
+      error: console.error,
+      warn: console.warn,
+      info: console.info,
+      debug: console.debug,
+    }
+
+    this.logger = {
+      ...logger,
+      debug: config.debug ? logger.debug : () => {},
+    }
+
+    this.fetch_ = this.initClient(config)
+  }
+
+  // Since the response is dynamically determined, we cannot know if it is JSON or not. Therefore, it is important to pass `Response` as the return type
+  fetch<T extends any>(input: FetchInput, init?: FetchArgs): Promise<T> {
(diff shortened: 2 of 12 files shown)


packages/core/js-sdk/src/store/index.ts
```diff
@@ -0,0 +1,267 @@
...
+      return this.client
+        .fetch(`/store/product-categories/${id}`, {
+          query: queryParams,
+          headers,
+        })
+        .then((resp) => resp.json())
+    },
+  }
+
+  public product = {
```
sradevski, line 77: I think these can be extracted into their own files, rather than implementing them inline, but I'll refrain from doing that until we decide on the few points I listed in the description.

srindom:
> 1. Should class Medusa export the store endpoints as a flat list (how it was originally) or under a store field (how it is now in this PR)? There are pros and cons with both approaches, so we need to decide what we prefer. Alternatively, we can export 2 separate MedusaAdmin and MedusaStore classes, and they will only contain either admin or store endpoints but not both.

Most of the time this SDK will be used for storefront implementations. I, therefore, suggest we stick to the old approach. I.e. do `medusa.region` -> `/store/region` and `medusa.admin.region` -> `/admin/region`.

> 2. How should our wrapper of fetch behave? Should we convert to JSON by default? Should we throw exceptions on non-200 statuses (similar to Axios)? For now I kept the API as the standard fetch, but I do think we can provide a bit better experience than that.

JSON is probably what people want 99/100 times. So, think it makes sense to convert to JSON. That said being able to opt-out of that would be nifty. 

> 3. Anything else we would like to support as part of the SDK?

Three things come to mind:

1. Built-in retries (with an `Idempotency-Key` header; either added by default or specified by the user). Some backed changes are needed before this works as intended, but I believe workflows will enable us to support idempotency on all operations very easily. 
2. Ability to use the SDK for sending requests to custom endpoints; could just be `medusa.request(method, path, body)` - important thing is that you can reuse auth and retry logic. A more advanced solution would allow you to extend the SDK (e.g., to enable `medusa.myResource.list`).
3. (Maybe look into realtime things - e.g., EventSourcing from a workflow subscription) this is a future thing, but would be really cool to have eventually.

kasperkristensen:
> 1. Should `class Medusa` export the `store` endpoints as a flat list (how it was originally) or under a `store` field (how it is now in this PR)? There are pros and cons with both approaches, so we need to decide what we prefer. Alternatively, we can export 2 separate `MedusaAdmin` and `MedusaStore` classes, and they will only contain either admin or store endpoints but not both.

Don't have any strong opinion this. But for the sake of getting all the arguments out there - I do agree with Seb that the majority of people will be using the SDK for storefront implementations. With that in mind returning them as a flat list is slightly easier to type out. On the other hand, I have seen a couple of issues/discord over the years of people not understanding why `medusa.regions.list()` was throwing a CORS error when they were building a admin application/extension. So the clarity of `.store/.admin` might be slightly better for readability. 

> 2. How should our wrapper of `fetch` behave? Should we convert to JSON by default? Should we throw exceptions on non-200 statuses (similar to Axios)? For now I kept the API as the standard `fetch`, but I do think we can provide a bit better experience than that.

I think we should convert to JSON by default, with an escape hatch if the user needs another format, since JSON conversion will be needed in 99% of cases. Also think we should do something like Axios for errors, as it makes for a better DX. We could also export a safeguard function similar to Axios' `axios.isAxiosError`.

+1 for allowing users to extend the SDK with their own domains, eg. `medusa.invoices.list()`, if you skim through old issues and requests you will see quite a few requests for this.

sradevski:
Hey @srindom @kasperkristensen thanks for the comments guys!

> Should class Medusa export the store endpoints as a flat list (how it was originally) or under a store field (how it is now in this PR)? There are pros and cons with both approaches, so we need to decide what we prefer. Alternatively, we can export 2 separate MedusaAdmin and MedusaStore classes, and they will only contain either admin or store endpoints but not both.

I was a bit concerned about what Kasper said where people mix the two up. Do you think it would be a better UX for both admin and store builders if we export `StoreClient` and `AdminClient` with pretty much identical constructors?

So if you are building a storefront, you'll do
`import {StoreClient} from "@medusajs/js-sdk"` instead of `import MedusaClient from "@medusajs/js-sdk"`?

This would be the only thing that gets impacted, but it will solve the concern that Kasper brought up, without compromising anything really. WDYT?


Otherwise:
1. Retries is something we will introduce, but not as part of this PR as it is just an enhancement. I will create tickets from the TODOs that are in the code of the SDK currently.
2. You can already access the "base" `fetch` even now, which will automatically have all headers and other things set up.
3. Streaming is definitely possible if the BE supports it, there are few routes we can take here but I feel like it is a discussion for another time
4. I will modify the current "base" `fetch` to throw on non-200 and add convert to JSON unless opted out from it.
5. Doesn't extending the client as `class MyClient extends StoreClient { mydomain: {}}` work as an interface? I can just ensure this works seamlessly

Let me know if you have any comments, but I think for now we have a consensus on how to proceed on the points I raised, except of how we should expose `Admin` and `Store`.

olivermrbl:
> This would be the only thing that gets impacted, but it will solve the concern that Kasper brought up, without compromising anything really. WDYT?

I think I am leaning toward having:
```
medusa.store.regions.list()
medusa.admin.regions.list()
medusa.agents.list()
medusa.request("POST", "/agents", ...)
```

This is not a strong opinion, so I am keen on another approach, too. But I think this gives us a clear separation between store and admin APIs, AND it ships a single entry point, which can be nifty in case you want to wrap it in a context, similar to what we did with `medusa-react.` I also personally just think it's a better DX to only have a single client and then have autocompletion help you along the way.

packages/core/js-sdk/package.json
```diff
@@ -0,0 +1,40 @@
...
+  "version": "0.0.1",
+  "description": "SDK for the Medusa API",
+  "main": "dist/index.js",
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/medusajs/medusa",
+    "directory": "packages/core/js-sdk"
+  },
+  "engines": {
+    "node": ">=16"
```
kasperkristensen: q: this would need to >=18 right?

packages/core/js-sdk/src/client.ts
```diff
@@ -0,0 +1,183 @@
...
+  // If we both requested JSON, we try to parse. Otherwise, we return the raw response.
+  const isJsonRequest = reqHeaders.get("accept")?.includes("application/json")
+  return isJsonRequest ? await resp.json() : resp
+}
+
+export class FetchError extends Error {
+  status: number | undefined
+
+  constructor(message: string, status?: number) {
+    super(`Status ${status}: ${message}`)
```
kasperkristensen: suggestion: I think it would be better to just have the clean error message here,  what status a request failed with is irrelevant to the end user, they just want to see "Missing field 'id'" when the error is shown in a toast or something else.

packages/core/js-sdk/package.json
sradevski: @kasperkristensen yes, good catch 👍

packages/core/js-sdk/src/client.ts
sradevski: @kasperkristensen ah yes that is true. I added a ticket for us to actually return a JSON response so we can eg. localize the errors, but I agree with you for the time being.

olivermrbl (approved):
LGTM, pending a few more approvals from discussion participants

kasperkristensen (approved):
LGTM.

Might be added later, but what are we doing with the `auth` namespace?

sradevski:
@kasperkristensen we will indeed add it later, and that can be its own scope.

But basically the Fetch Client will have a couple of methods to manage the lifecycle of authentication, I just didn't go into that yet, I want to get the basic setup in place first

sradevski:
@srindom I'll get this merged now, but if you have any comments please leave them and I will address them in another PR.

PR #7278, opened 2024-05-08
docs: added file module docs

shahednasser:
- Added docs about what a file module is
- Added docs about local file module
- Generated reference for the file provider module as a how-to guide (still need to actually add the docblocks within the source code for the reference to be meaningful)


shahednasser:

olivermrbl (approved):

PR #7282, opened 2024-05-09
feat: Add the basic implementation of notification module

sradevski:
As a follow-up I will implement at least 1 provider that we can test with, and add some end-to-end tests

### DIFF
--- a/packages/core/utils/src/notification/abstract-notification-provider.ts
+++ b/packages/core/utils/src/notification/abstract-notification-provider.ts
@@ -6,6 +6,10 @@ export class AbstractNotificationProviderService
   async send(
     notification: NotificationTypes.ProviderSendNotificationDTO
   ): Promise<NotificationTypes.ProviderSendNotificationResultsDTO> {
-    throw Error("send must be overridden by the child class")
+    throw Error(
+      `send is not implemented in ${
+        Object.getPrototypeOf(this).constructor.name
+      }`
+    )
   }
 }
--- a/packages/modules/notification/src/loaders/providers.ts
+++ b/packages/modules/notification/src/loaders/providers.ts
@@ -86,6 +86,8 @@ async function syncDatabaseProviders({
     }
   })
 
+  validateProviders(normalizedProviders)
+
   try {
     const providersInDb = await providerService.list({})
     const providersToDisable = providersInDb.filter(
@@ -113,4 +115,18 @@ async function syncDatabaseProviders({
   } catch (error) {
     logger.error(`Error syncing the notification providers: ${error.message}`)
   }
+}
+
+function validateProviders(providers: { channels: string[] }[]) {
+  const hasForChannel = {}
+  providers.forEach((provider) => {
+    provider.channels.forEach((channel) => {
+      if (hasForChannel[channel]) {
+        throw new Error(
+          `Multiple providers are configured for the same channel: ${channel}`
+        )
+      }
+      hasForChannel[channel] = true
+    })
+  })
 }
--- a/packages/modules/notification/src/models/notification.ts
+++ b/packages/modules/notification/src/models/notification.ts
@@ -103,7 +103,7 @@ export default class NotificationModel {
   @BeforeCreate()
   @OnInit()
   onCreate() {
-    this.id = generateEntityId(this.id, "not")
+    this.id = generateEntityId(this.id, "noti")
     this.provider_id ??= this.provider_id ?? this.provider?.id
   }
 }
--- a/packages/modules/notification/src/services/notification-module-service.ts
+++ b/packages/modules/notification/src/services/notification-module-service.ts
@@ -85,9 +85,7 @@ export default class NotificationModuleService<
 
     const serialized = await this.baseRepository_.serialize<
       NotificationTypes.NotificationDTO[]
-    >(createdNotifications, {
-      populate: true,
-    })
+    >(createdNotifications)
 
     return Array.isArray(data) ? serialized : serialized[0]
   }
--- a/packages/core/types/src/notification/service.ts
+++ b/packages/core/types/src/notification/service.ts
@@ -68,14 +68,14 @@ export interface INotificationModuleService extends IModuleService {
    *
    * ```ts
    * const notification =
-   *   await notificationModuleService.retrieve("not_123")
+   *   await notificationModuleService.retrieve("noti_123")
    * ```
    *
    * To specify relations that should be retrieved:
    *
    * ```ts
    * const notification = await notificationModuleService.retrieve(
-   *   "not_123",
+   *   "noti_123",
    *   {
    *     relations: ["provider"],
    *   }
@@ -103,7 +103,7 @@ export interface INotificationModuleService extends IModuleService {
    *
    * ```ts
    * const notifications = await notificationModuleService.list({
-   *   id: ["not_123", "not_321"],
+   *   id: ["noti_123", "noti_321"],
    * })
    * ```
    *
@@ -112,7 +112,7 @@ export interface INotificationModuleService extends IModuleService {
    * ```ts
    * const notifications = await notificationModuleService.list(
    *   {
-   *     id: ["not_123", "not_321"],
+   *     id: ["noti_123", "noti_321"],
    *   },
    *   {
    *     relations: ["provider"],
@@ -125,7 +125,7 @@ export interface INotificationModuleService extends IModuleService {
    * ```ts
    * const notifications = await notificationModuleService.list(
    *   {
-   *     id: ["not_123", "not_321"],
+   *     id: ["noti_123", "noti_321"],
    *   },
    *   {
    *     relations: ["provider"],
@@ -157,7 +157,7 @@ export interface INotificationModuleService extends IModuleService {
    * ```ts
    * const [notifications, count] =
    *   await notificationModuleService.listAndCount({
-   *     id: ["not_123", "not_321"],
+   *     id: ["noti_123", "noti_321"],
    *   })
    * ```
    *
@@ -167,7 +167,7 @@ export interface INotificationModuleService extends IModuleService {
    * const [notifications, count] =
    *   await notificationModuleService.listAndCount(
    *     {
-   *       id: ["not_123", "not_321"],
+   *       id: ["noti_123", "noti_321"],
    *     },
    *     {
    *       relations: ["provider"],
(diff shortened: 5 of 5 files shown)


packages/core/types/src/notification/providers/local.ts
```diff
@@ -0,0 +1 @@
+export interface LocalNotificationServiceOptions {}
```
sradevski, line 1: This is just a placeholder, I'll add the interface of some of the providers we will support out of the box

packages/modules/notification/integration-tests/__tests__/notification-module-service/index.spec.ts
```diff
@@ -0,0 +1,50 @@
...
+            name: "Test provider",
+            channels: ["email"],
+          },
+        },
+      },
+    },
+  ],
+}
+
+moduleIntegrationTestRunner({
```
sradevski, line 28: I will add more tests once I add a provider

packages/modules/notification/src/models/notification.ts
```diff
@@ -0,0 +1,94 @@
...
+  @Property({ columnType: "text", nullable: true })
+  resource_id?: string | null
+
+  // The typeame of the resource this notification is for, if applicable, eg. "order"
+  @Property({ columnType: "text", nullable: true })
+  resource_type?: string | null
+
+  // The ID of the customer this notification is for, if applicable.
+  @Property({ columnType: "text", nullable: true })
+  customer_id?: string | null
```
carlos-r-l-rodrigues: Maybe we could have an index for `customer_id`. I can foresee an useful link between Notification and Customer for the admin.

packages/modules/notification/src/services/notification-module-service.ts
```diff
@@ -0,0 +1,135 @@
...
+          provider,
+          entry
+        )
+        return { ...entry, provider_id: provider.id, external_id: res.id }
+      })
+    )
+
+    // Currently we store notifications after they are sent, which might result in a notification being sent that is not registered in the database.
+    // If necessary, we can switch to a two-step process where we first create the notification, send it, and update it after it being sent.
+    const createdNotifications = await this.notificationService_.create(
```
carlos-r-l-rodrigues, line 146: I believe we could save the notification before sending it. And after calling the provider we update its status.
This will have to be broken down into 3 different steps, because at the moment we are holding a DB transaction open while calling a 3rd party provider to send the notification.

.changeset/moody-days-drive.md
```diff
@@ -0,0 +1,8 @@
+---
+"@medusajs/modules-sdk": minor
+"@medusajs/types": minor
+"@medusajs/utils": minor
```
olivermrbl, line 4: nit:
```suggestion
"@medusajs/modules-sdk": patch
"@medusajs/types": patch
"@medusajs/utils": patch
```

packages/modules/notification/src/models/notification.ts
sradevski: @carlos-r-l-rodrigues I am actually thinking now, should this even be a field in the module? It might be better represented as a link, don't you think?

packages/modules/notification/src/services/notification-module-service.ts
sradevski: @carlos-r-l-rodrigues The reason I thought we can maybe avoid doing 2 writes is because it can be quite resource-intensive without a huge benefit. On the transaction being open that's a good point, I'll create a ticket as I want to get this merged first so I can get it to a working state with providers.

packages/modules/notification/src/models/notification.ts
sradevski: We discussed with Oli and decided to rename this to `receiver_id` instead as a more generic term. I also added an index for it.

packages/core/types/src/product/service.ts
```diff
@@ -79,7 +79,7 @@ export interface IProductModuleService extends IModuleService {
   ): Promise<ProductDTO>
 
   /**
-   * This method is used to retrieve a paginated list of price sets based on optional filters and configuration.
+   * This method is used to retrieve a paginated list of products based on optional filters and configuration.
```
adrien2p, line 82: 🙏 ahah nice catch

packages/core/utils/src/notification/abstract-notification-provider.ts
```diff
@@ -0,0 +1,11 @@
+import { NotificationTypes, INotificationProvider } from "@medusajs/types"
+
+export class AbstractNotificationProviderService
+  implements INotificationProvider
+{
+  async send(
+    notification: NotificationTypes.ProviderSendNotificationDTO
+  ): Promise<NotificationTypes.ProviderSendNotificationResultsDTO> {
+    throw Error("send must be overridden by the child class")
```
adrien2p: suggestion: 
```ts
throw Error(`send is not implemented in ${Object.getPrototypeOf(this).constructor.name}`)
```

packages/modules/notification/src/loaders/providers.ts
```diff
@@ -0,0 +1,116 @@
+import { moduleProviderLoader } from "@medusajs/modules-sdk"
```
adrien2p, line 1: note: We will have to rework this as we have it in 3 places now 👍

packages/modules/notification/src/migrations/Migration20240509083918_InitialSetupMigration.ts
```diff
@@ -0,0 +1,29 @@
+import { Migration } from "@mikro-orm/migrations"
+
+export class Migration20240509083918_InitialSetupMigration extends Migration {
+  async up(): Promise<void> {
+    this.addSql("drop table if exists notification")
+    this.addSql("drop table if exists notification_provider")
```
adrien2p, line 6: q: are we loosing any data from v1 if someone runs it in the same database? I suppose this has been validated elsewhere but just raising the question

packages/modules/notification/src/services/notification-module-service.ts
```diff
@@ -0,0 +1,155 @@
...
+  > {
+    const normalized = Array.isArray(data) ? data : [data]
+
+    const createdNotifications = await this.create_(normalized, sharedContext)
+
+    const serialized = await this.baseRepository_.serialize<
+      NotificationTypes.NotificationDTO[]
+    >(createdNotifications, {
+      populate: true,
+    })
```
adrien2p: nit: you can remove populate true, it is the default

packages/modules/notification/src/services/notification-provider.ts
```diff
@@ -0,0 +1,64 @@
...
+    if (!this.providersCache) {
+      const providers = await this.notificationProviderRepository_.find()
+      this.providersCache = new Map(
+        providers.flatMap((provider) =>
+          provider.channels.map((c) => [c, provider])
+        )
+      )
+    }
+
+    return this.providersCache.get(channel)
```
adrien2p, line 54: q: can't we have multiple providers for a single channel?

packages/modules/notification/src/loaders/providers.ts
sradevski: @adrien2p oh there is so much to improve around loading providers, but I didn't want to do it as part of this PR

packages/modules/notification/src/migrations/Migration20240509083918_InitialSetupMigration.ts
sradevski: @adrien2p our current setup doesn't support any migrations from v1 so I added this as a quick way to have v2 running (as we have other places where we eg. drop columns). I think we'll review all migrations before anyone attempts any sort of migration 😄

packages/modules/notification/src/services/notification-provider.ts
sradevski: @adrien2p I think it is better if we don't for the time being. We can always loosen this restriction, but for now I'd rather not push provider selection to the caller, as it introduces unnecessary complexity for almost all of the users.

We already discussed this in the RFC if you want to check that out

packages/modules/notification/src/loaders/providers.ts
adrien2p: yes definately, it is more a note that a suggestion :)

packages/modules/notification/src/services/notification-provider.ts
adrien2p: I do agree with that, but we might as well throw an error if it occurs by mistakes? wdyt?

sradevski: @adrien2p yup agreed, I'll add a check in the loader :+1

packages/modules/notification/src/joiner-config.ts
```diff
@@ -0,0 +1,33 @@
+import { Modules } from "@medusajs/modules-sdk"
+import { ModuleJoinerConfig } from "@medusajs/types"
+import { MapToConfig } from "@medusajs/utils"
+import { NotificationModel } from "@models"
+
+export const LinkableKeys: Record<string, string> = {
```
olivermrbl, line 6: suggestion(non-blocking): Add `receiver_id`

packages/modules/notification/src/models/notification.ts
```diff
@@ -0,0 +1,109 @@
...
+    onCreate: () => new Date(),
+    columnType: "timestamptz",
+    defaultRaw: "now()",
+  })
+  created_at: Date
+
+  @BeforeCreate()
+  @OnInit()
+  onCreate() {
+    this.id = generateEntityId(this.id, "not")
```
olivermrbl: very-nit(non-blocking): Use `noti` to avoid a clash with other entities, e.g. a Note with `not`.

olivermrbl (approved):
LGTM, nice work!

.changeset/moody-days-drive.md
sradevski: @olivermrbl Since these 3 are already at a 1.x version, based on semantic versioning they should be a minor. Should we follow a different logic instead?

packages/modules/notification/src/joiner-config.ts
sradevski: Let me create a ticket for it instead just to get PRs moving faster

packages/modules/notification/src/models/notification.ts
sradevski: Sounds good 👍

.changeset/moody-days-drive.md
olivermrbl: Actually, it doesn't matter much, as long as the changes land on `develop`.  We just need to stay away from major for the time being, and we should be fine.

PR #7285, opened 2024-05-10
chore: cleanup admin function

riqwan:


thetutlage:
Just for my curiosity. Don't we want to setup Admin panel when creating a new app, or we are setting it up in a different way and this is some old setup code?

riqwan:
@thetutlage thats right, this is legacy. Kasper is working on a shiny way. :)

thetutlage:
Cool!

adrien2p (approved):

PR #7286, opened 2024-05-10
chore(utils): Provide a mikro orm base entity

adrien2p:
**What**
Provide a new mikro orm base entity in order to abstract away the on init and on create which include the id auto generation if not provided. The main goal is to reduce complexity in entity definition

### DIFF
--- a/packages/admin-next/dashboard/public/locales/en-US/translation.json
+++ b/packages/admin-next/dashboard/public/locales/en-US/translation.json
@@ -1086,6 +1086,16 @@
         }
       }
     },
+    "promotions": {
+      "remove": {
+        "title": "Remove promotion from campaign",
+        "description": "You are about to remove {{count}} promotion(s) from the campaign. This action cannot be undone."
+      },
+      "alreadyAdded": "This promotion has already been added to the campaign.",
+      "toast": {
+        "success": "Successfully added {{count}} promotion(s) to campaign"
+      }
+    },
     "deleteCampaignWarning": "You are about to delete the campaign {{name}}. This action cannot be undone.",
     "totalSpend": "<0>{{amount}}</0> <1>{{currency}}</1>"
   },
--- a/packages/admin-next/dashboard/src/hooks/api/campaigns.tsx
+++ b/packages/admin-next/dashboard/src/hooks/api/campaigns.tsx
@@ -1,6 +1,7 @@
 import {
   AdminCampaignListResponse,
   AdminCampaignResponse,
+  LinkMethodRequest,
 } from "@medusajs/types"
 import {
   QueryKey,
@@ -14,6 +15,7 @@ import { queryClient } from "../../lib/medusa"
 import { queryKeysFactory } from "../../lib/query-key-factory"
 import { CreateCampaignReq, UpdateCampaignReq } from "../../types/api-payloads"
 import { CampaignDeleteRes } from "../../types/api-responses"
+import { promotionsQueryKeys } from "./promotions"
 
 const REGIONS_QUERY_KEY = "campaigns" as const
 export const campaignsQueryKeys = queryKeysFactory(REGIONS_QUERY_KEY)
@@ -82,7 +84,7 @@ export const useUpdateCampaign = (
     mutationFn: (payload) => client.campaigns.update(id, payload),
     onSuccess: (data, variables, context) => {
       queryClient.invalidateQueries({ queryKey: campaignsQueryKeys.lists() })
-      queryClient.invalidateQueries({ queryKey: campaignsQueryKeys.detail(id) })
+      queryClient.invalidateQueries({ queryKey: campaignsQueryKeys.details() })
 
       options?.onSuccess?.(data, variables, context)
     },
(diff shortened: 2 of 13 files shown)


adrien2p:
cc @shahednasser @olivermrbl does that correspond to what you had in mind?

adrien2p:
if we agree on the approach, would it make sense to move our models to use it or not? My thinking is that it is not a necessity for us to use it as we have already defined all our hooks. In that case we dont have to refactor right now and we can punt it to later if we want. It would be more to improve the DX as mentioned and asked by shahed :D

packages/core/utils/src/dal/mikro-orm/base-entity.ts
```diff
@@ -0,0 +1,71 @@
...
+  BeforeCreate,
+  Entity,
+  OnInit,
+  OptionalProps,
+  PrimaryKey,
+} from "@mikro-orm/core"
+import { generateEntityId } from "../../common"
+
+@Entity({ abstract: true })
+export class MikroOrmBaseEntity {
```
shahednasser: Would it make sense to just name it `BaseEntity`? What's the purpose of mentioning `MikroORM` in the name?

adrien2p: Just to avoid confusion if someone want to use a different orm in their module

shahednasser: But this just for some cases right? In general cases most people would use what we provide. For those that use a different ORM they can import it with a different name.

Not a big issue though, I just think since this is something that we're providing developers with (not something used internally) we don't need to be too explicit with the name. Easier names are easier to remember 😄

adrien2p: Yes I agree, my thinking was that people dont think they can use it with prisma for example, but in any case they will surely have an error to warn them, i did it the same way we prefixed our other utils for mikro orm. If you think it is better without I can remove it, just confirm me 😃

shahednasser: Not sure about other utils, but let's just name this one `BaseEntity` 👍🏻

adrien2p: 👍 '

adrien2p: done

olivermrbl:
I generally like this. We can use this going forward!

> My thinking is that it is not a necessity for us to use it as we have already defined all our hooks

We will have to visit the DX of our data model definition in the near future, so I would not refactor anything right now 👍

shahednasser (approved):
Cool stuff @adrien2p 🚀

packages/core/utils/src/dal/mikro-orm/base-entity.ts
```diff
@@ -0,0 +1,69 @@
...
+     */
+    const words = ensuredPrefixId
+      .split(/(?=[A-Z])/)
+      .filter((word) => !["entity", "model"].includes(word.toLowerCase()))
+    const wordsLength = words.length
+
+    /*
+     * if the class name (camel case) contains one word, the prefix id is the first three letters of the word
+     * if the class name (camel case) contains two words, the prefix id is the first two letters of the first word plus the first letter of the second one
+     * if the class name (camel case) contains more than two words, the prefix id is the first letter of each word
```
sradevski, line 47: I am not sure how readable this case would be, but I guess is fine since you can override it anyway.

packages/core/utils/src/dal/mikro-orm/__tests__/base-entity.spec.ts
```diff
@@ -0,0 +1,73 @@
...
+
+    const manager = orm.em.fork()
+    const entity1 = manager.create(Entity1, {})
+
+    expect(entity1.id).toMatch(/ent_[0-9]/)
+
+    await orm.close()
+  })
+
+  it("should handle the id generation without a provided prefix inferring it based on the words composing the entity name excluding model and entity as part of the name", async () => {
```
sradevski, line 45: Let's add a test with a model that has custom `OnInit` and `OnCreate` hooks, and ensure the ID is still generated

packages/core/utils/src/dal/mikro-orm/base-entity.ts
```diff
@@ -0,0 +1,69 @@
...
+  onInit() {
+    this.id ??= this.generateEntityId(this.__prefix_id__)
+  }
+
+  private generateEntityId(prefixId?: string): string {
+    let ensuredPrefixId: string = prefixId as string
+
+    if (ensuredPrefixId) {
+      return generateEntityId(undefined, ensuredPrefixId)
+    }
```
sradevski:
```suggestion
    if(prefixId) {
      return generateentityId(undefined, prefixId)
    }
```

packages/core/utils/src/dal/mikro-orm/base-entity.ts
```diff
@@ -0,0 +1,69 @@
...
+
+    ensuredPrefixId = Object.getPrototypeOf(this).constructor.name
```
sradevski:
```suggestion
   const ensuredPrefixId = Object.getPrototypeOf(this).constructor.name
```

packages/core/utils/src/dal/mikro-orm/base-entity.ts
```diff
@@ -0,0 +1,69 @@
...
+
+    /*
+     * Split the class name (camel case) into words and exclude model and entity from the words
+     */
+    const words = ensuredPrefixId
+      .split(/(?=[A-Z])/)
```
sradevski, line 40: Might be insignificant, but it would be nice to cache this calculation so we don't have to run regex every time an entry is created.

packages/core/utils/src/dal/mikro-orm/base-entity.ts
adrien2p: yes thats the idea, it is mainly for a user that does not want to provide a prefix

packages/core/utils/src/dal/mikro-orm/base-entity.ts
adrien2p: done 👍

packages/core/utils/src/dal/mikro-orm/__tests__/base-entity.spec.ts
adrien2p: done 👍

sradevski (approved):
Nice 👍

PR #7287, opened 2024-05-10
fix: change generate OAS v1 action to be based on v1.x branch

shahednasser:
Generating the OAS for v1.x requires the `packages/medusa/src/api` directory to be available, which is only available on the `v1.x` branch.

This PR changes the action that generates the OAS to checkout the `v1.x` branch when generating the reference, but still open the PR against the `develop` branch (with changes only in the `www/apps/api-reference/specs` directory)

### DIFF
--- a/.github/workflows/generate-public-references.yml
+++ b/.github/workflows/generate-public-references.yml
@@ -24,6 +24,7 @@ jobs:
         with:
           token: ${{ secrets.REFERENCE_PAT }}
           fetch-depth: 0
+          ref: "v1.x"
 
       - name: Setup Node.js environment
         uses: actions/setup-node@v3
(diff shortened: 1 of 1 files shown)


olivermrbl:
Can we remove the requirement on this branch instead?

shahednasser:
If you mean hosting the v1 OAS only on `v1.x`, that would require two different projects (in Vercel) to be created for the API reference. While that's not an issue (and might be the better approach to separate them completely), this would mean that in the long run there will be a big deviation in the projects such as their configurations, design, etc...

I'm fine with both approaches, so if you think this is the better approach then I'll look Monday into separating the projects (it will need some reconfiguration in certain areas)

PR #7289, opened 2024-05-10
fix(medusa): Missing middlewares export

adrien2p:


sradevski (approved):

carlos-r-l-rodrigues (approved):

adrien2p:
@sradevski did the tests pass in your pr?

packages/medusa/src/loaders/helpers/routing/index.ts
```diff
@@ -665,7 +665,7 @@ export class RoutesLoader {
         /**
          * Require the admin to be authenticated
          */
-        this.router.use(descriptor.route, authenticate())
+        this.router.use(descriptor.route, authenticateLegacy())
```
olivermrbl, line 668: question: Is this correct? I would expect no V1-related code to still be relevant in the core

adrien2p: It is temporary and will be migrated later, this pr was to fix the authenticate middleware not being exported to unblock shahed.

olivermrbl: Got it!

PR #7290, opened 2024-05-10
feat: Add support for sendgrid and logger notification providers

sradevski:


packages/modules/providers/notification-logger/src/services/logger.ts
```diff
@@ -0,0 +1,48 @@
...
+        `No notification information provided`
+      )
+    }
+
+    const message =
+      `Attempting to send a notification to: ${notification.to}` +
+      ` on the channel: ${notification.channel} with template: ${notification.template}` +
+      ` and data: ${JSON.stringify(notification.data)}`
+
+    this.logger_.info(message)
```
riqwan: Nothing to do for now, but if this is for logging, i think we should define a generic json template for log operations. Its much easier to view and faster to query json in apps like splunk.

Something like:
```
{
  class: "LoggerNotificationService",
  method: "send",
  data: {
    channel: notification.channel,
    template: notification.template,
    data: notification.data
  }
}
```

packages/modules/providers/notification-sendgrid/jest.config.js
```diff
@@ -0,0 +1,16 @@
+module.exports = {
+  globals: {
+    "ts-jest": {
+      tsconfig: "tsconfig.spec.json",
+      isolatedModules: false,
+    },
+  },
+  transform: {
+    "^.+\\.[jt]s?$": "ts-jest",
```
riqwan, line 9: transpile with swc-jest for any new projects, it speeds up the tests by so much!

packages/modules/providers/notification-sendgrid/src/services/sendgrid.ts
```diff
@@ -0,0 +1,73 @@
...
+        | { [key: string]: any }
+        | undefined,
+    }
+
+    try {
+      // Unfortunately we don't get anything useful back in the response
+      await sendgrid.send(message)
+      return {}
+    } catch (error) {
+      const errorCode = error.code
```
riqwan, line 63: should we update the notification status upon error?

packages/modules/providers/notification-stdout/tsconfig.json
```diff
@@ -0,0 +1,32 @@
...
+  },
+  "include": ["../notification-console/src"],
+  "exclude": [
+    "../notification-console/dist",
+    "../notification-console/build",
+    "../notification-console/src/**/__tests__",
+    "../notification-console/src/**/__mocks__",
+    "../notification-console/src/**/__fixtures__",
+    "../notification-console/node_modules",
+    "../notification-console/.eslintrc.js"
```
riqwan: why is this checking another package? 🤔

riqwan (approved):
some comments, but overall lgtm!

packages/modules/providers/notification-logger/src/services/logger.ts
riqwan: nvm, just saw that this is literally a provider that logs. Should we rename the package to notification-local/system? Similar to file/payment/shipping?

packages/modules/providers/notification-stdout/tsconfig.json
sradevski: @riqwan nice catch, I just renamed the package and something must have happened with my git, this needs to be removed (done!)

packages/modules/providers/notification-logger/src/services/logger.ts
sradevski: @riqwan sounds good, I can rename it to `local` 👍

packages/modules/providers/notification-sendgrid/jest.config.js
sradevski: Will do, bad copy-pasta practice 🍝

packages/modules/providers/notification-sendgrid/src/services/sendgrid.ts
sradevski: Currently we will just return an error, but you have a good point, we probably still want to store the notification, but with a status. I'll think about it and get this added 👍

sradevski: @riqwan I created a ticket as it will affect things outside of this PR, but I think it's a good idea to add a status to the ntofication model.

olivermrbl:
This is great 💪

PR #7291, opened 2024-05-10
chore(): Autoload module resources

adrien2p:
**What**
- automatically build and consume connection and container loader if not exported by the module
- therefore load the services and repositories automatically, including baseRepository
- automatically build run and revert migrations if not provided
- cleaup modules to remove extra unnecessary bits and pieces
- remove the `initializeFactory` in favor of using `medusaApp`

Should drastically improve the module building DX by removing a lot of boilerplate to handle by the user, that plus the base entity should simplify quite a lot the flow cc @shahednasser 

**Note**
I had to choose a way to identify connection and container loader from the exported loader from the module. I decided to go with named function `connectionLoader` and `containerLoader`, also, now the factories will return named function so if the user use the factories we are providing to build those loaders, the function will also be named and identified

### DIFF
--- a/packages/modules/sales-channel/integration-tests/__tests__/services/sales-channel-module.spec.ts
+++ b/packages/modules/sales-channel/integration-tests/__tests__/services/sales-channel-module.spec.ts
@@ -2,242 +2,241 @@ import { SqlEntityManager } from "@mikro-orm/postgresql"
 
 import { ISalesChannelModuleService } from "@medusajs/types"
 
-import { initialize } from "../../../src"
-
 import { createSalesChannels } from "../../__fixtures__"
-import { DB_URL, MikroOrmWrapper } from "../../utils"
+import { moduleIntegrationTestRunner, SuiteOptions } from "medusa-test-utils"
+import { Modules } from "@medusajs/modules-sdk"
 
 jest.setTimeout(30000)
 
-describe("Sales Channel Service", () => {
-  let service: ISalesChannelModuleService
-  let testManager: SqlEntityManager
-  let repositoryManager: SqlEntityManager
+moduleIntegrationTestRunner({
+  moduleName: Modules.SALES_CHANNEL,
+  testSuite: ({
+    MikroOrmWrapper,
+    medusaApp,
+  }: SuiteOptions<ISalesChannelModuleService>) => {
+    let service: ISalesChannelModuleService
 
-  beforeEach(async () => {
-    await MikroOrmWrapper.setupDatabase()
-    repositoryManager = await MikroOrmWrapper.forkManager()
-
-    service = await initialize({
-      database: {
-        clientUrl: DB_URL,
-        schema: process.env.MEDUSA_SALES_CHANNEL_DB_SCHEMA,
-      },
+    beforeEach(() => {
+      service = medusaApp.modules[Modules.SALES_CHANNEL]
     })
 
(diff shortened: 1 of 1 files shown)


packages/core/modules-sdk/src/loaders/utils/load-internal.ts
```diff
@@ -155,8 +173,229 @@ export async function loadModuleMigrations(
...
+   */
+  const connectionLoaderName = "connectionLoader"
+  const containerLoader = "containerLoader"
+
+  const hasConnectionLoader = loadedModuleLoaders.some(
+    (l) => l.name === connectionLoaderName
+  )
+
+  if (!hasConnectionLoader && models.length > 0) {
+    const connectionLoader = ModulesSdkUtils.mikroOrmConnectionLoaderFactory({
```
olivermrbl, line 363: would be good to get rid of MikroORM here eventually

packages/modules/auth/src/index.ts
```diff
@@ -1,11 +1,11 @@
...
-  runMigrations,
-} from "./module-definition"
+import { moduleDefinition } from "./module-definition"
+import { initializeFactory, Modules } from "@medusajs/modules-sdk"
 
 export default moduleDefinition
-export { revertMigration, runMigrations }
 
-export * from "./initialize"
+export const initialize = initializeFactory({
```
olivermrbl: very nitpicky, but I think the `Factory` suffix leaks an implementation detail; I liked initialize better.

olivermrbl: If we can't use initialize, maybe we could go with `initializeModule`?

olivermrbl (commented):
At first glance, this is an incredible improvement – nice work!! 

Imo, we can push this through, just want to keep it open for others to glance over it too

olivermrbl:
Next up is the module service boilerplate 😉

packages/core/modules-sdk/src/loaders/utils/load-internal.ts
adrien2p: This is the default connection loader, since we are using mikro orm by default. Could you elaborate on your idea?

packages/modules/auth/src/index.ts
adrien2p: This will disappear in the next commits. Basically we have everything to spin up module through medusaApp. So the module does not need to export the initialize and therefore the initialize will not even need to exists anymore 👍

adrien2p:
> Next up is the module service boilerplate 😉 

Yes, it will be the last one, then only the custom loaders will remain to export if you have any. This is because we need to keep the order in which they are exported as it might have an impact. Maybe we can think of something there too 👍

packages/modules/auth/src/index.ts
adrien2p: Also, i haven't changed the naming from what it was, initialize (for now) is still exported and the factory is not leaking outside of the module itself. Can you clarify on it please ☺️

olivermrbl: > This will disappear in the next commits. Basically we have everything to spin up module through medusaApp

Got it, even better!

adrien2p:
@olivermrbl should be good to re review if you want 👍

olivermrbl:
Will do!

olivermrbl (approved):
LGTM

adrien2p:
@shahednasser just for info it is merged now

PR #7295, opened 2024-05-10
fix: Remove medusa v2 flag from the medusa app package

sradevski:


packages/medusa/src/commands/migrate.js
```diff
@@ -76,30 +74,19 @@ const main = async function ({ directory }) {
...
-
-      await runLinkMigrations(directory)
-    }
+    const container = createMedusaContainer()
+    const pgConnection = await pgConnectionLoader({ configModule, container })
+    container.register({
+      [ContainerRegistrationKeys.CONFIG_MODULE]: asValue(configModule),
+      [ContainerRegistrationKeys.LOGGER]: asValue(Logger),
+      [ContainerRegistrationKeys.PG_CONNECTION]: asValue(pgConnection),
+      featureFlagRouter: asValue(featureFlagRouter),
```
adrien2p: suggestion: ContainerRegistrationKeys.FEATURE_FLAG_ROUTER

packages/medusa/src/commands/migrate.js
```diff
@@ -76,30 +74,19 @@ const main = async function ({ directory }) {
...
+    })
+    await migrateMedusaApp(
+      { configModule, container },
+      { registerInContainer: false }
+    )
     process.exit()
-
-    Logger.info("Migrations completed.")
```
adrien2p, line 102: suggestion: we can maybe re enable the log and put it before the process exit?

packages/medusa/src/commands/user.js
```diff
@@ -61,25 +24,34 @@ export default async function ({
...
-    const configModule = configModuleLoader(directory)
-    const featureFlagRouter = featureFlagLoader(configModule)
+    const userService = container.resolve(ModuleRegistrationName.USER)
+    const authService = container.resolve(ModuleRegistrationName.AUTH)
+    const provider = "emailpass"
 
-    if (featureFlagRouter.isFeatureEnabled(MedusaV2Flag.key)) {
-      await useV2Command({ email, password, isInvite: invite }, { container })
+    if (invite) {
+      // The invite flow only works with the V2 version of packages/admin-next/dashboard, so enable the V2 feature flag in admin before using this command
```
adrien2p: todo: you can rm the comment I believe

packages/medusa/src/loaders/index.ts
```diff
@@ -1,292 +1,97 @@
...
-
-      if (module.default?.models) {
-        module.default.models.map((model) =>
-          container.registerAdd("db_entities", asValue(model))
-        )
-      }
-    }
-  }
+const isWorkerMode = (configModule) => {
+  return configModule.projectConfig.worker_mode === "worker"
```
adrien2p, line 28: suggestion: outside of the scope of this pr, but I believe the start mode should be part of a constant of an enum what do you think, we are using it in multiple packages but we are string based and it can be prone to error

packages/medusa/src/loaders/medusa-app.ts
```diff
@@ -178,26 +175,24 @@ export const loadMedusaApp = async (
...
-        )
-      }
-    }
-
-    if (missingPackages.length) {
-      throw new Error(
-        `FeatureFlag medusa_v2 (MEDUSA_FF_MEDUSA_V2) requires the following packages/module registration: (${missingPackages.join(
-          ", "
-        )})`
+    if (!isModuleInstalled) {
```
adrien2p, line 182: todo: I believe we can refactor this part, the required module here is static and only contains product and pricing, but we already have this info in the module definition normally. Maybe not now in this pr but we can add a todo for that

packages/medusa/src/loaders/medusa-app.ts
```diff
@@ -178,26 +175,24 @@ export const loadMedusaApp = async (
...
+      missingPackages.push(
+        MODULE_PACKAGE_NAMES[requiredModuleKey] || requiredModuleKey
       )
     }
   }
 
+  if (missingPackages.length) {
+    throw new Error(
+      `FeatureFlag medusa_v2 (MEDUSA_FF_MEDUSA_V2) requires the following packages/module registration: (${missingPackages.join(
```
adrien2p: todo: we can remove the mention of the flag I believe since it work only for v2 now

packages/medusa/src/utils/middlewares/authenticate-customer.ts
```diff
@@ -1,34 +1,10 @@
...
 import { NextFunction, Request, RequestHandler, Response } from "express"
 
-import passport from "passport"
-
 // Optional customer authentication
 // If authenticated, middleware attaches customer to request (as user) otherwise we pass through
 // If you want to require authentication, use `requireCustomerAuthentication` in `packages/medusa/src/api/middlewares/require-customer-authentication.ts`
 export default (): RequestHandler => {
   return (req: Request, res: Response, next: NextFunction): void => {
-    const featureFlagRouter = req.scope.resolve(
```
adrien2p, line 11: Q: should we put a todo to do the necessary for v2?

packages/medusa/src/utils/middlewares/authenticate.ts
```diff
@@ -1,19 +1,7 @@
...
-    const featureFlagRouter = req.scope.resolve(
-      ContainerRegistrationKeys.FEATURE_FLAG_ROUTER
-    )
-    if (featureFlagRouter.isFeatureEnabled(MedusaV2Flag.key)) {
-      return next()
-    }
-    passport.authenticate(
-      ["admin-session", "admin-bearer", "admin-api-token"],
-      { session: false }
-    )(req, res, next)
```
adrien2p, line 17: Q: I believe this will make some tests to fail as well as the customer auth middleware?

adrien2p: I can see that there is not test failing, but I believe we have routing tests that check for 401, how comes they are passing? :D

packages/medusa/src/loaders/medusa-app.ts
sradevski: Adding a comment, I think doing it in a separate PR makes more sense.

packages/medusa/src/utils/middlewares/authenticate-customer.ts
sradevski: @adrien2p I can add yeah, although it is obvious that we need to make auth work for v2 😄

packages/medusa/src/utils/middlewares/authenticate.ts
sradevski: No idea 😄 I added a todo, let's tackle this in a separate PR (I mean auth in general)

packages/medusa/src/loaders/index.ts
sradevski: @adrien2p yeah I agree, but I'll leave it for another PR 👍

adrien2p (approved):

Issue #7263, opened 2024-05-07
`npx medusa develop` ignores `serve: false` from medusa-config.js


sonntag-philipp:
### Describe the bug

Starting the admin dashboard and the server in their own processes does not work as [stated in the documentation](https://docs.medusajs.com/admin/configuration#admin-cli). 

If I understand the documentation correctly, this configuration:
```javascript
{
    resolve: "@medusajs/admin",
    options: {
      serve: false,
      autoRebuild: false,
      develop: {
        open: process.env.OPEN_BROWSER !== "false",
      },
   },
},
```
should not start the Webpack dev server that is used for development. But that is exactly what happens.

### System information

Medusa version (including plugins): v1.20.6
Node.js version: v20.12.1
Database: Postgres v10.4 (Docker Container)
Operating system: maxOS Sonoma 14.3
Browser (if relevant): ---

### Steps to reproduce the behavior

1. `npx create-medusa-app@latest`
2. Configure database etc. without Next.js storefront
3. Stop the already running dev server
4. Open `medusa-config.js`
5. Configure as described in the additional context
    - Make sure `autoRebuild: false` and `serve: false` is set
6. Execute `yarn run dev`

--> The Webpack dev server ist started when executing `yarn run dev`

### Expected behavior

Do not start the Webpack dev server for the Admin Dashboard if serve is set to false in the medusa-config.js. This allows the user to start the dev server as it's own process like [the documentation states](https://docs.medusajs.com/admin/configuration#develop-command-options).

### Screenshots

If applicable, add screenshots to help explain your problem

### Code snippets

If applicable, add code samples to help explain your problem

#### medusa-config.js
```javascript
const dotenv = require("dotenv");

let ENV_FILE_NAME = "";
switch (process.env.NODE_ENV) {
  case "production":
    ENV_FILE_NAME = ".env.production";
    break;
  case "staging":
    ENV_FILE_NAME = ".env.staging";
    break;
  case "test":
    ENV_FILE_NAME = ".env.test";
    break;
  case "development":
  default:
    ENV_FILE_NAME = ".env";
    break;
}

try {
  dotenv.config({ path: process.cwd() + "/" + ENV_FILE_NAME });
} catch (e) {}

// CORS when consuming Medusa from admin
const ADMIN_CORS = process.env.ADMIN_CORS || "http://localhost:7000,http://localhost:7001";

// CORS to avoid issues when consuming Medusa from a client
[...]

PR #7299, opened 2024-05-11
fix: Abort the admin `develop` command if serve is `false`

adevinwild:
**Issue**
Based on this [issue](https://github.com/medusajs/medusa/issues/7263), the admin ui continues to start with the `serve` property set to `false`

**Solution**
In the `packages/medusa/src/commands/develop.js` I retrieved the config of the `@medusajs/admin` plugin, in case `serve` is set to `false` the `adminCLI`  process is not runned

> Note
> I don't know if it's the best way to handle it though


adevinwild:
Hi @riqwan, 
Do you think we should keep this fix?

riqwan (approved):
I think this is a good change. @kasperkristensen wdyt?

PR #7302, opened 2024-05-12
Avoid returning draft-orders from /orders endpoint

josetr:
I'm assuming the current behavior of returning all kind of orders is not the right one since we already have a /draft-orders endpoint.


carlos-r-l-rodrigues (approved):
Thank you

PR #7303, opened 2024-05-12
Feat(order): return reason and transaction methods

carlos-r-l-rodrigues:


packages/core/core-flows/src/return-reason/workflows/create-return-reasons.ts
```diff
@@ -0,0 +1,18 @@
+import {
+  CreateOrderReturnReasonDTO,
+  OrderReturnReasonDTO,
+} from "@medusajs/types"
+import { WorkflowData, createWorkflow } from "@medusajs/workflows-sdk"
+import { createReturnReasonsStep } from "../steps"
+
+type WorkflowInput = { customersData: CreateOrderReturnReasonDTO[] }
```
olivermrbl:
```suggestion
type WorkflowInput = { data: CreateOrderReturnReasonDTO[] }
```

olivermrbl (approved):
LGTM

PR #7307, opened 2024-05-13
docs: update api-reference project for v2

shahednasser:
- Make v2 the default version of api-reference
- Set base path of V2 API reference to `/v2`. For example, `/v2/api/admin`.
- Remove all versioning-related snippets and functionalities except for the version switcher
- Change version switcher to only take the user to v1 API reference
- Move rewrites to book project, making it the main routing project for paths starting with `/v2`.
- Normalize environment variable names to match across projects.

Note: renaming `specs-v2` directory to `specs` makes the number of file changes in this PR big, as that deletes v1 specs


shahednasser:

olivermrbl (approved):

PR #7310, opened 2024-05-13
Validate product and collection handles to be URL safe

thetutlage:
This PR validates the Product and the Collection handles to be URL safe using a Regex. Consult the following screenshot to see what is allowed

Screenshot 2024-05-13 at 4 42 32 PM

--- 

## Update

Ended up moving the normalization and validation both to the service. Also tested manually with the admin panel and following are the results.

### A title with `@` in it generates correct handle

Screenshot 2024-05-14 at 11 58 37 AM

### The backend returns error when handle it manually updated to contain invalid characters.

However, in this case frontend isn't showing errors. Something should be handled in a separate issue.

Screenshot 2024-05-14 at 11 59 12 AM

### DIFF
--- a/packages/modules/product/src/services/product-module-service.ts
+++ b/packages/modules/product/src/services/product-module-service.ts
@@ -1168,9 +1168,12 @@ export default class ProductModuleService<
   ): Promise<TProduct[]> {
     const normalizedInput = await promiseAll(
       data.map(async (d) => {
-        await this.normalizeCreateProductInput(d, sharedContext)
-        this.validateProductPayload(d)
-        return d
+        const normalized = await this.normalizeCreateProductInput(
+          d,
+          sharedContext
+        )
+        this.validateProductPayload(normalized)
+        return normalized
       })
     )
 
@@ -1228,9 +1231,12 @@ export default class ProductModuleService<
   ): Promise<TProduct[]> {
     const normalizedInput = await promiseAll(
       data.map(async (d) => {
-        await this.normalizeUpdateProductInput(d, sharedContext)
-        this.validateProductPayload(d)
-        return d
+        const normalized = await this.normalizeUpdateProductInput(
+          d,
+          sharedContext
+        )
+        this.validateProductPayload(normalized)
+        return normalized
       })
     )
 


thetutlage:
This is my first PR. So please feel free to familiarise me to the process and let me know if something needs to changed or done in a different way.

Also, I wasn't table to find any tests for the validators (maybe I looked at wrong places). So please guide me on how/where to write tests

packages/medusa/src/api-v2/admin/collections/validators.ts
```diff
@@ -29,13 +29,19 @@ export const AdminGetCollectionsParams = createFindParams({
 export type AdminCreateCollectionType = z.infer<typeof AdminCreateCollection>
 export const AdminCreateCollection = z.object({
   title: z.string(),
-  handle: z.string().optional(),
+  handle: z
+    .string()
+    .regex(/^[a-z0-9]+(?:-[a-z0-9]+)*$/)
+    .optional(),
```
riqwan: sugg: we have a `utils/common-validators/` folder to add common validators where you can add this, I'd assume this might be needed for categories and other `handle` requiring endpoints as well.

riqwan (approved):
💪🏻

srindom:
Good work with the validators - I'd expect these checks to happen further down the stack too. 

E.g., calling `productModuleService.create({ handle: "not allowed-cuz'-bad chars" })` should throw. 

Also, there is some logic somewhere that constructs a handle from the title, not sure if it's in the admin code or somewhere else. Currently, this code doesn't remove disallowed characters. E.g. `title: "Summer '24"` will result in `handle: "summer-'24"`. This logic should be modified too.

packages/medusa/src/api-v2/admin/collections/validators.ts
thetutlage: Indeed makes sense!

thetutlage:
> Good work with the validators - I'd expect these checks to happen further down the stack too.

Yeah, I had this conversation on the Slack too. I did not find any sort of schema validations happening in the services, so didn't feel comfortable to put it there.

But, if the goal is to keep formatting validations at the service layer. I can surely move the logic there as a manual validation. Or should we use Zod within services as-well?

thetutlage:
> Also, there is some logic somewhere that constructs a handle from the title, not sure if it's in the admin code or somewhere else. Currently, this code doesn't remove disallowed characters. E.g. title: "Summer '24" will result in handle: "summer-'24". This logic should be modified too.

Yeah, makes sense. The creation of handle should perform the normalization

packages/medusa/src/api-v2/utils/common-validators/common.ts
```diff
@@ -30,3 +30,11 @@ export const OptionalBooleanValidator = z.preprocess(
   (val: any) => optionalBooleanMapper.get(val?.toLowerCase()),
   z.boolean().optional()
 )
+
+/**
+ * Validates entity handle to have URL-safe characters
+ */
+export const HandleValidator = z
```
adrien2p: sugg: I think it is safe to have this validation in the product service, if we receiev a handle then we can normalize it otherwise we are already creating a handle ourselves as part of a entity hook if I am correct. Nice work though 👍

thetutlage: >  if we receiev a handle then we can normalize it otherwise we are already creating a handle ourselves as part of a entity hook if I am correct.

Should we normalize user submitted `handle`, or should we validate it?

packages/modules/product/src/services/product-module-service.ts
```diff
@@ -1315,9 +1316,13 @@ export default class ProductModuleService<
...
-    if (!productData.handle && productData.title) {
-      productData.handle = kebabCase(productData.title)
-    }
+    /**
+     * We are already computing handle from title in model. Do we
+     * need here again?
+     */
+    // if (!productData.handle && productData.title) {
+    //   productData.handle = kebabCase(productData.title)
+    // }
```
thetutlage: We are already computing handle from title in model. Do we need here again?

packages/modules/product/src/services/product-module-service.ts
```diff
@@ -1339,6 +1344,13 @@ export default class ProductModuleService<
       productData.discountable = false
     }
 
+    if (productData.handle && !isValidHandle(productData.handle)) {
+      throw new MedusaError(
+        MedusaError.Types.INVALID_DATA,
+        "Invalid product handle. It must contain URL safe characters"
+      )
+    }
+
```
thetutlage: Is this the right place to validate and throw a custom error?

thetutlage:
I ended up changing the implementation

- Create common helpers to generate and validate handles.
- Remove Zod validations and handle it at the service.

Let me know if the implementation looks right. If yes, I will repeat it for collections and other entities as well.

packages/medusa/src/api-v2/utils/common-validators/common.ts
adrien2p: Just validating the user submission and the normalisation can happen in the module i would say. Wdyt?

packages/modules/product/src/services/product-module-service.ts
adrien2p: I would say that we can do it in the service and remove it from the model. My thinking is that in the future the DAL will be replacable, i believe having the feature not tight to the model would be beneficial

thetutlage: Sure. I don't have any strong opinions :) So if this the only place that performs CREATE and UPDATE, then we can put the logic in the service directly.

packages/core/utils/src/common/to-handle.ts
```diff
@@ -0,0 +1,18 @@
...
+
+/**
+ * Helper method to create a to be URL friendly "handle" from
+ * a string value.
+ *
+ * - Works by converting the value to lowercase
+ * - Splits and remove accents from characters
+ * - Removes all unallowed characters like a '"%$ and so on.
+ */
+export const toHandle = (value: string): string => {
```
adrien2p, line 11: Suggestion : to handle looks a bit tight to a product specific knowledge. It kind of leaks an internal detail from a domain. Would it make sense to name it differently? I dont have a name in my head straight away though 🥹

packages/core/utils/src/common/validate-handle.ts
```diff
@@ -0,0 +1,9 @@
+import { kebabCase } from "./to-kebab-case"
+
+/**
+ * Helper method to validate entity "handle" to be URL
+ * friendly.
+ */
+export const isValidHandle = (value: string): boolean => {
```
adrien2p, line 7: Same suggestion as before

packages/core/utils/src/common/to-handle.ts
thetutlage: Isn't having handles a product wide thing for us? Or are these utils too generic that they should not be tied to Medusa at all?

thetutlage:
Anyone knows what is causing this error in integration tests?

Screenshot 2024-05-14 at 12 03 12 PM

packages/core/utils/src/common/to-handle.ts
adrien2p: Lets keep it as is for now and we can circle back on that naming later

adrien2p:
> Anyone knows what is causing this error in integration tests?
> 
> Screenshot 2024-05-14 at 12 03 12 PM

I believe this might be due to the upert with replace, @sradevski what do you think? I believe the reason is that an obj ref might be assigned at different level creating a circular ref on jsonifying it

thetutlage:
I got the error fixed after this commit https://github.com/medusajs/medusa/pull/7310/commits/ae4d9a7b92c09fb6f3380b0c8e68e835682dd1a5

Seems like removing the handle generation from the model hook breaks something and the error stack isn't able to point at it. 

I am not sure if I should remove the model hook or not. Because we have the same logic in service also and I remove the logic from service, then some other parts of the application break.

riqwan:
@thetutlage The CI seems to pass, might be an issue with an outdated package build. Can you try running `yarn build` from the root and try again?

thetutlage:
@riqwan I shared the reasoning about the CI failing earlier in this comment. https://github.com/medusajs/medusa/pull/7310#issuecomment-2109842281

packages/modules/product/src/services/product-module-service.ts
```diff
@@ -1339,6 +1341,13 @@ export default class ProductModuleService<
       productData.discountable = false
     }
 
+    if (productData.handle && !isValidHandle(productData.handle)) {
```
sradevski: I would create a new `validateUpdateProductInput` method, and move the validations there instead, and then we call it basically everywhere where we call `normalize*`.

thetutlage: > I would create a new validateUpdateProductInput method

This makes sense. Should this be a method on the model?

> and then we call it basically everywhere where we call normalize*.

Should we validate after normalizing? Because the normalization process is supposed to create valid handles in first place.

adrien2p: I believe it should be part of the service instead of the models, it create too much responsibilities for the model and if a user replace it then the validation needs to be re implemented. It can also be a util in the module itself that you consume in the service, since the function will not be tight to anything from the class

the place depends, if the handle is provided, then there might be a normalization step and it still needs to be validated right?

thetutlage: > the place depends, if the handle is provided, then there might be a normalization step and it still needs to be validated right?

I have been thinking it the other way around. Lemme share my thoughts.

- If handle is provided. It will only be validated. **Why not normalize?** Usually, its not a good idea to change the value of something a user has manually entered without any visual feedback. So if a custom handle is provided, it is only validated.
- The normalization (I think the wording is a bit off here). That is why I called it `toHandle`. This method is responsible for generate a URL-safe handle from the entity title. This method should be called when no `handle` is provided.

So in a nutshell. There is no normalization. There is

- Validation (user provided value)
- URL-safe handle generation (when no value is provided)

sradevski: @thetutlage we should validate after normalizing yes. The normalization, as Adrien said, will not create a valid handle if it is passed by the user, so the validation still makes sense.

thetutlage: Alright. Jumped on a call with @sradevski and everything is clear now.

We were talking about two different functions when said `normalize`. He meant the `normalizeUpdateProductInput` or `normalizeCreateProductInput` functions and I was thinking of my `toHandle` method as normalization function.

Basically, mis-communication. We are all good and I just have to abstract the validation to its own method within the service

adrien2p: no problem, sounds right 👍 , that happens sometimes especially when it is non verbal communication 😄

packages/modules/product/src/services/product-module-service.ts
```diff
@@ -1339,6 +1356,8 @@ export default class ProductModuleService<
       productData.discountable = false
     }
 
+    this.validateProductHandle(product)
```
sradevski: I think it's better to call the `validate*` outside of `normalize*`, WDYT? Alternatively we call validate at the end of the function, not in the middle of it. Also, maybe we can name it more generically, eg. `validateCreateProduct` and we will add more validations in it as we go.

sradevski (approved):

packages/modules/product/src/services/product-module-service.ts
thetutlage: Yeah, it indeed makes sense to keep validation outside of normalization. So I am going to move it to `create_` and `update_` methods.

> Also, maybe we can name it more generically, eg. validateCreateProduct and we will add more validations in it as we go.

Should we keep it even more generic like `validateProductPayload`? Because (atleast as of now) we are validating handle for both the actions.

sradevski: @thetutlage yeah makes sense 👍

adrien2p (approved):
Just need to fix the pipeline 💪 good work

sradevski (approved):

jackyfkj:
This PR essentially removes all the non-alphanumeric characters in the handle, which I think is too strict.

> handle: A slug representation of the product’s title. This can be used in the storefront to generate human-readable URLs for products, which can be beneficial for Search Engine Optimization (SEO) purposes. https://docs.medusajs.com/modules/products

I have noticed that the URL of some e-commerce sites can contain non-alphanumeric characters.  For example:

image

Given that, I think it is better to allow saving non-alphanumeric characters in the handle.

maybe we just do not need this line `re.sub(r"[^a-z0-9-_]", "", value)`?

jackjyq:
I am wondering if we need the handle to be URL-safe, for example:

We got the product URL `https://shopee.tw/恐龙-i.59046599.23945502411` which contains Chinese characters, we could use `恐龙-i.59046599.23945502411` as handle and let the frontend to encode the URL to `https://shopee.tw/%E6%81%90%E9%BE%99-i.59046599.23945502411`.

In this PR, the handle removes all non-alphanumeric characters, so we can not use `恐龙-i.59046599.23945502411`.

I suggest to remove this line `replace(/[^a-z0-9A-Z-_]/g, "")`

sradevski:
@jackjyq thanks for bringing it up. I agree, we went with a very restrictive check for URL-safety. Although we do want to handle this in the BE, we probably want to revisit the Regex we use for the check.

I'll open a ticket internally to track this and tackle it before the public v2 release.

PR #7314, opened 2024-05-13
feat(order): receive return and additional calculations

carlos-r-l-rodrigues:


packages/modules/order/src/services/order-module-service.ts
```diff
@@ -2238,8 +2238,8 @@ export default class OrderModuleService<
       return {
         action: ChangeActionType.RETURN_ITEM,
         internal_note: item.internal_note,
-        reference: data.reference,
-        reference_id: shippingMethodId,
+        reference: data.reference ?? "fulfillment",
+        reference_id: data.reference_id ?? shippingMethodId,
```
olivermrbl, line 2242: question: Why do we default to`shippingMethodId` for the `RETURN_ITEM` action

olivermrbl (approved):
A single question, but otherwise LGTM

packages/modules/order/src/services/order-module-service.ts
carlos-r-l-rodrigues: This reference is generic, and could be related to any "reference" and its ID.
We haven't yet created a "return" table. But I believe this will be referenced there.

PR #7319, opened 2024-05-14
feat(core-flows, types): Create return order

adrien2p:
**what**
- Create return workflow partial implementation
- Update some order domain types
- create order fulfillment link

**NOTE**
this PR is partially done but can still be merged as is, it will require some discussions around the flow and some unknowns or uncertainty in regards to some data and some behaviour

### DIFF
--- a/packages/core/core-flows/src/order/workflows/create-return.ts
+++ b/packages/core/core-flows/src/order/workflows/create-return.ts
@@ -192,7 +192,7 @@ function prepareFulfillmentData({
   if (!locationId) {
     throw new MedusaError(
       MedusaError.Types.INVALID_DATA,
-      `No location id provided and no stock location found in the fulfillment set of the shipping option with id ${returnShippingOption.id}.`
+      `Cannot create return without stock location, either provide a location or you should link the shipping option ${returnShippingOption.id} to a stock location.`
     )
   }
 


adrien2p:
@olivermrbl if you want you can start have a look to this PR and we can discuss the different points

packages/core/core-flows/src/common/steps/create-remote-links.ts
```diff
@@ -0,0 +1,29 @@
+import { LinkDefinition, RemoteLink } from "@medusajs/modules-sdk"
+import { createStep, StepResponse } from "@medusajs/workflows-sdk"
+
+import { ContainerRegistrationKeys } from "@medusajs/utils"
+
+type CreateRemoteLinksStepInput = LinkDefinition[]
+
+export const createRemoteLinkStepId = "create-remote-links"
```
olivermrbl:
```suggestion
export const createLinksStepId = "create-links"
```

packages/core/core-flows/src/order/workflows/create-return.ts
```diff
@@ -0,0 +1,305 @@
...
+import { createReturnStep } from "../steps/create-return"
+import { createFulfillmentWorkflow } from "../../fulfillment"
+
+function throwIfOrderIsCancelled({ order }: { order: OrderDTO }) {
+  // TODO: need work, check cancelled
+  if (false /*order.cancelled_at*/) {
+    throw new MedusaError(
+      MedusaError.Types.INVALID_DATA,
+      `Order with id ${order.id} has been cancelled.`
+    )
```
olivermrbl:
```suggestion
  if (false /*order.canceled_at*/) {
    throw new MedusaError(
      MedusaError.Types.INVALID_DATA,
      `Order with id ${order.id} has been canceled.`
    )
```

packages/core/core-flows/src/order/workflows/create-return.ts
```diff
@@ -0,0 +1,305 @@
...
+  }
+}
+
+function throwIfItemsDoesNotExistsInOrder({
```
olivermrbl, line 37: thought: This could probably be a more generic helper – I imagine we will need this elsewhere

packages/core/core-flows/src/order/workflows/create-return.ts
```diff
@@ -0,0 +1,305 @@
...
+    )
+  }
+}
+
+function validateReturnReasons(
+  {
+    order_id,
+    inputItems,
+  }: {
+    order_id: string
```
olivermrbl:
```suggestion
    orderId,
    inputItems,
  }: {
    orderId: string
```

packages/core/core-flows/src/order/workflows/create-return.ts
```diff
@@ -0,0 +1,305 @@
...
+  const returnReasons = remoteQuery({
+    entry_point: "return_reasons",
+    fields: ["return_reason_children.*"],
+    variables: { id: [inputItems.map((item) => item.reason_id)] },
+  })
+
+  const reasons = returnReasons.map((r) => r.id)
+  const hasUnusableReasons = reasons.filter(
+    (reason) => reason.return_reason_children.length === 0
+  )
```
olivermrbl: Can you add a comment in the code here explaining why this makes them unusable. I don't think that's immediately obvious

packages/core/core-flows/src/order/workflows/create-return.ts
```diff
@@ -0,0 +1,305 @@
...
+  const hasUnExistingReasons = arrayDifference(reasonIds, reasons)
+
+  if (hasUnExistingReasons.length) {
+    throw new MedusaError(
+      MedusaError.Types.INVALID_DATA,
+      `Return reason with id ${hasUnExistingReasons.join(
+        ", "
+      )} does not exists.`
+    )
+  }
```
olivermrbl:
```suggestion
  const hasNonExistingReasons = arrayDifference(reasonIds, reasons)

  if (hasNonExistingReasons.length) {
    throw new MedusaError(
      MedusaError.Types.INVALID_DATA,
      `Return reason with id ${hasNonExistingReasons.join(
        ", "
      )} does not exists.`
    )
  }
```

packages/core/core-flows/src/order/workflows/create-return.ts
```diff
@@ -0,0 +1,305 @@
...
+  const remoteQuery = container.resolve(ContainerRegistrationKeys.REMOTE_QUERY)
+
+  const returnReasons = remoteQuery({
+    entry_point: "return_reasons",
+    fields: ["return_reason_children.*"],
+    variables: { id: [inputItems.map((item) => item.reason_id)] },
+  })
+
+  const reasons = returnReasons.map((r) => r.id)
+  const hasUnusableReasons = reasons.filter(
```
olivermrbl: suggestion: Rename to `hasInvalidReasons`

packages/core/core-flows/src/order/workflows/create-return.ts
```diff
@@ -0,0 +1,305 @@
...
+
+function prepareShippingMethodData({
+  orderId,
+  inputShippingOption,
+  returnShippingOption,
+}: {
+  orderId: string
+  inputShippingOption: OrderWorkflow.CreateOrderReturnWorkflowInput["return_shipping"]
+  returnShippingOption: ShippingOptionDTO & {
+    calculated_price: { calculated_amount: number }
```
olivermrbl: suggestion: At some point, we should introduce a type `WithCalculatedPrice` that we can apply to all the places where we currently type it out explicitly. Think that will prove handy.

packages/core/core-flows/src/order/workflows/create-return.ts
```diff
@@ -0,0 +1,305 @@
...
+function validateCustomRefundAmount({
+  order,
+  refundAmount,
+}: {
+  order: Pick<OrderDTO, "item_total">
+  refundAmount?: number
+}) {
+  // validate that the refund prop input is less than order.item_total (item total)
+  // TODO: Probably this amount should be retrieved from the payments linked to the order
+  if (refundAmount && MathBN.gt(refundAmount, order.item_total)) {
```
olivermrbl, line 150: qusetion: We don't have a `refundable_amount` on the order? If not, I think we should consider it

packages/core/core-flows/src/order/workflows/create-return.ts
```diff
@@ -0,0 +1,305 @@
...
+      sku: orderItem.variant_sku || "",
+      barcode: orderItem.variant_barcode || "",
+    } as FulfillmentWorkflow.CreateFulfillmentItemWorkflowDTO
+  })
+
+  // TODO: should we get the stock location from location id and grab the address?
+
+  return {
+    input: {
+      location_id: input.location_id, // TODO: Do we ask the user to provider this, or do we grab the stock location from the fulfillment set of the service zone link to the shipping option id
```
olivermrbl: I would say default to the fulfillment set location ID, and make it an optional property of the input, so merchants can pass it

olivermrbl (commented):
Nice work, I left some initial comments

packages/core/core-flows/src/order/workflows/create-return.ts
adrien2p: yes definately, for now they are local, but as soon as we have other needs then we can move them to utils wdyt?

packages/core/core-flows/src/order/workflows/create-return.ts
adrien2p: Yes I can backport the comment from v1 :D

packages/core/core-flows/src/order/workflows/create-return.ts
adrien2p: yes for now we don't cc @carlos-r-l-rodrigues I think we discuss it as the refundable amount is basically everything that has been paid minus refunded

packages/core/workflows-sdk/src/utils/composer/helpers/resolve-value.ts
```diff
@@ -41,9 +41,11 @@ export async function resolveValue(input, transactionContext) {
     }
 
     if (Array.isArray(inputTOUnwrap)) {
-      return await promiseAll(
+      const res = await promiseAll(
```
carlos-r-l-rodrigues: nit: that can be reverted to just `return`

carlos-r-l-rodrigues (approved):

packages/core/workflows-sdk/src/utils/composer/helpers/resolve-value.ts
adrien2p: oups debug session left over ahah

adrien2p:
@olivermrbl I ve addressed your comments and a bit more, if you want to have another look

packages/core/core-flows/src/order/workflows/create-return.ts
```diff
@@ -0,0 +1,326 @@
...
+
+  const returnReasons = remoteQuery({
+    entry_point: "return_reasons",
+    fields: ["return_reason_children.*"],
+    variables: { id: [inputItems.map((item) => item.reason_id)] },
+  })
+
+  const reasons = returnReasons.map((r) => r.id)
+  const hasInvalidReasons = reasons.filter(
+    // We do not allow for root reason to be applied
```
olivermrbl, line 84: Ahhh yes, that was why – haha, I had forgotten 😄

packages/core/core-flows/src/order/workflows/create-return.ts
```diff
@@ -0,0 +1,326 @@
...
+      )} does not exists.`
+    )
+  }
+
+  if (hasInvalidReasons.length()) {
+    throw new MedusaError(
+      MedusaError.Types.INVALID_DATA,
+      `Cannot apply return reason category with id ${hasInvalidReasons.join(
+        ", "
+      )} to order with id ${orderId}.`
```
olivermrbl: suggestion(non-blocking):
```suggestion
      `Cannot apply return reason with id ${hasInvalidReasons.join(
        ", "
      )} to order with id ${orderId}. Return reason has nested reasons.`
```

packages/core/core-flows/src/order/workflows/create-return.ts
```diff
@@ -0,0 +1,326 @@
...
+
+  let locationId: string | undefined = input.location_id
+  if (!locationId) {
+    locationId = returnShippingOption.service_zone.fulfillment_set.location?.id
+  }
+
+  if (!locationId) {
+    throw new MedusaError(
+      MedusaError.Types.INVALID_DATA,
+      `No location id provided and no stock location found in the fulfillment set of the shipping option with id ${returnShippingOption.id}.`
```
olivermrbl: suggestion(non-blocking): I would just do:
```suggestion
      `Cannot create return without stock location`
```

In theory, this shouldn't happen right? Because a shipping option can't live outside the context of a location.

olivermrbl (approved):
Overall LGTM, don't have much to add right now

packages/core/core-flows/src/order/workflows/create-return.ts
adrien2p: I also fixed the condition 😂

packages/core/core-flows/src/order/workflows/create-return.ts
adrien2p: Yes it shouldn't, but I rather prevent a case where the user didn't create the link between the shipping option and the stock location wdyt?

olivermrbl: Ah yes, that makes sense

PR #7321, opened 2024-05-14
docs: update events reference

shahednasser:
- Update events reference to have separate references for each commerce modules + a combined one
- Remove occurrences of old events from old services in code snippets


shahednasser:

PR #7323, opened 2024-05-14
Remove usage of inventory_quantity from the codebase

thetutlage:
This PR removes the `inventory_quantity` from the API layer includes the `product` module.

However, the same property is referenced in the frontend for filtering as-well. I think we will need a separate issue to fix that.


packages/modules/product/src/migrations/InitialSetup20240401153642.ts
```diff
@@ -12,88 +12,192 @@ export class InitialSetup20240315083440 extends Migration {
         `alter table "product_option_value" alter column "variant_id" drop not null;`
```
carlos-r-l-rodrigues, line 12: nit: we could remove `table_schema = 'public'` from line 6.

carlos-r-l-rodrigues (approved):

sradevski (approved):
LGTM! As we discussed, we now need to see what the new interface around inventory quantity will be, both for creation and filtering. Let's take that discussion separately.

packages/modules/product/src/migrations/InitialSetup20240401153642.ts
thetutlage: Should this change be done to all migration files?

PR #7324, opened 2024-05-14
fix(medusa): fix type of data in SubscriberArgs

shahednasser:
When an event is emitted using the `MessageAggregator`, the data received by the subscriber is of the format `{ data; metadata }`.

While I believe it's better to standardize the type of the data received by the subscriber (for example, by moving `metadata` to a separate property in the `SubscriberArgs`), this feels like a part of a bigger discussion.

For now, this PR fixes the type in `SubscriberArgs` to avoid type errors in subscribers.

### DIFF
--- a/packages/medusa/src/types/subscribers.ts
+++ b/packages/medusa/src/types/subscribers.ts
@@ -10,7 +10,17 @@ export type SubscriberConfig = {
 }
 
 export type SubscriberArgs<T = unknown> = {
-  data: T
+  data:
+    | T
+    | {
+        metadata: {
+          service: string
+          action: string
+          object: string
+          eventGroupId?: string
+        }
+        data: T
+      }
   eventName: string
   container: MedusaContainer
   pluginOptions: Record<string, unknown>


sradevski (approved):
LGTM for now, but as we are cleaning up we'll move these types to the types package instead and ensure all emitters follow them strictly

packages/medusa/src/types/subscribers.ts
```diff
@@ -10,7 +10,17 @@ export type SubscriberConfig = {
 }
 
 export type SubscriberArgs<T = unknown> = {
-  data: T
+  data:
```
adrien2p, line 13: suggestion: you can use `MessageBody<T>` from the types package

adrien2p (commented):
sorry a bit later :D

packages/medusa/src/types/subscribers.ts
shahednasser: I'll open another PR for it 👍🏻

PR #7330, opened 2024-05-14
feat(medusa,dashboard,admin-sdk): Run admin dashboard from Medusa instance

kasperkristensen:
**What**
- Adds the admin dashboard as a core feature of `@medusajs/medusa`. 

**How**
- `@medusajs/admin-sdk` is responsbile for injecting a Vite dev server in development, and injecting a route handler for serving the static admin build in production.
- It also builds the static files, which is currently run every time `medusa start` is executed. In the future we will want to move this to `medusa build`.
- Options for the admin can be passed in `medusa-config.js`:

```js
module.exports = {
  admin: {
    disable: false, // if `true` we disable the admin entirely in both dev and prod
    path: "/app", // the path to the admin
    outDir: "./build", // the path for static files to be built to.
    vite: (config) => { ... }, // update the default Vite config, not currently implemented, so will be ignored.
  },
  projectConfig,
  plugins,
  modules,
  featureFlags: {
    medusa_v2: true,
  },
};
```
  

**TODO**
This is the initial implementation, in future PR's we will need to add additional features, that have currently been disabled or are missing.

- Move the admin build process to the future `medusa build` command.
- Improve logic for determining whether to run the development or production build. Currently, based on whether `medusa start` was initiated by `medusa develop` or not.
- Re-enable Admin extensions
- Finalize `admin` config, this includes how we want to handle building for hosting the admin separately.
- Add some way to automatically open the Admin in development.

### DIFF
--- a/packages/medusa/src/commands/build.ts
+++ b/packages/medusa/src/commands/build.ts
@@ -118,7 +118,7 @@ export default async function ({ directory }: BuildArgs) {
   )
 
   if (error) {
-    console.log(`Failed to load medusa-confg.js`)
+    console.log(`Failed to load medusa-config.js`)
     process.exit(1)
   }
 


packages/medusa/src/commands/utils/open-admin.ts
```diff
@@ -0,0 +1,30 @@
...
+
+/**
+ * A function to open the admin in the users default browser.
+ * @param port - The port the server is running on
+ * @param rootDirectory - The root directory of the project
+ * @returns Attempts to open the admin in the users default browser.
+ */
+export async function openAdmin(port: number, rootDirectory: string) {
+  // If we are not in development mode, we skip opening the admin
+  if (process.env.NODE_ENV !== "development") {
```
sradevski: When do we actually want to automatically open the admin? Doing so in production also doesn't make sense, as the backend will probably be deployed by some pipeline and therefore it will run headlessly.

packages/medusa/src/loaders/index.ts
```diff
@@ -54,6 +55,8 @@ async function loadEntrypoints(
     next()
   })
 
+  await adminLoader({ app: expressApp, configModule })
```
sradevski, line 58: Let's load the admin at the end, after all of the backend had already loaded.

sradevski (commented):
I am thinking if we can possibly completely decouple the backend and admin here. What I have in mind is the following:
1. We have a `dev` command in package.json that does this:
    1. Runs `yarn develop` on `admin-next/dashboard`
    2. Runs `yarn develop` on `medusa`

    In this case, the BE doesn't serve the admin, as it would make little sense to do so

2. We have a `start` command that does this:
    1. Calls `yarn build` on `admin-next/dashboard`
    2. Runs `yarn build && yarn serve` on the BE. The BE simply serves a pre-configured static folder where the output from step 1. was placed


Is there anything I am missing that wouldn't allow for this setup?

packages/medusa/src/commands/utils/open-admin.ts
kasperkristensen: In dev only, but I have removed it for now, we need to find a better solution then the `open` package, as that one is not able to re-use tabs, and will open a new tab each time you restart your server.

packages/medusa/src/loaders/index.ts
kasperkristensen: Is there any reason for why you would prefer this? It's quite important that the middleware for serving the admin dev server does not get superceded by another middleware, loading it before anything else makes sure that can't happen.

sradevski: @kasperkristensen simply because the BE API might not be ready by the time the admin has built and loaded

kasperkristensen: Also we aren't really loading anything here (currently we are building the admin in prod, but once we have `medusa build` command we will do that there instead), we are just appending middleware and route handlers.

sradevski: Got it, if the build is not happening here and is just route registration then we're good 👍

kasperkristensen:
@sradevski I am not entirely sure I get you concern here, but I can give some context to why we moved the admin to a server plugin a bit more than a year ago.

- We used to have the admin as a separate project, but that caused a lot of issues when people did changes to the project and then needed to pull in upstream changes. Thats why we moved to the admin being served from the server, and then customizations were made in UI Widgets and Routes, this means that a user can update their admin without any issues, as the app and their customizations are two separate things. 
- The reason we need to serve a dev server in development is to enable hot reloading of UI Routes and Widgets, if we were serving the static build that obviously wouldn't work as its static. In V1 we ran the admin dev server and Medusa server in two separate processes. There isn't any issue with that and we could do that here as well, but this is a better DX. No matter if you are dev or prod your admin is always on `/app`, which just makes everything easier.
- In the future, we want to move the Admin to RSC, and serving things directly inside of Medusa will allow us to expose the Medusa container inside of the project. This would allow for some very nifty things like accessing the product module directly in a server action and do whatever custom thing they want instead of having to add a new endpoint just for that one component.
- Running things as you suggest would cause a lot of issues because, as I understand it, your suggestion is to ship the `dashboard` package as the unbuilt project. And then run `yarn build` on it from inside a medusa project, so `"build:admin": "vite ./node_modules/@medusajs/dashboard"`. The `medusa-default-starter` project is a CJS project and the dashboard is an ESM project, which will result in dependencies getting loaded incorrectly.

olivermrbl (approved):
Legend @kasperkristensen!

I presume I can give this a spin now in a Medusa project? Given that I pull in the correct packages.

kasperkristensen:
> Legend @kasperkristensen!
> 
> I presume I can give this a spin now in a Medusa project? Given that I pull in the correct packages.
> 
> Update: I slapped an "Approve" on it now, but would like to play around with it a bit before we merge

Yea, I am guessing you already have resolutions setup for V2 work, so just need to build the packages, and add:

```
"resolutions": {
    "@medusajs/admin-sdk": "file:../monorepo/packages/admin-next/admin-sdk",
    "@medusajs/admin-vite-plugin": "file:../monorepo/packages/admin-next/admin-vite-plugin",
    "@medusajs/admin-shared": "file:../monorepo/packages/admin-next/admin-shared",
    "@medusajs/dashboard": "file:../monorepo/packages/admin-next/dashboard",
    "@medusajs/medusa: "file:../monorepo/packages/medusa",

**Comment by olivermrbl** [comment]
> Yea, I am guessing you already have resolutions setup for V2 work, so just need to build the packages, and add:

Yup, lemme try it out 🚢

**Comment by riqwan** [comment]
Mad work @kasperkristensen ! 🎉

File: packages/medusa/src/commands/build.ts [hunk 3]
```diff
@@ -0,0 +1,156 @@
...
+export default async function ({ directory }: BuildArgs) {
+  const started = Date.now()
+
+  const { configModule, error } = getConfigFile<ConfigModule>(
+    directory,
+    "medusa-config"
+  )
+
+  if (error) {
+    console.log(`Failed to load medusa-confg.js`)
```
olivermrbl: nit:
```suggestion
    console.log(`Failed to load medusa-config.js`)
```

olivermrbl (commented):
Let's get this merged 🚢

PR #7331, opened 2024-05-14
feat: Add a simple configurable notifications subscriber

sradevski:


integration-tests/modules/__tests__/notification/admin/notification.spec.ts
```diff
@@ -26,102 +30,140 @@ medusaIntegrationTestRunner({
...
+            data: {
+              order: {
+                id: "1234",
+                email: "test@medusajs.com",
+              },
+            },
           })
-        )
+
+          // TODO: Obviously a random sleep isn't great here, opening the PR to see if others have any ideas on how we can wait for the subscriber.
```
sradevski: To people reviewing, this is the only open point that I'd like your opinion on. The current implementation of the local event bus doesn't allow for easily waiting for all subscribers to finish, so I am curious to hear what you think would be a nice way to be able to do that in tests.

packages/medusa/src/loaders/subscribers.ts
```diff
@@ -2,21 +2,24 @@ import glob from "glob"
...
 /**
  * Registers all subscribers in the subscribers directory
  */
 export default ({ container }: { container: MedusaContainer }) => {
-  const isTest = process.env.NODE_ENV === "test"
-
-  const corePath = isTest
-    ? "../subscribers/__mocks__/*.js"
-    : "../subscribers/*.js"
+  const corePath = "../subscribers/*.js"
```
sradevski, line 11: Having different implementations of the subscribers beats the purpose of testing them, so I'd rather have us find a good way to wait for the subscribers' completion.

integration-tests/modules/__tests__/notification/admin/notification.spec.ts
carlos-r-l-rodrigues: for sake of testing, if you subscribe again to "order.created" and that resolves a promise reference, you can maybe just await for it.

```
let resolve
const waitPromise = new Promise((ok) => resolve = ok)

// ...subscribe(..., resolve)

// ... emit

await waitPromise
```

sradevski: @carlos-r-l-rodrigues this doesn't work (I've also tried it) because the notification subscriber is async and it takes a bit of time to complete due to the DB ops, so the second subscriber will finish before the first one does.

integration-tests/modules/__tests__/notification/admin/notification.spec.ts
```diff
@@ -1,19 +1,23 @@
 import { ModuleRegistrationName } from "@medusajs/modules-sdk"
 import {
   CreateNotificationDTO,
+  IEventBusModuleService,
   INotificationModuleService,
   Logger,
 } from "@medusajs/types"
 import { ContainerRegistrationKeys } from "@medusajs/utils"
 import { medusaIntegrationTestRunner } from "medusa-test-utils"
+import { promisify } from "util"
```
adrien2p: nit: `import { setTimeout } from "timers/promises"`

integration-tests/modules/__tests__/notification/admin/notification.spec.ts
```diff
@@ -1,19 +1,23 @@
...
+
+export const sleep = promisify(setTimeout)
```
adrien2p: Q: why do we have a sleep here?

integration-tests/modules/__tests__/notification/admin/notification.spec.ts
```diff
@@ -26,102 +30,140 @@ medusaIntegrationTestRunner({
...
+            })
+          )
+
+          delete fromDB.original_notification_id
+          delete fromDB.external_id
+          delete fromDB.receiver_id
+          delete (fromDB as any).idempotency_key
+          delete (fromDB as any).provider
+
+          expect(result).toEqual(fromDB)
```
adrien2p: suggestion: wouldn't it be better to expect.objectContaining here instead of deleting from the obj above, my thinking is that reading the test, I don't know what to expect, I just know what will not be present

integration-tests/modules/__tests__/notification/admin/notification.spec.ts
```diff
@@ -26,102 +30,140 @@ medusaIntegrationTestRunner({
...
+          expect(logSpy).toHaveBeenCalledWith(
+            'Attempting to send a notification to: test@medusajs.com on the channel: email with template: order-created and data: {"username":"john-doe"}'
```
adrien2p: nit: should we put a comma between the info, we don't know where the infor starts and stops

integration-tests/modules/__tests__/notification/admin/notification.spec.ts
carlos-r-l-rodrigues: In that case I'm fine with the sleep for now.
It doesn't make much sense to have a feature to await subscribers.
and everything else will involve changing the pubsub, like emmiting "local" events when jobs start/finish proccessing.
which by the way maybe can be the same thing used to handle graceful shutdown in this module.

sradevski: @carlos-r-l-rodrigues we came up with a pretty nice solution with @adrien2p , have a look when you can

packages/core/medusa-test-utils/src/events.ts
```diff
@@ -0,0 +1,28 @@
...
+
+// Allows you to wait for all subscribers to execute for a given event. Only works with the local event bus.
+export const waitSubscribersExecution = (
+  eventName: string,
+  eventBus: IEventBusModuleService
+) => {
+  const subscriberPromises: Promise<any>[] = []
+
+  ;(eventBus as any).eventEmitter_.listeners(eventName).forEach((listener) => {
+    ;(eventBus as any).eventEmitter_.removeListener("order.created", listener)
```
adrien2p, line 11:
```suggestion
    ;(eventBus as any).eventEmitter_.removeListener(eventName, listener)
```

packages/core/medusa-test-utils/src/events.ts
```diff
@@ -0,0 +1,28 @@
...
+      ok = resolve
+      nok = reject
+    })
+    subscriberPromises.push(promise)
+
+    const newListener = async (...args2) => {
+      return await listener.apply(eventBus, args2).then(ok).catch(nok)
+    }
+
+    ;(eventBus as any).eventEmitter_.on("order.created", newListener)
```
adrien2p, line 24:
```suggestion
    ;(eventBus as any).eventEmitter_.on(eventName, newListener)
```

packages/core/medusa-test-utils/src/module-test-runner.ts
```diff
@@ -78,7 +78,7 @@ export function moduleIntegrationTestRunner({
   const moduleOptions_: InitModulesOptions = {
     injectedDependencies: {
       [ContainerRegistrationKeys.PG_CONNECTION]: connection,
-      eventBusService: new MockEventBusService(),
+      ["eventBusModuleService"]: new MockEventBusService(),
```
adrien2p, line 81: nit: ModuleRegistrationNames?

packages/core/modules-sdk/src/__mocks__/stock-location-module.ts
```diff
@@ -6,7 +6,7 @@ export const StockLocationModule = {
     label: "StockLocationService",
     isRequired: false,
     isQueryable: true,
-    dependencies: ["eventBusService"],
+    dependencies: ["eventBusModuleService"],
```
adrien2p, line 9: nit: ModuleRegistrationNames

packages/medusa/src/loaders/subscribers.ts
```diff
@@ -2,21 +2,24 @@ import glob from "glob"
...
 
   const core = glob.sync(coreFull, { cwd: __dirname })
   core.forEach((fn) => {
     const loaded = require(fn).default
+    if (!loaded) {
+      throw new MedusaError(
+        MedusaError.Types.UNEXPECTED_STATE,
+        `Subscriber ${fn} does not have a default export`
+      )
+    }
```
adrien2p, line 22: suggestion: I wouldn't necessarely throw here but maybe warn that in this file there is no subscribers found. My thinking is that you could want to have a file that is utils for example here wdyt?

packages/medusa/src/subscribers/configurable-notifications.ts
```diff
@@ -0,0 +1,64 @@
...
+        async (data: any) => {
+          const payload = data.data
+
+          const notificationData = {
+            template: eventHandler.template,
+            channel: eventHandler.channel,
+            to: get(payload, eventHandler.to),
+            trigger_type: eventHandler.event,
+            resource_id: get(payload, eventHandler.resource_id),
+            data: Object.entries(eventHandler.data).reduce(
```
adrien2p, line 47: todo: I would remove the use of lodash in that file

adrien2p (approved):
LGTM, few suggestion and little todo's 💪

packages/core/medusa-test-utils/src/events.ts
sradevski: Oh how did I miss that, I thought we changed it in your commit??

packages/core/medusa-test-utils/src/module-test-runner.ts
sradevski: @adrien2p the build was complaining when importing the SDK, so I left it as is for now

adrien2p: oh yeah true, it should be in the utils I believe but lets think that later

packages/core/modules-sdk/src/__mocks__/stock-location-module.ts
sradevski: @adrien2p I think these mocks will get removed anyway so it didn't matter

packages/core/medusa-test-utils/src/events.ts
adrien2p: I thought too 😂

adrien2p: have you forced push at some point? cause I ve lost a commit here

packages/medusa/src/loaders/subscribers.ts
sradevski: Hmm I am not sure TBH, I thought it's better to "explode" early as it can otherwise be missed. Let's leave it like this for now as it is more explicit, WDYT @adrien2p ?

packages/medusa/src/subscribers/configurable-notifications.ts
sradevski: @adrien2p a question on this, why don't we use lodash more?? 😄

adrien2p: because it is yet another dependency that we dont fully use and we only need a few utils that we can carry ourselves

packages/core/medusa-test-utils/src/events.ts
sradevski: yeah after the rebase, maybe that's why...

packages/medusa/src/subscribers/configurable-notifications.ts
adrien2p: or just install the get 😄

sradevski: I think we can use more of it if it was installed everywhere, I honestly don't see the point of reimplementing some of the things that are already available there, but we can discuss it some other time. For this one, since we already use lodash in a couple of other places in `medusa` I'll leave it as reimplementing this would be unnecessary work, WDYT?

adrien2p: but all the utils that we use from lodash can be done easily natively so unless we use complex utils I dont see the need to rely on a third party imo

adrien2p: lets do that for now, but imo we should go away from lodash, not the opposite

sradevski: @adrien2p let's create a ticket maybe and make sure we get rid of all references in that case? I'll merge this for now and we can do a cleanup afterwards in one go

packages/core/medusa-test-utils/src/events.ts
adrien2p: If I can recommend that you always use force with lease instead 😄

PR #7333, opened 2024-05-14
feat(dashboard): collection product management

fPolic:
**WHAT**
- add product management to ProductCollection details view
- upgrade things to v2
- enable collections in the sidebar

**NOTE**
- When we want to remove all products from the collection we pass ` [ { id: 'pcol_123', product_ids: [] } ]` to the `productModuleSerivce.updateCollections_` method. In that case update is not properly applied i.e. products are not disassociated from the collection.

---

https://github.com/medusajs/medusa/assets/16856471/e181a5e4-3b32-4133-ad66-a27b3fc05346


olivermrbl:
Hmm, the delete doesn't work on single product deletion either (I guess it tries to do the same as if we were to delete them all). I'll create a ticket for it so we can move forward with this one.

olivermrbl (approved):
LGTM!

fPolic:
> Hmm, the delete doesn't work on single product deletion either (I guess it tries to do the same as if we were to delete them all). I'll create a ticket for it so we can move forward with this one.

Yes it's the same case - whenever the collection is left without any products

PR #7335, opened 2024-05-14
feat(payment): update payment collection status

carlos-r-l-rodrigues:


packages/modules/payment/src/services/payment-module.ts
```diff
@@ -756,4 +796,74 @@ export default class PaymentModuleService<
       count,
     ]
   }
+
+  @InjectTransactionManager("baseRepository_")
```
sradevski, line 804: What's the default isolation level for the transaction? I am thinking we should probably lock the `paymentCollectionId`'s row, as we can end up with race conditions otherwise. We have this problem in a lot of places, but esp. around payments we need to be quite careful 😄

sradevski (approved):

packages/modules/payment/src/services/payment-module.ts
carlos-r-l-rodrigues: Good call. I added a serializable transaction and created a test to make sure the behavior is correct.
However this module needs to be refactored. It has the same error of calling a provider inside a transaction.
Will create a ticket for that

packages/modules/payment/integration-tests/__tests__/services/payment-module/index.spec.ts
```diff
@@ -685,6 +683,208 @@ moduleIntegrationTestRunner({
           //   )
           // })
         })
+
+        describe("concurrency", () => {
+          it.only("should authorize, capture and refund multiple payment sessions", async () => {
```
olivermrbl: question: remove only?

olivermrbl (approved):
Nice!

PR #7336, opened 2024-05-15
feat: Destroy session + introduce `http` config

olivermrbl:
Introduce
- `DELETE /auth/session` to destroy session, used for sign outs
- New `http` config in `projectConfig`*

*This PR introduces a breaking change by replacing some existing configs with an `auth` property in `projectConfig`. 

We will go from
```ts
projectConfig: {
  ...,
  jwt_secret: "test",
  cookie_secret: "test",
  auth_cors: "..."
}
```
to
```ts
projectConfig: {
  ...,
  http: {
    jwtSecret: "test",
    cookieSecret: "test",
    authCors: "...",
    storeCors: "...",
    adminCors: "..."
  }
}
```

Additionally, an additional config `jwtExpiresIn` is introduced to control the expiry time of the issued JWT tokens. Default is 24 hours.

In the future, we should consider supporting settings per scope and provider, but I don't think this is very important right now.

### DIFF
--- a/packages/medusa/src/loaders/helpers/routing/__fixtures__/mocks/index.ts
+++ b/packages/medusa/src/loaders/helpers/routing/__fixtures__/mocks/index.ts
@@ -5,12 +5,12 @@ export const storeGlobalMiddlewareMock = jest.fn()
 export const config = {
   projectConfig: {
     database_logging: false,
-  },
-  http: {
-    storeCors: "http://localhost:8000",
-    adminCors: "http://localhost:7001",
-    jwtSecret: "supersecret",
-    cookieSecret: "superSecret",
+    http: {
+      storeCors: "http://localhost:8000",
+      adminCors: "http://localhost:7001",
+      jwtSecret: "supersecret",
+      cookieSecret: "superSecret",
+    },
   },
   featureFlags: {},
   plugins: [],
--- a/packages/medusa/src/loaders/helpers/routing/__fixtures__/server/index.js
+++ b/packages/medusa/src/loaders/helpers/routing/__fixtures__/server/index.js
@@ -124,7 +124,7 @@ export const createServer = async (rootDir) => {
             user_id: opts.adminSession.userId || opts.adminSession.jwt?.userId,
             domain: "admin",
           },
-          config.projectConfig.auth.jwtSecret
+          config.projectConfig.http.jwtSecret
         )
 
         headers.Authorization = `Bearer ${token}`
@@ -137,7 +137,7 @@ export const createServer = async (rootDir) => {
               opts.clientSession.jwt?.customer_id,
             domain: "store",
           },
-          config.projectConfig.auth.jwtSecret
+          config.projectConfig.http.jwtSecret
         )
 
         headers.Authorization = `Bearer ${token}`


packages/medusa/src/loaders/index.ts
```diff
@@ -107,6 +107,8 @@ export default async ({
     featureFlagRouter
   )
 
+  console.log("Entry points loaded: ")
```
sradevski: Let's remove this, or replace it with `track` as before?

packages/medusa/src/loaders/config.ts
```diff
@@ -72,8 +83,9 @@ export default (rootDirectory: string): ConfigModule => {
 
   return {
     projectConfig: {
-      jwt_secret: jwt_secret ?? "supersecret",
-      cookie_secret: cookie_secret ?? "supersecret",
+      // Question: Should we just introduce the breaking change now, or do we want to stay backward compatible?
```
sradevski: I'd say let's start with a breaking change, and once we have everything working and we are closer to a release we evaluate what we want to be backwards-compatible with and what not.

packages/medusa/src/loaders/config.ts
```diff
@@ -34,29 +67,7 @@ export default (rootDirectory: string): ConfigModule => {
...
-  if (!cookie_secret) {
-    errorHandler(
-      `[medusa-config] ⚠️ cookie_secret not found.${
-        isProduction
-          ? ""
-          : " fallback to either cookie_secret or default 'supersecret'."
-      }`
-    )
-  }
+  configModule.projectConfig.auth = buildAuthConfig(configModule.projectConfig)
```
sradevski: I suggest we do this for the entire `projectConfig` somewhere higher in the stack (something like `normalizeConfig` in the `loaders/index.ts` maybe), so you never need to do
```
const somevalue = projectConfig.someExpectedValue ?? "default-value
```
throughout the codebase, and all the defaults are set on load

packages/medusa/src/api-v2/auth/middlewares.ts
```diff
@@ -17,4 +17,9 @@ export const authRoutesMiddlewares: MiddlewareRoute[] = [
     matcher: "/auth/:scope/:auth_provider",
     middlewares: [],
   },
+  {
+    method: ["DELETE"],
+    matcher: "/auth/:scope",
```
sradevski: Should this be `/auth/session`? Currently we don't support invalidating JWTs, so it's better to be more explicit (similar to how we have `/auth/session`) WDYT?

packages/medusa/src/loaders/index.ts
olivermrbl: This sneaked itself in from testing. Will remove

packages/medusa/src/api-v2/auth/middlewares.ts
olivermrbl: Yeah, I think that makes sense. Will update the PR.

With respect to invalidating JWTs, I think we need to introduce a few changes now and then maybe add a more sophisticated invalidation mechanism in the future. 

On the backend, we introduce the expiry time (defaulting to 24 hours) and use that to sign JWTs upon authenticating. Pretty standard, but apparently we haven't had that.

On the client (i.e. admin), we bust the token storage and no longer attach it to the request automatically.

Ofc. you would still be able to authenticate with that token until it expires, but only if you can get a hold of it, which is very, very unlikely given that we removed it from local storage.

In the future, we can think about adding JWT token invalidation on the backend too, for example, through a token blacklist or similar, but I don't think we need that right now. Let me know if you have any ideas here.

packages/core/types/src/common/config-module.ts
```diff
@@ -590,6 +593,42 @@ export type ProjectConfigOptions = {
...
+     */
+    jwtSecret: string
+    /**
+     * The expiration time for the JWT token. If not provided, the default value is `24h`.
+     */
+    jwtExpiresIn: string
+    /**
+     * The secret to sign the cookie session, usually a randomly generated string. If not provided, the default value is `supersecret`.
+     */
+    cookieSecret: string
```
adrien2p: suggestion: probably optional since we have the default values in the config loader.

packages/core/types/src/common/config-module.ts
```diff
@@ -590,6 +593,42 @@ export type ProjectConfigOptions = {
...
+    /**
+     * Specify the accepted URLs or patterns for API Routes starting with `/auth`. It can either be one accepted origin, or a comma-separated list of accepted origins.
+     */
+    cors: string
```
adrien2p: Q: can't it be a regexp like string? like for the admin/store cors, in the comment it looks like it is still the old description

adrien2p (approved):
Sounds good to me

packages/core/types/src/common/config-module.ts
olivermrbl: Yup, haven't added the full JS comment yet. Wanted approval first 👍

packages/medusa/src/api-v2/auth/middlewares.ts
sradevski: @olivermrbl tokens with short expiry definitely reduce the attack surface and risk, so typically that should suffice.

For high-risk applications you usually have a `blacklist` of JWT tokens that were invalidated, and before every auth check you first make sure that the token wasn't invalidated. However, this beats the purpose of JWT tokens as you need to do a session check basically (or at least the same sort of operation), so I'd say we don't need to support it in the near future.

The point still holds though that we should have this specific to sessions, and we will probably keep it like that for a long time 😄

sradevski (approved):

olivermrbl:
Should we change the `auth` config to `http` and put everything related to that in there:
```ts
http: {
    jwtSecret: "test",
    cookieSecret: "test",
    authCors: "...",
    storeCors: "...",
    authCors: "...",  
    compression: { ... }
  }
```

cc @srindom

adrien2p:
I think it can make sense 👍 jwt is really specific to auth but I am find with defining the config in a kind of layer approach

carlos-r-l-rodrigues:
I believe `http` makes sense.

olivermrbl:
Done, let me know what you think :)

carlos-r-l-rodrigues (approved):
nice!

olivermrbl:
@shahednasser, would you be up for helping me ensure the documentation reflects these changes? 🙏

shahednasser:
@olivermrbl yes thanks for letting me know!

olivermrbl:
Btw. I still need to clean up after we decided to make all these changes breaking. Essentially, it means we won't be keeping any of the old secrets and configs that have been moved to `http`.

integration-tests/api/medusa-config.js
```diff
@@ -24,10 +24,12 @@ module.exports = {
...
-    jwt_secret: "test",
-    cookie_secret: "test",
-    http_compression: {
-      enabled: enableResponseCompression,
+    http: {
+      compression: {
+        enabled: enableResponseCompression,
+      },
+      jwtSecret: "test",
+      cookieSecret: "test",
```
adrien2p, line 32: suggestion: maybe we should allow to set the cookie domain and make it required when running in production to avoid being able to send the cookie everywhere else. cc @sradevski

adrien2p: unless it is automatically done under the hood?

olivermrbl: I think it defaults to the domain from which it was set

packages/admin-next/dashboard/src/components/layout/shell/shell.tsx
```diff
@@ -200,20 +203,20 @@ const ThemeToggle = () => {
...
-    //   },
-    // })
-    // noop
+    await logoutMutation(undefined, {
+      onSuccess: () => {
+        /**
+         * When the user logs out, we want to clear the query cache
+         */
+        queryClient.clear()
+        console.log("Hello world")
```
sradevski: Do we want to log hello world here? 😆

olivermrbl: Haha, I thought that would be a nice easter egg 😆

adrien2p: Maybe we can log our recruitment description ?? ☺️

integration-tests/api/medusa-config.js
sradevski: ok we'll need to look into this to ensure the cookie doesn't leak into 3rd party services (eg. sentry), but I'll tackle that as part of the bigger auth writeup.

packages/admin-next/dashboard/src/lib/client/auth.ts
```diff
@@ -14,9 +14,14 @@ async function login(token: string) {
   })
 }
 
+async function logout() {
```
sradevski: @olivermrbl let's add logout to the SDK now, and delete the `auth.ts` file. I believe after rebase it got recreated

sradevski: Ah nvm, you haven't rebased yet 😄 Well after you rebase, let's do that

PR #7340, opened 2024-05-15
fix(medusa, modules-sdk): Fix migrate command and fix revert support

adrien2p:
**What**
Fix the migrate command and cleanup

### DIFF
--- a/packages/core/modules-sdk/src/medusa-app.ts
+++ b/packages/core/modules-sdk/src/medusa-app.ts
@@ -389,12 +389,21 @@ async function MedusaApp_({
         }
       }
 
-      await MedusaModule.migrateUp(
-        moduleResolution.definition.key,
-        moduleResolution.resolutionPath as string,
-        moduleResolution.options,
-        moduleResolution.moduleExports
-      )
+      if (!revert) {
+        await MedusaModule.migrateUp(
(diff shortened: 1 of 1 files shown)


carlos-r-l-rodrigues (approved):

packages/core/modules-sdk/src/medusa-app.ts
```diff
@@ -379,24 +389,53 @@ async function MedusaApp_({
         }
       }
 
-      await MedusaModule.migrateUp(
-        moduleResolution.definition.key,
-        moduleResolution.resolutionPath as string,
-        moduleResolution.options,
-        moduleResolution.moduleExports
-      )
+      if (!revert) {
```
sradevski, line 392:
```suggestion
      if (revert) {
```
And then change the order they are performed. Optionally, and better IMO, is to rename the variable to `apply` instead of the negative form `revert`

sradevski (approved):
Very nice 👍

packages/core/modules-sdk/src/medusa-app.ts
adrien2p: do you mean, apply to specify the run migration and if it is false then I revert?

PR #7341, opened 2024-05-15
chore: Publish preview release + deploy staging every 3 hours

olivermrbl:


.github/workflows/trigger-staging-deployment.yml
```diff
@@ -0,0 +1,18 @@
...
+
+jobs:
+  deploy:
+    runs-on: ubuntu-latest
+    steps:
+      - run: |
+          curl -X POST \
+          -H "Authorization: Bearer ${{ secrets.STAGING_DEPLOY_ACCESS_TOKEN }}" \
+          -H "Accept: application/vnd.github.v3+json" \
+          https://api.github.com/repos/@medusajs/staging/actions/workflows/deploy.yml/dispatches \
```
srindom: shouldn't this be?
```suggestion
          https://api.github.com/repos/medusajs/staging/actions/workflows/deploy.yml/dispatches \
```

olivermrbl: yes nice catch!

adrien2p:
Question, couldn't we deploy everytime we merge and cancel the previous deployment if we have many merge one after each other? It might be a dumb question ahah

olivermrbl:
> Question, couldn't we deploy everytime we merge and cancel the previous deployment if we have many merge one after each other? It might be a dumb question ahah

Long-term, yes, I think we should do this. However, right now, we are merging quite a lot, so the schedule is just a measure to not overload the system 😅

adrien2p:
makes total sense 👍

adrien2p (approved):

PR #7342, opened 2024-05-15
chore(modules-sdk, medusa): Cleanup migration helpers

adrien2p:
**What**
Cleanup the migrations scripts


adrien2p:
@sradevski I went with just inveting the method calls as I think revert is pretty explicit and keeps the run as the default behaviour, wdyt?

olivermrbl (approved):

PR #7344, opened 2024-05-15
fix(core-flows): fix variant pricing validation for carts

riqwan:
what:

- not checking the calculated_amount was leading to a false success of the variant price validation step
- regions are randomly added if its not present in the create cart flow.
  - TODO: we should add the default region from the store instead of randomly picking one up.

### DIFF
--- a/packages/core/core-flows/src/definition/cart/steps/validate-variant-prices.ts
+++ b/packages/core/core-flows/src/definition/cart/steps/validate-variant-prices.ts
@@ -1,11 +1,13 @@
 import { BigNumberInput } from "@medusajs/types"
-import { MedusaError, isDefined } from "@medusajs/utils"
+import { MedusaError, isPresent } from "@medusajs/utils"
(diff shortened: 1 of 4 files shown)


olivermrbl (approved):

packages/core/core-flows/src/definition/cart/steps/validate-variant-prices.ts
```diff
@@ -1,11 +1,13 @@
...
+import { MedusaError, isPresent } from "@medusajs/utils"
 import { createStep } from "@medusajs/workflows-sdk"
 
 interface StepInput {
   variants: {
     id: string
-    calculated_price?: BigNumberInput
+    calculated_price?: {
+      calculated_amount?: BigNumberInput | null
+    }
```
adrien2p, line 10: suggestion: I ve create a WithCalculatedPrice in the types package, also I ve used number as once received from the module I believe we have numbers after serialization

PR #7345, opened 2024-05-16
fix(modules-sdk): Resolve discovery location properly

adrien2p:
**What**
The index file is always supposed to be next to the other potental directories to discover.

**Note**
I ve discussed something with virk to provide helpers to export the module definition and to build the module joiner config, in the future, this helpers could help us move the discovery away from the internal loaders in order to always load the modules the same way. The helper would take care of discovering and exporting the missing pieces


shahednasser (approved):
Tested in local project, LGTM

sradevski (approved):
LGTM for now, let's see if we'll need to improve it further later

adrien2p:
@sradevski i have some improvements in my head already that i ve shared in virk rfc. Let see that later but it would move this logic away and be more robust

carlos-r-l-rodrigues (approved):

PR #7346, opened 2024-05-16
feat(dashboard): manage groups from customer details

fPolic:
**WHAT**
- add support for adding customers to multiple customer groups from the customer details page
- remove customer from a group on the customer details page
- fix issue with unresponsive UI after prompt disaply


olivermrbl:
Should the row be selectable?
CleanShot 2024-05-16 at 12 56 02

olivermrbl:
After removing the customer, I can''t interact with anything. We've had this many times by now. Do we know what is going on? And can be fix it more generally, so we don't miss it in the future 😄

fPolic:
> After removing the customer, I can''t interact with anything. We've had this many times by now. Do we know what is going on? And can be fix it more generally, so we don't miss it in the future 😄

~~Yep it's the issue with prompt rendering in a different root~~ , I'll look into it now

EDIT: should be fixed

olivermrbl (approved):
LGTM!

@kasperkristensen would be good to have you look through it too

PR #7348, opened 2024-05-16
Fix issue related to removing last product from the collection

thetutlage:
We had an early return which was causing not running any Database queries when last product the removed. The fix was simply to remove the check

### DIFF
--- a/packages/modules/product/src/services/product-module-service.ts
+++ b/packages/modules/product/src/services/product-module-service.ts
@@ -928,8 +928,8 @@ export default class ProductModuleService<
 
     const collections: TProductCollection[] = []
 
-    const updateSelectorAndData = updatedCollections
-      .map((collectionData) => {
+    const updateSelectorAndData = updatedCollections.flatMap(
+      (collectionData) => {
         const input = normalizedInput.find((c) => c.id === collectionData.id)
         const productsToUpdate = (input as any)?.products
 
@@ -968,8 +968,8 @@ export default class ProductModuleService<
         })
 
         return result
-      })
-      .flat()
+      }
+    )
 
     await this.productService_.update(updateSelectorAndData, sharedContext)
     return collections


olivermrbl:
Nice @thetutlage – could you add a test too? 🙏

thetutlage:
Tests added. The PR is ready for review

thetutlage:
Some unrelated tests from the notification module are failing

sradevski (approved):

packages/modules/product/src/services/product-module-service.ts
```diff
@@ -941,7 +938,7 @@ export default class ProductModuleService<
           sharedContext
         )
 
-        if (productsToUpdate.length > 0) {
+        if (productsToUpdate?.length > 0) {
```
```suggestion
const updateSelectorAndData = updatedCollections.map((collectionData) => {
  const input = normalizedInput.find((c) => c.id === collection.id)
  const productsToUpdate = (input as any)?.products

  const dissociateSelector = {
    collection_id: collectionData.id,
  }
  const associateSelector = {}

  if (!!productsToUpdate?.length) {
    const productIds = productsToUpdate.map((p) => p.id)
    dissociateSelector["id"] = { $nin: productIds }
    associateSelector["id"] = { $in: productIds }
  }

  const result: Record<string, any>[] = [{
    selector: dissociateSelector,
    data: {
      collection_id: null,
    },
  }]
  
  if (isPresent(associateSelector)) {
    result.push({
      selector: associateSelector,
      data: {
        collection_id: collectionData.id,
      },
    })
  }
  
  collection['products'] = productsToUpdate ?? []
  
  return result
}).flat()

await this.productService_.update(
  updateSelectorAndData,
  sharedContext
)

return updatedCollections
```

can you tell me if this could work, so that we don't have to have a promise all of many promises of potentially unknown size.

packages/modules/product/src/services/product-module-service.ts
```diff
@@ -925,34 +926,53 @@ export default class ProductModuleService<
       sharedContext
     )
 
-    const collectionWithProducts = await promiseAll(
-      updatedCollections.map(async (collection, i) => {
-        const input = normalizedInput.find((c) => c.id === collection.id)
+    const collections: TProductCollection[] = []
+
+    const updateSelectorAndData = updatedCollections
+      .map((collectionData) => {
```
adrien2p: nit: flatMap

packages/modules/product/src/services/product-module-service.ts
```diff
@@ -925,34 +926,53 @@ export default class ProductModuleService<
...
-        return { ...collection, products: productsToUpdate || [] }
+        collections.push({
+          ...collectionData,
+          products: productsToUpdate ?? [],
+        })
+
+        return result
       })
-    )
+      .flat()
```
adrien2p: nit: rm if using flatMap

adrien2p (approved):
LGTM

olivermrbl:
@thetutlage minor thing, could I get you to use squash and merge going forward? 🙏

thetutlage:
Sure. Will keep that in mind

PR #7349, opened 2024-05-16
feat: Add support for authentication to the sdk, and plug it in the admin

sradevski:


packages/core/js-sdk/src/auth/index.ts
```diff
@@ -0,0 +1,33 @@
...
+  private client: Client
+  private config: Config
+
+  constructor(client: Client, config: Config) {
+    this.client = client
+    this.config = config
+  }
+
+  login = async (payload: { email: string; password: string }) => {
+    // TODO: It is a bit strange to eg. require to pass `scope` in `login`, it might be better for us to have auth methods in both `admin` and `store` classes instead?
```
adrien2p, line 14: I believe it would be easier to use as well as we would be inferring the scope ourselves, and if the user want a global auth with scope we can either provide it (in case the user customisations involves more scopes) or they can extends the client themselves no?

packages/core/js-sdk/src/client.ts
```diff
@@ -30,6 +37,8 @@ const normalizeRequest = (
   return {
     ...init,
     headers,
+    // TODO: Setting this to "include" poses some security risks, as it will send cookies to any domain. We should consider making this configurable.
+    credentials: config.auth?.type === "session" ? "include" : "omit",
```
adrien2p, line 41: theoratically it wont, the browser manage that depending on the cookie configuration I believe, and on the cookie you can set the domain

packages/core/js-sdk/src/client.ts
```diff
@@ -125,59 +152,92 @@ export class Client {
       )
 
       // Any non-request errors (eg. invalid JSON in the response) will be thrown as-is.
-      return fetch(normalizedInput, normalizeRequest(init, headers)).then(
-        (resp) => {
-          this.logger.debug(`Received response with status ${resp.status}\n`)
-          return normalizeResponse(resp, headers)
-        }
-      )
+      return fetch(
```
adrien2p: todo: add await, just to prevent miss behaviour inside other functions and try catch block just in case

packages/core/js-sdk/src/client.ts
```diff
@@ -125,59 +152,92 @@ export class Client {
...
+        normalizedInput,
+        normalizeRequest(init, headers, this.config)
+      ).then((resp) => {
+        this.logger.debug(`Received response with status ${resp.status}\n`)
```
adrien2p, line 159: Q: should we keep that?

packages/core/js-sdk/src/client.ts
```diff
@@ -125,59 +152,92 @@ export class Client {
...
+    }
+    if (!hasStorage("sessionStorage") && storageMethod === "session") {
+      throw new Error("Session JWT storage is only available in the browser")
+    }
+
+    return {
+      storageMethod,
+      storageKey,
+    }
+  }
```
adrien2p: nit 

```suggestion
protected getTokenStorageInfo_ = () => {
  const hasLocalStorage = hasStorage("localStorage")
  const hasSessionStorage = hasStorage("sessionStorage")
  const storageMethod =
    this.config.auth?.jwtTokenStorageMethod ||
    (hasLocalStorage ? "local" : "memory")
  
  const storageKey =
    this.config.auth?.jwtTokenStorageKey || this.DEFAULT_JWT_STORAGE_KEY

  if (!hasLocalStorage && storageMethod === "local") {
    throw new Error("Local JWT storage is only available in the browser")
  }
  
  if (!hasSessionStorage && storageMethod === "session") {
    throw new Error("Session JWT storage is only available in the browser")
  }

  return {
    storageMethod,
    storageKey,
  }
}
```

sradevski:
@fPolic @kasperkristensen I'd appreciate your review here 🙇‍♂️

adrien2p (approved):

packages/core/js-sdk/src/client.ts
sradevski: @adrien2p this is what happens now, I run the admin and BE on different ports, and with `include` it happily sends it to the BE (which is a different domain, so to speak).

packages/core/js-sdk/src/client.ts
sradevski: @adrien2p yup, if you can see how `debug` is defined, if you set `debug` in the config it will log these, otherwise it won't

packages/core/js-sdk/src/auth/index.ts
sradevski: @adrien2p yeah that's what I am thinking, but let's see what others think as well.

packages/core/js-sdk/src/client.ts
adrien2p: yes because the configuration locally is loose, but not in production

adrien2p: here is more info https://developer.mozilla.org/en-US/docs/Web/HTTP/Cookies#define_where_cookies_are_sent

sradevski: @adrien2p yeah I understand if you configure it correctly it won't but what I am saying is it isn't currently, and this is  what poses a security risk in this case, so I was thinking we maybe let the user decide if they want to allow it or not. As long as they don't meddle with the session config, and we don't mess it up it's ok, but it is not unheard of to misconfigure things 😄

adrien2p: But since the cookie is configured on the BE and we manage the configuration, couldn't we throw if we dont have the whole required config when starting ? It would eliminate security issue no? The thing is that if you receive a cookie it is for a reason, not sending it back could break the flow

adrien2p: Oli told be that it might default to the domain it was set from, so maybe just changing the port is not enough to test ahah

sradevski: @adrien2p the default config is `same-origin`, so the cookie would be sent, but only if it is the same origin as the web. So `localhost:9000` and `localhost:5173` are considered different origins, and `same-origin` doesn't work, but `include` does.

Anyway I'll look into it more, but even if we control the configuration it doesn't mean we can't have a bug that would expose the cookie to others by mistake.

packages/admin-next/dashboard/package.json
```diff
@@ -33,6 +33,7 @@
     "@dnd-kit/sortable": "^8.0.0",
     "@hookform/resolvers": "3.3.2",
     "@medusajs/icons": "1.2.1",
+    "@medusajs/js-sdk": "0.0.2-snapshot-20240516152256",
```
kasperkristensen: You should just add the version that is in the sdk's package.json. Then changeset will take care of updating it with each release, and yarn will point it towards the local dependency.

packages/admin-next/dashboard/src/lib/client/client.ts
```diff
@@ -57,3 +57,10 @@ export const client = {
   stockLocations: stockLocations,
   workflowExecutions: workflowExecutions,
 }
+
+export const sdk = new Medusa({
+  baseUrl: MEDUSA_BACKEND_URL,
```
kasperkristensen: replace with `__BACKEND_URL__ || "http://localhost:9000"`, since we are deleting the `medusa` file once we have removed all v1 code.

kasperkristensen (approved):
LGTM with minor comments

packages/core/js-sdk/src/auth/index.ts
srindom: I don't mind having auth in `store.` and `admin.` - but would still want there to be an `auth.` so you can access that endpoint easily if you have custom scopes.

packages/core/js-sdk/src/auth/index.ts
```diff
@@ -0,0 +1,33 @@
...
+export class Auth {
+  private client: Client
+  private config: Config
+
+  constructor(client: Client, config: Config) {
+    this.client = client
+    this.config = config
+  }
+
+  login = async (payload: { email: string; password: string }) => {
```
srindom, line 13: thought: this doesn't seem to support other provider logins. 

would expect a signature like this `authenticate("scope", "provider", { ...providerData })`

fPolic (approved):
LGTM

packages/core/js-sdk/src/auth/index.ts
sradevski: @srindom yeah makes sense, we can do that. `auth` can expose the more generic endpoint, whereas store and admin will simply call it with hard-coded scopes. I'll do it in a follow-up PR

packages/core/js-sdk/src/auth/index.ts
sradevski: @srindom yeah I think the exact interface needs to be finalized (as you can see, even the typings are hard-coded for now. I just wanted to replace what we had in admin to put the SDK in use. I will be doing more improvements today, and probably in the following week.

PR #7351, opened 2024-05-16
fix: Retrieve region in loader

olivermrbl:
old client doesn't accept a query


carlos-r-l-rodrigues (approved):

sradevski (approved):
After https://github.com/medusajs/medusa/pull/7349 is merged we can start using the SDK instead

PR #7357, opened 2024-05-17
feat(core-flows, fulfillment): Add create return specific method and add more tests

adrien2p:
**What**
- fix create return workflow
- add more integration tests for this iteration

### DIFF
--- a/packages/core/types/src/http/common/request.ts
+++ b/packages/core/types/src/http/common/request.ts
@@ -1,5 +1,5 @@
 export interface SelectParams {
-  fields?: string[]
+  fields?: string
 }
 
 export interface FindParams extends SelectParams {
--- a/packages/medusa/src/api/store/orders/query-config.ts
+++ b/packages/medusa/src/api/store/orders/query-config.ts
@@ -16,6 +16,7 @@ export const defaultStoreRetrieveOrderFields = [
   "version",
   "summary",
   "currency_code",
+  "display_id",
   "total",
   "subtotal",
   "tax_total",


packages/modules/order/src/joiner-config.ts
```diff
@@ -37,6 +37,7 @@ export const joinerConfig: ModuleJoinerConfig = {
       name: ["return_reason", "return_reasons"],
       args: {
         entity: ReturnReason.name,
+        methodSuffix: pluralize(ReturnReason.name),
```
adrien2p, line 40: @carlos-r-l-rodrigues do you think it would be wise by default to do that? if the entity if provided and it is not the first ailas then we do that under the hood? convention being the main entity is always the first

adrien2p:
@olivermrbl I have this CLI pipeline failing, I believe that you said something about it being fixed at some point is that correct?

packages/modules/fulfillment/src/services/fulfillment-module-service.ts
```diff
@@ -627,10 +604,43 @@ export default class FulfillmentModuleService<
...
+      fulfillment
+    )
+  }
+
+  @InjectManager("baseRepository_")
+  async createReturnFulfillment(
+    data: FulfillmentTypes.CreateFulfillmentDTO,
+    @MedusaContext() sharedContext: Context = {}
+  ): Promise<FulfillmentTypes.FulfillmentDTO> {
+    const { order, ...fulfillmentDataToCreate } = data
```
sradevski, line 616: Why do we have an order passed to the fulfillment module? I checked the typings and it seems it's an empty object now, but I guess it's better to remove it altogether?

packages/modules/order/src/joiner-config.ts
carlos-r-l-rodrigues: Yeah, it is alright to do that.
Maybe we could even infer the methodSuffix in the remote query from the property entity. and if it is different we can specify it.
But not in this PR.

carlos-r-l-rodrigues (approved):

packages/modules/fulfillment/src/services/fulfillment-module-service.ts
adrien2p: well I am not sure why we have that, it is also not used but it is required in the types and in the workflow for the fulfillment. I didn't want to do that in that pr until we know why we have that as I dont recall the need for it but I can still rm it everywhere once we are aware of why we added it in the first place. Also we have a link between order and fulfillment, so could it be for the provider?

cc @olivermrbl

carlos-r-l-rodrigues: v1 uses the order to pass some info to the providers.
we could rename that to something else in the future though

olivermrbl: Yeah, what @carlos-r-l-rodrigues said. This is just a naming issue. The order can hold details that are required for the provider to create the fulfillment

sradevski: Ok got it 👍 Since a fulfillment is not necessarily tied to an order, let's try to rename it to something more meaningful (In payments we have `data`, which is very generic but it might be OK since each provider might require something different)

carlos-r-l-rodrigues: I can change that to `data` in my follow up PR where I already removed `order_id` from types.

adrien2p: I believe The fulfillment already have a field data which is what is returned by the provider, we night need to think about something else that seems more meaningful ? extra_data?

adrien2p: Or something more specific like provider_data?

sradevski (approved):
BTW we can do the change in a separate PR, I just noticed it now :)

adrien2p:
@olivermrbl can I merge even with this CI failing?

olivermrbl:
Yes, sure. This is related to the starter, so should be fine.

PR #7367, opened 2024-05-19
fix: nullish BACKEND_URL in client

olivermrbl:
Nullish `BACKEND_URL`


olivermrbl:
I think we need a stricter config in the `dashboard`, such that things like this doesn't slip through

kasperkristensen (approved):

PR #7369, opened 2024-05-20
chore: pick the default store region

riqwan:
what:

- adds a default region to be created on first load
- when picking any region for cart, we pick the default region of the store.


integration-tests/modules/__tests__/cart/store/carts.spec.ts
```diff
@@ -88,6 +90,8 @@ medusaIntegrationTestRunner({
           name: "Default Region",
           currency_code: "dkk",
         })
+
+        await createDefaultsWorkflow(appContainer).run()
```
sradevski: Wouldn't this have already run at this point in time?

packages/core/core-flows/src/defaults/steps/create-default-region.ts
```diff
@@ -0,0 +1,43 @@
...
+    let shouldDelete = false
+    let [region] = await service.list({}, { take: 1 })
+
+    if (!region) {
+      const { result } = await createRegionsWorkflow(container).run({
+        input: { regions: [data.region] },
+      })
+
+      region = result[0]
+      shouldDelete = true
```
sradevski: I think instead of using this flag, what we can do is return 

```
return new StepResponse(region, { regionId: newlyCreatedRegion.id | undefined })
```

And in the reverse step we check if `regionId` is set (it will be only in the case of a newly created region)

packages/core/core-flows/src/defaults/steps/create-default-store.ts
```diff
@@ -11,25 +15,35 @@ export const createDefaultStoreStepId = "create-default-store"
...
+              default_currency_code: region?.currency_code ?? "usd",
+              default_region_id: region?.id,
             },
           ],
         },
       })
 
+      store = result[0]
+
       shouldDelete = true
```
sradevski: We can refactor it here as well if you like the proposal I had above.

packages/core/core-flows/src/defaults/steps/create-default-store.ts
```diff
@@ -11,25 +15,35 @@ export const createDefaultStoreStepId = "create-default-store"
...
+      ModuleRegistrationName.STORE
+    )
+    const regionService = container.resolve<IRegionModuleService>(
+      ModuleRegistrationName.REGION
+    )
 
     let shouldDelete = false
     let [store] = await storeService.list({}, { take: 1 })
 
     if (!store) {
```
sradevski: We can do an early return, i.e `if(store) return store`

packages/core/core-flows/src/definition/cart/steps/find-one-or-any-region.ts
```diff
@@ -11,18 +11,29 @@ export const findOneOrAnyRegionStep = createStep(
       ModuleRegistrationName.REGION
     )
 
+    const storeModule = container.resolve<IStoreModuleService>(
+      ModuleRegistrationName.STORE
+    )
+
     if (!data.regionId) {
```
sradevski: It might be easier to read if it is `if(data.regionId)` instead.

integration-tests/modules/__tests__/cart/store/carts.spec.ts
riqwan: Only on the first test, we truncate the database after each test

sradevski: aaaah yes, you're right 👍 that might be a bit unexpected I'd say, I'd probably imagine each test running in the exact same environment. But anyway nothing to action, maybe just leave a comment?

adrien2p: Jumping Quickly, but maybe this should be part of the runner? Wdyt?

riqwan: That was my first thought, I didn't know if the plans for the runner is to be core-specific. Shall we throw it in there?

adrien2p: I believe that the runner already run the main medusa loader and i thought it was already part if it. If you can check that in the runner and maybe we can validate that part, ill be back tomorrow

riqwan: no rush :) I'll take a look

packages/core/core-flows/src/defaults/workflows/create-defaults.ts
```diff
@@ -1,17 +1,23 @@
 import { createWorkflow } from "@medusajs/workflows-sdk"
 import { createDefaultSalesChannelStep } from "../../sales-channel"
+import { createDefaultRegionStep } from "../steps"
 import { createDefaultStoreStep } from "../steps/create-default-store"
 
 export const createDefaultsWorkflowID = "create-defaults"
 export const createDefaultsWorkflow = createWorkflow(
```
olivermrbl: thought(non-blocking): Might make sense to kill this defaults workflow, except for the store part, in favor of a seed script. There will be many other resources we need to load in by default to prepare a simple store setup, and I thought we might as well make the core less opinionated about which those resources are, if we can.

sradevski: @olivermrbl do you think it might be a good idea for the seed itself to be a workflow? What I am thinking is, rather than having yet another concept for seeding data, you can define your own `create-defaults` workflow. 

Then, we support "hot replacement" of workflows - if you attempt to register a workflow with the same name, the original one gets replaced. I think this replacement mechanism would be useful if eg. you want to use medusa as-is, except `create-products` should do one more step, you can just create a new workflow and that's it, you wouldn't need to touch the HTTP endpoints, for example.

Just thinking out loud, I am also not 100% sure this is a good idea, but it might also be a good intro to workflows for users.

olivermrbl: > do you think it might be a good idea for the seed itself to be a workflow?

It was definitely the plan to build the default seed script as a workflow. The idea was actually just to call the workflow we have now in a script rather in the core loader.

> rather than having yet another concept for seeding data

I am not sure I understand – as mentioned above this would be a command executing a script like so:
```
// in package.json
"yarn seed": "medusa script ../workflows/defaults.ts"
```

As such, seeding data will not be a separate concept, but something enabled via our CLI + a workflow. If users want to seed their own data, they have the defaults workflow locally in their project and can make the changes directly to it. Thereby, effectively replacing it as you also mention. 

> you want to use medusa as-is, except create-products should do one more step, you can just create a new workflow and that's it, you wouldn't need to touch the HTTP endpoints, for example.

From what I understand, you are talking about extending the flows in `core-flows`, which is definitely on the radar for further improvements to the Workflows SDK. 

Right now, you can, in theory, import the workflow in your own project and append or remove steps. However, we will likely discourage this in favor of dedicated "hooks" in some of the flows we expect to be relevant for extensions, e.g. create product or complete cart. The hooks would be a measure against users messing up the business-critical flows. However, this is a separate discussions to have in the future.


Nothing is set in stone regarding seeding, but this was just my current thinking.

sradevski: @olivermrbl yeah I am talking about extending workflows, but also supporting replacement (and not just a purely additive extension). I wasn't aware of the hooks approach we've thought about 👍 

Yeah ok it makes sense for the workflow to be ran outside of the medusa app, I agree. I just think it would be nice if it is a workflow, so you also don't end up with a partial seed data.

riqwan: @olivermrbl I can move this over to the test seeds, but I'm guessing we still keep the changes made in find-one-or-any-region or do we close this PR?

olivermrbl: > but I'm guessing we still keep the changes made in find-one-or-any-region

Yeah, think we can keep those in 👍

integration-tests/modules/__tests__/cart/store/cart.workflows.spec.ts
```diff
@@ -89,11 +90,9 @@ medusaIntegrationTestRunner({
       beforeEach(async () => {
         await createAdminUser(dbConnection, adminHeaders, appContainer)
 
-        // Here, so we don't have to create a region for each test
-        defaultRegion = await regionModuleService.create({
-          name: "Default Region",
-          currency_code: "dkk",
-        })
+        const { region } = await seedStorefrontDefaults(appContainer, "dkk")
```
riqwan, line 93: @adrien2p @sradevski keeping this in the test file as its not required for most specs.

sradevski (approved):

integration-tests/modules/__tests__/cart/store/cart.workflows.spec.ts
adrien2p: sounds good 👍

PR #7370, opened 2024-05-20
feat: expose backend url in admin options

olivermrbl:
```
projectConfig: {
  admin: {
    backendUrl: "http://localhost:3000",
  },
}
```

### DIFF
--- a/packages/core/types/src/common/config-module.ts
+++ b/packages/core/types/src/common/config-module.ts
@@ -36,7 +36,7 @@ export type AdminOptions = {
   /**
    * The URL of your Medusa server. Defaults to an empty string, which means requests will hit the same server that serves the dashboard.
    */
-  backend?: string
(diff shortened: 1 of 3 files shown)


packages/core/types/src/common/config-module.ts
```diff
@@ -33,6 +33,10 @@ export type AdminOptions = {
    * The default value is `./build`.
    */
   outDir?: string
+  /**
+   * The URL of your Medusa server. Defaults to an empty string, which means requests will hit the same server that serves the dashboard.
+   */
+  backend?: string
```
sradevski: should we call it `backendUrl` instead? I am fine either way, but it doesn't hurt to be explicit.

sradevski (approved):

fPolic (approved):

PR #7371, opened 2024-05-20
feat(sdk): Replace region calls with the SDK in admin, apply typings …

sradevski:


packages/admin-next/dashboard/src/hooks/table/filters/use-order-table-filters.tsx
```diff
@@ -27,7 +27,7 @@ export const useOrderTableFilters = (): Filter[] => {
       label: t("fields.region"),
       type: "select",
       options: regions.map((r) => ({
-        label: r.name,
+        label: r.name ?? r.id ?? "",
```
riqwan: todo: The ID shouldn't be exposed to the user in the UI, any reason why this is being done?

riqwan (approved):
nice!

packages/admin-next/dashboard/src/hooks/table/filters/use-order-table-filters.tsx
sradevski: @riqwan ah it's just a mistake, I later changed the typings saying the name is always defined. Let me fix it

PR #7372, opened 2024-05-20
Introduce directories.srcDir for configuring source code location

thetutlage:
Okay, so here is a bit of back story to explain the change I am trying to make.

## The goal

The goal is to eventually use a JIT (TS-Node + SWC) to run the backend API in development. Once this change goes live, there won't be any `dist` folder in the project. We should discover resources from the `src` folder.

Now, this change must be done inside the loaders so that instead of importing files from `dist`, loaders will import files from the `src` folder. BUT ONLY IF IT IS AN APP (NOT A PACKAGE).

## Avoiding breaking everything

This is going to be a two-fold change. First in the `medusa/medusa` package and next in the starter kit (we will configure TS-Node + SWC in starter kit). So, in the meantime, old starter kits with new medusa package will break.

So what I am doing is, introducing a temporary config called `directories.srcDir`. The default value of this config is `dist`. So old apps will continue to discover the resources from the `dist` folder.

And for the new apps (using JIT compiler), we will define `directories.srcDir` under the `medusa-config.js` file. 

Maybe in couple of days when everything is stable. We can remove this config altogether.

### DIFF
--- a/packages/medusa/src/loaders/index.ts
+++ b/packages/medusa/src/loaders/index.ts
@@ -14,7 +14,7 @@ import expressLoader from "./express"
 import featureFlagsLoader from "./feature-flags"
 import { registerWorkflows } from "./helpers/register-workflows"
 import Logger from "./logger"
-import loadModules from "./medusa-app"
+import loadMedusaApp from "./medusa-app"
 import registerPgConnection from "./pg-connection"
 import { SubscriberLoader } from "./helpers/subscribers"
 import { getResolvedPlugins } from "./helpers/resolve-plugins"
@@ -136,7 +136,7 @@ export default async ({
   const {
     onApplicationShutdown: medusaAppOnApplicationShutdown,
     onApplicationPrepareShutdown: medusaAppOnApplicationPrepareShutdown,
-  } = await loadModules({
+  } = await loadMedusaApp({
     container,
   })
 
--- a/packages/medusa/src/loaders/medusa-app.ts
+++ b/packages/medusa/src/loaders/medusa-app.ts
@@ -142,7 +142,7 @@ export async function revertMedusaApp({
   })
 }
 
-export const loadModules = async (
+export const loadMedusaApp = async (
   {
     container,
   }: {
@@ -262,4 +262,4 @@ export async function runModulesLoader({
   })
 }
 
-export default loadModules
+export default loadMedusaApp
--- a/packages/medusa/src/commands/migrate.js
+++ b/packages/medusa/src/commands/migrate.js
@@ -9,7 +9,7 @@ import configModuleLoader from "../loaders/config"
 import featureFlagLoader from "../loaders/feature-flags"
 import Logger from "../loaders/logger"
 import {
-  loadModules,
+  loadMedusaApp,
   migrateMedusaApp,
   revertMedusaApp,
 } from "../loaders/medusa-app"
@@ -27,7 +27,7 @@ const runLinkMigrations = async (directory) => {
 
   await pgConnectionLoader({ configModule, container })
 
-  const { runMigrations } = await loadModules(
+  const { runMigrations } = await loadMedusaApp(
     { configModule, container },
     { registerInContainer: false }
   )


packages/medusa/src/loaders/helpers/register-workflows.ts
```diff
@@ -16,6 +16,11 @@ export async function registerProjectWorkflows({
   configModule,
 }) {
   const [resolved] =
-    getResolvedPlugins(rootDirectory, configModule, "dist", true) || []
+    getResolvedPlugins(
```
thetutlage: Is this the only place where we discover resources of an app. REMEMBER, WE DON'T WANT TO TOUCH PACKAGES HERE

carlos-r-l-rodrigues (approved):

olivermrbl (approved):
LGTM

riqwan (approved):

thetutlage:
Alright. After an intense session with @sradevski this PR is going to get bigger, infact a log bigger 😇

olivermrbl:
> After an intense session with @sradevski this PR is going to get bigger, infact a log bigger

Building up the suspense. I am excited to hear what will go into it 😄

packages/medusa/src/loaders/admin.ts
```diff
@@ -1,27 +1,27 @@
 import { AdminOptions, ConfigModule } from "@medusajs/types"
+import { ContainerRegistrationKeys } from "@medusajs/utils"
+import { AwilixContainer } from "awilix"
```
sradevski: Is this needed? If it is used somewhere, let's use MedusaContainer instead.

packages/medusa/src/loaders/api.ts
```diff
@@ -35,14 +34,27 @@ export default async ({ app, configModule }: Options) => {
      */
     await new RoutesLoader({
       app: app,
+      configModule: container.resolve(ContainerRegistrationKeys.CONFIG_MODULE),
       rootDir: path.join(__dirname, "../api-v2"),
```
sradevski: NIT: Let's move this into a variable

packages/medusa/src/loaders/api.ts
```diff
@@ -35,14 +34,27 @@ export default async ({ app, configModule }: Options) => {
      */
     await new RoutesLoader({
       app: app,
+      configModule: container.resolve(ContainerRegistrationKeys.CONFIG_MODULE),
```
sradevski: Let's put the configModule in a variable, and then pass it both here and below

sradevski (commented):
Looks much better 👍

thetutlage:
Okay. So here's the GIST of the changes.

In a huddle, me and @sradevski discovered that the resources from the starter kit are exposed as a plugin to the rest of the codebase. Here's the block of code that returns starter kit resources as a plugin. https://github.com/medusajs/medusa/blob/becf903cea42ba4b69d9aa9b3fdbd206dc6c765f/packages/medusa/src/loaders/helpers/resolve-plugins.ts#L115-L134

So our goal was to make all the `*loader` functions imports their specific resources from the list of plugins loaded via the `resolvePlugins` methods.

Also, at the same time we found some inconsistencies in the codebase. For example:

- In case places, we were using the `AwilixContainer` as a type and in other places we were using `MedusaContainer`. I have clean up some reference, but not all. Can do that in a separate PR.
- Some parts of our codebase accepts `ConfigModule`, some accepts a subset of `ConfigModule`. Our goal is to be as specific as possible with what a function accepts and pass exactly that info at the time of calling the function. This makes it a little easier to reason about the code. Again, haven't cleaned up everything.

Finally, we are now able to load and register routes from the starter kit after the last commit.

packages/medusa/src/loaders/admin.ts
thetutlage: I enabled `"noUnusedLocals": true,` in tsconfig and it discovered so many unused imports. Should I keep it on?

sradevski: @thetutlage if we can clean them up easily, let's keep it, otherwise we can leave it for another PR.

thetutlage:
Working on fixing the unit tests

packages/medusa/src/loaders/admin.ts
```diff
@@ -3,25 +3,23 @@ import { Express } from "express"
...
-export default async function adminLoader({ app, configModule }: Options) {
-  const { admin } = configModule
-
+export default async function adminLoader({ app, adminConfig }: Options) {
   const adminOptions: IntializedOptions = {
     disable: false,
     path: "/app",
     outDir: "./build",
-    ...admin,
+    ...adminConfig,
```
riqwan, line 19: since this is optional 
```suggestion
    ...(adminConfig || {}),
```

packages/medusa/src/loaders/helpers/resolve-plugins.ts
```diff
@@ -126,7 +133,8 @@ export function getResolvedPlugins(
     ]
   }
 
-  const resolved = plugins.map((plugin) => {
+  const extensionDirectoryPath = "dist"
+  const resolved = configModule?.plugins.map((plugin) => {
```
riqwan, line 137: sugg: This might break if plugins are not added to the config

```suggestion
  const resolved = (configModule?.plugins || []).map((plugin) => {
```

packages/medusa/src/loaders/helpers/routing/__fixtures__/server/index.ts
```diff
@@ -42,7 +44,7 @@ export const createServer = async (rootDir) => {
     )[moduleKey]
   })
 
-  const container = createContainer()
+  const container = createContainer() as MedusaContainer
```
riqwan: sugg: I think we have a `createMedusaContainer()` somewhere

packages/medusa/src/loaders/medusa-app.ts
```diff
@@ -141,15 +142,10 @@ export async function revertMedusaApp({
   })
 }
 
-export const loadMedusaApp = async (
+export const loadModules = async (
```
riqwan: thought: might be relevant to keep this as `loadMedusaApp`. MedusaApp in modules-sdk is the one loading the modules, among other things and here it just seems to be initializing it.

riqwan (commented):
💯 will continue tomorrow

packages/medusa/src/loaders/medusa-app.ts
thetutlage: From what I saw was this method was just loading the modules and rest of the stuff like loading routes, workflows are happened somewhere else.

Maybe I am wrong, but I thought `loadingApp` means, loading everything and not just modules :)

packages/medusa/src/loaders/helpers/routing/__fixtures__/server/index.ts
thetutlage: Sure, lemme check

packages/medusa/src/loaders/admin.ts
thetutlage: Will it matter? Because, you can spread `undefined` values too.

packages/medusa/src/loaders/helpers/resolve-plugins.ts
thetutlage: Yeah. Makes sense

thetutlage: Okay. Just checked. It looks like the `plugins` property in `ConfigModule` type is required. Also configLoader returns an empty plugins array when one is not defined.

https://github.com/medusajs/medusa/blob/develop/packages/core/types/src/common/config-module.ts#L785

https://github.com/medusajs/medusa/blob/develop/packages/medusa/src/loaders/config.ts#L104

But still, if I am missing something, then do let me know :)

thetutlage:
Tests were passing before merging develop and right now it seems like some payments related tests are failing. Can someone point me in the direction on how to fix it?

olivermrbl:
Nice clean-up!

Regarding the added back Subscribers and API Routes registration, would you be up for running a few tests in your own project? 

You can publish snapshot versions of the changes in your PR by commenting `/snapshot-this`. Then, after ~2-3 mins a snapshot version will be added to the PR as a comment. 

Not super important though, but thought I'd share so you know about the snapshots going forward.

packages/medusa/src/loaders/admin.ts
thetutlage: Cleaned them up and kept it on

adrien2p:
One question, couldn't we use something like `process[Symbol.for("ts-node.register.instance")]` in order to identify if we are running under ts-node or not? In that case, when true we know that we have to discover the resources from the src and otherwise from the dist, we can still store it in the config for our convinience but don't think it should be made available in the medusa config wdyt?

packages/core/types/src/common/config-module.ts
```diff
@@ -862,6 +862,10 @@ export type ConfigModule = {
    * :::
    */
   featureFlags: Record<string, boolean | string>
+
+  directories?: {
```
adrien2p, line 866: Q: this one goes with my main question :D

packages/medusa/src/loaders/api.ts
```diff
@@ -35,14 +33,27 @@ export default async ({ app, configModule }: Options) => {
...
     }).load()
   } catch (err) {
     throw Error(
-      "An error occurred while registering Medusa Core API Routes. See error in logs for more details."
+      "An error occurred while registering Medusa Core API Routes. See error in logs for more details.",
+      { cause: err }
     )
   }
 
+  await Promise.all(
```
adrien2p, line 50: Question: I believe the route from the plugins was loaded first in v1, making it possible to override or prepend middlewares to an existing route from the core. Here, we will not be able to to that anymore. I believe it requires a bit more discussion but I wanted to mention it there, wdyt?

packages/medusa/src/loaders/helpers/register-workflows.ts
```diff
@@ -1,21 +1,20 @@
...
-  const [resolved] =
-    getResolvedPlugins(rootDirectory, configModule, "dist", true) || []
-  await registerWorkflows(resolved)
+export async function registerWorkflows(
+  plugins: PluginDetails[]
+): Promise<void> {
+  await Promise.all(
+    plugins.map(async (pluginDetails) => {
+      const files = glob.sync(
+        `${pluginDetails.resolve}/workflows/*(.js|.ts|.mjs)`,
```
adrien2p: suggestion: I believe we should exclude definition and map file wdyt?

adrien2p: Q: Does the resolve include the dist or src?

packages/medusa/src/loaders/helpers/routing/__fixtures__/server/index.ts
```diff
@@ -3,19 +3,24 @@ import {
...
 import querystring from "querystring"
 import supertest from "supertest"
 import apiLoader from "../../../../api"
+import { getResolvedPlugins } from "../../../../helpers/resolve-plugins"
 import featureFlagLoader, { featureFlagRouter } from "../../../../feature-flags"
 import passportLoader from "../../../../passport"
 
 import RoutesLoader from "../.."
 import { config } from "../mocks"
+import { MedusaContainer } from "medusa-core-utils"
```
adrien2p: todo: I suggest to use the one from the types package, this one should be removed

thetutlage:
@adrien2p 

Yeah we can do that. But, if you ask me, I would love the internals of the Medusa to be not such aware of where the source code lives. It should come from some config. 

The end-user don't have to write this config (if we want to keep things minimal), but the core logic should be free of knowledge about the directory structure.

I know we are not there yet, but atleast that's my dream end state.

packages/medusa/src/loaders/index.ts
```diff
@@ -29,17 +31,43 @@ const isWorkerMode = (configModule) => {
...
+    container
+  ).load()
+
+  /**
+   * Load subscribers from all the plugins.
+   */
+  await Promise.all(
+    plugins.map(async (pluginDetails) => {
+      const files = glob.sync(
+        `${pluginDetails.resolve}/subscribers/*(.js|.ts|.mjs)`,
```
adrien2p: Q:should we exclude definition and map in those glob?

adrien2p:
> @adrien2p
> 
> Yeah we can do that. But, if you ask me, I would love the internals of the Medusa to be not such aware of where the source code lives. It should come from some config.
> 
> The end-user don't have to write this config (if we want to keep things minimal), but the core logic should be free of knowledge about the directory structure.
> 
> I know we are not there yet, but atleast that's my dream end state.

What I thought is that it could be part of the config loader, not getting this value from the user but instead conpute it, in that way, the core is not aware of it and can still be based on the config module but we do not make it available to the end user. does that make sense?

packages/medusa/src/loaders/admin.ts
riqwan: TIL 😄

packages/medusa/src/loaders/helpers/register-workflows.ts
riqwan: I think barrel files as well in the exclusion

packages/core/types/src/common/config-module.ts
thetutlage: For now it goes in the `medusa-config` file. But, we can remove it later and go for detection of `ts-node`. But, for now it will allow to experiment with `TS-Node + SWC` and see if it fits our expectations.

adrien2p: Lets do that and see later then 👍

packages/medusa/src/loaders/api.ts
thetutlage: No idea. I am not aware how things were done in v1 :)

But, if plugin is loaded after, wouldn't it be able to override middleware loaded before it?

packages/medusa/src/loaders/helpers/register-workflows.ts
thetutlage: > Q: Does the resolve include the dist or src?

I think it will be both. In development it will be from the `src` folder and in prod it will be from the compiled output, ie `dist`.

packages/medusa/src/loaders/helpers/routing/__fixtures__/server/index.ts
thetutlage: 👍

packages/medusa/src/loaders/medusa-app.ts
riqwan: MedusaApp is a generic concept that can be run outside of the core as well, like a serverless func. I know the naming is confusing because the core can be considered an app as well, but this is a different concept. No strong feelings about the naming of this, just thought I'll put it out there. :)

packages/medusa/src/loaders/api.ts
adrien2p: No i dont think so as it will be added after therefore in order to resolve a request express will go through the matchers and encounter the core middlewares and then the core route which returns and stop there, so the plugins wont be reached for this scenario

packages/medusa/src/loaders/medusa-app.ts
adrien2p: Yes indeed, it is possible to spinup the modules without going through the medusa package, which in fact is just holding the http layer and loaders. But i am fine with that for now

adrien2p:
We need to think about adding some tests as well as mentioned by @riqwan just to prevent regression, for the router there is some tests already i think

packages/medusa/src/loaders/medusa-app.ts
thetutlage: I have no strong feelings about the name either :). But, a bit confused with naming and what exactly this method is supposed to do 🤔

packages/medusa/src/loaders/api.ts
thetutlage: Okay, I think I will have to pair with you to understand this :)

thetutlage:
Most of the feedback has been implemented in this PR. There are a couple of things that I want to understand from others on a call.

So I will merge this PR, do the call and will create a new PR (if needed). Otherwise, this branch keeps on getting stale and I have resolve merge conflicts

packages/medusa/src/loaders/medusa-app.ts
carlos-r-l-rodrigues: To help you a bit with the naming: `MedusaApp` is the real "core" loading the modules (internal and soon external), setting up the links between them, and exposing a Remote Query and Remote Link instances.
Medusa (@medusajs/medusa) is the ecommerce application using all the modules, using all the workflows from `@medusajs/core-flows`, exposing a REST Api and loading custom things.

thetutlage: Okay. That makes sense.

carlos-r-l-rodrigues:
great stuff 🥇

adrien2p (approved):

PR #7373, opened 2024-05-20
chore(order): Paid Total and Refunded Total

carlos-r-l-rodrigues:


packages/modules/order/src/services/order-module-service.ts
```diff
@@ -2306,106 +2342,144 @@ export default class OrderModuleService<
...
-    const toUpdate = Array.isArray(idOrDataOrSelector)
-      ? idOrDataOrSelector
-      : [
-          {
-            selector: idOrDataOrSelector,
-            data: data,
-          } as OrderTypes.UpdateOrderTransactionWithSelectorDTO,
-        ]
+  @InjectTransactionManager("baseRepository_")
+  // @ts-ignore
```
sradevski, line 2365: There are several ts-ignore and `any`'s, let's see if we can resolve those.

packages/modules/order/src/services/order-module-service.ts
```diff
@@ -2306,106 +2342,144 @@ export default class OrderModuleService<
...
-        toUpdate.push({
-          ...data,
-          id: trx.id,
-        })
-      })
-    }
+      if (!trxs.length) {
+        return
+      }
+      transformPropertiesToBigNumber(trxs)
```
sradevski, line 2454: I thought conversion to BN happens automatically?

packages/modules/order/src/services/order-module-service.ts
carlos-r-l-rodrigues: it fixed the types but will keep the ignore as these methods are conflicting because TS is recoginizing them as `member property` and when declaring as `member functions` it complains.
I won't spend much time in it.

packages/modules/order/src/services/order-module-service.ts
carlos-r-l-rodrigues: This property `totals` is a generic JSONB object, so the conversion won't happen automatically as it does when the entity is a Big Number instance on the model.
This helper will basically create the `raw_` properties of their correspondent values.

olivermrbl (approved):
These transactions are created under the hood right? Consumers won't really know of them

packages/modules/order/src/services/order-module-service.ts
sradevski: Got it 👍

carlos-r-l-rodrigues:
> These transactions are created under the hood right? Consumers won't really know of them

Yeah, in Medusa they will be created when payments are captured or refunded. But the module will have the methods available to be used elsewhere if needed.

PR #7374, opened 2024-05-20
feat(dashboard): restructure create product flow

fPolic:
**WHAT**
- restructure the product create flow

**Other**
- [x] fix pricing formating on create/edit
- [x] make edit pricing form work properly 

**NOTE**
- inventory kit tab is incomplete ATM until support is added to the workflow


olivermrbl:
Very nice!

Minor thing: We should default to having all variants checked in the first step of the flow

olivermrbl:
Also, another thing: When selecting "No variants", we should create a default variant under the hood. The details should just say "Default variant" etc.

I should have mentioned this earlier, my bad.

fPolic:
> Minor thing: We should default to having all variants checked in the first step of the flow

@olivermrbl should this be for the first option only or? What if the user creates an option, selects only some variants for creation and then creates a new option, which variants are checked then?

olivermrbl:
If we can have a control flow like so, that would be great:

Scenario 1:
- Create option
- *All variants are selected

Scenario 2:
- Create option
- *All variants are selected
- Select only two variants
- Create _new_ option
- The same two remain selected

Scenario 3:
- Create option
- *All variants are selected
- Create _new_ option
- *All variants (old + new) are selected

fPolic:
> Also, another thing: When selecting "No variants", we should create a default variant under the hood. The details should just say "Default variant" etc.

@olivermrbl we create that on the admin, right? Do we create the default variant right after the first step is complete or when we send the payload? Because if we create a default variant for the user should we also create default pricing and inventory stuff for that variant?

olivermrbl:
> we create that on the admin, right? Do we create the default variant right after the first step is complete or when we send the payload? Because if we create a default variant for the user should we also create default pricing and inventory stuff for that variant?

Yeah, we should still allow merchants to specify pricing and inventory for the default variant. Something like [this](https://www.figma.com/design/kHUEEgotLYDLpW8n4Q3Xre/Products-%2B-Inventory-v2-Q2-2024?node-id=14882-18706&t=n6Pd4zcJc8y9smbL-4). Lemme just validate the design with Ludvig

packages/admin-next/dashboard/src/routes/products/product-create/components/product-create-form/product-create-form.tsx
```diff
@@ -105,12 +107,12 @@ export const ProductCreateForm = () => {
...
         handleSuccess(`../${product.id}`)
       } catch (error) {
-        if (isFetchError(error) && error.status === 400) {
-          toast.error(t("general.error"), {
-            description: error.message,
-            dismissLabel: t("general.close"),
-          })
-        }
+        // if (isFetchError(error) && error.status === 400) {
+        toast.error(t("general.error"), {
```
adrien2p: Suggestion: could we revert that, add a return in the if and add the following after the id
```ts
          toast.error(t("general.error"), {
            description: t("errors.serverError"),
            dismissLabel: t("general.close"),
          })
```

fPolic:
@olivermrbl @kasperkristensen would love to hear your thoughts on this one

olivermrbl:
Will give this a spin today!

Are there any larger todos remaining?

fPolic:
No, should be ready for review

packages/admin-next/dashboard/src/routes/products/product-create/components/product-create-details-form/components/product-create-details-general-section/product-create-general-section.tsx
```diff
@@ -18,7 +18,7 @@ export const ProductCreateGeneralSection = ({
   return (
     <div id="general" className="flex flex-col gap-y-8">
       <div className="flex flex-col gap-y-2">
-        <div className="grid grid-cols-2 gap-x-4">
+        <div className="grid grid-cols-3 gap-x-4">
```
kasperkristensen:
```suggestion
        <div className="grid grid-cols-1 md:grid-cols-3 gap-4">
```

Should be responsive

kasperkristensen (commented):
Couple of things:

We need to initialize the values of each field when we create a variant in-memory. If not we get two issues: The input changes from uncontrolled to controlled, and if you add a value to a field that isn't inited, and then hit cmd + z it will insert the value "undefined" into the field.

The grid gets laggy when you have many variants, and you are horizontally scrolled so all of the boolean cells are rendered. I saw a good amount of performance increase when I changed the country select cell to render a placeholder when its not the anchor cell. Think we should do this for all "complex" cells, but fine to leave for now, as the grid is still WIP (need to address the things I've outlined in the comment above the component).

fPolic:
> The grid gets laggy when you have many variants...

I addressed the other comment but I think that we should address bulk editor WIP stuff and performance optimisations in a separate ticket

kasperkristensen (approved):
Lets merge it in, I think most if not all current issues are related to the DataGrid, which we can tackle later.

PR #7376, opened 2024-05-20
feat(medusa,medusa-cli): add an exec command

shahednasser:
**Edit: After discussing it, the command is now renamed to `exec`.**

Added a `script` command to Medusa's CLI tool that accepts a file parameter as a path to run its exported function. It passes the function the Medusa Container as a parameter to the function.

An example of a script:

```ts
// src/helpers/hello-world.ts
import { ExecArgs } from "@medusajs/types"

export default function seed({
  container,
  args
}: ExecArgs) {
  console.log("Hello, world!")
}
```

You can then run it with the following command:

```bash
yarn build
npx medusa script ./dist/helpers/hello-world.js
```


packages/medusa/src/loaders/index.ts
```diff
@@ -22,7 +22,6 @@ import { SubscriberLoader } from "./helpers/subscribers"
 type Options = {
   directory: string
   expressApp: Express
-  isTest: boolean
```
shahednasser, line 25: Note: removed this as it's a required option for the `loaders` function but didn't find it used anywhere

packages/cli/medusa-cli/src/create-cli.ts
```diff
@@ -321,6 +321,19 @@ function buildLocalCommands(cli, isLocalProject) {
         })
       ),
     })
+    .command({
+      command: `script [file]`,
```
sradevski: How about `exec` or `execute`? `script` is not a verb so it reads a bit strangely.

packages/medusa/src/commands/script.ts
```diff
@@ -0,0 +1,30 @@
+import loaders from "../loaders"
+import express from "express"
+import path from "path"
+
+type Options = {
+  file: string
+}
+
+export default async function script({ file }: Options) {
```
sradevski: NIT: Adding some logs would be helpful when debugging

sradevski (approved):

packages/cli/medusa-cli/src/create-cli.ts
olivermrbl: The idea was to expose a command with an immediately obvious purpose. That said, I imagine most are familiar with the concept of `exec`, so I am also fine with that. @shahednasser made a similar comment about it actually.

cc @srindom

shahednasser:
I addressed the PR feedback and renamed the command to `exec` (I think it makes a lot of sense 😄)

@olivermrbl I also added a new feature to it, which is the ability to pass arguments from the command line to the script. I imagine that a lot of usage for this command could be to create a custom CLI command, which, in many cases, would perform something with CLI arguments. Now, any arguments passed after the file like this:

```
npx medusa bash ./dist/helpers/hello-world.js test
```

The script receives it as a second parameter:

```ts
// src/helpers/hello-world.ts
import { MedusaContainer } from "@medusajs/medusa"

export default function seed(container: MedusaContainer, args:  string[]) {
  console.log("Hello, world!")
}
```

Let me know what you think

olivermrbl (commented):
Think this is great @shahednasser!

QQ: Should we use the `Logger` instead of the console?

adrien2p (approved):
LGTM 💪 should we add a type somewhere for the input args for the user to use in there function definition? just so that they know how they will receive the args from the command line

shahednasser:
> QQ: Should we use the Logger instead of the console?

Makes sense, I'll make the change 👍🏻

> should we add a type somewhere for the input args for the user to use in there function definition?

@adrien2p Yea we can do that, maybe change the parameter they receive to be an object with a `container` and `args` properties instead of two parameters

adrien2p:
> > QQ: Should we use the Logger instead of the console?
> 
> Makes sense, I'll make the change 👍🏻
> 
> > should we add a type somewhere for the input args for the user to use in there function definition?
> 
> @adrien2p Yea we can do that, maybe change the parameter they receive to be an object with a `container` and `args` properties instead of two parameters

Yes that's perfect. Also, I can see in the description that MedusaContainer is imported from medusa but it comes from the types package 👍

shahednasser:
@adrien2p true, but it now changes to `ExecArgs` from the `types` package anyway. Updated the description to reflect that 👍🏻

olivermrbl:
Merging now, if cool with you all?

packages/medusa/src/commands/exec.ts
```diff
@@ -0,0 +1,42 @@
...
+    // set worker mode
+    process.env.MEDUSA_WORKER_MODE = "worker"
+
+    const { container } = await loaders({
+      directory,
+      expressApp: app,
+    })
+
+    const scriptFile = (await import(path.resolve(directory, file))).default
+
```
srindom, line 27: thought(non-blocking): maybe some light validation would be good so we can shoot off a nice error message.

shahednasser:
Fine by me 👍🏻

packages/medusa/src/commands/exec.ts
shahednasser: Will add it in a separate PR later to not block merging for now 👍🏻

PR #7378, opened 2024-05-21
fix: Minor updates to the http typings

sradevski:


packages/core/types/src/http/order/common.ts
```diff
@@ -98,16 +103,26 @@ interface BaseOrderShippingMethod {
   metadata: Record<string, unknown> | null
   tax_lines?: BaseOrderShippingMethodTaxLine[]
   adjustments?: BaseOrderShippingMethodAdjustment[]
+  original_total: BigNumberValue
```
adrien2p, line 106: Note: I think the totals are optional and present if selected, maybe they should be marked optional and selectable ?

adrien2p (approved):

packages/core/types/src/http/order/common.ts
sradevski: @adrien2p I took it from the orderDTO, so I assumed it's correct. Since this is an order, maybe they are stored in the DB at this point, and always available? I'll check just to be sure

adrien2p: Sure, @carlos-r-l-rodrigues will know better for that part

sradevski: @adrien2p I see they are non-optional all over the place in the order module, let's revisit as a bigger part of improving the orders in general

PR #7381, opened 2024-05-21
feat(create-medusa-app): install v2 by default

shahednasser:
- Remove `v2` option and install V2 starter branch by default.
- Use new `exec` command to seed data
- Install v2 branch of next.js starter if the option is provided.
- Change the opened browser to `localhost:9000/app`.
- Added a bunch of todos for onboarding flows once we have that

### DIFF
--- a/.github/workflows/test-cli-with-database.yml
+++ b/.github/workflows/test-cli-with-database.yml
@@ -52,7 +52,7 @@ jobs:
 
      - name: Create Medusa project
        run: |
-         medusa new cli-test --skip-db --v2
+         medusa new cli-test --skip-db --v2 --branch feat/v2-ci
        working-directory: ..
 
      - name: run medusa dev
--- a/integration-tests/modules/__tests__/order/workflows/create-return.spec.ts
+++ b/integration-tests/modules/__tests__/order/workflows/create-return.spec.ts
@@ -1,5 +1,10 @@
-import { ModuleRegistrationName, Modules } from "@medusajs/modules-sdk"
 import {
+  ModuleRegistrationName,
+  Modules,
+  RemoteLink,
+} from "@medusajs/modules-sdk"
+import {
+  FulfillmentSetDTO,
   FulfillmentWorkflow,
   IOrderModuleService,
   IRegionModuleService,
@@ -208,6 +213,7 @@ async function prepareDataFixtures({ container }) {
     salesChannel,
     location,
     product,
+    fulfillmentSet,
   }
 }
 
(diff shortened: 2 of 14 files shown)


packages/cli/create-medusa-app/src/utils/nextjs-utils.ts
```diff
@@ -56,15 +56,24 @@ export async function installNextjsStarter({
...
-            npm_config_yes: "yes",
-          },
+          env: process.env,
+        },
+      ],
+      { verbose }
+    )
+    await execute(
+      [
+        `npm install`,
```
sradevski: We currently use yarn in the starter, and ideally we use it here as well (although it might not be installed on some machines, which is the main issue. I guess). I guess if we use create-next-app it won't matter

sradevski (approved):
LGTM!

packages/cli/create-medusa-app/src/utils/nextjs-utils.ts
shahednasser: Got it, I can do what we do when installing Medusa's dependencies which is try to install with `yarn` first and, if not available, then install with `npm`

sradevski: It might not matter much, it's just the lock file being interpreted correctly that MIGHT cause issues, but in general it shouldn't.

thetutlage:
Curious:

Does this mean anyone publicly creating a new app will get `v2`?

shahednasser:
@thetutlage only if they're using the `preview` version.

olivermrbl (approved):

packages/cli/create-medusa-app/src/utils/nextjs-utils.ts
shahednasser: Changed it now to try to install with `yarn` first.

jelen07:
When I'm trying to install v2 via `npx create-medusa-app@preview`, I'm getting:

```bash
Need to install the following packages:
create-medusa-app@1.2.9-preview-20240820060557
Ok to proceed? (y) n
```
How can I install the the v2? 🤔

carlos-r-l-rodrigues:
> When I'm trying to install v2 via `npx create-medusa-app@preview`, I'm getting:
> 
> ```shell
> Need to install the following packages:
> create-medusa-app@1.2.9-preview-20240820060557
> Ok to proceed? (y) n
> ```
> 
> How can I install the the v2? 🤔

Just type `y`to confirm 😄 
The package versions are not bumped to 2.x yet, they are all under `-preview` suffix

PR #7382, opened 2024-05-21
docs-util: fixes to base OAS + circular-patch redocly plugin

shahednasser:
- Remove `MultipleErrors` schema which we had in V1 API
- Fix circular patch redocly plugin error when no match is found for a schema
- Fix duplicate parameters.


shahednasser:

carlos-r-l-rodrigues (approved):

PR #7383, opened 2024-05-21
feat(dashboard,admin-sdk,admin-shared,admin-vite-plugin): Add support for UI extensions

kasperkristensen:
**What**
- Re-adds support for UI Widgets and UI Routes.
- Removes UI Settings as a concept. Settings are instead added under `/src/admin/routes/settings`. This change was made, as there are no difference between what a route and a settings route is in the new admin, both now uses sidebar menu items for navigation, and in the new dashboard settings can now have multiple levels of nesting - `/settings/provider/:id/edit` etc.

**Examples**
```tsx
// /src/admin/widgets/test.tsx
import { defineWidgetConfig } from "@medusajs/admin-shared"

const Widget = () => {
  return (
    <div>Test</div>
  )
}

export const config = defineWidgetConfig({
  zone: "product.details.before"
})

export default Widget
```
```tsx
// /src/admin/routes/test/page.tsx
import { defineRouteConfig } from "@medusajs/admin-shared";

const TestPage = () => {
  return <div>Test Page!</div>;
};

    label: "Test"
})

export default TestPage;
```
```tsx
// /src/admin/routes/settings/test/page.tsx
import { defineRouteConfig } from "@medusajs/admin-shared";

const TestSettingsPage = () => {
  return <div>Test Settings Page!</div>;
};

    label: "Test"
})

export default TestSettingsPage;
```

**Note**
- I wanted to re-export the `defineWidgetConfig` and `defineRouteConfig` from `@medusajs/medusa`, but that is not possible as the medusa package leaks Node code into the admin build. For the time being the helpers can be imported from `@medusajs/admin-shared`.

**TODO in follow up PR(s)**
- Add InjectionZones to all domains in `/dashboard`.
- Pass props to widgets, such as the domain entity. I have omitted it for now, as we need to discuss the pattern we want to follow for this. Same as before, eg. we pass <Product> to the `product.details.*` widgets, or do we want people to fetch the data themselves, etc.
- Validate admin `path`, to prevent serving the admin from reserved paths: `/admin`, `/store`, `/auth`, `/`.
- Add better error messages to Vite plugin. Currently, if the plugin fails to parse a file, it will silently ignore it, we should alert the user so they can fix the file if it needs to be considered for extensions.
- Add support for passing breadcrumbs and loaders to UI Routes.
- (potentially) Add support for manually changing how routes are nested. The intend would be to allow developers to use the same patterns we do, such as rendering modals in a separate route through a Outlet. Currently, we nest routes based on their path names, but to follow our pattern developers would need to be allowed to override the nesting, an example would be to add a `/create` route as a child of the another route in order to render the modal on top instead of in an entirely different route.


packages/admin-next/admin-sdk/src/lib/serve.ts
```diff
@@ -41,6 +42,8 @@ export async function serve(options: ServeOptions) {
     res.setHeader("Vary", "Origin, Cache-Control")
   }
 
+  router.use(compression())
```
sradevski, line 45: Should we pull the compression settings from the config (I see we have some there)?

packages/admin-next/admin-shared/package.json
```diff
@@ -5,14 +5,17 @@
   "author": "Kasper Kristensen <kasper@medusajs.com>",
```
sradevski, line 5: Maybe the timing is not the best, but I feel like we can do better with the naming. To me
`admin-sdk` would be closer to what `admin-shared` does. And the current `admin-sdk` can be more something like `admin-builder`. I am not saying we need to do it in this PR, but something to think about before officially releasing v2.

packages/admin-next/dashboard/src/components/layout/settings-layout/settings-layout.tsx
```diff
@@ -80,9 +84,27 @@ const useDeveloperRoutes = (): NavItemProps[] => {
   )
 }
 
+const useExtensionRoutes = (): NavItemProps[] => {
+  const links = routes.links
+  const settingsLinks = links.filter((link: any) =>
```
sradevski: We can move the filtering to the useMemo?

packages/admin-next/dashboard/src/lib/extension-helpers.ts
```diff
@@ -0,0 +1,56 @@
...
+  ignore?: string
+): RouteObject[] => {
+  const root: RouteObject[] = []
+
+  const addRoute = (
+    pathSegments: string[],
+    file: string,
+    currentLevel: RouteObject[]
+  ) => {
+    if (pathSegments.length === 0) {
```
sradevski:
```suggestion
    if (!pathSegments.length) {
```

packages/admin-next/admin-sdk/src/lib/serve.ts
kasperkristensen: I think we might want to have a separate compression config in `projectConfig.admin` that we could pass here, as I don't necessarily think the compressions options you want generally for your server === the compression you want for the admin dashboard. But I think we can leave that for later, the default compression options seems to work great for this use case, but we should definetly open them up to people if they want to modify it for whatever reason.

packages/admin-next/admin-shared/package.json
kasperkristensen: Agree that the naming can be better, and something we need to revisit, but would hold off on it to post RC, as it requires we update the packages a couple of different places.

packages/admin-next/admin-shared/src/extensions/virtual/utils.ts
```diff
@@ -0,0 +1,29 @@
+import { InjectionZone } from "../widgets"
+
+const PREFIX = "virtual:medusa/"
+
+export const id = (name: string) => {
```
adrien2p: nitsuggestion: id sounds like a value but it is a function, could we name it like resolveId for example?

packages/admin-next/dashboard/src/components/layout/main-layout/main-layout.tsx
```diff
@@ -157,35 +160,39 @@ const CoreRouteSection = () => {
...
-  if (!extensions.links || extensions.links.length === 0) {
+  const { t } = useTranslation()
+
+  const links = routes.links
+
+  const extensionLinks = links.filter(
+    (link) => !settingsRouteRegex.test(link.path)
+  )
+
+  if (extensionLinks.length === 0) {
```
adrien2p:
```suggestion
  if (!extensionLinks.length) {
```

packages/admin-next/dashboard/src/components/layout/main-layout/main-layout.tsx
```diff
@@ -196,7 +203,7 @@ const ExtensionRouteSection = () => {
           </div>
           <Collapsible.Content>
             <div className="flex flex-col gap-y-1 py-1 pb-4">
-              {extensions.links.map((link) => {
+              {extensionLinks.map((link: any) => {
```
adrien2p: nit suggestion: should we add a type here?

packages/admin-next/dashboard/src/lib/extension-helpers.ts
```diff
@@ -0,0 +1,56 @@
...
+    const [currentSegment, ...remainingSegments] = pathSegments
+    let route = currentLevel.find((r) => r.path === currentSegment)
+
+    if (!route) {
+      route = { path: currentSegment, children: [] }
+      currentLevel.push(route)
+    }
+
+    if (remainingSegments.length === 0) {
+      route.children = route.children || []
```
adrien2p:
```suggestion
    if (!remainingSegments.length) {
      route.children ||= []
```

packages/admin-next/dashboard/src/lib/extension-helpers.ts
```diff
@@ -0,0 +1,56 @@
...
+      route.children.push({
+        path: "",
+        async lazy() {
+          const { default: Component } = await import(/* @vite-ignore */ file)
+          return { Component }
+        },
+      })
+    } else {
+      route.children = route.children || []
```
adrien2p:
```suggestion
      route.children ||= []
```

packages/medusa/src/loaders/admin.ts
```diff
@@ -1,22 +1,43 @@
 import { AdminOptions, ConfigModule } from "@medusajs/types"
 import { Express } from "express"
+import fs from "fs"
+import path from "path"
 
 type Options = {
   app: Express
-  adminConfig: ConfigModule["admin"]
+  configModule: ConfigModule
```
adrien2p, line 8:
```suggestion
  configModule: Pick<ConfigModule>
```

The idea being that you are not require to pass the entire configuration as only the admin is consumed

adrien2p (approved):
Wonderful

packages/medusa/src/loaders/admin.ts
kasperkristensen: When we have a "plugins" concept again, we will need to read that to resolve where to look for possible admin extensions. So right now we only need the admin part later we will need more :)

olivermrbl:
>  For the time being the helpers can be imported from @medusajs/admin-shared.

@kasperkristensen, this means users will have to separately install `@medusajs/admin-shared` in their project right? This package is not depended on by any already-installed packages as far as I can see.

If so, could I get you to include a changeset for it? I don't think we have that, so I can't test

packages/medusa/src/loaders/admin.ts
```diff
@@ -1,22 +1,43 @@
...
+export default async function adminLoader({
+  app,
+  configModule,
+  rootDirectory,
+}: Options) {
+  const { admin } = configModule
+
+  const sources: string[] = []
+
+  const projectSource = path.join(rootDirectory, "src", "admin")
```
riqwan, line 28: q: Not entirely sure myself, but will the `yarn dev` changes made on the project affect this at all in discovering these files?

riqwan (approved):
nice!

packages/medusa/src/loaders/admin.ts
adrien2p: oh yes thats true 😂 my bad 💪

kasperkristensen:
> > For the time being the helpers can be imported from @medusajs/admin-shared.
> 
> @kasperkristensen, this means users will have to separately install `@medusajs/admin-shared` in their project right? This package is not depended on by any already-installed packages as far as I can see.
> 
> If so, could I get you to include a changeset for it? I don't think we have that, so I can't test

`admin-sdk` depends on it, so it will be in your node_modules. Auto import might not suggest it for you, but if you type it out it should work.

olivermrbl:
> `admin-sdk` depends on it, so it will be in your node_modules. Auto import might not suggest it for you, but if you type it out it should work.

Ah lol sorry, a versioning mismatch on my end caused my lock-file to be incorrect. Thx!

packages/medusa/src/loaders/admin.ts
kasperkristensen: I think we are still getting the rootDirectory param, no? And we are reading from src files not dist, so we shouldn't have an issue due to missing dist files.

adrien2p: yeah in theory it shouldnt change anything here, the root dir does not change 👍

kasperkristensen:
@olivermrbl are you cool with merging this now, so I can jump on adding InjectionZones back to admin

olivermrbl:
Let's do it! 🚢

adrien2p (approved):

packages/medusa/src/loaders/admin.ts
riqwan: Yes, you're right!

PR #7384, opened 2024-05-21
feat(medusa-cli): added branch option

shahednasser:
~Added CI option that clones the `feat/v2-ci` branch of the starter. This branch has the admin disabled to avoid errors.~

Edit: Added `branch` option to specify the branch to clone of the repo

### DIFF
--- a/packages/cli/medusa-cli/src/commands/new.ts
+++ b/packages/cli/medusa-cli/src/commands/new.ts
@@ -179,11 +179,11 @@ const clone = async (hostInfo, rootPath, v2 = false, inputBranch) => {
     url = hostInfo.https({ noCommittish: true, noGitPlus: true })
   }
 
-const branch = (inputBranch || hostInfo.committish ? [`-b`, inputBranch || hostInfo.committish] : [])
+  let branch = (inputBranch || hostInfo.committish ? [`-b`, inputBranch || hostInfo.committish] : [])
 
-if (v2) {
-  branch = [`-b`, inputBranch || "feat/v2"]
-}
+  if (v2) {
+    branch = [`-b`, inputBranch || "feat/v2"]
+  }
 
   const createAct = reporter.activity(`Creating new project from git: ${url}`)
 


packages/cli/medusa-cli/src/commands/new.ts
```diff
@@ -169,7 +169,7 @@ const copy = async (starterPath, rootPath) => {
 }
 
 // Clones starter from URI.
-const clone = async (hostInfo, rootPath, v2 = false) => {
+const clone = async (hostInfo, rootPath, v2 = false, ci = false) => {
```
olivermrbl: thought: Would it make sense to be able to pass a repo URL (with a branch) instead? That would allow us to eliminate all these conditionals further down. And we won't have to change the command in the future.

Update: Not sure if this is possible actually, but thought I'd throw it out there.

shahednasser: I think it's possible. We would instead just supply a `branch` option and use it

shahednasser: @olivermrbl changed it to a `branch` option. let me know what you think

shahednasser:
@olivermrbl should I re-disable the admin in the `feat/v2` branch until we get this working and published on `preview` so to not have all PRs failing?

packages/cli/medusa-cli/src/create-cli.ts
```diff
@@ -119,10 +119,9 @@ function buildLocalCommands(cli, isLocalProject) {
             describe: `Install Medusa with the V2 feature flag enabled. WARNING: Medusa V2 is still in development and shouldn't be used in production.`,
             default: false,
           })
-          .option(`ci`, {
-            type: `boolean`,
+          .option(`branch`, {
+            type: `string`,
             describe: `Indicate that the installation is part of the CI Pipeline`,
```
sradevski: This needs updating as well

shahednasser: Nice catch 😅 fixed

packages/cli/medusa-cli/src/commands/new.ts
olivermrbl: Think something like this could work. Fine to move forward with this. 

What I had in mind was actually something like this: `medusa new ci https://github.com/medusajs/medusa-starter-default/[branch]`

But this is what I am unsure if we can do without having to do too many gymnastics around extracting the branch from the url.

olivermrbl (approved):

packages/cli/medusa-cli/src/commands/new.ts
shahednasser: Makes sense, actually I think it would be nice to re-do the entire CLI tool because there are a lot of things that could be fixed/improved, but a task for another day 😄

sradevski (approved):

packages/cli/medusa-cli/src/commands/new.ts
```diff
@@ -179,8 +179,8 @@ const clone = async (hostInfo, rootPath, v2 = false) => {
     url = hostInfo.https({ noCommittish: true, noGitPlus: true })
   }
 
-  const branch = v2 ? [`-b`, "feat/v2"] : 
-    hostInfo.committish ? [`-b`, hostInfo.committish] : []
+  const branch = v2 ? [`-b`, inputBranch || "feat/v2"]
+      : (inputBranch || hostInfo.committish ? [`-b`, inputBranch || hostInfo.committish] : [])
```
adrien2p:
```suggestion
let branch = (inputBranch || hostInfo.committish ? [`-b`, inputBranch || hostInfo.committish] : [])

if (v2) {
  branch = [`-b`, inputBranch || "feat/v2"]
}
```

adrien2p: nit: just to make it a bit more digestable 😂

adrien2p (approved):

PR #7385, opened 2024-05-21
feat(order): fulfillment workflow

carlos-r-l-rodrigues:
FIXES: CORE-2162 CORE-2167 CORE-2041


integration-tests/modules/__tests__/order/workflows/create-return.spec.ts
```diff
@@ -394,7 +394,7 @@ medusaIntegrationTestRunner({
 
         await createReturnOrderWorkflow(container).run({
           input: createReturnOrderData,
-          throwOnError: false,
+          throwOnError: true,
```
adrien2p, line 397: nit: rm

packages/core/core-flows/src/fulfillment/workflows/create-fulfillment.ts
```diff
@@ -16,20 +10,6 @@ export const createFulfillmentWorkflow = createWorkflow(
...
-        return [
-          {
-            [Modules.ORDER]: { order_id: data.order_id },
-            [Modules.FULFILLMENT]: { fulfillment_id: data.fulfillment.id },
-          },
-        ]
-      }
-    )
-
-    createLinkStep(link)
```
adrien2p, line 31: suggestion: should we then remove the order_id from the input?

packages/core/core-flows/src/order/workflows/create-return.ts
```diff
@@ -7,28 +8,27 @@ import {
...
+import { createLinkStep, useRemoteQueryStep } from "../../common"
+import { createFulfillmentWorkflow } from "../../fulfillment"
 import { updateOrderTaxLinesStep } from "../steps"
 import { createReturnStep } from "../steps/create-return"
-import { createFulfillmentWorkflow } from "../../fulfillment"
 
 function throwIfOrderIsCancelled({ order }: { order: OrderDTO }) {
-  // TODO: need work, check canceled
-  if (false /*order.canceled_at*/) {
+  if (order.status === OrderStatus.CANCELED) {
```
adrien2p: note: as mentioned on the other pr we might move those functions to the utils dir next to the order workflow, so that we can create our different validation step in those workflow using those utils functions. wdyt?

packages/core/core-flows/src/order/workflows/create-return.ts
```diff
@@ -339,7 +339,21 @@ export const createReturnOrderWorkflow = createWorkflow(
       prepareFulfillmentData
     )
 
-    createFulfillmentWorkflow.runAsStep(fulfillmentData)
+    const returnFulfillment =
+      createFulfillmentWorkflow.runAsStep(fulfillmentData)
```
adrien2p: Q: I think you have changed it in the other pr right?

packages/modules/inventory-next/src/services/inventory.ts
```diff
@@ -799,33 +800,68 @@ export default class InventoryModuleService<
...
+        data.locationId,
+        data.adjustment,
+        context
+      )
+      results.push(result)
+
+      context.messageAggregator?.saveRawMessageData({
+        eventName: InventoryEvents.inventory_level_updated,
+        metadata: {
+          service: this.constructor.name,
```
adrien2p, line 855: discussion: I ve seen that in many places, the event service is the constructor name of the class, but on my side I use the Modules.MY_MODULE instead. I think it should be the module identifier rather than the class name cause in a distributed system I believe the module (e.g microservice) name is more relevant than the underlying service class name which can have a random name. wdyt? dont change it I ll do it as part of the event task

adrien2p (approved):

packages/core/core-flows/src/fulfillment/workflows/create-fulfillment.ts
carlos-r-l-rodrigues: Which input do you mean?

packages/core/core-flows/src/order/workflows/create-return.ts
carlos-r-l-rodrigues: yes, I did

packages/modules/inventory-next/src/services/inventory.ts
carlos-r-l-rodrigues: Yes, it has to be the Module definition key.
I didn't even pay attention to these events in this PR :D

packages/core/core-flows/src/fulfillment/workflows/create-fulfillment.ts
adrien2p: the workflow input

packages/modules/inventory-next/src/services/inventory.ts
adrien2p: ahah no worries, I started to change them in another pr

adrien2p: cc @olivermrbl just for info

packages/core/core-flows/src/fulfillment/workflows/create-fulfillment.ts
carlos-r-l-rodrigues: I removed it from `CreateFulfillmentWorkflowInput` but we need the order_id in the input to fetch the order.

PR #7387, opened 2024-05-21
feat(auth): Revamp authentication setup

sradevski:
Apologies for the big-bang PR, but it was much easier to do all of them in one go.

Here are the changes:
- Renamed AuthUser to AuthIdentity
- Revamped the auth context carried in the request
- Created a link between User and AuthIdentity
- Created a link between Customer and AuthIdentity
- Used links for authentication of user and customer


packages/medusa/src/api/auth/[scope]/[auth_provider]/callback/route.ts
```diff
@@ -22,14 +22,26 @@ export const GET = async (req: MedusaRequest, res: MedusaResponse) => {
...
+  const { success, error, authIdentity, successRedirectUrl } = authResult
 
   if (success) {
     const { http } = req.scope.resolve("configModule").projectConfig
 
     const { jwtSecret, jwtExpiresIn } = http
-
-    const token = jwt.sign(authUser, jwtSecret, { expiresIn: jwtExpiresIn })
+    // TODO: Dynamically determine the actor information
+    const token = generateJwtToken(
```
Comment (sradevski) [inline packages/medusa/src/api/auth/[scope]/[auth_provider]/callback/route.ts:47]: This obviously needs to be done, but I'd prefer to do it in a separate PR, as I have some suggestion for improvements of this endpoint that we can discuss separately.

packages/core/core-flows/src/customer/workflows/create-customer-account.ts
```diff
@@ -20,12 +21,19 @@ export const createCustomerAccountWorkflow = createWorkflow(
...
+        return [
+          {
+            [Modules.CUSTOMER]: { customer_id: data.customer.id },
+            [Modules.AUTH]: { auth_identity_id: data.authIdentityId },
+          },
+        ]
+      }
+    )
 
+    createLinkStep(link)
```
srindom, line 36: thought: think this part could use a better abstraction. Something like linkPermission or similar. Under the hood it wouldn't do anything differently but suspect people will have a hard time writing/understanding this

integration-tests/modules/__tests__/customer/store/create-customer.spec.ts
```diff
@@ -34,15 +34,14 @@ medusaIntegrationTestRunner({
         const authService: IAuthModuleService = appContainer.resolve(
           ModuleRegistrationName.AUTH
         )
-        const { http } =
-          appContainer.resolve("configModule").projectConfig
-        const authUser = await authService.create({
+        const { http } = appContainer.resolve("configModule").projectConfig
```
adrien2p: nit: ContainerRegistrationKeys.CONFIG_MODULE

packages/core/types/src/auth/service.ts
```diff
@@ -210,35 +210,38 @@ export interface IAuthModuleService extends IModuleService {
...
    *
-   * @param {UpdateAuthUserDTO[]} data - The attributes to update in the auth users.
+   * @param {UpdateAuthIdentityDTO[]} data - The attributes to update in the auth identitys.
    * @param {Context} sharedContext - A context used to share resources, such as transaction manager, between the application and the module.
-   * @returns {Promise<AuthUserDTO[]>} The updated auths.
+   * @returns {Promise<AuthIdentityDTO[]>} The updated auths.
    *
    * @example
-   * const authUsers = await authModuleService.update([
+   * const authIdentitys = await authModuleService.update([
```
adrien2p:
```suggestion
   * const authIdentities = await authModuleService.update([
```

adrien2p: nit: There might be over places with that typo  :D

packages/medusa/src/api/admin/users/me/route.ts
```diff
@@ -12,7 +12,7 @@ export const GET = async (
   req: AuthenticatedMedusaRequest,
   res: MedusaResponse
 ) => {
-  const id = req.auth.app_metadata.user_id
+  const id = req.auth_context.app_metadata.user_id
```
adrien2p: question: isn't the actor id more appropriate?

packages/medusa/src/loaders/helpers/routing/index.ts
```diff
@@ -645,27 +640,6 @@ export class RoutesLoader {
...
-         */
-        this.router.use(descriptor.route, requireCustomerAuthentication())
-      }
-
-      if (descriptor.config.shouldRequireAdminAuth) {
-        /**
-         * Require the admin to be authenticated
-         */
-        this.router.use(descriptor.route, authenticateLegacy())
-      }
```
adrien2p, line 667: Question: Why did we remove that?

packages/medusa/src/utils/middlewares/authenticate-middleware.ts
```diff
@@ -36,63 +36,65 @@ export const authenticate = (
...
-      (authUser?.app_metadata?.customer_id &&
-        stringEqualsOrRegexMatch(authScope, "store"))
-
+    // If the entity is authenticated, and it is a registered user/customer we can continue
     if (
-      authUser &&
-      (isRegistered || (!isRegistered && options.allowUnregistered))
+      authContext &&
+      !!authContext.actor_id &&
+      authContext.actor_type !== "unknown"
```
adrien2p:
```suggestion
      !!authContext?.actor_id &&
      authContext?.actor_type !== "unknown"
```

packages/medusa/src/utils/middlewares/authenticate-middleware.ts
```diff
@@ -36,63 +36,65 @@ export const authenticate = (
     next: NextFunction
   ): Promise<void> => {
     const authTypes = Array.isArray(authType) ? authType : [authType]
+    const scopes = Array.isArray(authScope) ? authScope : [authScope]
```
adrien2p, line 39:
```suggestion
    const scopes = Array.isArray(authScope) ? authScope : [authScope]
    const req_ = req as AuthenticatedMedusaRequest
```

then you can use `req_` without having to cast all the time 👍

packages/medusa/src/utils/middlewares/authenticate-middleware.ts
```diff
@@ -36,63 +36,65 @@ export const authenticate = (
...
-      }
+      ;(req as AuthenticatedMedusaRequest).auth_context = authContext
+      return next()
+    }
 
+    // If the entity is authenticated, but there is no user/customer yet, we can continue (eg. in the case of a user invite) if allow unregistered is set
+    if (
+      authContext &&
+      authContext.auth_identity_id &&
+      options.allowUnregistered
```
adrien2p:
```suggestion
      authContext?.auth_identity_id &&
      options.allowUnregistered
```

adrien2p (approved):
LGTM 💪 a few comments here and there

carlos-r-l-rodrigues (approved):
nit: can we have an Enum for the different `actor_type` values?

packages/medusa/src/types/routing.ts
```diff
@@ -59,15 +59,17 @@ export interface MedusaRequest<Body = unknown>
   pricingContext?: MedusaPricingContext
 }
 
+export interface AuthContext {
+  actor_id: string
+  actor_type: "api-key" | "user" | "customer" | "unknown"
```
srindom, line 65: thought: this is likely something people will want to extend. Can we come up with something that is more open?

packages/medusa/src/utils/middlewares/authenticate-middleware.ts
```diff
@@ -20,13 +15,18 @@ const API_KEY_AUTH = "api-key"
...
+
 type MedusaSession = {
-  auth_user: AuthUserDTO
-  scope: string
+  auth_context: AuthContext
 }
 
 export const authenticate = (
-  authScope: string | RegExp,
+  authScope: Scope | Scope[],
```
srindom, line 29: todo: this should allow anything. If I create my own auth scope - e.g vendor.

packages/core/types/src/auth/service.ts
sradevski: @adrien2p ah nice catch, yeah there very few due to poor find and replace strategy 😄

packages/medusa/src/api/admin/users/me/route.ts
sradevski: @adrien2p this is already fixed, it was a mistake that I noticed as well 😄

packages/medusa/src/loaders/helpers/routing/index.ts
sradevski: @adrien2p these specific middlewares didn't do anything, I will message everyone on the behavior we want tomorrow and add whatever logic we decide to have.

packages/core/core-flows/src/customer/workflows/create-customer-account.ts
sradevski: @srindom I am taking it one step at a time so as not to introduce too many changes at once, but I wanted to take it one step further, where you have a generic workflow like `createXAccount` and we use that for users, customers, etc. I'll ping you guys tomorrow to raise few more points and improvements we can do.

packages/medusa/src/types/routing.ts
sradevski: @srindom since this is just typings I think we can keep it as-is until we get everything in order internally, and then we can always switch to eg. `string` when it's needed. This is also why I didn't use an enum for now.

Let me add a TODO comment, so we can review it once our auth setup is cleaned up.

packages/medusa/src/utils/middlewares/authenticate-middleware.ts
sradevski: @srindom I'll bring this up tomorrow as well, I have a few more improvements I want to suggest, but I want to get everything working first.

PR #7391, opened 2024-05-22
feat: Event emitting part 1/N (Fulfillment)

adrien2p:
**What**
Add support for event emitting in the fulfillment module

**NOTE**
It does not include the review of the events for the abstract module factory if the method is not implemented in the module itself and rely on the default implementation

### DIFF
--- a/packages/core/utils/src/auth/abstract-auth-provider.ts
+++ b/packages/core/utils/src/auth/abstract-auth-provider.ts
@@ -5,19 +5,99 @@ import {
   IAuthProvider,
 } from "@medusajs/types"
 
+/**
+ * ### constructor
+ * 
+ * The constructor allows you to access resources from the module's container using the first parameter,
+ * and the module's options using the second parameter.
+ * 
+ * If you're creating a client or establishing a connection with a third-party service, do it in the constructor.
+ * 
+ * In the constructor, you must pass to the parent constructor two parameters:
+ * 
+ * 1. The first one is an empty object.
+ * 2. The second is an object having two properties:
+ *    - `provider`: The ID of the provider. For example, `emailpass`.
+ *    - `displayName`: The label or displayable name of the provider. For example, `Email and Password Authentication`.
+ * 
+ * #### Example
+ * 
+ * ```ts
+ * import { AbstractAuthModuleProvider } from "@medusajs/utils"
+ * import { Logger } from "@medusajs/types"
+ * 
+ * type InjectedDependencies = {
+ *   logger: Logger
+ * }
+ * 
+ * type Options = {
+ *   apiKey: string
+ * }
+ * 
+ * class MyAuthProviderService extends AbstractAuthModuleProvider {
+ *   protected logger_: Logger
+ *   protected options_: Options
+ *   // assuming you're initializing a client
+ *   protected client
+ * 
+ *   constructor (
+ *     { logger }: InjectedDependencies,
+ *     options: Options
+ *   ) {
+ *     super(
+ *       {},
+ *       {
+ *         provider: "my-auth",
+ *         displayName: "My Custom Authentication"
+ *       }
+ *     )
+ * 
+ *     this.logger_ = logger
+ *     this.options_ = options
+ * 
+ *     // assuming you're initializing a client
+ *     this.client = new Client(options)
+ *   }
+ * 
+ *   // ...
+ * }
+ * 
+ * export default MyAuthProviderService
+ * ```
+ */
 export abstract class AbstractAuthModuleProvider implements IAuthProvider {
+  /**
+   * @ignore
+   */
   private static PROVIDER: string
+  /**
+   * @ignore
+   */
   private static DISPLAY_NAME: string
-
+  /**
+   * @ignore
+   */
   protected readonly container_: any
+  /**
+   * @ignore
+   */
   public get provider() {
     return (this.constructor as typeof AbstractAuthModuleProvider).PROVIDER
   }
-
+  /**
+   * @ignore
+   */
   public get displayName() {
     return (this.constructor as typeof AbstractAuthModuleProvider).DISPLAY_NAME
   }
 
+  /**
+   * @ignore
+   * 
+   * @privateRemarks
+   * Documenting the constructor in the class's TSDocs as it's difficult to relay
+   * the necessary information with this constructor's signature.
+   */
   protected constructor({}, config: { provider: string; displayName: string }) {
     this.container_ = arguments[0]
     ;(this.constructor as typeof AbstractAuthModuleProvider).PROVIDER ??=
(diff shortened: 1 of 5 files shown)


adrien2p:
@olivermrbl could I get your eye on those first changes please? I d like your opinion on the approach before I do too many things and have to change everything :D

olivermrbl:
@adrien2p, this is great – let's pause this until we are all on board the direction. Will give this a thorough review tomorrow after the preview release 👍

adrien2p:
@olivermrbl sounds good to me let's discuss further tomorrow then 👍 just to stress that I ve followed an explicit path as we discussed while not mixing up everything in loop etc etc

adrien2p:
@olivermrbl do you think you ll have a bit of time to look at it so that we can conclude and I can continue this work?

packages/modules/fulfillment/src/utils/events.ts
```diff
@@ -0,0 +1,157 @@
...
+  if (!fulfillmentSets.length) {
+    return
+  }
+
+  const aggregator = sharedContext.messageAggregator!
+  const messages: EventBusTypes.RawMessageFormat[] = []
+
+  fulfillmentSets.forEach((fulfillmentSet) => {
+    messages.push({
+      service: Modules.FULFILLMENT,
```
olivermrbl: question: I cannot recall, what purpose does `service` serve?

adrien2p: It is actually used no where, it is more of an extra information on the provenance of the event. But we could start with making those metadata optional and re add them later if you prefer. But at least the idea is that you can provide some extra metadata

packages/modules/fulfillment/src/services/fulfillment-module-service.ts
```diff
@@ -1416,28 +1687,141 @@ export default class FulfillmentModuleService<
...
+     * @note
+     * Since the relation is a one to many, the deletion, update and creation of labels
+     * is handled b the orm. That means that we dont have to perform any manual deletions or update.
+     * For some reason we use to have upsert and replace handled manually but we could simplify all that just like
+     * we do below which will create the label, update some and delete the one that does not exists in the new data.
+     *
+     * There is a bit of logic as we need to reassign the data of those we want to keep
+     * and we also need to emit the events later on.
+     */
+    if (isDefined(data.labels) && isPresent(data.labels)) {
```
adrien2p, line 1710: cc @sradevski here the upsert with replace works smoothly and handled by the orm on oneToMany relations, which is what we wanted, the only thing I am doing here is that if the user only provides the id I reassign the existing data to it. and also few things to emit the events as I need to know what will happen. Do you have an idea why we havn't done that in the product?

packages/core/utils/src/event-bus/build-event-messages.ts
```diff
@@ -48,27 +13,29 @@ export function composeMessage(
...
     action?: string
     context?: Context
-    options?: Record<string, unknown>
+    options?: Record<string, any>
   }
 ): EventBusTypes.Message {
   const act = action || eventName.split(".").pop()
-  if (!action && !Object.values(CommonEvents).includes(act as CommonEvents)) {
+  if (
+    !action /* && !Object.values(CommonEvents).includes(act as CommonEvents)*/
```
carlos-r-l-rodrigues, line 31: uncomment?

packages/modules/fulfillment/src/services/fulfillment-module-service.ts
```diff
@@ -1416,31 +1687,145 @@ export default class FulfillmentModuleService<
...
+      const dataLabelIds: string[] = data.labels
+        .filter((label): label is { id: string } => "id" in label)
+        .map((label) => label.id)
+
+      deletedLabelIds = arrayDifference(existingLabelIds, dataLabelIds)
+
+      for (let label of data.labels) {
+        if ("id" in label) {
```
carlos-r-l-rodrigues:
```suggestion
        if (!("id" in label)) {
          continue
        }
```

carlos-r-l-rodrigues (approved):
LGTM

I would just add the module name to the event name as discussed.

packages/core/utils/src/event-bus/build-event-messages.ts
adrien2p: I did comment that just in the sense that this utils could be used by an external user for his own events and I was thinking they might have other actions that they would like to provide here. wdyt?

carlos-r-l-rodrigues: fair enough. so I guess we can remove that then

adrien2p: yeah I originally kept it to open a discussion 😂

packages/core/types/src/event-bus/common.ts
```diff
@@ -39,13 +41,12 @@ export type Message<T = unknown> = {
   options?: Record<string, unknown>
 }
 
-export type MessageFormat<T = unknown> = {
+export type RawMessageFormat<T = any> = {
```
sradevski, line 44: Why the rename, what is the non-raw message format?

packages/core/utils/src/inventory/events.ts
```diff
@@ -1,14 +1,13 @@
...
-  reservation_item_deleted: "reservation-item." + CommonEvents.DELETED,
-  inventory_level_deleted: "inventory-level." + CommonEvents.DELETED,
-  inventory_level_created: "inventory-level." + CommonEvents.CREATED,
-  inventory_level_updated: "inventory-level." + CommonEvents.UPDATED,
-}
+const eventBaseNames: ["inventoryItem", "reservationItem", "inventoryLevel"] = [
+  "inventoryItem",
+  "reservationItem",
+  "inventoryLevel",
+]
```
sradevski, line 8: Wouldn't `as const` do the same without having to explicitly list the eventnames?

packages/core/utils/src/event-bus/message-aggregator.ts
```diff
@@ -28,11 +29,25 @@ export class MessageAggregator implements IMessageAggregator {
...
-    options?: Record<string, unknown>
+      | EventBusTypes.RawMessageFormat<T>
+      | EventBusTypes.RawMessageFormat<T>[],
+    {
+      options,
+      sharedContext,
+    }: { options?: Record<string, unknown>; sharedContext?: Context } = {}
   ): void {
-    this.save(buildEventMessages(messageData, options))
+    const messages = Array.isArray(messageData) ? messageData : [messageData]
```
sradevski, line 39: Why did we move this out of the compose? I guess we'll have to write the same thing over and over again

packages/modules/fulfillment/src/services/fulfillment-module-service.ts
```diff
@@ -997,6 +1138,32 @@ export default class FulfillmentModuleService<
       sharedContext
     )
 
+    buildServiceZoneEvents({
```
sradevski: Is the decorator necessary, or should we just do `emitServiceZoneEvents`? What does the decorator do?

packages/modules/fulfillment/src/services/fulfillment-module-service.ts
sradevski: I don't get the question, do you mean why did we use upsertWithReplace and not write it out manually like this?

packages/modules/fulfillment/src/utils/events.ts
```diff
@@ -0,0 +1,515 @@
...
+  GeoZone,
+  ServiceZone,
+  ShippingOption,
+  ShippingOptionRule,
+  ShippingOptionType,
+} from "@models"
+import { Context, EventBusTypes } from "@medusajs/types"
+import { CommonEvents, FulfillmentUtils, Modules } from "@medusajs/utils"
+
+export function buildFulfillmentAddressEvents({
```
sradevski: This is a lot of code to build out events, I am fine with it for now, but we'll really need to improve how the events are constructed and sent.

carlos-r-l-rodrigues: can you clarify a bit this comment?
this is building the message and sending it to the aggregator first. What are the improvements you see to simplify this flow?

packages/core/types/src/event-bus/common.ts
adrien2p: well it is the `Message` once the raw transformed

packages/core/utils/src/inventory/events.ts
adrien2p: No it wouldn't and we would lose the type in the following function call

packages/core/utils/src/event-bus/message-aggregator.ts
adrien2p: I ve removed the old one in favor of the new one, but they was doing the exact same thing, if you look at the function at the end we call the same as well with the composed message

packages/modules/fulfillment/src/services/fulfillment-module-service.ts
adrien2p: The decorator inject the MessageAggregator and wrap the function such that the emit is called after the function execution. In those utils I never emit anything, otherwise we could emit things within a running transaction and then the transaction fail

packages/modules/fulfillment/src/utils/events.ts
adrien2p: yes could you provide an example of improvemens you think about?

packages/modules/fulfillment/src/services/fulfillment-module-service.ts
adrien2p: I am not writting it manually here, I need some info to prepare the event later i.e knowing what is created, updated, deleted but thats pretty much it. Here I don't have any custom entity manipulation, no playing with the entity map. I am kind of validating the input and gather information for event seggregation.

Eventually, the only thing I am doing is that if the user only provides an id, I assign the original values to it just for the orm to not complain, but thats pretty much it.

My quesstion was more related to what we did not go with the ORM handling it and wrote something completly custom, I dont recall

packages/core/types/src/event-bus/common.ts
sradevski: @adrien2p  Should we call it `RawMessage` without Format maybe? Just a suggestion, fine to keep it as is as well

packages/modules/fulfillment/src/utils/events.ts
sradevski: @adrien2p @carlos-r-l-rodrigues  I am thinking a generic `buildEvents` function, as a lot of the `build*` functions do more or less the same thing, with just different parameters.

adrien2p: It is already using a generic build event under the hood, here I am only passing things that are specific to my actions in this specific module. If I didnt created the module, I would have done that in the service itself just passing a few props to the raw message to build the event. Here I just moved it out of the way of the service thats it.

adrien2p: all the properties I am providing here are specific to eac action, I can't abstract them away

packages/modules/fulfillment/src/services/fulfillment-module-service.ts
sradevski: Ah ok got it. I think the issue was when also handling many-to-many which the ORM didn't handle well, and it was an all-or-nothing when using the entity map, so we went with not using it at all. But I think now that we are getting to a more feature-complete state, we will have time to revisit the code in the modules and see what makes most sense to do.

packages/modules/fulfillment/src/utils/events.ts
adrien2p: @sradevski @carlos-r-l-rodrigues what do you think like this? https://github.com/medusajs/medusa/pull/7391/commits/2fe0e6ffe6fb2a7f7fc68157bdf3dec077bcacf2 I think it is a good start then we can improve from that

packages/core/utils/src/modules-sdk/event-builder-factory.ts
```diff
@@ -0,0 +1,44 @@
...
+    if (!data.length) {
+      return
+    }
+
+    const aggregator = sharedContext.messageAggregator!
+    const messages: EventBusTypes.RawMessageFormat[] = []
+
+    data.forEach((dataItem) => {
+      messages.push({
+        service: Modules.FULFILLMENT,
```
sradevski: This will need to be passed as an argument

sradevski (approved):
Looks so much better 👍 Approving since it just needs a small change

packages/core/utils/src/modules-sdk/event-builder-factory.ts
adrien2p: done 😂

packages/modules/fulfillment/src/utils/events.ts
```diff
@@ -0,0 +1,293 @@
...
+} from "@models"
+import { Context } from "@medusajs/types"
+import {
+  CommonEvents,
+  eventBuilderFactory,
+  FulfillmentEvents,
+  Modules,
+} from "@medusajs/utils"
+
+export const eventBuilders = {
```
adrien2p, line 18: @carlos-r-l-rodrigues @sradevski in theory, we could even export that from the utils/fulfillment alongside the event enum wdyt?

adrien2p: for now I ll keep it here but I was just thinking out loud

adrien2p:
@thetutlage @olivermrbl if you want to have a look before I merge :D

packages/modules/fulfillment/src/utils/events.ts
sradevski: Yeah let's keep it for now, it's already well-defined so extracting it will be very easy. I do see the point of epxorting it TBH, but we can do it later

thetutlage:
Looks good to me 👍. As always have some questions, but they are separate from the PR :)

PR #7393, opened 2024-05-22
fix: Accept invite in admin

olivermrbl:


packages/admin-next/dashboard/src/hooks/api/invites.tsx
```diff
@@ -5,29 +5,34 @@ import {
...
 const invitesQueryKeys = queryKeysFactory(INVITES_QUERY_KEY)
 
 export const useInvite = (
   id: string,
   options?: Omit<
-    UseQueryOptions<InviteRes, Error, InviteRes, QueryKey>,
+    UseQueryOptions<
+      { invite: HttpTypes.AdminInviteResponse },
+      Error,
+      { invite: HttpTypes.AdminInviteResponse },
```
adrien2p, line 27: Q: why do we have the res type twice in many places? It might be a dumb question 😅

sradevski (approved):

adrien2p (approved):

PR #7395, opened 2024-05-22
Register plugin routes before the Medusa core routes

thetutlage:
Load plugin routes before the Medusa core routes, since it will allow the plugin to define routes with higher priority than Medusa. Here are couple of examples.

- Plugin registers a route called "/products/active"
- Medusa registers a route called "/products/:id"

Now, if Medusa routes gets registered first, then the `/products/active` route will never be resolved, because it will be handled by the `/products/:id` route.

### DIFF
--- a/packages/medusa/src/loaders/api.ts
+++ b/packages/medusa/src/loaders/api.ts
@@ -28,6 +28,28 @@ export default async ({ app, container, plugins }: Options) => {
     ContainerRegistrationKeys.CONFIG_MODULE
   )
 
+  /**
+   * Always load plugin routes before the Medusa core routes, since it
+   * will allow the plugin to define routes with higher priority
+   * than Medusa. Here are couple of examples.
+   *
+   * - Plugin registers a route called "/products/active"
+   * - Medusa registers a route called "/products/:id"
+   *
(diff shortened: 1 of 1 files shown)


sradevski (approved):
We need to revisit our entire routing TBH as it is a bit of a mess, but this works for now

packages/medusa/src/loaders/api.ts
```diff
@@ -28,6 +28,28 @@ export default async ({ app, container, plugins }: Options) => {
...
+   */
+  await Promise.all(
+    plugins.map(async (pluginDetails) => {
+      return new RoutesLoader({
+        app: app,
+        configModule,
+        rootDir: path.join(pluginDetails.resolve, "api"),
+      }).load()
+    })
+  )
```
riqwan, line 51: sugg: to wrap this baby up with a user friendly error

riqwan (approved):
LGTM!

packages/medusa/src/loaders/api.ts
thetutlage: Agree 100%.

But usually we have these sorts of loaders in many places and most of them are not wrapped in `try/catch` and then showing a user friendly error.

Maybe, we have to find all of them and settle on a pattern to use.

PR #7396, opened 2024-05-22
feat: Apply authentication middleware globally

sradevski:


packages/medusa/src/api/admin/users/route.ts
```diff
@@ -90,3 +90,5 @@ export const POST = async (
 
   res.status(200).json({ user, token })
 }
+
+export const AUTHENTICATE = false
```
adrien2p, line 88: Question: Does that mean that listing all users does not require to be auth?

packages/medusa/src/loaders/helpers/routing/__fixtures__/server/index.ts
```diff
@@ -135,10 +136,11 @@ export const createServer = async (rootDir) => {
       if (opts.clientSession) {
         const token = jwt.sign(
           {
-            customer_id:
+            actor_id:
```
adrien2p, line 142: nice catch 🙏

packages/medusa/src/loaders/helpers/routing/__tests__/index.spec.ts
```diff
@@ -156,7 +156,7 @@ describe("RoutesLoader", function () {
     })
 
     it("should return 200 when customer is authenticated", async () => {
-      const res = await request("GET", "/store/me/protected", {
+      const res = await request("GET", "/store/customers/me/protected", {
```
adrien2p: Question: are we sure we don't want to stick to store/me? wouldn't having the customer segment leak some info?

packages/medusa/src/loaders/helpers/routing/index.ts
```diff
@@ -332,21 +331,19 @@ export class RoutesLoader {
           }
 
           if (route.startsWith("/store")) {
-            config.shouldAppendCustomer = true
-
             if (shouldAddCors) {
               config.shouldAppendStoreCors = true
             }
+
+            if (shouldRequireAuth && route.startsWith("/store/customers/me")) {
```
adrien2p: suggestion: /store/customers?

packages/medusa/src/api/admin/users/route.ts
sradevski: @adrien2p ah nice catch 👍 This is why I don't like this approach too much 😄

packages/medusa/src/loaders/helpers/routing/index.ts
sradevski: @adrien2p I am sticking to what was already there. Also, the `/store/customers` endpoints apply a different type of authentication that allows for unregistered access

packages/medusa/src/api/admin/users/route.ts
sradevski: @adrien2p ah ok, I already left the locally defined authentication, so it was good, nothing to do

packages/medusa/src/loaders/helpers/routing/index.ts
adrien2p: would that mean that if you have GET store/customeres you would be able to list the customers?

sradevski: @adrien2p there is no `GET store/customers` handler defined, so you won't be able to do it. You can only create a new one.

packages/medusa/src/loaders/helpers/routing/__tests__/index.spec.ts
sradevski: @adrien2p no idea TBH, I think having `POST store/customers` for registration is a bit cleaner, but if you think we should change it maybe raise it in Slack?

packages/medusa/src/loaders/helpers/routing/index.ts
adrien2p: I am fine with keeping it, was just wondering ahah

packages/medusa/src/loaders/helpers/routing/__tests__/index.spec.ts
```diff
@@ -155,34 +155,6 @@ describe("RoutesLoader", function () {
       expect(res.text).toBe("GET /admin/unprotected")
     })
 
-    it("should return 200 when customer is authenticated", async () => {
```
sradevski, line 157: @olivermrbl @adrien2p after looking at the code, I think it makes sense to not apply any authentication middleware for the storefront.

Currently we have 3 routes that require authentication
- `POST /store/customers` with allowUnregistered
- `ALL /store/customers/me*`
- `GET /store/orders`

It is very weird if we apply the middleware globally for some routes, but not others, so I think it's easier to reason and understand if we say:
**We apply auth middleware globally for admin**
**We don't apply auth middleware globally for store**

The auth context will be filled in on all routes btw, this is just for restricting access.
WDYT?

adrien2p: Yes I think it makes sense and looks closer to what we had no? we still have restricted the few store end points that requires it but then the user can restrict what he needs if necessary

shahednasser: No strong opinions on this, but I do think restricting `/store/customers/me` is a good way of maintaining consistency between the developers' customizations and our API routes. It's also an approach that makes sense (so not difficult for developers to understand/expect). I do agree it doesn't need to be three different route paths/patterns though.

olivermrbl: Yes, this sounds right to me. It doesn't make sense to have the store-facing API authenticated by default.

sradevski: Maybe I wasn't clear, but what I meant is, it is better to not apply global auth middleware for `store` at all (before we did have it applied like that). The effect is the same, we would still require auth for `store/customers/me`, but only because it's applied in `customers/middleware.ts`.

shahednasser: Ah ok I misunderstood then 😄 sounds good to me 👍🏻

adrien2p (approved):

PR #7401, opened 2024-05-22
fix(admin-sdk): error when admin needs build

shahednasser:
Fix the error shown when admin is not built instructing the user to run the `build` command.

### DIFF
--- a/packages/admin-next/admin-sdk/src/lib/serve.ts
+++ b/packages/admin-next/admin-sdk/src/lib/serve.ts
@@ -24,7 +24,7 @@ export async function serve(options: ServeOptions) {
 
   if (!indexExists) {
     throw new Error(
-      `Could not find the admin UI build files. Please run \`npm run build\` command and try again.`
+      `Could not find the admin UI build files. Please run \`npm run build\` or \`yarn build\` command and try again.`
     )
   }
 


packages/admin-next/admin-sdk/src/lib/serve.ts
```diff
@@ -24,7 +24,7 @@ export async function serve(options: ServeOptions) {
 
   if (!indexExists) {
     throw new Error(
-      `Could not find the admin UI build files. Please run "medusa-admin build" or enable "autoRebuild" in the plugin options to build the admin UI.`
+      `Could not find the admin UI build files. Please run \`npm run build\` command and try again.`
```
riqwan: just to be pushing to both :D 

```suggestion
      `Could not find the admin UI build files. Please run \`npm run build\` or \`yarn build\` command and try again.`
```

riqwan (approved):

packages/admin-next/admin-sdk/src/lib/serve.ts
adrien2p: ahah should we add pnpm run build too? :D

shahednasser: I think having npm and yarn should be enough 😄

riqwan: If you're running pnpm, you're on your own unfortunately. ;)

adrien2p: 😂 fair enough

PR #7402, opened 2024-05-22
chore: Change default log level

olivermrbl:
The `LOG_LEVEL` has defaulted to `silly` which will print all types of messages, debug, info, warn, error, etc. 

My suggestion is to only print at the `info` level and above and let the rest be up to the user. 

Separately, changed the log level to `debug` for API Routes.


olivermrbl:
If we want all messages in `develop` by default, perhaps we should set the log level in the `develop` command? I would still default to only logging at the `info` level.

riqwan (approved):

adrien2p (approved):
💪

sradevski (approved):

PR #7404, opened 2024-05-22
fix: Commands loading entrypoints

thetutlage:
This PR does a couple of changes

- It uses auto-detection to know if we are using `ts-node` to run the development server
- Next, it runs the `user` command in worker mode.

### DIFF
--- a/packages/medusa/src/commands/user.js
+++ b/packages/medusa/src/commands/user.js
@@ -22,7 +22,7 @@ export default async function ({
   try {
     /**
      * Enabling worker mode to prevent discovering/loading
-     * starter kit resources
+     * of API routes from the starter kit
      */
     process.env.MEDUSA_WORKER_MODE = "worker"
 


adrien2p (approved):

packages/medusa/src/commands/user.js
```diff
@@ -20,6 +20,12 @@ export default async function ({
   track("CLI_USER", { with_id: !!id })
   const app = express()
   try {
+    /**
+     * Enabling worker mode to prevent discovering/loading
+     * starter kit resources
```
shahednasser: Maybe we should be clear here that it specifically doesn't register API routes, but not all resources

packages/medusa/src/loaders/config.ts
```diff
@@ -102,8 +102,8 @@ export default (rootDirectory: string): ConfigModule => {
     modules: configModule.modules ?? {},
     featureFlags: configModule?.featureFlags ?? {},
     plugins: configModule?.plugins ?? [],
-    directories: configModule?.directories ?? {
-      srcDir: "dist",
+    directories: {
+      srcDir: process[Symbol.for("ts-node.register.instance")] ? "src" : "dist",
```
riqwan: is this really needed if its not used in a lot of places and not configurable?

adrien2p: We are curretly discussing it in slack :D

packages/medusa/src/commands/user.js
thetutlage: Okay, is that the case. I thought it disables all the resources. Let me update the comment.

packages/medusa/src/loaders/config.ts
thetutlage: Moved it to the place where it belongs. No longer in the config

packages/medusa/src/commands/user.js
shahednasser: No just the API routes, it still loads modules and their resources

PR #7406, opened 2024-05-22
docs: fix data payload of events in subscribers

shahednasser:
Fix data payload in subscribers to consider data either in `data.data` or directly in `data`.


shahednasser:

PR #7407, opened 2024-05-22
feat(order): shipment workflow

carlos-r-l-rodrigues:


packages/core/core-flows/src/order/steps/register-shipment.ts
```diff
@@ -0,0 +1,29 @@
...
+  async (orderId, { container }) => {
+    if (!orderId) {
+      return
+    }
+
+    const service = container.resolve<IOrderModuleService>(
+      ModuleRegistrationName.ORDER
+    )
+
+    await service.revertLastVersion(orderId)
```
adrien2p, line 27: question: if another workflow is also updating the order, can it revert to the wrong last version? maybe it will be part of the locking mechanism right?

packages/core/core-flows/src/order/workflows/create-shipment.ts
```diff
@@ -0,0 +1,140 @@
...
+
+  if (diff.length) {
+    throw new MedusaError(
+      MedusaError.Types.INVALID_DATA,
+      `Items with ids ${diff.join(", ")} does not exist in order with id ${
+        order.id
+      }.`
+    )
+  }
+}
```
adrien2p: suggestion: what do you think if we move some validations in the order utils dir next to the workflow? I believe many workflows will needs those validations right?

packages/core/types/src/workflow/order/create-shipment.ts
```diff
@@ -0,0 +1,17 @@
+import { BigNumberInput } from "../../totals"
```
adrien2p, line 1: suggestion: Maybe we can start moving the workflow types that are located in src/order, no necessarely in that pr though

packages/medusa/src/api/admin/orders/[id]/fulfillments/[fulfillment_id]/shipment/route.ts
```diff
@@ -0,0 +1,40 @@
...
+  AuthenticatedMedusaRequest,
+  MedusaResponse,
+} from "../../../../../../../types/routing"
+import { AdminOrderCreateShipmentType } from "../../../../validators"
+
+export const POST = async (
+  req: AuthenticatedMedusaRequest,
+  res: MedusaResponse
+) => {
+  const remoteQuery = req.scope.resolve("remoteQuery")
```
Comment (adrien2p) [inline packages/medusa/src/api/admin/orders/[id]/fulfillments/[fulfillment_id]/shipment/route.ts]:
```suggestion
  const remoteQuery = req.scope.resolve(ContainerRegistrationKeys.REMOTE_QUERY")
```

packages/medusa/src/api/admin/orders/[id]/fulfillments/[fulfillment_id]/shipment/route.ts
```diff
@@ -0,0 +1,40 @@
...
+  }
+
+  const { errors } = await createOrderShipmentWorkflow(req.scope).run({
+    input,
+    throwOnError: false,
+  })
+
+  if (Array.isArray(errors) && errors[0]) {
+    throw errors[0].error
+  }
```
Comment (adrien2p) [inline packages/medusa/src/api/admin/orders/[id]/fulfillments/[fulfillment_id]/shipment/route.ts:34]: wdyt if by default the workflows always re throw the first error? And if you want to handle it manually then you can set the flag and check how many errors they are and how to handle them?

packages/medusa/src/api/admin/orders/[id]/fulfillments/[fulfillment_id]/shipment/route.ts
```diff
@@ -0,0 +1,40 @@
+import { createOrderShipmentWorkflow } from "@medusajs/core-flows"
+import { remoteQueryObjectFromString } from "@medusajs/utils"
+import {
+  AuthenticatedMedusaRequest,
+  MedusaResponse,
+} from "../../../../../../../types/routing"
+import { AdminOrderCreateShipmentType } from "../../../../validators"
+
+export const POST = async (
+  req: AuthenticatedMedusaRequest,
```
Comment (adrien2p) [inline packages/medusa/src/api/admin/orders/[id]/fulfillments/[fulfillment_id]/shipment/route.ts]: suggestion: we could add the type in the template arg here

packages/medusa/src/api/admin/orders/[id]/fulfillments/route.ts
```diff
@@ -13,7 +14,10 @@ export const POST = async (
 
   const variables = { id: req.params.id }
 
-  const input = req.validatedBody
+  const input = {
+    ...((req.validatedBody as AdminOrderCreateFulfillmentType) ?? {}),
```
Comment (adrien2p) [inline packages/medusa/src/api/admin/orders/[id]/fulfillments/route.ts]: suggestion: if you add the type in the template arg, this cast wont be necessary

packages/medusa/src/api/admin/orders/validators.ts
```diff
@@ -62,3 +62,20 @@ export const AdminOrderCreateFulfillment = z.object({
...
+
+const Label = z.object({
+  tracking_number: z.string(),
+  tracking_url: z.string(),
+  label_url: z.string(),
+})
+
+export const AdminOrderCreateShipment = z.object({
+  items: z.array(Item),
+  labels: z.array(Label),
```
adrien2p: Q: do you think it is required?

adrien2p (approved):
LGTM, few suggestions 💪

packages/core/core-flows/src/order/steps/register-shipment.ts
carlos-r-l-rodrigues: Yeah, if there is concurrency we can revert the wrong change.
However this flow shouldn't be very concerning in regards concurrency. 
But I would love to have the locking as a Module so we can apply it to whatever flow we want to control this type of situation.

carlos-r-l-rodrigues: But I will very likely change this to return a specific Order Change ID and try to delele only that. At least if something happen, the worst case scenario will fail to revert the change.

packages/core/core-flows/src/order/workflows/create-shipment.ts
carlos-r-l-rodrigues: good call.

packages/medusa/src/api/admin/orders/validators.ts
carlos-r-l-rodrigues: no, it isn't. good catch.

PR #7410, opened 2024-05-22
fix: Remove log for read only links

olivermrbl:


adrien2p (approved):
I thought Carlos added that check ahah

carlos-r-l-rodrigues (approved):

PR #7416, opened 2024-05-23
fix(admin-next, utils, medusa): Handle db errors properly and fix form

adrien2p:
Screenshot 2024-05-23 at 12 17 19


packages/admin-next/dashboard/src/routes/invite/invite.tsx
```diff
@@ -220,7 +220,7 @@ const CreateView = ({
       {
         onSuccess,
         onError: (error) => {
-          if (isAxiosError(error) && error.response?.status === 400) {
+          if ("type" in error && error.type === "invalid_data") {
```
adrien2p: @kasperkristensen is there another way to do that? here we don't have an axios error but still we have an invalid data error

adrien2p: Or should I use a toast like other places?

packages/admin-next/dashboard/src/v2-routes/invite/invite.tsx
```diff
@@ -232,8 +232,8 @@ const CreateView = ({
         description: t("invite.toast.accepted"),
         dismissLabel: t("actions.close"),
       })
-    } catch (error) {
-      if (isAxiosError(error) && error.response?.status === 400) {
+    } catch (error: any) {
+      if ("type" in error && error.type === "invalid_data") {
```
adrien2p: @olivermrbl I might be wrong but other places are showing a toast with the error message, should I do that?

packages/admin-next/dashboard/src/routes/invite/invite.tsx
thetutlage: Not sure if its worth handling. But if error turns out to be a `string`, then the `in` check will fail

adrien2p: I don't think it will be a string, but we can add an handler, my question is more if there is a normalized way to handle those errors already, can seem to find it before I jump into creating one :D

kasperkristensen: We want to display the error in the form and not a toast here, that is why we were previously checking if the error was 400, so we could show "Invalid invite" message, and if not then we would show "Server error - try again". I think this fine, but as i mentioned in the PR for the new SDK I think it would be great to have a utility like isAxiosError for handling these things.

packages/admin-next/dashboard/src/v2-routes/invite/invite.tsx
olivermrbl: I don't think we should if the form is showing the error. Is it doing that?

adrien2p: No it does not because there is no response on the error, there is the type, message and no code as well. So currently it does not show anything

adrien2p: here the error I was playing with is duplicated email

packages/admin-next/dashboard/src/routes/invite/invite.tsx
sradevski: @kasperkristensen @adrien2p you can do `err instanceof FetchError` already, but we can wrap that in a util if you think it will be easier to use it.

adrien2p: I ll clean up

adrien2p:
@sradevski @kasperkristensen I ve been able to finally cleaned it up

sradevski (approved):

PR #7417, opened 2024-05-23
feat(create-medusa-app): remove admin email prompt

shahednasser:
Remove prompt asking for email as the user will enter it in the invite form in the admin


adrien2p (approved):

shahednasser:
It doesn't affect the pipeline since it doesn't use `create-medusa-app`

PR #7418, opened 2024-05-23
Use medusajs/medusa package to register the CLI entrypoint

thetutlage:
I see that multiple packages registers itself to the bin as `medusa` binary. I went ahead and made `@medusajs/medusa` to do it only. 

Let me know if my approach is wrong :)

### DIFF
--- a/packages/admin-next/dashboard/src/lib/client/client.ts
+++ b/packages/admin-next/dashboard/src/lib/client/client.ts
@@ -7,15 +7,15 @@ import { currencies } from "./currencies"
 import { customerGroups } from "./customer-groups"
 import { customers } from "./customers"
 import { fulfillmentProviders } from "./fulfillment-providers"
+import { fulfillments } from "./fulfillments"
 import { inventoryItems } from "./inventory"
 import { invites } from "./invites"
+import { orders } from "./orders"
 import { payments } from "./payments"
 import { priceLists } from "./price-lists"
 import { productTypes } from "./product-types"
 import { products } from "./products"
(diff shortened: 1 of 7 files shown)


packages/cli/medusa-cli/package.json
```diff
@@ -3,9 +3,6 @@
   "version": "1.3.22",
   "description": "Command Line interface for Medusa Commerce",
   "main": "dist/index.js",
-  "bin": {
-    "medusa": "cli.js"
-  },
```
adrien2p: are you sure this can be removed?

packages/medusa/package.json
```diff
@@ -3,7 +3,9 @@
   "version": "1.20.5",
   "description": "Building blocks for digital commerce",
   "main": "dist/index.js",
-  "bin": "./cli.js",
```
thetutlage, line 6: Then should we remove this @adrien2p ?

packages/cli/medusa-cli/package.json
thetutlage: Added back

packages/medusa/package.json
thetutlage: Removed

sradevski (approved):

thetutlage:
@adrien2p Want to take a final look?

olivermrbl:
@thetutlage should we test this with a snapshot? It seems the pipeline is failing 🤔

adrien2p (approved):

PR #7419, opened 2024-05-23
Revamp the authentication setup

sradevski:
This ended up being a pretty big PR, but it should be pretty easy to review if you go commit-by-commit.
The changes are as follows:

1. Create an emailpass auth provider package
2. Refactor the auth module package to be more consistent with others, and remove the concept of `scope`
3. Switch to using actor types instead of scopes for authentication
4. Add support for per-actor auth methods (what `scope` did before)

The only remaining piece is to move the Google auth provider to a package, but that is for another PR


integration-tests/api/medusa-config.js
```diff
@@ -42,16 +42,14 @@ module.exports = {
...
           {
-            name: "emailpass",
-            scopes: {
-              admin: {},
-              store: {},
+            resolve: "@medusajs/auth-emailpass",
+            options: {
+              config: {
+                emailpass: {},
+              },
```
olivermrbl: question: Why did we choose to add a separate package?


comment: I think we need to do something about these configs some time. Not for this PR.

packages/modules/auth/src/models/auth-identity.ts
```diff
@@ -29,15 +29,9 @@ export default class AuthIdentity {
...
   provider: string
 
-  @Property({ columnType: "text" })
-  scope: string
-
   @Property({ columnType: "jsonb", nullable: true })
   user_metadata: Record<string, unknown> | null
 
-  @Property({ columnType: "jsonb" })
-  app_metadata: Record<string, unknown> = {}
```
olivermrbl, line 39: question: I thought we wanted to keep `app_metadata` for other auth-specific props?

adrien2p: seb was in favour to remove it as we have the provider data

integration-tests/api/medusa-config.js
adrien2p: yes definately

packages/modules/auth/src/providers/google.ts
```diff
@@ -1,13 +1,12 @@
...
 import { AbstractAuthModuleProvider, MedusaError } from "@medusajs/utils"
-import { AuthIdentityService } from "@services"
 import jwt, { JwtPayload } from "jsonwebtoken"
 
 import { AuthorizationCode } from "simple-oauth2"
 import url from "url"
 
 type InjectedDependencies = {
-  authIdentityService: AuthIdentityService
+  authIdentityService: any
```
sradevski, line 9: Once this is moved to a package this will be injected as a parameter instead, so just typed it as any for now.

integration-tests/api/medusa-config.js
sradevski: @olivermrbl I think having it as a separate package makes it easier if you want to swap it out, plus it follows the same architecture as every other module. What's your thought process, why shouldn't we do it as a package?

packages/modules/auth/src/models/auth-identity.ts
olivermrbl: Okay I might have missed that discussion, so I don't think I have the context to review the PR. Would just be a bunch of questions 😅

sradevski: @olivermrbl that's fine, I can answer all the questions you have, it's important we get this right :)

integration-tests/api/medusa-config.js
olivermrbl: I just asked out of curiosity because I wanted to hear your thinking. 

I don't have a strong opinion on whether to have a separate package or not, but the general idea with the local providers (in payments, fulfillment, and other domains, too) has been to reduce friction for the common provider use cases and make it easy to get started. 

We can eliminate this potential friction by installing the package and enabling it in the auth module by default, but it is still more error-prone than just baking it in. If you don't have this configured correctly, the admin is effectively useless, as you cannot sign in. 

That said, let's go with this for now. We can revisit if this becomes a huge pain for devs getting started. I think striving to have a consistent provider architecture makes a lot of sense and weighs heavier for now. And maybe auth specifically, is a slightly different use case than e.g. payments.

olivermrbl: > I think striving to have a consistent provider architecture makes a lot of sense and weighs heavier for now.

Should we migrate google to a package too then?

sradevski: @olivermrbl yup as you can see in the PR description I wrote that as a remaining TODO 😄 

I see where you're coming from, and I do agree that this can cause friction. But I think us having a consistent setup and architecture will allow us to improve the UX of all modules, not just auth, and if we are worried the UX is bad or the possibility for misconfiguring is high, we should focus on fixing that instead. At least that's how I see it 😄

olivermrbl: > @olivermrbl yup as you can see in the PR description I wrote that as a remaining TODO 😄

Missed that, my bad 😅

srindom:
A few questions and thoughts:

1. Doesn't look like this addresses the provider-specific access, is that correct?
    - `AuthIdentity.entity_id` is meant to be an id from the external provider (e.g., your google user id). This means that these current data model doesn't support multiple providers for the same AuthIdentity. This means two actor configurations must share the same providers to be certain that an AuthIdentity can act as different actors. 
    - Think an AuthIdentityProviderId entity with `auth_id, provider_id, external_id` would solve this?
2. We need to address how customer/admin account creation works. Currently once you are authenticated in the correct scope you are able to call `POST /store/customers` to create a customer account. With the changes here the auth middleware only accepts requests that already have an AuthIdentity<>Customer connection so you will never be able to create an account. Correct me if I am misunderstanding something.
3. Let's wait with moving providers into separate packages - think there is great value in baking them in directly and being able to ship new providers easily for people. We can always look into moving them out or enabling external providers as well, but think we should start with a more locked down approach and expand from there.

sradevski:
@srindom 
1. The issue with `entity_id` needs to be tackled separately, as that constraint already existed (even before you couldn't have 2 providers for the same identity). This PR does take care of provider-specific access, though, that is enforced in the `auth/[actor_type]/[provider]` route. Maybe we can huddle and I can explain why it works.
2. When you create an AuthUser it is not associated with anything yet. `store/customers` and `admin/invite` for example don't require that an `actor_id` is present, so after the authUser was created you can call either of them and it is only once they are created that the actor type is enforced. I am doing this testing now, and it is possible I have a bug, but this should work.
3. I think we can have them as separate packages, and still have them baked-in just by directly installing the package in the module. This still provides the benefit of isolation (which was a bit broken before, since the entire AuthEntityService was injected) and isolated testing, while not requiring the user to do anything, WDYT?

srindom:
> even before you couldn't have 2 providers for the same identity
My comment was not as much related to a requirement of two providers for the same identity (although this is a nice benefit of these changes 👍)  but rather that you can tie access based on the provider. This was possible before because of the scopes - i.e., you were able to say: "only these providers can get access through the admin scope". Since we are removing the scope we now need to enable multiple providers for the same identity.

2. Hadn't seen the `allowUnregistered` option, but looks like that is the key to making that part work. 
3. Sounds good! As long as `npm install @medusajs/auth` gets you the providers I don't mind 👍

sradevski:
@srindom now https://github.com/medusajs/medusa/pull/7419/files#diff-687ef806f7d3228e1b747d07558bc53c3cf9c1cac01b332bd8267e9b81aa4370R22 will enforce which providers can be used for which actors. So in your config you can say:
```
{
  'user': ["google"]
  'customer': ["emailass"]
}
```

So you achieve the same thing, except now the auth module is agnostic of the concept of scopes (which is an HTTP layer concern) and is enforced in the HTTP layer

olivermrbl (approved):

olivermrbl:
@sradevski, looks like we might need to update the `user` command to reflect these changes

sradevski:
@olivermrbl yeah there is just a minor issue with app_metadata not being nullable, pushing a fix now (just testing it)

olivermrbl:
It passed! 😄

Should we merge?

sradevski:
@olivermrbl finally 😄 For some weird reason the table already existed, so the auth migrations were running as expected. I will check how exactly the CLI test runs

PR #7422, opened 2024-05-23
fix(workflows-sdk): use loaded modules when container is empty

carlos-r-l-rodrigues:
what:
  - use loaded modules if there is no container
  - remove remote query callback resolving legacy v1 services.
  - cleanup isInternalService config


packages/core/types/src/joiner/index.ts
```diff
@@ -3,11 +3,6 @@ export type JoinerRelationship = {
   foreignKey: string
   primaryKey: string
   serviceName: string
-  /**
-   * If true, the relationship is an internal service from the medusa core
-   * TODO: Remove when there are no more "internal" services
-   */
-  isInternalService?: boolean
```
adrien2p, line 10: nice cleanup :D

packages/modules/link-modules/src/utils/generate-schema.ts
```diff
@@ -32,11 +32,7 @@ export function generateGraphQLSchema(
 
   for (const extend of joinerConfig.extends ?? []) {
     const extendedModule = MedusaModule.getModuleInstance(extend.serviceName)
-    if (
-      !isReadOnlyLink &&
-      !extendedModule &&
-      !extend.relationship.isInternalService
-    ) {
+    if (!extendedModule) {
```
adrien2p, line 35: why do we changed that?

carlos-r-l-rodrigues: This is more a safe guard for us. In practice this condition will never happen because the read only links were already filtered against loaded modules.

adrien2p (approved):

PR #7424, opened 2024-05-23
Silently load ts-node

thetutlage:
Okay, so this change will allow us to run all the Medusa command inside a TypeScript project without creating a `dist` folder.

It means when we create a new project, we can skip the `npm run build` part altogether and run `migrations run` or `user` commands directly. 

But yeah happy to hear what others think.

### DIFF
--- a/packages/cli/medusa-cli/cli.js
+++ b/packages/cli/medusa-cli/cli.js
@@ -1,3 +1,6 @@
 #!/usr/bin/env node
+try {
+  require('ts-node').register({})
+} catch {}
 require("dotenv").config()
 require("./dist/index.js")


thetutlage:
@adrien2p 

Yup tested it locally with all the commands and seems to be working great so far!

riqwan (approved):
if this works, this is mad!

adrien2p (approved):

olivermrbl (approved):
Very nice!

thetutlage:
@riqwan 

> if this works, this is mad!

Yeah, sometimes simplest of the changes brings great benefit.

PR #7428, opened 2024-05-23
feat(types,medusa,link-modules): add cart links to promotions/product

riqwan:
what:

- Adds links between cart + products and cart + promotions.
- Introduces attributes necessary for promotions in cart response


riqwan:
@adrien2p yes, for now I just wanted to get the bare minimum for the cart. Only tiny PRs from now onward :D

PR #7430, opened 2024-05-23
feat: Load custom link definitions

adrien2p:
**What**
Load custom links from the projects


carlos-r-l-rodrigues (approved):

packages/medusa/src/loaders/helpers/resolve-plugins-links.ts
```diff
@@ -0,0 +1,46 @@
...
+/**
+ * import files from the links directory to retrieve the links to be loaded
+ * @param plugins
+ * @param container
+ */
+export async function resolvePluginsLinks(
+  plugins: PluginDetails[],
+  container: MedusaContainer
+): Promise<ModuleJoinerConfig[]> {
+  const logger = container.resolve("logger") ?? console
```
sradevski: ContainerRegistrationKeys?

packages/medusa/src/loaders/index.ts
```diff
@@ -138,6 +139,7 @@ export default async ({
     onApplicationPrepareShutdown: medusaAppOnApplicationPrepareShutdown,
   } = await loadMedusaApp({
     container,
+    customLinkModules: await resolvePluginsLinks(plugins, container),
```
sradevski: Should this happen inside `loadMedusaApp` instead? 

Also, I don't think we should make a difference between custom and non-custom link definitions, we should load all of them the same way IMO

sradevski (commented):
Nice :) Let's add some tests to ensure they are loaded as expected

packages/medusa/src/loaders/index.ts
adrien2p: they are all loaded the same way, it is just naming, I can rename it

packages/medusa/src/loaders/helpers/resolve-plugins-links.ts
adrien2p: nice catch :D

packages/medusa/src/loaders/index.ts
carlos-r-l-rodrigues: I think it is fine to keep this name. MedusaApp is trying to import "@medusajs/link-modules", those would be the ones the system provides

adrien2p: I choose not to do it in medusa app because then it becomes suddenly aware of the plugins, I would like to keep this concept outside IMO

adrien2p: I ve changed to link modules, but since we also have custom modules, it might make more sense to revert to customLinkModules @sradevski

adrien2p:
@sradevski we already have the link-modules tested which is the one ensuring everything is loaded, the only think I added is getting the objects from the file through glob which is also tested not sure we need t add more no?

sradevski (approved):
I'm ok if you think there is no need for an additional test, it's just nice to cover the plugin loading aspect as well but I'm fine without it.

On the naming and where the resolution happens, I have no strong preference it was just what I thought might be a bit cleaner, I leave it up to you 👍

packages/medusa/src/loaders/__tests__/resolve-plugins-links.spec.ts
```diff
@@ -0,0 +1,62 @@
+import { mkdirSync, rmSync, writeFileSync } from "fs"
```
adrien2p, line 1: @sradevski to ensure we are loading correctly from links wdyt?

adrien2p:
@olivermrbl any feedback on this one?

olivermrbl (approved):

PR #7431, opened 2024-05-23
chore(modules-sdk): throw when failing to load module

carlos-r-l-rodrigues:


olivermrbl (approved):
Nice, I think this makes sense!

PR #7436, opened 2024-05-24
Update develop command to use JIT compiler

thetutlage:
The `develop` command can now be used to run the Medusa app in development. It uses the JIT compiler and a file watcher to run the application with compilation.

Also, at the same time we have removed the concept of `COMMAND_INITIATED_BY` and instead use `NODE_ENV` to decide if we should run the Admin in development or production mode. The same has been confirmed with @kasperkristensen

### DIFF
--- a/packages/medusa/src/commands/develop.js
+++ b/packages/medusa/src/commands/develop.js
@@ -57,8 +57,8 @@ export default async function ({ port, directory }) {
         execArgv: argv,
       })
       this.childProcess.on("error", (error) => {
-        Logger.error("Dev server died with error", error)
-        Logger.info("Still watching filesystem to restart the dev server")
+        Logger.error("Dev server failed to start", error)
+        Logger.info("The server will restart automatically after your changes")
       })
     },
 
@@ -94,28 +94,19 @@ export default async function ({ port, directory }) {
 
       this.watcher.on("add", (file) => {
         Logger.info(
-          `${path.relative(
-            directory,
-            file
-          )} file created: restarting dev server...`
+          `${path.relative(directory, file)} created: Restarting dev server`
         )
         this.restart()
       })
       this.watcher.on("change", (file) => {
         Logger.info(
-          `${path.relative(
-            directory,
-            file
-          )} file modified: restarting dev server...`
+          `${path.relative(directory, file)} modified: Restarting dev server`
         )
         this.restart()
       })
       this.watcher.on("unlink", (file) => {
         Logger.info(
-          `${path.relative(
-            directory,
-            file
-          )} file removed: restarting dev server...`
+          `${path.relative(directory, file)} removed: Restarting dev server`
         )
         this.restart()
       })


packages/medusa/src/commands/develop.js
```diff
@@ -23,6 +23,109 @@ export default async function ({ port, directory }) {
...
+      this.childProcess = fork(cliPath, ["start", ...args], {
+        cwd: directory,
+        env: {
+          ...process.env,
+          NODE_ENV: "development",
+        },
+        execArgv: argv,
+      })
+      this.childProcess.on("error", (error) => {
+        Logger.error("Dev server died with error", error)
```
adrien2p: Suggestion: Dev server cannot start, error

packages/medusa/src/commands/develop.js
```diff
@@ -23,6 +23,109 @@ export default async function ({ port, directory }) {
...
+        Logger.info("Still watching filesystem to restart the dev server")
```
adrien2p: Suggestion: The server will restart automatically after your changes

packages/medusa/src/commands/develop.js
```diff
@@ -23,6 +23,109 @@ export default async function ({ port, directory }) {
...
+        cwd: process.cwd(),
+        ignored: [/(^|[\\/\\])\../, "node_modules", "dist", "src/admin/**/*"],
+      })
+
+      this.watcher.on("add", (file) => {
+        Logger.info(
+          `${path.relative(
+            directory,
+            file
+          )} file created: restarting dev server...`
```
adrien2p: Suggestion: we can maybe remove the file word? So that it looks more like 
service.ts created

packages/medusa/src/commands/develop.js
```diff
@@ -23,6 +23,109 @@ export default async function ({ port, directory }) {
...
+        )
+        this.restart()
+      })
+      this.watcher.on("change", (file) => {
+        Logger.info(
+          `${path.relative(
+            directory,
+            file
+          )} file modified: restarting dev server...`
```
adrien2p: Same suggestion

packages/medusa/src/commands/develop.js
```diff
@@ -23,6 +23,109 @@ export default async function ({ port, directory }) {
...
+        )
+        this.restart()
+      })
+      this.watcher.on("unlink", (file) => {
+        Logger.info(
+          `${path.relative(
+            directory,
+            file
+          )} file removed: restarting dev server...`
```
adrien2p: Same

adrien2p (approved):
Very nice 😊

packages/medusa/src/commands/develop.js
thetutlage: Can we instead say "Dev server exited with error"

The reason is "Dev server was running, but it crash or exited during the watch mode". WDYT?

kasperkristensen (approved):

packages/medusa/src/commands/develop.js
adrien2p: Sounds good to me 👌

packages/medusa/src/commands/develop.js
```diff
@@ -23,6 +23,100 @@ export default async function ({ port, directory }) {
...
+    },
+
+    /**
+     * Restarts the development server by cleaning up the existing
+     * child process and forking a new one
+     */
+    restart() {
+      if (this.childProcess) {
+        this.childProcess.removeAllListeners()
+        if (process.platform === "win32") {
```
sradevski, line 72: Of course... Windows 😄

sradevski (approved):
Good stuff 👍

packages/medusa/src/commands/develop.js
thetutlage: Hahah :)

PR #7438, opened 2024-05-24
feat(medusa,core-flows,types): fix bugs with cart promotions

riqwan:
what:

The true source of promotions applied are present in the adjustments, previously we had links as the source of truth, but some issues popped up when you're trying to delete a line item as it was holding onto stale promotion data.

For this to truly work properly in the current form, we would have to pass data from the deleted line item step to the refresh promotion step and mutate the links twice to get the correct promotions for a cart. Additionally, the final list of codes to apply, comes from computeAction, because some previously applied codes could no longer be valid. The combination of passing codes from adjustments to computeActions and the codes returned from the computedActions are how we can accurately calculate the right promotions to apply. 

With this change, the promotions step always look at the latest line item adjustments of the cart, pass them to compute actions, pick the correct codes to apply from the compute actions and then mutate the links based on the final result. This way, we perform the correct calculations for a cart and have promotion links that are correct.

### DIFF
--- a/packages/core/core-flows/src/definition/cart/steps/get-promotion-codes-to-apply.ts
+++ b/packages/core/core-flows/src/definition/cart/steps/get-promotion-codes-to-apply.ts
@@ -1,5 +1,5 @@
 import { ModuleRegistrationName } from "@medusajs/modules-sdk"
-import { CartDTO, IPromotionModuleService } from "@medusajs/types"
+import { IPromotionModuleService } from "@medusajs/types"
 import { PromotionActions } from "@medusajs/utils"
 import { StepResponse, createStep } from "@medusajs/workflows-sdk"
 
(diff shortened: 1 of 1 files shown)


packages/core/core-flows/src/definition/cart/steps/get-promotion-codes-to-apply.ts
```diff
@@ -0,0 +1,65 @@
...
+      promotionCodesToApply = promotionCodesToApply.filter(
+        (code) => !promo_codes.includes(code!)
+      )
+    }
+
+    if (action === PromotionActions.REPLACE) {
+      promotionCodesToApply = promo_codes
+    }
+
+    return new StepResponse(promotionCodesToApply)
```
adrien2p:
```suggestion
    const { promo_codes = [], cart, action = PromotionActions.ADD } = data
    const { items = [], shipping_methods = [] } = cart
    const adjustmentCodes: string[] = []
    const promotionService = container.resolve<IPromotionModuleService>(
      ModuleRegistrationName.PROMOTION
    )

    const objects = items.concat(shipping_methods)

    objects.forEach(object => {
      object.adjustments?.forEach(adjustment => {
        if (adjustment.code && !adjustmentCodes.includes(adjustment.code)) {
          adjustmentCodes.push(adjustment.code)
        }
      })
    })

    const promotionCodesToApply: Set<string> = new Set((
      await promotionService.list(
        { code: adjustmentCodes },
        { select: ["code"], take: null }
      )
    ).map((p) => p.code!))

    if (action === PromotionActions.ADD) {
      promo_codes.forEach(promotionCodesToApply.add)
    }

    if (action === PromotionActions.REMOVE) {
      promo_codes.forEach(promotionCodesToApply.delete)
    }

    if (action === PromotionActions.REPLACE) {
      promotionCodesToApply.clear()
      promo_codes.forEach(promotionCodesToApply.add)
    }

    return new StepResponse(Array.from(promotionCodesToApply))
```
take or leave, let me know what you think?

packages/core/core-flows/src/definition/cart/steps/prepare-adjustments-from-promotion-actions.ts
```diff
@@ -65,11 +65,17 @@ export const prepareAdjustmentsFromPromotionActionsStep = createStep(
       )
       .map((a) => (a as RemoveShippingMethodAdjustment).adjustment_id)
 
+    const computedPromotionCodes = [
+      ...lineItemAdjustmentsToCreate.map((adjustment) => adjustment.code),
+      ...shippingMethodAdjustmentsToCreate.map((adjustment) => adjustment.code),
+    ]
```
adrien2p:
```suggestion
    const computedPromotionCodes = [
      ...lineItemAdjustmentsToCreate,
      ...shippingMethodAdjustmentsToCreate,
    ].map((adjustment) => adjustment.code)
```

packages/core/core-flows/src/definition/cart/steps/get-promotion-codes-to-apply.ts
```diff
@@ -0,0 +1,65 @@
+import { ModuleRegistrationName } from "@medusajs/modules-sdk"
+import { CartDTO, IPromotionModuleService } from "@medusajs/types"
+import { PromotionActions } from "@medusajs/utils"
+import { StepResponse, createStep } from "@medusajs/workflows-sdk"
+
+interface StepInput {
+  cart: CartDTO
```
adrien2p:
```suggestion
  cart: {
    items?: { adjustments?: { code?: string }[] }[]
    shipping_methods?: { adjustments?: { code?: string }[] }[]
  }
```

adrien2p: you might be able to reuse some existing types, but the idea stays that we dont really need an entire Cart here but rather a specific data shape 👍

adrien2p (approved):

PR #7439, opened 2024-05-24
fix(customer): Unique constraint on customer email

adrien2p:
**What**
Prevent creating multiple customers with the same email


packages/modules/customer/src/models/customer.ts
```diff
@@ -22,8 +27,16 @@ type OptionalCustomerProps =
   | "addresses"
   | DAL.SoftDeletableEntityDateColumns
 
+const CustomerUniqueEmail = createPsqlIndexStatementHelper({
+  tableName: "customer",
```
sradevski, line 31: It should be a composite index on email, has_account right?

adrien2p: do we accept multiple guest account with the same email?

adrien2p: but you are probably right, @olivermrbl ?

sradevski: @adrien2p AFAIK, we accept 1 guest and 1 non-guest account per email.

adrien2p: so would one unique on has account true and one on has account false would be more appropriate? wdyt?

adrien2p: Like that you can have one of each or if you swith has account to true it works too

sradevski: @adrien2p that would work as well, it just might be a bit more costly performance-wise (on every update, you'll need to update 2 indices, rather than 1) But `UNIQUE INDEX ON (email, has_account)` should have the same effect, right?

adrien2p: yes thats right, I ll do that 💪

adrien2p: should be good for a last review @sradevski

sradevski (approved):
Nice 👍

PR #7440, opened 2024-05-24
feat: Update customer related typings and sdk methods

sradevski:


packages/core/types/src/http/customer/common.ts
```diff
@@ -0,0 +1,97 @@
...
+  address_1: string | null
+  address_2: string | null
+  city: string | null
+  country_code: string | null
+  province: string | null
+  postal_code: string | null
+  phone: string | null
+  metadata: Record<string, unknown> | null
+  created_at: string
+  updated_at: string
```
adrien2p, line 30: Q: isn't it always Date | string?

packages/medusa/src/api/store/orders/query-config.ts
```diff
@@ -59,5 +59,5 @@ export const retrieveTransformQueryConfig = {
 
 export const listTransformQueryConfig = {
   defaults: defaultStoreOrderFields,
-  isList: false,
+  isList: true,
```
adrien2p, line 62: nice catch 🙏

adrien2p (approved):
We are getting there 💪

packages/core/types/src/http/customer/common.ts
sradevski: @adrien2p I think when sending to the BE it first needs to be serialized to string, and the response will also need to be parsed from string to Date, that doesn't happen automatically currently.

PR #7442, opened 2024-05-24
chore(medusa, workflows-sdk): Workflow default to throw the first error

adrien2p:
**What**
Currently when a workflow fail it will throw an error which contains the messages of all error thrown durring the workflow lifetime. Therefore, in our cases we are always preventing workflow to throw and grab the first error that we then re throw.
This pr eliminate that need by throwing back the first error among the error thrown by a workflow as it is the main case. In case someone need a special handling they can still set the option throwOnError to false and handle the error the way they need

### DIFF
--- a/packages/medusa/src/loaders/index.ts
+++ b/packages/medusa/src/loaders/index.ts
@@ -53,8 +53,10 @@ async function subscribersLoader(
           ignore: ["**/*.d.ts", "**/*.map"],
         }
       )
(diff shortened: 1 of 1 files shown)


sradevski (approved):
So much better 👍

olivermrbl:
This is great @adrien2p 💪

carlos-r-l-rodrigues (approved):

PR #7450, opened 2024-05-24
chore: Start cleaning up medusa-core-utils

adrien2p:
**What**
- remove medusa-core-utils
- dispatch the utils where they belongs
- update usage

**NOTE**
I have been wondering if the graceful class should go into the utils package or medusa package, I ve put it in the medusa package as it seems to be the best place I can see for now and is tight to the server as well. Also, I wanted to avoid the utils package to depends on http and net dependencies, happy to change that if you feel like it

### DIFF
--- a/.changeset/little-books-reply.md
+++ b/.changeset/little-books-reply.md
@@ -17,7 +17,6 @@
 "@medusajs/ui": patch
 "@medusajs/ui-preset": patch
 "@medusajs/medusa": patch
-"medusa-core-utils": patch
 "medusa-telemetry": patch
 "@medusajs/api-key": patch
 "@medusajs/auth": patch


yarn.lock
```diff
@@ -23831,18 +23831,14 @@ __metadata:
   languageName: node
   linkType: hard
 
-"medusa-core-utils@^1.2.0, medusa-core-utils@^1.2.2, medusa-core-utils@workspace:packages/medusa-core-utils":
-  version: 0.0.0-use.local
-  resolution: "medusa-core-utils@workspace:packages/medusa-core-utils"
+"medusa-core-utils@npm:^1.2.0, medusa-core-utils@npm:^1.2.2":
```
carlos-r-l-rodrigues: shouldn't this be removed?

carlos-r-l-rodrigues (approved):
🥇

yarn.lock
adrien2p: Yes ill re run a yarn

packages/core/medusa-test-utils/src/medusa-test-runner-utils/bootstrap-app.js
```diff
@@ -2,7 +2,9 @@ const path = require("path")
 const express = require("express")
 const getPort = require("get-port")
 const { isObject, promiseAll } = require("@medusajs/utils")
-const { GracefulShutdownServer } = require("medusa-core-utils")
+const {
```
sradevski, line 6: Obviously this isn't great but it'll do for now (let's just add a TODO comment to remember and handle it at some later point)

packages/medusa/src/loaders/helpers/resolve-plugins.ts
```diff
@@ -14,6 +14,19 @@ function createFileContentHash(path, files): string {
...
+  const mod = new Module(filename)
+
+  mod.filename = filename
+  mod.paths = (Module as any)._nodeModulePaths(path.dirname(filename))
+  ;(mod as any)._compile(`module.exports = require;`, filename)
+
+  return mod.exports
+}
+
+// Polyfill Node's `Module.createRequireFromPath` if not present (added in Node v10.12.0)
```
sradevski: I don't think we need this polyfill anymore, right?

sradevski (approved):
LGTM, approving, just a couple of comments

packages/core/medusa-test-utils/src/medusa-test-runner-utils/bootstrap-app.js
sradevski: Btw I am not sure what the best solution is here, but we don't need to solve for it now, we can revisit at a later point.

olivermrbl (approved):

PR #7455, opened 2024-05-25
fix: Add typings to list shipping options

sradevski:


adrien2p (commented):
I think something went wrong with your rebase 🤣 my pr is in yours ahah

sradevski:
@adrien2p yeah weird, fixed it now.

packages/core/types/src/http/fulfillment/store/index.ts
```diff
@@ -0,0 +1,22 @@
+// TODO: The way the cart shipping options are listed now differs from most other endpoints as it is fetched in a workflow.
```
olivermrbl, line 1: question: I think we'll have other GET endpoints that will need to use workflows as well – what's your thinking?

olivermrbl (approved):

packages/core/types/src/http/fulfillment/store/index.ts
sradevski: @oliverheaps I think it's fine to use workflows, but selecting fields etc. doesn't quite work now, so that part might need to be fixed. Otherwise it's fine to use either remote query or workflows (although for purely reading data I don't think workflows are really necessary as they do have overhead)

adrien2p: I think workflow can be useful if we have computed result which add an extra logic step to it, otherwise the remote query should suffice. In reguards to the fields, I would say that the workflow needs an update to accept the requested fields etc wdyt?

sradevski: @adrien2p yeah exactly @adrien2p , the workflow needs to take in similar arguments to `remote-query` so that, before the result is returned, we select everything the user has asked for.

On when to use workflows and when to use remove-query, I think we'll need to look at it on a case-by-case basis, I just feel like if it is purely a reading function, it can easily be a normal JS function that does multiple `remoteQuery` calls instead

PR #7464, opened 2024-05-27
chore: Cleanup medusa test utils

adrien2p:
This pr does not include the integration tests api cleanup

### DIFF
--- a/integration-tests/modules/__tests__/modules/load-standalone.ts
+++ b/integration-tests/modules/__tests__/modules/load-standalone.ts
@@ -5,10 +5,10 @@ import { IProductModuleService } from "@medusajs/types"
 jest.setTimeout(30000)
 
 medusaIntegrationTestRunner({
-  testSuite: ({ dbConnection }) => {
+  testSuite: ({ dbConfig: { clientUrl } }) => {
     describe("Standalone Modules", () => {
       beforeAll(async () => {
-        process.env.POSTGRES_URL = dbConnection.manager.connection.options.url
+        process.env.POSTGRES_URL = clientUrl
       })
 
       afterAll(async () => {


packages/core/medusa-test-utils/package.json
```diff
@@ -54,9 +53,6 @@
     },
     "pg-god": {
       "optional": true
-    },
-    "typeorm": {
```
sradevski, line 58: Uh, nice :D

packages/core/medusa-test-utils/src/medusa-test-runner-utils/bootstrap-app.ts
```diff
@@ -0,0 +1,83 @@
+import express from "express"
```
sradevski, line 1: It would be nice if this and the init DB process was a bit simpler. Nothing to do for now, just talking out loud 😄

sradevski (approved):

yarn.lock
```diff
@@ -23712,7 +23712,6 @@ __metadata:
     express: ^4.18.3
     get-port: ^5.1.0
     pg-god: ^1.0.12
-    typeorm: ^0.2.43
```
olivermrbl, line 23715: yay!

packages/core/medusa-test-utils/src/medusa-test-runner-utils/bootstrap-app.ts
adrien2p: yes I do agree, but probably another time in another pr, it would also be nice that those utils can be used by any users. As of now, they could probably use the medusa runner in their project, but the module runner should not care about the orm and probably just use knex for a util if you still need some sort of interaction level like we do for the medusa runner. But lets keep that for another time, I have some ideas but it is not yet a prio

olivermrbl (approved):

PR #7470, opened 2024-05-27
feat(core-flows,types,utils,medusa): Update existing line items when adding the same variant to cart

riqwan:
what:

- when you're adding the same variant with the same metadata to cart, we update the quantity of an existing line item instead of creating a new line item

### DIFF
--- a/packages/core/core-flows/src/definition/cart/workflows/add-to-cart.ts
+++ b/packages/core/core-flows/src/definition/cart/workflows/add-to-cart.ts
@@ -5,6 +5,7 @@ import {
 import {
   WorkflowData,
   createWorkflow,
+  parallelize,
   transform,
 } from "@medusajs/workflows-sdk"
 import { useRemoteQueryStep } from "../../../common/steps/use-remote-query"
@@ -88,15 +89,16 @@ export const addToCartWorkflow = createWorkflow(
       items: lineItems,
     })
 
-    const createdItems = createLineItemsStep({
-      id: input.cart.id,
-      items: itemsToCreate,
-    })
-
-    const updatedItems = updateLineItemsStep({
-      id: input.cart.id,
-      items: itemsToUpdate,
-    })
+    const [createdItems, updatedItems] = parallelize(
+      createLineItemsStep({
+        id: input.cart.id,
+        items: itemsToCreate,
+      }),
+      updateLineItemsStep({
+        id: input.cart.id,
+        items: itemsToUpdate,
+      })
+    )
 
     const items = transform({ createdItems, updatedItems }, (data) => {
       return [...(data.createdItems || []), ...(data.updatedItems || [])]


packages/core/core-flows/src/definition/cart/steps/add-to-cart.ts
```diff
@@ -1,31 +1,103 @@
...
+          selector: { id: existingItem.id },
+          data: { id: existingItem.id, quantity: quantity },
+        })
+      } else {
+        itemsToCreate.push(item)
+      }
+    }
+
+    const createdItems = itemsToCreate.length
+      ? await cartModule.addLineItems(itemsToCreate)
```
carlos-r-l-rodrigues: do we have an upsert method?
"addLineItems" and "updateLineItems" should happen in 2 different steps to be able to revert.

adrien2p: I believe it should work here since we are reverting both in the rollback no? it acts as an upsert, but indeed would be nice to have an upsert method instead on the module

packages/core/core-flows/src/definition/cart/steps/add-to-cart.ts
```diff
@@ -1,31 +1,103 @@
...
+
+      for (const item of itemsBeforeUpdate) {
+        itemsToUpdate.push({
+          selector: { id: item.id },
+          data: { id: item.id, quantity: item.quantity },
+        })
+      }
+
+      await cartModule.updateLineItems(itemsToUpdate)
+    }
```
adrien2p: suggestion:

```ts
    
    const promises: Promise<any>[] = []

    if (itemsBeforeUpdate.length) {
      const itemsToUpdate: UpdateLineItemWithSelectorDTO[] = []

      for (const item of itemsBeforeUpdate) {
        itemsToUpdate.push({
          selector: { id: item.id },
          data: { id: item.id, quantity: item.quantity },
        })
      }

      promises.push(cartModule.updateLineItems(itemsToUpdate))
    }

    if (createdItems.length) {
      promises.push(cartModule.deleteLineItems(createdItems.map((c) => c.id)))
    }
    
    await Promise.all(promises)
```

adrien2p (approved):

packages/core/core-flows/src/definition/cart/steps/add-to-cart.ts
carlos-r-l-rodrigues: if one of them fail we have a problem. for example the `add` works and `update` fails. we don't have anything on the compensation step as response.
Each step can only contain one operation with their equivalent compensation.

adrien2p: yes thats true 👍

packages/core/core-flows/src/definition/cart/steps/add-to-cart.ts
adrien2p: nvm this comment as it needs some reworks

packages/core/core-flows/src/definition/cart/steps/add-to-cart.ts
riqwan: Good catch, I've split the step into create and update in 2 steps 🤞🏻

packages/core/core-flows/src/definition/cart/workflows/add-to-cart.ts
```diff
@@ -78,7 +83,24 @@ export const addToCartWorkflow = createWorkflow(
...
+      id: input.cart.id,
+      items: lineItems,
+    })
+
+    const createdItems = createLineItemsStep({
+      id: input.cart.id,
+      items: itemsToCreate,
+    })
+
+    const updatedItems = updateLineItemsStep({
```
carlos-r-l-rodrigues: that can be parallelized

carlos-r-l-rodrigues (approved):

PR #7474, opened 2024-05-27
fix: Listening events

adrien2p:
Fix listening event on app listen


thetutlage:
Just wondering, how it was working earlier.

adrien2p:
> Just wondering, how it was working earlier. 

Probably a missmatch version installed?

sradevski (approved):

PR #7492, opened 2024-05-28
fix: Use the correct header for publishable key

sradevski:


packages/core/js-sdk/src/__tests__/client.spec.ts
```diff
@@ -44,7 +44,7 @@ const server = setupServer(
     }
   }),
   http.get(`${baseUrl}/pubkey`, ({ request, params, cookies }) => {
-    if (request.headers.get("x-medusa-pub-key") === "test-pub-key") {
+    if (request.headers.get("x-publishable-api-key") === "test-pub-key") {
```
adrien2p: Q: should we put that in a constant in the package? in case of future changes?

adrien2p (approved):

packages/core/js-sdk/src/__tests__/client.spec.ts
sradevski: Yeah we can, let me do that 👍

fPolic (approved):

PR #7497, opened 2024-05-28
chore(order): aggregate statuses

carlos-r-l-rodrigues:


packages/admin-next/dashboard/src/routes/orders/order-detail/components/order-activity-section/order-activity-section.tsx
```diff
@@ -1,11 +1,11 @@
+import { OrderDetailDTO } from "@medusajs/types"
 import { Container, Heading } from "@medusajs/ui"
 import { useTranslation } from "react-i18next"
 import { OrderNoteForm } from "./order-note-form"
 import { OrderTimeline } from "./order-timeline"
-import { OrderDTO } from "@medusajs/types"
 
 type OrderActivityProps = {
-  order: OrderDTO
+  order: OrderDetailDTO
```
sradevski: Can we maybe create the type in `http/orders` instead of using the module DTO?

packages/core/core-flows/src/order/workflows/get-order-detail.ts
```diff
@@ -0,0 +1,159 @@
+import { OrderDTO, OrderDetailDTO } from "@medusajs/types"
+import { deduplicate } from "@medusajs/utils"
+import {
+  WorkflowData,
+  createWorkflow,
+  transform,
+} from "@medusajs/workflows-sdk"
+import { useRemoteQueryStep } from "../../common"
+
+enum PaymentStatus {
```
sradevski: I am not sure what is best to do around enums, but at least for the http typings we said we'll just go with a string union instead. Now, for workflows and BE enums we can continue adding them to `utils`, or maybe we also go with string unions? WDYT?

At the very least, these should be in either `types` or `utils`, not inside the workflow

packages/core/core-flows/src/order/workflows/get-order-detail.ts
```diff
@@ -0,0 +1,159 @@
...
+  REQUIRES_ACTION = "requires_action",
+}
+
+export const getOrderDetailWorkflowId = "get-order-detail"
+export const getOrderDetailWorkflow = createWorkflow(
+  getOrderDetailWorkflowId,
+  (
+    input: WorkflowData<{ fields: string[]; order_id: string }>
+  ): WorkflowData<OrderDetailDTO> => {
+    const fields = transform(input, ({ fields }) => {
```
sradevski, line 20: Should this happen in the HTTP layer instead? Prob. in the middleware

packages/core/core-flows/src/order/workflows/get-order-detail.ts
```diff
@@ -0,0 +1,159 @@
...
+      fields,
+      variables: { id: input.order_id },
+      list: false,
+      throw_if_key_not_found: true,
+    })
+
+    const aggregatedOrder = transform({ order }, ({ order }) => {
+      const order_ = order as OrderDetailDTO
+
+      const getLastPaymentStatus = (order) => {
```
sradevski: Should we move these outside of the workflow, and write some unit tests for them? I feel like they can get quite complex and easy to get wrong.

adrien2p (approved):
LGTM, so moves to do around enums/types but otherwise 💪

packages/core/core-flows/src/order/workflows/get-order-detail.ts
carlos-r-l-rodrigues: This is a read only workflow where we have to make sure a couple of relations are always present. overall I don't like this flow, nor the aggregations it does, I am just trying to keep the fields we have in v1 for now.

packages/core/core-flows/src/order/workflows/get-order-detail.ts
carlos-r-l-rodrigues: will move them to utils.

packages/core/core-flows/src/order/workflows/get-order-detail.ts
carlos-r-l-rodrigues: The typings are using string union, here I just need these locally. Will bring it to the functions on the utils

sradevski: 👍

PR #7498, opened 2024-05-28
chore: Rename config properties to camelCase

thetutlage:
Renaming config properties to be camel case.

Side note: Not able to run build on the local. So the only option I have is to push changes and use the CI as the only feedback channel.


adrien2p (approved):
Looks good to me, did you go through the entire code base?

olivermrbl (approved):

PR #7500, opened 2024-05-28
chore: buyget templates add default target + buy rules

riqwan:
what:

- adds a default rule set to buyget templates
- cleanup

### DIFF
--- a/packages/medusa/src/api/admin/promotions/middlewares.ts
+++ b/packages/medusa/src/api/admin/promotions/middlewares.ts
@@ -1,4 +1,5 @@
 import { MiddlewareRoute } from "../../../loaders/helpers/routing/types"
+import { unlessPath } from "../../utils/unless-path"
 import { validateAndTransformBody } from "../../utils/validate-body"
 import { validateAndTransformQuery } from "../../utils/validate-query"
 import { createBatchBody } from "../../utils/validators"
@@ -62,9 +63,12 @@ export const adminPromotionRoutesMiddlewares: MiddlewareRoute[] = [
     method: ["GET"],
     matcher: "/admin/promotions/:id/:rule_type",
     middlewares: [
-      validateAndTransformQuery(
-        AdminGetPromotionRuleTypeParams,
-        QueryConfig.retrieveTransformQueryConfig
+      unlessPath(
+        /.*\/promotions\/rule-attribute-options/,
+        validateAndTransformQuery(
+          AdminGetPromotionRuleTypeParams,
+          QueryConfig.retrieveTransformQueryConfig
+        )
       ),
     ],
   },
@@ -115,6 +119,16 @@ export const adminPromotionRoutesMiddlewares: MiddlewareRoute[] = [
       validateAndTransformQuery(
         AdminGetPromotionsRuleValueParams,
         QueryConfig.listRuleValueTransformQueryConfig
+      ),
+    ],
+  },
+  {
+    method: ["GET"],
+    matcher: "/admin/promotions/rule-attribute-options/:rule_type",
+    middlewares: [
+      validateAndTransformQuery(
+        AdminGetPromotionRuleParams,
+        QueryConfig.listRuleTransformQueryConfig
       ),
     ],
   },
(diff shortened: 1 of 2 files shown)


packages/admin-next/dashboard/src/lib/client/promotions.ts
```diff
@@ -81,15 +81,22 @@ async function removePromotionRules(
...
 
-async function listPromotionRules(id: string | null, ruleType: string) {
+async function listPromotionRules(
+  id: string | null,
+  ruleType: string,
+  promotionType?: string
+) {
   return getRequest<PromotionRuleAttributesListRes>(
-    `/admin/promotions/${id}/${ruleType}`
+    `/admin/promotions/${id}/${ruleType}?promotion_type=${promotionType}`
```
sradevski: Should we do 
```suggestion
if(promotionType) {
const promoFilter = `promotion_type=${promotionType}`
...
```

Or does passing an empty `promotion_type=` work?

packages/admin-next/dashboard/src/routes/promotions/common/edit-rules/components/edit-rules-form/edit-rules-form.tsx
```diff
@@ -39,65 +37,38 @@ const EditRules = zod.object({
...
-
   const handleFormSubmit = form.handleSubmit(handleSubmit(rulesToRemove))
 
   return (
     <RouteDrawer.Form form={form}>
       <form onSubmit={handleFormSubmit} className="flex h-full flex-col">
         <RouteDrawer.Body>
           <RulesFormField
-            form={form}
+            form={form as any}
```
sradevski: What's the issue with the typings here, maybe we can resolve it without any?

packages/admin-next/dashboard/src/routes/promotions/common/edit-rules/components/rules-form-field/rules-form-field.tsx
```diff
@@ -34,20 +26,84 @@ type RulesFormFieldType<TSchema extends FieldValues> = {
...
+  const { attributes } = usePromotionRuleAttributes(ruleType, formData.type)
+  const { operators } = usePromotionRuleOperators()
+
+  const { fields, append, remove, update, replace } = useFieldArray({
+    control: form.control,
+    name: scope,
+    keyName: scope,
+  })
+
+  const watchType = useWatch({
```
sradevski: Can we rename this to something more obvious (it sounds like a function now), eg. `ruleTypeValue`

packages/admin-next/dashboard/src/routes/promotions/common/edit-rules/components/edit-rules-form/edit-rules-form.tsx
riqwan: This was legacy, forgot to remove it. 🤞🏻 Good catch

packages/admin-next/dashboard/src/routes/promotions/common/edit-rules/components/edit-rules-wrapper/edit-rules-wrapper.tsx
```diff
@@ -60,36 +40,24 @@ export const EditRulesWrapper = ({
...
 
   const handleSubmit = (rulesToRemove?: { id: string }[]) => {
-    return async function (data: { rules: PromotionRuleDTO[] }) {
+    return async function (data: { rules: PromotionRuleResponse[] }) {
       const applicationMethodData: Record<any, any> = {}
       const { rules: allRules = [] } = data
-      const disguisedRulesData = allRules.filter((rule) =>
-        disguisedRules.map((rule) => rule.id).includes(rule.id!)
-      )
+      const diguisedRules = allRules.filter((rule) => rule.disguised)
```
fPolic: nit: typo

fPolic (commented):
Not sure if this is directly related to this PR but when I select "By Get" type, I cannot submit request successfully since the rules array is always empty?

fPolic (approved):
Looks good!

One minor thing we could address before merge is that when you click "clear all" required fields are not reset to default/empty values.

Also, I noticed that we set default values for amount and percentage to 100 which is not great from the UX perspective. Think that we shouldn't set defaults there and use placeholders.

riqwan:
Thanks @fPolic , I'll create a ticket with these changes. I can't seem to easily make the reset work for the select without creating additional bugs. 😅 Ticket - CORE-2272

PR #7502, opened 2024-05-28
feat(dashboard,types,js-sdk): Cleanup collection domain

kasperkristensen:
**What**
- Adds missing collection HttpTypes
- Adds missing sdk functions
- Adds usage of sdk to collection domain.


packages/admin-next/dashboard/src/hooks/api/collections.tsx
```diff
@@ -90,19 +90,24 @@ export const useUpdateCollection = (
 export const useUpdateCollectionProducts = (
   id: string,
   options?: UseMutationOptions<
-    ProductCollectionRes,
-    Error,
-    UpdateProductCollectionProductsReq
+    { collection: HttpTypes.AdminCollection },
```
adrien2p, line 85: note: I ve seen that we have some types that include `Response` at the end of the types and other are not, we will have to align. not for this pr but more of a general comment :D

packages/admin-next/dashboard/src/hooks/api/collections.tsx
```diff
@@ -90,19 +90,24 @@ export const useUpdateCollection = (
...
+    FetchError,
+    AdminUpdateCollectionProducts
```
adrien2p: suggestion: should we take all the types from the HttpTypes to stay coherent through the usage?

sradevski (approved):
Nice 👍 Approving, but check Adrien's comments

adrien2p (approved):

packages/admin-next/dashboard/src/hooks/api/collections.tsx
kasperkristensen: If we exported the the type as `{ collection: AdminCollection }` then I think it should be suffixed with Response. But none of the other recent ones are exported like that, only the `Admin<Entity>` is exported, so I stuck with that for this PR as well. But yea we need to align on this, for sure.

PR #7508, opened 2024-05-28
chore: Remove currency + country logs

olivermrbl:


olivermrbl:
Just a minor clean-up - doesn't look good that it's printed when the dev server is in watch mode 😅

PR #7509, opened 2024-05-28
feat(admin-sdk): Merge users Vite config

kasperkristensen:
**What**
- Allows users to customize the Vite config from `admin.vite` in `medusa-config.js`

**How**
```
/** @type {import('@medusajs/types').ConfigModule} */
module.exports = {
  projectConfig,
  admin: {
    backendUrl: "http://localhost:9000",
    // the config param is our default config
    vite: (config) => {
      return {
        define: {
          __TEST__: JSON.stringify("test"), // this will now be defined as a const that can be accessed in widgets
        },
      };
    },
  },
 // ...
};
```

The `vite` param is a callback that takes our default config as a param. If the user does not need to access our config they don't have to use it, as we merge their config with our own, but accessing the default config can be useful in some cases.


packages/admin-next/admin-sdk/src/lib/config.ts
```diff
@@ -59,6 +59,14 @@ export async function getViteConfig(
       }),
     ],
   }
+
+  if (options.vite) {
+    const customConfig = options.vite(baseConfig)
+
+    return mergeConfig(baseConfig, customConfig)
```
olivermrbl, line 66: question: Do we need some validation on what is passed to `vite: () => {}`?

olivermrbl (approved):

packages/admin-next/admin-sdk/src/lib/config.ts
kasperkristensen: Don't think its realistic to validate what is passed, as they the config is very deep (it has almost all rollup and esbuild options). When we get the `defineConfig` the type will be there to help guide people. Also Vite won't crash if you return an invalid config, it will just discard it.

olivermrbl: > Also Vite won't crash if you return an invalid config, it will just discard it.

Cool, sounds good

PR #7510, opened 2024-05-28
Ensure sales channel updates don't remove sales channel on other products

sradevski:


packages/core/core-flows/src/product/workflows/update-products.ts
```diff
@@ -38,6 +38,7 @@ function prepareUpdateProductInput({
 }: {
   input: WorkflowInput
 }): UpdateProductsStepInput {
+  console.log("INPUT", input)
```
adrien2p: Todo rm

packages/core/core-flows/src/product/workflows/update-products.ts
```diff
@@ -81,6 +83,8 @@ function prepareSalesChannelLinks({
   updatedProducts: ProductTypes.ProductDTO[]
   input: WorkflowInput
 }): Record<string, Record<string, any>>[] {
+  console.log("TO CREATE", input, updatedProducts)
```
adrien2p: Todo rm

adrien2p (commented):
Why did we remove all inventory_quantity? It seems unrelated to the pr scope no?

packages/core/core-flows/src/product/workflows/update-products.ts
sradevski: @adrien2p gosh :/ done

sradevski:
@adrien2p they were already removed everywhere else, it's just that they were not removed in the api tests. The tests were failing because of that.

adrien2p (approved):
It seems to be good for me 👍

carlos-r-l-rodrigues (approved):

olivermrbl (approved):

PR #7511, opened 2024-05-28
fix: Change default from empty string to undefined

olivermrbl:
You can't update more than one variant on a product right now because of uniqueness constraints on `sku`, `ean`,`upc`, and `barcode`. 

These values are by default set to empty strings and will be sent like in the requests to the backend. 

Example:
- Update variant one title to "Variant One"
```
// payload
{ "title": "Variant One", ean: "", barcode: "", sku: "" }
```
- Update variant two title to "Variant two"
```
// payload
{ "title": "Variant Two", ean: "", barcode: "", sku: "" }
```

Second request fails because the empty string on ean, barcode, and sku already exists.


adrien2p (approved):
Question, should we also update the uniqueness on not null and not empty? Or should it still be unique ?

olivermrbl:
Yeah, I thought of that. So on the nullable column we would have a `unique index where [column name] <> ''`

adrien2p:
> Yeah, I thought of that. We can do that on the optional columns?

Yes we can, and our util allow to specify such constraints

olivermrbl:
Sorry, I didn't mean to phrase it as a question. I meant, I was gonna do it 😅

olivermrbl:
I'll do that in a separate PR

adrien2p:
Alright, sounds good to me

kasperkristensen (changes requested):
This isn't allowed in React, as it would cause the component to change from uncontrolled to controlled. We should instead clean the request before we send it off.

olivermrbl:
> This isn't allowed in React

Got it, I think I'll go ahead with the index approach first then, and let you guys handle the admin side of it. 

Just out of curiosity, was happens exactly making this incorrect? Everything seems to be working fine with the change, and I "borrowed" this approach from the old v1 routes.

olivermrbl:
Also, how do we clean empty strings from the request without conflicting with something you want to actually set as the empty string? Do we need to convert them to `null` to unset values?

kasperkristensen:
> > This isn't allowed in React
> 
> Got it, I think I'll go ahead with the index approach first then, and let you guys handle the admin side of it.
> 
> Just out of curiosity, was happens exactly making this incorrect? Everything seems to be working fine with the change, and I "borrowed" this approach from the old v1 routes.

If you look at your console you will see that React throws an error each time you edit one of the fields that have been initialized with `undefined`. In React a form input can be controlled or uncontrolled and it should not change between these two states. React will continue to working, but it throws an error each time an input switches between these two states, so we should avoid it 👍

PR #7512, opened 2024-05-28
fix(dashboard): Clean Edit Variant form payload of empty strings

kasperkristensen:


kasperkristensen:
@olivermrbl we can do this instead, I think this needs to be applied all over the place, so I have added a util for cleaning up payloads.

olivermrbl:
@kasperkristensen, how do you unset a value with this? This was part of the reason for the other PR. We can't clean up empty strings, because they could have been passed intentionally. 

E.g. changing `barcode` from `"BC1234"` to `""`.

packages/admin-next/dashboard/src/lib/form-helpers.ts
```diff
@@ -0,0 +1,25 @@
+import { castNumber } from "./cast-number"
+
+export function parseOptionalFormValue<T>(value: T): T | undefined {
+  if (typeof value === "string" && value.trim() === "") {
```
adrien2p, line 7: Q: how do we set something to an empty string if wanted to?

packages/admin-next/dashboard/src/lib/form-helpers.ts
```diff
@@ -0,0 +1,25 @@
...
+    return undefined
+  }
+
+  if (Array.isArray(value) && value.length === 0) {
```
adrien2p, line 11: Todo: it would mean that we can remove all items from an object anymore. Not sure we want that

packages/admin-next/dashboard/src/lib/form-helpers.ts
```diff
@@ -0,0 +1,25 @@
...
+    return undefined
+  }
+
+  return value
+}
+
+export function parseOptionalFormNumber(value?: string | number) {
+  if (typeof value === "undefined" || value === "") {
```
adrien2p: Q: should we also trim?

kasperkristensen:
@olivermrbl You are right, so I am thinking the helper can take a second param of `nullify` which should be true for "Edit" forms, where an empty string or an empty array should be translated to null to unset the value. For create forms we would still want it to be undefined, so we don't pass the values along at all. I'll add it to this PR, but I also need to update the validatator for the endpoint, as you currently aren't allowed to pass anything but string or undefined to alot of the fields 😅

packages/admin-next/dashboard/src/lib/form-helpers.ts
kasperkristensen: Why would you want to set anything to an empty string? Don't think that makes sense from a form input perspective, as an empty string equals undefined or null in practice as you haven't input anything.

adrien2p: Undefined is stripped from the json once sent no? and for us, undefined means a property as not been provided if i am right. In general for the value we look for null, empty and value cc @olivermrbl

olivermrbl:
> You are right, so I am thinking the helper can take a second param of nullify which should be true for "Edit" forms, where an empty string or an empty array should be translated to null to unset the value

Think this sounds good, let's give it a shot. 

> I also need to update the validatator for the endpoint, as you currently aren't allowed to pass anything but string or undefined to alot of the fields

Yeah, I imagine we will need to make many fields nullable. I really think we need to solve this duplicate typings issue once and for all soon so we don't have to maintain a single type in four layers of the stack 😅

adrien2p:
@olivermrbl updating the index would help for that as well wdyt?

olivermrbl:
> @olivermrbl updating the index would help for that as well wdyt?

Yup, combined with owning the data model, I think we can solve it for good.

packages/admin-next/dashboard/src/lib/form-helpers.ts
sradevski: You maybe want to set a product subtitle to an empty string (or in other words, unset it). With this approach, that won't be possible.

What we can do is check if the initial value differs from the current one (i.e initially it wasn't empty, but it got set to empty when submitting).

packages/admin-next/dashboard/src/lib/form-helpers.ts
sradevski: We can use the same mechanism as I mentioned above.

packages/admin-next/dashboard/src/lib/form-helpers.ts
kasperkristensen: If you want to unset the subtitle you should send null to the backend. I have just updated the PR to also allow passing null for all the nullable fields, as well as this util so it now transforms empty string into null.

kasperkristensen:
So checking the validators we have a large cleanup task in making fields nullable in our validators, and in the types for our core-workflows. We properly need to do this quite quickly as it makes it impossible to unset things from admin without simply changing everything to empty strings which is not the correct solution.

packages/admin-next/dashboard/src/lib/form-helpers.ts
sradevski: @kasperkristensen how will the caller of this function decide what to nullify though? Isn't it easy if we do that calculation based on initial and current value?

kasperkristensen: @sradevski if it's a create form, you properly don't want to nullify the fields, but instead convert them to `undefined` so the values aren't sent off to the backend. But it does not really matter, as the fields that are nullable, are also null by default in our models. But in most cases you would set nullify to `false` here. In edit forms, you want to send null because an empty field means you want to remove the value. We could also check the diff between initial values and current values, but I don't see the point in adding additional code for this, when the above pattern is always true. But maybe I am missing something, but my opinions is that:

1. Our validators should accept `null` for all fields that are nullable, in order to allow users to correctly unset the value, and not have to set empty strings for all of their fields, which also does not work for number fields that should be nullable.
2. In edit forms an empty input === null, in create forms an empty input === undefined.

sradevski: @kasperkristensen The only concern I had is in your second point, on edit an empty input === null, it could be that we want an empty string instead. But maybe it's an edge case so le's try this and see if we stumble upon issues.

olivermrbl (approved):

PR #7513, opened 2024-05-28
chore: payment auto capture + remove enum from utils

carlos-r-l-rodrigues:
What:

- handles payments session that are auto captured
- removes all the enums from `@medusajs/types`
- move all `@medusajs/types` to devDependencies


packages/core/types/src/payment/common.ts
```diff
@@ -2,69 +2,21 @@ import { BaseFilterable } from "../dal"
...
-   * An error occurred while processing the payment.
-   */
-  ERROR = "error",
-
-  /**
-   * The payment is canceled.
-   */
-  CANCELED = "canceled",
-}
+export type PaymentCollectionStatus =
```
riqwan, line 6: Not entirely sure if this is a convention, but in other places I saw this pattern where the type is postfixed with Values:

```
enum PaymentStatus = {
  PAID = "paid",
  PENDING = "pending"
}
type PaymentStatusValues = "paid" | "pending"
```

packages/modules/payment/src/services/payment-module.ts
```diff
@@ -434,14 +436,19 @@ export default class PaymentModuleService<
...
 
-    const { data, status } =
-      await this.paymentProviderService_.authorizePayment(
-        {
-          provider_id: session.provider_id,
-          data: session.data,
-        },
-        context
-      )
+    let { data, status } = await this.paymentProviderService_.authorizePayment(
```
riqwan, line 439: q: should we try/catch here and throw a generic medusa error?

riqwan (approved):
💯

packages/modules/payment/src/services/payment-module.ts
adrien2p: we already handle that in the payment provider service

adrien2p (approved):

packages/modules/payment/src/services/payment-module.ts
carlos-r-l-rodrigues: That's a good point, there is a `PaymentSessionStatus.ERROR` as a type but at the moment the only provider we have is never returning that.
We have to refactor the Payment module to not block the db transaction when calling the provider, and I will test this entire flow a bit more in a next PR.

packages/core/types/src/payment/common.ts
carlos-r-l-rodrigues: But in this case we'll keep the enum in the package, which will generated JS code.
The `@medusajs/types` should only be used as devDependency, and none package should reference it in the code.

packages/modules/payment/src/services/payment-module.ts
adrien2p: but if you look at what we have done in the providers, we return some errors then the provider-service get that error and throw a MedusaError, so you should be good in that regards

carlos-r-l-rodrigues: you mean in v1? because in v2, I don't see that.

adrien2p: Yes in v2 as well, if the provider sdk throw we catch and return a formatted error and in the provider service then this.throwPaymentProviderError(error), otherwise we throw a medusa error from the provider if it is un handled. You can look at the payment/src/services/provider service

carlos-r-l-rodrigues: I see, I was looking at the stripe implementation where authorizePayment isn't handling anything because it is just reading the status.

PR #7515, opened 2024-05-29
fix: Use correct config for auth methods per actor

sradevski:


packages/medusa/src/api/auth/[actor_type]/[auth_provider]/callback/route.ts
```diff
@@ -14,10 +14,11 @@ export const GET = async (req: MedusaRequest, res: MedusaResponse) => {
...
   )
 
-  const authMethods = (config.projectConfig?.http as any)?.authMethods ?? {}
+  const authMethodsPerActor =
+    config.projectConfig?.http?.authMethodsPerActor ?? {}
   // Not having the config defined would allow for all auth providers for the particular actor.
-  if (authMethods[actor_type]) {
-    if (!authMethods[actor_type].includes(auth_provider)) {
+  if (authMethodsPerActor[actor_type]) {
+    if (!authMethodsPerActor[actor_type].includes(auth_provider)) {
```
Comment (adrien2p) [inline packages/medusa/src/api/auth/[actor_type]/[auth_provider]/callback/route.ts:21]:
```suggestion
    if (!authMethodsPerActor[actor_type]?.includes(auth_provider)) {
```

Comment (adrien2p) [reply packages/medusa/src/api/auth/[actor_type]/[auth_provider]/callback/route.ts:21]: otherwise if the auth actor provided is not configured, it will go through, wdyt?

packages/medusa/src/api/auth/[actor_type]/[auth_provider]/route.ts
```diff
@@ -14,10 +14,11 @@ export const GET = async (req: MedusaRequest, res: MedusaResponse) => {
...
   )
 
-  const authMethods = (config.projectConfig?.http as any)?.authMethods ?? {}
+  const authMethodsPerActor =
+    config.projectConfig?.http?.authMethodsPerActor ?? {}
   // Not having the config defined would allow for all auth providers for the particular actor.
-  if (authMethods[actor_type]) {
-    if (!authMethods[actor_type].includes(auth_provider)) {
+  if (authMethodsPerActor[actor_type]) {
+    if (!authMethodsPerActor[actor_type].includes(auth_provider)) {
```
Comment (adrien2p) [inline packages/medusa/src/api/auth/[actor_type]/[auth_provider]/route.ts:21]:
```suggestion
    if (!authMethodsPerActor[actor_type]?.includes(auth_provider)) {
```

packages/medusa/src/api/auth/[actor_type]/[auth_provider]/callback/route.ts
Comment (sradevski) [reply packages/medusa/src/api/auth/[actor_type]/[auth_provider]/callback/route.ts:21]: @adrien2p that's what we want to achieve, if it is not defined it should allow any login method by default

packages/medusa/src/api/auth/[actor_type]/[auth_provider]/callback/route.ts
```diff
@@ -14,10 +14,11 @@ export const GET = async (req: MedusaRequest, res: MedusaResponse) => {
     ContainerRegistrationKeys.CONFIG_MODULE
   )
 
-  const authMethods = (config.projectConfig?.http as any)?.authMethods ?? {}
+  const authMethodsPerActor =
+    config.projectConfig?.http?.authMethodsPerActor ?? {}
```
Comment (thetutlage) [inline packages/medusa/src/api/auth/[actor_type]/[auth_provider]/callback/route.ts:18]: Okay. Looks like we can update the `ConfigModule` type to make it optional.

Comment (sradevski) [reply packages/medusa/src/api/auth/[actor_type]/[auth_provider]/callback/route.ts:18]: @thetutlage yeah it should be optional

packages/medusa/src/api/auth/[actor_type]/[auth_provider]/callback/route.ts
Comment (adrien2p) [reply packages/medusa/src/api/auth/[actor_type]/[auth_provider]/callback/route.ts:21]: ok got it then maybe we can reduce the number of if and do
```ts
if (authMethodsPerActor[actor_type] && !authMethodsPerActor[actor_type].includes(auth_provider)) {
```

wdyt?

adrien2p (approved):

thetutlage (approved):

packages/medusa/src/api/auth/[actor_type]/[auth_provider]/callback/route.ts
Comment (sradevski) [reply packages/medusa/src/api/auth/[actor_type]/[auth_provider]/callback/route.ts:21]: @adrien2p I found this slightly more readable as it is not so long, but it doesn't really matter IMO

Comment (adrien2p) [reply packages/medusa/src/api/auth/[actor_type]/[auth_provider]/callback/route.ts:21]: yeah it doesnt really, But I know that usually you are trying to prevent those from happening elsewhere so I thoughtI would mention it, but it doesnt change the behaviour so fine to keep it

PR #7517, opened 2024-05-29
Add defineConfig helper

thetutlage:
Adding `defineConfig` helper that could be used to define configuration within the `medusa-config.js` file.

- It will provide better auto-complete experience and type-safety.
- It will reduce the visual debt from the `medusa-config.js` file.

Ideally the `defineConfig` can be called without any arguments and it will work. However, we have discussed to redefine certain config sections in the starter-kit.


thetutlage:
Needs https://github.com/medusajs/medusa/pull/7515#pullrequestreview-2084650913 for the type errors to solve

packages/core/utils/src/common/define-config.ts
```diff
@@ -0,0 +1,127 @@
...
+/**
+ * The "defineConfig" helper can be used to define the configuration
+ * of a medusa application.
+ *
+ * The helper under the hood merges your config with a set of defaults to
+ * make an application work seamlessly, but still provide you the ability
+ * to override configuration as needed.
+ */
+export function defineConfig(config: Partial<ConfigModule> = {}): ConfigModule {
+  const { http, ...restOrProjectConfig } = config.projectConfig || {}
```
adrien2p:
```suggestion
  const { http, ...restOfProjectConfig } = config.projectConfig || {}
```

typeo I believe?

packages/core/utils/src/common/define-config.ts
```diff
@@ -0,0 +1,127 @@
...
+    backendUrl: DEFAULT_ADMIN_URL,
+    ...config.admin,
+  }
+
+  /**
+   * The defaults to use for the feature flags config. They are shallow merged
+   * with the user defined config
+   */
+  const featureFlags: ConfigModule["featureFlags"] = {
+    medusa_v2: true,
```
adrien2p: todo: I believe we can remove that now

packages/core/utils/src/common/define-config.ts
```diff
@@ -0,0 +1,127 @@
+import { ConfigModule } from "@medusajs/types"
+import { Modules } from "@medusajs/modules-sdk"
+
+const DEFAULT_SECRET = "supersecret"
+const DEFAULT_ADMIN_URL = "http://localhost:9000"
+const DEFAULT_STORE_CORS = "http://localhost:8000"
+const DEFAULT_DATABASE_URL = "postgres://localhost/medusa-starter-default"
+const DEFAULT_ADMIN_CORS = "http://localhost:7000,http://localhost:7001"
```
riqwan: Adding the default vite port:

```suggestion
const DEFAULT_ADMIN_CORS = "http://localhost:7000,http://localhost:7001,http://localhost:5173"
```

packages/core/utils/src/common/define-config.ts
```diff
@@ -0,0 +1,127 @@
...
+              config: {
+                manual: {},
+              },
+            },
+          },
+        ],
+      },
+    },
+    ...config.modules,
+  }
```
adrien2p, line 118:
```suggestion
   const modules: ConfigModule["modules"] = {
    ...Object.keys(MODULE_PACKAGE_NAMES).reduce((acc, key) => {
      acc[key] = true;
      return acc;
    }, {}),
    [Modules.USER]: {
      resolve: "@medusajs/user",
      options: {
        jwt_secret: process.env.JWT_SECRET ?? DEFAULT_SECRET,
      },
    },
    [Modules.FILE]: {
      resolve: "@medusajs/file",
      options: {
        providers: [
          {
            resolve: "@medusajs/file-local-next",
            options: {
              config: {
                local: {},
              },
            },
          },
        ],
      },
    },
    [Modules.FULFILLMENT]: {
      resolve: "@medusajs/fulfillment",
      options: {
        providers: [
          {
            resolve: "@medusajs/fulfillment-manual",
            options: {
              config: {
                manual: {},
              },
            },
          },
        ],
      },
    },
    ...config.modules,
  }
```

Wdyt? Take it or leave it :D, I am just suggesting that such that if we add a new module then it will be present in most cases. But otherwise I am fine with the current one

adrien2p (approved):

olivermrbl (approved):
Love it @thetutlage!

packages/core/utils/src/common/define-config.ts
riqwan: I don't think we use this anymore.

riqwan (approved):
🥇

packages/core/utils/src/common/define-config.ts
thetutlage: Hahah. True

packages/core/utils/src/common/define-config.ts
thetutlage: Cool

thetutlage: Sure

packages/core/utils/src/common/define-config.ts
thetutlage: Yeah, in one instance I like it. But in another one, I think let it be a explicit of explicit module names. Not sure, its hard to decide 😬

thetutlage:
Build failing because of circular dependency between `@medusajs/modules-sdk` and `@medusajs/utils`.

sradevski (approved):

PR #7520, opened 2024-05-29
chore: Allow modules to discover resources from ts/js and local directories

adrien2p:
**What**
Update the module resources discovery to account to the source directory to look into.

example:
```ts
user: { resolve: 'user' }
```
The above config will load the resources from the resolved path looking from the node modules
```ts
user: { resolve: './modules/user' }
```
The above config will load the resources from the local directory under dist if not run with ts node and under src otherwise


packages/core/modules-sdk/src/loaders/utils/load-internal.ts
```diff
@@ -208,17 +208,24 @@ export async function loadModuleMigrations(
 async function importAllFromDir(path: string) {
   let filesToLoad: string[] = []
 
+  const excludedExtensionsRegexp = /(\.ts\.map|\.js\.map|\.d\.ts)/
```
carlos-r-l-rodrigues: nit:
```suggestion
  const excludedExtensionsRegexp = /(\.ts\.map|\.js\.map|\.d\.ts)$/
```

carlos-r-l-rodrigues (approved):

thetutlage:
Okay, a bit confused (maybe I am missing the needed knowledge for this part).

Are we trying to load resources like `models` from the starter kit?

packages/core/modules-sdk/src/loaders/utils/load-internal.ts
adrien2p: Ends with, ill commit it for extra safety🤘

adrien2p:
@thetutlage we are from the modules, so if in the starter you have a custom local modules that you specify on your config and resolve to it, we will load its resources

thetutlage:
Okay, and the path of that module should not have `./dist` or `./src` inside it. Because, we want to pick `./src` or `./dist` via auto-detection (same as plugins).

packages/core/modules-sdk/src/loaders/utils/load-internal.ts
```diff
@@ -208,17 +208,24 @@ export async function loadModuleMigrations(
...
-        const filePath = join(path, file)
-        const stats = statSync(filePath)
-
-        if (stats.isDirectory()) {
-          // TODO: should we handle that? dont think so but I put that here for discussion
-        } else if (stats.isFile()) {
-          filesToLoad.push(filePath)
-        }
+      if (
+        file.startsWith("index.") ||
```
sradevski, line 216: We could also do `index.js || index.ts`, since that's what we are doing below as well, no need for Regex IMO

sradevski (approved):

packages/core/modules-sdk/src/loaders/utils/load-internal.ts
adrien2p: we dont want the index.ts.map nor the index.js.map etc and we dont want any files not index but ending with those extensions

adrien2p: for example we don't want to import my-service.ts.map

thetutlage (approved):

packages/core/modules-sdk/src/loaders/utils/load-internal.ts
sradevski: @adrien2p ah sorry I wasn't clear, I meant we can say `file === "index.js" || file === "index.ts"`, since later down the code we replace the index.js and index.ts explicitly. Anyway it's not a big deal, I just thought maybe we can match the two

adrien2p: we are replacing it only for the module path, here it is the resources :)

olivermrbl (approved):
Ncie!

adrien2p:
@shahednasser sorry for the delay, you can test it 💪

PR #7521, opened 2024-05-29
Revamp auth module to support multiple providers linked to a single auth identity

sradevski:
In order to migrate you can either just create a new user with the CLI, or:
- Before running the migrations, copy `provider` , `entity_id` and `provider_metadata` from `auth_identity` table
- Run the migration
- Create a row in `provider_identity` with the previously copied data, where the `auth_identity_id` will be the row you copied from.

### DIFF
--- a/packages/modules/auth/src/services/auth-module.ts
+++ b/packages/modules/auth/src/services/auth-module.ts
@@ -139,7 +139,7 @@ export default class AuthModuleService<
       return await this.authProviderService_.authenticate(
         provider,
         authenticationData,
-        this.getAuthIdentityProviderService()
+        this.getAuthIdentityProviderService(provider)
       )
     } catch (error) {
       return { success: false, error: error.message }
@@ -154,16 +154,18 @@ export default class AuthModuleService<
       return await this.authProviderService_.validateCallback(
         provider,
         authenticationData,
-        this.getAuthIdentityProviderService()
+        this.getAuthIdentityProviderService(provider)
       )
     } catch (error) {
       return { success: false, error: error.message }
     }
   }
 
-  getAuthIdentityProviderService(): AuthIdentityProviderService {
+  getAuthIdentityProviderService(
+    provider: string
+  ): AuthIdentityProviderService {
     return {
-      retrieve: async ({ entity_id, provider }) => {
+      retrieve: async ({ entity_id }) => {
         const authIdentities = await this.authIdentityService_.list(
           {
             provider_identities: {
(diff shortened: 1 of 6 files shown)


srindom (commented):
Two thoughts:

**First**
wonder if it would be better to stick with the current API for creation. I.e.:

```ts
authService.create({ provider, entity_id, provider_metadata })
```

or maybe:

```ts
authService.create({ provider_identity: { provider, entity_id, provider_metadata } })
```

And then have a dedicated mechanism for linking additional providers. E.g.: 
```ts
authService.linkProviderWithIdentity(identityId, { provider, entity_id, provider_metadata })
```

My thinking is that there won't be multiple providers at creation time so supporting that in the API might be confusing? 

It would mean the CreateDTO drifts a bit from the underlying data model, but might be worth it. Not a strong opinion but curious to hear what you think?

**Second**
These changes highlights the need for changing the mechanism that creates the AuthIdentity. Currently, the actual creation seems to happen in the auth provider. But since the authentication now can lead to a "linking" instead of a creation we will need to change that. 

This also poses the question of whether the auth providers should be allowed to call AuthModuleService directly. A more sensible approach might be to give the AuthProviders a strict API they need to adhere to (e.g., must return { entity_id, provider_metadata }) and then the AuthModule will decide what to do with that information.

This is outside the scope of this PR, but wanted to address it so we are aware.

sradevski:
@srindom thanks for the comments!

1. I kept it barebones for now, but in principle I agree with you. Let me see what I can do and how it will look like.
2. With my changes the auth providers don't call the auth module directly, they are given 2 functions that they invoke (that just happen to do just retrieve and create now, but can do more when needed). Now, those two functions will need to be reworked a bit when we add support for linking, but the main goal of this PR is for us to do the breaking changes to the model as early as possible, and then the only changes needed would be at the API level.

Let me know if 2. sounds good to you, and I'll have a look into 1.

srindom:
2 sounds reasonable 👍

sradevski:
@srindom I revisited the interfaces, and this is what I think is best:
1. The Auth module should create the provider identities in the raw form (i.e. by passing provider_identities). Since modules can be used in many ways, I think it's better to allow users to eg. create 2 identities at the same time if they wanted to. Note that the workflows and HTTP endpoints won't allow this, so for most people it won't matter anyway.

2. I limited the parameters that auth providers pass to make it simpler. They only pass the data they know about (see the last commit).

Let me know if you have any other comments 👍

packages/modules/auth/src/services/auth-module.ts
```diff
@@ -194,8 +196,26 @@ export default class AuthModuleService<
...
+              provider_metadata: data.provider_metadata,
+              user_metadata: data.user_metadata,
+              provider,
+            },
+          ],
+        }
+
+        const createdAuthIdentity = await this.authIdentityService_.create(
+          normalizedRequest
+        )
```
srindom, line 218: Wouldn't this have to be a conditional - e.g., if the identity already exists we shouldn't create a new one but just link this provider to an existing identity?

packages/modules/auth/src/services/auth-module.ts
```diff
@@ -194,8 +196,26 @@ export default class AuthModuleService<
           authIdentities[0]
         )
       },
-      create: async (data: AuthTypes.CreateAuthIdentityDTO) => {
-        const createdAuthIdentity = await this.authIdentityService_.create(data)
+
+      create: async (data: {
```
srindom, line 200: thought: maybe this should be renamed to `createOrLink` to make it clear what is it will do?

srindom (approved):
LGTM - we still need to figure out where linking happens. Can be here or in a separate PR 👍

packages/modules/auth/src/services/auth-module.ts
sradevski: @srindom when we add linking yes, but currently all we can do is create a new auth identity

packages/modules/auth/src/services/auth-module.ts
sradevski: @srindom I already added a comment to this interface, but yes, the interface will need to change once linking support is added

sradevski:
@srindom I'd leave linking for a separate PR (I can scope it out today), that will be a purely additive change (except for providers, but no one is building them just yet)

PR #7523, opened 2024-05-29
feat: Remove returning token from customer and user endpoints

sradevski:


packages/medusa/src/api/admin/users/route.ts
```diff
@@ -57,32 +56,13 @@ export const POST = async (
 
   const { result } = await createUserAccountWorkflow(req.scope).run(input)
 
-  const { http } = req.scope.resolve(
```
sradevski, line 60: We should actually remove this entire endpoint, but I'll do that in a separate PR

olivermrbl (approved):

PR #7524, opened 2024-05-29
chore: query inventory items by location id

riqwan:
what:

- allows querying inventory items by the correct location_id


kasperkristensen (approved):
Just tested and seems to fix the issue. Only thing I noticed is that the validator did not fail when I haden't updated my request to no longer pass `location_id` directly. Might be because `.strict()` is missing?

riqwan:
@kasperkristensen added the strict in there now

PR #7526, opened 2024-05-29
chore(modules-sdk): Small fix to error wording

shahednasser:
Small fix to the wording when a module can't be loaded.


packages/core/modules-sdk/src/loaders/module-provider-loader.ts
```diff
@@ -43,7 +43,7 @@ export async function loadModuleProvider(
     }
   } catch (error) {
     throw new Error(
-      `Unable to find plugin ${pluginName} -- perhaps you need to install its package?`
+      `Unable to find module ${pluginName} -- perhaps you need to install its package?`
```
sradevski: If you feel like it, maybe let's rename the variable as well :)

sradevski (approved):

packages/core/modules-sdk/src/loaders/module-provider-loader.ts
shahednasser: good point, I just used the github editor to do this one so didn't wanna mess up anything 😄

sradevski (approved):

PR #7530, opened 2024-05-29
feat(core-flows,medusa,types): creating inventory items attaches it to variant

riqwan:
what:

- updates workflow to attach inventory items to variant
- fixes skipped tests

### DIFF
--- a/packages/core/core-flows/src/inventory/steps/create-inventory-items.ts
+++ b/packages/core/core-flows/src/inventory/steps/create-inventory-items.ts
@@ -13,23 +13,23 @@ export const createInventoryItemsStep = createStep(
     const inventoryService: IInventoryServiceNext = container.resolve(
       ModuleRegistrationName.INVENTORY
     )
-    const variantItemDataMap = new Map<
+    const variantItemDataMap: Record<
       string,
       InventoryNext.CreateInventoryItemInput[]
-    >()
-    const variantItemsMap = new Map<string, InventoryNext.InventoryItemDTO[]>()
+    > = {}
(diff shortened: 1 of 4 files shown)


packages/core/core-flows/src/inventory/steps/create-inventory-items.ts
```diff
@@ -1,31 +1,50 @@
...
+
+    for (const { variant_id, ...inventoryData } of data) {
+      const array = variantItemDataMap.get(variant_id) || []
+
+      array.push(inventoryData)
+      variantItemDataMap.set(variant_id, array)
+    }
+
+    for (const [variantId, itemsData] of variantItemDataMap.entries()) {
+      const createdItems = await inventoryService.create(itemsData)
```
carlos-r-l-rodrigues, line 30: don't we have this create in bulk on the module?
that can be a problem if you have multiple and one of them fail, the compensation will have no id.
so, worst case scenario would be bring that loop to the module, so you can reuse the same db transaction

riqwan: Hmm. This is creating in bulk (partially). The problem here is that we still need to keep the map of the variant_id to created items. Since there aren't any unique attributes in inventory module, its hard to do it in a single bulk call.

riqwan: I'll change this to be a purely create item workflow and create a different workflow that is to attach an items to a variant.

carlos-r-l-rodrigues: usually in cases like this we must guarantee the order we sent to be created is the same the method will return, so we can remap it

PR #7534, opened 2024-05-29
fix(dashboard) editor columns

fPolic:
**What**
- reorganize create variants grid columns
- add region prices cells
  - _note: backend only handles currency prices atm_
- automatically select default sales channel  
- add placeholders on Options fields on product create flow

---

CLOSES CORE-2190 CORE-2135


packages/admin-next/dashboard/src/routes/products/product-create/utils.ts
```diff
@@ -56,10 +56,19 @@ export const normalizeVariants = (
...
-    prices: Object.entries(variant.prices || {}).map(([key, value]: any) => ({
-      currency_code: key,
-      amount: getDbAmount(castNumber(value), key),
-    })),
+    prices: Object.entries(variant.prices || {})
+      .map(([key, value]: any) => {
+        if (key.startsWith("reg_")) {
+          // TODO: route needs to accept region prices as well
+          return undefined
+        } else {
```
srindom, line 64: @fPolic - is someone working on fixing this?

fPolic: Not ATM as far as I know

packages/admin-next/dashboard/src/routes/products/product-create/components/product-create-variants-form/product-create-variants-form.tsx
```diff
@@ -19,6 +20,8 @@ type ProductCreateVariantsFormProps = {
 export const ProductCreateVariantsForm = ({
   form,
 }: ProductCreateVariantsFormProps) => {
+  const { regions } = useRegions()
```
kasperkristensen: This would causes issues if the store has more regions then what we return by default. Properly very unlikely but could happen if people have created a region for all countries they sell to. 

I think we should set the limit to something like 999, and only select the fields that we require for the editor to work, so id, title, and currency_code.

packages/admin-next/dashboard/src/routes/products/product-create/utils.ts
```diff
@@ -70,19 +79,19 @@ export const decorateVariantsWithDefaultValues = (
...
-    material: variant.material || "",
-    origin_country: variant.origin_country || "",
+    // mid_code: variant.mid_code || "",
+    // hs_code: variant.hs_code || "",
+    // width: variant.width || "",
+    // height: variant.height || "",
+    // weight: variant.weight || "",
+    // length: variant.length || "",
+    // material: variant.material || "",
+    // origin_country: variant.origin_country || "",
```
kasperkristensen: question: should we not just delete all of these now?

packages/admin-next/dashboard/src/routes/products/product-detail/product-detail.tsx
```diff
@@ -45,7 +45,7 @@ export const ProductDetail = () => {
         )
       })}
       <div className="flex flex-col gap-x-4 xl:flex-row xl:items-start">
```
kasperkristensen: add gap-y-3 here as well

kasperkristensen: and remove `mt-2` and add `gap-y-3` on line 64

packages/admin-next/dashboard/src/routes/products/product-create/components/product-create-variants-form/product-create-variants-form.tsx
fPolic: I intended to do that but I passed 9999 limit to the store hook below 🤦

packages/admin-next/dashboard/src/routes/products/product-create/utils.ts
fPolic: yes, I was just waiting for confirmation on linear but I'll remove this

olivermrbl:
Should we merge this?

PR #7535, opened 2024-05-29
feat: Improve zod error messages

sradevski:


olivermrbl:
This is really great, how much can we customize the error messages? Think the `at` reads a little bit weird

sradevski:
@olivermrbl I was just checking that, I think eventually we might want to roll out our own formatter, esp. when it comes to localization, but I'll try and see if we can at least make this a bit more user-friendly

sradevski:
@olivermrbl ignore what you saw, check it again 😄 At the end it seems just implementing it manually is better.

packages/medusa/src/api/utils/__tests__/zod-helper.spec.ts
```diff
@@ -0,0 +1,134 @@
+import { z } from "zod"
```
sradevski, line 1: I renamed validate-body to zod-helper tests, I didn't really delete anything 😄

packages/medusa/src/api/utils/zod-helper.ts
```diff
@@ -0,0 +1,129 @@
...
+const formatUnionError = (issue: ZodInvalidUnionIssue) => {
+  const issues = issue.unionErrors.flatMap((e) => e.issues)
+  return (
+    formatInvalidType(issues) || formatRequiredField(issues) || issue.message
+  )
+}
+
+const formatError = (err: ZodError) => {
+  const issueMessages = err.issues.slice(0, 3).map((issue) => {
+    console.log(issue)
```
olivermrbl:
```suggestion
```

olivermrbl (approved):
Very nice @sradevski !

PR #7536, opened 2024-05-29
chore(payment): provider call outside db transaction

carlos-r-l-rodrigues:


adrien2p (approved):
Sounds good, i believe we had the same approach for fulfillment 🤘

sradevski (approved):
LGTM, but I am a bit confused on the difference between InjectManager and InjectTransactionManager.

IIRC from early on when Adrien was explaining to me, InjectManager also runs the code in the function in a transaction. Is the difference that with InjectManager the transaction gets commited when the function ends, and InjectTransactionManager it doesn't until some other time (when is that actually?)

adrien2p:
> LGTM, but I am a bit confused on the difference between InjectManager and InjectTransactionManager.
> 
> IIRC from early on when Adrien was explaining to me, InjectManager also runs the code in the function in a transaction. Is the difference that with InjectManager the transaction gets commited when the function ends, and InjectTransactionManager it doesn't until some other time (when is that actually?)

@sradevski , Inject manager only inject the entity manager, without starting a transaction. the InjectTransactionManager wrap the function in a transaction which gets commited at the end of the function block.

So, so far with those function communicating with external providers we manage them a bit differently to not start a transaction which lock the tuple while we communicate with an external provider which can take some time and create long running transactions

sradevski:
@adrien2p ok clear, thanks! (btw, you pinged some random person ahaha)

adrien2p:
> @adrien2p ok clear, thanks! (btw, you pinged some random person ahaha)

Shish, thanks for letting me know :D

PR #7537, opened 2024-05-29
chore(promotion): big number calc

carlos-r-l-rodrigues:


packages/modules/promotion/src/services/promotion-module.ts
```diff
@@ -202,12 +206,12 @@ export default class PromotionModuleService<
 
         const campaignBudgetData = {
           id: campaignBudget.id,
-          used: (campaignBudget.used ?? 0) + 1,
+          used: MathBN.add(campaignBudget.used ?? 0, 1),
```
carlos-r-l-rodrigues, line 209: @riqwan can you confirm if this `+ 1` here is correct?

riqwan: Yes, this is correct

riqwan (approved):
💪🏻

sradevski (approved):

PR #7541, opened 2024-05-30
feat(types,medusa): add inventory quantity to products endpoint

riqwan:
what:

- when inventory_quantity is requested through the API, we calculate the inventory based on sales channels + stock locations and return the total available inventory.

A variant can have multiple inventory items. As an example:

Table: (variant)
  - 4 (required_quantity via link) x legs (inventory item)
  - 2 x table top

Only if all individual inventory items of a variant are available, do we mark the variant as available as a single unit. 

RESOLVES CORE-2187

### DIFF
--- a/integration-tests/modules/__tests__/product/store/index.spec.ts
+++ b/integration-tests/modules/__tests__/product/store/index.spec.ts
@@ -451,6 +451,7 @@ medusaIntegrationTestRunner({
               ContainerRegistrationKeys.REMOTE_LINK
             )
 
+            // TODO: Missing API endpoint. Remove this when its available
             await remoteLink.create([
               {
                 [Modules.SALES_CHANNEL]: { sales_channel_id: salesChannel1.id },
@@ -468,6 +469,7 @@ medusaIntegrationTestRunner({
               ContainerRegistrationKeys.REMOTE_LINK
             )
 
+            // TODO: Missing API endpoint. Remove this when its available
             await remoteLink.create([
               {
                 [Modules.PRODUCT]: { variant_id: variant.id },
@@ -514,6 +516,7 @@ medusaIntegrationTestRunner({
               ContainerRegistrationKeys.REMOTE_LINK
             )
 
+            // TODO: Missing API endpoint. Remove this when its available
             await remoteLink.create([
               {
                 [Modules.PRODUCT]: { variant_id: variant.id },


packages/medusa/src/api/store/products/helpers.ts
```diff
@@ -8,3 +14,96 @@ export const refetchProduct = async (
...
 }
+
+type VariantInventoryType = {
+  variant_id: string
+  variant: { manage_inventory: boolean }
+  required_quantity: number
+  inventory: InventoryItemDTO
+}
+
+export const wrapVariantsWithInventoryQuantity = async (
```
riqwan, line 25: @carlos-r-l-rodrigues @olivermrbl - I didn't need to query the module directly, so I moved it into a helper. Does this still need to be a workflow?

olivermrbl: This is nice @riqwan!

I am fine keeping it like this for the time being

packages/medusa/src/api/store/products/middlewares.ts
```diff
@@ -23,6 +26,35 @@ export const storeProductRoutesMiddlewares: MiddlewareRoute[] = [
...
+          const withInventoryQuantity = req.remoteQueryConfig.fields.some(
+            (field) => field.includes("variants.inventory_quantity")
+          )
+
+          req.remoteQueryConfig.fields = req.remoteQueryConfig.fields.filter(
+            (field) => !field.includes("variants.inventory_quantity")
+          )
+
+          return withInventoryQuantity
+        },
```
olivermrbl: suggestion(non-blocking): We could kill this dedicated context property, move the logic here to `stock_location_id`, and check the `remoteQueryConfig.fields` in the route? Just to have fewer magic props

packages/medusa/src/api/utils/middlewares/common/set-context.ts
```diff
@@ -0,0 +1,22 @@
+import { NextFunction } from "express"
+import { MedusaRequest } from "../../../../types/routing"
+
+export function setContext(context: Record<string, any>) {
```
olivermrbl, line 4: nice!

olivermrbl (approved):
LGTM, very nice!

adrien2p (approved):
LGTM, Nothing more to say than the already provided comment, maybe one thing, I would extract the context function from the middlewares to some helper to not pollute the middleware file but otherwise 👍

packages/medusa/src/api/store/products/helpers.ts
```diff
@@ -8,3 +14,96 @@ export const refetchProduct = async (
...
+
+    for (const link of links) {
+      const requiredQuantity = link.required_quantity
+      const availableQuantity = (link.inventory.location_levels || []).reduce(
+        (sum, level) => sum + level.available_quantity || 0,
+        0
+      )
+
+      // This will give us the maximum deliverable quantities for each inventory item
+      const maxInventoryQuantity = Math.round(
```
carlos-r-l-rodrigues: `Math.floor`?

carlos-r-l-rodrigues (approved):

packages/medusa/src/api/store/products/helpers.ts
riqwan: Good catch! :D 
```suggestion
      const maxInventoryQuantity = Math.floor(
```

integration-tests/modules/__tests__/product/store/index.spec.ts
```diff
@@ -356,6 +364,212 @@ medusaIntegrationTestRunner({
...
+              },
+            ])
+          })
+
+          it("should list all inventory items for a variant", async () => {
+            const remoteLink = appContainer.resolve(
+              ContainerRegistrationKeys.REMOTE_LINK
+            )
+
+            await remoteLink.create([
```
sradevski, line 473: Shouldn't this be supported through the product HTTP API? If we are doing this simply because it isn't yet, let's add a TODO here

packages/medusa/src/api/store/products/middlewares.ts
sradevski: Yup I agree with this approach as well

packages/medusa/src/api/store/products/helpers.ts
sradevski, line 25: The function feels a bit overwhelming, so maybe we can reduce complexity by splitting the code up a bit? Not a blocker for sure, just something I felt while reading the code.

sradevski (approved):
LGTM overall!

packages/medusa/src/api/store/products/helpers.ts
riqwan: Any recommendations? Most of the function is just gathering data and a few lines of code to calculate

packages/medusa/src/api/store/products/helpers.ts
```diff
@@ -8,3 +14,117 @@ export const refetchProduct = async (
...
+  links.forEach((link) => {
+    const array: VariantInventoryType[] =
+      variantInventoriesMap.get(link.variant_id) || []
+
+    array.push(link)
+
+    variantInventoriesMap.set(link.variant_id, array)
+  })
+
+  for (const variant of variants || []) {
```
sradevski, line 80: @riqwan maybe from here until the end of the loop can be extracted into a function, WDYT? Anyway I leave it up to you, it isn't a big deal TBH

PR #7548, opened 2024-05-30
chore: use product variant workflow in product create workflow

riqwan:
what:

- reuses the product variant create workflow inside product create workflow


adrien2p (approved):
Looks good, does that mean we dont use the product create cascade, I think it is fine since we are doing plenty of other things that are hard to maintain/compute

riqwan:
@adrien2p yeah we lose that aspect, but this one is more maintainable as we add more stuff to it like inventory creation as a part of variants.

sradevski (approved):
Nice 👍

PR #7550, opened 2024-05-30
chore(utils): added TSDocs to AbstractAuthModuleProvider

shahednasser:
Added TSDocs to AbstractAuthModuleProvider. These are used to generate the reference on how to create an auth provider module.


packages/core/utils/src/auth/abstract-auth-provider.ts
```diff
@@ -26,11 +106,195 @@ export abstract class AbstractAuthModuleProvider implements IAuthProvider {
       config.displayName
   }
 
+  /**
+   * This method authenticates the user.
+   * 
+   * If the authentication requires an action to perform with a third-party service, such as login with
```
sradevski: "The authentication happens either by directly authenticating or returning a redirect URL to continue the authentication with a third party provider."

Or something along those lines, to make it a bit clearer that there are basically 2 ways `authenticate` behaves

sradevski (approved):

sradevski (approved):

PR #7561, opened 2024-05-31
Loosely typed container

thetutlage:
Alright. So we went ahead and added some types to the container. It means using `container.resolve` method will show a list of auto-suggestions and also define the return type for the known container bindings. 

As a result of this, a lot of code within the medusa repo that was typed as `any` has received strict types and needed refactoring.

Also, some of the refactoring might have taken more time. So I have decided to annotate that code with `@ts-expect-error` vs fixing it. But if you have any feedback, please let me know and we can fix that code.

### DIFF
--- a/packages/core/core-flows/src/defaults/steps/create-default-store.ts
+++ b/packages/core/core-flows/src/defaults/steps/create-default-store.ts
@@ -20,10 +20,6 @@ export const createDefaultStoreStep = createStep(
      * @todo
      * Seems like we are missing an integration test when the
      * following conditional as true.
-     *
-     * As per types, the result from "createStoresWorkflow.run" was
-     * an array of "StoreDTO", whereas we needed just "StoreDTO" at
-     * top-level.
      */
     if (!store) {
       const stores = await createStoresWorkflow(container).run({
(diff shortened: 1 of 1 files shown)


adrien2p (approved):
LGTM, as we already went through it 😄

thetutlage:
@adrien2p Thanks a ton dude for spending so much time on this with me 😇

packages/core/modules-sdk/src/medusa-app.ts
```diff
@@ -36,6 +39,16 @@ import { cleanGraphQLSchema } from "./utils"
 
 const LinkModulePackage = MODULE_PACKAGE_NAMES[Modules.LINK]
 
+declare module "@medusajs/types" {
+  export interface ModuleImplementations {
```
riqwan, line 43: very cool! I'm guessing you could do something similar in the starter and it would work for any custom module? 

q: does this override or append to the ModuleImplementations type?

riqwan (approved):
very cool guys! @adrien2p @thetutlage 🎉

packages/core/modules-sdk/src/medusa-app.ts
adrien2p: This will merge the interfaces (declaration merging, module augmentation), this is the approach I ve also provided to the community in medusa v1 for extending interfaces for customization if you remember :D when for example they want the product type to include the newly added props accross the code base

thetutlage: It merges. So what you add in the starter will added on top and the existing one's will be replaced (only if the types are compatible with each other)

thetutlage:
Should we merge this? Snapshots are not working for some reason

packages/core/modules-sdk/src/medusa-app.ts
riqwan: noice!

PR #7562, opened 2024-05-31
chore: preserve ordering of variant rank when creating variants

riqwan:
what:

when passing the inputs over to create variants workflow from create product workflow, we lose the ability to order correctly, so we pass the ranking information on from the product workflow.

### DIFF
--- a/integration-tests/http/__tests__/product/admin/product.spec.ts
+++ b/integration-tests/http/__tests__/product/admin/product.spec.ts
@@ -1,8 +1,8 @@
-import {
-  createAdminUser,
-  adminHeaders,
-} from "../../../../helpers/create-admin-user"
 import { medusaIntegrationTestRunner } from "medusa-test-utils"
+import {
+  adminHeaders,
+  createAdminUser,
+} from "../../../../helpers/create-admin-user"
 
 jest.setTimeout(50000)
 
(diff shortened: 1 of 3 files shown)


packages/core/core-flows/src/product/workflows/create-products.ts
```diff
@@ -58,13 +58,17 @@ export const createProductsWorkflow = createWorkflow(
 
       data.createdProducts.forEach((product, i) => {
         const inputProduct = data.input.products[i]
+        let rank = 0
```
sradevski, line 61: I would rather move this responsibility to the module as it can be nicely abstracted away

integration-tests/http/__tests__/product/admin/product.spec.ts
```diff
@@ -1262,14 +1262,16 @@ medusaIntegrationTestRunner({
           expect(response.data.product).toEqual(
             expect.objectContaining({
               title: "Test product - 1",
-              variants: [
+              variants: expect.arrayContaining([
```
sradevski, line 1265: If we use `arrayContaining` then the result can come in either order, and we want to check here that they do come in the specified order right?

riqwan:
Closing this in favor of moving it to the module 💯

PR #7563, opened 2024-05-31
fix(utils): support omitting modules in config

srindom:
**What**

- With the new `defineConfig` we automatically enable all modules. However, users may not want to use all modules. 
- This PR modifies `defineConfig`'s behavior to not include module configs for modules that are marked as false.

## Usage

```
// medusa-config.ts

export default defineConfig({
  ...,
  modules: {
    // Location and inventory modules will not be required or loaded
    [Modules.STOCK_LOCATION]: false,
    [Modules.INVENTORY]: false,
  },
})

```

~~Note: I also considered whether disabling a module with `false` should be supported in the loading mechanism for MedusaApp. This is not necessary, so can be removed if we don't think that makes sense.~~

^ have reverted this for now, don't think it's necessary.


packages/core/modules-sdk/src/medusa-app.ts
srindom, line 1: note: the change in this file is not necessary, but wanted to see what people think of being able to disable modules here too?

packages/core/modules-sdk/src/medusa-app.ts
```diff
@@ -85,6 +85,16 @@ export async function loadModules(
       let declaration: any = {}
       let definition: Partial<ModuleDefinition> | undefined = undefined
 
+      if (!isObject(mod) && typeof mod !== "boolean") {
+        throw new Error(
+          `Invalid module configuration for ${moduleName}. Expected boolean or a module configuration object, got ${typeof mod}. Check your medusa-config file.`
+        )
+      }
```
adrien2p: suggestion: I dont think we should mention the medusa-config file here as the module sdk can be used for standalone modules but also should not be aware of where the config is coming from wdyt?

srindom: agree - good point

packages/core/utils/src/common/define-config.ts
```diff
@@ -117,6 +117,13 @@ export function defineConfig(config: Partial<ConfigModule> = {}): ConfigModule {
     ...config.modules,
   }
 
+  // Remove any modules set to false
+  Object.keys(modules).forEach((key) => {
+    if (modules[key] === false) {
+      delete modules[key]
+    }
+  })
```
adrien2p, line 125: Q: should we be able to disabled modules that have a default package as well?

adrien2p (approved):
LGTM

packages/core/utils/src/common/__tests__/define-config.spec.ts
```diff
@@ -256,4 +257,93 @@ describe("defineConfig", function () {
       }
     `)
   })
+
+  it("should not include disabled modules", function () {
```
adrien2p, line 261: suggestion: should we add a test where we disable a module and therefore it should not be resolvable from the container. wdyt? it could go there integration-tests/modules/__tests__/modules/load-standalone.ts

riqwan (approved):
💯

thetutlage:
During my initial conversation with Oli, he mentioned there won't be any use-cases to disable base set of modules, so I did not consider adding the support for it.

packages/core/utils/src/common/define-config.ts
srindom: yes - e.g., in platform we don't want stock/inventory/etc. 

it will break functionality in the core API because those modules aren't there, but if you want to use medusa partially you should be able to config your way to that.

olivermrbl:
> During my initial conversation with Oli, he mentioned there won't be any use-cases to disable base set of modules, so I did not consider adding the support for it.

I believe what I said was in the case where we have module "bundles"/"templates", I don't see a use case for a disabling function. And specifically, my idea was to not offer a `without` function it in the "bundles" API, when (if) we choose to introduce that for the sake of having a simple API. At least that is what I meant.

E.g. I imagine in a world where we have bundles, you'll need something like:
- A commerce bundle
- An events bundle
- A cache bundle
- A workflows bundle

And all modules within a bundle would likely be strictly required otherwise you would just install the modules yourself.

That said, I see why the two cases might have been mixed together as they are very closely related. Tbh. I just didn't think of this disabling mechanism when we chose to add the default modules without supporting bundles.

packages/core/utils/src/common/define-config.ts
adrien2p: then this wont be enough, you can take a look at the mergeDefaults maybe, but there we get the modules with a default package and auto assign them to the result

sradevski (approved):

adrien2p (changes requested):
As mention as per seb feedback, this wont be enough, the mergeDefaults needs some update

packages/core/utils/src/common/define-config.ts
```diff
@@ -117,6 +117,13 @@ export function defineConfig(config: Partial<ConfigModule> = {}): ConfigModule {
     ...config.modules,
   }
 
+  // Remove any modules set to false
```
sradevski, line 120: Although I approved, I think it might be good to already look into adding these defaults in the form of a `template` rather than having them hard-coded in `define-config`.

srindom:
> @adrien2p - As mention as per seb feedback, this wont be enough, the mergeDefaults needs some update

Not sure I follow?

adrien2p:
> > @adrien2p - As mention as per seb feedback, this wont be enough, the mergeDefaults needs some update
> 
> Not sure I follow?

@srindom , based on what you are saying, you should be able to disable a module even though there is a default package, though, we are going through this function `mergeDefaultModules` which will assign those modules to the resulted modules if there is a default package configured. Also, I believe it would be easier to test that it works if you add a test in the place I ve mentioned in one of the comment. That way you can ensure it is behaving as expected

adrien2p:
@srindom does that make sense or do you want me to look at it?

thetutlage:
@olivermrbl 

Cool, makes sense. I bring it up, so that my understanding of the approach is on the same page as the rest 😄

srindom:
@adrien2p - I see. Think I was a bit confused about what `defaultPackage` referred to. Thought it would be something like the `defaultPackage` for cart module is `@medusajs/cart`, but that doesn't seem to be the case. 

The changes in the PR satisfy the need I have to disable the commerce related modules that I don't need for a partial installation. The `defaultPackage` thing might need to be considered separately.

adrien2p:
> @adrien2p - I see. Think I was a bit confused about what `defaultPackage` referred to. Thought it would be something like the `defaultPackage` for cart module is `@medusajs/cart`, but that doesn't seem to be the case.
> 
> The changes in the PR satisfy the need I have to disable the commerce related modules that I don't need for a partial installation. The `defaultPackage` thing might need to be considered separately.

Ok got it, I see 👍

adrien2p (approved):

PR #7569, opened 2024-05-31
chore: Move product category integration tests

olivermrbl:
Only missing tests are `DELETE`

### DIFF
--- a/integration-tests/http/__tests__/product-category/admin/product-category.spec.ts
+++ b/integration-tests/http/__tests__/product-category/admin/product-category.spec.ts
@@ -1,5 +1,3 @@
-import { ModuleRegistrationName } from "@medusajs/modules-sdk"
-import { IProductModuleService } from "@medusajs/types"
 import { medusaIntegrationTestRunner } from "medusa-test-utils"
 import {
   adminHeaders,
@@ -10,7 +8,6 @@ jest.setTimeout(30000)
 
 medusaIntegrationTestRunner({
   testSuite: ({ dbConnection, getContainer, api }) => {
-    let appContainer
     let productCategory
     let productCategory1
     let productCategory2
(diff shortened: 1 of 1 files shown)


olivermrbl:
@sradevski, just saw you migrated a different product category test suite. I'll merge the two

integration-tests/http/__tests__/product/admin/product-category.spec.ts
```diff
@@ -0,0 +1,956 @@
...
+import {
+  adminHeaders,
+  createAdminUser,
+} from "../../../../helpers/create-admin-user"
+
+jest.setTimeout(30000)
+
+medusaIntegrationTestRunner({
+  testSuite: ({ dbConnection, getContainer, api }) => {
+    let appContainer
```
sradevski: We can remove this as it's not used naywhere outside of `beforeEach`

integration-tests/http/__tests__/product/admin/product-category.spec.ts
```diff
@@ -0,0 +1,956 @@
...
+      productModuleService = appContainer.resolve(
+        ModuleRegistrationName.PRODUCT
+      )
+
+      await createAdminUser(dbConnection, adminHeaders, appContainer)
+    })
+
+    describe("GET /admin/product-categories/:id", () => {
+      beforeEach(async () => {
+        productCategoryParent = await productModuleService.createCategory({
```
sradevski: Let's perform the creation through HTTP, so we are closer to what an end-user would do (and module changes won't break the tests if the HTTP layer remains constant)

olivermrbl:
Btw. did you notice that `rank` is returned as string even though it is stored as `numeric` in the DB?

I found that strange. Thinking some serialization is done incorrectly but I can't dig it out

sradevski (approved):
Epic stuff, just the seeding can be done through HTTP, otherwise LGTM

integration-tests/http/__tests__/product/admin/product-category.spec.ts
olivermrbl: Good push, will do

sradevski:
@olivermrbl hey I missed your comments :) Yes please merge the two, I only moved the one that was written for v2. Weird about the rank, I didn't notice that, it is strange that it only happens for categories, probably something to do with the fact that we are using a custom repository layer (I see the module's tests are also expecting a string).

Update: https://github.com/mikro-orm/mikro-orm/issues/2295 it is on purpose in this case, we can serialize it to number

olivermrbl:
@sradevski, thanks for pushing the fix. I updated the suite to only use the HTTP layer now and merged the two category suites. Should be good to merge.

sradevski (approved):
Re-approving, nice one 👍

PR #7572, opened 2024-06-01
fix: Use integer for variant and category ranks

sradevski:


sradevski:
@olivermrbl here's a fix :)

olivermrbl (approved):
ah ofc, thanks! don't know why I didn't catch this 😅

PR #7573, opened 2024-06-01
feat(order): cancel fulfillment

carlos-r-l-rodrigues:


packages/modules/order/src/utils/actions/cancel-item-fulfillment.ts
```diff
@@ -0,0 +1,74 @@
...
+          "Reference ID is required."
+        )
+      }
+
+      const existing = currentOrder.items.find((item) => item.id === refId)
+
+      if (!existing) {
+        throw new MedusaError(
+          MedusaError.Types.INVALID_DATA,
+          `Reference ID "${refId}" not found.`
```
olivermrbl: thought(non-blocking): Would it make sense to say `Item ID ${refId} not found`?

packages/medusa/src/api/admin/orders/[id]/fulfillments/[fulfillment_id]/cancel/route.ts
```diff
@@ -6,16 +7,29 @@ import {
...
 
   const variables = { id: req.params.id }
 
-  // TODO: Workflow to cancel fulfillment + adjust inventory
+  const input = {
+    ...req.validatedBody,
+    order_id: req.params.id,
+  }
+
+  const { errors } = await cancelOrderFulfillmentWorkflow(req.scope).run({
```
Comment (olivermrbl) [inline packages/medusa/src/api/admin/orders/[id]/fulfillments/[fulfillment_id]/cancel/route.ts:25]: thought(non-blocking): I guess we could have a single Fulfillment Rest API and then depending on the input, run different workflows.

What I mean is, in the case of canceling a fulfillment, we could have:
```
POST /admin/fulfillments/:id/cancel
```

And then have mutually exclusive properties for each of the entities, that can be associated with a fulfillment, i.e. `order_id`, `exchange_id`, etc.

It would eliminate a few endpoints at the cost of having to do some work in the route to call the correct workflow. I am not sure what is best, but thought I'd share the idea.

packages/medusa/src/api/admin/orders/[id]/fulfillments/[fulfillment_id]/cancel/route.ts
```diff
@@ -6,16 +7,29 @@ import {
...
+    input,
+    throwOnError: false,
+  })
+
+  if (Array.isArray(errors) && errors[0]) {
```
Comment (olivermrbl) [inline packages/medusa/src/api/admin/orders/[id]/fulfillments/[fulfillment_id]/cancel/route.ts:30]: Do we need this anymore?

packages/medusa/src/api/admin/orders/validators.ts
```diff
@@ -79,3 +79,12 @@ export const AdminOrderCreateShipment = z.object({
 export type AdminOrderCreateShipmentType = z.infer<
   typeof AdminOrderCreateShipment
 >
+
+export const AdminOrderCancelFulfillment = z.object({
```
olivermrbl, line 83: The reason I suggested the single Fulfillment Rest API is that I guess this payload will be the same across all separate cancel fulfillment APIs.

packages/core/core-flows/src/order/workflows/cancel-order-fulfillment.ts
```diff
@@ -0,0 +1,140 @@
...
+import { useRemoteQueryStep } from "../../common"
+import { cancelFulfillmentWorkflow } from "../../fulfillment"
+import { adjustInventoryLevelsStep } from "../../inventory"
+import { cancelOrderFulfillmentStep } from "../steps/cancel-fulfillment"
+import {
+  throwIfItemsDoesNotExistsInOrder,
+  throwIfOrderIsCancelled,
+} from "../utils/order-validation"
+
+const validateOrder = createStep(
```
olivermrbl, line 19: Are we checking if the items in the fulfillment has been shipped? Those cannot be canceled.

olivermrbl (commented):
Nice, I left a few comments here and there, but otherwise LGTM

packages/modules/order/src/utils/actions/cancel-item-fulfillment.ts
carlos-r-l-rodrigues: ~~In this case no much, because the engine is generic.
But we should never see this message working with the module methods.~~

it does makes sense actually.

packages/medusa/src/api/admin/orders/[id]/fulfillments/[fulfillment_id]/cancel/route.ts
Comment (carlos-r-l-rodrigues) [reply packages/medusa/src/api/admin/orders/[id]/fulfillments/[fulfillment_id]/cancel/route.ts:25]: I see. The create would also behave the same? which is the current problem you caught. admin is not calling `orders/:id/fulfillment` to create it

packages/medusa/src/api/admin/orders/[id]/fulfillments/[fulfillment_id]/cancel/route.ts
Comment (carlos-r-l-rodrigues) [reply packages/medusa/src/api/admin/orders/[id]/fulfillments/[fulfillment_id]/cancel/route.ts:30]: nope. I will remove it

packages/core/core-flows/src/order/workflows/cancel-order-fulfillment.ts
carlos-r-l-rodrigues: Oh, I removed it by mistake... good catch.

packages/medusa/src/api/admin/orders/[id]/fulfillments/[fulfillment_id]/cancel/route.ts
Comment (olivermrbl) [reply packages/medusa/src/api/admin/orders/[id]/fulfillments/[fulfillment_id]/cancel/route.ts:25]: Yeah, I think we should be able to have one fulfillment Rest API for all operations. But we would still need to switch on the resource ID in the right and call the correct workflow, e.g. `createOrderFulfillmentWorkflow`, etc.

Comment (olivermrbl) [reply packages/medusa/src/api/admin/orders/[id]/fulfillments/[fulfillment_id]/cancel/route.ts:25]: Not necessarily something we have to do now. We can do it in a later PR.

olivermrbl (approved):
LGTM

PR #7576, opened 2024-06-02
feat(core-flows,modules-sdk,types,medusa,link-modules): adds variant <> inventory item link endpoints

riqwan:
what:

- adds variant inventory link management endpoints:
```
Link inventory item to variant
POST /products/:id/variants/:vid/inventory-items

Update variant's inventory item link
POST /products/:id/variants/:vid/inventory-items/:iid

Unlink variant's inventory item
DELETE /products/:id/variants/:vid/inventory-items/:iid
```

- a batch endpoint that does the above 3 across variants
```
POST /products/:id/variants/inventory-items
```

### DIFF
--- a/integration-tests/modules/__tests__/common/workflows.spec.ts
+++ b/integration-tests/modules/__tests__/common/workflows.spec.ts
@@ -191,7 +191,7 @@ medusaIntegrationTestRunner({
 
             expect(errors).toEqual([
               {
-                action: "update-links-step",
+                action: "update-remote-links-step",
                 handlerType: "invoke",
                 error: expect.objectContaining({
                   message: `Could not find all existing links from data`,
--- a/packages/core/core-flows/src/common/steps/dismiss-remote-links.ts
+++ b/packages/core/core-flows/src/common/steps/dismiss-remote-links.ts
@@ -5,6 +5,7 @@ import { ContainerRegistrationKeys } from "@medusajs/utils"
 
 type DismissRemoteLinksStepInput = LinkDefinition | LinkDefinition[]
 
+// TODO: add ability for this step to restore links from only foreign keys
 export const dismissRemoteLinkStepId = "dismiss-remote-links"
 export const dismissRemoteLinkStep = createStep(
   dismissRemoteLinkStepId,
--- a/packages/modules/link-modules/src/services/dynamic-service-class.ts
+++ b/packages/modules/link-modules/src/services/dynamic-service-class.ts
@@ -1,4 +1,5 @@
 import { Constructor, ILinkModule, ModuleJoinerConfig } from "@medusajs/types"
+import { isDefined } from "@medusajs/utils"
 import { LinkModuleService } from "@services"
 
 export function getModuleService(
@@ -11,9 +12,9 @@ export function getModuleService(
 
   // If extraDataFields is not defined, pick the fields to populate and validate from the
   // database config if any fields are provided.
-  if (typeof joinerConfig_.extraDataFields === "undefined") {
+  if (!isDefined(joinerConfig_.extraDataFields)) {
     joinerConfig_.extraDataFields = Object.keys(
-      databaseConfig.extraDataFields || {}
+      databaseConfig.extraFields || {}
     )
   }
 
--- a/packages/core/core-flows/src/common/steps/update-remote-links.ts
+++ b/packages/core/core-flows/src/common/steps/update-remote-links.ts
@@ -6,6 +6,10 @@ export const updateRemoteLinksStepId = "update-remote-links-step"
 export const updateRemoteLinksStep = createStep(
   updateRemoteLinksStepId,
   async (data: LinkDefinition[], { container }) => {
+    if (!data.length) {
+      return new StepResponse([], [])
+    }
+
     const link = container.resolve<RemoteLink>(
       ContainerRegistrationKeys.REMOTE_LINK
     )


packages/core/modules-sdk/src/remote-link.ts
```diff
@@ -453,4 +455,72 @@ export class RemoteLink {
...
+
+    for (const [serviceName, filters] of serviceLinks) {
+      const service = this.modulesMap.get(serviceName)!
+
+      promises.push(
+        service
+          .list({ $or: filters })
+          .then((links: any[]) =>
+            options?.asLinkDefinition
+              ? convertRecordsToLinkDefinition(links, service)
```
riqwan, line 499: @carlos-r-l-rodrigues please take a look at this method, more specifically this transformation.

Outside of this class, its hard for workflows to accurately perform a revert because it doesn't have information about which fields in the database are the primary keys or which one are extra "data" fields.

I've added a helper in this class to convert database records to link definition shape, which the workflows can then pass around to all the create/update/dismiss methods. I quite liked using it as it made the workflows super easy. 

wdyt? Let me know if you have a better idea 🤞🏻

packages/core/modules-sdk/src/remote-link.ts
```diff
@@ -453,4 +455,72 @@ export class RemoteLink {
...
+      if (!service) {
+        throw new Error(
+          linkingErrorMessage({
+            moduleA,
+            moduleAKey,
+            moduleB,
+            moduleBKey,
+            type: "link",
+          })
+        )
```
carlos-r-l-rodrigues: do you mind to create a private function `getLinkModuleOrThrow`? it can also be public.
there are now 3 places doing that same thing.

packages/core/modules-sdk/src/remote-link.ts
carlos-r-l-rodrigues: This looks good. Nice addition to it.

packages/core/types/src/modules-sdk/index.ts
```diff
@@ -137,6 +137,7 @@ export type ModuleJoinerConfig = Omit<
    */
   schema?: string
   relationships?: ModuleJoinerRelationship[]
+  extraFields?: string[]
```
carlos-r-l-rodrigues: suggestion: `extraDataFields`
at a first glance I thought this was the same that is inside `databaseConfig`.
And if you don't mind, we could add a comment below saying that `databaseConfig` won't be present on the public `__joinerConfig`, and the extra fields from there will be populated here if not defined.

packages/modules/link-modules/src/services/dynamic-service-class.ts
```diff
@@ -5,7 +5,12 @@ export function getModuleService(
   joinerConfig: ModuleJoinerConfig
 ): Constructor<ILinkModule> {
   const joinerConfig_ = JSON.parse(JSON.stringify(joinerConfig))
+  const databaseConfig = joinerConfig_.databaseConfig
+
   delete joinerConfig_.databaseConfig
+
+  joinerConfig_.extraFields = Object.keys(databaseConfig.extraFields || {})
```
carlos-r-l-rodrigues: I think we can rename this to extraDataFields to do not be mixed up to the one inside databaseConfig, and I  believe if there is something already defined on this field we must validate if all of them are defined inside databaseConfig, or throw.

packages/core/types/src/modules-sdk/index.ts
riqwan: Ah wait, I think I added this in the wrong place. Should be in the loaded module type and it's a generated read-only attribute derived from the database config.

Does this need to be a configurable field?

packages/core/modules-sdk/src/remote-link.ts
riqwan: Yes sure. There's a TODO on line 464 to do this exactly. I wanted to get your ✅ on the approach before I do that

packages/core/types/src/modules-sdk/index.ts
carlos-r-l-rodrigues: The use cases I see to keep it here, is if someone wants extra fields added to the database, but don't want to update them from the application.
Or if you replace the link module by a custom module or an external module and you can't infer these fields.
So I think it is a good thing to keep it configurable.

riqwan: Added 👍🏻

carlos-r-l-rodrigues (approved):

integration-tests/modules/__tests__/common/workflows.spec.ts
```diff
@@ -0,0 +1,314 @@
...
+  updateLinksWorkflow,
+  updateLinksWorkflowId,
+} from "@medusajs/core-flows"
+import { Modules } from "@medusajs/utils"
+import { medusaIntegrationTestRunner } from "medusa-test-utils/dist"
+import { createAdminUser } from "../../../helpers/create-admin-user"
+
+jest.setTimeout(50000)
+
+export const adminHeaders = {
```
sradevski: You can import this from the helpers

packages/core/core-flows/src/common/steps/dismiss-remote-links.ts
```diff
@@ -13,18 +13,27 @@ export const dismissRemoteLinkStep = createStep(
     const link = container.resolve<RemoteLink>(
       ContainerRegistrationKeys.REMOTE_LINK
     )
+
+    // Our current revert strategy for dismissed links are to recreate it again.
```
sradevski, line 23: Should we open a ticket to support this out of the box? I think having to keep the original data in memory here is a lot more wasteful

packages/medusa/src/api/admin/products/validators.ts
```diff
@@ -235,3 +236,56 @@ export const AdminBatchUpdateProduct = AdminUpdateProduct.extend({
...
+  })
+  .strict()
+export type AdminBatchUpdateVariantInventoryItemType = z.infer<
+  typeof AdminBatchUpdateVariantInventoryItem
+>
+
+export const AdminBatchDeleteVariantInventoryItem = z
+  .object({
+    inventory_item_id: z.string(),
+    variant_id: z.string(),
```
sradevski, line 280: Shouldn't the variant be picked up from the route?

riqwan: The non batch ones do, this one is a batch across variants `/products/:id/variants/inventory-items`

packages/core/core-flows/src/common/steps/dismiss-remote-links.ts
riqwan: I think so. The only way to not do this would be to soft delete links. Would have to check if this approach was taken intentionally cc. @carlos-r-l-rodrigues

packages/medusa/src/api/admin/products/validators.ts
sradevski: @riqwan ah ok, cause eg. for the variants batch method we scope it to a single product (so in this case it would be `products/:id/variants/:id/inventory-items`), but if we have this use-case to manage inventory items across multiple variants then it's fine I guess 👍

packages/core/core-flows/src/common/steps/dismiss-remote-links.ts
sradevski: Yeah I was thinking we do soft-deletes like everywhere else, but let's see what Carlos says

carlos-r-l-rodrigues: They are soft deleted by default, even when you call the method RemoteLink.delete.
And the method create is an upsert under the hood. You can try to just call the create again with the PK and FK and it will just set the deleted_at null.
If this removes the extra fields, we'll need another method `restoreLink` or something...

packages/modules/link-modules/src/services/dynamic-service-class.ts
```diff
@@ -5,7 +5,18 @@ export function getModuleService(
   joinerConfig: ModuleJoinerConfig
 ): Constructor<ILinkModule> {
   const joinerConfig_ = JSON.parse(JSON.stringify(joinerConfig))
+  const databaseConfig = joinerConfig_.databaseConfig
+
   delete joinerConfig_.databaseConfig
+
+  // If extraDataFields is not defined, pick the fields to populate and validate from the
+  // database config if any fields are provided.
+  if (typeof joinerConfig_.extraDataFields === "undefined") {
```
carlos-r-l-rodrigues:
```suggestion
  if (!isDefined(joinerConfig_.extraDataFields)) {
```

packages/core/core-flows/src/common/workflows/batch-links.ts
```diff
@@ -0,0 +1,32 @@
...
+  batchLinksWorkflowId,
+  (
+    input: WorkflowData<
+      BatchWorkflowInput<LinkDefinition, LinkDefinition, LinkDefinition>
+    >
+  ) => {
+    const [created, updated, deleted] = parallelize(
+      createRemoteLinkStep(input.create || []),
+      updateRemoteLinksStep(input.update || []),
+      dismissRemoteLinkStep(input.delete || [])
```
adrien2p, line 23: suggestion: Can we check in all the steps that there is an input and otherwise return prematurely?

adrien2p (approved):

packages/core/core-flows/src/common/workflows/batch-links.ts
riqwan: all of them are doing that already :)

packages/core/core-flows/src/common/steps/dismiss-remote-links.ts
riqwan: Change doesn't seem as easy, I've added a TODO on the file.

PR #7577, opened 2024-06-02
chore: Move customer + customer group integration tests and fixes issues

olivermrbl:
**What**
- Migrate to V2 and move admin customer integration tests from `/api` to `/http`
- Migrate to V2 and move admin customer group integration tests from `/api` to `/http`
- Add unique constraint to customer group name to stay backward compatible
- Add `GET /admin/customer-groups/:id/customers` endpoint
- Resolves minor issues spawning from migrating tests to V2

Note, a bunch of `/batch` tests were removed in the migration.


riqwan (approved):
✨ clean!

packages/modules/customer/src/models/customer-group.ts
```diff
@@ -1,21 +1,33 @@
...
-  ManyToMany,
-  Collection,
-  Filter,
 } from "@mikro-orm/core"
 import Customer from "./customer"
 import CustomerGroupCustomer from "./customer-group-customer"
 
 type OptionalGroupProps = DAL.SoftDeletableEntityDateColumns // TODO: To be revisited when more clear
 
+const CustomerGroupUniqueName = createPsqlIndexStatementHelper({
```
sradevski, line 24: Should we maybe normalize the group name to be case-insensitive?

packages/medusa/src/api/admin/customer-groups/middlewares.ts
```diff
@@ -64,4 +66,14 @@ export const adminCustomerGroupRoutesMiddlewares: MiddlewareRoute[] = [
       ),
     ],
   },
+  {
```
sradevski: I believe having `/admin/customers?customer_group=123` is better than having these relationship endpoints (and is the pattern we followed elsewhere, like price list products). 

We could always have this endpoint before releasing for backwards compatibility, but it should only reroute to the `/customers` endpoint. WDYT?

olivermrbl: Good push. I added this mainly to stay backward compatible, but I am fine removing them tbh. I don't imagine these endpoints are being used a lot, so maybe we should just introduce the breaking change. One less endpoint for us to maintain. Wdyt?

sradevski: @olivermrbl yeah I think we should go for it, and then when we think about migrations we look into what we want to keep as in v1 (and add a proxy for redirecting/remapping) and what we want to just go with a breaking change. But for now I think it's good to go from first principles and design the API as we want it.

packages/modules/customer/src/models/customer-group.ts
olivermrbl: To me names are a bit different from e.g. handles, and therefore I think we can have looser restrictions for them. 

E.g. I would be fine with having "VIP Customers" and "vip customers".

sradevski (approved):

PR #7580, opened 2024-06-03
chore: move CI to node 21

riqwan:


adrien2p (approved):

olivermrbl:
LGTM!

Only thing I thought about was that maybe it makes sense to always have our pipeline use the minimum supported node version, which will be 20 in 2.0. Wdyt?

riqwan:
Fair enough, looks like the performance benefit between 20 and 21 isn't significant. Will close this one! 🤞🏻

PR #7581, opened 2024-06-03
chore: Move currency, collection, and auth tests to http folder

sradevski:


olivermrbl (approved):
LGTM

integration-tests/http/__tests__/currency/admin/currency.spec.ts
```diff
@@ -0,0 +1,66 @@
...
+          expect.objectContaining({
+            code: "usd",
+            name: "US Dollar",
+          }),
+        ])
+      })
+    })
+  },
+
+  // BREAKING: There was an "should update currency includes_tax" test that no longer applies in v2 (realted to MEDUSA_FF_TAX_INCLUSIVE_PRICING)
```
olivermrbl, line 65: We will need to add back the tax inclusivity tests when the feature has been implemented

sradevski: Got it, let's see if it still makes sense to put that flag on the currency or it should be in some other place (eg. in tax module)

PR #7583, opened 2024-06-03
chore: fix random product batch fail spec

riqwan:


riqwan:
Closing this as @sradevski & @adrien2p are doing a proper fix

PR #7584, opened 2024-06-03
WIP: Implementing events to the pricing module

thetutlage:
This PR is not ready to be merged yet. However, I am opening it as a draft for early feedback


packages/modules/pricing/integration-tests/__tests__/services/pricing-module/price-set.spec.ts
```diff
@@ -639,6 +682,24 @@ moduleIntegrationTestRunner({
...
+              object: "price",
+              data: { id: priceSet.prices![1].id },
+            })
+          )
+          expect(eventBusEmitSpy.mock.calls[0][0][1]).toEqual(
+            composeMessage(PricingEvents.price_rule_created, {
+              service: Modules.PRICING,
+              action: CommonEvents.CREATED,
+              object: "price_rule",
+              data: { id: priceSet.prices![1].price_rules[0].id },
```
thetutlage, line 708: As per types, there is no property called `price_rules` on a price.

packages/modules/pricing/src/services/pricing-module.ts
```diff
@@ -1021,12 +1072,53 @@ export default class PricingModuleService<
             price_set_id: priceSetId,
             title: "test", // TODO: accept title
             rules_count: numberOfRules,
-            priceRules,
+            price_rules: priceRules,
```
thetutlage, line 1040: Works with `price_rules`, but not with `priceRules`. However, in both the cases, the `CreatePriceDTO` does not this property

thetutlage:
@adrien2p Can I some early review on this PR? And also, if possible need a bit of help deciding which methods to ignore when emitting events :)

packages/core/utils/src/pricing/events.ts
```diff
@@ -0,0 +1,27 @@
+import { buildEventNamesFromEntityName } from "../event-bus"
+import { Modules } from "../modules-sdk"
+
+const eventBaseNames: [
```
sradevski, line 4: Do we even need these explicitly typed now? I guess it will be the `PricingEvents` that will be used directly, right?

packages/modules/pricing/integration-tests/__tests__/services/pricing-module/price-list.spec.ts
```diff
@@ -488,6 +503,40 @@ moduleIntegrationTestRunner({
               ]),
             })
           )
+
+          expect(eventBusEmitSpy.mock.calls[0][0]).toHaveLength(4)
+          expect(eventBusEmitSpy.mock.calls[0][0][0]).toEqual(
+            composeMessage(PricingEvents.price_list_created, {
```
sradevski, line 510: One suggestion on the event names, maybe we can have them `<entity_name>.<action>` instead of using snake case for all of it. Like that, you can know what is the entity name, and what is the action, so it is a bit cleaner.

packages/modules/pricing/integration-tests/__tests__/services/pricing-module/price-set.spec.ts
```diff
@@ -422,6 +437,40 @@ moduleIntegrationTestRunner({
               ],
             })
           )
+
+          const [priceRules] = await service.listPriceRules({
+            price_set_id: [priceSet.id],
+          })
+
+          expect(eventBusEmitSpy.mock.calls[0][0]).toHaveLength(3)
+          expect(eventBusEmitSpy.mock.calls[0][0][0]).toEqual(
```
sradevski: Those [0][0][0] are so ugly to read 😄 Maybe the mock event bus can expose some helper methods for asserting calls instead?

packages/modules/pricing/src/services/pricing-module.ts
```diff
@@ -626,6 +636,7 @@ export default class PricingModuleService<
   }
 
   @InjectManager("baseRepository_")
+  @EmitEvents()
```
sradevski, line 639: What events would this emit?

packages/modules/pricing/integration-tests/__tests__/services/pricing-module/price-list.spec.ts
```diff
@@ -488,6 +503,40 @@ moduleIntegrationTestRunner({
               ]),
             })
           )
+
+          expect(eventBusEmitSpy.mock.calls[0][0]).toHaveLength(4)
+          expect(eventBusEmitSpy.mock.calls[0][0][0]).toEqual(
```
adrien2p: nit: maybe we can store the emitted events such as `const emittedEvents = eventBusEmitSpy.mock.calls[0][0]` and then from it later you can do `expect(emittedEvents[0]`

packages/modules/pricing/integration-tests/__tests__/services/pricing-module/price-list.spec.ts
```diff
@@ -809,6 +858,24 @@ moduleIntegrationTestRunner({
               price_list_rules: [],
             })
           )
+
+          expect(eventBusEmitSpy.mock.calls[0][0]).toHaveLength(2)
```
adrien2p: todo: l 796 add `jest.clearAllMocks()` to not be bothered if any of the previous method to populate some data emit some events. You can usually do that just before the method you want to check the events from. I let you validate where you need it in all the tests :D

packages/modules/pricing/integration-tests/__tests__/services/pricing-module/price-set.spec.ts
```diff
@@ -422,6 +437,40 @@ moduleIntegrationTestRunner({
...
+          expect(eventBusEmitSpy.mock.calls[0][0][2]).toEqual(
+            composeMessage(PricingEvents.price_rule_created, {
+              service: Modules.PRICING,
+              action: CommonEvents.CREATED,
+              object: "price_rule",
+              data: {
+                id: priceRules.id,
+              },
+            })
+          )
```
adrien2p, line 474: todo: we need to add some expectation in the test bellow, I can see that we jumped to some tests later on

packages/modules/pricing/src/services/pricing-module.ts
adrien2p: as discussed we can fix it by changing to snake case as you mentioned

adrien2p (commented):
Looks good so far 💪

packages/core/utils/src/pricing/events.ts
adrien2p: we need it to infer the types of the PricingEvents, so yes

packages/modules/pricing/integration-tests/__tests__/services/pricing-module/price-list.spec.ts
adrien2p: if you look at the code, the event name is already composed like that `<source>.<object>.<action>`, also, I dont think we should tight the object to the event based on some calculation because it happens to be the case here. wdyt?

packages/modules/pricing/integration-tests/__tests__/services/pricing-module/price-set.spec.ts
adrien2p: normally we just use `toHaveBeenCalledWith`, which is already the helper to do that

packages/modules/pricing/src/services/pricing-module.ts
adrien2p: all the events in the message aggregator

packages/modules/pricing/integration-tests/__tests__/services/pricing-module/price-list.spec.ts
sradevski: maybe I am missing something, but if I look in `event-builder-factory.ts` we have
```
isMainEntity
          ? eventsEnum[action]
          : eventsEnum[`${object}_${action}`],
```

I am suggesting we do
```
isMainEntity
          ? eventsEnum[action]
          : eventsEnum[`${object}.${action}`],
```
instead.

packages/modules/pricing/integration-tests/__tests__/services/pricing-module/price-set.spec.ts
sradevski: Ah indeed, so maybe we can use that here instead.

packages/modules/pricing/src/services/pricing-module.ts
sradevski: I mean, what would they be, since I don't see us explicitly sending events in this method, but we are in some of the other ones. Or is there some convention around it?

packages/modules/pricing/integration-tests/__tests__/services/pricing-module/price-list.spec.ts
adrien2p: that would feel weird in my opinion, now if you use the enum to grab an event from the user perspective you would always have to go through `eventEnums['something.created']` instead of being able to do `eventEnums.some_entity_created`. Remember, it is an enum that can be used for subscribers for example

packages/modules/pricing/src/services/pricing-module.ts
adrien2p: When you apply this decorator, what it does are the following
- Inject the message aggregator to the shared context
- wrap the function in order to execute the emit after the function is done

Now the shared context is passed down to any function called by this function, so any function aggregating messages to the current message aggregator, will get emitted once this function reach the end. Does that make sense?


Now, this pr is WIP and not yet ready to be reviewed so it is normal that you see things not finished or no message being aggregated everywhere as described in the description but it will come 😄

packages/modules/pricing/integration-tests/__tests__/services/pricing-module/price-set.spec.ts
thetutlage: The debugging experience with `toHaveBeenCalledWith` is not great. When you give it an array of events and one of the events mis-match, then there is no feedback on which index there is a mis-match.

Whereas with individual assertion calls, you know exactly which assertion failed

packages/modules/pricing/integration-tests/__tests__/services/pricing-module/price-list.spec.ts
thetutlage: Aren't we doing this in `beforeEach`?

packages/modules/pricing/integration-tests/__tests__/services/pricing-module/price-list.spec.ts
sradevski: It is just a common pattern in naming events in most of the event-driven architecture resources I've read, but I see your point as well. @thetutlage any opinions?

packages/modules/pricing/src/services/pricing-module.ts
sradevski: Got it yes, it makes sense 😄

packages/modules/pricing/integration-tests/__tests__/services/pricing-module/price-list.spec.ts
adrien2p: @sradevski are you sure you are not mixing up the event name and the enum property name?

thetutlage: @sradevski The events are named with dot `.` separator. Its just the enum helpers has `_` convention. Let me share an example:

- Let's say the event is called `product.created`
- Now you want to reference this event somewhere in your code. You have two options.

Either you can write the name manually as

```ts
composeEvent('product.created')
```

Or you may use some Object/enum as a auto-complete unit.

```ts
composeEvent(EventsListEnum['product.created'])
```

Now the above syntax is less ergonomic in typing. Because you have first type square brackets and then maybe get autocomplete for available events list.

Whereas, when this helper use `snake_case`, then it becomes a bit easier to reference. Like this

```ts
composeEvent(EventsListEnum.product_created)
```

sradevski: Yes I was 🤦  😄

PR #7586, opened 2024-06-03
chore(order): cancel order

carlos-r-l-rodrigues:
Added a few missing awaits that can be causing intermittent issues in the pipeline

### DIFF
--- a/packages/core/core-flows/src/order/workflows/cancel-order.ts
+++ b/packages/core/core-flows/src/order/workflows/cancel-order.ts
@@ -114,7 +114,7 @@ export const cancelOrderWorkflow = createWorkflow(
     })
     deleteReservationsByLineItemsStep(lineItemIds)
 
-    const payment_ids = transform({ order }, ({ order }) => {
+    const paymentIds = transform({ order }, ({ order }) => {
       return deepFlatMap(
         order,
         "payment_collections.payments",
@@ -123,7 +123,7 @@ export const cancelOrderWorkflow = createWorkflow(
         }
       )
     })
-    cancelPaymentStep({ payment_ids })
+    cancelPaymentStep({ paymentIds })
 
     cancelOrdersStep({ orderIds: [order.id] })
   }


packages/core/core-flows/src/order/steps/cancel-orders.ts
```diff
@@ -0,0 +1,50 @@
...
+    )
+
+    const orders = await service.list(
+      {
+        id: data.orderIds,
+      },
+      {
+        select: ["id", "status"],
+      }
+    )
```
olivermrbl, line 24: question: Would it be OK to pass the `data.orderIds` directly to `cancel`?

packages/modules/order/src/services/order-module-service.ts
```diff
@@ -2753,10 +2753,55 @@ export default class OrderModuleService<
     }
 
     await this.orderService_.update(
+      orderIds.map((id) => {
+        return {
+          id,
+          status: OrderStatus.COMPLETED,
+        }
+      }),
```
olivermrbl, line 2761: question: Do we have a `this.update()` that accepts a selector?

packages/modules/order/src/services/order-module-service.ts
```diff
@@ -2753,10 +2753,55 @@ export default class OrderModuleService<
...
+      order.canceled_at = canceled_at
+    }
+
+    await this.orderService_.update(
+      orderIds.map((id) => {
+        return {
+          id,
+          status: OrderStatus.CANCELED,
+          canceled_at,
+        }
```
olivermrbl, line 2803: Same question

packages/core/types/src/order/service.ts
```diff
@@ -1512,6 +1515,13 @@ export interface IOrderModuleService extends IModuleService {
     sharedContext?: Context
   ): Promise<void>
 
+  /*
+  cancelReturn(
+    returnData: CancelOrderReturnDTO,
+    sharedContext?: Context
+  ): Promise<void>
+  */
```
olivermrbl, line 1523: question: Coming later or safe to remove?

packages/core/core-flows/src/order/workflows/cancel-order.ts
```diff
@@ -0,0 +1,130 @@
...
+    }
+
+    throwIfOrderIsCancelled({ order })
+
+    let refunds = 0
+    let captures = 0
+
+    deepFlatMap(order_, "payment_collections.payments", ({ payments }) => {
+      refunds += payments?.refunds?.length ?? 0
+      captures += payments?.captures?.length ?? 0
```
olivermrbl, line 40: thought: Guess this should be migrated to BigNumber no?

packages/core/core-flows/src/payment/steps/cancel-payment.ts
```diff
@@ -0,0 +1,35 @@
+import { ModuleRegistrationName } from "@medusajs/modules-sdk"
+import { IPaymentModuleService, Logger } from "@medusajs/types"
+import { ContainerRegistrationKeys, promiseAll } from "@medusajs/utils"
+import { createStep } from "@medusajs/workflows-sdk"
+
+type StepInput = {
+  payment_ids: string | string[]
```
olivermrbl: nit(non-blocking): camelCase?

olivermrbl (approved):
Overall LGTM, think we need to play around a bit with all the order operations to know if we have convered all cases. This is not limited to the cancel order flow.

packages/core/core-flows/src/order/steps/cancel-orders.ts
carlos-r-l-rodrigues: I need to query it before to be able to revert on the compensation.

packages/core/core-flows/src/order/workflows/cancel-order.ts
carlos-r-l-rodrigues: This is only counting if there is any refund or capture. Not their amount.

packages/core/types/src/order/service.ts
carlos-r-l-rodrigues: This is coming later. The return will have its own id, and this is not yet implemented in the module.

packages/modules/order/src/services/order-module-service.ts
carlos-r-l-rodrigues: yes, but that will also call this same internal service.
do you think we should change it?

packages/core/core-flows/src/order/workflows/cancel-order.ts
olivermrbl: Ah sorry, I missed that 😅

packages/core/types/src/order/service.ts
olivermrbl: 👍

packages/modules/order/src/services/order-module-service.ts
olivermrbl: I am fine keeping it as is, just thought it would read a bit simpler. Nothing to address now

packages/core/core-flows/src/order/steps/cancel-orders.ts
carlos-r-l-rodrigues: but I'm not sure I understood your question correctly.

PR #7589, opened 2024-06-03
feat(dashboard,core-flows,medusa): update fulfillment flows

fPolic:
**What**
- update fulfillment create/cancel flows to work with latest API
- cancel order action

**Fix**
- fix cancel fulfillment endpoint to accept params correctly
- fix inventory adjustment step in the cancel workflow if the item's variant doesn't have `manage_inventory`
- make cancel order endpoint POST

---

CLOSES CORE-2234


olivermrbl:
Getting this error on creating fulfillment right now:
CleanShot 2024-06-04 at 10 29 40

I think this will be fixed by #7592, as this is because there are no reservations for the variant, because we have `manage_inventory=false`. 

Can you confirm?

fPolic:
@olivermrbl - exactly, this is fixed with the https://github.com/medusajs/medusa/pull/7594

packages/admin-next/dashboard/src/components/table/table-cells/order/fulfillment-status-cell/fulfillment-status-cell.tsx
```diff
@@ -1,4 +1,4 @@
-import type { FulfillmentStatus } from "@medusajs/medusa"
+import { FulfillmentStatus } from "@medusajs/medusa"
```
carlos-r-l-rodrigues: The admin should not import `@medusa/medusa`, should it?

carlos-r-l-rodrigues (approved):
LGTM. just 1 import issue

packages/admin-next/dashboard/src/components/table/table-cells/order/fulfillment-status-cell/fulfillment-status-cell.tsx
fPolic: Yes, good catch this was wrong autoimport

packages/admin-next/dashboard/src/hooks/api/orders.ts
```diff
@@ -38,3 +45,57 @@ export const useOrders = (
 
   return { ...data, ...rest }
 }
+
+export const useCreateOrderFulfillment = (
```
olivermrbl: thought: Should we migrate the client endpoints to use the SDK?

packages/admin-next/dashboard/src/lib/client/orders.ts
```diff
@@ -1,5 +1,10 @@
-import { getRequest } from "./common"
-import { OrderListRes, OrderRes } from "../../types/api-responses"
+import { getRequest, postRequest } from "./common"
```
olivermrbl: I guess we can remove this now right?

packages/core/js-sdk/src/admin/order.ts
```diff
@@ -0,0 +1,81 @@
...
+  AdminCancelOrderFulfillment,
+  FindParams,
+  HttpTypes,
+  PaginatedResponse,
+  SelectParams,
+} from "@medusajs/types"
+import { Client } from "../client"
+import { ClientHeaders } from "../types"
+
+export class Order {
```
olivermrbl, line 11: Nice, think we should create types for this, but to not extend the scope of the PR further, we could save that for a follow-up PR

olivermrbl (approved):
LGTM, just one comment

PR #7590, opened 2024-06-03
fix(workflows-sdk): workflow registration

carlos-r-l-rodrigues:


sradevski (approved):
Should we maybe log an info?

PR #7591, opened 2024-06-03
Feat: Add product tag endpoints, move tests to HTTP folder

sradevski:


packages/core/core-flows/tsconfig.json
```diff
@@ -18,7 +18,7 @@
     "skipLibCheck": true,
     "downlevelIteration": true
   },
-  "include": ["src"],
+  "include": ["src", "src/product/steps/delete-product-tags/.ts"],
```
olivermrbl:
```suggestion
  "include": ["src"],
```

packages/medusa/src/api/admin/product-tags/validators.ts
```diff
@@ -0,0 +1,46 @@
...
+  })
+)
+
+export type AdminCreateProductTagType = z.infer<typeof AdminCreateProductTag>
+export const AdminCreateProductTag = z
+  .object({
+    value: z.string(),
+    metadata: z.record(z.string(), z.unknown()).optional(),
+  })
+  .strict()
```
olivermrbl, line 38: thought: Should we default to `strict()` validation? This has fooled me a few times.

Is there a use case for wanting to let unexpected props pass through?

olivermrbl (approved):

packages/medusa/src/api/admin/product-tags/validators.ts
sradevski: @olivermrbl Yes I think we can be strict by default. The only issue is that we can't apply it globally in an easy way, since if we use preprocess (eg. to parse the data) the return type is ZodEffect, which can't have `strict()` applied to it. 

Anyway, let me just open a ticket and we can see if we can find a way around it.

PR #7592, opened 2024-06-03
chore(workflows-sdk): check exported workflow

carlos-r-l-rodrigues:


adrien2p (approved):

packages/core/workflows-sdk/src/helper/__tests__/compose.ts
```diff
@@ -1972,4 +1974,52 @@ describe("Workflow composer", function () {
     expect(mockStep1Fn.mock.calls[0][0]).toEqual(workflowInput)
     expect(mockCompensateSte1.mock.calls[0][0]).toEqual({ data: "data" })
   })
+
+  it("should throw if the same workflow is defined using different steps", async () => {
```
sradevski, line 1978: I was thinking, once we have plugins, maybe people would want to replace one workflow with another one (eg. they want to change the behavior without rewiring everything else). Should we maybe allow for this in that case?

sradevski (approved):

packages/core/workflows-sdk/src/helper/__tests__/compose.ts
adrien2p: We need to be careful with that and also think about the way we want to approach that ability. But if we allow it, you can't know if many extensions are trying to replace the same workflow and in that case the last one will be the only one left without you noticing

sradevski: @adrien2p definitely agreed that we need to be careful how to approach it, but I think it would be an important thing to support if we want to allow for easy extensibility. I think as long as it's clear that how you load your plugins matters, it should be ok for the last one to win. 

Anyway, just bringing it up as a discussion topic, we don't need to change anything in this PR.

carlos-r-l-rodrigues: I agree that we should let the entire flow be replaced, but I think this should be explicity, not by mistake. And this is what we're trying to prevent here.
I believe we have to have this discussion when we enable workflow edits. Where you could change the global definition of any given workflow (GlobalWorkflow), or just change one instance of a flow before its execution (LocalWorkflow).

PR #7594, opened 2024-06-03
fix(order): ignore reservation when manage_inventory is false

carlos-r-l-rodrigues:


fPolic (approved):
LGTM!

adrien2p (approved):

packages/core/core-flows/src/order/workflows/create-fulfillment.ts
```diff
@@ -157,6 +154,15 @@ function prepareInventoryUpdate({ reservations, order, input }) {
...
     const reservation = reservationMap[item.id]
+    if (!reservation) {
+      if (item.manage_inventory) {
+        throw new Error(
+          `No stock reservation found for item ${item.id} - ${item.title} (${item.variant_title})`
+        )
+      }
+      continue
+    }
+
```
sradevski, line 165: Should we refactor it to
```suggestion
      if (item.manage_inventory) {
       continue
      }
      
      const reservation = reservationMap[item.id]
    if (!reservation) {
        throw new Error(
          `No stock reservation found for item ${item.id} - ${item.title} (${item.variant_title})`
        )
    }

```

sradevski (approved):

olivermrbl (approved):

packages/core/core-flows/src/order/workflows/create-fulfillment.ts
carlos-r-l-rodrigues: That would change the behavior. Custom items won't have "manage_inventory" but they can or cannot have a reservation.
and adding extra `&&` or `||` to the `if` wouldn't make it easier to read.

PR #7597, opened 2024-06-04
Remove redundant methods from pricing module

thetutlage:
During events implementation process, we have identified certain methods that are not required in the Pricing module service


adrien2p (approved):
LGTM, as discussed, those methods did not make sense to exists in the first place, just for info

riqwan (approved):
🥇

sradevski (approved):

PR #7599, opened 2024-06-04
feat(core-flows,medusa): adds inventory kit creation to variants endpoint

riqwan:
what:

When creating a variant, we can now create inventory as a part of the product and variants create endpoint.

This applies only to variants where `manage_inventory=true`. 2 cases present itself:

1. When inventory_items are present
  - Link an inventory item with required_quantity to the variant
  - the inventory item already needs to be present
2. When inventory_items are not present
  - A default inventory item will be created
  - links the created item to the variant with a default required_quantity
  
  
RESOLVES CORE-2220

### DIFF
--- a/packages/core/core-flows/src/product/workflows/create-product-variants.ts
+++ b/packages/core/core-flows/src/product/workflows/create-product-variants.ts
@@ -59,7 +59,7 @@ const buildLinksToCreate = (data: {
     if (!hasInventoryItems) {
       const inventoryItem = data.inventoryIndexMap[index]
 
-      linksToCreate.push(buildLink(variant.id, inventoryItem.id))
+      linksToCreate.push(buildLink(variant.id, inventoryItem.id, 1))
 
       continue
     }


packages/core/core-flows/src/product/workflows/create-product-variants.ts
```diff
@@ -12,9 +16,109 @@ import { createVariantPricingLinkStep } from "../steps/create-variant-pricing-li
 type WorkflowInput = {
   product_variants: (ProductTypes.CreateProductVariantDTO & {
     prices?: PricingTypes.CreateMoneyAmountDTO[]
+  } & {
+    inventory_items?: {
+      inventory_item_id: string
+      required_quantity: number
```
carlos-r-l-rodrigues: should this be optional and we default it to 1?

packages/core/core-flows/src/product/workflows/create-product-variants.ts
```diff
@@ -12,9 +16,109 @@ import { createVariantPricingLinkStep } from "../steps/create-variant-pricing-li
...
+
+      continue
+    }
+
+    for (const inventoryInput of variantInput.inventory_items || []) {
+      linksToCreate.push(
+        buildLink(
+          variant.id,
+          inventoryInput.inventory_item_id,
+          inventoryInput.required_quantity
```
carlos-r-l-rodrigues: should we default it to 1 at the application level? even if the db has 1 by default.

riqwan: only on the workflow or the http as well? I like the idea of requiring it on the http, but happy to make this optional and default to 1. wdyt?

carlos-r-l-rodrigues: Yeah, agree. on the http I think we can require it. But the workflow should set it to 1 by default and be optional

packages/core/core-flows/src/product/workflows/create-product-variants.ts
riqwan:
```suggestion
      required_quantity?: number
```

packages/core/core-flows/src/product/workflows/create-product-variants.ts
riqwan:
```suggestion
          inventoryInput.required_quantity ?? 1
```

packages/core/core-flows/src/product/workflows/create-product-variants.ts
```diff
@@ -12,9 +16,109 @@ import { createVariantPricingLinkStep } from "../steps/create-variant-pricing-li
...
+      inventory_item_id: string
+      required_quantity?: number
+    }[]
   })[]
 }
 
+const buildLink = (
+  variant_id: string,
+  inventory_item_id: string,
+  required_quantity?: number
```
sradevski: This can be required maybe?

sradevski (approved):

packages/core/core-flows/src/product/workflows/create-product-variants.ts
adrien2p: I believe the idea was to require it from the http layer, but make it optional from the workflow layer and default to 1 (as per the discussion above) which I think make sense wdyt?

sradevski: the default is set in the caller of `buildLink`, which means this parameter should be required (otherwise `required_quantity` will be undefined)

riqwan: that is true, good catch. 🤞🏻

adrien2p: nice catch 👍

PR #7601, opened 2024-06-04
fix: Several fixes to store product endpoints, moved several test suites to HTTP

sradevski:


integration-tests/http/__tests__/api-key/admin/publishable-key.spec.ts
```diff
@@ -0,0 +1,291 @@
...
+          await api.post(
+            "/admin/api-keys",
+            // BREAKING: The type field is now required
+            { title: "just a title", type: "publishable" },
+            adminHeaders
+          )
+        ).data.api_key
+      })
+
+      // BREAKING: The URL changed from /admin/api-keys to /admin/api-keys, as well as the response field
```
riqwan: whats the difference in the url here?

packages/medusa/src/api/admin/api-keys/[id]/route.ts
```diff
@@ -20,6 +21,13 @@ export const GET = async (
     req.remoteQueryConfig.fields
   )
 
+  if (!apiKey) {
+    throw new MedusaError(
```
Comment (riqwan) [inline packages/medusa/src/api/admin/api-keys/[id]/route.ts:25]: sugg: This will be the behavior for all route endpoints, might be good to build this into refetchEntity helper

packages/medusa/src/api/store/products/middlewares.ts
```diff
@@ -57,6 +58,7 @@ export const storeProductRoutesMiddlewares: MiddlewareRoute[] = [
         StoreGetProductsParams,
         QueryConfig.retrieveProductQueryConfig
       ),
+      applyParamsAsFilters({ id: "id" }),
```
riqwan, line 61: There's already a `applyDefaultFilters` that you can use for this.

packages/medusa/src/api/utils/middlewares/products/filter-by-valid-sales-channels.ts
```diff
@@ -1,79 +1,74 @@
...
-    if (!apiKey) {
-      try {
-        throw new MedusaError(
-          MedusaError.Types.INVALID_DATA,
-          `Publishable API key not found`
-        )
-      } catch (e) {
-        return next(e)
-      }
+    if (salesChannelIds) {
```
riqwan:
```suggestion
    if (salesChannelIds?.length) {
```

packages/medusa/src/api/utils/middlewares/products/filter-by-valid-sales-channels.ts
```diff
@@ -1,79 +1,74 @@
...
-      } catch (e) {
-        return next(e)
-      }
+    if (!store) {
+      return next(
+        new MedusaError(MedusaError.Types.INVALID_DATA, `Store not found`)
+      )
     }
 
+    req.filterableFields.sales_channel_id = [store.default_sales_channel_id]
```
riqwan: sugg: this could still be undefined, might have to throw here in those cases

riqwan (approved):
Added some comments, otherwise lgtm!

integration-tests/http/__tests__/api-key/admin/publishable-key.spec.ts
sradevski: Ah god damn autocomplete 😄 It was `/admin/publishable-api-keys`

packages/medusa/src/api/admin/api-keys/[id]/route.ts
Comment (sradevski) [reply packages/medusa/src/api/admin/api-keys/[id]/route.ts:25]: Yeah I agree, probably worth going through all endpoints and apply it in 1 go. Let me open a ticket for it.

packages/medusa/src/api/store/products/middlewares.ts
sradevski: Ah I guess I can use the function there, let me switch 👍

sradevski: The thing is, the filter value comes from the path params, and currently you can't do that with `applyDefaultFilters`. I think having this makes sense as it has different semantics

packages/medusa/src/api/utils/middlewares/products/filter-by-valid-sales-channels.ts
sradevski: Good point, I guess there is no guarantee that there is a default sales channel. I don't think we need to throw though, we just don't apply any sales channels in that case

PR #7603, opened 2024-06-04
fix(dashboard): Use ChipInput for option forms

kasperkristensen:
Resolves CORE-2069, CORE-2068


sradevski (approved):

fPolic (approved):
LGTM

Suggestion: can we use `autocomplete="off"` on options input so we don't get this weird suggestions with commas.

PR #7605, opened 2024-06-04
chore: parallelize steps

carlos-r-l-rodrigues:


sradevski (approved):
Nice cleanup 👍

packages/core/core-flows/src/product/workflows/update-products.ts
```diff
@@ -166,14 +167,15 @@ export const updateProductsWorkflow = createWorkflow(
...
 
-    dismissRemoteLinkStep(toDeleteLinks)
-
     const salesChannelLinks = transform(
       { input, updatedProducts },
       prepareSalesChannelLinks
     )
 
-    createRemoteLinkStep(salesChannelLinks)
+    parallelize(
```
adrien2p: suggestion: Can you check that this is being executed, I believe I had the case that if you dont store the result in a variable used elsewhere this does not get executed :D I d like to get your feedback

adrien2p (commented):
Very nice cleanup, just one thing to check :D

packages/core/core-flows/src/product/workflows/update-products.ts
carlos-r-l-rodrigues: just reverted it. that one can't be in parallel.

packages/core/core-flows/src/definition/cart/workflows/add-shipping-method-to-cart.ts
```diff
@@ -78,10 +79,13 @@ export const addShippingMethodToWorkflow = createWorkflow(
...
     })
-    refreshCartPromotionsStep({ id: input.cart_id })
-    updateTaxLinesStep({
-      cart_or_cart_id: input.cart_id,
-      shipping_methods: shippingMethods,
-    })
+
+    parallelize(
+      refreshCartPromotionsStep({ id: input.cart_id }),
+      updateTaxLinesStep({
```
riqwan, line 85: One thing to check here is that the adjustments created by promotions would be accounted for in tax lines. Im not sure if this can be paralrllized.

carlos-r-l-rodrigues: I don't see adjustments in the update tax line workflow

riqwan: Ah ok, might be a missed implementation. Ignore this comment, I'll create a ticket 🤞🏻

carlos-r-l-rodrigues: so that can't be in parallel?
should I merge this PR and later you remove it from the parallelize?

carlos-r-l-rodrigues: But I believe the taxes and adjustments are independent and only calculated when we display totals, isn't it?

riqwan: You're right! Sorry, I got confused between tax rates and tax amount

PR #7609, opened 2024-06-04
feat: Move shipping option + profile tests and support updating profiles

olivermrbl:
**What**
- Move shipping options http tests from `modules/` to `http/`
- Move shipping profiles http tests from `modules/` to `http/`
- Add support for updating and upsert shipping profiles
- Delete old shipping options test

@adrien2p had already created some good tests for options and profiles in the `modules/` folder, so I just moved them over and replaced the few things that were consuming the modules directly with HTTP calls.


integration-tests/http/__tests__/shipping-option/admin/shipping-option.spec.ts
```diff
@@ -363,6 +379,280 @@ medusaIntegrationTestRunner({
...
+          expect(response.data).toEqual({
+            type: "invalid_data",
+            message:
+              "each value in rule_ids must be a string, rule_ids should not be empty",
+          })
+        })
+
+        // FYI: This test was the only run being run for shipping option rules.
+        // TODO: The other tests should probably be updated?
+        it("should throw error when shipping option does not exist", async () => {
```
olivermrbl: Among the shipping option rules, this was the only test currently being run.

I have kept it this way and will get back later to see why other tests were skipped previously.

olivermrbl:
I didn't replace most of the old shipping options test as the domain in 2.0 has diverged significantly from 1.0.

integration-tests/http/__tests__/shipping-option/admin/shipping-option.spec.ts
```diff
@@ -30,38 +20,64 @@ medusaIntegrationTestRunner({
...
-      beforeAll(async () => {
-        appContainer = getContainer()
-        fulfillmentModule = appContainer.resolve(
-          ModuleRegistrationName.FULFILLMENT
-        )
-        regionService = appContainer.resolve(ModuleRegistrationName.REGION)
-      })
-
       beforeEach(async () => {
+        const appContainer = getContainer()
```
sradevski, line 25: NIT: I think just adding a `BREAKING: Shipping setup has significantly changed from v1, exact migration needs more investigation` or something along those lines would be nice

olivermrbl: Good push, added that in 👍

PR #7615, opened 2024-06-05
chore: Move stock location tests to HTTP folder

sradevski:


integration-tests/http/__tests__/stock-location/admin/stock-location.spec.ts
```diff
@@ -1,91 +1,100 @@
...
+          `/admin/sales-channels`,
+          {
+            name: "Test SC 2",
+          },
+          adminHeaders
+        )
+      ).data.sales_channel
+    })
+
+    describe("create stock location", () => {
```
riqwan, line 70: while you're moving these tests, can you also rename the describes to the actual endpoint names?

riqwan (approved):

integration-tests/http/__tests__/stock-location/admin/stock-location.spec.ts
sradevski: Let me create a ticket for it, as we've already gone through most of the admin tests 😄

PR #7616, opened 2024-06-05
chore: return meta information on created/updated/deleted object from upsertWithReplace

adrien2p:
**What**
- Return the updated/deleted/created entities from the upsert with replace according to the configuration

This will help to emit the events from the product module as it rely on this abstraction in many places

### DIFF
--- a/integration-tests/http/__tests__/customer/admin/customer.spec.ts
+++ b/integration-tests/http/__tests__/customer/admin/customer.spec.ts
@@ -174,21 +174,16 @@ medusaIntegrationTestRunner({
 
     describe("POST /admin/customers", () => {
       it("should create a customer", async () => {
-        const response = await api
-          .post(
-            "/admin/customers",
-            {
-              first_name: "newf",
-              last_name: "newl",
-              email: "new@email.com",
-              password: "newpassword",
-              metadata: { foo: "bar" },
-            },
-            adminHeaders
-          )
-          .catch((err) => {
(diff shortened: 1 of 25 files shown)


adrien2p:
@sradevski can I get you to have a look at this one please 💪

packages/core/utils/src/dal/mikro-orm/mikro-orm-repository.ts
```diff
@@ -622,18 +661,38 @@ export function mikroOrmBaseRepositoryFactory<T extends object = object>(
             })
           })
 
-          await this.upsertMany_(manager, relation.type, normalizedData)
+          const { performedActions: performedActions_ } =
+            await this.upsertMany_(manager, relation.type, normalizedData)
+          this.mergePerformedActions(performedActions, performedActions_)
         }
 
+        const toDeleteEntities = await manager.find<any>(
```
sradevski, line 669: I assume this should work as the entityMap will be empty, so it will go straight to the DB, right?

sradevski (approved):
Looking great!

riqwan (approved):
🥇

packages/core/utils/src/dal/mikro-orm/mikro-orm-repository.ts
adrien2p: Ues that's right 👍

PR #7617, opened 2024-06-05
feat: Fix subscribers loading + add order <> cart link

olivermrbl:
What
- Add link between orders and carts
- Fix subscribers loading*

*We were trying to load each subscriber file with our Subscribers class instead of `subscribers/` directories, and the tool isn't built for that. I am not entirely sure if I have missed something, but I reverted it back to how it worked in V1.


packages/medusa/src/api/store/carts/[id]/complete/route.ts
```diff
@@ -1,22 +1,48 @@
...
+        const result = await refetchEntity(
+          "order_cart",
+          { cart_id },
+          req.scope,
+          ["order_id", "cart_id"]
+        )
+
+        await eventBusModule.emit("order.placed", { id: result.order_id })
+      },
+    },
```
Comment (olivermrbl) [inline packages/medusa/src/api/store/carts/[id]/complete/route.ts]: @carlos-r-l-rodrigues @adrien2p, what do you think about this?

I kept the idempotency out of this PR. We can address it in a follow up, when we have time to discuss the options.

packages/medusa/src/api/store/carts/[id]/complete/route.ts
```diff
@@ -1,22 +1,48 @@
...
+        }
+
+        const result = await refetchEntity(
+          "order_cart",
+          { cart_id },
+          req.scope,
+          ["order_id", "cart_id"]
+        )
+
+        await eventBusModule.emit("order.placed", { id: result.order_id })
```
Comment (adrien2p) [inline packages/medusa/src/api/store/carts/[id]/complete/route.ts]: nit: I would just create an events.ts file in the utils package under the order dir with the following
```ts
const eventBaseNames: [
  "order"
] = [
  "order",
]

export const OrderEvents = {
  ...buildEventNamesFromEntityName(
    eventBaseNames,
    Modules.ORDER
  ),
  'order_placed': 'order.placed'
}
```
and then consume it here, not necessarely in that pr but just a thoughts

adrien2p (commented):
LGTM 💪

packages/medusa/src/api/store/carts/[id]/complete/route.ts
Comment (olivermrbl) [reply packages/medusa/src/api/store/carts/[id]/complete/route.ts]: This was added under the assumption that the endpoint `POST /store/carts/:id/complete` would only be used to complete carts converting them into a regular order.

If we choose to also handle for example exchange/swap completion through this endpoint, we will need to rethink the approach.

packages/medusa/src/api/store/carts/[id]/complete/route.ts
Comment (olivermrbl) [reply packages/medusa/src/api/store/carts/[id]/complete/route.ts]: Yeah think this is a good idea!

packages/medusa/src/api/store/carts/[id]/complete/route.ts
```diff
@@ -1,22 +1,48 @@
...
   const { errors, result } = await completeCartWorkflow(req.scope).run({
     input: { id: cart_id },
     context: { transactionId: cart_id },
     throwOnError: false,
+    events: {
+      onFinish: async ({ transaction }) => {
+        const state = transaction.getState()
+
+        // We only want to emit the order.placed event if the workflow completed
+        if (state !== TransactionState.DONE) {
```
Comment (sradevski) [inline packages/medusa/src/api/store/carts/[id]/complete/route.ts]: A general question - should the event emission be part of the workflow? I.e. if we fail to emit the order placed event, do we want to rollback?

I think if another critical system depends on these events, that might be the desired behavior, so having reliable events is probably quite important

sradevski (approved):

packages/core/core-flows/src/definition/cart/workflows/complete-cart.ts
```diff
@@ -86,6 +87,13 @@ export const completeCartWorkflow = createWorkflow(
 
     linkOrderAndPaymentCollectionsStep(linkOrderPaymentCollection)
 
+    createRemoteLinkStep([
```
carlos-r-l-rodrigues: we can use `paralellize` to call `linkOrderAndPaymentCollectionsStep` and `createRemoteLinkStep`

packages/medusa/src/api/store/carts/[id]/complete/route.ts
Comment (srindom) [reply packages/medusa/src/api/store/carts/[id]/complete/route.ts]: I am not a big fan of the API route owning the event emission. Couldn't this be a step in the workflow instead? 

My thinking is that people might want to build an experience on top of the cart+order flows, (e.g., a quotation flow). This experience will maybe have some different validation than the default cart endpoints, but still rely on the cartCompletionWorkflow. I'd expect `order.placed` to be emitted in this case without having to explicitly add the onFinish handler like this.

Comment (carlos-r-l-rodrigues) [reply packages/medusa/src/api/store/carts/[id]/complete/route.ts]: I don't believe exchange and swap will use this flow. The order edits in general will have their own flows.
In case we want to reuse the cart for some reason to complete these other flows, they will be new carts anyway and they wouldn't generate new orders

Comment (carlos-r-l-rodrigues) [reply packages/medusa/src/api/store/carts/[id]/complete/route.ts]: But I think you can use the workflow engine here and use the result to send the event. The idempotency part will come in later

packages/medusa/src/api/store/carts/[id]/complete/route.ts
Comment (srindom) [reply packages/medusa/src/api/store/carts/[id]/complete/route.ts]: +1

packages/modules/link-modules/src/definitions/order-cart.ts
```diff
@@ -0,0 +1,58 @@
...
+      },
+      relationship: {
+        serviceName: LINKS.OrderCart,
+        primaryKey: "order_id",
+        foreignKey: "id",
+        alias: "cart_link",
+      },
+    },
+    {
+      serviceName: Modules.CART,
```
carlos-r-l-rodrigues, line 49: I think we can also add the fieldAlias to have `order` in the Cart as well

packages/medusa/src/api/store/carts/[id]/complete/route.ts
Comment (olivermrbl) [reply packages/medusa/src/api/store/carts/[id]/complete/route.ts]: Hmm, I did think of this. 

There are many ways of using events in the context of workflows, and I think we will need to differentiate the types of events we want, to know where something fits well. 

The types of events I can think of right now are:
- Pure module events: Events emitted within modules when entities are created, updated, or deleted, and there can be A LOT of these during the lifetime of a workflow execution. These will need to be grouped somehow and emitted at the very end (or have an option specifying that these should be emitted immediately)
- Workflow lifecycle events: Events that are emitted when workflow-specific state changes, e.g. step succeeds, workflow reverts, etc.
- Domain-specific events: Events that are human-readable and specific to the context in which the workflow is executed, e.g. completing a cart (which is what we do here). 

There are likely more types of events than this.

Maybe we have to dictate what guarantees the different types of events offer? For example, I am not sure I want to roll back the entire cart-completion because `order.placed` failed to be emitted. Maybe you should use `order.created` if you want that guarantee for your use case.

-----------------

Separately, if we choose to incorporate the event as a step, it **must** be the last step, which is something we currently can't guarantee because workflows can be extended anywhere in the flow.

That said, with controlled extensions, this would work, so if we introduce that concept this is definitely a path we can go down.

packages/medusa/src/api/store/carts/[id]/complete/route.ts
Comment (olivermrbl) [reply packages/medusa/src/api/store/carts/[id]/complete/route.ts]: > Couldn't this be a step in the workflow instead?

The reason I didn't add it as a step is that we currently can't guarantee it is the last step in the workflow. As mentioned in a separate comment, if we add controlled extensions, this is reasonable and maybe even the preferred approach.

Should we instead introduce this functionality under the assumption, that we will do that in the near future?

packages/medusa/src/api/store/carts/[id]/complete/route.ts
Comment (sradevski) [reply packages/medusa/src/api/store/carts/[id]/complete/route.ts]: Whether you want to rollback or not will depend on how you use events. If it is just for emails, then sure, you don't want to rollback. However, if you have another system you want to keep in sync (eg. reporting taxes to the tax office), having these events unreliable will be a big issue.

And I agree with you, we need to specify what sort of guarantees we offer for events and let people build their systems around them. However, I think that is orthogonal to where the events are emitted, because you can have an optional workflow step that is sending an event, and if it fails we can still complete the workflow, which might be a better solution in this case. WDYT?

As a side note, if users have an overview of all the steps that have failed, (and ideally they can retry), they can emit the events manually and still ensure eventual consistency.

Comment (olivermrbl) [reply packages/medusa/src/api/store/carts/[id]/complete/route.ts]: > Whether you want to rollback or not will depend on how you use events. If it is just for emails, then sure, you don't want to rollback. However, if you have another system you want to keep in sync (eg. reporting taxes to the tax office), having these events unreliable will be a big issue.

Yes, exactly – I agree, which is why I wanted the differentiation between these events and their guarantees.

> you can have an optional workflow step that is sending an event, and if it fails we can still complete the workflow, which might be a better solution in this case

I am fine going with this for now. As I mentioned, the main reason for choosing the `onFinish` over a step right now is that we can't control the extensions to a workflow, and I wanted to be guaranteed this was emitted at the end.

Lemme try to change it to use a step and I'll ask for a re-review from everyone.

packages/medusa/src/api/store/carts/[id]/complete/route.ts
Comment (carlos-r-l-rodrigues) [reply packages/medusa/src/api/store/carts/[id]/complete/route.ts]: When we have events grouped, even if the event is not the last step, it will only be fired when the flow is completed.

packages/medusa/src/api/store/carts/[id]/complete/route.ts
Comment (sradevski) [reply packages/medusa/src/api/store/carts/[id]/complete/route.ts]: @olivermrbl I think it's actually fine even if it gets extended right? If someone extends it to do something else after an order is placed, they can emit another event that they can react to. 

Semantically speaking, after this workflow is done, an order has been placed. Now, if you want extend the workflow to print a **thank you** note, you can have your workflow send a `order.thank_you_printed` (or whatever 😄) and you can listen to that instead/as well.

Now if you want to inject something before the `order.placed` event is triggered, then we'll need these injection/extension zones we've talked about

Comment (riqwan) [reply packages/medusa/src/api/store/carts/[id]/complete/route.ts]: @olivermrbl if you pass in the group_id to the event in the steps/workflow, it will only be emitted once the workflow is completed.

Comment (olivermrbl) [reply packages/medusa/src/api/store/carts/[id]/complete/route.ts]: @riqwan, has this been implemented? I know we talked about it previously, but I can't remember if we have built it.

Comment (riqwan) [reply packages/medusa/src/api/store/carts/[id]/complete/route.ts]: Good point, its on step 3 of the events project (we're currently at 1).

Comment (carlos-r-l-rodrigues) [reply packages/medusa/src/api/store/carts/[id]/complete/route.ts]: this is not implemented yet. but this will be the behavior and probably the priority for that will be increased.

Comment (olivermrbl) [reply packages/medusa/src/api/store/carts/[id]/complete/route.ts]: > its on step 3 of the events project

@riqwan where do you see these steps? I can't find the project

Comment (riqwan) [reply packages/medusa/src/api/store/carts/[id]/complete/route.ts]: on the events RFC, sending you the link

olivermrbl:
Okay, so I have removed everything related to the `order.placed` event, since we, as per the discussion internally, wait for the grouped events mechanism to be in place.

Now, this PR only does two things:
- Fixes the subscribers loading
- Add the order <> link

Without the `order.placed` event, these changes are completely unrelated, but let's keep it like this for now. No reason to separate it into two PRs 😄

packages/core/core-flows/src/definition/cart/workflows/complete-cart.ts
```diff
@@ -75,16 +75,20 @@ export const completeCartWorkflow = createWorkflow(
 
     const order = createOrderFromCartStep({ cart: finalCart })
 
-    const linkOrderPaymentCollection = transform({ order, cart }, (data) => ({
-      links: [
+    parallelize(
```
carlos-r-l-rodrigues:
```suggestion

```

carlos-r-l-rodrigues (approved):

packages/core/core-flows/src/definition/cart/workflows/complete-cart.ts
olivermrbl: Thx 🤕

sradevski (approved):
I guess we'll just leave the event emission for later?

adrien2p (approved):

PR #7618, opened 2024-06-05
chore: Move fulfillment set tests to http folder

sradevski:


integration-tests/http/__tests__/fulfillment/admin/fulfillment-sets.spec.ts
```diff
@@ -1,50 +1,51 @@
...
+      // ).data.fulfillment_set
+
+      // TODO: Add support for creating fulfillment sets through HTTP
+      const fulfillmentModule: IFulfillmentModuleService =
+        getContainer().resolve(ModuleRegistrationName.FULFILLMENT)
+
+      fulfillmentSet1 = await fulfillmentModule.create({
+        name: "Test fulfillment set",
+        type: "pickup",
+      })
```
olivermrbl, line 37: You could replace this with:

```ts
let location = (
  await api.post(
    `/admin/stock-locations`,
    {
      name: "Test location",
    },
    adminHeaders
  )
).data.stock_location

location = (
  await api.post(
    `/admin/stock-locations/${location.id}/fulfillment-sets?fields=*fulfillment_sets`,
    {
      name: "Test",
      type: "test-type",
    },
    adminHeaders
  )
).data.stock_location

fulfillmentSet = (
  await api.post(
    `/admin/fulfillment-sets/${location.fulfillment_sets[0].id}/service-zones`,
    {
      name: "Test",
      geo_zones: [{ type: "country", country_code: "us" }],
    },
    adminHeaders
  )
).data.fulfillment_set
```

if we really want to get rid of the module usage. I did this in a different test suite.

olivermrbl: Btw. this is how fulfillment sets are created at the HTTP level. They don't have a "standalone" endpoint.

olivermrbl (approved):

integration-tests/http/__tests__/fulfillment/admin/fulfillment-sets.spec.ts
sradevski: I wonder if we need to revisit how fulfillment sets are created though, what we did in other places (like inventory items) is to create the inventory item on its own route, but then link it to a variant (in this case, stock location) on the `stock-locations/:id/fulfillment-sets` route.

I'll leave it for now, and once we have all tests in place we can clean them up.

PR #7620, opened 2024-06-05
chore: fix inventory items + reservations specs

riqwan:
what:

- splits inventory item and reservation specs
- fixes all skipped tests, except 1


packages/medusa/src/api/admin/inventory-items/validators.ts
```diff
@@ -72,7 +72,6 @@ export const AdminCreateInventoryLocationLevel = z
     location_id: z.string(),
     stocked_quantity: z.number().min(0).optional(),
     incoming_quantity: z.number().min(0).optional(),
-    reserved_quantity: z.number().min(0).optional(),
```
sradevski, line 75: @riqwan the tests used to create this as well, was that a mistake?

sradevski (approved):
Nice cleanup 👍

packages/medusa/src/api/admin/inventory-items/validators.ts
riqwan: Yeah, must've been created at a time when reservations weren't available yet.

PR #7622, opened 2024-06-05
fix(modules-sdk): dependencies

carlos-r-l-rodrigues:
```
moduleB: {
  resolve: __dirname + "/modules/modB",
  dependencies: ["moduleC"],
},
moduleC: {
  resolve: __dirname + "/modules/modC",
}
```

it enables `moduleB` to access `this.container.moduleC` outside its constructor

### DIFF
--- a/packages/core/modules-sdk/src/medusa-module.ts
+++ b/packages/core/modules-sdk/src/medusa-module.ts
@@ -300,6 +300,7 @@ export class MedusaModule {
         resources: declaration?.resources || MODULE_RESOURCE_TYPE.ISOLATED,
         resolve: defaultPath,
         options: declaration?.options ?? declaration,
+        dependencies: declaration?.dependencies ?? [],
         alias: declaration?.alias,
         main: declaration?.main,
         worker_mode: workerMode,


sradevski (approved):

sradevski:
@carlos-r-l-rodrigues I assume this is for things like the event bus and logger right? We wouldn't want people to start calling eg. product module from inside order module

carlos-r-l-rodrigues:
> @carlos-r-l-rodrigues I assume this is for things like the event bus and logger right? We wouldn't want people to start calling eg. product module from inside order module

This is only for custom modules actually.
The ones we own the definition are working fine. Eg: to inject `logger` as dependency.

We won't use this for our modules, but we can't prevent people from creating modules/utils and have them injected in the container.

@edast needs to share the FeatureFlag module to be used inside other methods from other modules.

The drawback now is that they won't be available in the constructor. If that turns to be a requirement in the future we have to change the way modules are loaded.

sradevski:
@carlos-r-l-rodrigues ok sounds good. I think this is good for now, I am just worried when people start implementing modules and they start injecting them, it can end up in a big spaghetti mess, so maybe we want to be more opinionated in what we allow (which in turn should give us more control when the time comes for breaking changes and migrations). Just food for thought, I am fine with the PR btw

PR #7625, opened 2024-06-05
feat(types,js-sdk,medusa,fulfillment): Types and SDK methods related to Stock Locations

kasperkristensen:
Since we still haven't come to an agreement with which fields to make optional I have just keep the types as is, and just cleaned up the naming + adding missing types.

I have updated the query param validators for some endpoints, as they were either completely wrong (e.g. using createSelectParams), or missing any fields other than `limit`, `offset`, and `order`.

### DIFF
--- a/packages/admin-next/dashboard/src/components/common/switch-box/index.ts
+++ b/packages/admin-next/dashboard/src/components/common/switch-box/index.ts
@@ -1,0 +1,1 @@
+export * from "./switch-box"
--- a/packages/admin-next/dashboard/src/components/common/switch-box/switch-box.tsx
+++ b/packages/admin-next/dashboard/src/components/common/switch-box/switch-box.tsx
@@ -1,0 +1,70 @@
+import { Switch } from "@medusajs/ui"
+import { ReactNode } from "react"
+import { ControllerProps, FieldPath, FieldValues } from "react-hook-form"
+
+import { Form } from "../../common/form"
+
+interface HeadlessControllerProps<
+  TFieldValues extends FieldValues = FieldValues,
+  TName extends FieldPath<TFieldValues> = FieldPath<TFieldValues>
+>  extends Omit<ControllerProps<TFieldValues, TName>, "render"> {}
+
+interface SwitchBoxProps<
+  TFieldValues extends FieldValues = FieldValues,
+  TName extends FieldPath<TFieldValues> = FieldPath<TFieldValues>
+> extends HeadlessControllerProps<TFieldValues, TName>{
+  label: string
+  description: string
+  optional?: boolean
+  tooltip?: ReactNode
+  /**
+   * Callback for performing additional actions when the checked state changes.
+   * This does not intercept the form control, it is only used for injecting side-effects.
+   */
+  onCheckedChange?: (checked: boolean) => void
+}
+
+/**
+ * Wrapper for the Switch component to be used with `react-hook-form`.
+ * 
+ * Use this component whenever a design calls for wrapping the Switch component
+ * in a container with a label and description.
+ */
+export const SwitchBox = <
+  TFieldValues extends FieldValues = FieldValues,
+  TName extends FieldPath<TFieldValues> = FieldPath<TFieldValues>
+>({
+  label,
+  description,
+  optional = false,
+  tooltip,
+  onCheckedChange,
+  ...props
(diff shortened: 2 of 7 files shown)


packages/modules/fulfillment/src/models/fulfillment-provider.ts
```diff
@@ -9,6 +9,7 @@ import {
 
 @Entity()
 export default class FulfillmentProvider {
+  @Searchable()
```
kasperkristensen, line 12: Does adding this require modifying the migration?

packages/modules/fulfillment/src/models/shipping-option.ts
```diff
@@ -68,6 +69,7 @@ export default class ShippingOption {
   @PrimaryKey({ columnType: "text" })
   id: string
 
+  @Searchable()
```
kasperkristensen, line 72: Does adding this require modifying the migration?

packages/core/types/src/http/pricing/admin/entitites.ts
```diff
@@ -0,0 +1,31 @@
...
+  currency_code: string
+  amount: number
+  raw_amount: Record<string, unknown>
+  min_quantity: number | null
+  max_quantity: number | null
+  price_set_id: string
+  created_at: string
+  updated_at: string
+  deleted_at: string | null
+}
```
kasperkristensen, line 31: self review: I had no idea on how to type this. The model has several fields that never seem to be returned even when expanding everything.

packages/core/types/src/http/fulfillment/admin/payloads.ts
```diff
@@ -0,0 +1,46 @@
...
+export interface AdminCreateFulfillment {
+  location_id: string
+  provider_id: string
+  delivery_address: AdminFulfillmentDeliveryAddress
+  items: AdminCreateFulfillmentItem[]
+  labels: AdminCreateFulfillmentLabel[]
+  // Validator requires an empty object for the order field
+  // seems like an issue with the validator but including it here
+  // to make the type match the validator.
+  order: Record<string, unknown>
```
kasperkristensen, line 39: The validator for this payload seems to be off, but I have matched it to prevent any errors.

sradevski: Yeah I also had some issues now that I was moving some tests, this is fine for now, just add a TODO please so it's a bit easier to find it.

packages/medusa/src/api/admin/fulfillment-sets/validators.ts
```diff
@@ -44,20 +44,9 @@ export const AdminUpdateFulfillmentSetServiceZonesSchema = z
...
-    id: z.union([z.string(), z.array(z.string())]).optional(),
-    name: z.union([z.string(), z.array(z.string())]).optional(),
-    type: z.union([z.string(), z.array(z.string())]).optional(),
-    service_zone_id: z.union([z.string(), z.array(z.string())]).optional(),
-    created_at: createOperatorMap().optional(),
-    updated_at: createOperatorMap().optional(),
-    deleted_at: createOperatorMap().optional(),
-  })
-)
+createSelectParams()
```
sradevski: This seems like a mistake?

packages/modules/fulfillment/src/models/fulfillment-provider.ts
sradevski: No it doesn't, this is all you need

sradevski (approved):
Great stuff 👍

olivermrbl (commented):
Very nice @kasperkristensen 💪

packages/medusa/src/api/admin/fulfillment-sets/validators.ts
kasperkristensen: There isn’t any list endpoint for fulfillment sets, and this was being used for endpoint that were returning a single set

olivermrbl: I think @sradevski just means the free-floating `createSelectParams()`. Think that can be deleted no?

kasperkristensen: Ahh okay, sorry yea thats a mistake will remove it 👍

PR #7629, opened 2024-06-06
chore(core-flows,inventory,types,medusa): add fixes to inventory module + location levels api

riqwan:
what:

- santizes inputs to prevent reserved_quantity from being updated directly
- inventory items create api can create location levels
- add validation to update quantity of reservation items
- general cleanup

RESOLVES CORE-2254


integration-tests/http/__tests__/inventory/admin/inventory.spec.ts
```diff
@@ -612,6 +612,38 @@ medusaIntegrationTestRunner({
             })
           )
         })
+
+        it("should create inventory items along with location levels", async () => {
+          const response = await api.post(
+            `/admin/inventory-items?fields=%2blocation_levels.*`,
```
sradevski: I think you can just write 
```suggestion
            `/admin/inventory-items?fields=*location_levels`,
```

integration-tests/http/__tests__/reservations/admin/reservations.spec.ts
```diff
@@ -88,16 +88,25 @@ medusaIntegrationTestRunner({
...
+          quantity: 10000,
           inventory_item_id: inventoryItem1.id,
           line_item_id: "line-item-id",
           location_id: stockLocation1.id,
         }
 
         const res = await api
-          .post(`/admin/reservations`, payload, adminHeaders)
+          .post(
+            `/admin/reservations?field=inventory_item.location_levels.*`,
```
sradevski: Similarly, the * goes in the beginning now

packages/modules/inventory-next/src/services/inventory-module.ts
```diff
@@ -621,26 +670,47 @@ export default class InventoryModuleService<
...
 
+    const diff = arrayDifference(
+      ids,
+      reservationItems.map((i) => i.id)
+    )
+
+    if (diff.length) {
+      throw new MedusaError(
+        MedusaError.Types.INVALID_DATA,
+        `Reservation item with id ${ids} not found`
```
sradevski:
```suggestion
        `Reservation item with id ${ids.join(', ')} not found`
```
Also, shouldn't we only print the diff?

packages/core/core-flows/src/inventory/workflows/create-inventory-items.ts
```diff
@@ -3,23 +3,86 @@ import {
...
+  items: (InventoryNext.CreateInventoryItemInput & {
+    location_levels?: LocationLevelWithoutInventory[]
+  })[]
+}
+
+const buildLocationLevelMapAndItemData = (data: WorkflowInput) => {
+  const inventoryItems: InventoryNext.CreateInventoryItemInput[] = []
+  // Keep an index to location levels mapping to inject the created inventory item
+  // id into the location levels workflow input
+  const locationLevelMap: Record<number, LocationLevelWithoutInventory[]> = {}
```
sradevski, line 26: I feel this index-based mapping will come to bite us in the butt at one point where the ordering of the operations isn't preserved 😄. I know we don't have much choice though, just thinking out loud.

packages/core/core-flows/src/inventory/workflows/create-inventory-items.ts
```diff
@@ -3,23 +3,86 @@ import {
...
+  let index = 0
+
+  for (const { location_levels, ...inventoryItem } of data.items || []) {
+    locationLevelMap[index] = location_levels?.length ? location_levels : []
```
sradevski, line 29:
```suggestion
    locationLevelMap[index] = location_levels ? location_levels : []
```

packages/core/core-flows/src/inventory/workflows/create-inventory-items.ts
```diff
@@ -3,23 +3,86 @@ import {
...
+}
+
+const buildLocationLevelMapAndItemData = (data: WorkflowInput) => {
+  const inventoryItems: InventoryNext.CreateInventoryItemInput[] = []
+  // Keep an index to location levels mapping to inject the created inventory item
+  // id into the location levels workflow input
+  const locationLevelMap: Record<number, LocationLevelWithoutInventory[]> = {}
+  let index = 0
+
+  for (const { location_levels, ...inventoryItem } of data.items || []) {
```
sradevski: Should we do `data.items.forEach` so you don't have to keep track of the index manually?

packages/core/core-flows/src/inventory/workflows/create-inventory-items.ts
```diff
@@ -3,23 +3,86 @@ import {
...
+const buildInventoryLevelsInput = (data: {
+  locationLevelMap: Record<number, LocationLevelWithoutInventory[]>
+  items: InventoryNext.InventoryItemDTO[]
+}) => {
+  const inventoryLevels: InventoryNext.CreateInventoryLevelInput[] = []
+  let index = 0
+
+  // The order of the input is critical to accurately create location levels for
+  // the right inventory item
+  for (const item of data.items || []) {
```
sradevski: Same here, maybe let's use forEach and get the index from there.

packages/core/core-flows/src/inventory/workflows/create-inventory-items.ts
riqwan: I'm on the same page as you are 😄 I really hate this, but there isn't an alternative here without the module accepting and returning a bulk identifier. Even in that case, you're still accepting the fact that the database is returning the right order its creating, which isn't guaranteed.

carlos-r-l-rodrigues: Should we think about having a `reference_id` column where we control what is inserted and can 100% guarantee to map it back to the correspondent row?
maybe it is better to implement that now, before this happens in production

carlos-r-l-rodrigues: or maybe `map_id`, something internal that we can always use in cases we need that mapping and there isn't any key yet

sradevski: We will definitely need to consider something like that @carlos-r-l-rodrigues . Let me at least open a ticket, and we can revisit this after the public preview release.

riqwan: Yes, thats the only way we can guarantee it

carlos-r-l-rodrigues: but we generate the id of the entities in mikro-orm before inserting them. so there is a great chance that this is already being mapped correctly after inserting.

carlos-r-l-rodrigues (approved):

PR #7630, opened 2024-06-06
fix(dashboard): stale params

fPolic:
**What**
- fix places where we send wrong params (strict validation is now returning 400s for these)
- use SDK instead client for order details loader


sradevski (approved):

kasperkristensen (approved):

riqwan (approved):
lgtm!

PR #7633, opened 2024-06-06
chore: Remove generated client types folder

sradevski:
It seems this isn't used and it didn't have purpose. Autogeneration will probably change if/when we want to introduce it anyway, so removing to keep the codebase clean


adrien2p:
I ve tried to remove it but it gets re generated 😂 we need to change something else I think

sradevski:
> I ve tried to remove it but it gets re generated 😂 we need to change something else I think

oh really?? Ahaha 😄 I couldn't find any references to this folder in the codebase, let me check again.

sradevski:
@shahednasser I can't find anything where this would get generated now, maybe you have more info on where/if this is still happening?

shahednasser (commented):
@sradevski  I've discussed this with @adrien2p before, for documentation purposes we don't need it and we're ok with removing it. However, I remember when Adrien tried to remove it he found out it was used in clients and the admin dashboard. So, not sure if this is related or resolved.

As far as I know, these are generated when you run the `build` command, so maybe remove the `build` script of `openapi-typescript-codegen`? Also, no we don't need the `openapi-typescript-codegen` package (again, for documentation), so we need to check if it's necessary anywhere else.

shahednasser (approved):

adrien2p (approved):
LGTM

PR #7634, opened 2024-06-06
fix: should disable db connection ssl when some url params are provided

adrien2p:
**What**
Enable ssl identification from the client url query params as well


packages/core/utils/src/modules-sdk/load-module-database-config.ts
```diff
@@ -31,7 +31,9 @@ function getDefaultDriverOptions(clientUrl: string) {
   }
 
   if (clientUrl) {
-    return clientUrl.match(/localhost/i) ? localOptions : remoteOptions
+    return clientUrl.match(/localhost|ssl_mode=(disable|false)/i)
```
carlos-r-l-rodrigues: We could check `127.0.0.1` as well. And maybe also if ssl_mode is true?

carlos-r-l-rodrigues (approved):

packages/core/utils/src/modules-sdk/load-module-database-config.ts
adrien2p: let me improve

adrien2p: done 👍

sradevski (approved):

riqwan (approved):
nice!

PR #7636, opened 2024-06-06
feat: emit product tags events

adrien2p:
**What**
emit product tags events

**note**
Stil does not take into consideration non overriden method, delete and soft delete emits event under the hood and we will have to migrate that at the end


sradevski (approved):

riqwan (approved):
🥇

PR #7637, opened 2024-06-06
fix(create-medusa-app): set the database host and port explicitely

shahednasser:
- set the database host and port explicitely (this doesn't apply when the `--db-url` option is used)
- other: set the `AUTH_CORS` environment variable.


adrien2p (approved):
LGTM should we add some unit tests later?

shahednasser:
@adrien2p makes sense, i can look into it maybe tomorrow

riqwan (approved):
🥇

sradevski (approved):

PR #7638, opened 2024-06-06
fix(medusa): return a single category instead of an array in retrieve category route

shahednasser:
The `/store/product-categories/[id]` route returns an array of categories rather than the single category. This PR fixes it

### DIFF
--- a/packages/medusa/src/api/store/product-categories/[id]/route.ts
+++ b/packages/medusa/src/api/store/product-categories/[id]/route.ts
@@ -4,26 +4,25 @@ import {
   AuthenticatedMedusaRequest,
   MedusaResponse,
 } from "../../../../types/routing"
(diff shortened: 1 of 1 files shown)


packages/medusa/src/api/store/product-categories/[id]/route.ts
```diff
@@ -19,11 +19,11 @@ export const GET = async (
     req.remoteQueryConfig.pagination
   )
 
-  if (!category) {
+  if (!category?.length) {
```
Comment (adrien2p) [inline packages/medusa/src/api/store/product-categories/[id]/route.ts]: suggestion: use `refetchEntity` instead of `refetchEntities`?

Comment (shahednasser) [reply packages/medusa/src/api/store/product-categories/[id]/route.ts]: Oh honestly I wasn't aware of it 😄 changing it now

adrien2p (approved):

PR #7643, opened 2024-06-07
feat: Add emitEvent step + cleanup

adrien2p:
**What**
- Generic step to emit event in a workflow, allow no input in order to be able to by pass it if a constraint would like to prevent emitting an event based on some contextual data
- Cleanup all `service` from messages to be `source` which makes more sense and is more align with general event formatting and more agnostic

**note**
This does not include the event group id work to validate it is properly present in the transaction metadata and the place it gets either generated or captured

FIXES CORE-2266


packages/core/core-flows/src/common/steps/emit-event.ts
```diff
@@ -0,0 +1,41 @@
+import { composeMessage, ModuleRegistrationName } from "@medusajs/utils"
+import { createStep, StepExecutionContext } from "@medusajs/workflows-sdk"
+
+type Input = {
+  eventName: string
+  source: string
+  object: string
+  action?: string
+  options?: Record<string, any>
```
riqwan, line 9: what are these here?

packages/core/core-flows/src/common/steps/release-event.ts
```diff
@@ -0,0 +1,20 @@
+import { ModuleRegistrationName } from "@medusajs/utils"
+import { createStep } from "@medusajs/workflows-sdk"
+
+export const releaseEventsStepId = "emit-event-step"
```
sradevski:
```suggestion
export const releaseEventsStepId = "release-event-step"
```

packages/core/core-flows/src/common/steps/emit-event.ts
adrien2p: potential options for the choosen event bus

packages/core/core-flows/src/common/steps/release-event.ts
adrien2p: good catch 😂

olivermrbl (approved):

PR #7644, opened 2024-06-07
Initial implementation of data types and base schema

thetutlage:
The PR is not ready to be merged. But open for early feedback. I have finalized the core data-types and base schema. 

**It is a good time to dis-agree on certain design choices.**

### DIFF
--- a/packages/core/utils/src/dml/__tests__/base_relationship.spec.ts
+++ b/packages/core/utils/src/dml/__tests__/base_relationship.spec.ts
@@ -1,5 +1,5 @@
 import { expectTypeOf } from "expect-type"
-import { StringSchema } from "../schema/string"
+import { TextSchema } from "../schema/text"
 import { BaseRelationship } from "../relations/base"
 
 describe("Base relationship", () => {
@@ -9,7 +9,7 @@ describe("Base relationship", () => {
     }
 
     const user = {
-      username: new StringSchema(),
+      username: new TextSchema(),
     }
 
     const entityRef = () => user
--- a/packages/core/utils/src/dml/__tests__/entity_builder.spec.ts
+++ b/packages/core/utils/src/dml/__tests__/entity_builder.spec.ts
@@ -7,8 +7,8 @@ describe("Entity builder", () => {
     const model = new EntityBuilder()
     const user = model.define("user", {
       id: model.number(),
-      username: model.string(),
-      email: model.string(),
+      username: model.text(),
+      email: model.text(),
     })
 
     expectTypeOf<InstanceType<Infer<typeof user>>>().toMatchTypeOf<{
@@ -21,13 +21,13 @@ describe("Entity builder", () => {
   test("define an entity with relationships", () => {
     const model = new EntityBuilder()
     const email = model.define("email", {
-      email: model.string(),
+      email: model.text(),
       isVerified: model.boolean(),
     })
 
     const user = model.define("user", {
       id: model.number(),
-      username: model.string(),
+      username: model.text(),
       emails: model.hasMany(() => email),
     })
 
@@ -47,7 +47,7 @@ describe("Entity builder", () => {
 
     const user = model.define("user", {
       id: model.number(),
-      username: model.string(),
+      username: model.text(),
       orders: model.hasMany(() => order),
     })
 
--- a/packages/core/utils/src/dml/__tests__/has_many_relationship.spec.ts
+++ b/packages/core/utils/src/dml/__tests__/has_many_relationship.spec.ts
@@ -1,11 +1,11 @@
 import { expectTypeOf } from "expect-type"
 import { HasMany } from "../relations/has_many"
-import { StringSchema } from "../schema/string"
+import { TextSchema } from "../schema/text"
 
 describe("HasMany relationship", () => {
   test("define hasMany relationship", () => {
     const user = {
-      username: new StringSchema(),
+      username: new TextSchema(),
     }
 
     const entityRef = () => user
--- a/packages/core/utils/src/dml/__tests__/has_one_relationship.spec.ts
+++ b/packages/core/utils/src/dml/__tests__/has_one_relationship.spec.ts
@@ -1,11 +1,11 @@
 import { expectTypeOf } from "expect-type"
 import { HasOne } from "../relations/has_one"
-import { StringSchema } from "../schema/string"
+import { TextSchema } from "../schema/text"
 
 describe("HasOne relationship", () => {
   test("define hasOne relationship", () => {
     const user = {
-      username: new StringSchema(),
+      username: new TextSchema(),
     }
 
     const entityRef = () => user
--- a/packages/core/utils/src/dml/__tests__/many_to_many.spec.ts
+++ b/packages/core/utils/src/dml/__tests__/many_to_many.spec.ts
@@ -1,11 +1,11 @@
 import { expectTypeOf } from "expect-type"
+import { TextSchema } from "../schema/text"
 import { ManyToMany } from "../relations/many_to_many"
-import { StringSchema } from "../schema/string"
 
 describe("ManyToMany relationship", () => {
   test("define manyToMany relationship", () => {
     const user = {
-      username: new StringSchema(),
+      username: new TextSchema(),
     }
 
     const entityRef = () => user
--- a/packages/core/utils/src/dml/__tests__/text_schema.spec.ts
+++ b/packages/core/utils/src/dml/__tests__/text_schema.spec.ts
@@ -1,9 +1,9 @@
 import { expectTypeOf } from "expect-type"
-import { StringSchema } from "../schema/string"
+import { TextSchema } from "../schema/text"
 
 describe("String schema", () => {
   test("create string schema type", () => {
-    const schema = new StringSchema()
+    const schema = new TextSchema()
 
     expectTypeOf(schema["$dataType"]).toEqualTypeOf<string>()
     expect(schema.parse("username")).toEqual({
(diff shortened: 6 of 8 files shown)


packages/core/utils/src/dml/types.ts
```diff
@@ -0,0 +1,44 @@
+/**
+ * The supported data types
+ */
+export type KnownDataTypes =
+  | "string"
+  | "boolean"
+  | "enum"
+  | "number"
+  | "date"
+  | "dateTime"
```
olivermrbl: thought: Should we eliminate `dateTime` and enforce that `date` applies the `timestamptz` type?

thetutlage: What if I need a `date` data-type without the time?

olivermrbl: We haven't had the need for it yet in any of our modules, which motivated the suggestion. I personally think the more we can reduce the amount of options the better. Ofc. it should not come at the cost of being too inflexible, but I don't think that's the case here.

What do you think? This is not a strong opinion, so would love to hear what others think too.

olivermrbl: We can always add the other type further down the line, if we see a demand for it.

thetutlage: If we want to add it back later, then we should go with `dateTime` and add back `date` later. Because, otherwise it will be a breaking change.

olivermrbl: Imo. we don't need a direct translation between the schema types and what gets applied under the hood. I would use the most intuitive naming for the most common use cases.

So if we need to add the `date` data type later, we could call it something else, e.g. `simpleDate()` or add an option to `date(type: "simple")`.

thetutlage: I won't push much. But is there any reason that `date` feels better than `dateTime`. Also, introducing a very confusing naming like `simpleDate` won't help much either.

adrien2p: I believe dateTime would be more appropriate, users are also more use to see dataTime when it comes to date + time IMO, simple date I have never seen date anywhere which would add confusion if we add that as it does not exists either in typescript or the DB types or other languages. wdyt?

olivermrbl: > But is there any reason that date feels better than dateTime

It's not a strong opinion, but I just find it nicer and more intuitive. I am also fine with `dateTime`. It is a very minor difference, I agree.

@srindom, do you have any thoughts on this?

adrien2p (approved):

adrien2p:
I like the approach personnaly 👍

packages/core/utils/src/dml/types.ts
olivermrbl: @adrien2p `simpleDate` was maybe a bad example, but it was just to get the point across 😅

olivermrbl (approved):

packages/core/utils/src/dml/types.ts
adrien2p: ahah 😄 no problem I see what you mean, but we still will need to find a name which will be completely different from what users are use too IMO

packages/core/utils/src/dml/modifiers/optional.ts
```diff
@@ -0,0 +1,36 @@
...
+   * A type-only property to infer the JavScript data-type
+   * of the schema property
+   */
+  declare $dataType: T | undefined
+
+  /**
+   * The parent schema on which the nullable modifier is
+   * applied
+   */
+  #schema: SchemaType<T>
```
sradevski, line 15: Should we use `_` as in other places? Same for other private variables

packages/core/utils/src/dml/types.ts
sradevski: I would also go with dateTime, as it is more explicit and obvious what it holds (you don't need to wonder what precision this is for).  Another option is to have `date`, which takes in modifiers such as `time | date | full` or something along those lines.

packages/core/utils/src/dml/types.ts
```diff
@@ -0,0 +1,44 @@
...
+  | "json"
+  | "any"
+
+/**
+ * The meta-data returned by the schema parse method
+ */
+export type SchemaMetaData = {
```
sradevski:
```suggestion
export type SchemaMetadata = {
```

packages/core/utils/src/dml/modifiers/optional.ts
adrien2p: we used `_` because we were using a loose private attribute which is accessible on the prototype outside of the class. With `#` private menber they are not available and you can't conflict with a getter as it is prefixed

packages/core/utils/src/dml/types.ts
olivermrbl: QQ: When you guys say it is more obvious and known to developers, what other usage of `dateTime` are you referring to?

thetutlage: @olivermrbl In general:

Anytime I want to represent `time`, `date` and `datetime` separately. I will use the following naming.

- `time` - To represent time without date
- `date` - To represent date without time
- `dateTime` - To represent date time as a timestamp (precision option can be configured on this )

Yes, usually these terms are connected to a database. But even outside that if I have to talk about these three unit of values, I will use the same naming and won't invent it.

packages/core/utils/src/dml/types.ts
thetutlage: Okay, I have always wrote it as `metaData`. So it should be `metadata` and `SchemaMetadata` right?

packages/core/utils/src/dml/modifiers/optional.ts
thetutlage: Yeah, same as @adrien2p . We don't need convention here, since the field is private natively

packages/core/utils/src/dml/types.ts
adrien2p: On my side I am referring to many languages, database, packages an browser as well

packages/core/utils/src/dml/types.ts
sradevski: Yeah, that's how it is everywhere (and TBH how it's commonly typed, as it is a single word in the dictionary 😄 )

packages/core/utils/src/dml/types.ts
srindom: I don't have strong opinions here - fine with having both date and datetime. 

Would maybe suggest we are opinionated about always including timezone on datetime though?

adrien2p: I believe timezone is always good to have to manage date properly so fine with including it

packages/core/utils/src/dml/types.ts
thetutlage: Good to know 😄

thetutlage:
I have added the initial implementation of entity builder. You must reference this test file to get an idea of how the end-user API looks like. 

https://github.com/medusajs/medusa/blob/f16eedc5239399498f9daf79258b2ef35d323f63/packages/core/utils/src/dml/__tests__/entity_builder.spec.ts

I have changed the relationships API a bit but do not hold any strong opinions for it. So please provide your feedback and help shape the API :)

Also, I would love to get feedback on the approach I have taken to define a type-only representation of a Mikro ORM entity. https://github.com/medusajs/medusa/blob/f16eedc5239399498f9daf79258b2ef35d323f63/packages/core/utils/src/dml/types.ts

CC @olivermrbl @sradevski @adrien2p

olivermrbl:
@thetutlage great work! Overall, LGTM.

Should we use `text()` instead of `string()`? To me, `text` is slightly more user-friendly, and we will be consistent with a few other similar APIs, e.g. Drizzle

We are missing a few modifiers, including the `id()` and indexes – is this something we will add on top in follow-up PRs? That's totally fine with me, but I thought I would highlight in case we missed those

adrien2p:
It looks got to me 💪

thetutlage:
> Should we use text() instead of string()? To me, text is slightly more user-friendly, and we will be consistent with a few other similar APIs, e.g. Drizzle

Yeah, I am cool with both. Just one question, if we decide to generate migrations from this. How will we differentiate with `varchar` and `text`?

> We are missing a few modifiers, including the id() and indexes – is this something we will add on top in follow-up PRs? That's totally fine with me, but I thought I would highlight in case we missed those

Yeah, I have left indexes for the last. Trying to go through the unknown territory first

olivermrbl:
> How will we differentiate with varchar and text?

Hmm, do you think we will need varchar? In the context of Postgres, I imagined text to be a sufficient option. We haven't had the need ourselves yet.

thetutlage:
> Hmm, do you think we will need varchar? In the context of Postgres, don't you think text will suffice? We haven't had the need ourselves yet.

No idea to be honest. I have been super strict with types myself. For example: Properties like `title`, `email` are always `varchar` for me.

adrien2p:
@olivermrbl in some cases the user might want to limit the size of the text and will use varchar to limit the space disk that will be allocated for it, on the other hand text takes more space in the disk because the size is unknow and text allows for unlimited length

adrien2p:
Could one solution be to use for example `text()` and some options such as `{ size?: number }` and if the size is provided then we fallback to varchar?

olivermrbl:
> Could one solution be to use for example text() and some options such as { size?: number } and if the size is provided then we fallback to varchar?

I think something like this is worth considering, but imo. not super important right now. The reason I am pushing `text` is that I just want us to make the API as simple as possible in the first iteration. I am not sure that developers defining models will ever worry about the disk space difference between the two types, which is why I am leaning toward trading off the optionality for simplicity.

thetutlage:
Cool. Let's go with `text` then.

sradevski:
@olivermrbl @adrien2p @thetutlage just FYI, in Postgres `text` is more space-efficient than `varchar`,  because varchar requires fixes space (eg. 255, where you only use 10), but text is dynamic and takes as much space as it actually needs. Being able to put constraints is nice, but I think passing params to `text` is much nicer IMO, so we can do that when we need it.

PR #7646, opened 2024-06-07
feat: Ensure the event group id is either generated or picked up from the context

adrien2p:
**What**
- Pick the context event group id or generate it to be accessible from the workflow and down the module calls

FIXES CORE-2273

### DIFF
--- a/packages/admin-next/dashboard/src/components/data-grid/utils.ts
+++ b/packages/admin-next/dashboard/src/components/data-grid/utils.ts
@@ -97,7 +97,7 @@ export function getFieldsInRange(
 }
 
 export function convertArrayToPrimitive<
-  T extends "boolean" | "number" | "string",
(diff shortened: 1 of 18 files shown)


sradevski (approved):

riqwan (approved):
nice!

packages/core/workflows-sdk/src/helper/workflow-export.ts
```diff
@@ -140,6 +140,7 @@ function createContextualWorkflowRunner<
     }
 
     context.transactionId ??= ulid()
+    context.eventGroupId ??= ulid()
```
riqwan, line 143: should we consider using the same as transactionId?

adrien2p: Not sure 🤔 we can see that later? Its on my todo 😉

PR #7647, opened 2024-06-07
chore: Remove global workflow in favor of workflow manager storage

sradevski:
The global workflow was no longer used, as the same definition was fetched from the workflow manager.


olivermrbl (commented):
No idea if we can do this, but it's always nice to eliminate redundant concepts 😄 

Will let Adrien and Carlos review

adrien2p:
The idea with global workflow is that it would hold the instance of a workflow for which the definition will stay the initial one VS Local workflow which is what we are running in our core and which can be edited on the fly. So I think it would be nice to not remove it now and keep it until we validate the concept

sradevski:
@adrien2p From what I understand, currently the entire definition is stored in the WorkflowManager (and that's what it is used in the code) so I am not sure what benefit we have from keeping yet another concept in the codebase. I'd say if we ever want to go back to it we can easily find the PR where we removed it, and until then it is better to keep the codebase clean. WDYT?

carlos-r-l-rodrigues:
Currently the Workflow Manager is being used as the global workflow but I believe this will change.
I foresee in the near future the need of having 3 distincts workflow definitions.

* Workflow Manager - The immutable initial definition of any workflow (at the moment in memory, but in the future, comming from a database). Very likely managed and editable by an UI.
* Global Workflow - The global definition of the workflow that is editable by external sources, like plugins. When edited, these changes will affect all the executions without modifying the original definition on the manager
* Local Workflow - The instance of a global workflow, that can also be edited, but the changes are only applied to that particular instance.

I am not sure exactly what is the concern of keeping this file in the repo, but I am afraid that this concept can be lost in the future without this reference here.

sradevski:
@carlos-r-l-rodrigues I am just trying to reduce the amount of unused code, that's all. But if you guys want to keep it, it's fine with me, let me close it and we can revisit at another time.

PR #7648, opened 2024-06-07
chore: Ensure the transactionId/eventGroupId are passed down the child workflows

adrien2p:
**What**
- Make eventGroupId also available at the top level of step context
- ensure both transaction id and event group id are shared with child workflows execution

FIXES CORE-2274


sradevski (approved):

riqwan (approved):
nice!

PR #7649, opened 2024-06-07
chore: group & release events for local eventbus

riqwan:
what:

- adds ability to group, release and clear events on local eventbus

RESOLVES CORE-2262

### DIFF
--- a/packages/modules/event-bus-local/src/services/__tests__/event-bus-local.ts
+++ b/packages/modules/event-bus-local/src/services/__tests__/event-bus-local.ts
@@ -24,18 +24,10 @@ describe("LocalEventBusService", () => {
         eventBus = new LocalEventBusService(moduleDeps as any)
       })
 
-      it("should emit an event", () => {
-        eventBus = new LocalEventBusService(
-          moduleDeps,
-          {},
-          {
-            resources: "shared",
-          }
-        )
-
+      it("should emit an event", async () => {
         eventBus.eventEmitter_.emit.mockImplementationOnce((data) => data)
 
-        eventBus.emit("eventName", { hi: "1234" })
+        await eventBus.emit("eventName", { hi: "1234" })
 
         expect(eventBus.eventEmitter_.emit).toHaveBeenCalledTimes(1)
         expect(eventBus.eventEmitter_.emit).toHaveBeenCalledWith("eventName", {
@@ -78,6 +70,7 @@ describe("LocalEventBusService", () => {
 
         jest.clearAllMocks()
 
+        eventBus.eventEmitter_.emit.mockImplementationOnce((data) => data)
         eventBus.emit("test-event", { test: "1234", eventGroupId: "test" })
         eventBus.emit("test-event", { test: "test-1" })
 
@@ -89,7 +82,10 @@ describe("LocalEventBusService", () => {
           expect.objectContaining({ eventName: "test-event" }),
         ])
 
-        eventBus.emit("test-event", { test: "1234", eventGroupId: "test-2" })
+        await eventBus.emit("test-event", {
+          test: "1234",
+          eventGroupId: "test-2",
+        })
 
         expect(eventBus.groupedEventsMap_.get("test-2")).toEqual([
           expect.objectContaining({ eventName: "test-event" }),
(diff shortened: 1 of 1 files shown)


packages/core/utils/src/event-bus/index.ts
```diff
@@ -24,6 +24,15 @@ export abstract class AbstractEventBusModuleService
   abstract emit<T>(data: EventBusTypes.EmitData<T>[]): Promise<void>
   abstract emit<T>(data: EventBusTypes.Message<T>[]): Promise<void>
 
+  abstract stageEvent<TData = unknown>(
```
adrien2p: suggestion: should probably support the same API as emit no?

packages/modules/event-bus-local/src/services/__tests__/event-bus-local.js
```diff
@@ -65,6 +65,124 @@ describe("LocalEventBusService", () => {
...
+
+        eventBus.emit("test-event", { test: "1234", eventGroupId: "test" })
+
+        expect(eventBus.eventEmitter_.emit).not.toHaveBeenCalled()
+        expect(stageEventFn).toHaveBeenCalledTimes(1)
+        expect(stageEventFn).toHaveBeenCalledWith("test", "test-event", {
+          test: "1234",
+        })
+
+        eventBus.emit("test-event", { test: "1234", eventGroupId: "test" })
```
adrien2p: suggestion: can we also emit without group id to ensure the behaviour is fully tested ?

adrien2p (approved):
LGTM, few suggestions still

packages/modules/event-bus-local/src/services/event-bus-local.ts
```diff
@@ -70,8 +74,56 @@ export default class LocalEventBusService extends AbstractEventBusModuleService
...
+      await this.stageOrEmitEvent(event.eventName, data)
+    }
+  }
+
+  // If the data of the event consists of a eventGroupId, we don't emit the event, instead
+  // we add them to a staging queue grouped by the eventGroupId and release them when
+  // explicitly requested.
+  // This is useful in the event of a distributed transaction where you'd want to emit
+  // events only once the transaction ends.
+  async stageOrEmitEvent(
```
carlos-r-l-rodrigues:
```suggestion
  private async stageOrEmitEvent(
```

packages/core/utils/src/event-bus/index.ts
carlos-r-l-rodrigues: I believe this should be private, no?

packages/core/utils/src/event-bus/index.ts
```diff
@@ -24,6 +24,15 @@ export abstract class AbstractEventBusModuleService
...
+    eventGroupId: string,
+    eventName: string,
+    data: TData
+  ): Promise<void>
+
+  abstract releaseStagedEvents(eventGroupId: string): Promise<void>
+  abstract clearStagedEvents(eventGroupId: string): Promise<void>
```
carlos-r-l-rodrigues: Maybe we can find better names here?
I like only `release` and `dismiss`
or `releaseGroupedEvents` and `cancelGroupedEvents`/`dismissGroupedEvents`

carlos-r-l-rodrigues (approved):

packages/core/utils/src/event-bus/index.ts
riqwan: yeah, lets make this private

Discussed shape on chat. 👍🏻

packages/core/utils/src/event-bus/index.ts
riqwan: can used grouped instead, but wrt to a queue, the action words are release and clear right?

sradevski (approved):
So far so good! Talking concretely about bullmq, do we have the possibility to group events like this without too much performance overhead?

riqwan:
@sradevski we're pushing it straight to redis, potentially just passing in a luascript instead of performing the actions here directly

packages/core/utils/src/event-bus/index.ts
riqwan: resolved over chat - decision to use group, but the action words the same 💪🏻

PR #7651, opened 2024-06-07
feat: Add support for scheduled workflows

sradevski:
We still need to:
But wanted to open the PR for early feedback on the approach

### DIFF
--- a/packages/core/orchestration/src/transaction/datastore/abstract-storage.ts
+++ b/packages/core/orchestration/src/transaction/datastore/abstract-storage.ts
@@ -6,14 +6,14 @@ import { TransactionStep } from "../transaction-step"
 import { SchedulerOptions } from "../types"
 
 export interface IDistributedSchedulerStorage {
-  scheduleJob(
+  schedule(
     jobDefinition: string | { jobId: string },
     schedulerOptions: SchedulerOptions
   ): Promise<void>
 
-  cancelJob(jobId: string): Promise<void>
+  remove(jobId: string): Promise<void>
 
-  cancelAllJobs(): Promise<void>
+  removeAll(): Promise<void>
 }
 
 export interface IDistributedTransactionStorage {
@@ -55,19 +55,19 @@ export abstract class DistributedSchedulerStorage
     /* noop */
   }
 
-  async scheduleJob(
+  async schedule(
     jobDefinition: string | { jobId: string },
     schedulerOptions: SchedulerOptions
   ): Promise<void> {
-    throw new Error("Method 'scheduleExecution' not implemented.")
+    throw new Error("Method 'schedule' not implemented.")
   }
 
-  async cancelJob(jobId: string): Promise<void> {
-    throw new Error("Method 'cancelJob' not implemented.")
+  async remove(jobId: string): Promise<void> {
+    throw new Error("Method 'remove' not implemented.")
   }
 
-  async cancelAllJobs(): Promise<void> {
-    throw new Error("Method 'cancelAllJobs' not implemented.")
+  async removeAll(): Promise<void> {
+    throw new Error("Method 'removeAll' not implemented.")
   }
 }
 
--- a/packages/core/orchestration/src/workflow/scheduler.ts
+++ b/packages/core/orchestration/src/workflow/scheduler.ts
@@ -3,9 +3,9 @@ import { IDistributedSchedulerStorage, SchedulerOptions } from "../transaction"
 import { WorkflowDefinition } from "./workflow-manager"
 
 export class WorkflowScheduler {
-  private static keyValueStore: IDistributedSchedulerStorage
+  private static storage: IDistributedSchedulerStorage
   public static setStorage(storage: IDistributedSchedulerStorage) {
-    this.keyValueStore = storage
+    this.storage = storage
   }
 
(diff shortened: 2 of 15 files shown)


packages/modules/workflow-engine-redis/src/utils/workflow-orchestrator-storage.ts
```diff
@@ -108,6 +119,17 @@ export class RedisDistributedTransactionStorage extends DistributedTransactionSt
     })
   }
 
+  private async executeScheduledJob(
+    jobId: string,
+    schedulerOptions: SchedulerOptions
+  ) {
+    // TODO: In the case of concurrency being forbidden, we want to generate a predictable transaction ID and rely on the idempotency
```
carlos-r-l-rodrigues, line 126: I have plans to support concurrency control in the workflow, so I believe we will be able to reuse it here.
the idea is to support max concurrent executions by workflow id

carlos-r-l-rodrigues (commented):
This looks good to me.

packages/core/orchestration/src/transaction/datastore/abstract-storage.ts
```diff
@@ -36,6 +48,29 @@ export interface IDistributedTransactionStorage {
...
+{
+  constructor() {
+    /* noop */
+  }
+
+  async scheduleJob(
+    jobDefinition: string | { jobId: string },
+    schedulerOptions: SchedulerOptions
+  ): Promise<void> {
+    throw new Error("Method 'scheduleExecution' not implemented.")
```
olivermrbl:
```suggestion
    throw new Error("Method 'scheduleJob' not implemented.")
```

packages/modules/workflow-engine-redis/src/utils/workflow-orchestrator-storage.ts
sradevski: Yeah I am not planning on adding anything yet, just added it here as something we need to consider (as oftentimes, you don't want more than 1 instance of a cronjob running at the same time)

olivermrbl (approved):
LGTM

packages/core/orchestration/src/workflow/scheduler.ts
```diff
@@ -0,0 +1,45 @@
...
+            numberOfExecutions: schedule.numberOfExecutions,
+          }
+
+    await WorkflowScheduler.keyValueStore.scheduleJob(
+      workflow.id,
+      normalizedSchedule
+    )
+  }
+
+  public async clearWorkflow(workflow: WorkflowDefinition) {
```
riqwan, line 35: sugg: this can just accept the id directly right?

packages/modules/workflow-engine-redis/src/utils/workflow-orchestrator-storage.ts
```diff
@@ -290,4 +312,42 @@ export class RedisDistributedTransactionStorage extends DistributedTransactionSt
...
+  async scheduleJob(
+    jobDefinition: string | { jobId: string },
+    schedulerOptions: SchedulerOptions
+  ): Promise<void> {
+    const jobId =
+      typeof jobDefinition === "string" ? jobDefinition : jobDefinition.jobId
+
+    // In order to ensure that the schedule configuration is always up to date, we first cancel an existing job, if there was one
+    // any only then we add the new one.
+    await this.cancelJob(jobId)
```
riqwan: q: Whats the behavior here when we try to cancel a job thats already running? Does the workflow get gracefully killed or is it a complete rug pull?

riqwan (approved):
one question, otherwise lgtm!

packages/modules/workflow-engine-redis/src/utils/workflow-orchestrator-storage.ts
sradevski: @riqwan it is not documented, but based on how their repeatable jobs work it should not kill any running job, it will just remove the scheduling. That being said, I guess `cancel` isn't the best verb to use here. I'll need to add a test to be 100% sure though.

packages/core/orchestration/src/workflow/scheduler.ts
sradevski: What I was thinking is that maybe we want to have some options set on the workflow in terms of how clearing works, so I opted for passing the entire object. I am happy to change it to id only as well, I think for now either is fine

packages/modules/workflow-engine-redis/src/utils/workflow-orchestrator-storage.ts
```diff
@@ -290,4 +312,42 @@ export class RedisDistributedTransactionStorage extends DistributedTransactionSt
...
+        jobId: `${JobType.SCHEDULE}_${jobId}`,
+      }
+    )
+  }
+
+  async cancelJob(jobId: string): Promise<void> {
+    await this.queue.removeRepeatableByKey(`${JobType.SCHEDULE}_${jobId}`)
+  }
+
+  async cancelAllJobs(): Promise<void> {
```
olivermrbl: Maybe we can have `remove` and `removeAll` instead

adrien2p (approved):
It looks pretty clean to me 💪

olivermrbl (approved):
Nice!

PR #7652, opened 2024-06-07
chore(create-medusa-app): Add tests

shahednasser:
- Add a `--quiet` option to `create-medusa-app` that doesn't show prompts and uses default values.
- Add tests for `create-medusa-app`
- General fixes

Note: the tests' runtime is relevantly long since each time a project has to be created, dependencies installed, etc... I couldn't think of a workaround so if anyone has any ideas I'm all ears

### DIFF
--- a/packages/cli/create-medusa-app/src/__tests__/create-command.spec.ts
+++ b/packages/cli/create-medusa-app/src/__tests__/create-command.spec.ts
@@ -1,35 +1,65 @@
 import execa from "execa"
 import path from "path"
-import { rm, readFile, stat } from "fs/promises"
+import { 
+  rm, 
+  readFile, 
+  stat, 
+  readdir,
+  mkdir
+} from "fs/promises"
 import pg from "pg"
 const { Client } = pg
 
-const DB_REGEX = /DATABASE_URL=(?<url>.+?)/
+const DB_REGEX = /DATABASE_URL=(?<url>.+)/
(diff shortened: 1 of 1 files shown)


adrien2p (approved):
I believe it is good to have the integration tests, but I would have probably started with the unit tests of each function it is relying on, but nice work 💪

shahednasser:
@adrien2p I was gonna do that, but in the case of the `create-medusa-app` I believe it's more valuable to do them as integration tests, as it's important to test that the entire setup works as expected with the different conditions (for example, just testing that the database is created successfully is less valuable than making sure that the connection in Medusa to that database + migrations run as expected as well).

However, if you believe there's an argument for doing unit tests instead of integration tests (or both) I'm happy to revisit 😄

adrien2p:
No you are right, integration tests should suffice, but we will need to go further, but lets keep that for later

fPolic (approved):

shahednasser:
@adrien2p seems like the integration test won't work after all, since the tests take forever to finish 😅 I'll open another PR with unit tests instead as originally intended.

adrien2p:
It was part of my concerns but i thought it did work 🤣 let test the function individually, if they behaviour is tested then is should work overall and it should be enough ✌️

shahednasser:
It worked locally, but kept hanging in the github action 😅 my main concern from the tests was to ensure the entire setup works, but let's go for unit tests like you said for now

PR #7654, opened 2024-06-07
chore: Provide an easy way to infer joiner config if needed

adrien2p:
**What**
If not configured manually, the util provide an easy way to get a joiner config up and running


adrien2p:
@olivermrbl what do you think of this approach as a first step?

packages/modules/fulfillment/src/joiner-config.ts
```diff
@@ -11,84 +9,24 @@ import {
...
-
-const entityLinkableKeysMap: MapToConfig = {}
-Object.entries(LinkableKeys).forEach(([key, value]) => {
-  entityLinkableKeysMap[value] ??= []
-  entityLinkableKeysMap[value].push({
-    mapTo: key,
-    valueFrom: key.split("_").pop()!,
-  })
+export const joinerConfig = defineJoinerConfig(Modules.FULFILLMENT, {
+  publicEntityObjects: [
```
olivermrbl: thought: What if this was opt-out instead?

By default, we assume that all models in `/models` should be part of the joiner config. If not, you can add the `internal` ones here.

That way, we don't need to export the joiner config in the majority of use cases right? We can use this util under the hood, and only if you need to opt-out, you define the config here.

olivermrbl: Maybe a good way of gauging whether this makes sense or not is to check how many modules we have that have more internal models than public ones. I bet it's few to none.

adrien2p: I see, my thinking is that since the config is supposed to be generated for external consumption, specifying what is internal seems a bit going the other way around but i see your point. 

Though, as you mention it now We could go with a first approach with auto generating the joiner config and explaining that if the default is not what they need then they can completely opt out or export their own config either manually created or using the util. Also, i still need to work the util a bit to give more manual control on the alias config for example. 

Though, we have to take into consideration that you might not have a model per say but just providing some methods which you might want to make available externally to the remote query. What do you see for those cases @olivermrbl maybe one approach would be all models by default and if you configure it yourself then you specify all the models manually as opting out would't work in all cases as it is more virtual

It is still a poc and ill changed it on the side later ✌️

olivermrbl: > auto generating the joiner config and explaining that if the default is not what they need then they can completely opt out or export their own config either manually created or using the util

Yeah exactly. This is what I had in mind. I imagine in most cases the default autogenerated one would be enough.

adrien2p: @olivermrbl i have updated my message 😅 also i dont think we can take our usage as the majority, for example in the medusa eat repo it is the opposite, less models are public

olivermrbl: > providing some methods which you might want to make available externally to the remote query

What does this look like? I don't think I've seen this use case until now. Do we have that?

adrien2p: We have that when it comes to third party yes like the agency was asking for for example, in that case you dont have any models as proper model but more the concept of entry points representing the public data objects available, again database model and public object representation can differ

adrien2p: It can be a cms, pim, a translation tool or other thing

adrien2p: That is also why the entity object i am expecting is just requesting a name as it would be the virtual name représention of that object that they want to expose

adrien2p: An example could be the following

alias: { name: [translation, transalations], args: { entity: 'translation', methodSuffix: 'Translations' }

Here the translation object is a representation of what would be returned by a third party but i dont have a model in the module. 

In that example there is nothing to opt out but you would still want to benefit from the util and opt in for it to not have to write everything manually

adrien2p: In the above example that is what the gql was describing in our schemas, what object are publically available and provided by a module without necessarily relying on a model as the db model and the returned data are two different thing. But in most cases i agree that they will be backed by a model

olivermrbl: I am not sure I understand how that changes the joiner config. I thought the purpose of the joiner config was to define what keys of the models (regardless of source MikroORM / Stripe / CMS) you can link to?

Using your example, you would still like to have `translation_id` be a linkable key, even though translation is a "model" with data sourced from a third-party system.

The `translation` "model" would still live in the `/models` folder like any other model.

That is at least how I imagined the interplay between our DML and the joiner config, but maybe that's not possible.

olivermrbl: Let's maybe huddle about this monday 😅

adrien2p: Yes for sure, in the example above there is no model. But it will be easier to explaon in a huddle 😅

adrien2p: @olivermrbl I ve done some improvements but we will still need to figure out some things, same for the DML, we can discuss that later if you want

PR #7655, opened 2024-06-08
chore: ability to group events on redis event bus

riqwan:
what:

- update event bus redis to stage events before emitting
- use a singular shape to emit an event

coming up:

- use luascript with ioredis instead

RESOLVES CORE-2263

### DIFF
--- a/packages/core/utils/src/event-bus/__tests__/message-aggregator.spec.ts
+++ b/packages/core/utils/src/event-bus/__tests__/message-aggregator.spec.ts
@@ -59,10 +59,10 @@ describe("MessageAggregator", function () {
     })
 
     const format = {
-      groupBy: ["eventName", "body.metadata.object", "body.metadata.action"],
+      groupBy: ["eventName", "metadata.object", "metadata.action"],
       sortBy: {
-        "body.metadata.object": ["ProductType", "ProductVariant", "Product"],
-        "body.data.id": "asc",
+        "metadata.object": ["ProductType", "ProductVariant", "Product"],
+        "data.id": "asc",
       },
     }
 


packages/core/utils/src/modules-sdk/abstract-module-service-factory.ts
```diff
@@ -501,7 +503,12 @@ export function abstractModuleServiceFactory<
           await this.eventBusModuleService_?.emit(
             softDeletedEntities.map(({ id }) => ({
               eventName: `${kebabCase(model.name)}.deleted`,
-              data: { id },
+              body: {
+                id,
+                // TODO: fulfil these
```
olivermrbl: Let's create a ticket for this if we haven't already 🙏

packages/modules/event-bus-redis/src/services/event-bus-redis.ts
```diff
@@ -71,78 +81,114 @@ export default class RedisEventBusService extends AbstractEventBusModuleService
...
+      const emitData = this.buildEvents(eventsToEmit, options)
+
+      promises.push(this.queue_.addBulk(emitData))
+    }
+
+    for (const [groupId, events] of groupEventsMap.entries()) {
+      const eventsData = this.buildEvents(events, options)
+
+      promises.push(this.groupEvents(groupId, eventsData))
+    }
```
olivermrbl: suggestion(non-blocking): Wrap in `if (groupEventsMap.size())`

packages/modules/event-bus-redis/src/services/event-bus-redis.ts
```diff
@@ -71,78 +81,114 @@ export default class RedisEventBusService extends AbstractEventBusModuleService
...
-  async releaseGroupedEvents(eventGroupId: string) {}
-  async clearGroupedEvents(eventGroupId: string) {}
+  private async groupEvents<T = unknown>(
+    eventGroupId: string,
+    events: IORedisEventType<T>[]
+  ) {
+    await this.eventBusRedisConnection_.rpush(
+      `staging:${eventGroupId}`,
+      ...events.map((event) => JSON.stringify(event))
+    )
```
olivermrbl, line 187: thought: Wondering if we should have a TTL on this

olivermrbl (approved):
Very nice @riqwan, no major comments from my side! 💪

packages/core/utils/src/modules-sdk/abstract-module-service-factory.ts
adrien2p: this is part of the other events project to find a way to refactor this 👍

packages/modules/event-bus-local/src/services/__tests__/event-bus-local.ts
```diff
@@ -14,156 +14,210 @@ const moduleDeps = {
 }
 
 describe("LocalEventBusService", () => {
-  let eventBus
+  let eventBus: LocalEventBusService
```
adrien2p, line 17: nit: we can refactor this suite to use the moduleTestRunner wdyt?

packages/modules/event-bus-redis/src/services/__tests__/event-bus.ts
```diff
@@ -0,0 +1,511 @@
...
+  logger: loggerMock,
+  eventBusRedisConnection: redisMock,
+}
+
+describe("RedisEventBusService", () => {
+  let eventBus: RedisEventBusService
+  let queue
+  let redis
+
+  describe("constructor", () => {
```
adrien2p, line 43: suggestion: we can use the module test runner wdyt?

packages/modules/event-bus-redis/src/services/event-bus-redis.ts
```diff
@@ -1,15 +1,25 @@
...
+import { BullJob, EventBusRedisModuleOptions } from "../types"
 
 type InjectedDependencies = {
   logger: Logger
   eventBusRedisConnection: Redis
 }
 
+type IORedisEventType<T = unknown> = {
+  name: string
+  data: Omit<Message<T>, "body"> & { data: Message<T>["body"]["data"] }
```
adrien2p: Q: are we loosing the metadata as part of the emitted events?

packages/modules/event-bus-redis/src/services/event-bus-redis.ts
```diff
@@ -71,78 +81,114 @@ export default class RedisEventBusService extends AbstractEventBusModuleService
...
+    const promises: Promise<unknown>[] = []
+
+    if (eventsToEmit.length) {
+      const emitData = this.buildEvents(eventsToEmit, options)
+
+      promises.push(this.queue_.addBulk(emitData))
+    }
+
+    for (const [groupId, events] of groupEventsMap.entries()) {
+      const eventsData = this.buildEvents(events, options)
```
adrien2p, line 164: suggestion: if (!events.length) continue

packages/modules/event-bus-redis/src/services/event-bus-redis.ts
adrien2p: I think it is good to not have this check as it wont loop if there is nothing

packages/modules/event-bus-redis/src/services/event-bus-redis.ts
```diff
@@ -71,78 +81,114 @@ export default class RedisEventBusService extends AbstractEventBusModuleService
...
+      ...options,
+    }
+
+    return eventsData.map((eventData) => {
+      return {
+        name: eventData.eventName,
+        data: {
+          eventName: eventData.eventName,
+          data: eventData.body.data,
+        },
```
adrien2p: suggestion: Looks like we are loosing the metadata and we expect the subscriber to receive those metadata

packages/modules/event-bus-redis/src/services/event-bus-redis.ts
```diff
@@ -71,78 +81,114 @@ export default class RedisEventBusService extends AbstractEventBusModuleService
...
+  private async getGroupedEvents(eventGroupId: string) {
+    return await this.eventBusRedisConnection_
+      .lrange(`staging:${eventGroupId}`, 0, -1)
+      .then((result) => {
+        return result.map((jsonString) => JSON.parse(jsonString))
+      })
+  }
+
+  async releaseGroupedEvents(eventGroupId: string) {
+    const groupedEvents: IORedisEventType[] = await this.getGroupedEvents(
```
adrien2p: suggestion: you could type the return type of getGroupedEvents and not have to type any const relying on that result

packages/modules/event-bus-redis/src/services/event-bus-redis.ts
```diff
@@ -71,78 +81,114 @@ export default class RedisEventBusService extends AbstractEventBusModuleService
...
+      eventGroupId
+    )
+
+    await this.queue_.addBulk(groupedEvents)
```
adrien2p, line 203: suggestion: what I thought we could do is the following,
we grab the grouped events, we make the event group id of their metadata undefined and then we call the emit, such that we always rely on the same behaviour no matter what once we are releasing, wdyt?

packages/modules/event-bus-redis/src/services/event-bus-redis.ts
```diff
@@ -71,78 +81,114 @@ export default class RedisEventBusService extends AbstractEventBusModuleService
...
+
+    await this.clearGroupedEvents(eventGroupId)
+  }
+
+  async clearGroupedEvents(eventGroupId: string) {
+    if (eventGroupId) {
+      await this.eventBusRedisConnection_.del(`staging:${eventGroupId}`)
+    }
```
adrien2p:
```suggestion
    if (!eventGroupId) {
      return
    }
    await this.eventBusRedisConnection_.del(`staging:${eventGroupId}`)
```

packages/modules/event-bus-redis/src/services/event-bus-redis.ts
olivermrbl: I actually noticed this too but I thought it was intentional. The case is the same for the local event bus.

adrien2p: I don't think the metadata should be removed as it is metadata information that we send to the user, but it could also be a place where the user add some metadata such as timestamp of any other information, it could be open ended

packages/core/utils/src/modules-sdk/abstract-module-service-factory.ts
```diff
@@ -466,9 +466,11 @@ export function abstractModuleServiceFactory<
           await this.eventBusModuleService_?.emit(
             primaryKeyValues_.map((primaryKeyValue) => ({
               eventName: `${kebabCase(model.name)}.deleted`,
-              data: isString(primaryKeyValue)
-                ? { id: primaryKeyValue }
-                : primaryKeyValue,
+              body: {
```
sradevski: Why not have `data` one level higher as before? It feels like unnecessary nesting IMO, since `data` already implies `body` to me.

packages/modules/event-bus-redis/src/services/event-bus-redis.ts
sradevski: At the very least maybe an overview in the admin (similar to workflows) that you can action.

packages/modules/event-bus-redis/src/services/event-bus-redis.ts
```diff
@@ -71,78 +81,114 @@ export default class RedisEventBusService extends AbstractEventBusModuleService
...
+  }
+
+  async releaseGroupedEvents(eventGroupId: string) {
+    const groupedEvents: IORedisEventType[] = await this.getGroupedEvents(
+      eventGroupId
+    )
+
+    await this.queue_.addBulk(groupedEvents)
+
+    await this.clearGroupedEvents(eventGroupId)
```
sradevski, line 205: We might need a lock in this method to avoid concurrency. Also, we will probably want to have a flag that is set before we start adding new events to the main queue, as we could easily get the events, emit them, but then the clearing fails, which will probably result in duplicate events.

sradevski (commented):
Nice stuff!

packages/modules/event-bus-redis/src/services/event-bus-redis.ts
adrien2p: maybe we should grab, clear and emit instead of grab, emit and clear. could this work?

carlos-r-l-rodrigues: If we can have all these operations in a Lua script we reduce a lot the points of failure. I believe for this PR this is ok, but ideally we try to do all that in one single call and hope Redis won't fail in the middle of the operation. That will be the single point of failure.
We just need a script that sets all the messages in the format bullmq is expecting.

adrien2p: we have indeed a separete ticket for that

sradevski: grab, clear, and emit is a bit worse, since you might lose messages (at least once delivery is almost always better than at most once). Doing it all in Redis would increase the reliability for sure, I am just thinking if it will cause issues with eg. bullmq changing something under the surface (but not the public API) and our solution breaking. But we can definitely discuss it, I just wanted to raise it here so at the very least we add a TODO or a ticket.

carlos-r-l-rodrigues: we probably need to write a test to read all the keys created by bullmq, at least if we bump the version and it has changed we know.

packages/modules/event-bus-redis/src/services/event-bus-redis.ts
riqwan: @sradevski can you elaborate on this ^ one? Is this related to the TTL? Or the ability to retry/resend an event?

packages/modules/event-bus-redis/src/services/event-bus-redis.ts
riqwan: I created a separate ticket for the lua scripts because I think that will open up some questions of its own :D 

Wrt to the fail in clearing events, i think we can remove clearing events from this function and pass that responsibility onto the consumer to handle the failure scenario. If the consumer ends up not clearing them, the default TTL should clear it up eventually

wdyt?

carlos-r-l-rodrigues: We can just ignore or add a log if clearing events fail. THe worst case scenario will be garbage left in Redis.
BullMQ won't manage this separeted queue, so nothing will be consumed.

sradevski: @riqwan I think just noting it down in comments as a possibility is enough for now. 

The main problem is you don't know if the events were emitted before clearing or not, so you don't know if you should retry the operation. It doesn't really matter who does the clearing, the issue is if a machine crashes, you don't know where to proceed from without a flag (similar to workflows).

packages/modules/event-bus-redis/src/services/event-bus-redis.ts
sradevski: @riqwan I was thinking it would be nice to be able to CRUD event groups from the UI, so you can see what's staged, what is maybe stuck, etc. It is orthogonal to TTLs, I was just throwing it out as an idea :)

adrien2p: I am not sure the ui should be handled by us right? They already have tools for redis and bull to monitor and see this directly

sradevski: I was just thinking having this embedded experience similar to the workflows UI would be nice, esp. when it comes to grouped events, but of course it's not something that is pressing to implement

olivermrbl: Somewhat related, I have always found Stripe's Events API (and the related UI) very nice and super useful for debugging. A feature like this would probably need its own module, and the events persisted to the DB, so a bit different from this. But maybe a combination could make sense in the future 😄

PR #7661, opened 2024-06-10
fix(medusa): TX isolation level in batch job subscriber

### LINKED ISSUE #7611: Inconsistent orders when exporting
## Bug report

When exporting orders, we sometimes get inconsistent export results (duplicate orders,..). 
Looking at the export job, it seems to use "REPEATABLE READ" isolation level which looks good

image

But the issue seems to be in this code section:

image

Since this code already opens a transaction, so atomicPhase reuses this transaction, but this transaction is "Read committed".
So during export, if someone creates new orders matching the query condition of the export, the offset is messed up by these orders, making the results inconsistent.

We are able to reproduce this in this way, then we try openning the transaction in `BatchJobSubscriber.processBatchJob` as repeatable read to confirm if this is the root cause. That trial fix seems to work.

### System information

Medusa version (including plugins): 1.17.4
Node.js version: 18
Database: Postgres

olivermrbl:
Fixes #7611

What 
- Change isolation level on tx initiated by the batch job processing subscriber to "REPEATABLE READ"

Why
- Currently, the subscriber initiates a tx with the default isolation level "READ COMMITTED", which is reused by `atomicPhase` in the batch job strategy, overwriting the "REPEATABLE READ" in the batch job `processJob` method

### DIFF
--- a/packages/medusa/src/subscribers/batch-job.ts
+++ b/packages/medusa/src/subscribers/batch-job.ts
@@ -63,7 +63,7 @@ class BatchJobSubscriber {
 
   processBatchJob = async (data): Promise<void> => {
     try {
-      await this.manager_.transaction(async (manager) => {
+      await this.manager_.transaction("REPEATABLE READ", async (manager) => {
         const batchJobServiceTx = this.batchJobService_.withTransaction(manager)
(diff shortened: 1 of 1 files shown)


adrien2p (approved):
💪

PR #7663, opened 2024-06-10
Clean up product typings

sradevski:
There are a lot of issues in the admin after applying the correct typings, but fixing those should be done gradually, it's better to keep it out of this PR


packages/admin-next/dashboard/src/hooks/api/product-types.tsx
```diff
@@ -10,7 +10,12 @@ export const useProductType = (
   id: string,
   query?: Record<string, any>,
   options?: Omit<
-    UseQueryOptions<ProductTypeRes, Error, ProductTypeRes, QueryKey>,
+    UseQueryOptions<
+      { product_type: HttpTypes.AdminProductType },
```
olivermrbl, line 14: Should use dedicated response type: `HttpTypes.AdminProductTypeResponse`

packages/admin-next/dashboard/src/hooks/api/product-types.tsx
```diff
@@ -26,7 +31,12 @@ export const useProductType = (
 export const useProductTypes = (
   query?: Record<string, any>,
   options?: Omit<
-    UseQueryOptions<ProductTypeListRes, Error, ProductTypeListRes, QueryKey>,
+    UseQueryOptions<
+      { product_types: HttpTypes.AdminProductType[] },
```
olivermrbl, line 35: Should use dedicated response type: `HttpTypes.AdminProductTypeListResponse`

packages/admin-next/dashboard/src/lib/client/product-types.ts
```diff
@@ -1,12 +1,18 @@
-import { ProductTypeListRes, ProductTypeRes } from "../../types/api-responses"
+import { HttpTypes } from "@medusajs/types"
```
olivermrbl, line 1: question: Can we remove this file now?

packages/medusa/src/api/admin/products/helpers.ts
```diff
@@ -70,26 +76,23 @@ export const remapVariantResponse = (variant: ProductVariantDTO) => {
...
-    entryPoint: "product",
-    variables: {
-      filters: { id: productId },
-    },
-    fields: remapKeysForProduct(fields ?? []),
-  })
-
-  const products = await remoteQuery(queryObject)
-  return products[0]
+  return refetchEntity(
```
olivermrbl:
```suggestion
  return await refetchEntity(
```

packages/medusa/src/api/admin/products/[id]/variants/batch/route.ts
```diff
@@ -3,25 +3,16 @@ import {
...
-    >
-  >,
-  res: MedusaResponse
+  req: AuthenticatedMedusaRequest<HttpTypes.AdminBatchProductVariantRequest>,
+  res: MedusaResponse<HttpTypes.AdminBatchProductVariantResponse>
 ) => {
   const productId = req.params.id
 
-  const normalizedInput = {
+  const normalizedInput: any = {
```
Comment (olivermrbl) [inline packages/medusa/src/api/admin/products/[id]/variants/batch/route.ts]: I guess we should be able to type this no?

packages/core/js-sdk/src/admin/product.ts
```diff
@@ -8,12 +8,27 @@ export class Product {
     this.client = client
```
olivermrbl, line 8: Some client calls are missing `return await`

olivermrbl (approved):
Overall LGTM, nice work. Will let @kasperkristensen glance over it too

packages/admin-next/dashboard/src/lib/client/product-types.ts
sradevski: @olivermrbl I only removed the product type typings, but nothing else around product types is done currently. I can tackle it during downtime, but I thought it's too much for this PR.

packages/medusa/src/api/admin/products/helpers.ts
sradevski: It's not really needed since the caller to refetchProduct will call await, but I can actually remove the entire function since. I use refetchEntity now

packages/admin-next/dashboard/src/lib/client/product-types.ts
olivermrbl: Ah got it, makes sense.

packages/medusa/src/api/admin/products/[id]/variants/batch/route.ts
Comment (sradevski) [reply packages/medusa/src/api/admin/products/[id]/variants/batch/route.ts]: Yes I forgot it 👍

PR #7664, opened 2024-06-10
feat: workflows release events

adrien2p:
**What**
- release grouped event on workflow finish
- update common step release events

**NOTE**
rely on https://github.com/medusajs/medusa/pull/7655

FIXES CORE-2264
FIXES CORE-2267

### DIFF
--- a/packages/medusa/src/loaders/__fixtures__/mock-scheduler-storage.ts
+++ b/packages/medusa/src/loaders/__fixtures__/mock-scheduler-storage.ts
@@ -1,0 +1,21 @@
+import {
+  IDistributedSchedulerStorage,
+  SchedulerOptions,
+} from "@medusajs/orchestration"
+
+export class MockSchedulerStorage implements IDistributedSchedulerStorage {
+  async schedule(
+    jobDefinition: string | { jobId: string },
+    schedulerOptions: SchedulerOptions
+  ): Promise<void> {
+    return Promise.resolve()
+  }
+
+  async remove(jobId: string): Promise<void> {
+    return Promise.resolve()
+  }
+
+  async removeAll(): Promise<void> {
+    return Promise.resolve()
+  }
+}
--- a/packages/medusa/src/loaders/__fixtures__/plugin/jobs/order-summary.ts
+++ b/packages/medusa/src/loaders/__fixtures__/plugin/jobs/order-summary.ts
@@ -1,0 +1,11 @@
+import { MedusaContainer } from "@medusajs/types"
+
+export default async function handler(container: MedusaContainer) {
+  console.log(`You have received 5 orders today`)
+}
+
+export const config = {
+  name: "summarize-orders-job",
+  schedule: "* * * * * *",
+  numberOfExecutions: 2,
+}
--- a/packages/medusa/src/loaders/__tests__/register-jobs.spec.ts
+++ b/packages/medusa/src/loaders/__tests__/register-jobs.spec.ts
@@ -1,0 +1,19 @@
+import path from "path"
+import { registerJobs } from "../helpers/register-jobs"
+import { WorkflowManager, WorkflowScheduler } from "@medusajs/orchestration"
+import { MockSchedulerStorage } from "../__fixtures__/mock-scheduler-storage"
+
+describe("register jobs", () => {
+  WorkflowScheduler.setStorage(new MockSchedulerStorage())
+  it("registers jobs from plugins", async () => {
+    await registerJobs([
+      { resolve: path.join(__dirname, "../__fixtures__/plugin") },
+    ])
+    const workflow = WorkflowManager.getWorkflow("summarize-orders-job")
+    expect(workflow).toBeDefined()
+    expect(workflow?.options.schedule).toEqual({
+      cron: "* * * * * *",
+      numberOfExecutions: 2,
+    })
+  })
+})
--- a/packages/medusa/src/loaders/helpers/register-jobs.ts
+++ b/packages/medusa/src/loaders/helpers/register-jobs.ts
@@ -1,0 +1,92 @@
+import {
+  StepResponse,
+  createStep,
+  createWorkflow,
+} from "@medusajs/workflows-sdk"
+import { glob } from "glob"
+import logger from "../logger"
+import { MedusaError } from "@medusajs/utils"
+
+export const registerJobs = async (plugins) => {
+  await Promise.all(
+    plugins.map(async (pluginDetails) => {
+      const files = glob.sync(
+        `${pluginDetails.resolve}/jobs/*.{ts,js,mjs,mts}`,
+        {
+          ignore: ["**/*.d.ts", "**/*.map"],
+        }
+      )
+
+      logger.debug(
+        `Registering ${files.length} jobs from ${pluginDetails.resolve}`
+      )
+
+      const jobs = await Promise.all(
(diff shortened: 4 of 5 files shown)


packages/core/workflows-sdk/src/helper/workflow-export.ts
```diff
@@ -448,3 +465,38 @@ export const exportWorkflow = <TData = unknown, TResult = unknown>(
   MedusaWorkflow.registerWorkflow(workflowId, exportedWorkflow)
   return exportedWorkflow as MainExportedWorkflow<TData, TResult>
 }
+
+function attachOnFinishReleaseEvents(
```
olivermrbl, line 478: question: Just to be sure I understand, this means that releasing grouped events happens separately from the workflow execution, i.e. in case emitting the grouped events fails, the compensation flow is **not** run. Is that correct?

adrien2p: Yes thats right, do you think i should instead create a step instead of listening to finish ? Basically applying the release step at the end of the workflow ?

Cc @carlos-r-l-rodrigues

olivermrbl: > do you think i should instead create a step instead of listening to finish ?

This is what I had in mind. This is based on the assumption that most times - if not all - the event emission done in the workflow is critical to the operation performed, so in case the emission fails, I would expect it to be rolled back and we can try again.

carlos-r-l-rodrigues: Either we have it as a separeted step, or we explictly cancel the flow with `transaction.cancel()` if the emit fails.
It depends if we find a problem to display this new magic step in the UI for example.

adrien2p: what do you think would be the most appropriate guys?

olivermrbl: If we can identify the final injected release events step, so we can hide it in the UI (if we want to), then I am leaning towards that approach. The reasoning here is that then we rely on an existing "native" workflow mechanism, instead of having an additional, separate path that needs testing and maintenance. 

Wdyt Carlos?

carlos-r-l-rodrigues: I prefer to not have the magic step, unless it is explicity added. Otherwise we also need some other complex changes to make sure this will be the last step even if the workflow is edited.
I would just handle the emitGroupedEvents and add cancel the transaction if that fails, and add it to the errors returned.

olivermrbl: Ah OK, so we can still cancel the workflow in the `onFinish`? I thought it was after the execution was done.

adrien2p: I ll go with the cancel guys then. First I need to finish my other future pr that I am on at the moment for another part of the group id management 😂

adrien2p:
@olivermrbl could you re review it please

olivermrbl:
Don't we need the `cancel` functionality or will that be added later @adrien2p ?

adrien2p:
sorry @olivermrbl I might have lost it as I ve done it yesterday the cancel

packages/core/workflows-sdk/src/helper/workflow-export.ts
```diff
@@ -448,3 +465,38 @@ export const exportWorkflow = <TData = unknown, TResult = unknown>(
...
+    result?: unknown
+    errors?: unknown[]
+  }) => {
+    await onFinish?.(args)
+
+    const { transaction } = args
+    const failedStatus = [TransactionState.FAILED, TransactionState.REVERTED]
+
+    if (failedStatus.includes(transaction.getState())) {
+      return
```
sradevski: Here we should clear the staged events right?

adrien2p: already pushed probably while you written this comment 😂 @olivermrbl fortunately I ve retrieve what I ve done in my local build 😂

packages/core/workflows-sdk/src/helper/workflow-export.ts
```diff
@@ -448,3 +461,43 @@ export const exportWorkflow = <TData = unknown, TResult = unknown>(
...
+
+    if (!eventBusService || !eventGroupId) {
+      return
+    }
+
+    const { transaction } = args
+    const failedStatus = [TransactionState.FAILED, TransactionState.REVERTED]
+
+    if (failedStatus.includes(transaction.getState())) {
+      return await eventBusService.clearGroupedEvents(eventGroupId)
```
riqwan:
```suggestion
      try {
        await eventBusService.clearGroupedEvents(eventGroupId)
      } catch (e) {
        console.warn(`Failed to clear events for eventGroupId - ${eventGroupId}`)
      } finally {
        return
      }
```

packages/core/workflows-sdk/src/helper/workflow-export.ts
```diff
@@ -448,3 +461,43 @@ export const exportWorkflow = <TData = unknown, TResult = unknown>(
...
+    }
+
+    await eventBusService
+      .releaseGroupedEvents(eventGroupId)
+      .catch(async (e) => {
+        await flow.cancel(transaction)
```
olivermrbl: question: Do we need to log or something?

olivermrbl (approved):
Nice! 💪

riqwan (approved):
nice!

packages/core/workflows-sdk/src/helper/workflow-export.ts
adrien2p: Sorry @olivermrbl but again we are probably doing things at the same time, me adding tests and improvements and you guys commenting 😂 so it is already in

packages/core/workflows-sdk/src/helper/workflow-export.ts
adrien2p: true, I ve added it but differently

sradevski (approved):

packages/core/workflows-sdk/src/helper/workflow-export.ts
```diff
@@ -466,3 +474,56 @@ export const exportWorkflow = <TData = unknown, TResult = unknown>(
...
+        { allowUnregistered: true }
+      ) || console
+
+    const { transaction } = args
+    const failedStatus = [TransactionState.FAILED, TransactionState.REVERTED]
+
+    if (failedStatus.includes(transaction.getState())) {
+      await eventBusService.clearGroupedEvents(eventGroupId).catch(() => {
+        logger.warn(`Failed to clear events for eventGroupId - ${eventGroupId}`)
+      })
```
riqwan: skipped the return here 

```suggestion
      await eventBusService.clearGroupedEvents(eventGroupId).catch(() => {
        logger.warn(`Failed to clear events for eventGroupId - ${eventGroupId}`)
      })
      
      return
```

adrien2p: oups I ve removed it by mistake, readded it but differently as well 💪

adrien2p:
@riqwan we must not forget to update the interface of the event bus module when we add new things that are publically available 😄 dont do anything I ve included it here

sradevski (approved):

PR #7666, opened 2024-06-10
Feat(order): post purchase support

carlos-r-l-rodrigues:
What:
- Returns, Claims and Exchanges will be supported using the same Order structure.
  The Order will always contain all the histoy of changes in it.
  Returns, Claims and Exchanges will reference items, shipping methods and transactions that are affected by them.
- Return entity

### DIFF
--- a/integration-tests/modules/__tests__/order/order.spec.ts
+++ b/integration-tests/modules/__tests__/order/order.spec.ts
@@ -257,7 +257,7 @@ medusaIntegrationTestRunner({
                 value: "1",
                 precision: 20,
               },
-              detail: {
+              detail: expect.objectContaining({
                 id: expect.any(String),
                 order_id: expect.any(String),
                 version: 1,
@@ -301,7 +301,7 @@ medusaIntegrationTestRunner({
                 return_received_quantity: 0,
                 return_dismissed_quantity: 0,
                 written_off_quantity: 0,
-              },
+              }),
               subtotal: 50,
               total: 50,
               original_total: 50,


packages/modules/order/src/types/order-change.ts
```diff
@@ -0,0 +1,35 @@
...
+  PENDING = "pending",
+  /**
+   * The order change is canceled.
+   */
+  CANCELED = "canceled",
+}
+
+export enum OrderChangeType {
+  RETURN = "return",
+  SWAP = "swap",
```
olivermrbl: suggestion: 
```suggestion
  EXCHANGE = "exchange",
```

packages/modules/order/src/types/order-change.ts
```diff
@@ -0,0 +1,35 @@
...
+  CLAIM = "claim",
+  ORDER_EDIT = "order_edit",
```
olivermrbl: suggestion:
```suggestion
  EDIT = "edit",
```

packages/modules/order/src/services/order-module-service.ts
```diff
@@ -2807,4 +2610,76 @@ export default class OrderModuleService<
 
     return Array.isArray(orderId) ? orders : orders[0]
   }
+
+  // ------------------- Bundled Order Actions
```
olivermrbl, line 2614: question: Could I get you to explain in a few sentences what we changed in this file?

packages/modules/order/src/models/order-item.ts
```diff
@@ -59,15 +66,30 @@ export default class OrderItem {
...
 
+  @ManyToOne({
+    entity: () => Return,
+    mapToPk: true,
+    fieldName: "return_id",
+    columnType: "text",
+    nullable: true,
+  })
+  @ReturnIdIndex.MikroORMIndex()
+  return_id: string | null = null
```
olivermrbl, line 82: question: In the migration, we add references to the other types of edits, e.g. `exchange_id` and `claim_id`. Should they not be added here too?

olivermrbl: Think this might apply to the other model changes as well.

packages/modules/order/src/utils/base-repository-find.ts
```diff
@@ -0,0 +1,151 @@
...
+    config.options.populateWhere.shipping_methods.version = version
+
+    if (!config.options.orderBy) {
+      config.options.orderBy = { id: "ASC" }
+    }
+
+    return await manager.find(entity, config.where, config.options)
+  }
+
+  klass.prototype.findAndCound = async function findAndCount(
```
olivermrbl:
```suggestion
  klass.prototype.findAndCount = async function findAndCount(
```

packages/modules/order/src/models/order-item.ts
carlos-r-l-rodrigues: Initially I was planning to add them while developing each feature. But I already added them everywhere haha.
I'll add here too.

packages/modules/order/src/services/order-module-service.ts
carlos-r-l-rodrigues: I just reorganized the functions to be separated by files in another folder. It was really hard to read.
The behavior is the same.

packages/modules/order/src/models/order-item.ts
carlos-r-l-rodrigues: I will leave these changes for the PRs that swap and exchange are implemented. There are multiple changes to do to create the models related to these 2 things, and it will make this PR even bigger.
is that ok?

packages/modules/order/src/services/actions/create-return.ts
```diff
@@ -0,0 +1,113 @@
...
+    sharedContext
+  )
+
+  const [returnRef] = await this.createReturns(
+    [
+      {
+        order_id: data.order_id,
+        order_version: order.version,
+        status: ReturnStatus.REQUESTED,
+        // refund_amount: data.refund_amount ?? null,
```
olivermrbl, line 33: Do we need a todo here?

carlos-r-l-rodrigues: I already have it in my new branch.
Will get back to it when it is clearer where we'll calculate this amount.

PR #7667, opened 2024-06-10
Initial implementation of DML to Mikro ORM entity generation

thetutlage:
This PR focuses on create a helper method that can take the DML entity and generate a type-safe Mikro-ORM entity from it.

The relationships implementation is not 100% done. But I wanted to open this PR for the initial feedback on the approach.

### DIFF
--- a/integration-tests/http/__tests__/product-category/admin/product-category.spec.ts
+++ b/integration-tests/http/__tests__/product-category/admin/product-category.spec.ts
@@ -639,18 +639,79 @@ medusaIntegrationTestRunner({
       })
     })
 
-    // TODO: Should be migrate to V2
-    describe.skip("DELETE /admin/product-categories/:id", () => {
-      it("returns successfully with an invalid ID", async () => {
-        const response = await api.delete(
-          `/admin/product-categories/invalid-id`,
-          adminHeaders
-        )
+    describe("DELETE /admin/product-categories/:id", () => {
+      beforeEach(async () => {
+        productCategoryParent = (
+          await api.post(
+            "/admin/product-categories",
+            {
+              name: "category parent",
+              description: "category parent",
+            },
+            adminHeaders
+          )
+        ).data.product_category
 
-        expect(response.status).toEqual(200)
-        expect(response.data.id).toEqual("invalid-id")
-        expect(response.data.deleted).toBeTruthy()
-        expect(response.data.object).toEqual("product-category")
+        productCategory = (
+          await api.post(
+            "/admin/product-categories",
+            {
+              name: "category-0",
+              parent_category_id: productCategoryParent.id,
+              rank: 0,
+              description: "category-0",
+            },
+            adminHeaders
+          )
+        ).data.product_category
+
+        productCategoryChild = (
+          await api.post(
+            "/admin/product-categories",
+            {
+              name: "category child",
+              parent_category_id: productCategory.id,
+              rank: 0,
+              description: "category child",
+            },
+            adminHeaders
+          )
+        ).data.product_category
+
+        productCategoryChild2 = (
+          await api.post(
+            "/admin/product-categories",
+            {
+              name: "category child 2",
+              parent_category_id: productCategoryChild.id,
+              rank: 2,
+              description: "category child 2",
+            },
+            adminHeaders
+          )
+        ).data.product_category
+
+        productCategoryChild3 = (
+          await api.post(
+            "/admin/product-categories",
+            {
+              name: "category child 3",
+              parent_category_id: productCategoryChild.id,
+              rank: 2,
+              description: "category child 3",
+            },
+            adminHeaders
+          )
+        ).data.product_category
+      })
+
+      // BREAKING: In v1 we would return 200 on an invalid ID, in v2 we return 404
+      it("returns successfully with an invalid ID", async () => {
+        const err = await api
+          .delete(`/admin/product-categories/invalid-id`, adminHeaders)
+          .catch((e) => e)
+
+        expect(err.response.status).toEqual(404)
       })
 
       it("throws a not allowed error for a category with children", async () => {
@@ -669,49 +730,43 @@ medusaIntegrationTestRunner({
       })
 
       it("deletes a product category with no children successfully", async () => {
-        const deleteResponse = await api
-          .delete(
-            `/admin/product-categories/${productCategory.id}`,
-            adminHeaders
-          )
-          .catch((e) => e)
+        const deleteResponse = await api.delete(
+          `/admin/product-categories/${productCategoryChild2.id}`,
+          adminHeaders
+        )
 
         expect(deleteResponse.status).toEqual(200)
-        expect(deleteResponse.data.id).toEqual(productCategory.id)
+        expect(deleteResponse.data.id).toEqual(productCategoryChild2.id)
         expect(deleteResponse.data.deleted).toBeTruthy()
-        expect(deleteResponse.data.object).toEqual("product-category")
+        expect(deleteResponse.data.object).toEqual("product_category")
 
         const errorFetchingDeleted = await api
-          .get(`/admin/product-categories/${productCategory.id}`, adminHeaders)
+          .get(
+            `/admin/product-categories/${productCategoryChild2.id}`,
+            adminHeaders
+          )
           .catch((e) => e)
 
         expect(errorFetchingDeleted.response.status).toEqual(404)
       })
 
       it("deleting a product category reorders siblings accurately", async () => {
-        const deleteResponse = await api
-          .delete(
-            `/admin/product-categories/${productCategory.id}`,
-            adminHeaders
-          )
-          .catch((e) => e)
+        const deleteResponse = await api.delete(
+          `/admin/product-categories/${productCategoryChild2.id}`,
+          adminHeaders
+        )
 
         expect(deleteResponse.status).toEqual(200)
-
         const siblingsResponse = await api
           .get(
-            `/admin/product-categories?parent_category_id=${productCategoryParent.id}`,
+            `/admin/product-categories?parent_category_id=${productCategoryChild.id}`,
             adminHeaders
           )
           .catch((e) => e)
 
         expect(siblingsResponse.data.product_categories).toEqual([
           expect.objectContaining({
-            id: productCategory1.id,
-            rank: 0,
-          }),
-          expect.objectContaining({
-            id: productCategory2.id,
+            id: productCategoryChild3.id,
             rank: 1,
           }),
         ])
--- a/packages/core/core-flows/src/common/steps/release-event.ts
+++ b/packages/core/core-flows/src/common/steps/release-event.ts
@@ -4,17 +4,16 @@ import { createStep } from "@medusajs/workflows-sdk"
 export const releaseEventsStepId = "release-events-step"
 export const releaseEventsStep = createStep(
   releaseEventsStepId,
-  async (
-    input: void,
-    {
-      container,
-      metadata: {
-        /* eventGroupId */
-      },
+  async (input: void, { container, eventGroupId }) => {
+    const eventBusService = container.resolve(
+      ModuleRegistrationName.EVENT_BUS,
+      { allowUnregistered: true }
+    )
+    if (!eventBusService || !eventGroupId) {
+      return
     }
-  ) => {
-    const eventBus = container.resolve(ModuleRegistrationName.EVENT_BUS)
-    // await eventBus.release
+
+    await eventBusService.releaseGroupedEvents(eventGroupId)
   },
   async (data: void) => {}
 )
--- a/packages/core/core-flows/src/product-category/steps/delete-product-category.ts
+++ b/packages/core/core-flows/src/product-category/steps/delete-product-category.ts
@@ -1,0 +1,28 @@
+import { ModuleRegistrationName } from "@medusajs/modules-sdk"
+import { IProductModuleService } from "@medusajs/types"
+import { StepResponse, createStep } from "@medusajs/workflows-sdk"
+
+export const deleteProductCategoryStepId = "delete-product-category"
+export const deleteProductCategoryStep = createStep(
+  deleteProductCategoryStepId,
+  async (id: string, { container }) => {
+    const service = container.resolve<IProductModuleService>(
+      ModuleRegistrationName.PRODUCT
+    )
+
+    await service.deleteCategory(id)
+    return new StepResponse(void 0, id)
+  },
+  async (prevId, { container }) => {
+    if (!prevId) {
+      return
+    }
+
+    const service = container.resolve<IProductModuleService>(
+      ModuleRegistrationName.PRODUCT
+    )
+
+    // TODO: There is no soft delete support for categories yet
+    // await service.restoreCategory(prevId)
+  }
+)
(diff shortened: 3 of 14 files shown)


packages/core/utils/src/dml/helpers/create_mikro_orm_entity.ts
```diff
@@ -0,0 +1,93 @@
...
+    const field = entity.schema[property].parse(property)
+
+    /**
+     * Define field as a property on the class
+     */
+    if ("fieldName" in field) {
+      Property({ type: field.dataType.name, columnType: field.dataType.name })(
+        MikroORMEntity.prototype,
+        property
+      )
```
sradevski: Return early and remove the `else` step

packages/core/utils/src/dml/helpers/create_mikro_orm_entity.ts
```diff
@@ -0,0 +1,93 @@
...
+    },
+  })
+
+  Object.keys(entity.schema).forEach((property) => {
+    const field = entity.schema[property].parse(property)
+
+    /**
+     * Define field as a property on the class
+     */
+    if ("fieldName" in field) {
```
sradevski, line 196: Will the relations not have a field name?

packages/core/utils/src/dml/helpers/create_mikro_orm_entity.ts
```diff
@@ -0,0 +1,93 @@
...
+        )
+      }
+
+      const relatedModelName = upperCaseFirst(relatedEntity.name)
+
+      /**
+       * Defining relationships
+       */
+      switch (field.type) {
+        case "hasOne":
```
sradevski: On the other side of the relationship of, eg. a `hasMany`, wouldn't that result in `ManyToOne` relation?

packages/core/utils/src/dml/helpers/create_mikro_orm_entity.ts
```diff
@@ -0,0 +1,93 @@
...
+          `Invalid relationship reference for "${entity.name}.${property}"`
+        )
+      }
+
+      const relatedModelName = upperCaseFirst(relatedEntity.name)
+
+      /**
+       * Defining relationships
+       */
+      switch (field.type) {
```
sradevski: How will the resolution of relationships work? Will we first load all models, and then ensure that the `hasOne` on one side has a matching `hasMany` on the other side?

packages/core/utils/src/dml/helpers/create_mikro_orm_entity.ts
```diff
@@ -0,0 +1,93 @@
...
+   * the DML entity name
+   */
+  Object.defineProperty(MikroORMEntity, "name", {
+    get: function () {
+      return className
+    },
+  })
+
+  Object.keys(entity.schema).forEach((property) => {
+    const field = entity.schema[property].parse(property)
```
olivermrbl, line 195: thought: I would expect our schemas to map to the column properties that we have used in the core. 

~~For example, the following are the columns for a `fulfillment_set` (simplified for brevity)~~

So what I mean is we would have a mapping between our schema primitives and the MikroORM properties that look something along these lines:
- `text()` -> `@Property({ columnType: "text" })`
- `dateTime()` -> `@Property({ columnType: "timestamptz" })`
- `number()` -> `@Property({ columnType: "integer" })`

And so on and so forth. 

What do you think? Also, let me know if I am missing something, I might not have the full context of this change.

sradevski (commented):
Overall looks sane, I don't see anything wrong with the approach 😄

packages/core/utils/src/dml/helpers/create_mikro_orm_entity.ts
olivermrbl: Maybe this is actually more about how the schema primitives define their `dataType` than this entity generation utility. Because this approach here looks good to me.

adrien2p: I am not sure I understand what you mean? there is still some changes to perform in regards to type vs columnType I believe as one is ts type and the other one is db type. maybe this is what you meant?

carlos-r-l-rodrigues: I am not sure I understand, but the mikro-orm schema from what I understand will be composed dinamically. We are not generating typescript from the DML.

olivermrbl: Ah OK, maybe I am missing something or my message was phrased incorrectly. Does the `Property` on line 42-45 not correspond to the Property decorators we use on the MikroORM models?

adrien2p: it does, except that in our model when we apply the decorator it can infer the typescript type from the property, but with dynamic generation you dont have a typescript type and you have to set it manually. Here I think the usage must be changed as it is incorrect at the moment, but it is similar to what we have been doing with big number for example where we set both columnType and type, the first one for the db, the second one for the typing

adrien2p: here is an example for dynamic apply of the decorator we used for big number
```
Property({
      type: "any",
      columnType: "numeric",
      trackChanges: false,
      ...options,
    })(target, columnName)
```

adrien2p: if what you mean is the mapping is incorrect then yes it is, but I believe @thetutlage wanted to get early feedback on the approach mostly :)

olivermrbl: Cool, then I believe my comment/question still stands. 

Essentially, what I mean is it looks like we are not generating the same `columnType`s for the DB here as we have used in the core.

riqwan (commented):
Not sure if you came across this or if this will be useful, but adding it here just incase it is - https://mikro-orm.io/docs/entity-generator

packages/core/utils/src/dml/helpers/create_mikro_orm_entity.ts
adrien2p: yes, as mentioned above, I believe virk has not done anything around the mapping yet

adrien2p: it is still a draft to validate the approach, i.e the way the class will be auto generated and the decorators applied, then the options of the decorators will be fixed 😄

olivermrbl: Got it, as long as we are aware of it, it's all good. Approach looks solid 💪

adrien2p: yes ahah, I have a call with virk tomorrow morning, we will go through it 👍

packages/core/utils/src/dml/helpers/create_mikro_orm_entity.ts
thetutlage: Nope. Field name is strictly for columns.

packages/core/utils/src/dml/helpers/create_mikro_orm_entity.ts
thetutlage: Yeah, have a call with @adrien2p tomorrow. My goal is to fill the gaps between our schema and MikroORM entities in that call.

thetutlage:
@riqwan Seems like EntityGenerator uses some sort of discovery from directories?

riqwan:
@thetutlage just had a chance to take a deeper look. It generates entities after looking at the db schema and optionally creates a file out of it in your project. I thought it might offer some builder primitives to aid in your work, but seems like its very purpose made. Ignore my previous comment 😄

packages/core/utils/src/dml/helpers/create_mikro_orm_entity.ts
```diff
@@ -0,0 +1,93 @@
...
+          )
+          break
+        case "hasMany":
+          OneToMany({ entity: relatedModelName, mappedBy: "" })(
+            MikroORMEntity.prototype,
+            property
+          )
+          break
+        case "manyToMany":
+          ManyToMany({ entity: relatedModelName, mappedBy: "" })(
```
riqwan: q: How would this look like with an owning side and an inverse side? eg:

```
@Entity()
export class Book {
  @ManyToMany(() => BookTag, 'books', { owner: true })
  tags4 = new Collection<BookTag>(this);
}

@Entity()
export class BookTag {
  @ManyToMany(() => Book, book => book.tags)
  books = new Collection<Book>(this);
}
```

riqwan (commented):
very cool!

packages/core/utils/src/dml/helpers/create_mikro_orm_entity.ts
thetutlage: Cleaned up this part of the codebase to have its own functions

thetutlage:
# Run down of DML entity to Mikro ORM entity generation

Let me give you a quick overview of how this process works under the hood, and then we can discuss it further.

## Properties/Fields

Mikro ORM uses the term properties (via their `@Property` decorator), and I am using fields in the DML.

Following are the mappings between the DML fields and Mikro ORM properties.

- `boolean`: columnType (boolean)
- `dateTime`: columnType (timestamptz)
- `number`: columnType (integer),
- `text`: columnType (text),
- `json`: columnType (jsonb),
- `enum`: Use the `@Enum` decorator instead.

Following is the list of options accepted by Mikro ORM via their `@Property` decorator.

- `name`: Not set
- `fieldName`: Not set
- `fieldNames`: Not set
- `customType`: Not set
- `columnType`: **Computed from above shared mapping**
- `type`: Computed
- `length`: Not set
- `precision`: Not set
- `scale`: Not set
- `autoincrement`: Not set
- `returning`: Not set
- `onCreate`: Not set
- `onUpdate`: Not set
- `default`: **Will be computed from the DML schema**
- `defaultRaw`: Not set
- `formula`: Not set
- `nullable` **Will be computed from the DML schema**
- `unsigned`: Not set
- `persist`: Not set
- `hydrate`: Not set
- `trackChanges`: Not set
- `hidden`: Not set
- `version`: Not set
- `concurrencyCheck`: Not set
- `index`: **Will be defined once we add indexes support**
- `unique`: **Will be defined once we add indexes support**
- `check`: Not set
- `lazy`: Not set
- `primary`: **Will be defined once we add indexes support**
- `setter`: Not set
- `getter`: Not set
- `serializedPrimaryKey`: Not set
- `serializer`: Not set
- `serializedName`: Not set
- `customOrder`: Not set
- `comment`: Not set
- `extra`: Not set
- `ignoreSchemaChanges`: Not set

## Relationships

The relationships in DML are named differently from Mikro ORM. However, to properly configure relationships, we still have to offer some of Mikro ORM's capabilities in DML as well.

### `hasOne -> OneToOne`

The `OneToOne` relationship of Mikro ORM is represented as a `hasOne` relationship in the DML. 

We must expose an API to configure the following properties in the DML.

- `nullable`: Will be supported by chaining the `nullable` modifier.
- `onDelete`: Same as Mikro ORM
- `owner`: Same as Mikro ORM
- `cascade`: Same as Mikro ORM
- `mappedBy`: : Same as Mikro ORM

**The following options are ignored.**

- `inversedBy`
- `wrappedReference`
- `ref`
- `mapToPK`
- `onUpdateIntegrity`
- `primary`
- `orphanRemoval`
- `joinColumn`
- `joinColumns`
- `inverseJoinColumn`
- `inverseJoinColumns`
- `referenceColumnName`
- `eager`
- `strategy`

### `hasMany -> OneToMany`

The `OneToMany` relationship of Mikro ORM is represented as a `hasMany` relationship in the DML. 

We must expose an API to configure the following properties in the DML.

- `onDelete`: Same as Mikro ORM
- `orphanRemoval`: Same as Mikro ORM
- `cascade`: Same as Mikro ORM
- `mappedBy`: : Same as Mikro ORM

**The following options are ignored.**

- `joinColumn`
- `joinColumns`
- `orderBy`
- `inverseJoinColumn`
- `inverseJoinColumns`
- `referenceColumnName`
- `eager`
- `strategy`

### `hasOneThroughMany -> ManyToOne`

I haven't been able to come up with a great name for `hasOneThroughMany`. So please feel free to suggest a better name.

The `ManyToOne` relationship of Mikro ORM is represented as the `hasOneThroughMany` relationship in the DML. 

We must expose an API to configure the following properties in the DML.

- `nullable`: Will be supported by chaining the `nullable` modifier.
- `onDelete`:  Same as Mikro ORM
- `fieldName`: Same as Mikro ORM but auto-generated when not provided.
- `columnType`: Same as Mikro ORM. The column type for the foreign key. It will be inferred from the primary key of the related entity.

**The following options are ignored.**

- `inversedBy`
- `wrappedReference`
- `ref`
- `primary`
- `mapToPk`
- `joinColumn`
- `joinColumns`
- `referenceColumnName`
- `referencedColumnNames`
- `onUpdateIntegrity`


### `manyToMany - ManyToMany`

Here, the naming is identical.

We must expose an API to configure the following properties in the DML.

- `mappedBy`: Same as Mikro ORM
- `pivotTable`: Same as Mikro ORM
- `pivotEntity`: Unsure if we should allow this and what the API will look like.

 **The following options are ignored.**

- `owner`
- `inversedBy`
- `mappedBy`
- `orderBy`
- `fixedOrder`
- `fixedOrderColumn`
- `joinColumn`
- `joinColumns`
- `inverseJoinColumn`
- `inverseJoinColumns`
- `referenceColumnName`
- `referencedColumnNames`

CC @olivermrbl @sradevski @riqwan @carlos-r-l-rodrigues @adrien2p 

Special thanks for @adrien2p for spending more than 3 hours pairing with me on this

sradevski:
@thetutlage Here are my thoughts:


1. I think we should not have both `onDelete` and `cascade` as part of the DML. This is a frequent point of confusion, so we should look to make it clearer
2. The owner and mappedBy can be made more intuitive IMO compared to what mikroORM has
3. How the relationships are defined is not very consistent and clean IMO. What I would imagine us doing is only have `hasOne` and `hasMany` methods, and determine if it is oneToOne or oneToMany based on what the other side has defined. An example:
```
      email: model.text(),
    }).hasOne("profile")
    
const profile = model.define("profile", {
first_name: model.text()
}).hasOne("user")
```
This would result in a oneToOne relationship.

However,
```
      email: model.text(),
    }).hasMany("address")
    
const address = model.define("address", {
address: model.text()
}).hasOne("user")
```

Would be a oneToMany relationship.

WDYT?

thetutlage:
> How the relationships are defined is not very consistent and clean IMO. What I would imagine us doing is only have hasOne and hasMany methods, and determine if it is oneToOne or oneToMany based on what the other side has defined. An example:

The only concern here is we are forcing the other side of relationship to exist. What if I do not have to define a relation from `user` to `addresses`? Or do we want to force it always?

thetutlage:
If everything looks fine, maybe we can start reviewing this PR from implementation point of view.

I will further refine the relationships API in a different PR and also see if we can incorporate suggestions of @sradevski in that. 

Had a discussion with @adrien2p in Slack and we will start exploring the possibility to automatically detect the correct relationship between `OneToOne` or `ManyToOne` (as suggested by @sradevski).

olivermrbl (approved):
Strong work @thetutlage 🔥 

+1 for doing improvements/refinements in follow up PRs

adrien2p (approved):

sradevski (approved):
This looks good so far, let's go, one PR at a time :)

PR #7668, opened 2024-06-10
chore: ensure the event group id is kept for async workflow

adrien2p:
**What**
Ensure that the event group id is the same for every steps of async workflow as well

FIXES CORE-2278


carlos-r-l-rodrigues (approved):

riqwan (approved):
💪🏻

sradevski (approved):

PR #7674, opened 2024-06-11
feat: Add support for scheduled jobs

sradevski:


packages/medusa/src/loaders/index.ts
```diff
@@ -57,6 +58,14 @@ async function subscribersLoader(
   )
 }
 
+async function jobsLoader(plugins: PluginDetails[]) {
+  /**
+   * Load subscribers from the medusa/medusa package. Remove once the medusa core is converted to a plugin
```
adrien2p: nit: jobs instead of subscribers?

adrien2p (approved):

olivermrbl (approved):
Super clean!

PR #7675, opened 2024-06-11
feat(dashboard,types,sdk,medusa,ui): ProductCategory domain

kasperkristensen:
**What**
- Add missing features to ProductCategory domain in admin
- Add types
- Add SDK

**UI**
- Moves the TooltipProvider from the component to an export. Users should now wrap their entire application in a TooltipProvider. This change was made to take advantage of the built-in features of Radix Tooltip, and allows us to skip the delayDuration when moving the cursor from one tooltip to another within 500ms.
- Fixes the layout of the Hint component, as the create form revealed that it was off.
- Fixes an issue where focus styles were missing from the dropdown menu.

**Note**
- ~~We currently don't have an endpoint for deleting categories, so I have disabled the button in the admin. See CORE--2286~~ PR has been opened to add delete endpoint, so I have re-enabled the delete button.
- The update category workflow seems to be broken, it's possible for the `mpath` of a category to reach an invalid state, that breaks `include_descendants_tree` from working. See CORE-2287.
- The ProductCategory model is incorrect. All fields are optional and it's not possible to set the description to null, which means the only way of unsetting it is to set it to `""`. See CORE-2276.
- The design for the Organize drag-n-drop form is not final. Ludvig will create a final design, and we can then update the form.
- Currently, all things related to Metadata is left out, as we need to update the flow for metadata according to the latest designs.

RESOLVES CORE-1960, CORE-2230
*except for the above mentioned issues.


packages/core/js-sdk/src/admin/product-category.ts
```diff
@@ -0,0 +1,97 @@
...
+    return this.client.fetch<HttpTypes.AdminProductCategoryDeleteResponse>(
+      `/admin/product-categories/${id}`,
+      {
+        method: "DELETE",
+        headers,
+      }
+    )
+  }
+
+  async updateProducts(
```
sradevski, line 81: Maybe we can use different names for the relationship-setting actions, like `category.setProducts` or `assignProducts`?

packages/core/types/src/http/product-category/common.ts
```diff
@@ -1,15 +1,43 @@
...
+  include_descendants_tree?: boolean
+  include_ancestors_tree?: boolean
+  created_at?: OperatorMap<string>
+  updated_at?: OperatorMap<string>
+  deleted_at?: OperatorMap<string>
+}
 
-  parent_category: ProductCategoryResponse
-  category_children: ProductCategoryResponse[]
+export interface BaseProductCategoryParams extends SelectParams {
```
sradevski, line 40: We are redeclaring the same interface twice, we need to use a different name here.

packages/medusa/src/api/admin/product-categories/validators.ts
```diff
@@ -65,7 +65,7 @@ export const AdminUpdateProductCategory = z
     handle: z.string().optional(),
```
sradevski, line 65: `description` should also be nullable right?

sradevski (commented):
I only checked the non-FE changes, I'll check the FE in a separate review batch.
One note, can we also update the api routes (can be a separate PR)?

packages/medusa/src/api/admin/product-categories/validators.ts
kasperkristensen: Not supported by the model atm, I've opened an issue for it here CORE-2276

packages/core/types/src/http/product-category/common.ts
kasperkristensen: Ahh missed that thanks!

packages/core/js-sdk/src/admin/product-category.ts
kasperkristensen: But it's both removing and adding products, so assign doesn't work, and set seems more like of a source of truth action as in this will now be the the products that are part of this category.

sradevski: Yeah I agree neither are great, but `updateProducts` seems more like it will update the product data, not set products on categories, so `setProducts` is a tiny bit better IMO, but I leave it up to you

olivermrbl: ~~Maybe `sdk.category.modifyProducts()`?~~

Nvm. think it is not any better than update.

olivermrbl:
There is something off about the visibility and status selects in the edit drawer:

CleanShot 2024-06-11 at 19 52 07

olivermrbl (approved):

packages/core/js-sdk/src/admin/product-category.ts
kasperkristensen: I'll just leave it for now, and we can come up with a proper name for this type of action for all the relevant domains

PR #7681, opened 2024-06-11
feat(order): Claim and Exchange entities

carlos-r-l-rodrigues:


olivermrbl (approved):
Overall LGTM – if I had to be nitpicky, I would say we should try to shorten the entity prefixes 😄

carlos-r-l-rodrigues:
> I would say we should try to shorten the entity prefixes 😄

I can do that in the next PR

PR #7682, opened 2024-06-11
feat: `order.placed` event + simplify `emitEventStep`

olivermrbl:
**What**
- Injects `order.placed` event in cart-completion
- Simplifies the `emitEventStep`*

*The message composer is meant to be used for module entity events and requires properties only used by the index engine. Therefore, I removed the usage of the composer and instead we emit an event with `data` and an open-ended `metadata` property.

### DIFF
--- a/packages/core/core-flows/src/common/steps/emit-event.ts
+++ b/packages/core/core-flows/src/common/steps/emit-event.ts
@@ -1,15 +1,14 @@
-import { composeMessage, ModuleRegistrationName } from "@medusajs/utils"
-import { createStep, StepExecutionContext } from "@medusajs/workflows-sdk"
+import { EventBusTypes, IEventBusModuleService } from "@medusajs/types"
+import { ModuleRegistrationName } from "@medusajs/utils"
+import { StepExecutionContext, createStep } from "@medusajs/workflows-sdk"
 
 type Input = {
   eventName: string
-  source: string
-  object: string
-  action?: string
   options?: Record<string, any>
-  data: (
-    context: StepExecutionContext
-  ) => Promise<Record<any, any>> | Record<any, any>
+  metadata?: Record<string, any>
+  data:
+    | ((context: StepExecutionContext) => Promise<Record<any, any>>)
+    | Record<any, any>
 }
 
 export const emitEventStepId = "emit-event-step"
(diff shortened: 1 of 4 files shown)


riqwan (approved):
🥂

packages/core/core-flows/src/common/steps/emit-event.ts
```diff
@@ -22,18 +21,27 @@ export const emitEventStep = createStep(
...
+    const message = {
+      eventName: input.eventName,
       data: data_,
-      action: input.action ?? "",
-      object: input.object,
-      source: input.source,
       options: input.options,
-      context,
-    })
+      metadata: input.metadata,
```
adrien2p: Suggestion: should we use metadata instead here?

packages/core/utils/src/order/events.ts
```diff
@@ -0,0 +1,6 @@
+export const orderEvents = {
+  PLACED: "order.placed",
+  UPDATED: "order.updated",
+  CANCELED: "order.canceled",
+  COMPLETED: "order.completed",
+}
```
adrien2p: suggestion: this does not follow the convention, I would suggest to lower case the keys and remove the updated
```
export const orderEvents = {
  placed: "order.placed",
  canceled: "order.canceled",
  completed: "order.completed",
}
```

packages/core/core-flows/src/common/steps/emit-event.ts
olivermrbl: Yes we def. should. Nice catch :)

packages/core/utils/src/order/events.ts
```diff
@@ -0,0 +1,6 @@
+export const orderEvents = {
```
adrien2p: nit: OrderEvents

packages/core/utils/src/order/events.ts
olivermrbl: I would maybe push to have us change the convention to use an uppercased version of the action:

So:
```
OrderEvents.PLACED
```
instead of 
```
OrderEvents.placed
```

Makes it more obvious that it is a "global constant".

Not necessarily right now, but thought I'd share my thinking.

adrien2p: yes for sure, we just need to update the utils that build the constants 👍  and then the usage

sradevski (approved):

packages/core/utils/src/order/events.ts
sradevski: +1 for uppercase, a bit easier to read.

PR #7687, opened 2024-06-12
Code cleanup and add support for default values and nullable relationships

thetutlage:
- Add support for nullable modifier in relationships.
- Rename datatype of `text` property to `text`. Earlier we renamed the public method name, but the underlying data was still called `string`.
- Add support for defining default values for properties. **NOT RELATIONSHIPS**
- Add code comments and perform general cleanup

### DIFF
--- a/packages/core/utils/src/dml/__tests__/entity_builder.spec.ts
+++ b/packages/core/utils/src/dml/__tests__/entity_builder.spec.ts
@@ -126,7 +126,7 @@ describe("Entity builder", () => {
     const model = new EntityBuilder()
     const user = model.define("user", {
       id: model.number(),
-      username: model.text().defaultsTo("foo"),
+      username: model.text().default("foo"),
       email: model.text(),
     })
 
--- a/packages/core/utils/src/dml/__tests__/base_schema.spec.ts
+++ b/packages/core/utils/src/dml/__tests__/base_schema.spec.ts
@@ -52,7 +52,7 @@ describe("Base schema", () => {
       }
     }
 
-    const schema = new StringSchema().defaultsTo("foo")
+    const schema = new StringSchema().default("foo")
 
     expectTypeOf(schema["$dataType"]).toEqualTypeOf<string>()
     expect(schema.parse("username")).toEqual({
--- a/packages/core/utils/src/dml/schema/base.ts
+++ b/packages/core/utils/src/dml/schema/base.ts
@@ -32,7 +32,7 @@ export abstract class BaseSchema<T> implements SchemaType<T> {
   /**
    * Define default value for the property
    */
-  defaultsTo(value: T) {
+  default(value: T) {
     this.#defaultValue = value
     return this
   }


packages/core/utils/src/dml/__tests__/base_relationship.spec.ts
```diff
@@ -21,6 +21,37 @@ describe("Base relationship", () => {
...
+    }).nullable()
+
+    expectTypeOf(relationship["$dataType"]).toEqualTypeOf<
+      (() => typeof user) | null
+    >()
+    expect(relationship.parse("user")).toEqual({
+      name: "user",
+      type: "hasOne",
+      nullable: true,
+      optional: false,
```
adrien2p: Q: if it is nullable, shouldn't it be optional? in case you create a relation, you shouldn't be forced to pass the user

sradevski (approved):
Nice stuff 👍

packages/core/utils/src/dml/__tests__/entity_builder.spec.ts
```diff
@@ -122,6 +122,56 @@ describe("Entity builder", () => {
     ])
   })
 
+  test("define an entity with default value", () => {
+    const model = new EntityBuilder()
+    const user = model.define("user", {
+      id: model.number(),
+      username: model.text().defaultsTo("foo"),
```
adrien2p: suggestion: I would prefer `default` instead of `defaultsTo` wdyt?

packages/core/utils/src/dml/__tests__/entity_builder.spec.ts
```diff
@@ -122,6 +122,56 @@ describe("Entity builder", () => {
...
+    }>()
+
+    const metaData = MetadataStorage.getMetadataFromDecorator(User)
+    expect(metaData.className).toEqual("User")
+    expect(metaData.path).toEqual("User")
+    expect(metaData.properties).toEqual({
+      id: {
+        reference: "scalar",
+        type: "number",
+        columnType: "integer",
```
adrien2p, line 147: Q: we use integer by default, should we allow to configure that?? maybe not now but later a user could easily want numeric or big int no?

packages/core/utils/src/dml/modifiers/nullable.ts
```diff
@@ -12,9 +15,11 @@ export class NullableModifier<T> {
    * The parent schema on which the nullable modifier is
    * applied
    */
-  #schema: SchemaType<T>
+  #schema: {
+    parse(fieldName: string): MaybeFieldMetadata
+  }
 
-  constructor(schema: SchemaType<T>) {
+  constructor(schema: { parse(fieldName: string): MaybeFieldMetadata }) {
```
adrien2p, line 21: Q: should we use `Pick<SchemaType<T>, 'parse'>` instead? just to prevent divergeance?

adrien2p: If we agree on that then there is other places as well

packages/core/utils/src/dml/modifiers/nullable.ts
```diff
@@ -1,6 +1,9 @@
-import { SchemaType } from "../types"
+import { MaybeFieldMetadata } from "../types"
 import { OptionalModifier } from "./optional"
 
+/**
+ * Nullable modifier marks a schema node as nullable
```
adrien2p, line 4: suggestion: currently, for all nullable field our strategy was to make it optional and default to null by auto assigning, but I believe we should backport that

packages/core/utils/src/dml/__tests__/entity_builder.spec.ts
olivermrbl: +1

packages/core/utils/src/dml/__tests__/base_relationship.spec.ts
thetutlage: Ideally, optional and nullable are two separate things. I am not sure if we need optional in our case. Here's my understanding of two.

- `nullable`: I want to explicitly override the existing value by assign `null` to it.
- `optional`: If a new value is not provided, keep the old value as it is. And do not force me to provide a new value.

olivermrbl (approved):
Nice, LGTM – only thing would be the suggestion into change `defaultsTo` -> `default`

packages/core/utils/src/dml/__tests__/base_relationship.spec.ts
adrien2p: yes, but in my memories if a property is not auto assigned or marked as optional then mikro orm will force you to provide a value. That is why we generally go with an auto assign null which makes it optional without marking it as optional which is what we could eventually do and if we can't we can mark it as optional with a default to null such that you are not forced to provide at least null. I dont know if I am clear 😂 we can discuss it .

Also, null is not necessarely to override an existing value, it makes the db column nullable and we dont necessarely want to provide the same value if required and we dont want to update it

packages/core/utils/src/dml/__tests__/entity_builder.spec.ts
thetutlage: I do not get it. We have a schema type called `model.number` which maps to `integer` data-type. Do you want to call it `model.integer` instead?

packages/core/utils/src/dml/modifiers/nullable.ts
thetutlage: The modifier could be used with both `RelationshipType` and `SchemaType`. Ideally it should care what the `parse` returns except `nullable` and `optional` fields.

So it feels natural to have a sub-type that represents a MaybeField. Be it relationship or a property

packages/core/utils/src/dml/__tests__/entity_builder.spec.ts
adrien2p: no no we can keep it as it, I was just thinking about the other "number" types 😂

packages/core/utils/src/dml/__tests__/entity_builder.spec.ts
thetutlage: Sure. Seems like Oli asked for the same.

packages/core/utils/src/dml/__tests__/base_relationship.spec.ts
thetutlage: Yeah, I think we should jump on a call to understand it better :)

packages/core/utils/src/dml/modifiers/nullable.ts
adrien2p: ah yes indeed it can be used with both 👍

packages/core/utils/src/dml/__tests__/base_relationship.spec.ts
olivermrbl: Would it make sense to only have the `nullable` modifier? Or conversely, the `optional` one.

thetutlage: Yeah. I was thinking the same. Remove the optional modifier. Let me do it

riqwan (approved):
✨ niice!

packages/core/utils/src/dml/modifiers/nullable.ts
thetutlage: Let's discuss over the call and then handle it in next PR :)

PR #7688, opened 2024-06-12
chore: Refactor and improve abstract module service factory

adrien2p:
**What**
Refactor and cleanup the modules service API

**NOTE**
Fix create/update mapped types not working, now they are working but the API must be generic any as not all modules follow the same convention and that can't be define through the types currently

**NOTE 2**
We have to add back the ts-ignore now that the types have been fixed and therefore we get back to the known ts issue about mapped types instance method of dynamic classes (https://github.com/microsoft/TypeScript/issues/48125)

FIXES CORE-2294

### DIFF
--- a/packages/admin-next/dashboard/src/hooks/api/price-lists.tsx
+++ b/packages/admin-next/dashboard/src/hooks/api/price-lists.tsx
@@ -1,3 +1,5 @@
+import { FetchError } from "@medusajs/js-sdk"
+import { HttpTypes } from "@medusajs/types"
 import {
   QueryKey,
   UseMutationOptions,
@@ -5,7 +7,7 @@ import {
   useMutation,
   useQuery,
 } from "@tanstack/react-query"
-import { client } from "../../lib/client"
+import { client, sdk } from "../../lib/client"
 import { queryClient } from "../../lib/query-client"
 import { queryKeysFactory } from "../../lib/query-key-factory"
 import {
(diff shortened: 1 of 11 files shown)


adrien2p:
@sradevski @thetutlage @olivermrbl should be good to review and merge, then I can open the other pr with the fondational changes

thetutlage:
Looks pretty neat to me!

thetutlage (approved):

adrien2p:
@shahednasser just for info here how it will look like for simple users, some more changes will come soon though
```
export class TestModel {
  id: string
}

export class OtherModel {
  title: string
  id: string
}

class Service extends MedusaService(TestModel, { OtherModel }) {
  constructor(container: InjectedDependencies) {
    super(container)
  }

  async test() {
    const entities = await super.listOtherModels()
    const id = entities[0].title
  }
}
```

shahednasser:
@adrien2p looks much cleaner 😄 can you let me know when it's ok to update the docs with this change? (as in, when it's ok for users to use this).

Also, do we not need type arguments anymore?

olivermrbl (approved):
Sublime work!

olivermrbl:
cc @VariableVic

adrien2p:
> @adrien2p looks much cleaner 😄 can you let me know when it's ok to update the docs with this change? (as in, when it's ok for users to use this).
> 
> 
> 
> Also, do we not need type arguments anymore?

Yes i let you know, yes the template args are not required but can be used if you want to specify a specific dto for example

PR #7690, opened 2024-06-12
Overall revamp of relationships

thetutlage:
Okay, so the relationships have been revamped a bit. Let me start by explaining how it all works under the hood.

### hasOne
The `hasOne` relationship defines a `OneToOne` relationship with Mikro ORM with `owner: undefined`. 

- You can mark a hasOne as nullable using the `nullable` modifier.
- You may define a `mappedBy` key for `hasOne` relationship by chaining the `mappedBy` function. If not defined, it will be set to `undefined`.

### hasMany
The `hasMany` relationship defines a `OneToMany` relationship with Mikro ORM.

- The `orphanRemoval` option always set to `true` and cannot customized.
- The `nullable` modifier is not supported.
- The `mappedBy` key can be defined explicitly or computed internally using conventions. Mikro ORM forces `mappedBy` to be defined when creating `OneToMany` relationship.

### belongsTo
The `belongsTo` uses some auto-detection and defines either `ManyToOne` or `OneToOne (owner: true)`  relationship. This is how it works.

- You can mark a belongsTo as nullable using the `nullable` modifier.
- It uses the `mappedBy` key to inspect the relationship on the related entity.
- A runtime exception will be thrown if no relationship is defined on the related entity.
- A runtime exception will be thrown if `mappedBy` key on the related entity defines a relationship other than `hasOne` or `hasMany`. 

Let's look at some examples:

```ts
// ✅ OneToOne relationship is set between email -> user

  email: model.text(),
  isVerified: model.boolean(),
  user: model.belongsTo(() => user),
})

  id: model.number(),
  username: model.text(),
})
```

```ts
// ✅ ManyToOne relationship is set between email -> user

  email: model.text(),
  isVerified: model.boolean(),
  user: model.belongsTo(() => user).mappedBy("emails"),
})

  id: model.number(),
  username: model.text(),
  emails: model.hasMany(() => email),
})
```

```ts
// ❌ Exception raised because user does not define relationship back to email
  email: model.text(),
  isVerified: model.boolean(),
  user: model.belongsTo(() => user),
})

  id: model.number(),
  username: model.text(),
})
```

```ts
// ❌ Exception raised because user -> email is using an unsupported relationship
  email: model.text(),
  isVerified: model.boolean(),
  user: model.belongsTo(() => user).mappedBy('emails'),
})

  id: model.number(),
  username: model.text(),
  emails: model.manyToMany(() => user)
})
```

### manyToMany

The `manyToMany` relationship defines a `ManyToMany` relationship with Mikro ORM.

- Nullable modifier is not supported.
- You may define a `mappedBy` key for `manyToMany` relationship by chaining the `mappedBy` function. If not defined, it will be set to `undefined`.


adrien2p (approved):

thetutlage:
Going to leave the PR open for a while to see if others want to chip-in and provide feedback

packages/core/utils/src/dml/relations/base.ts
```diff
@@ -22,14 +23,16 @@ export abstract class BaseRelationship<T> implements RelationshipType<T> {
...
   }
 
   /**
-   * Apply nullable modifier on the schema
+   * Define the property name on the related entity that
+   * defines the inverse of this relationship.
    */
-  nullable() {
-    return new NullableModifier<T>(this)
+  mappedBy(property: string): this {
```
adrien2p: nit: I might just put the mappedBy as an argument of the relation but no strong opinion

olivermrbl: @adrien2p, do you mean something like `belongsTo(() => User, { mappedBy: "emails" })` ?

If so, I think I agree – the strict coupling between the concepts makes sense. 

In case we add more chainable APIs, we will avoid something like this:
```
user: belongsTo(()) => user).somethingElse().someOtherThing().mappedBy()
```

where the `mappedBy` is chained separately from the relationship. I know this is probably a farfetched example, but maybe something like this could happen.

Wdyt?

adrien2p: yes I was thinking more like `belongsTo(() => User,  "emails")` but your example works nicely too

riqwan: I have mixed feels about this. Only because I can't tell what the difference between these two are:

```
  email: model.text().optional(),
  user: model.belongsTo(() => user, { mappedBy: "emails" }).optional(),
})
```

what makes optional chainable here, but not mappedBy?

riqwan (approved):
delicious! 🤤

packages/core/utils/src/dml/relations/base.ts
adrien2p: Because optional is an option of the relation while the mapped refer to a property of the schema you are referring too. Therefore it can be almost read as user.emails

riqwan:
@thetutlage supernit: the file names seems to snakecase, but should be dasherized.

packages/core/utils/src/dml/relations/base.ts
olivermrbl: `mappedBy` cannot live without the relation (e.g. `belongsTo`) while `optional()` is a standalone concept and can live on its own

Update: I think we killed `optional()` but the argument still stands for other modifiers.

thetutlage: Both are equally correct. It all boils down to what you like more subjectively.

I'll go with the popular opinion here 😇

riqwan: That makes sense, thanks guys!

thetutlage: Done!

thetutlage:
> the file names seems to snakecase, but should be dasherized.

@riqwan Done ✌️

PR #7691, opened 2024-06-12
fix(medusa): allow calculating prices without explicit fields

riqwan:
what:

- fix: adding currency_code or region_id as a filter with store product endpoints will return calculated prices without needing to pass in calculated_price field explicitly
- chore: move store product test to http from module

RESOLVES CORE-2250


integration-tests/http/__tests__/product/store/product.spec.ts
```diff
@@ -1,24 +1,88 @@
...
+} from "@medusajs/utils"
 import { medusaIntegrationTestRunner } from "medusa-test-utils"
 import {
   adminHeaders,
   createAdminUser,
 } from "../../../../helpers/create-admin-user"
 import { getProductFixture } from "../../../../helpers/fixtures"
-import { ModuleRegistrationName } from "@medusajs/utils"
-import { IStoreModuleService } from "@medusajs/types"
+import { createDefaultRuleTypes } from "../../../../modules/helpers/create-default-rule-types"
```
sradevski, line 13: I think it's better for the http tests to not depend on modules helpers. We can either just do the fixture inline, do it as part of the `defaultsWorkflow` when booting medusa (if that's what we want), or move the helper higher up the stack.

Maybe creating these as part of the `defaultsWorkflow` makes sense, WDYT?

integration-tests/http/__tests__/product/store/product.spec.ts
```diff
@@ -401,5 +469,640 @@ medusaIntegrationTestRunner({
...
+          status: ProductStatus.DRAFT,
+          variants: [{ title: "test variant 4", prices: [] }],
+        })
+
+        const defaultSalesChannel = await createSalesChannel(
+          { name: "default sales channel" },
+          [product.id, product2.id, product3.id, product4.id]
+        )
+
+        const service = appContainer.resolve(ModuleRegistrationName.STORE)
```
sradevski, line 534: Let's move this to a `beforeEach`, and let's not store the `container` globally as we are trying to avoid people calling modules directly if possible

integration-tests/http/__tests__/product/store/product.spec.ts
```diff
@@ -401,5 +469,640 @@ medusaIntegrationTestRunner({
         expect(response.status).toEqual(200)
       })
     })
+
+    describe("GET /store/products", () => {
```
sradevski, line 473: Were these tests aimed at v2 only since the beginning? I don't see any `BREAKING` comments so maybe that's why, I am sure we do have breaking changes though 😄

integration-tests/http/__tests__/product/store/product.spec.ts
```diff
@@ -401,5 +469,640 @@ medusaIntegrationTestRunner({
...
+            },
+          ],
+        })
+
+        const defaultSalesChannel = await createSalesChannel(
+          { name: "default sales channel" },
+          [product.id]
+        )
+
+        const service = appContainer.resolve(ModuleRegistrationName.STORE)
```
sradevski, line 971: Same here

shahednasser (approved):
LGTM works as expected

integration-tests/http/__tests__/product/store/product.spec.ts
riqwan: yeah, this is coming in from modules, so v2 specific. We surely do, but those will come in when the store products tests are ported over from api

integration-tests/http/__tests__/product/store/product.spec.ts
riqwan: I'll create a ticket to address this when i'm back from vacation 🤞🏻

integration-tests/http/__tests__/product/store/product.spec.ts
riqwan: same

PR #7695, opened 2024-06-12
feat: Revamp of product categories

sradevski:
There are several changes, namely:
1. Refactored the product category interface to follow the general module standards
2. Refactored the product category repository to be much easier to understand
3. Fixes CORE-2287
4. Minor fixes to how ranks were normalized

### DIFF
--- a/integration-tests/http/__tests__/product-category/admin/product-category.spec.ts
+++ b/integration-tests/http/__tests__/product-category/admin/product-category.spec.ts
@@ -684,7 +684,6 @@ medusaIntegrationTestRunner({
             {
               name: "category child 2",
               parent_category_id: productCategoryChild.id,
-              rank: 2,
               description: "category child 2",
             },
             adminHeaders
@@ -697,7 +696,6 @@ medusaIntegrationTestRunner({
             {
               name: "category child 3",
               parent_category_id: productCategoryChild.id,
-              rank: 2,
               description: "category child 3",
             },
             adminHeaders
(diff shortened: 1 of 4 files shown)


packages/modules/product/src/models/product-category.ts
```diff
@@ -130,22 +130,7 @@ class ProductCategory {
...
-        this.parent_category_id
-      )
-    }
-
-    if (parentCategory) {
-      this.mpath = `${parentCategory?.mpath}${this.id}.`
-    } else {
-      this.mpath = `${this.id}.`
-    }
+    this.mpath = `${this.mpath ? this.mpath + "." : ""}${this.id}`
```
carlos-r-l-rodrigues, line 133: if updating the `parent_id`, this will get the incorrect reference no?

carlos-r-l-rodrigues: nevermind. just checked the repository.

carlos-r-l-rodrigues (approved):

olivermrbl (commented):
Very nice work and clean-up!

I think we still have issues with the ranking though. I captured some of the behaviour I was seeing below:

https://github.com/medusajs/medusa/assets/59018053/af5b78ed-1743-4add-a2d2-5817d705065f

sradevski:
@olivermrbl yes I noticed the same thing, I already messaged Kasper in Slack. The rank number is correctly updated as you can see, so it could be a UI bug. If not, there might be something not working correctly in the `find` method.

I just checked, `find` doesn't do any rank sorting currently. I'll do it in a separate PR, this one is big enough as it is.

PR #7699, opened 2024-06-12
fix(dashboard, types): ability to update price list status on admin

riqwan:
what:

- adds ability to update price list status on admin

price-list

RESOLVES CORE-2258

### DIFF
--- a/packages/admin-next/dashboard/src/routes/pricing/common/utils.ts
+++ b/packages/admin-next/dashboard/src/routes/pricing/common/utils.ts
@@ -1,6 +1,6 @@
-import { PriceListDTO, HttpTypes } from "@medusajs/types"
+import { HttpTypes, PriceListDTO } from "@medusajs/types"
 import { TFunction } from "i18next"
-import { PriceListStatus } from "./constants"
+import { PriceListDateStatus, PriceListStatus } from "./constants"
 
 const getValues = (priceList: PriceListDTO) => {
   const startsAt = priceList.starts_at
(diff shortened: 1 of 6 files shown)


packages/admin-next/dashboard/src/routes/pricing/common/utils.ts
```diff
@@ -25,25 +25,30 @@ export const getPriceListStatus = (
...
-  if (isScheduled) {
-    color = "orange"
-    text = t("pricing.status.scheduled")
-  }
+  let status: string = PriceListStatus.ACTIVE
 
   if (isDraft) {
     color = "grey"
     text = t("pricing.status.draft")
+    status = PriceListStatus.DRAFT
```
sradevski, line 33: Why are we using 2 different enums here for draft and the other two states?

shahednasser (approved):
LGTM works as expected

packages/admin-next/dashboard/src/routes/pricing/common/utils.ts
riqwan: One is actual db status, the other is computed status that we display on the UI based on the date

sradevski: @riqwan I see, ideally the UI then defines all statuses, basically a union of the computed + db in one enum

riqwan: Hm, I don't know if that adds much value here for a simple badge. If we use it in other places, sure but that doesn't seem to be the case yet.

PR #7700, opened 2024-06-12
feat(dashboard,js-sdk,types): prices of a product can be deleted in a price list

riqwan:
what:

- allows deleting prices of a product in price list

https://github.com/medusajs/medusa/assets/5105988/3a8aa444-ae5d-4d4f-b1e6-350032d05014

RESOLVES CORE-2295


olivermrbl (approved):
LGTM

PR #7694, opened 2024-06-12
feat(dashboard): display inventory levels in variants table

fPolic:
**What**
- inventory item / kit levels in product variants table
- add inventory actions to variant rows

---

https://github.com/medusajs/medusa/assets/16856471/c988ea18-6f32-45c8-811a-459c16b1001d

PR #7701, opened 2024-06-13
feat: Variant inventory quantity in GET requests

olivermrbl:
**What**
- Add `inventory_quantity` to `GET /admin/products/:id/variants`
- Refactor the utils for `inventory_quantity` computation to be exposed to both `/store` and `/admin` routes


olivermrbl:
@fPolic, in #7694, the available quantity that we show will live top-level on variants after merging this PR. So you can do:
`variant.inventory_quantity`

packages/medusa/src/api/utils/middlewares/products/variant-inventory-quantity.ts
```diff
@@ -0,0 +1,106 @@
...
+  if (!variantIds.length) {
+    return
+  }
+
+  const variantInventoriesMap = await getVariantInventoryItems({
+    req,
+    variantIds,
+  })
+
+  for (const variant of variants || []) {
```
fPolic: nit: if `variants` here can be undefined then we also need check on L:85

fPolic (approved):

packages/medusa/src/api/utils/middlewares/products/variant-inventory-quantity.ts
```diff
@@ -0,0 +1,106 @@
...
+export async function getVariantInventoryItems({
+  req,
+  variantIds,
+  additionalFilters = {},
+  asMap = true,
+}) {
+  const remoteQuery = req.scope.resolve(ContainerRegistrationKeys.REMOTE_QUERY)
+
+  const linkQuery = remoteQueryObjectFromString({
+    entryPoint: "product_variant_inventory_item",
```
adrien2p: suggestion: for the link query, you can use `service: LINKS.ProductVariantInventoryItem`

packages/medusa/src/api/utils/middlewares/products/variant-inventory-quantity.ts
```diff
@@ -0,0 +1,106 @@
...
+
+export const wrapVariantsWithInventoryQuantity = async (
+  req: MedusaRequest,
+  variants: {
+    id: string
+    inventory_quantity?: number
+    manage_inventory?: boolean
+  }[]
+) => {
+  const variantIds = variants.map((variant) => variant.id).flat(1)
```
adrien2p, line 87:
```suggestion
  variants ??= [] // you might need to type it
  const variantIds = variants.map((variant) => variant.id).flat(1)
```

packages/medusa/src/api/utils/middlewares/products/variant-inventory-quantity.ts
adrien2p: This approach should be the preferred one a bit like resolving from ModuleRegistrationName etc wdyt?

adrien2p (approved):

packages/medusa/src/api/utils/middlewares/products/variant-inventory-quantity.ts
olivermrbl: Yes good push, let me add that in

PR #7702, opened 2024-06-13
fix: Sort product categories based on rank

sradevski:


packages/modules/product/src/repositories/product-category.ts
```diff
@@ -86,14 +88,31 @@ export class ProductCategoryRepository extends DALUtils.MikroOrmBaseTreeReposito
...
+  }
+
+  sortCategoriesByRank(categories: ProductCategory[]): ProductCategory[] {
+    const sortedCategories = categories.sort((a, b) => a.rank - b.rank)
+
+    for (const category of sortedCategories) {
+      if (category.category_children) {
+        // All data up to this point is manipulated as an array, but it is a Collection<ProductCategory> type under the hood, so we are casting to any here.
+        category.category_children = this.sortCategoriesByRank(
+          category.category_children as any
```
adrien2p, line 110: Q: I belive you can turn it to an array by doing`[...category.category_children]` or `Array.from(category.category_children)` let me know what you think

adrien2p (approved):

olivermrbl (approved):

packages/modules/product/src/repositories/product-category.ts
sradevski: it is actually an array already (if you see how the data is created in `buildProductCategoriesWithTree`), it's just that the typings are not incorrect. But I'll try to clean it up as you suggested

PR #7704, opened 2024-06-13
Add onDelete hook

thetutlage:
> [!IMPORTANT]
> The following description is incorrect because after further discussions we discarded the concept of hooks and went with cascades. Instead follow this PR https://github.com/medusajs/medusa/pull/7721

Okay, so here is the first iteration of `onDelete` entity hook.

First, I will explain the generate approach towards entity hooks and then we can get into specifics.

- We want to define hooks on an entity using the chainable API.
- Think of hooks as side-effects you want to perform when an action happens.
- In the future, we might allow to callbacks to run any sort of side-effects. But for now, we are opting for a more restrictive API

## onDelete

I have implemented a very simple `onDelete` hook that can be used to cleanup relationships data in the database. This is how it looks like:

```ts
  email: model.text(),
  isVerified: model.boolean(),
})

const user = model
  .define("user", {
    id: model.number(),
    username: model.text(),
  })
  // 👇
  .onDelete({
    remove: ["email"],
  })
```

So essentially. I am telling my entity to remove the `email` anytime the user is deleted. How does it work under the hood is not so important from the DML point of view. That is responsibility of the adapter that converts the DML schema to a more concrete output.

## Mikro ORM adapter

The Mikro ORM adapter (the one that generates Mikro ORM entity from DML) will use the `onDelete` hook and its options to define the following options at Mikro ORM level.

```ts
@OneToOne({
  onDelete: 'cascade',
  cascade: ['soft-remove']
})
```

I took the options to apply from the fulfillment module

Screenshot 2024-06-13 at 4 14 14 PM


packages/core/utils/src/dml/helpers/create-mikro-orm-entity.ts
```diff
@@ -97,13 +98,20 @@ function defineHasOneRelationship(
...
+  const shouldRemoveRelated = !!hooks.deleted?.remove.includes(
+    relationship.name
+  )
+
   OneToOne({
     entity: relatedModelName,
     nullable: relationship.nullable,
     mappedBy: relationship.mappedBy as any,
+    onDelete: shouldRemoveRelated ? "cascade" : undefined,
+    cascade: shouldRemoveRelated ? (["soft-remove"] as any) : undefined,
```
adrien2p: Let's discuss after, but here be careful, the soft remove should be applied on the other side, also, specifying cascade remove the defaults like persists for example

olivermrbl:
@srindom, you might also want to have a look at this

srindom:
Looks cool - one thing I remember which might need to be thought of for the MikroORM adapter is that onDelete: cascade and cascade: "soft-remove" work in opposite directions (unless this has been modified).

This would mean the adapter would need to put the db level cascade on one entity and the soft-remove on the other.

adrien2p:
Yes i already mentioned that to virk and we will discuss it, basically application level cascade in general are applied in the opposite way than db level cascade

packages/core/utils/src/dml/entity.ts
```diff
@@ -7,5 +12,39 @@ import { RelationshipType, SchemaType } from "./types"
...
+    }
+  }
+
+  /**
+   * Delete actions to be performed when the entity is deleted. For example:
+   *
+   * You can configure relationship data to be deleted when the current
+   * entity is deleted.
+   */
+  onDelete(options: { remove: ExtractEntityRelations<Schema> }) {
```
sradevski: I am just thinking if `onDelete` is the best name or not. (this is especially true for the `onCreate` hook, is it before or after creation? It isn't obvious).

Other options:
`onDeleted`
`deleted`
`afterDelete`


What does everyone think? @olivermrbl @adrien2p @srindom

adrien2p: yes indeed, I think the issue here is that we are using the `onDelete` hook name to configure an option related to a cascade which occurs upon true deletion. if you agree, when you are deleting something on cascade, the child must be deleted first and therefore we could assume that it is before, but then let say we open the configuration of hooks to beforeDelete, afterDelete now we can see that configuring a cascade cannot necessarely be represented by the hook configuration.

Cant we separate the two and have the onDelete where we configure the cascade and later on add the notion of `hooks` where you can configure this before/after actions?

olivermrbl: I don't have any strong opinion on this. Think `afterDelete` is also a good option. If we ever want to add a hook for something that happens before, then the explicitness of the timing makes the difference between the two obvious. So I am slightly leaning toward `afterDelete`.

adrien2p: so it would mean that after delete would be able to take the configuration for cascading plus hooks and before delete only for the hooks? doesnt that seems strange?

thetutlage:
I am dealing with a situation here. Lemme share that with all you guys and would love to have your input on the same.

Problem in hand.

We have a relationship that is needed to be configured from two sides. 

- First is the actual relationship registration.
- Second is the hook trying to reach to the related entity relationship.

```ts
const email = model.define('email', () => {
  email: model.string(),
  user: model.belongsTo(() => user, { mappedBy: 'email' })
})

  id: model.number(),
  username: model.text(),
})
```

The above code example, will register a relationship for `Email` entity with Mikro ORM that looks as follows.

```
user: {
  reference: "1:1",
  name: "user",
  entity: "User",
  nullable: false,
  owner: true,
  mappedBy: "email",
},
```

Next, we use a hook called `onDelete` to remove emails when user is deleted. Now this hook should update the relationship metadata registered with TypeORM and add the following property to it.

```ts
const user = model.define("user", {...})
.onDelete({
  remove: 'emails'
})

user: {
  reference: "1:1",
  name: "user",
  entity: "User",
  nullable: false,
  owner: true,
  mappedBy: "email",
  +++
  onDelete: 'cascade'
  +++
}
```

- If we convert the `email` model to Mikro ORM entity first, then the `onDelete` hook will update the metadata.
- If we convert the `user` model to Mikro ORM entity first, then the `onDelete` hook will register the relationship and the `belongsTo` will have to update the metadata.

packages/core/utils/src/dml/entity.ts
sradevski: I agree it is a bit strange, butt I don't think `onDelete` makes it any less strange, it is just naming after all (since onDelete would be the `after` hook in our discussions). Let me know if you have any better suggestions. Also, it's important to not think about DB or mikroORM cascades, but only think of the DML API and how it reads at this point.

adrien2p: I agree on the thinking no worries I am the one that keep repeating it 😂 , I am thinking more the information we want to provide to the user, which is indeed a cascade information that happen when a tuple gets deleted 😂 what I am thinking is that `onDelete` could be the one taking the casacade configuration and all hooks would be `(before/after)Action` that way there is a clear distinction and it is also impossible to have a hook during deletion wdyt?

thetutlage:
Closing in favor of https://github.com/medusajs/medusa/pull/7721

PR #7706, opened 2024-06-13
fix: Sort orders by `display_id` desc by default

olivermrbl:


carlos-r-l-rodrigues:
We don 't have an index in this field. Should we sort by "display_id" desc?

olivermrbl:
Nice catch, yes let's do that

carlos-r-l-rodrigues (approved):

sradevski (approved):

PR #7711, opened 2024-06-13
Chore: Remove main model concept from currency domain

adrien2p:
**What**

Remove main model concept from currency domain

FIXES CORE-2315

### DIFF
--- a/packages/modules/currency/src/services/currency-module-service.ts
+++ b/packages/modules/currency/src/services/currency-module-service.ts
@@ -10,24 +10,24 @@ import {
   ModuleJoinerConfig,
   ModulesSdkTypes,
 } from "@medusajs/types"
-import { ModulesSdkUtils } from "@medusajs/utils"
 
 import { Currency } from "@models"
 import { entityNameToLinkableKeysMap, joinerConfig } from "../joiner-config"
+import { MedusaService } from "@medusajs/utils"
 
 type InjectedDependencies = {
   baseRepository: DAL.RepositoryService
   currencyService: ModulesSdkTypes.IMedusaInternalService<any>
 }
 
-export default class CurrencyModuleService<TEntity extends Currency = Currency>
-  extends ModulesSdkUtils.MedusaService<{
+export default class CurrencyModuleService
+  extends MedusaService<{
     Currency: { dto: CurrencyTypes.CurrencyDTO }
   }>({ Currency }, entityNameToLinkableKeysMap)
   implements ICurrencyModuleService
 {
   protected baseRepository_: DAL.RepositoryService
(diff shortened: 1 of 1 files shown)


adrien2p:
I didn't find anything more relying on it cc @sradevski @olivermrbl

packages/modules/currency/src/services/currency-module-service.ts
```diff
@@ -15,20 +15,15 @@ import { ModulesSdkUtils } from "@medusajs/utils"
...
 }
 
 export default class CurrencyModuleService<TEntity extends Currency = Currency>
-  extends ModulesSdkUtils.MedusaService<
-    CurrencyTypes.CurrencyDTO,
-    {
-      Currency: { dto: CurrencyTypes.CurrencyDTO }
-    }
-  >(Currency, generateMethodForModels, entityNameToLinkableKeysMap)
+  extends ModulesSdkUtils.MedusaService<{
```
sradevski: NIT: Maybe we import `MedusaService` directly, a bit easier to read IMO

packages/modules/currency/src/services/currency-module-service.ts
```diff
@@ -15,20 +15,15 @@ import { ModulesSdkUtils } from "@medusajs/utils"
...
 import { entityNameToLinkableKeysMap, joinerConfig } from "../joiner-config"
 
-const generateMethodForModels = {}
-
 type InjectedDependencies = {
   baseRepository: DAL.RepositoryService
   currencyService: ModulesSdkTypes.IMedusaInternalService<any>
 }
 
 export default class CurrencyModuleService<TEntity extends Currency = Currency>
```
sradevski: Do we need the TEntity extends Currency = Currency now?

packages/modules/currency/src/services/currency-module-service.ts
olivermrbl: +1, think this should be the convention even though it's an extra import

packages/modules/currency/src/services/currency-module-service.ts
adrien2p: That is part of the class, not the base class. But I believe we did that in case the user change the Repository layer, in that case he could specify something different here and be applied to the entire class. But with the DML comming I believe we could think about removing it. Lets maybe tackle this one separately as it can have many impacts on some modules more complex

packages/modules/currency/src/services/currency-module-service.ts
adrien2p: done, for the tests unfortunately pipeline unfortunately we will have to wait for the main branch to be finished

sradevski (approved):
LGTM!

PR #7716, opened 2024-06-13
feat: Update the api key module to follow latest conventions

sradevski:


packages/modules/api-key/src/services/api-key-module-service.ts
```diff
@@ -253,7 +254,8 @@ export default class ApiKeyModuleService<TEntity extends ApiKey = ApiKey>
   }
 
   @InjectManager("baseRepository_")
-  async retrieve(
+  // @ts-expect-error
+  async retrieveApiKey(
```
olivermrbl, line 258: question: Do we need to migrate this service too? It seems we have redundant method implementations no?

olivermrbl (approved):
Also, looks good.

However, it seems all PRs are failing because of the seed script removal:


CleanShot 2024-06-13 at 22 05 08@2x


Probably minor, but something to fix across all PRs before merging.

packages/modules/api-key/src/services/api-key-module-service.ts
sradevski: @olivermrbl they are not redundant in the case of the api keys, since we remove the token before returning

olivermrbl: Ah sorry, I only checked the method signature. Forgot about the token. My bad

PR #7720, opened 2024-06-13
chore: Remove main entity from user

sradevski:
This one has a lot more cleanup in it than the removal of the main entity


packages/modules/user/src/services/user-module.ts
```diff
@@ -62,21 +62,42 @@ export default class UserModuleService<
...
   ): Promise<UserTypes.InviteDTO> {
-    const invite = await this.inviteService_.validateInviteToken(
-      token,
+    const jwtSecret = this.moduleDeclaration["jwt_secret"]
+    if (!jwtSecret) {
+      throw new MedusaError(
+        MedusaError.Types.INVALID_DATA,
+        "No jwt_secret was provided in the UserModule's options. Please add one."
+      )
+    }
```
olivermrbl: thought: We throw this error in two places in critical flows. Would it make sense to check for this in the service constructor and throw at registration-time instead?

So when set on `this.config`

olivermrbl (approved):
Overall, LGTM

packages/modules/user/src/services/user-module.ts
adrien2p: Yes I think it would be better to not get the error too later

adrien2p (approved):

packages/modules/user/src/services/user-module.ts
sradevski: Yes definitely, let me clean it up.

PR #7721, opened 2024-06-14
Add support for cascades to DML

thetutlage:
This PR adds the concept of cascades to DML entities. We should not confuse DML cascades with database cascades, as they are slightly different and looking at them from fresh perspective will help you understand them better.

Fixes: CORE-2299

## Flows from parent to child
DML cascades are defined on the parent entity and they can impact the children relationships. For example: A user entity can define a cascade to remove emails when user is deleted. However, the opposite cannot be defined.

- `hasOne` relationship creates a parent to child relation.
- `hasMany` relationship creates a parent to child relation.
- `belongsTo` relationship creates a child to parent. Hence, cascades are not allowed in this case.

Let's look at some examples.

### Unidirectional relationship
In the following example, we have a relationship defined on only one entity. The `email` does not point back to the `user`.

```ts
  email: model.text(),
  isVerified: model.boolean(),
})

  id: model.number(),
  username: model.text(),
})
// 👇 Cascading delete on child email
.cascades({
  delete: ["email"],
})
```

The Mikro ORM adapter will define the `user -> email` relationship as follows.

```diff
email: {
  reference: "1:1",
  name: "email",
  entity: "Email",
  nullable: false,
  mappedBy: "user",
+ cascade: ["perist", "soft-remove"],
}
```

### Bidirectional relationship
In the following example, both entities are creating a recursive relationship with each other.

```ts
  email: model.text(),
  isVerified: model.boolean(),
  // 👇 Pointing back to user
  user: model.belongsTo(() => user),
})

const user = model
  .define("user", {
    id: model.number(),
    username: model.text(),
  })
  // 👇 Cascading delete on children
  .cascades({
    delete: ["email"],
  })
```

The Mikro ORM adapter will define the same relationship between `user -> email` as earlier. However, this time the `email -> user` relationship will have additional `onDelete: cascade` option.

```diff
user: {
  entity: "User",
  mappedBy: "email",
  name: "user",
  nullable: false,
+ onDelete: "cascade",
  owner: true,
  reference: "1:1",
}
```

### ❌ Not allowed - Attempting to delete parent from child
The following example will give both typing error and runtime exception.

```ts
model
  .define("email", {
    email: model.text(),
    isVerified: model.boolean(),
    user: model.belongsTo(() => user),
  })
  .cascades({
    // @ts-expect-error "User cannot be mentioned in cascades"
    delete: ["user"],
  })
```

### DIFF
--- a/packages/core/utils/src/dml/entity.ts
+++ b/packages/core/utils/src/dml/entity.ts
@@ -42,13 +42,19 @@ export class DmlEntity<
       ExtractEntityRelations<Schema, "hasOne" | "hasMany">
     >
   ) {
-    options.delete?.forEach((relationship) => {
-      if (this.schema[relationship] instanceof BelongsTo) {
-        throw new Error(
-          `Cannot cascade delete "${relationship}" relationship from "${this.name}" entity. Child to parent cascades are not allowed`
-        )
-      }
+    const childToParentCascades = options.delete?.filter((relationship) => {
+      return this.schema[relationship] instanceof BelongsTo
     })
+
+    if (childToParentCascades?.length) {
+      throw new Error(
+        `Cannot cascade delete "${childToParentCascades.join(
+          ", "
+        )}" relationship(s) from "${
+          this.name
+        }" entity. Child to parent cascades are not allowed`
+      )
+    }
 
     this.#cascades = options
     return this
--- a/packages/core/utils/src/dml/helpers/create-mikro-orm-entity.ts
+++ b/packages/core/utils/src/dml/helpers/create-mikro-orm-entity.ts
@@ -337,8 +337,8 @@ export function createMikrORMEntity<T extends DmlEntity<any>>(
   /**
    * Processing schema fields
    */
-  Object.keys(schema).forEach((property) => {
-    const field = schema[property].parse(property)
+  Object.entries(schema).forEach(([name, property]) => {
+    const field = property.parse(name)
     if ("fieldName" in field) {
       defineProperty(MikroORMEntity, field)
     } else {
--- a/packages/core/utils/src/dml/__tests__/entity-builder.spec.ts
+++ b/packages/core/utils/src/dml/__tests__/entity-builder.spec.ts
@@ -1464,7 +1464,7 @@ describe("Entity builder", () => {
           })
 
       expect(defineEmail).toThrow(
-        'Cannot cascade delete "user" relationship from "email" entity. Child to parent cascades are not allowed'
+        'Cannot cascade delete "user" relationship(s) from "email" entity. Child to parent cascades are not allowed'
       )
     })
   })


packages/core/utils/src/dml/entity.ts
```diff
@@ -7,5 +13,44 @@ import { RelationshipType, SchemaType } from "./types"
...
+      ExtractEntityRelations<Schema, "hasOne" | "hasMany">
+    >
+  ) {
+    options.delete?.forEach((relationship) => {
+      if (this.schema[relationship] instanceof BelongsTo) {
+        throw new Error(
+          `Cannot cascade delete "${relationship}" relationship from "${this.name}" entity. Child to parent cascades are not allowed`
+        )
+      }
+    })
```
adrien2p:
```suggestion
const wrongDeleteCascadeRelation = options.delete?.filter((relationship) => {
  return this.schema[relationship] instanceof BelongsTo
})

if (!!wrongDeleteCascadeRelation.length) {
  throw new Error(
    `Cannot cascade delete "${wrongDeleteCascadeRelation.join(
      ", "
    )}" relationship from "${this.name}" entity. Child to parent cascades are not allowed`
  )
}
```

The idea behind this suggestion being that you don't have to run after fixing the first issue and then get the next one, giving the opportunity to the user to fix all of them on the entity at once but also can be easily improve forfuture configuration with little effort. wdyt?

packages/core/utils/src/dml/helpers/create-mikro-orm-entity.ts
```diff
@@ -165,6 +189,7 @@ function defineBelongsToRelationship(
       mapToPk: true,
       fieldName: camelToSnakeCase(`${relationship.name}Id`),
       nullable: relationship.nullable,
+      onDelete: shouldCascade ? "cascade" : undefined,
```
sradevski, line 192: I think the default case should be to `set null`, right? Otherwise you Postgres won't allow you to delete the data (maybe that's how it is already?), since it will reference a non-existent foreign key

packages/core/utils/src/dml/helpers/create-mikro-orm-entity.ts
```diff
@@ -300,12 +337,12 @@ export function createMikrORMEntity<
   /**
    * Processing schema fields
    */
-  Object.keys(entity.schema).forEach((property) => {
-    const field = entity.schema[property].parse(property)
+  Object.keys(schema).forEach((property) => {
```
sradevski: NIT: You can use `.entries` so you don't ned the `schema[property]`

packages/core/utils/src/dml/types.ts
```diff
@@ -108,3 +108,25 @@ export type Infer<T> = T extends DmlEntity<infer Schema>
...
   : never
+
+/**
+ * Extracts names of relationships from a schema
+ */
+export type ExtractEntityRelations<
+  Schema extends Record<string, any>,
+  OfType extends RelationshipTypes
+> = {
+  [K in keyof Schema & string]: Schema[K] extends RelationshipType<any>
```
adrien2p, line 119:
```suggestion
  [K in keyof Schema]: Schema[K] extends RelationshipType<any>
```
K is already typed as string in record, it should not be necessary

sradevski (approved):
Looks very nice, 2 minor comments!

packages/core/utils/src/dml/types.ts
```diff
@@ -108,3 +108,25 @@ export type Infer<T> = T extends DmlEntity<infer Schema>
...
+    ? Schema[K] extends { type: OfType }
+      ? K
+      : never
+    : never
+}[keyof Schema & string][]
```
adrien2p, line 124:
```suggestion
export type ExtractEntityRelations<
  Schema extends Record<string, any>,
  OfType extends RelationshipTypes
> = {
  [K in keyof Schema]: Schema[K] extends RelationshipType<any>
    ? Schema[K] extends { type: OfType }
      ? K
      : never
    : never
}
```

adrien2p (approved):
💪

packages/core/utils/src/dml/helpers/create-mikro-orm-entity.ts
thetutlage: When I looked at our relationships in modules, I either saw `onDelete: "cascade"` or nothing. So not sure if we should define `set null`. Have to verify once

packages/core/utils/src/dml/helpers/create-mikro-orm-entity.ts
thetutlage: Sure thing!

packages/core/utils/src/dml/types.ts
thetutlage: Why did we remove the `[keyof Schema]` at the end after mapping the types 🤔

packages/core/utils/src/dml/entity.ts
thetutlage: Makes sense!

packages/core/utils/src/dml/types.ts
thetutlage: Just tried this in TypeScript playground and seems to expand to `string | number | symbol` when a generic is passed to the type. 

https://www.typescriptlang.org/play/?ts=5.1.6#code/C4TwDgpgBAysBOBLAdgcxgYwBYQLYEMAeAFSggA9gJkATAZygCUIMB7eGwuhFVAGijIArrgBGEeAD5JUALxQA3gCgoUANoBpKCigBrCCFYAzKMQC6ALlOazZStXqCR4+FAD8ULVeQQAbhKUAXzV9QxNzNTMlJSMhZAxgRFZkPQM6QkwcAjsqWgZmNg4uHjQBYTEJSQEtClzHOCQ0TLwiZoJpAApdKw0ASkUVPQA6AjAOjsR+2RllVVVEIeBWABlWAHcJAGF8OggO3sHAg8CgA

packages/core/utils/src/dml/helpers/create-mikro-orm-entity.ts
thetutlage: @adrien2p Any insights on this?

PR #7728, opened 2024-06-14
chore: Migrate payment module main entity

adrien2p:
FIXES CORE-2320


packages/modules/payment/src/services/payment-module.ts
```diff
@@ -822,7 +871,7 @@ export default class PaymentModuleService<
       case PaymentActions.SUCCESSFUL: {
         const [payment] = await this.listPayments(
           {
-            session_id: event.data.resource_id,
+            id: event.data.resource_id,
```
adrien2p: @carlos-r-l-rodrigues I d like your input here :D

carlos-r-l-rodrigues: I believe here it is the `session_id` we want.
On the provider we only work with the session_id, there isn't the notion of "Payment".

sradevski (approved):
LGTM other than the ID change, let's ensure what that data should be

PR #7734, opened 2024-06-15
feat(order): create claim and exchange

carlos-r-l-rodrigues:
Cancel methods and better Typescript Types will come in a next PR


package.json
```diff
@@ -44,8 +44,7 @@
     "resolve-cwd": "^3.0.0",
     "ts-jest": "^29.1.1",
     "ts-node": "^10.9.1",
-    "turbo": "^1.6.3",
-    "typeorm": "^0.3.16"
```
olivermrbl, line 48: 🥳

packages/modules/order/src/utils/transform-order.ts
```diff
@@ -5,18 +5,32 @@ import {
...
   orders.map((order) => {
-    order.items = order.items?.map((orderItem) => {
+    let mainOrder = order
+
+    const isRelatedEntity = options?.entity !== Order
+
+    if (isRelatedEntity) {
+      if (!order.order) {
+        return order
+      }
```
olivermrbl, line 28: question: What case are we covering here?

packages/modules/order/src/utils/transform-order.ts
```diff
@@ -29,30 +43,118 @@ export function formatOrder(
...
+      (item) => item.is_additional_item
+    )
+
+    claim.additional_items.forEach((orderItem) => {
+      const item = claim.order.items?.find(
+        (item) => item.id === orderItem.item_id
+      )
+
+      orderItem.detail = item?.detail
+    })
```
olivermrbl, line 110: question: I am not sure I understand this part. 

First, we filter away `additional_items` for items that are not additional – what items are these then?

Then, we for each additional item, we retrieve its corresponding item on the claim order – wouldn't these be new items and therefore not be part of the order (yet)?

packages/modules/order/src/utils/transform-order.ts
```diff
@@ -29,30 +43,118 @@ export function formatOrder(
...
+  claim.claim_items = claim.claim_items.filter(
+    (item) => !item.is_additional_item
+  )
+  claim.claim_items.forEach((orderItem) => {
+    const item = claim.order.items?.find(
+      (item) => item.id === orderItem.item_id
+    )
+
+    orderItem.detail = item?.detail
+  })
```
olivermrbl, line 126: The same questions I have here

packages/modules/order/src/services/actions/receive-return.ts
```diff
@@ -1,39 +1,44 @@
...
 import { ChangeActionType } from "../../utils"
 
-export async function receiveReturn(
-  this: any,
-  data: OrderTypes.ReceiveOrderReturnDTO,
-  sharedContext?: Context
-) {
-  const returnEntry = await this.retrieveReturn(
-    data.return_id,
+async function retrieveReturn(service, returnId, sharedContext) {
```
olivermrbl: thought: I guess we can just use the service method directly in the module service, now that we no longer decorate with return items. No need for this util

packages/modules/order/src/services/actions/create-return.ts
```diff
@@ -7,43 +7,72 @@ import {
...
-  this: any,
-  data: OrderTypes.CreateOrderReturnDTO,
-  sharedContext?: Context
-) {
-  const order = await this.orderService_.retrieve(
-    data.order_id,
-    {
-      relations: ["items"],
-    },
+async function retrieveOrder(service, orderId, sharedContext) {
```
olivermrbl: Same comment here.

olivermrbl (commented):
Nice work @carlos-r-l-rodrigues – just a few questions

packages/modules/order/src/utils/transform-order.ts
carlos-r-l-rodrigues: Returns, Claims and Exchanges have the entity `order`. Just in case it is not present we don't try to format it.

packages/modules/order/src/utils/transform-order.ts
carlos-r-l-rodrigues: Claimed items and additional items, both are in the same table and just reference to the line item of the order.
Here we just filter and split them by the claimed items and the additional ones.
That filter could be at the SQL level as well, but that would make it more complex.

packages/modules/order/src/services/actions/receive-return.ts
carlos-r-l-rodrigues: Yeah sure, I just split everything and composed the main methods like workflows definition haha...

olivermrbl (approved):

packages/modules/order/src/utils/transform-order.ts
olivermrbl: Got it. I think what seems strange to me is that this piece of code tells me that `claim.additional_items` can have items that are not additional items and have to be filtered out. I am probably just missing some context here, but thought I'd share.

On line 100-102:
```
claim.additional_items = claim.additional_items.filter(
  (item) => item.is_additional_item
)
```

carlos-r-l-rodrigues:
@olivermrbl, I will merge this one to avoid conflicts on the naming conventions @adrien2p is changing. I can push changes in a diff PR if needed

PR #7739, opened 2024-06-16
chore: move to swc/jest

riqwan:
what:

- moves most suites to use swc/jest
- fixes TS strictness that comes with using swc
- fixes unit test script to scope the shards correctly


packages/core/modules-sdk/src/medusa-module.ts
```diff
@@ -576,4 +576,6 @@ export class MedusaModule {
 }
 
 global.MedusaModule ??= MedusaModule
-exports.MedusaModule = global.MedusaModule
+const GlobalMedusaModule = global.MedusaModule as typeof MedusaModule
+
+export { GlobalMedusaModule as MedusaModule }
```
riqwan, line 594: @carlos-r-l-rodrigues can you check if this is correct.

scripts/run-workspace-unit-tests-in-chunks.sh
```diff
@@ -21,5 +21,7 @@ for workspace in $(echo "$workspaces" | jq -r '.[]'); do
   filters+=" --filter=${workspace}"
 done
 
+command="yarn run test $filters"
+echo "$command"
```
carlos-r-l-rodrigues: rm

carlos-r-l-rodrigues (commented):
This looks great man.

scripts/run-workspace-unit-tests-in-chunks.sh
riqwan:
```suggestion
```

sradevski (approved):
Nice, did you notice any speed improvements in the pipeline?

riqwan:
@sradevski I didn't think there would be since the tests affected were fairly small, but I remember being a decent improvement. Lets find out with the new run 😄

riqwan:
@sradevski looks like there a decent improvement

thetutlage:
Breath of fresh air 💨

package.json
```diff
@@ -58,7 +58,7 @@
     "lint:path": "eslint --ignore-path .eslintignore --ext .js,.ts,.tsx",
     "prettier": "prettier",
     "jest": "jest",
-    "test": "turbo run test --concurrency=50% --no-daemon --no-cache --force --filter='./packages/*' --filter='./packages/core/*' --filter='./packages/cli/*' --filter='./packages/modules/*' --filter='./packages/modules/providers/*'",
+    "test": "turbo run test --concurrency=50% --no-daemon --no-cache --force",
```
adrien2p, line 61: Q: I remember we added those filters to be sure it only runs the unit tests from packages but that the filter did not allow for deep regexp, are we sure only the unit tests are ran from the packages directory here?

packages/core/modules-sdk/src/medusa-module.ts
adrien2p: should work the same 👍

adrien2p (approved):
Looks good to me 💪

package.json
riqwan: the .sh script will filter to run all packages in the workspace. Which ones need to be filtered out?

adrien2p: oh ok perfect

PR #7741, opened 2024-06-17
Identify the owner when both sides defines a many to many relationship

thetutlage:
Last week I played a bit with Mikro ORM to understand how its many to many relationship works and this PR aims to output entities as per Mikro ORM expectations. Here's how it works.

Fixes: CORE-2299

### When one side defines the relationship

In this case, Mikro ORM will mark the side defining the relationship as the owner implicitly. In the following example, the `User.teams` is the owner.

```ts
@Entity()
export class User {
  @PrimaryKey()
  declare id: number;

  @Property()
  declare email: string;

  @ManyToMany({
    entity: () => Team,
  })
  teams = new Collection<Team>(this);
}

@Entity()
export class Team {
  @PrimaryKey()
  declare id: number;

  @Property()
  declare name: string;
}
```

### When both side defines the relationship

When both side defines the relationships, then any one side is marked as an owner by some internal runtime logic of Mikro ORM.

```ts
@Entity()
export class User {
  @PrimaryKey()
  declare id: number;

  @Property()
  declare email: string;

  @ManyToMany({
    entity: () => Team,
  })
  teams = new Collection<Team>(this);
}

@Entity()
export class Team {
  @PrimaryKey()
  declare id: number;

  @Property()
  declare name: string;

  @ManyToMany({
    entity: () => User,
  })
  users = new Collection<User>(this);
}
```

### When one side defines the mappedBy

The `mappedBy` side of the relationship will be marked as the owner. Now if the other side uses `mappedBy` or `owner`, then an exception will be raised.

In this case, the other side should either use `inversedBy` or not define any config that results in identifying an owner.

## DML to Mikro ORM generation

Okay, now that we understand how Mikro ORM works. Let's see how we can generate proper relationships using the `DML -> Mikro ORM generator`.

We have the constraint of not introducing `inversedBy` option. Because, DML should be free from Mikro ORM jargon. Because of this, we will have to internally keep a track of `ManyToMany` relationships between two entities and then smartly define either `inversedBy` or `mappedBy` options when defining the relationships.

This is how it works under the hood.

### One side defines the relationship
No special logic needed. We define a standard `ManyToMany` Mikro ORM relationship.

### Both sides defines the relationship without `mappedBy` option
Again, we leave it to Mikro ORM to identify the owner. 

### Relationship on both ends and one side uses the `mappedBy` option
Again, we leave it to Mikro ORM to mark the owner based upon the existence of `mappedBy` option.

### Both sides are using the `mappedBy` option
This is where things gets tricky. Because, we have to convert one of those `mappedBy` to `inversedBy` under the hood.

- We internally keep a track of many to many relationships that uses the `mappedBy` option. For example: If `user -> teams` relationship has a `mappedBy` option, then we will store it internally as `user.teams: isOwner: true`.
- Now when the team model points back to the user model via `team -> users`, we will check if the related entity, which is `user` in this case is already tracked for the `teams` relationship (which comes via mappedBy`), then we will replace `mappedBy` with `inversedBy`

All this might seem complicated (which it is as it took me a whole day to understand). But in implementation, it is quite simple.

### DIFF
--- a/packages/core/utils/src/dml/__tests__/entity-builder.spec.ts
+++ b/packages/core/utils/src/dml/__tests__/entity-builder.spec.ts
@@ -1962,5 +1962,148 @@ describe("Entity builder", () => {
         },
       })
     })
+
+    test("define multiple many to many relationships to the same entity", () => {
+      const model = new EntityBuilder()
+      const team = model.define("team", {
+        id: model.number(),
+        name: model.text(),
+        activeTeamsUsers: model.manyToMany(() => user, {
+          mappedBy: "activeTeams",
+        }),
+        users: model.manyToMany(() => user, { mappedBy: "teams" }),
+      })
+
+      const user = model.define("user", {
+        id: model.number(),
+        username: model.text(),
+        activeTeams: model.manyToMany(() => team, {
+          mappedBy: "activeTeamsUsers",
+        }),
+        teams: model.manyToMany(() => team, { mappedBy: "users" }),
+      })
+
+      const entityBuilder = createMikrORMEntity()
+      const Team = entityBuilder(team)
+      const User = entityBuilder(user)
+
+      expectTypeOf(new User()).toMatchTypeOf<{
+        id: number
+        username: string
+        teams: EntityConstructor<{
+          id: number
+          name: string
+          users: EntityConstructor<{
+            id: number
+            username: string
+          }>
+        }>
+        activeTeams: EntityConstructor<{
+          id: number
+          name: string
+          users: EntityConstructor<{
+            id: number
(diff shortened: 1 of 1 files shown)


olivermrbl (approved):
Nice one! 💪

riqwan (approved):
This is a masterclass PR! 🔥 

Q: Are we going to support many to many through pivot entity?

thetutlage:
> Are we going to support many to many through pivot entity?

Not yet. Keeping the scope as simple as possible for now.

thetutlage:
Now I need @adrien2p approval too before I merge it

PR #7742, opened 2024-06-17
chore: Migrate pricing module main entity

adrien2p:
FIXES CORE-2307


olivermrbl (approved):
LGTM!

carlos-r-l-rodrigues:
We'll have to update the Remote Joiner callback to always add the entity name as a suffix.

adrien2p:
> We'll have to update the Remote Joiner callback to always add the entity name as a suffix.

@carlos-r-l-rodrigues there is a dedicated ticket for that part. The util will be applied to all the modules and the joiner config will be auto generated

PR #7746, opened 2024-06-17
feat(dashboard,types,promotion,medusa): hide fields on promotions depending on templates

riqwan:
**what:**

- hides different fields depending on the chosen template
- remove operator values API
- fixes to edit promotion rules
- make currency optional for promotion

RESOLVES CORE-2297


riqwan:
@olivermrbl can you run this and see what you think about the hidden fields?

PS: not ready for review yet, just testing

olivermrbl:
> @olivermrbl can you run this and see what you think about the hidden fields?
> 
> PS: not ready for review yet, just testing

@riqwan, this is exactly what I had in mind – super nice. We might need to rework the wording of the different sections + tooltips, but that's a separate thing that can go in a follow-up PR.

packages/admin-next/dashboard/src/routes/promotions/common/edit-rules/components/edit-rules-wrapper/edit-rules-wrapper.tsx
```diff
@@ -77,14 +85,14 @@ export const EditRulesWrapper = ({
...
-              values: rule.operator === "eq" ? rule.values[0] : rule.values,
+              values: rule.values,
             } as any
           }),
         }))
 
       rulesToRemove?.length &&
         (await removePromotionRules({
-          rule_ids: rulesToRemove.map((r) => r.id!),
+          rule_ids: rulesToRemove.filter((r) => !!r.id).map((r) => r.id!),
```
adrien2p:
```suggestion
          rule_ids: rulesToRemove.map((r) => r.id).filter(Boolean),
```

packages/admin-next/dashboard/src/routes/promotions/common/edit-rules/components/rules-form-field/rules-form-field.tsx
```diff
@@ -26,54 +26,49 @@ type RulesFormFieldType = {
...
 
-  const promotionType: string = useWatch({
-    control: form.control,
-    name: "type",
-  })
+  const promotionType: string =
+    useWatch({
+      control: form.control,
+      name: "type",
+    }) || promotion?.type
```
adrien2p: Q: why dont we use default value in the useWatch?

packages/modules/promotion/src/migrations/Migration20240617102917.ts
```diff
@@ -0,0 +1,22 @@
+import { Migration } from "@mikro-orm/migrations"
+
+export class Migration20240617102917 extends Migration {
```
adrien2p, line 3: suggestion: I dont see the snapshot being updated, would that create issue if we generate another migration?

adrien2p (approved):

packages/admin-next/dashboard/src/routes/promotions/common/edit-rules/components/rules-form-field/rules-form-field.tsx
riqwan: because I didn't know we could do that 😄 nice!

adrien2p: ahah no problem, I thought I might be missing a behavior 😄

packages/modules/promotion/src/migrations/Migration20240617102917.ts
riqwan: This migration file doesn't change anything in the schema because the schema was already changed. I think a migration file got deleted at some point, but the snapshot remained in the updated state. This file brings it to its correct form.

adrien2p: oh got it 💪

olivermrbl:
Will just give it a final spin before we merge, if that's cool 🙏

olivermrbl (approved):

olivermrbl:
@riqwan, added a comment in Slack. It would be great to add the template headers to the flow, so users won't get confused about what type of promotion they are creating

PR #7747, opened 2024-06-17
feat(dashboard): Add Optimistic UI to category rank form + style updates

kasperkristensen:
**What**
- Makes rank updates optimistic, meaning that we override the local state (ranking) with what we expect the outcome of the request to be. If a request fails then we revert to the last known server state.
- Updates the style of dragged items.
- Fixes an issue where the tree would flicker when submitting the create form.


olivermrbl:
Damn, this is smooth. 

+1 for more optimistic UI plz 🔥

olivermrbl (approved):
LGTM

PR #7748, opened 2024-06-17
chore(payment-stripe): smallest unit

carlos-r-l-rodrigues:


packages/modules/providers/payment-stripe/src/utils/get-smallest-unit.ts
```diff
@@ -0,0 +1,67 @@
+import { BigNumberInput } from "@medusajs/types"
+import { BigNumber, MathBN } from "@medusajs/utils"
+
+function getCurrencyMultiplier(currency) {
+  const currencyMultipliers = {
```
riqwan, line 5: q: should this come in as a config? that should potentially come in from the currency module?

packages/modules/providers/payment-stripe/src/utils/get-smallest-unit.ts
```diff
@@ -0,0 +1,67 @@
...
+    if (value.includes(currency)) {
+      power = parseInt(key, 10)
+      break
+    }
+  }
+  return Math.pow(10, power)
+}
+
+/**
+ * Converts an amount to the format required by Stripe based on currency.
```
riqwan, line 39: q: is this only for stripe or are they used by other providers as well?

carlos-r-l-rodrigues:
@olivermrbl @adrien2p perhaps we can remove the concept of "decimal_digits" and "rounding" to the currency module.
that was only used for this Stripe conversions.

packages/modules/providers/payment-stripe/src/utils/get-smallest-unit.ts
carlos-r-l-rodrigues: This is very specific for this provider, so I was thinking even about removing these configs from the currency module.
Any suggestion where to place such config? Because if this changes, the provider very likely has to be updated.

riqwan: I see, if those aren't used for anything else, I don't see a reason why the currency module should hold this info. 

wrt to config, we already support something like this right? By default the module does a power of 2, unless specified in the config like this:

```
[Modules.PAYMENT]: {
  resolve: "@medusajs/payment",
  options: {
    providers: [
      {
        resolve: "@medusajs/stripe",
        options: {
          multipliers: {...},
        },
      },
    ],
  },
},
```

riqwan: If this doesn't change between users, I don't know if we even need to do this. We could just have an exhaustive list

carlos-r-l-rodrigues: The plugin has to support all the list as in the docs. Open it to be configured would add little value to it, as only new added currencies would be affected. A very edge case that would require us to update the plugin,

If some day stripe supports crypto, very likely all that will change as this smallest unit doesn't make much sense.

carlos-r-l-rodrigues: > If this doesn't change between users, I don't know if we even need to do this. We could just have an exhaustive list

This doesn't change. But I also wouldn't add the list of 2 digits.

riqwan: Got it, makes sense. 💪🏻

riqwan (approved):
lgtm!

olivermrbl (approved):

PR #7751, opened 2024-06-17
fix(core-flows): delete existing payment session before creating new

riqwan:
what:

Multiple active sessions are used for split payments. We don't currently support split payments. Until we have that implemented, this change is to ensure that whenever we create a new payment session, we delete an existing session if present. 

RESOLVES CORE-2284


packages/core/core-flows/src/payment-collection/workflows/create-payment-session.ts
```diff
@@ -41,7 +40,17 @@ export const createPaymentSessionsWorkflow = createWorkflow(
...
 
-    const created = createPaymentSessionStep(paymentSessionInput)
+    // Note: We are deleting an existing active session before creating a new one
+    // for a payment collection as we don't support split payments at the moment.
+    // When we are ready to accept split payments, this along with other workflows
+    // need to be handled correctly
+    const [created] = parallelize(
+      createPaymentSessionStep(paymentSessionInput),
+      deletePaymentSessionStep({
+        payment_session_id: paymentCollection.payment_sessions?.[0].id,
```
carlos-r-l-rodrigues: We can pass all the IDs here to delete them all.
In case we support multiple, all the sessions have to be created at the same time anyway, therefore all the existing ones can be removed.

packages/core/core-flows/src/payment-collection/steps/delete-payment-sessions.ts
```diff
@@ -0,0 +1,101 @@
...
+  PaymentSessionDTO,
+} from "@medusajs/types"
+import { ContainerRegistrationKeys } from "@medusajs/utils"
+import { StepResponse, createStep } from "@medusajs/workflows-sdk"
+
+interface StepInput {
+  ids: string[]
+}
+
+// Note: This workflow should not be used alone as it doesn't consider a revert
```
carlos-r-l-rodrigues:
```suggestion
// Note: This step should not be used alone as it doesn't consider a revert
```

carlos-r-l-rodrigues: Is this comment still relevant?

carlos-r-l-rodrigues (approved):
LGTM

packages/core/core-flows/src/payment-collection/steps/delete-payment-sessions.ts
riqwan: yes, because the next step in the workflow will revert this if something goes wrong.

olivermrbl (approved):

PR #7752, opened 2024-06-17
chore(orchestration): remote joiner - relationships as map

carlos-r-l-rodrigues:


olivermrbl (approved):
LGTM, just out of curiosity, what sparked the need for the change?

carlos-r-l-rodrigues:
Just to use the right data format here. The number of relationships can grow significantly.

PR #7756, opened 2024-06-18
Add support for indexes

thetutlage:
Adding support for basic indexes. We are ignoring composite indexes for now, but we can also add them as needed.

FIXES CORE-2281


olivermrbl (approved):
Nice work!

PR #7758, opened 2024-06-18
fix: Medusa service base method transaction

adrien2p:
**What**
Remove transaction decorator from the base medusa service method, the transaction will always be coming from the shared context. It fixes the issue that when you consume a base method directly it will return a proper tuple from the DB instead of the one from the entity map cc @VariableVic 

**NOTE**
This pr also fix some categories issues in the product module which was preventing the tests from working. if @sradevski you could have a look later, in the mean time we can still merge it

FIXES CORE-2342


olivermrbl (approved):

adrien2p:
@olivermrbl if you could have another review on it as I had to fix some part of the product module

olivermrbl (approved):
LGTM

PR #7759, opened 2024-06-18
feat: defineJoinerConfig

adrien2p:
**What**
Create `defineJoinerConfig` util and apply it to the module. By default, the util will load all models exported from the module located inside the models directory, it can be configured if necessary

FIXES CORE-2329

### DIFF
--- a/packages/core/types/src/file/mutations.ts
+++ b/packages/core/types/src/file/mutations.ts
@@ -15,5 +15,5 @@ export interface CreateFileDTO {
   /**
    * The file content as a binary-encoded string
    */
-  content: string | Buffer
+  content: string
 }
--- a/packages/core/types/src/file/provider.ts
+++ b/packages/core/types/src/file/provider.ts
@@ -64,7 +64,7 @@ export type ProviderUploadFileDTO = {
   /**
    * The file content as a binary-encoded string
    */
-  content: string | Buffer
+  content: string
 }
 
 export interface IFileProvider {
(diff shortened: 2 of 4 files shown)


packages/modules/api-key/src/joiner-config.ts
```diff
@@ -1,31 +1,11 @@
...
-  primaryKeys: ["id"],
-  linkableKeys: LinkableKeys,
-  alias: [
-    {
-      name: ["api_key", "api_keys"],
-      args: { entity: ApiKey.name },
-    },
-  ],
-} as ModuleJoinerConfig
+export const entityNameToLinkableKeysMap: MapToConfig =
```
olivermrbl, line 10: thought: Can we automagically export the joiner config? It would be great to not require an explicit export in the module. Similar to the migration helpers.

olivermrbl (commented):
Really nice 🔥

packages/modules/api-key/src/joiner-config.ts
adrien2p: yes I can move it to the service itself if thats what you mean?

adrien2p: or if you are talking about what you mention about handling it under the hood, then it wont be part of this pr yet, it requires a bit more work. we still need to discuss some point and I prefer to not extend the scope of the main pr

olivermrbl: > handling it under the hood

Yeah, this is what I had in mind. Totally fine to save this for a follow-up PR. I was just curious to hear if this was in the pipeline of improvements.

adrien2p: it is in the pipeline but not this one, since it will touch the internal loading mechanism I prefer to keep it separated after we merged this one, do we have a ticket for it or should I create on in and in which project?

olivermrbl: I'll create one 👍

PR #7760, opened 2024-06-18
fix: extend timeout on flaky spec

riqwan:


adrien2p (approved):
ahah funny https://github.com/medusajs/medusa/pull/7758 💪

PR #7762, opened 2024-06-18
chore: Migrate order module main entity

adrien2p:
FIXES CORE-2343


adrien2p:
cc @carlos-r-l-rodrigues I tried to do a lightweight migration to reduce the friction in case we have conflicts

olivermrbl (approved):

PR #7763, opened 2024-06-18
chore: make apis nullable

riqwan:
what:

- makes top level attributes of each object an optional field in the http layer where possible

RESOLVES CORE-2229


packages/medusa/src/api/admin/fulfillments/validators.ts
```diff
@@ -32,13 +32,13 @@ export const AdminCreateFulfillment = z.object({
...
-  delivered_at: z.coerce.date().optional().nullable(),
-  canceled_at: z.coerce.date().optional().nullable(),
-  metadata: z.record(z.unknown()).optional().nullable(),
+  shipping_option_id: z.string().nullish(),
+  data: z.record(z.unknown()).nullable(),
+  packed_at: z.coerce.date().nullish(),
+  shipped_at: z.coerce.date().nullish(),
+  delivered_at: z.coerce.date().nullish(),
+  canceled_at: z.coerce.date().nullish(),
+  metadata: z.record(z.unknown()).nullable(),
```
olivermrbl, line 41: why are soem of these not `nullish`? `e.g. `metadata` is not required

olivermrbl (approved):
Hard to know if we have covered all by looking at the PR, but it is definitely an improvement. Let's get this merged and we can address the rest of the missing or incorrect configurations on the fly.

packages/medusa/src/api/admin/fulfillments/validators.ts
riqwan: Good catch, replace all screw up. Some modules are setup for metadata to not be nullable. Shall we make them all `metadata: z.record(z.unknown()).nullish()`?

olivermrbl: I would say so yeah

packages/medusa/src/api/admin/campaigns/validators.ts
```diff
@@ -61,10 +61,10 @@ export const AdminCreateCampaign = z
   .object({
     name: z.string(),
     campaign_identifier: z.string(),
-    description: z.string().optional(),
```
sradevski, line 64: I don't think we need to explicitly set fields to `null` on create, leaving them undefined is better IMO. This also applies to the typings. WDYT?

riqwan: You don't have to set this to null explicitly, it accepts undefined as well. It just doesn't throw an error if you do send null

PR #7765, opened 2024-06-18
Add support for created and updated at timestamps

thetutlage:
After this PR all DML entities will implicitly have `created_at` and `updated_at` timestamps. There is no way to opt out of it.


adrien2p (approved):

packages/core/utils/src/dml/helpers/create-mikro-orm-entity.ts
```diff
@@ -92,6 +125,11 @@ export function createMikrORMEntity() {
     MikroORMEntity: EntityConstructor<any>,
     field: PropertyMetadata
   ) {
+    if (SPECIAL_PROPERTIES[field.fieldName]) {
+      SPECIAL_PROPERTIES[field.fieldName](MikroORMEntity, field)
+      return
+    }
```
olivermrbl, line 131: thought: Should we either log or throw to inform the end user of this case?

olivermrbl (approved):

packages/core/utils/src/dml/helpers/create-mikro-orm-entity.ts
thetutlage: Naah, this is when we are generating Mikro ORM entities from DML. I think the exception should be when DML receives it as part of the schema.

Let me add it there.

adrien2p: I dont think we have to log it as it will part of the documentation

thetutlage: @adrien2p Throwing exception?

adrien2p: I believe we can throw an exception only if the user provides it as well, as you mention

joekendal:
This doesn't seem to work when updating objects afterwards

PR #7769, opened 2024-06-18
fix(medusa): Load subscribers and scheduled jobs in worker mode

olivermrbl:
Only load subscribers and scheduled jobs in worker and shared mode


srindom (approved):
LGTM

PR #7771, opened 2024-06-18
chore(orchestration): idempotent

carlos-r-l-rodrigues:
What:
`idempotent` flag on workflows.
If set to true, before executing the workflow, it tries to load from memory/Redis and then from Postgres. If available the same transaction is returned making the flow idempotent.
The `idempotency-key` is the transaction id itself.

### DIFF
--- a/packages/core/orchestration/src/transaction/types.ts
+++ b/packages/core/orchestration/src/transaction/types.ts
@@ -119,7 +119,7 @@ export type TransactionModelOptions = {
   storeExecution?: boolean
 
   /**
-   * If true, the workflow will be executed only once with the same transaction id.
+   * If true, the workflow will use the transaction ID as the key to ensure only-once execution
    */
   idempotent?: boolean
 


packages/core/orchestration/src/transaction/transaction-orchestrator.ts
```diff
@@ -798,7 +814,7 @@ export class TransactionOrchestrator extends EventEmitter {
       flow.state = TransactionState.INVOKING
       flow.startedAt = Date.now()
 
-      if (this.options?.store) {
+      if (this.getOptions().store) {
```
riqwan, line 817: is store guaranteed here?

```suggestion
      if (this.getOptions()?.store) {
```

riqwan (approved):
clean idea! nice 🎉

packages/core/orchestration/src/transaction/types.ts
```diff
@@ -118,6 +118,11 @@ export type TransactionModelOptions = {
    */
   storeExecution?: boolean
 
+  /**
+   * If true, the workflow will be executed only once with the same transaction id.
```
sradevski:
```suggestion
   * If true, the workflow will use the transaction ID as the key to ensure only-once execution
```

sradevski (approved):
LGTM 👍

packages/core/orchestration/src/transaction/transaction-orchestrator.ts
riqwan: ignore this. I read it wrong :D

PR #7773, opened 2024-06-19
Allow entities to contain pg schema name in their name

thetutlage:
This PR adds support for defining entity names that also contains the PostgreSQL schema name inside them.

FIXES CORE-2348

In the following example

- The pg schema name will be `platform`.
- The table name will be `users`.
- The model class name will be `User`.

```ts
model.define('platform.user', {
})
```

### DIFF
--- a/packages/core/utils/src/dml/__tests__/entity-builder.spec.ts
+++ b/packages/core/utils/src/dml/__tests__/entity-builder.spec.ts
@@ -667,7 +667,6 @@ describe("Entity builder", () => {
       }>()
 
       const metaData = MetadataStorage.getMetadataFromDecorator(User)
-      console.log(metaData)
       expect(metaData.className).toEqual("User")
       expect(metaData.path).toEqual("User")
 
@@ -768,14 +767,14 @@ describe("Entity builder", () => {
 
       expect(metaData.indexes).toEqual([
         {
-          name: "IDX_users_id",
+          name: "IDX_user_id",
           expression:
-            'CREATE INDEX IF NOT EXISTS "IDX_users_id" ON "users" (id) WHERE deleted_at IS NULL',
+            'CREATE INDEX IF NOT EXISTS "IDX_user_id" ON "user" (id) WHERE deleted_at IS NULL',
         },
         {
-          name: "IDX_users_email",
+          name: "IDX_user_email",
           expression:
-            'CREATE UNIQUE INDEX IF NOT EXISTS "IDX_users_email" ON "users" (email) WHERE deleted_at IS NULL',
+            'CREATE UNIQUE INDEX IF NOT EXISTS "IDX_user_email" ON "user" (email) WHERE deleted_at IS NULL',
         },
       ])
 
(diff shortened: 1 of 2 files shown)


packages/core/utils/src/dml/__tests__/entity-builder.spec.ts
```diff
@@ -646,6 +646,102 @@ describe("Entity builder", () => {
...
+        id: number
+        username: string
+        email: string
+        created_at: Date
+        updated_at: Date
+        deleted_at: Date | null
+      }>()
+
+      const metaData = MetadataStorage.getMetadataFromDecorator(User)
+      console.log(metaData)
```
adrien2p: suggestion: we can pobably rm the log

packages/core/utils/src/dml/__tests__/entity-builder.spec.ts
```diff
@@ -754,6 +850,112 @@ describe("Entity builder", () => {
...
+
+      const metaData = MetadataStorage.getMetadataFromDecorator(User)
+      expect(metaData.className).toEqual("User")
+      expect(metaData.path).toEqual("User")
+
+      expect(metaData.indexes).toEqual([
+        {
+          name: "IDX_users_id",
+          expression:
+            'CREATE INDEX IF NOT EXISTS "IDX_users_id" ON "platform"."users" (id) WHERE deleted_at IS NULL',
```
adrien2p: todo: the table name should be singular, this is the convention we went with

packages/core/utils/src/dml/helpers/create-mikro-orm-entity.ts
```diff
@@ -170,7 +174,7 @@ export function createMikrORMEntity() {
 
       const providerEntityIdIndexStatement = createPsqlIndexStatementHelper({
         name,
-        tableName,
+        tableName: pgSchema ? `${pgSchema}"."${tableName}` : tableName,
```
adrien2p, line 177: Q: shouldnt it be like this?

```suggestion
        tableName: pgSchema ? `"${pgSchema}"."${tableName}"` : tableName,
```

packages/core/utils/src/dml/__tests__/entity-builder.spec.ts
thetutlage: Ohhh. Left it there by mistake

packages/core/utils/src/dml/__tests__/entity-builder.spec.ts
thetutlage: Ohh okay! Usually I have seen it to be plural

packages/core/utils/src/dml/helpers/create-mikro-orm-entity.ts
thetutlage: Naah, the surrounding quotes are added by Mikro ORM already. If you look the output in the test

packages/core/utils/src/dml/__tests__/entity-builder.spec.ts
adrien2p: yeah me too, but medusa convention is singular :D

packages/core/utils/src/dml/helpers/create-mikro-orm-entity.ts
adrien2p: yes indeed it is done by our util itself, not mikro orm. one question, should we instead handle the quotes in the util directly. It seems unintuitive to have to know the underlying impl to use the util

thetutlage: Yup. that makes sense. But let's do it in next PR.

adrien2p: deal 👍

adrien2p (approved):

PR #7774, opened 2024-06-19
Handle embedded pg schema name inside the table name when generating indexes

thetutlage:
This PR updates the implementation of `createPsqlIndexStatementHelper` method to handle use-cases where the schema name is embedded within the table name. Such as `public.users`.

FIXES CORE-2351

### DIFF
--- a/packages/core/utils/src/dml/__tests__/entity-builder.spec.ts
+++ b/packages/core/utils/src/dml/__tests__/entity-builder.spec.ts
@@ -2926,6 +2926,187 @@ describe("Entity builder", () => {
         },
       })
     })
+
+    test("define relationships between cross pg schemas entities", () => {
+      const model = new EntityBuilder()
+
+      const email = model.define("platform.email", {
+        email: model.text(),
+        isVerified: model.boolean(),
+        user: model.belongsTo(() => user),
+      })
+
+      const user = model.define("public.user", {
+        id: model.number(),
+        username: model.text(),
+        email: model.hasOne(() => email),
+      })
+
+      const entityBuilder = createMikrORMEntity()
+      const User = entityBuilder(user)
+      const Email = entityBuilder(email)
+
+      expectTypeOf(new User()).toMatchTypeOf<{
+        id: number
+        username: string
+        deleted_at: Date | null
+        email: EntityConstructor<{
+          email: string
+          isVerified: boolean
+          deleted_at: Date | null
+          user: EntityConstructor<{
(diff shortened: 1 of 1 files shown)


packages/core/utils/package.json
```diff
@@ -46,7 +46,7 @@
     "prepublishOnly": "cross-env NODE_ENV=production tsc --build",
     "build": "rimraf dist && tsc --build",
     "watch": "tsc --build --watch",
-    "test": "jest --silent --bail --maxWorkers=50% --forceExit --testPathIgnorePatterns='/integration-tests/' -- src/**/__tests__/**/*.ts",
+    "test": "jest --silent=false --bail --maxWorkers=50% --forceExit --testPathIgnorePatterns='/integration-tests/' -- src/**/__tests__/**/*.ts",
```
adrien2p: suggestion: can you revert this please 😄

packages/core/utils/src/dml/__tests__/entity-builder.spec.ts
```diff
@@ -667,8 +667,8 @@ describe("Entity builder", () => {
       }>()
 
       const metaData = MetadataStorage.getMetadataFromDecorator(User)
-      expect(metaData.className).toEqual("User")
-      expect(metaData.path).toEqual("User")
+      expect(metaData.className).toEqual("PublicUser")
+      expect(metaData.path).toEqual("PublicUser")
```
adrien2p: Q: I am not sure the name should include the schema, I believe it has many other implication that we can discuss

packages/core/utils/package.json
thetutlage: haha. Just curious, why do we silent logs by default?

packages/core/utils/src/dml/__tests__/entity-builder.spec.ts
thetutlage: Sure. Earlier I removed it, but then I thought why not?

packages/core/utils/package.json
adrien2p: because it was creating a lot of noise in the pipeline and becoming infinite scroll, sometimes bugging the view 😂

adrien2p: it also slowdown the pipeline

riqwan (approved):

adrien2p (approved):

olivermrbl (approved):

PR #7775, opened 2024-06-19
Add support for id property type

thetutlage:
FIXES CORE-2350

The `id` property is a special property that leads to generation of Mikro ORM hooks that runs on `onCreate` and `onInit`. 

The idea is to simplify the generation id and mark it is primary using a branded type.


packages/core/utils/package.json
```diff
@@ -46,7 +46,7 @@
     "prepublishOnly": "cross-env NODE_ENV=production tsc --build",
     "build": "rimraf dist && tsc --build",
     "watch": "tsc --build --watch",
-    "test": "jest --silent --bail --maxWorkers=50% --forceExit --testPathIgnorePatterns='/integration-tests/' -- src/**/__tests__/**/*.ts",
+    "test": "jest --silent=false --bail --maxWorkers=50% --forceExit --testPathIgnorePatterns='/integration-tests/' -- src/**/__tests__/**/*.ts",
```
adrien2p: rm 😄

packages/core/utils/src/dml/helpers/create-mikro-orm-entity.ts
```diff
@@ -170,6 +174,39 @@ export function createMikrORMEntity() {
       return
     }
 
+    /**
+     * Defining an id property
+     */
+    if (field.dataType.name === "id") {
+      const IdDecorator = field.dataType.options?.primaryKey
```
adrien2p, line 181: q: would it have the same effect to use the property decorator with primary: true?

packages/core/utils/src/dml/helpers/create-mikro-orm-entity.ts
```diff
@@ -170,6 +174,39 @@ export function createMikrORMEntity() {
...
+            nullable: field.nullable,
+          })
+
+      IdDecorator(MikroORMEntity.prototype, field.fieldName)
+
+      /**
+       * Hook to generate entity within the code
+       */
+      MikroORMEntity.prototype.generateId = function () {
+        this.id = generateEntityId(this.id, "acc")
```
thetutlage: @adrien2p What should be the prefix of id generation. Do we follow some convention?

adrien2p: todo: the prefix should be configuration, also, if not provided maybe we can infer a default value that depends on the model name as well wdyt?

adrien2p: ahah I ve put a comment about it, we generally use a short name that correspond to the model name and that is meaningful a bit. so it should be configurable and if not provided we can infer it from the model name I would say. Something along the line: transform to snake case, split on `_`, if length 1 then the first 3 letters, if length 2 then the first two letters of each and if more then the first letter of each. wdyt?

packages/core/utils/package.json
thetutlage: I think I should write a commit hook to fail if this substring exists

packages/core/utils/src/dml/helpers/create-mikro-orm-entity.ts
thetutlage: No idea. I decided to stick with what rest of the codebase is using.

adrien2p: Lets go with that and de can revisit this later it is nit 😅

adrien2p (approved):

PR #7776, opened 2024-06-19
chore: export DML builders + fix default undefined values in SQL

riqwan:
**what:**

- exports DML builders through utils package + as a bundle
- fix generated SQL from setting `set default undefined` in SQL
- upgrades typescript to 5+ as the dml utils uses TS 5 features

### DIFF
--- a/packages/core/utils/src/dml/helpers/create-mikro-orm-entity.ts
+++ b/packages/core/utils/src/dml/helpers/create-mikro-orm-entity.ts
@@ -16,6 +16,7 @@ import {
   camelToSnakeCase,
   createPsqlIndexStatementHelper,
   generateEntityId,
+  isDefined,
   pluralize,
   toCamelCase,
 } from "../../common"
@@ -174,7 +175,7 @@ export function createMikrORMEntity() {
          * the database schema SQL. Conditionally add it here to prevent undefined
          * from being set as default value in SQL.
          */
-        ...(field.defaultValue && { default: field.defaultValue }),
+        ...(isDefined(field.defaultValue) && { default: field.defaultValue }),
       })(MikroORMEntity.prototype, field.fieldName)
       return
     }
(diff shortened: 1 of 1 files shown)


packages/core/utils/src/dml/helpers/create-mikro-orm-entity.ts
```diff
@@ -169,7 +169,12 @@ export function createMikrORMEntity() {
       Enum({
         items: () => field.dataType.options!.choices,
         nullable: field.nullable,
-        default: field.defaultValue,
+        /**
+         * MikroORM does not ignore undefined values for default when generating
+         * the database schema SQL. Conditionally add it here to prevent undefined
+         * from being set as default value in SQL.
+         */
+        ...(field.defaultValue && { default: field.defaultValue }),
```
adrien2p: todo: if the defaultValue is a false or null or '' then this wont work. Is would suggest to use the isDefined util here wdyt?

packages/core/utils/src/dml/helpers/create-mikro-orm-entity.ts
```diff
@@ -217,7 +222,12 @@ export function createMikrORMEntity() {
       columnType,
       type: propertyType,
       nullable: field.nullable,
-      default: field.defaultValue,
+      /**
+       * MikroORM does not ignore undefined values for default when generating
+       * the database schema SQL. Conditionally add it here to prevent undefined
+       * from being set as default value in SQL.
+       */
+      ...(field.defaultValue && { default: field.defaultValue }),
```
adrien2p: same comment as the previous one :)

packages/core/utils/src/dml/helpers/create-mikro-orm-entity.ts
riqwan: good catch!

adrien2p (approved):

sradevski (approved):

thetutlage (approved):

PR #7779, opened 2024-06-20
Add support for pivot table and entity in manyToMany relationships

thetutlage:
FIXES CORE-2349

This PR adds support for defining custom `pivotTable` and `pivotEntity` for a `manyToMany` relationship

### DIFF
--- a/packages/admin-next/dashboard/src/components/layout/settings-layout/settings-layout.tsx
+++ b/packages/admin-next/dashboard/src/components/layout/settings-layout/settings-layout.tsx
@@ -54,10 +54,6 @@ const useSettingRoutes = (): NavItemProps[] => {
         to: "/settings/product-types",
       },
       {
-        label: t("shippingProfile.domain"),
-        to: "/settings/shipping-profiles",
-      },
-      {
         label: t("stockLocations.domain"),
         to: "/settings/locations",
       },
--- a/packages/admin-next/dashboard/src/i18n/translations/en.json
+++ b/packages/admin-next/dashboard/src/i18n/translations/en.json
@@ -804,6 +804,14 @@
         "shipping": "Shipping was successfully enabled."
       }
     },
+    "sidebar": {
+      "shippingConfiguration": "Shipping configuration",
+      "shippingProfiles": "Shipping profiles",
+      "shippingProfilesDesc": "Shipping rules for different types of products",
+      "shippingOptionTypes": "Shipping Option Types",
+      "shippingOptionTypesDesc": "Group options based on characteristic"
+
+    },
     "salesChannels": {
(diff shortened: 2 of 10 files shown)


olivermrbl (approved):
LGTM, nice work

riqwan (approved):
🔥

thetutlage:
I will wait for @adrien2p to approve it as well :)

packages/core/utils/src/dml/__tests__/entity-builder.spec.ts
```diff
@@ -3023,7 +3023,7 @@ describe("Entity builder", () => {
       const user = model.define("user", {
         id: model.number(),
         username: model.text(),
-        email: model.manyToMany(() => email),
+        email: model.belongsTo(() => email),
```
adrien2p, line 3026: nice catch 🙏

packages/core/utils/src/dml/entity-builder.ts
```diff
@@ -162,7 +162,13 @@ export class EntityBuilder {
    *   relationship requires a pivot table to establish a many to many
    *   relationship between two entities
    */
-  manyToMany<T>(entityBuilder: T, options?: RelationshipOptions) {
+  manyToMany<T>(
+    entityBuilder: T,
+    options?: RelationshipOptions & {
+      pivotTable?: string
+      pivotEntity?: () => DmlEntity<any>
+    }
```
adrien2p:
```suggestion
    options?: RelationshipOptions & ({
      pivotTable?: string
      pivotEntity?: never
    } | {
      pivotTable?: never
      pivotEntity?: () => DmlEntity<any>
    })
```

The idea being that we only allow one or the other wdyt? Unless it makes sense to have both but not sure

adrien2p (approved):
LGTM, few suggestions

packages/core/utils/src/dml/entity-builder.ts
thetutlage: What if I want to use a custom pivot table name with a custom pivot entity. Remember, we do not allow defining a custom table name via entities. So there is no way for me to configure that on the pivot entity

adrien2p: yes thats true, but don't we want to have the same behaviour has any other entity? meaning that the table name will be constructed from the entity name, here the pivot entity name would be the one defining this table name. If you look at the migration level it is also probably what will happen, a table will be created for the pivot entity which might in that case defer from configuration

adrien2p: Validated through a call discussion 💪

thetutlage: Makes sense!

PR #7781, opened 2024-06-20
chore: make module loaders DML aware and auto generate joiner config

adrien2p:
**What**
When a module is automatically loaded from its location, if the module service does not provides us with a joiner config then we will auto generate the joiner config based on the models that are present in the module no matter if they are mikro orm entities or dml entities.

This pr also allow a module to load with dml entities as models.

This pr also include refactoring for the `Modules` usage to cleanup inconsistencies and always import it from the `utils` package.

Also, now the DML export `model` and `toMikroOrmEntity` utilities

**Note**

This auto loading mechanism only works if we load the resources ourselves from a path, which will likely be all cases, if someone provides the module exports directly then they ll have to handle resources manually but i think it is fine

FIXES CORE-2353
FIXES CORE-2339

### DIFF
--- a/integration-tests/modules/__tests__/link-modules/product-variant-price-set.spec.ts
+++ b/integration-tests/modules/__tests__/link-modules/product-variant-price-set.spec.ts
@@ -1,7 +1,8 @@
-import { ModuleRegistrationName, Modules } from "@medusajs/modules-sdk"
+import { ModuleRegistrationName } from "@medusajs/modules-sdk"
 import { IPricingModuleService, IProductModuleService } from "@medusajs/types"
 import {
   ContainerRegistrationKeys,
+  Modules,
   remoteQueryObjectFromString,
 } from "@medusajs/utils"
 import { medusaIntegrationTestRunner } from "medusa-test-utils"
--- a/integration-tests/modules/__tests__/modules/remote-query.spec.ts
+++ b/integration-tests/modules/__tests__/modules/remote-query.spec.ts
@@ -1,6 +1,6 @@
-import { ModuleRegistrationName, Modules } from "@medusajs/modules-sdk"
+import { ModuleRegistrationName } from "@medusajs/modules-sdk"
 import { IRegionModuleService } from "@medusajs/types"
-import { ContainerRegistrationKeys } from "@medusajs/utils"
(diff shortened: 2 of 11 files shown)


packages/core/utils/src/dml/entity.ts
```diff
@@ -13,9 +13,25 @@ import {
...
+   * A static method to check if an entity is an instance of DmlEntity.
+   * It allows us to identify a specific object as being an instance of
+   * DmlEntity.
+   *
+   * @param entity
+   */
+  static isDmlEntity(entity: any): entity is DmlEntity<any> {
+    return (
+      entity instanceof DmlEntity ||
+      entity.__dmlEntitySymbol__ === this.__dmlEntitySymbol__
```
thetutlage: Will `__dmlEntitySymbol__` exist on the instance?

adrien2p: yes I fixed it in my commit ahah

adrien2p:
The review focus here can be on `modules-sdk` and `utils` package, I havent migrated the modules to rely on the joiner config autogeneration and they are not yet using the DML

riqwan (approved):
nice!

olivermrbl (approved):
Nice work and cleanup!

sradevski (approved):
Nice 👍

olivermrbl:
@adrien2p the starter will need to be updated to import `Modules` from utils as well. This is what is causing the CI to fail

adrien2p:
@olivermrbl the v2 branch of the starter already import from the utils

olivermrbl:
Ah sorry, myb ad!

olivermrbl:
@adrien2p, I believe it hasn't been updated actually: https://github.com/medusajs/medusa-starter-default/tree/feat/v2-ci

adrien2p:
> @adrien2p, I believe it hasn't been updated actually: https://github.com/medusajs/medusa-starter-default/tree/feat/v2-ci

@olivermrbl from what I ve seen it is importing from utils, it is not new that the modules enum is exported from utils :), but our packages were inconsistent

adrien2p:
@olivermrbl shall we? 😊

olivermrbl:
Let's do it!

PR #7786, opened 2024-06-20
fix(region): Soft deleting region unsets region on countries

olivermrbl:
**What**

Unset `region_id` on countries when soft deleting the associated region. Otherwise, countries cannot be added to new or other regions

Alternatively, we hard delete the region, which will trigger the DB level cascade

In both cases, we cannot easily restore the region countries.


fPolic (approved):

packages/modules/region/src/services/region-module.ts
```diff
@@ -129,6 +130,25 @@ export default class RegionModuleService
     return result
   }
 
+  @InjectManager("baseRepository_")
+  // @ts-ignore
+  async softDeleteRegions(
```
sradevski, line 135: Let's just add a comment that it is not an idempotent operation, so it cannot be reverted simply by restoring the region

sradevski (approved):

packages/modules/region/src/services/region-module.ts
```diff
@@ -129,6 +130,26 @@ export default class RegionModuleService
     return result
   }
 
+  @InjectManager("baseRepository_")
+  // @ts-ignore
```
adrien2p, line 134: nit: I prefer ts-expect-error now in those cases (not all)😂

adrien2p (approved):

PR #7792, opened 2024-06-21
chore: Internal medusa service proper typings with DML

adrien2p:
FIXES CORE-2372

### DIFF
--- a/packages/modules/inventory-next/src/services/inventory-level.ts
+++ b/packages/modules/inventory-next/src/services/inventory-level.ts
@@ -8,11 +8,9 @@ type InjectedDependencies = {
   inventoryLevelRepository: InventoryLevelRepository
 }
 
-export default class InventoryLevelService<
-  TEntity extends InventoryLevel = InventoryLevel
-> extends ModulesSdkUtils.MedusaInternalService<InjectedDependencies>(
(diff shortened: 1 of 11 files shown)


thetutlage:
Will DML types be imported from a subpath like?

```ts
import { Infer } from '@medusajs/types/dml'
```

sradevski (approved):
Nice 👍

adrien2p:
> Will DML types be imported from a subpath like?
> 
> ```ts
> import { Infer } from '@medusajs/types/dml'
> ```

We can do that later if we want (which I like and would prefer, but the convention is different 😄 ) but for now the usage is 
- `import { Infer } from "@medusaj/types`
- `import { DmlType } from "@medusaj/types`
  - `DmlTypes.Infer`

thetutlage:
> We can do that later if we want (which I like and would prefer, but the convention is different 😄 ) but for now the usage is

Okay, but then our naming is too generic for now. So maybe, I another PR we can make these names more specific to DML

thetutlage (approved):

adrien2p:
Yes @thetutlage I think about that everytime but you can still import from the bundle export alias so i think it is ok for now and it will only be used internally as of now right? At least for most cases. But at some point it would be noce to short to subpaths instead for all bundles export

PR #7795, opened 2024-06-21
feat: initialize nullable properties with null value

thetutlage:
FIXES CORE-2370

### DIFF
--- a/packages/core/utils/src/dml/__tests__/entity-builder.spec.ts
+++ b/packages/core/utils/src/dml/__tests__/entity-builder.spec.ts
@@ -236,10 +236,11 @@ describe("Entity builder", () => {
       expect(metaData.className).toEqual("User")
       expect(metaData.path).toEqual("User")
 
-      const u = new User()
-      expect(u.username).toEqual(null)
-      u.username = "john"
-      expect(u.username).toEqual("john")
+      const userInstance = new User()
+      expect(userInstance.username).toEqual(null)
+
+      userInstance.username = "john"
+      expect(userInstance.username).toEqual("john")
 
       expect(metaData.filters).toEqual({
         softDeletable: {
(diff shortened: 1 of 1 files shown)


packages/core/utils/src/dml/__tests__/entity-builder.spec.ts
```diff
@@ -236,6 +236,9 @@ describe("Entity builder", () => {
       expect(metaData.className).toEqual("User")
       expect(metaData.path).toEqual("User")
 
+      const u = new User()
+      expect(u.username).toEqual(null)
```
adrien2p:
```suggestion
      expect(new User().username).toEqual(null)
```

just to avoid having a one letter naming here 😄 wdyt?, same comment for the others

packages/core/utils/src/dml/helpers/create-mikro-orm-entity.ts
```diff
@@ -188,12 +208,12 @@ export function createMikrORMEntity() {
         ? PrimaryKey({
             columnType: "text",
             type: "string",
-            nullable: field.nullable,
```
adrien2p, line 191: nice catch

adrien2p (approved):

packages/core/utils/src/dml/__tests__/entity-builder.spec.ts
thetutlage: I have added another assertion that mutates the value. So we will need a variable now 😇

thetutlage: Renamed variable to `userInstance`

adrien2p: wonderful :)

adrien2p (approved):

PR #7801, opened 2024-06-21
feat(types,utils): DML can create a bigNumber property

riqwan:
what:

- adds bigNumber as a property to DML
- creates a bigNumber options field (`raw_{{ field }}`) as a part of the schema

RESOLVES CORE-2375


packages/core/utils/src/dml/__tests__/entity-builder.spec.ts
```diff
@@ -180,6 +207,26 @@ describe("Entity builder", () => {
...
+        spend_limit: {
+          columnType: "numeric",
+          default: 500.4,
+          getter: true,
+          name: "spend_limit",
+          nullable: false,
+          reference: "scalar",
+          setter: true,
+          trackChanges: false,
+          type: "any",
```
riqwan, line 219: @adrien2p what do you think the type should be here for this and raw_ field?

packages/core/utils/src/dml/helpers/create-mikro-orm-entity.ts
```diff
@@ -162,19 +165,58 @@ export function createMikrORMEntity() {
...
+        MikroORMEntity.prototype,
+        field.fieldName,
+        nullableOptions
+      )
+
+      /**
+       * If dataType is a bigNumber, the nullable clause gets passed
+       * onto the bigNumber's raw field as well
+       */
+      if (field.dataType.name === "bigNumber") {
```
riqwan: todo: This should be reverted as this is coming in directly from the schema

packages/core/utils/src/dml/__tests__/entity-builder.spec.ts
adrien2p: I think we can keep it as any cause I am not sure how to tell mikro orm it can be BigNumberInput. what they do it that they have a util which takes the type of the input as follow and compare it. so any seems the best for now
```ts
const objectType = Object.prototype.toString.call(value);
return objectType.match(/\[object (\w+)]/)[1].toLowerCase();
```

adrien2p (approved):
Nice one 💪

olivermrbl (approved):
Nice!

PR #7807, opened 2024-06-24
feat(utils,currency): Migrate currency to use DML

riqwan:
what:

- migrates currency module to use DML
- updates the id field to not generate ids upon a flag
- fixes the key for a custom id field
- joiner config fixes by @adrien2p 

RESOLVES CORE-2338

### DIFF
--- a/packages/core/utils/src/common/__tests__/define-config.spec.ts
+++ b/packages/core/utils/src/common/__tests__/define-config.spec.ts
@@ -48,6 +48,21 @@ describe("defineConfig", function () {
             "resolve": "@medusajs/fulfillment",
           },
           "inventoryService": true,
+          "notification": {
+            "options": {
+              "providers": [
+                {
+                  "options": {
+                    "config": {
+                      "local": {},
+                    },
+                  },
+                  "resolve": "@medusajs/notification-local",
+                },
+              ],
+            },
+            "resolve": "@medusajs/notification",
+          },
           "order": true,
           "payment": true,
           "pricingService": true,
@@ -138,6 +153,21 @@ describe("defineConfig", function () {
             "resolve": "./modules/github",
           },
           "inventoryService": true,
+          "notification": {
+            "options": {
+              "providers": [
+                {
+                  "options": {
+                    "config": {
+                      "local": {},
+                    },
+                  },
+                  "resolve": "@medusajs/notification-local",
+                },
+              ],
+            },
+            "resolve": "@medusajs/notification",
+          },
           "order": true,
           "payment": true,
           "pricingService": true,
@@ -225,6 +255,21 @@ describe("defineConfig", function () {
             "resolve": "@medusajs/fulfillment",
           },
           "inventoryService": true,
+          "notification": {
+            "options": {
+              "providers": [
+                {
+                  "options": {
+                    "config": {
+                      "local": {},
+                    },
+                  },
+                  "resolve": "@medusajs/notification-local",
+                },
+              ],
+            },
+            "resolve": "@medusajs/notification",
+          },
           "order": true,
           "payment": true,
           "pricingService": true,
@@ -314,6 +359,21 @@ describe("defineConfig", function () {
             "resolve": "@medusajs/fulfillment",
           },
           "inventoryService": true,
+          "notification": {
+            "options": {
+              "providers": [
+                {
+                  "options": {
+                    "config": {
+                      "local": {},
+                    },
+                  },
+                  "resolve": "@medusajs/notification-local",
+                },
+              ],
+            },
+            "resolve": "@medusajs/notification",
+          },
           "order": true,
           "payment": true,
           "pricingService": true,
--- a/packages/core/utils/src/common/define-config.ts
+++ b/packages/core/utils/src/common/define-config.ts
@@ -114,6 +114,21 @@ export function defineConfig(config: Partial<ConfigModule> = {}): ConfigModule {
         ],
       },
     },
+    [Modules.NOTIFICATION]: {
+      resolve: "@medusajs/notification",
+      options: {
+        providers: [
+          {
+            resolve: "@medusajs/notification-local",
+            options: {
+              config: {
+                local: {},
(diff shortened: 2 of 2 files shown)


integration-tests/modules/__tests__/currency/admin/currency.spec.ts
```diff
@@ -60,9 +60,10 @@ medusaIntegrationTestRunner({
           adminHeaders
         )
 
-        expect(listResp.data.currencies).toHaveLength(1)
-        expect(listResp.data.currencies[0]).toEqual(
-          expect.objectContaining({ code: "zwl", name: "Zimbabwean Dollar" })
+        expect(listResp.data.currencies).toEqual(
```
riqwan, line 63: @adrien2p do you know what change in this PR is causing the module loaders to run on each test?

I had to add this fix to the test because earlier the module loader would be run once in a test and then all the data created in that loader will be deleted on any subsequent test. 

But now, its running on every test.

packages/modules/currency/src/models/currency.ts
```diff
@@ -1,50 +1,10 @@
...
-    defaultRaw: "now()",
-  })
-  updated_at: Date
-}
-
-export default Currency
+import { model } from "@medusajs/utils"
+
+export default model.define("currency", {
+  code: model.id({ generateId: false }),
```
carlos-r-l-rodrigues: is this the same as `.text().primaryKey()`?

carlos-r-l-rodrigues (approved):

packages/modules/currency/src/models/currency.ts
olivermrbl: Related to this comment, I don't think we should add `generateId` to the `id` property. It seems like we are solving a problem at the wrong place.

The `id` is a specific-purpose property that covers the vast majority of model definition use cases, and we shouldn't compromise its simplicity for edge cases like this.

Instead, we should encourage the approach @carlos-r-l-rodrigues highlights here, and if this is not currently possible, I think we should make it possible instead of adding this optionality to the id column.

Wdyt?

shahednasser: I came here to same the same thing as @carlos-r-l-rodrigues 😄 i think using `id` is very confusing. `primaryKey` or just `primary` should be good

packages/modules/currency/mikro-orm.config.dev.ts
```diff
@@ -1,8 +1,7 @@
...
 import * as entities from "./src/models"
 
-module.exports = {
-  entities: Object.values(entities),
-  schema: "public",
-  clientUrl: "postgres://postgres@localhost/medusa-currency",
-  type: "postgresql",
-}
+defineMikroOrmCliConfig({
+  entities: Object.values(entities) as any[],
```
sradevski: Why is the cast needed here?

packages/core/utils/src/modules-sdk/loaders/load-models.ts
```diff
@@ -29,7 +29,7 @@ export function loadModels(basePath: string) {
         try {
           const required = require(filePath)
           return Object.values(required).filter(
-            (resource) => typeof resource === "function" && !!resource.name
+            (resource: any) => !!resource.name
```
sradevski: Same here, can we maybe type it closer to what it represents?

packages/core/utils/src/modules-sdk/joiner-config-builder.ts
```diff
@@ -79,16 +79,22 @@ export function defineJoinerConfig(
...
-          `${pluralize(camelToSnakeCase(entity.name).toLowerCase())}`,
-        ],
-        args: {
-          entity: entity.name,
-          methodSuffix: pluralize(upperCaseFirst(entity.name)),
-        },
-      })),
+      ...models
+        .filter((model) => {
+          return !alias
```
sradevski: It's a bit strange to have a tertiary check in a filter, and it's difficult to understand what it does. Maybe we can rewrite it to
```
!alias || !alias.some(...)
```

packages/modules/currency/src/models/currency.ts
riqwan: updated 🤞🏻

shahednasser (approved):

integration-tests/modules/__tests__/currency/admin/currency.spec.ts
adrien2p: weird, I didnt changed that part

packages/core/utils/src/modules-sdk/loaders/load-models.ts
adrien2p: We choose any because a user can decide to export anything he wants and we can't know, so any or unknown seamed the most appropriate. I ve changed it still

packages/core/utils/src/dml/__tests__/entity-builder.spec.ts
```diff
@@ -1166,6 +1166,109 @@ describe("Entity builder", () => {
     })
   })
 
+  describe("Entity builder | primaryKey", () => {
+    test("should create both id fields and primaryKey fields", () => {
+      const model = new EntityBuilder()
+      const user = model.define("user", {
+        id: model.id({ primaryKey: false }),
```
olivermrbl, line 1173: The main reason for my other suggestion was to eliminate this option entirely, so that it doesn't pollute the `id` implementation. Fine to merge now, but let's see if we can get rid of it later on with an "inference" mechanism

riqwan: Did a dirty check for the inference, looks like we can get rid of this easily in the next PR.

olivermrbl (approved):
Noice!

packages/core/utils/src/dml/__tests__/entity-builder.spec.ts
olivermrbl: I like it! 💪

PR #7808, opened 2024-06-24
Add support for tax inclusivity to region and store

sradevski:
This also includes rework of the currency model for the Store module.

This change is breaking as existing stores won't have any supported currencies set, so users would need to go to the store settings again and choose the supported currencies there.


packages/modules/store/src/models/currency.ts
```diff
@@ -0,0 +1,84 @@
...
+@Filter(DALUtils.mikroOrmSoftDeletableFilterOptions)
+export default class StoreCurrency {
+  @PrimaryKey({ columnType: "text" })
+  id: string
+
+  @Searchable()
+  @Property({ columnType: "text" })
+  currency_code: string
+
+  @Property({ columnType: "boolean", default: true })
```
olivermrbl: thought: I would have expected this to default to false. Tax-inclusive pricing is commonly a feature you opt-in for.  What are your thoughts here?

packages/modules/region/src/models/region.ts
```diff
@@ -35,6 +35,9 @@ export default class Region {
   @Property({ columnType: "boolean" })
   automatic_taxes: boolean = true
 
+  @Property({ columnType: "boolean" })
+  is_tax_inclusive: boolean = true
```
olivermrbl: Same as other comment.

olivermrbl (commented):
Nice work, BE looks good. Will give the admin a spin now

packages/modules/store/src/models/currency.ts
sradevski: @olivermrbl I am fine either way. I just checked, 102 countries use tax-inclusive pricing, notably most of europe, china, japan, and india. I think the biggest market where it's tax-exclusive is the US, so it depends which markets we want to optimize for. Let me know what you think.

sradevski: Source: https://help.shopify.com/en/manual/international/pricing/dynamic-tax-inclusive-pricing

olivermrbl: I don't have any strong opinion on this, and I think your argument on target market is reasonable. My thought was rooted in the fact that most other commerce platforms come with this setting disabled by default, so for users migrating to Medusa, this could be unexpected behavior. That said, this is a very minor thing and can be solved with proper education. 

Will let @srindom decide on this one.

olivermrbl (approved):
Admin works like a charm too!

packages/modules/store/src/models/currency.ts
sradevski: I removed the is_tax_inclusive flag as it will be moved to the pricing module, resolving the comment for now (we can re-discuss once we have it implemented again)

PR #7810, opened 2024-06-24
fix: improve error message when config fails to load in build

srindom:
**What**

Throws the original error when failing to load medusa-config during build.

Before it only threw "Failed to load medusa-config". Now it does:
CleanShot 2024-06-24 at 11  17 46@2x

### DIFF
--- a/packages/medusa/src/commands/build.ts
+++ b/packages/medusa/src/commands/build.ts
@@ -36,7 +36,8 @@ async function findFiles(dir: string): Promise<string[]> {
     )
     return paths.flat()
   } catch (e) {
-    console.log(`Failed to read directory ${dir}`)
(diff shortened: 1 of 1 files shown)


packages/medusa/src/commands/build.ts
```diff
@@ -122,6 +122,7 @@ export default async function ({ directory }: BuildArgs) {
 
   if (error) {
     console.log(`Failed to load medusa-config.js`)
```
sradevski: This shouldn't be a `.log`, it should be an `.error` as well

sradevski (approved):

packages/medusa/src/commands/build.ts
adrien2p: shouldn't we import and use the logger instead?

PR #7811, opened 2024-06-24
feat(dashboard): handle region pricing on create/edit

fPolic:
**What**
- allow for region prices creation on product create flow
- editing of region prices in prices edit form


olivermrbl (approved):
LGTM!

PR #7813, opened 2024-06-24
docs: fixed response tag for reset password API ref

ank-2222:
image

- Error tag should be replaced with success tag in Store API reference - Reset password Response
- Changed the condition to check if code starts with 20 or not.


ank-2222 (commented):
checking if code starts with 20 or not

shahednasser (commented):
Thanks for the fix! But we also need it on the `v1.x` branch. Can you open another PR for that branch?

ank-2222:
@shahednasser [New PR](https://github.com/medusajs/medusa/pull/7815) on v1.x branch

PR #7821, opened 2024-06-24
fix(promotion): scope campaign identifier index with deleted at

riqwan:
what:

- scopes campaign indentifer unique index with deleted at

RESOLVES CORE-2378


olivermrbl (approved):
LGTM

PR #7822, opened 2024-06-24
fix(utils): Uppercase event action name

olivermrbl:
Currently, the keys of the events contain a lowercased action, e.g. `INVENTORY_ITEM_created`, which is breaking our usage across multiple modules

### DIFF
--- a/packages/core/utils/src/modules-sdk/event-builder-factory.ts
+++ b/packages/core/utils/src/modules-sdk/event-builder-factory.ts
@@ -46,13 +46,17 @@ export function eventBuilderFactory({
     const aggregator = sharedContext.messageAggregator!
     const messages: EventBusTypes.RawMessageFormat[] = []
 
+    // The event enums contains event formatted like so [object]_[action] e.g. PRODUCT_CREATED
+    // We expect the keys of events to be fully uppercased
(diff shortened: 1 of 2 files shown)


adrien2p (approved):
Nice catch ✌️

packages/core/utils/src/event-bus/__tests__/build-event-names-from-entity-name.spec.ts
```diff
@@ -0,0 +1,34 @@
+import { buildEventNamesFromEntityName } from "../../../dist"
```
riqwan: this is importing from dist, not sure if this could cause problems somewhere

riqwan (approved):
🥇

packages/core/utils/src/event-bus/__tests__/build-event-names-from-entity-name.spec.ts
adrien2p: It should not be dist, i missed it 👍

olivermrbl: nice catch fixed it

PR #7823, opened 2024-06-24
chore(medusa): return endpoints

carlos-r-l-rodrigues:


packages/core/core-flows/src/order/workflows/create-return.ts
```diff
@@ -139,6 +141,37 @@ function validateCustomRefundAmount({
   }
 }
 
+function prepareReceiveItems({
+  receiveNow,
+  returnId,
+  items,
+  createdBy,
+}: {
+  receiveNow: boolean
```
sradevski, line 150: What is the meaning of `receiveNow`?

packages/core/core-flows/src/order/utils/order-validation.ts
```diff
@@ -30,3 +30,37 @@ export function throwIfItemsDoesNotExistsInOrder({
...
+}) {
+  if (orderReturn.canceled_at) {
+    throw new MedusaError(
+      MedusaError.Types.INVALID_DATA,
+      `return with id ${orderReturn.id} has been canceled.`
+    )
+  }
+}
+
+export function throwIfItemsDoesNotExistsInReturn({
```
sradevski, line 47:
```suggestion
export function validateItemExists({
```

Should suffice, you are passing an `orderReturn` input anyway, and throwing is a common way to handle invalid data anyway. Alternatively `validateItemExistsInReturn` is also fine, but I think there is no need to have the `throwIf` part

sradevski (approved):
LGTM, just some minor comments

packages/core/core-flows/src/order/workflows/create-return.ts
carlos-r-l-rodrigues: If true, the return is created and marked as received in the same flow.

sradevski: Ah got it, maybe we can call it `markAsReceived` instead?

sradevski: And in that case, why don't we set the `created_by` and `items`? Shouldn't we have that information regardless, and manage whether something is received or not by a status?

carlos-r-l-rodrigues: `receive_now` is the property used in v1 API. I just kept the name.
I didn't understand the `created_by` and `items` question. This would be a conditional step, as we don't have it yet, we send an empty `items` so the step is skipped

packages/core/core-flows/src/order/workflows/receive-return.ts
```diff
@@ -0,0 +1,53 @@
+import { OrderWorkflow } from "@medusajs/types"
+import {
+  WorkflowData,
+  createStep,
+  createWorkflow,
+} from "@medusajs/workflows-sdk"
+import { useRemoteQueryStep } from "../../common"
+
+import { ReturnDTO } from "@medusajs/types/src"
```
adrien2p: todo: can you check that we dont have other imports from src directly please 😄

adrien2p (approved):
LGTM, one little todo 💪

packages/core/core-flows/src/order/workflows/create-return.ts
sradevski: ah ok understood, I thought they are stored in the DB 👍

packages/core/core-flows/src/order/workflows/receive-return.ts
carlos-r-l-rodrigues: nice catch...

carlos-r-l-rodrigues: I replaced from the entire project. there were many.

adrien2p: nice catch 💪 I was almost sure some must have gone through, maybe worth checking dist as well 😓

packages/core/core-flows/src/api-key/steps/link-sales-channels-to-publishable-key.ts
```diff
@@ -1,6 +1,6 @@
-import { LinkWorkflowInput } from "@medusajs/types/src"
+import { LinkWorkflowInput } from "@medusajs/types"
```
adrien2p, line 1: suggestion: Lin -> Link 😃

carlos-r-l-rodrigues: `Lin`?

adrien2p: my bad nvm I read it wrong 😂

adrien2p:
Nice cleanup like I Like to see 🚀 😅

PR #7824, opened 2024-06-25
docs(v1.x): generate references

shahednasser:
Generate references after change in PR #7812


olivermrbl:
CleanShot 2024-06-25 at 21 10 45 Is this correct? 👀 😅

shahednasser:
@olivermrbl yea it is 😅 Aside from the OAS, we also need to re-generate the references for the JS Client and Medusa React, but I re-generated everything (which we usually do on release) just to be on the safe side of making sure the change is reflected everywhere necessary. I can instead just generate for JS Client and Medusa React if that's better?

PR #7827, opened 2024-06-25
feat: Add necessary middlewares for tax inclusive pricing

sradevski:
We are adding tax inclusive pricing calculation when listing products.

Two things to keep in mind:
- `region_id` will be required if you request calculated prices.
- We won't accept `currency_code` anymore, as that will come from the region info (since ultimately a cart and its currency are tied to a region)

REF CORE-2376
DEPENDS ON #8003

### DIFF
--- a/integration-tests/http/__tests__/product/store/product.spec.ts
+++ b/integration-tests/http/__tests__/product/store/product.spec.ts
@@ -1269,22 +1269,6 @@ medusaIntegrationTestRunner({
       let euCart
 
       beforeEach(async () => {
-        product1 = (
-          await api.post(
-            "/admin/products",
-            getProductFixture({ title: "test1", status: "published" }),
-            adminHeaders
-          )
-        ).data.product
-
-        product2 = (
-          await api.post(
-            "/admin/products",
-            getProductFixture({ title: "test2", status: "published" }),
-            adminHeaders
-          )
-        ).data.product
-
         usRegion = (
           await api.post(
             "/admin/regions",
@@ -1326,6 +1310,39 @@ medusaIntegrationTestRunner({
             adminHeaders
           )
         ).data.region
+
+        product1 = (
+          await api.post(
+            "/admin/products",
+            getProductFixture({
+              title: "test1",
+              status: "published",
+              variants: [
+                {
+                  title: "Test taxes",
+                  prices: [
+                    {
+                      amount: 45,
+                      currency_code: "eur",
+                      rules: { region_id: euRegion.id },
+                    },
+                    {
+                      amount: 100,
+                      currency_code: "usd",
+                      rules: { region_id: usRegion.id },
+                    },
+                    {
+                      amount: 30,
+                      currency_code: "dkk",
+                      rules: { region_id: dkRegion.id },
+                    },
+                  ],
+                },
+              ],
+            }),
+            adminHeaders
+          )
+        ).data.product
 
         euCart = (await api.post("/store/carts", { region_id: euRegion.id }))
           .data.cart
@@ -1377,7 +1394,7 @@ medusaIntegrationTestRunner({
           )
         ).data.products
 
-        expect(products.length).toBe(2)
+        expect(products.length).toBe(1)
         expect(products[0].variants[0].calculated_price).not.toHaveProperty(
           "calculated_amount_with_tax"
         )
@@ -1393,7 +1410,7 @@ medusaIntegrationTestRunner({
           )
         ).data.products
 
-        expect(products.length).toBe(2)
+        expect(products.length).toBe(1)
         expect(products[0].variants[0].calculated_price).not.toHaveProperty(
           "calculated_amount_with_tax"
         )
@@ -1409,7 +1426,7 @@ medusaIntegrationTestRunner({
           )
         ).data.products
 
-        expect(products.length).toBe(2)
+        expect(products.length).toBe(1)
         expect(products[0].variants).toEqual(
           expect.arrayContaining([
             expect.objectContaining({
@@ -1430,14 +1447,14 @@ medusaIntegrationTestRunner({
         ).toEqual("40.9")
       })
 
-      it.skip("should return prices with and without tax for a tax exclusive region when listing products", async () => {
+      it("should return prices with and without tax for a tax exclusive region when listing products", async () => {
         const products = (
           await api.get(
             `/store/products?fields=id,*variants.calculated_price&region_id=${dkRegion.id}&country_code=dk`
           )
         ).data.products
 
-        expect(products.length).toBe(2)
+        expect(products.length).toBe(1)
         expect(products[0].variants).toEqual(
           expect.arrayContaining([
             expect.objectContaining({
@@ -1459,7 +1476,7 @@ medusaIntegrationTestRunner({
           )
         ).data.products
 
-        expect(products.length).toBe(2)
+        expect(products.length).toBe(1)
         expect(products[0].variants).toEqual(
           expect.arrayContaining([
             expect.objectContaining({
@@ -1493,7 +1510,7 @@ medusaIntegrationTestRunner({
           )
         ).data.products
 
-        expect(products.length).toBe(2)
+        expect(products.length).toBe(1)
         expect(products[0].variants).toEqual(
           expect.arrayContaining([
             expect.objectContaining({
--- a/packages/medusa/src/api/store/products/helpers.ts
+++ b/packages/medusa/src/api/store/products/helpers.ts
@@ -97,11 +97,8 @@ export const wrapProductsWithTaxPrices = async <T>(
           variant.calculated_price!.is_calculated_price_tax_inclusive!,
       })
 
-      // TODO: Resolve the typings when tax inclusivity is finalized.
-      ;(variant.calculated_price as any).calculated_amount_with_tax =
-        priceWithTax
-      ;(variant.calculated_price as any).calculated_amount_without_tax =
-        priceWithoutTax
+      variant.calculated_price.calculated_amount_with_tax = priceWithTax
+      variant.calculated_price.calculated_amount_without_tax = priceWithoutTax
     })
   })
 }
--- a/integration-tests/helpers/fixtures.ts
+++ b/integration-tests/helpers/fixtures.ts
@@ -1,7 +1,7 @@
 import { HttpTypes } from "@medusajs/types"
 
 export const getProductFixture = (
-  overrides: Partial<HttpTypes.AdminProduct>
+  overrides: Partial<HttpTypes.AdminCreateProduct>
 ) => ({
   title: "Test fixture",
   description: "test-product-description",
--- a/packages/core/types/src/http/pricing/common.ts
+++ b/packages/core/types/src/http/pricing/common.ts
@@ -21,6 +21,16 @@ export interface BaseCalculatedPriceSet {
   calculated_amount: number | null
 
   /**
+   * The amount of the calculated price with taxes included. If the calculated price is tax inclusive, this field will be the same as `calculated_amount`.
+   */
+  calculated_amount_with_tax?: number | null
+
+  /**
+   * The amount of the calculated price without taxes included. If the calculated price is tax exclusive, this field will be the same as `calculated_amount`.
+   */
+  calculated_amount_without_tax?: number | null
+
+  /**
    * Whether the original price is associated with a price list. During the calculation process, if the price list of the calculated price is of type override,
    * the original price will be the same as the calculated price. In that case, the value of this property is `true`.
    */
--- a/packages/core/types/src/http/product/admin/payloads.ts
+++ b/packages/core/types/src/http/product/admin/payloads.ts
@@ -22,6 +22,8 @@ export interface AdminCreateProductVariantPrice {
   amount: number
   min_quantity?: number | null
   max_quantity?: number | null
+  // Note: Although the BE is generic, we only use region_id for price rules for now, so it's better to keep the typings stricter.
+  rules?: { region_id: string } | null
 }
 
 export interface AdminCreateProductVariant {
--- a/packages/modules/pricing/integration-tests/__tests__/services/pricing-module/calculate-price.spec.ts
+++ b/packages/modules/pricing/integration-tests/__tests__/services/pricing-module/calculate-price.spec.ts
@@ -1449,6 +1449,39 @@ moduleIntegrationTestRunner<IPricingModuleService>({
             ])
           })
 
+          it("should return the region tax inclusivity for the selected price when there are multiple region preferences", async () => {
+            await (service as any).createPricePreferences([
+              {
+                attribute: "region_id",
+                value: "DE",
+                is_tax_inclusive: false,
+              },
+              {
+                attribute: "region_id",
+                value: "PL",
+                is_tax_inclusive: true,
+              },
+            ])
+
+            const priceSetsResult = await service.calculatePrices(
+              { id: ["price-set-PLN"] },
+              {
+                context: { currency_code: "PLN", region_id: "PL" },
+              }
+            )
+
+            expect(priceSetsResult).toEqual([
+              expect.objectContaining({
+                id: "price-set-PLN",
+                is_calculated_price_tax_inclusive: true,
+                calculated_amount: 300,
+                is_original_price_tax_inclusive: true,
+                original_amount: 300,
+                currency_code: "PLN",
+              }),
+            ])
+          })
+
           it("should return the appropriate tax inclusive setting for each calculated and original price", async () => {
             await createPriceLists(service, {}, {})
             await (service as any).createPricePreferences([
--- a/packages/modules/pricing/src/services/pricing-module.ts
+++ b/packages/modules/pricing/src/services/pricing-module.ts
@@ -334,7 +334,9 @@ export default class PricingModuleService
             is_calculated_price_price_list: !!calculatedPrice?.price_list_id,
             is_calculated_price_tax_inclusive: isTaxInclusive(
               priceRulesPriceMap.get(calculatedPrice.id),
-              pricingPreferences
+              pricingPreferences,
+              calculatedPrice.currency_code!,
+              pricingContext.context?.region_id as string
             ),
             calculated_amount: parseInt(calculatedPrice?.amount || "") || null,
 
@@ -342,7 +344,9 @@ export default class PricingModuleService
             is_original_price_tax_inclusive: originalPrice?.id
               ? isTaxInclusive(
                   priceRulesPriceMap.get(originalPrice.id),
-                  pricingPreferences
+                  pricingPreferences,
+                  originalPrice.currency_code || calculatedPrice.currency_code!,
+                  pricingContext.context?.region_id as string
                 )
               : false,
             original_amount: parseInt(originalPrice?.amount || "") || null,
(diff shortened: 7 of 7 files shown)


packages/medusa/src/api/store/products/helpers.ts
```diff
@@ -30,3 +47,101 @@ export const maybeApplyStockLocationId = async (req: MedusaRequest, ctx) => {
...
+}
+
+const getTaxPricesForVariant = (
+  variant: HttpTypes.StoreProductVariant,
+  taxRatesForVariant: ItemTaxLineDTO[],
+  isTaxInclusive: boolean
+) => {
+  const tax = calculateTaxTotal({
+    // TODO: See why rate is nullable in tax
+    taxLines: taxRatesForVariant as any,
```
olivermrbl: Can we avoid the `any` here?

packages/medusa/src/api/store/products/helpers.ts
```diff
@@ -30,3 +47,101 @@ export const maybeApplyStockLocationId = async (req: MedusaRequest, ctx) => {
...
+      const { priceWithTax, priceWithoutTax } = getTaxPricesForVariant(
+        variant,
+        taxRatesForVariant,
+        isTaxInclusive
+      )
+
+      ;(variant.calculated_price as any).calculated_amount_with_tax =
+        priceWithTax
+      ;(variant.calculated_price as any).calculated_amount_without_tax =
+        priceWithoutTax
```
olivermrbl, line 97:
```suggestion
      ;variant.calculated_price["calculated_amount_with_tax"] =
        priceWithTax
      ;variant.calculated_price["calculated_amount_without_tax"] =
        priceWithoutTax
```

packages/medusa/src/api/store/products/helpers.ts
```diff
@@ -30,3 +47,101 @@ export const maybeApplyStockLocationId = async (req: MedusaRequest, ctx) => {
...
+) => {
+  const tax = calculateTaxTotal({
+    // TODO: See why rate is nullable in tax
+    taxLines: taxRatesForVariant as any,
+    taxableAmount: variant.calculated_price!.calculated_amount!,
+    includesTax: isTaxInclusive,
+  })
+
+  const priceWithoutTax = isTaxInclusive
+    ? tax.minus(variant.calculated_price!.calculated_amount!).abs().toNumber()
```
olivermrbl: Wouldn't it be more human readable to do `price - tax` instead of `absolute(tax - price)`?

olivermrbl (dismissed):
Overall LGTM, there are a few todos here and there, but I assume these will be tackled in follow-up PRs?

integration-tests/http/__tests__/product/store/product.spec.ts
```diff
@@ -894,7 +894,7 @@ medusaIntegrationTestRunner({
         expect(error.response.status).toEqual(400)
         expect(error.response.data).toEqual({
           message:
-            "Missing required pricing context to calculate prices - currency_code or region_id",
+            "Missing required pricing context to calculate prices - region_id",
```
riqwan, line 895: I'd imagine North America as a region having multiple currencies - CAD, USD. We'd lose out on specifying these if we remove currency_code right? Same case with EU as a region with multiple currencies.

From the looks of this PR, are we not merging the concepts of tax region and region here?

packages/medusa/src/api/store/products/helpers.ts
```diff
@@ -30,3 +47,101 @@ export const maybeApplyStockLocationId = async (req: MedusaRequest, ctx) => {
...
+    return
+  }
+
+  const { isTaxInclusive } = req.taxContext.taxInclusivityContext
+  const taxService = req.scope.resolve(ModuleRegistrationName.TAX)
+
+  const taxRates = (await taxService.getTaxLines(
+    products.map(asTaxItem).flat(),
+    req.taxContext.taxLineContext
+  )) as unknown as ItemTaxLineDTO[]
```
riqwan, line 75: q: Isn't it our convention that we don't resolve modules outside of workflows?

packages/medusa/src/api/store/products/helpers.ts
```diff
@@ -30,3 +47,101 @@ export const maybeApplyStockLocationId = async (req: MedusaRequest, ctx) => {
...
+
+  const taxRatesAsMap = new Map<string, ItemTaxLineDTO[]>()
+  taxRates.forEach((taxRate) => {
+    if (!taxRatesAsMap.has(taxRate.line_item_id)) {
+      taxRatesAsMap.set(taxRate.line_item_id, [])
+    }
+
+    taxRatesAsMap.get(taxRate.line_item_id)?.push(taxRate)
+  })
```
riqwan: Can just be called taxRatesMap

```suggestion
  const taxRatesMap = new Map<string, ItemTaxLineDTO[]>()
  
  taxRates.forEach((taxRate) => {
    if (!taxRatesMap.has(taxRate.line_item_id)) {
      taxRatesMap.set(taxRate.line_item_id, [])
    }

    taxRatesMap.get(taxRate.line_item_id)?.push(taxRate)
  })
```

packages/medusa/src/api/store/products/helpers.ts
```diff
@@ -30,3 +47,101 @@ export const maybeApplyStockLocationId = async (req: MedusaRequest, ctx) => {
...
+  products.forEach((product) => {
+    product.variants?.forEach((variant) => {
+      const taxRatesForVariant = taxRatesAsMap.get(variant.id) || []
+      const { priceWithTax, priceWithoutTax } = getTaxPricesForVariant(
+        variant,
+        taxRatesForVariant,
+        isTaxInclusive
+      )
+
+      ;(variant.calculated_price as any).calculated_amount_with_tax =
```
riqwan: sugg: `calculated_price` is not guaranteed here, we might want to return early, earlier in the function if thats the case.

packages/medusa/src/api/utils/middlewares/products/set-tax-context.ts
```diff
@@ -0,0 +1,75 @@
...
+  if (!region) {
+    throw new MedusaError(
+      MedusaError.Types.INVALID_DATA,
+      `Region with id ${req.filterableFields.region_id} not found when populating the tax context`
+    )
+  }
+
+  // TODO: Add is_tax_inclusive here from pricing module
+  return {
+    isTaxInclusive: true,
```
olivermrbl: question: Just for clarity, how would this flag get populated in a future iteration?

packages/medusa/src/api/store/products/helpers.ts
```diff
@@ -30,3 +47,101 @@ export const maybeApplyStockLocationId = async (req: MedusaRequest, ctx) => {
...
+      if (!variant.calculated_price) {
+        return
+      }
+
+      return {
+        id: variant.id,
+        product_id: product.id,
+        product_name: product.title,
+        product_categories: product.categories?.map((c) => c.name),
+        product_category_id: product.categories?.[0]?.id,
```
riqwan: q: this looks weird here, especially the decision to choose the first category. Shouldn't it be `product_category_ids: [...]?`

packages/medusa/src/api/utils/middlewares/products/normalize-data-for-context.ts
```diff
@@ -0,0 +1,74 @@
...
+import { refetchEntities, refetchEntity } from "../../refetch-entity"
+
+export function normalizeDataForContext() {
+  return async (req: AuthenticatedMedusaRequest, _, next: NextFunction) => {
+    // If the product pricing is not requested, we don't need region information
+    let withCalculatedPrice = req.remoteQueryConfig.fields.some((field) =>
+      field.startsWith("variants.calculated_price")
+    )
+
+    // If the region is passed, we calculate the prices without requesting them.
```
olivermrbl, line 13: Why do we do that?

packages/medusa/src/api/utils/middlewares/common/clear-context-fields.ts
```diff
@@ -0,0 +1,12 @@
+import { NextFunction } from "express"
+import { MedusaRequest } from "../../../../types/routing"
+
+export function clearContextFields(fields: string[]) {
```
riqwan: suggestion:

```suggestion
export function clearFiltersByKey(keys: string[]) {
```

packages/medusa/src/api/store/products/helpers.ts
```diff
@@ -30,3 +47,101 @@ export const maybeApplyStockLocationId = async (req: MedusaRequest, ctx) => {
...
+    includesTax: isTaxInclusive,
+  })
+
+  const priceWithoutTax = isTaxInclusive
+    ? tax.minus(variant.calculated_price!.calculated_amount!).abs().toNumber()
+    : variant.calculated_price!.calculated_amount
+
+  const priceWithTax = isTaxInclusive
+    ? variant.calculated_price!.calculated_amount
+    : tax.plus(variant.calculated_price!.calculated_amount!).abs().toNumber()
```
riqwan: sugg: This should come from a generic helper, maybe even the `calculateTaxTotal` to return an object with these parts.

We'll be doing this in multiple places, might be good to consolidate them

packages/medusa/src/api/utils/middlewares/products/normalize-data-for-context.ts
riqwan: Context - [CORE-2250](https://linear.app/medusajs/issue/CORE-2250)

packages/medusa/src/api/utils/middlewares/products/normalize-data-for-context.ts
```diff
@@ -0,0 +1,74 @@
...
+      }
+    }
+
+    // Finally, try to get it from the store defaults if not available
+    if (!regionId) {
+      const stores = await refetchEntities("store", {}, req.scope, [
+        "default_region_id",
+      ])
+      regionId = stores[0]?.default_region_id
+    }
```
riqwan, line 55: sugg: If region_id is not found from the cart or the query param, we should either throw or return without prices. If both the cart or query fail, we're already in a weird place. Picking from the store might put the query/customer in a weirder place if its a multi region store 😄 

I prefer the throw because I'd rather know an error is popping up as a dev so that I can fix it than silently returning.

riqwan (commented):
Will write some general observations on the slack thread thats confusing me in this flow tomorrow. 🤞🏻

packages/medusa/src/api/store/products/helpers.ts
```diff
@@ -30,3 +47,101 @@ export const maybeApplyStockLocationId = async (req: MedusaRequest, ctx) => {
 
   return entities.map((entity) => entity.stock_location_id)
 }
+
+export const wrapProductsWithTaxPrices = async <T>(
```
olivermrbl, line 53: We discussed the middleware approach briefly with @carlos-r-l-rodrigues yesterday evening, and he had a few good thoughts on this. Sharing here this morning, so we can kick off the discussion before his AM:

- Performance for store product list and retrieve endpoints is critical, so we should be careful about introducing too many operations at the HTTP layer that would harm it. Tax calculation has the potential to be in this category of operations. We want to use the search module as much as possible for the performance-critical endpoints.
- Performing updates on prices at the HTTP layer will make the sorting capabilities in the search module (when introduced) less useful. You can't rely on the sorting from the search engine because you are modifying your prices at a higher level in the stack.

I think both are valid points and good to open a discussion around. 

Right now (without the search module), there are at least two paths to go down:
- Leave the endpoint implementation as is, knowing we will have to rethink (and refactor) it when introducing the search module
- Put product variant price calculations into a separate endpoint, e.g. `GET /store/products/prices?id[]=...` (suggested by @carlos-r-l-rodrigues). This would mean a product page and product grid would to perform two requests to get the commonly required information to build such pages:
```
GET /store/products
GET /store/products/prices?id[]=...
```

Maybe there are other approaches to consider. Feel free to pitch in.

I don't have any strong opinion on this, there are pros and cons to both approaches. However, I do see the benefit of having a separate endpoint for price calculations in case we want to introduce more dynamic stuff in that context. Another case to consider in the future is inventory which will need to be computed and decorated dynamically on every request as well. We should try to follow a similar approach depending on what we decide here.

-----------------------

**On the search module**

We can index anything we want in the search module, but in the first iteration, it relies on the entity module events and doesn't offer much flexibility (correct me if I am wrong, @adrien2p, @carlos-r-l-rodrigues). Further down the line, we could imagine that the engine would expose "pre-indexing hooks," allowing us to perform calculations on resources before they are stored in the engine DB. 

An example use case of such pre-indexing hook would be price calculations. We could calculate product variant prices for each region and currency, so queries like the one we are discussing would require only a single request to the search engine.

integration-tests/http/__tests__/product/store/product.spec.ts
sradevski: @riqwan currently a region is tied to a single currency, so you would have to create two regions if you want to support checkout in multiple currencies. I already discussed this with Seb, and I agree with you, it is a limitation that we have in the system that we might want to address, but until then it doesn't make sense to accept currency code.

packages/medusa/src/api/store/products/helpers.ts
sradevski: we are calling a non-standard endpoint here, so we don't have another way of doing it.

packages/medusa/src/api/store/products/helpers.ts
sradevski: Yes I agree, and I think the tax module needs a rework, but it is outside the scope of this PR

packages/medusa/src/api/utils/middlewares/products/normalize-data-for-context.ts
sradevski: If we do that, then the store shouldn't have a default region at all, do we agree on that? Let's see what others think about this.

packages/medusa/src/api/store/products/helpers.ts
sradevski: @olivermrbl I agree on having the products endpoint very performant. In this case, unless you ask for the calculation by passing a cart or country context (we can even be more explicit), no calculations will happen, so it won't have an impact unless you need that information. 

Adding a separate endpoint won't solve anything, as it will have the exact same problems as doing `fields=*variants.calculated_price`. Doing that will actually introduce more issues
1. It will make the interface more complicated. 
2. The FE will need to do a join between products and prices
3. Sorting by price is a very common action, and that won't work on the `store/products` endpoint anymore.

If we want this tax calculation, we either do it like this, or we need to move that calculation to the pricing module. However, that won't reduce the number of operations, as you still need to list the tax lines and pass them to the pricing module.

Basically, our current architecture doesn't allow for significant performance improvements at the core. The main performance gains will have to come from a good caching layer.

packages/medusa/src/api/utils/middlewares/products/set-tax-context.ts
sradevski: @olivermrbl as I posted in the tax inclusive channel, we will store this as part of the pricing module (to begin with, set with either a currency or region rule), and we will just fetch it here before returning. The rest of the logic should remain unchanged (this will happen in a separate PR)

packages/medusa/src/api/store/products/helpers.ts
sradevski: Yeah it is more readable, it's just that calculated amount is not a big number here, so instead of converting I did this. Let me revisit it though, I agree with you

packages/medusa/src/api/store/products/helpers.ts
sradevski: Yes, I just need to check why the rate is nullable in the tax module, that's what the cast is for (and the TODO) 😄

packages/medusa/src/api/store/products/helpers.ts
olivermrbl: If the resources are indexed in the search module, the request should be (still to be tested) highly performant and support sorting across modules, e.g. sort product variant prices. This is one of its main purposes. If we can index the prices with taxes taken into account, I think that would be ideal and probably what we should aim for to solve this. The two approaches I outlined above are both temporary imo., I just want us to consider which one we feel most comfortable with in the near term. There is no pretty way to do this kind of stuff right now, which is why we have all of this middleware acrobatics going on. 

I agree we will need to introduce heavy usage of caching, but I think that's a separate topic in this context and something we will need to do irregardless of this.

olivermrbl: I think the idea with the separate endpoint was also to have a place to introduce even more dynamic calculations in the future, in case we need that. E.g. dynamic to the point of being real-time, which wouldn't be possible to index in the search engine. Will let @carlos-r-l-rodrigues pitch in here.

packages/medusa/src/api/store/products/helpers.ts
olivermrbl: Ah I see, makes sense 👍

packages/medusa/src/api/utils/middlewares/products/set-tax-context.ts
olivermrbl: As a separate data model in the pricing module right?

sradevski: Yes exactly, that is also rule-based (at least that was the last thing we discussed with Seb)

olivermrbl: Makes sense

packages/medusa/src/api/store/products/helpers.ts
carlos-r-l-rodrigues: Can we transform this into a workflow to fetch products?
This does not seem to be the case of middlewares, this is a cross domain calculation. It is more business logic than anything else at this point.

sradevski: I know where you are coming from, and I agree with having most of the BL outside the HTTP layer, but I don't think putting it in a workflow is the right solution here (or for any read-only operations). What we could do is create a simple helper function that wraps all of this logic and put it higher up the stack, WDYT?

packages/medusa/src/api/store/products/helpers.ts
```diff
@@ -30,3 +49,81 @@ export const maybeApplyStockLocationId = async (req: MedusaRequest, ctx) => {
...
+        return
+      }
+
+      return {
+        id: variant.id,
+        product_id: product.id,
+        product_name: product.title,
+        product_categories: product.categories?.map((c) => c.name),
+        // TODO: It is strange that we only accept a single category, revisit the tax module implementation
+        product_category_id: product.categories?.[0]?.id,
```
olivermrbl, line 119: comment: I don't even think we support categories in the tax module right now. Last I remember the rules are fixed to a few props, but maybe this has changed

packages/medusa/src/api/utils/middlewares/products/normalize-data-for-context.ts
olivermrbl: > If we do that, then the store shouldn't have a default region at all

Couldn't there be other use cases for a default region on the store? I am not it was added for the sole purpose of having a fallback for price calculations

olivermrbl (approved):
Overall LGTM. There's a lot going on in the HTTP layer, but it's hard to get around right now, so I am down for mergin as is

sradevski:
@olivermrbl yeah agreed, I think we need to decide how routes are constructed and where such logic lies (we have several places where the middlewares and handlers are quite heavy code-wise), but it can be an independent cleanup task

PR #7833, opened 2024-06-25
chore: various DML improvements

adrien2p:
**What**
- Add support for `searchable` modifier
- Replace instance of checks with static util checks
- Allow define config to take an object with name and table name

FIXES CORE-2395
FIXES CORE-2394
FIXES CORE-2389

### DIFF
--- a/packages/admin-next/dashboard/src/routes/product-variants/product-variant-detail/components/variant-prices-section/variant-prices-section.tsx
+++ b/packages/admin-next/dashboard/src/routes/product-variants/product-variant-detail/components/variant-prices-section/variant-prices-section.tsx
@@ -17,8 +17,8 @@ export function VariantPricesSection({ variant }: VariantPricesSectionProps) {
   const { t } = useTranslation()
 
   const prices = variant.prices
-    .filter((p) => !p.rules?.length)
-    .sort((p1, p2) => p1.currency_code?.localeCompare(p2.currency_code)) // display just currency prices
+    .filter((p) => !Object.keys(p.rules || {}).length) // display just currency prices
+    .sort((p1, p2) => p1.currency_code?.localeCompare(p2.currency_code))
 
   const [current, setCurrent] = useState(Math.min(prices.length, 3))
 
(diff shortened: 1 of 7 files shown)


olivermrbl (approved):
Very nice, LGTM

packages/core/utils/src/dml/entity.ts
```diff
@@ -16,8 +19,34 @@ import { BelongsTo } from "./relations/belongs-to"
...
   [IsDmlEntity]: true = true
 
+  name: string
+
+  readonly #tableName: string
   #cascades: EntityCascades<string[]> = {}
-  constructor(public name: string, public schema: Schema) {}
+
+  constructor(nameOrConfig: Config, public schema: Schema) {
+    if (isString(nameOrConfig)) {
```
sradevski: NIT: Maybe create a helper function `const {name, tableName} = getNameAndTable(nameOrConfig)` so we don't have so much logic in the constructor.

packages/core/utils/src/dml/entity.ts
```diff
@@ -16,8 +19,34 @@ import { BelongsTo } from "./relations/belongs-to"
...
+      const [schema, ...rest] = nameOrConfig.split(".")
+      const name = rest.length ? rest.join(".") : schema
+      this.name = toCamelCase(name)
```
sradevski: Let's rename `name` to `modelName` everywhere so it's clear what it is (I wasn't sure what it represents until I saw it below in the code)

adrien2p: I wont do this one because here it is the name of the object, not the model name. there is no notion of model until we reach the mikro orm helper and here we use the name to define the model name. Same thing as the name of a function or a class that we emulate here `Object.name` basically 😄

sradevski: But isn't this the DML model name? eg. in `const user = model.define("user", ...` `"user"` would be the name of the model (thus, the function being called `model.define`?

adrien2p: no it is not, same as when you passed the entity to the MedusaService, we were using the Class.name to infer everything, so using the Class.name which happen to be the model name in mikro orm, here it is the exact same behavior but instead of the name of the class (cause you can still pass mikro orm class or other classes if you want) we are using the DML.name.

sradevski: Got it 👍

sradevski (approved):

packages/core/utils/src/dml/entity.ts
adrien2p: To summarize, mikro orm infer the model name and table name from the class name unless configured differently, we are doing the same here, the name you are defining is the object name and we use that to infer the model name to have the exact same behaviour no matter if you are using a class, object, function etc. It makes everything works with a unique behavior and since the object don't have a name by default we are defining it ourselves. You can also take a look at the MedusaService which is based on name as well and not model name

PR #7835, opened 2024-06-25
chore: Migrate notification module to DML

sradevski:


packages/modules/notification/src/migrations/Migration20240627153332.ts
```diff
@@ -0,0 +1,24 @@
...
+    )
+    this.addSql(
+      'alter table if exists "notification_provider" alter column "channels" set default \'{}\';'
+    )
+
+    this.addSql(
+      'alter table if exists "notification" add column if not exists "updated_at" timestamptz not null default now(), add column "deleted_at" timestamptz null;'
+    )
+    this.addSql('drop index if exists "IDX_notification_idempotency_key";')
+    this.addSql('drop index if exists "IDX_notification_provider_id";')
```
sradevski: @riqwan I would have expected this to not be dropped, since it's a `belongsTo` relationship. Did I maybe misunderstand how it works?

riqwan: It shouldn't be removed. Let me take a look

riqwan: good catch - https://github.com/medusajs/medusa/pull/7858/files

packages/modules/notification/src/loaders/providers.ts
```diff
@@ -89,7 +88,7 @@ async function syncDatabaseProviders({
   validateProviders(normalizedProviders)
 
   try {
-    const providersInDb = await providerService.list({})
+    const providersInDb = await providerService.find()
```
olivermrbl: question: Why does this service have `find` and not `list`? Should be an internally generated service no?

packages/modules/notification/src/migrations/Migration20240628075401.ts
```diff
@@ -0,0 +1,23 @@
+import { Migration } from "@mikro-orm/migrations"
+
+export class Migration20240628075401 extends Migration {
+  async up(): Promise<void> {
```
olivermrbl, line 4: question: should we have `down` too?

packages/modules/notification/src/loaders/providers.ts
olivermrbl: Ah, I guess this is because we are resolving the repo instead. We try to avoid those leaking outside of services. 

If this is absolutely necessary, I would maybe suggest we rename the `providerService` to `providerRepo`. Up to you.

olivermrbl: But shouldn't we be able to do `resolve("notificationProviderService")`

olivermrbl (approved):
LGTM, just a few comments 💪

thetutlage:
LGTM!

PR #7836, opened 2024-06-25
feat: Add support for array properties DML

adrien2p:
**What**
Add support for `array` properties.
Also include a refactoring to lighten the mikro orm helper and instead split into utils

FIXES CORE-2396


riqwan (approved):
nice categorization! 🏅 lgtm!

PR #7837, opened 2024-06-25
chore: Migrate region module to use DML

olivermrbl:


packages/modules/region/src/models/country.ts
```diff
@@ -1,44 +1,19 @@
...
+    iso_3: model.text(),
+    num_code: model.text(),
+    name: model.text(),
+    display_name: model.text(),
+    region: model.belongsTo(() => Region, { mappedBy: "countries" }).nullable(),
+    metadata: model.json().nullable(),
+  }
+)
+
+// .indexes([{ fields: ["region_id", "iso_2"], unique: true, name: "IDX_region_country_region_id_iso_2_unique" ])
```
olivermrbl: PR blocked by missing composite indexes API

packages/modules/region/src/models/country.ts
```diff
@@ -1,44 +1,28 @@
...
-  region?: Region | null
-}
+const Country = model
+  .define(
+    { name: "Country", tableName: "region_country" },
+    {
+      iso_2: model.text().primaryKey(),
+      iso_3: model.text(),
+      num_code: model.text(),
+      name: model.text(),
```
sradevski: Should the country code and name be searchable as well?

sradevski (approved):
Nice 👍

riqwan (approved):
nice!

packages/modules/region/src/models/country.ts
olivermrbl: Yeah, I guess it doesn't hurt. Will add that in

PR #7839, opened 2024-06-25
feat(utils): infer primaryKeys from a DML model

riqwan:
what:

- depending on other properties in a DML model, we infer primaryKeys between id properties and primaryKey-able properties. 

```
Example:
  Model 1:
    id: model.id() -> primary key
    code: model.text()

  Model 2:
    id: model.id()
    code: model.text().primaryKey() -> primary key

  Model 3:
    id: model.id()
    code: model.text().primaryKey() -> composite primary key
    name: model.text().primaryKey() -> composite primary key
```


packages/core/utils/src/dml/helpers/entity-builder/infer-primary-key-properties.ts
```diff
@@ -0,0 +1,70 @@
...
+    id: model.id(), -> implicit primaryKey field,
+    text: model.text(),
+    textPrimary: model.text().primaryKey(), -> explicit primaryKey field
+    numberPrimary: model.number().primaryKey(), -> explicit primaryKey field
+    belongsTo: model.belongsTo(() => belongsToAnother),
+  })
+*/
+function getExplicitPrimaryKeyFields(schema: DMLSchema) {
+  return Object.entries(schema).filter(([field, property]) => {
+    if (property instanceof IdProperty) {
```
carlos-r-l-rodrigues: Nothing to be done in this PR, but maybe in the future we need to replace these instanceof with a `__type` property to hold the type of the property and avoid issues if multiple versions of the same Node package are loaded.

adrien2p: I have cleaned them all already in develop, we need to do the same here ☺️ there was no more instance of ahah

olivermrbl: +1 - we already started replacing the instanceof for the relations classes, because it was causing issues and didn't detect matches. 

Think we might as well change this now, since it will likely cause issues almost right away for someone.

riqwan: TIL 😄 I was just thinking "that reads very pretty"

riqwan: changed 🤞🏻

packages/core/utils/src/dml/helpers/entity-builder/infer-primary-key-properties.ts
```diff
@@ -0,0 +1,74 @@
...
+  // set to primaryKey.
+  for (const [field, property] of Object.entries(schema)) {
+    const parsed = property.parse(field)
+    const isRelationshipType = "type" in parsed
+
+    if (isRelationshipType) {
+      continue
+    }
+
+    if (parsed.dataType.name === "id") {
```
adrien2p, line 43: Nit: for consistency and make it easier can you use the same approach as the other ones and if there is no type you can add a symbol.for to which we can refer too without being shown in the auto completion and then we can use it every time we need without having to think about which value to compare to know that it is an id.

You can look at the dml, properties and relations they all provide a static method ☺️

adrien2p (approved):

packages/core/utils/src/dml/helpers/entity-builder/infer-primary-key-properties.ts
riqwan: I gave this a shot, but doesn't look nice when i tried it. Lets pair on this in the next PR. 👍🏻

PR #7842, opened 2024-06-26
feat(utils,types): DML can apply composite indexes

riqwan:
**what:**

- DML can apply composite indexes
- Where clause is currently a string, QB version will come as a follow up

```
model.define("user", {
  email: model.text(),
  account: model.text(),
}).indexes([
  {
    name: "IDX-unique-name",
    unique: true,
    on: ["email", "account"],
    where: "email is NOT NULL",
  },
])
```

RESOLVES CORE-2391

### DIFF
--- a/packages/core/utils/src/dml/__tests__/entity-builder.spec.ts
+++ b/packages/core/utils/src/dml/__tests__/entity-builder.spec.ts
@@ -862,6 +862,7 @@ describe("Entity builder", () => {
           items: expect.any(Function),
           nullable: false,
           name: "role",
+          type: "string",
         },
         created_at: {
           reference: "scalar",
@@ -970,6 +971,7 @@ describe("Entity builder", () => {
           items: expect.any(Function),
           nullable: false,
           name: "role",
+          type: "string",
         },
         created_at: {
           reference: "scalar",
@@ -1082,6 +1084,7 @@ describe("Entity builder", () => {
           items: expect.any(Function),
           nullable: true,
           name: "role",
+          type: "string",
         },
         created_at: {
           reference: "scalar",
--- a/packages/core/utils/src/dml/helpers/entity-builder/define-property.ts
+++ b/packages/core/utils/src/dml/helpers/entity-builder/define-property.ts
@@ -12,6 +12,7 @@ import {
   OnInit,
   PrimaryKey,
   Property,
+  Utils,
 } from "@mikro-orm/core"
 
 /**
@@ -152,6 +153,7 @@ export function defineProperty(
     Enum({
       items: () => field.dataType.options!.choices,
       nullable: field.nullable,
+      type: Utils.getObjectType(field.dataType.options!.choices[0]),
       /**
        * MikroORM does not ignore undefined values for default when generating
        * the database schema SQL. Conditionally add it here to prevent undefined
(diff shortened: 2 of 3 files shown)


packages/core/types/src/dml/index.ts
```diff
@@ -163,3 +163,10 @@ export type InferTypeOf<T extends IDmlEntity<any>> = InstanceType<Infer<T>>
 export type InferEntityType<T extends any> = T extends IDmlEntity<any>
   ? InferTypeOf<T>
   : T
+
+export type EntityIndex = {
+  name?: string
+  unique?: boolean
+  fields: string[]
```
shahednasser: thought: I suggest going back to using `on` instead of `fields`. According to the DML terminology, we refer to the items in a schema as properties, not fields. I think we should be consistent with that.

riqwan: Agree on consolidating the terminology. 💯 Is there a docu on this yet that I can read?

From what I see on the codebase, we have this:

```
define('modelName', {
  [field]: model.text() -> property
})
```

We seem to define the key of the schema as a field and the value of the schema field a property.

shahednasser: I read it in the Notion document that Virk wrote.

The field/property detail from the codebase is interesting, but ultimately, the developer doesn't need to know its details. If we call it a property/field, then it's best to stick with the choice.

riqwan: @shahednasser should we switch to on?

shahednasser: I think that's better

riqwan: done! 🥇

shahednasser (approved):
LGTM!

packages/core/utils/src/dml/__tests__/entity-builder.spec.ts
```diff
@@ -2484,6 +2519,145 @@ describe("Entity builder", () => {
...
+
+        toMikroORMEntity(user)
+
+        throw "should not reach"
+      } catch (e) {
+        expect(e.message).toEqual(
+          "Fields (doesnotexist, anotherdoesnotexist) are not found when applying indexes from DML entity"
+        )
+      }
+    })
```
adrien2p, line 2658:
```suggestion
        const group = model.define("group", {
          id: model.number(),
          name: model.text(),
          users: model.hasMany(() => user),
        })

        const user = model
          .define("user", {
            email: model.text(),
            account: model.text(),
            organization: model.text(),
            group: model.belongsTo(() => group, { mappedBy: "users" }),
          })
          .indexes([
            {
              on: ["email", "account", "doesnotexist", "anotherdoesnotexist"],
            },
          ])

    let err: any
    try {
        toMikroORMEntity(user)
      } catch (e) {
        err = e
      }
      
        expect(e.message).toEqual(
          "Fields (doesnotexist, anotherdoesnotexist) are not found when applying indexes from DML entity"
        )
    })
```

packages/core/utils/src/dml/entity.ts
```diff
@@ -119,4 +120,14 @@ export class DmlEntity<Schema extends DMLSchema> implements IDmlEntity<Schema> {
     this.#cascades = options
     return this
   }
+
+  indexes(indexes: EntityIndex<Schema>[]) {
+    for (const index of indexes) {
+      index.where = transformIndexWhere(index)
+      index.unique = index.unique ?? false
```
adrien2p, line 127:
```suggestion
      index.unique ??= false
```

adrien2p (approved):

PR #7844, opened 2024-06-26
chore: infer fields on indexes from schema

riqwan:
what: 

- allows fields in indexes to be inferred from schema

Opening this as a separate PR to provide additional feedback on how to better typescript 😄 

Screenshot 2024-06-26 at 15 04 23

### DIFF
--- a/packages/core/utils/src/dml/__tests__/entity-builder.spec.ts
+++ b/packages/core/utils/src/dml/__tests__/entity-builder.spec.ts
@@ -2537,7 +2537,7 @@ describe("Entity builder", () => {
           {
             name: "IDX_unique-name",
             unique: true,
-            fields: ["organization", "test"],
+            fields: ["organization", "account"],
           },
         ])
 


packages/core/types/src/dml/index.ts
```diff
@@ -164,9 +169,9 @@ export type InferEntityType<T extends any> = T extends IDmlEntity<any>
   ? InferTypeOf<T>
   : T
 
-export type EntityIndex = {
+export type EntityIndex<TSchema extends DMLSchema = DMLSchema> = {
   name?: string
   unique?: boolean
-  fields: string[]
+  fields: (keyof TSchema)[]
```
adrien2p: suggestion: Can we maybe use `Infer<TSchema>` instead? so that you can see the future generated foreign keys as well

packages/core/utils/src/dml/helpers/entity-builder/build-indexes.ts
```diff
@@ -1,14 +1,15 @@
...
   The DML provides an opinionated soft deletable entity as a part of every model
   We assume that deleted_at would be scoped in indexes in all cases as an index without the scope
   doesn't seem to be valid. If a case presents itself where one would like to remove the scope, 
   this will need to be updated to include that case.
 */
-export function transformIndexWhere(index: EntityIndex) {
+export function transformIndexWhere<TSchema extends DMLSchema>(
+  index: EntityIndex<TSchema>
+) {
   let where = index.where
```
adrien2p, line 13:
```suggestion
  let where = index.where + ''
```

adrien2p: such that you can remove the check on where line 20

adrien2p: and line 26 you can then change to `!where`

adrien2p (approved):

packages/core/types/src/dml/index.ts
adrien2p: unless for that we handle it automatically as I suggested in slack 😄 let me know, and if they want to add some more indexes related to foreign keys then maybe they go with the .index on the relation? let me know

packages/core/utils/src/dml/helpers/entity-builder/build-indexes.ts
riqwan: will bring this to the base PR 🤞🏻

packages/core/types/src/dml/index.ts
riqwan: resolved over huddle 💪🏻

riqwan:
Thanks to @adrien2p we now have belongs_to foreign key in the type as well. 🔥 

Screenshot 2024-06-26 at 16 31 20

adrien2p (approved):

packages/core/utils/src/dml/__tests__/entity-builder.spec.ts
```diff
@@ -2537,7 +2537,7 @@ describe("Entity builder", () => {
           {
             name: "IDX_unique-name",
             unique: true,
-            fields: ["organization", "account"],
+            fields: ["organization", "test"],
```
adrien2p: Q: should we change to account?

adrien2p (approved):

packages/core/utils/src/dml/__tests__/entity-builder.spec.ts
riqwan:
```suggestion
            fields: ["organization", "account"],
```

PR #7846, opened 2024-06-26
fix: DML Enum property

adrien2p:
FIXES CORE-2381


adrien2p:
@shahednasser here is a fix for the enum

riqwan (approved):
nice!

PR #7848, opened 2024-06-26
feat(dashboard, order, medusa, types, js-sdk): Request return e2e flow

fPolic:
**What**
- add create return flow
- fix routing
- fix migrations issue with existing enum types

**Note**
- `custom_price` can be set but it cannot be inferred back when we refresh since the field is not saved explicitly in the database

---

- DEPENDS ON https://github.com/medusajs/medusa/pull/8190

---

CLOSES CC-73
CLOSES CC-240

### DIFF
--- a/packages/admin-next/dashboard/src/i18n/translations/en.json
+++ b/packages/admin-next/dashboard/src/i18n/translations/en.json
@@ -925,7 +925,9 @@
           "items_other": "{{count}} items"
         },
         "return": {
-          "created": "Return created"
+          "created": "Return created",
+          "items_one": "{{count}} item returned",
+          "items_other": "{{count}} items returned"
         },
         "note": {
           "comment": "Comment",
--- a/packages/admin-next/dashboard/src/routes/orders/order-detail/components/order-activity-section/order-timeline.tsx
+++ b/packages/admin-next/dashboard/src/routes/orders/order-detail/components/order-activity-section/order-timeline.tsx
@@ -5,11 +5,12 @@ import { PropsWithChildren, ReactNode, useMemo, useState } from "react"
 import { Link } from "react-router-dom"
 
 import { XMarkMini } from "@medusajs/icons"
-import { AdminFulfillment, AdminOrder } from "@medusajs/types"
+import { AdminFulfillment, AdminOrder, AdminReturn } from "@medusajs/types"
 import { useTranslation } from "react-i18next"
 
 import { useDate } from "../../../../../hooks/use-date"
 import { getStylizedAmount } from "../../../../../lib/money-amount-helpers"
+import { useReturns } from "../../../../../hooks/api/returns"
 
 type OrderTimelineProps = {
   order: AdminOrder
@@ -79,6 +80,8 @@ type Activity = {
 
 const useActivityItems = (order: AdminOrder) => {
   const { t } = useTranslation()
+
+  const { returns = [] } = useReturns({ order_id: order.id, fields: "*items" })
 
   const notes = []
   const isLoading = false
@@ -165,16 +168,13 @@ const useActivityItems = (order: AdminOrder) => {
       }
     }
 
-    /**
-     * TODO: revisit when API is fixed to fetch returns of an order
-     */
-
-    // for (const ret of order.returns) {
-    //   items.push({
-    //     title: t("orders.activity.events.return.created"),
-    //     timestamp: ret.created_at,
-    //   })
-    // }
+    for (const ret of returns) {
+      items.push({
+        title: t("orders.activity.events.return.created"),
+        timestamp: ret.created_at,
+        children: <ReturnCreatedBody orderReturn={ret} />,
+      })
+    }
 
     // for (const note of notes || []) {
     //   items.push({
@@ -385,6 +385,24 @@ const FulfillmentCreatedBody = ({
     <div>
       <Text size="small" className="text-ui-fg-subtle">
         {t("orders.activity.events.fulfillment.items", {
+          count: numberOfItems,
+        })}
+      </Text>
+    </div>
+  )
+}
+
+const ReturnCreatedBody = ({ orderReturn }: { orderReturn: AdminReturn }) => {
+  const { t } = useTranslation()
+
+  const numberOfItems = orderReturn.items.reduce((acc, item) => {
+    return acc + item.quantity
+  }, 0)
+
+  return (
+    <div>
+      <Text size="small" className="text-ui-fg-subtle">
+        {t("orders.activity.events.return.items", {
           count: numberOfItems,
         })}
       </Text>


srindom:
Two things from watching the screen recording:
1. The empty state for items to return doesn't align with the design ([Link](https://www.figma.com/design/6iTfljFmXXsdYJV4mjUXX2/RMA-Flows-Q2-2024?node-id=1018-1024&t=nlau55XRsQsMLXj1-11))
2. When the return has been requested the order details page should reflect the change - in your recording it looks like nothing happened.

> 1. to test the flow, a user needs to create a return reason in the database
> 2. I'm not sure how we would infer the shipping price once the shipping option is selected and also how to display tax percentage when user inputs the price of the shipping

1. A return reason should not be required. 
2. Can you elaborate on what you mean by inferring the shipping price? I would expect the workflow that adds shipping to the order change to calculate the price.

fPolic:
> Two things from watching the screen recording:
> 
> 1. The empty state for items to return doesn't align with the design ([Link](https://www.figma.com/design/6iTfljFmXXsdYJV4mjUXX2/RMA-Flows-Q2-2024?node-id=1018-1024&t=nlau55XRsQsMLXj1-11))

Have that in mind, I was waiting for Ludvig to export an image I could use to for this that but it had a lower priority. Will implement this with pure CSS.

> 2. When the return has been requested the order details page should reflect the change - in your recording it looks like nothing happened.

I was planning to implement this in an immediate followup so I'm not blocking this PR. I'm currently on Create shipping flow and will tackle this afterwards.

> 3. A return reason should not be required.

Good catch, will push a fix.

> 4. Can you elaborate on what you mean by inferring the shipping price? I would expect the workflow that adds shipping to the order change to calculate the price.

Ahh makes sense (I was expecting that when the user changes the shipping option, we calculate the price on the client...).

srindom:
>I was planning to implement this in an immediate followup so I'm not blocking this PR. I'm currently on Create shipping flow and will tackle this afterwards.

Perfectly fine 👍

srindom:
>Ahh makes sense (I was expecting that when the user changes the shipping option, we calculate the price on the client...)

I thought the return flow was implemented as a multi-request flow, but understand why this is difficult right now. Let's not do anything to crazy in the client for now and then refactor when we have the multi-request flow in place.

packages/admin-next/dashboard/src/routes/orders/order-create-return/components/add-return-items-table/add-return-items-table.tsx
```diff
@@ -0,0 +1,275 @@
...
+    return results.slice(offset, offset + limit)
+  }, [items, currencyCode, searchParams])
+
+  const columns = useReturnItemTableColumns(currencyCode)
+  const filters = useReturnItemTableFilters()
+
+  const { table } = useDataTable({
+    data: queriedItems as AdminOrderLineItem[],
+    columns: columns,
+    count: items.length,
```
kasperkristensen:
```suggestion
    count: queriedItems.length,
```

packages/core/types/src/http/return-reason/common.ts
```diff
@@ -0,0 +1,23 @@
+import { BaseFilterable, OperatorMap } from "../../dal"
+
+export interface BaseReturnReason {
+  id: string
+  value: string
+  label: string
+  description?: string
+  metadata?: Record<string, any> | null
+  created_at?: string
+  updated_at?: string
```
kasperkristensen: q: These can't be undefined right?

packages/core/types/src/http/return-reason/common.ts
```diff
@@ -0,0 +1,23 @@
+import { BaseFilterable, OperatorMap } from "../../dal"
+
+export interface BaseReturnReason {
+  id: string
+  value: string
+  label: string
+  description?: string
```
kasperkristensen: q: Can this be null?

packages/core/types/src/http/return-reason/common.ts
```diff
@@ -0,0 +1,23 @@
...
+  extends BaseFilterable<BaseReturnReason> {
+  q?: string
+  id?: string[] | string | OperatorMap<string | string[]>
+  value?: string | OperatorMap<string>
+  label?: string | OperatorMap<string>
+  description?: string | OperatorMap<string>
+  parent_return_reason_id?: string | OperatorMap<string | string[]>
+  created_at?: OperatorMap<string>
+  updated_at?: OperatorMap<string>
+}
```
kasperkristensen: Don't seem like this is really used for anything, as it's just being re-exported from the admin file, but in there there is a separate definition for AdminReturnReasonListParams

packages/core/types/src/http/return-reason/admin.ts
```diff
@@ -0,0 +1,27 @@
...
+  value: string
+  label: string
+  description?: string
+}
+
+export interface AdminReturnReasonsResponse {
+  return_reasons: AdminReturnReason[]
+}
+
+export interface AdminReturnReasonListParams extends FindParams {
```
kasperkristensen: needs to also extend BaseFilterable

packages/core/types/src/http/return-reason/admin.ts
```diff
@@ -0,0 +1,27 @@
+import { BaseReturnReason, BaseReturnReasonFilters } from "./common"
+import { FindParams } from "../common"
+import { OperatorMap } from "../../dal"
+
+export interface AdminReturnReason extends BaseReturnReason {}
+export interface AdminReturnReasonFilters extends BaseReturnReasonFilters {}
```
kasperkristensen: rm

packages/core/types/src/http/return-reason/admin.ts
```diff
@@ -0,0 +1,27 @@
+import { BaseReturnReason, BaseReturnReasonFilters } from "./common"
```
kasperkristensen: suggestion: follow file structure of other domains:

```
- admin
  -- entities.ts
  -- queries.ts
  -- payloads.ts
  -- responses.ts
  -- index.ts
- common.ts
- index.ts
```

packages/core/types/src/http/return/admin.ts
```diff
@@ -22,6 +22,10 @@ export interface AdminReturnResponse {
   items: BaseReturnItem[]
```
kasperkristensen: suggestion: Use AdminBaseReturnItem, to keep the naming consistent

packages/admin-next/dashboard/src/routes/orders/order-create-return/components/return-create-form/return-item.tsx
```diff
@@ -0,0 +1,243 @@
...
+                    )
+                  }}
+                />
+              </div>
+              <IconButton
+                type="button"
+                className="flex-shrink"
+                variant="transparent"
+                onClick={() => {
+                  form.setValue(`items.${index}.note`, null)
```
kasperkristensen:
```suggestion
                  form.setValue(`items.${index}.note`, null, { shouldDirty: true, shouldTouch: true })
```

packages/admin-next/dashboard/src/routes/orders/order-create-return/components/return-create-form/return-create-form.tsx
```diff
@@ -0,0 +1,694 @@
...
+                    onSelectionChange={(s) => (selectedItems = s)}
+                  />
+                  <StackedFocusModal.Footer>
+                    <div className="flex w-full items-center justify-end gap-x-4">
+                      <div className="flex items-center justify-end gap-x-2">
+                        <RouteFocusModal.Close
+                          asChild
+                          onClick={(e) => {
+                            e.stopPropagation()
+                          }}
```
kasperkristensen: Not needed if you set the ***type*** on the child button to `button`

packages/admin-next/dashboard/src/routes/orders/order-create-return/components/return-create-form/return-create-form.tsx
```diff
@@ -0,0 +1,694 @@
...
+              />
+            </div>
+          </div>
+        </RouteFocusModal.Body>
+        <RouteFocusModal.Footer>
+          <div className="flex w-full items-center justify-end gap-x-4">
+            <div className="flex items-center justify-end gap-x-2">
+              <RouteFocusModal.Close asChild>
+                <Button
+                  // onClick={() => cancelReturnRequest()}
```
kasperkristensen: type="button"

olivermrbl:
@fPolic, this PR will not include return receival and the activity list changes, right? For the latter see image below:

CleanShot 2024-07-24 at 16 22 34@2x

fPolic:
> @fPolic, this PR will not include return receival and the activity list changes, right? For the latter see image below:

I started working on the receival flow in a separate branch. Regarding activity list, good reminder, I'll add it to this PR.

olivermrbl:
Cool, sounds good!

packages/modules/order/src/migrations/Migration20240604100512.ts
```diff
@@ -129,15 +129,17 @@ export class Migration20240604100512 extends Migration {
...
-    
-    
 
-    CREATE TYPE return_status_enum AS ENUM (
-        'requested',
-        'received',
-        'partially_received',
-        'canceled'
-    );
+    DO $$
```
olivermrbl, line 133: question: What triggered these changes?

packages/admin-next/dashboard/src/lib/rma.ts
```diff
@@ -1,75 +1,15 @@
...
-      }),
-    new Map<string, Omit<LineItem, "beforeInsert">>()
+import { AdminOrderLineItem } from "@medusajs/types"
+
+export function getReturnableQuantity(item: AdminOrderLineItem): number {
+  const {
+    // TODO: this should be `fulfilled_quantity`? now there is check on the BD that we can't return more quantity than we shipped but some items don't require shipping
+    shipped_quantity,
+    return_received_quantity,
+    return_dismissed_quantity, // TODO: check this
```
olivermrbl, line 8: A few TODOs here, just adding a comment to ensure we are aware of these

olivermrbl (approved):

packages/modules/order/src/migrations/Migration20240604100512.ts
fPolic: I had migration failing because I already had that type in the DB but since Postgres doesn't support `create type if not exist...` this is equivalent.

PR #7849, opened 2024-06-26
feat(utils,types): DML Index can generate where SQL from a query builder

riqwan:
what:

- introduces a simple query builder
- uses the query builder to tranform an object in where to SQL when applying indexes

```
Examples:
  { where: { column: null } } -> column IS NULL
  { where: { column: { $ne: null } } } -> column IS NOT NULL
  { where: { boolean_column: true } } -> boolean_column IS TRUE
  { where: { column: "value", another_column: { $ne: 30 } } } -> column = "value" AND another_column != 30
```

```
const user = model
  .define("user", {
    email: model.text(),
    account: model.text(),
    organization: model.text(),
  })
  .indexes([
    {
      on: ["organization", "account"],
      where: { email: { $ne: null } },
    },
    {
      name: "IDX-email-account-special",
      on: ["organization", "account"],
      where: {
        email: { $ne: null },
        account: null,
      },
    },
```

RESOLVES CORE-2392

### DIFF
--- a/integration-tests/http/__tests__/inventory/admin/inventory.spec.ts
+++ b/integration-tests/http/__tests__/inventory/admin/inventory.spec.ts
@@ -581,6 +581,79 @@ medusaIntegrationTestRunner({
             })
           )
         })
+        
+        it("should retrieve the inventory item with correct stocked quantity given location levels have been deleted", async () => {
+          await api.post(
+            `/admin/inventory-items/${inventoryItem1.id}/location-levels`,
+            {
+              location_id: stockLocation1.id,
+              stocked_quantity: 10,
+              incoming_quantity: 0,
+            },
+            adminHeaders
+          )
+
+          const item = (
+            await api.post(
+              `/admin/inventory-items/${inventoryItem1.id}/location-levels`,
+              {
+                location_id: stockLocation2.id,
+                stocked_quantity: 10,
+                incoming_quantity: 0,
+              },
+              adminHeaders
+            )
+          ).data.inventory_item
+
+          await api.post(
+            `/admin/reservations`,
+            {
+              line_item_id: "line-item-id-1",
+              inventory_item_id: inventoryItem1.id,
+              location_id: stockLocation2.id,
+              description: "test description",
+              quantity: 1,
+            },
+            adminHeaders
+          )
+
+          const reservation = (
+            await api.post(
+              `/admin/reservations`,
+              {
+                line_item_id: "line-item-id-2",
+                inventory_item_id: inventoryItem1.id,
+                location_id: stockLocation2.id,
+                description: "test description 2",
+                quantity: 1,
+              },
+              adminHeaders
+            )
+          ).data.reservation
+
+          await api.delete(
+            `/admin/inventory-items/${inventoryItem1.id}/location-levels/${item.location_levels[0].id}`,
+            adminHeaders
+          )
+
+          await api.delete(
+            `/admin/reservations/${reservation.id}`,
+            adminHeaders
+          )
+
+          const response = await api.get(
+            `/admin/inventory-items/${inventoryItem1.id}`,
+            adminHeaders
+          )
+
+          expect(response.data.inventory_item).toEqual(
+            expect.objectContaining({
+              id: inventoryItem1.id,
+              stocked_quantity: 10,
+              reserved_quantity: 1,
+            })
+          )
+        })
 
         it("should throw if inventory item doesn't exist", async () => {
           const error = await api
--- a/packages/modules/inventory-next/src/models/inventory-item.ts
+++ b/packages/modules/inventory-next/src/models/inventory-item.ts
@@ -131,14 +131,14 @@ export class InventoryItem {
 
   @Formula(
     (item) =>
-      `(SELECT SUM(reserved_quantity) FROM inventory_level il WHERE il.inventory_item_id = ${item}.id)`,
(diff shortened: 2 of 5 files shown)


packages/core/utils/src/dml/helpers/entity-builder/query-builder.ts
```diff
@@ -0,0 +1,57 @@
...
+  }
+
+  return conditions.join(" AND ")
+}
+
+function isQueryCondition(value: any): value is QueryCondition {
+  return typeof value === "object" && value !== null && !("$ne" in value)
+}
+
+function buildCondition(key: string, value: QueryValue): string {
```
adrien2p, line 34: Suggestion: i would add the boolean support through is true is false because for nullable boolean you can have null and != false would gives you true and null boolean.

packages/core/utils/src/dml/helpers/entity-builder/query-builder.ts
```diff
@@ -0,0 +1,57 @@
...
+  if (value === null) {
+    return `${key} IS NULL`
+  } else if (typeof value === "object" && "$ne" in value) {
```
adrien2p: Todo: you can use isObject, otherwise null is of type object as well for example

packages/core/utils/src/dml/helpers/entity-builder/query-builder.ts
```diff
@@ -0,0 +1,57 @@
...
+    if (value.$ne === null) {
+      return `${key} IS NOT NULL`
+    } else {
+      return `${key} != ${formatValue(value.$ne)}`
+    }
+  } else {
+    return `${key} = ${formatValue(value)}`
```
adrien2p, line 48: Q: What happen for undefined?

adrien2p (approved):

packages/core/types/src/dml/index.ts
```diff
@@ -180,9 +180,20 @@ export type InferIndexableProperties<T> = keyof (T extends IDmlEntity<
     } & InferForeignKeys<T>
   : never)
 
-export type EntityIndex<TSchema extends DMLSchema = DMLSchema> = {
+export type EntityIndex<
+  TSchema extends DMLSchema = DMLSchema,
+  TWhere = string
```
adrien2p, line 185: Suggestion : shouldn't it be string or QueryConditions? And the keys for record are the inferInxableProperties wdyt?

packages/core/utils/src/dml/__tests__/entity-builder.spec.ts
```diff
@@ -2626,35 +2626,94 @@ describe("Entity builder", () => {
...
+        toMikroORMEntity(user)
       } catch (e) {
-        expect(e.message).toEqual(
-          "Fields (doesnotexist, anotherdoesnotexist) are not found when applying indexes from DML entity"
-        )
+        err = e
       }
+
+      expect(err.message).toEqual(
+        "Fields (doesnotexist, anotherdoesnotexist) are not found when applying indexes from DML entity"
```
adrien2p: Suggestion: cannot apply indexes (x, y) on field from entity entityName

packages/core/types/src/dml/index.ts
```diff
@@ -180,9 +180,20 @@ export type InferIndexableProperties<T> = keyof (T extends IDmlEntity<
...
   on: InferIndexableProperties<IDmlEntity<TSchema>>[]
-  where?: string
+  where?: TWhere
+}
+
+export type SimpleQueryValue = string | number | boolean | null
+export type NeQueryValue = { $ne: SimpleQueryValue }
+export type QueryValue = SimpleQueryValue | NeQueryValue
+
+export interface QueryCondition {
```
adrien2p, line 197: Suggestion: could take schema and restrict properties to InferIndexable wdyt?

packages/core/types/src/dml/index.ts
riqwan: only at the entry point, everywhere else it'll be a string. I've added an override at the entry point

packages/core/utils/src/dml/helpers/entity-builder/query-builder.ts
riqwan: We ignore it. Nothing much we can do with an undefined.

adrien2p: Agreed, didnt see where it was dropped 😅

packages/core/types/src/dml/index.ts
adrien2p: Ok got it 😉

packages/core/types/src/dml/index.ts
riqwan: I'll do this in a separate PR similar to the previous try. I'll pick it up tomorrow 🤞🏻

packages/core/utils/src/dml/__tests__/entity-builder.spec.ts
olivermrbl: Are we sure that's what we want? To me, the "apply indexes (x, ,y) on field" sounds a little bit off.

Shouldn't it be:
`Cannot apply indexes on fields (x, y) for model User`

adrien2p: Sorry the x, y are the field, i ve typed it wrong

shahednasser (approved):

sradevski (approved):
Nice one 👍

PR #7852, opened 2024-06-26
feat: Completely revamp the pricing module

sradevski:
This will be a big breaking change, but it's better done sooner than later.
This PR achieves the following:

1. Removes the concept of rule types (will be reintroduced in a much simpler format, used only in the calculatePrices function without any relations). This also doesn't require creating the rules before using them, which simplifies the UX
2. Removes custom services
3. Improved performance of price calculation (both because of joins with price values and rule types, but also removed a redundant join with prices which doubled the amount of data processed)
4. Brings the rules data model closer to the rest of our rule-based models (promotions, fulfillment)
5. Some bug fixes in create/update flows (eg. rules count not being updated correctly)
6. Simplifies the code overall

### DIFF
--- a/packages/modules/pricing/src/migrations/.snapshot-medusa-pricing.json
+++ b/packages/modules/pricing/src/migrations/.snapshot-medusa-pricing.json
@@ -635,14 +635,14 @@
       "schema": "public",
       "indexes": [
         {
-          "keyName": "IDX_price_rule_price_id_rule_type_id_unique",
+          "keyName": "IDX_price_rule_price_id_attribute_unique",
           "columnNames": [
             "price_id"
           ],
           "composite": false,
           "primary": false,
           "unique": false,
-          "expression": "CREATE UNIQUE INDEX IF NOT EXISTS \"IDX_price_rule_price_id_rule_type_id_unique\" ON \"price_rule\" (price_id, rule_type_id) WHERE deleted_at IS NULL"
+          "expression": "CREATE UNIQUE INDEX IF NOT EXISTS \"IDX_price_rule_price_id_attribute_unique\" ON \"price_rule\" (price_id, attribute) WHERE deleted_at IS NULL"
         },
         {
           "keyName": "IDX_price_rule_deleted_at",
--- a/packages/modules/pricing/src/migrations/Migration20240626133555.ts
+++ b/packages/modules/pricing/src/migrations/Migration20240626133555.ts
@@ -48,6 +48,11 @@ export class Migration20240626133555 extends Migration {
     )
     this.addSql('drop index if exists "IDX_price_rule_price_set_id";')
     this.addSql('drop index if exists "IDX_price_rule_rule_type_id";')
+    this.addSql('drop index if exists "IDX_price_rule_price_id_unique";')
+    this.addSql(
+      'CREATE UNIQUE INDEX IF NOT EXISTS "IDX_price_rule_price_id_attribute_unique" ON "price_rule" (price_id, attribute) WHERE deleted_at IS NULL;'
+    )
+
     this.addSql(
       'alter table if exists "price_rule" drop column if exists "price_set_id";'
     )
--- a/packages/modules/pricing/src/models/price-rule.ts
+++ b/packages/modules/pricing/src/models/price-rule.ts
@@ -28,7 +28,7 @@ const PriceRuleDeletedAtIndex = createPsqlIndexStatementHelper({
 
 const PriceRulePriceIdIndex = createPsqlIndexStatementHelper({
   tableName: tableName,
-  columns: ["price_id", "rule_type_id"],
+  columns: ["price_id", "attribute"],
   where: "deleted_at IS NULL",
   unique: true,
 })


sradevski:
@riqwan thanks for checking it, I know it's a lot to take in 😄 

@olivermrbl do you think it is worth spending time on having a data migration so this is non breaking? I do agree it would be nice to do, but it will take a bit of time to prepare a query to do that.

olivermrbl:
How much work will follow this PR? I would love to understand the end result of the revamped pricing module, as I believe it would ease the review process. 

Could I get you to outline that (if any) in detail? 🙏

sradevski:
@olivermrbl The PR is ready as it is, so any follow-up work will just be regular enhancements and improvements.

I am working on a couple of fixes now, but they are unrelated to this PR (although this PR makes it a lot easier to implement them)

The only thing we might want to add (but as we discussed yesterday, even this can be done at a later date) is the ability to be able to specify weights per attribute and use it for ranking, which would look something like this:
1. A table that has `attribute, weight` fields. Simple CRUD on it
2. In `calculatePrices` we will load all data from this table (or cache it, as it will unlikely have more than 5-10 rows), and use those values in the last part of the calculation (the sorting)

And that's it, like that we will support weight-based ranking.

There are few more improvements we can do, but those are also purely additive and optional:
1. The `price_set` won't hold any prices. Instead, we will have a `price_list` with a `DEFAULT` type (or simply with empty rules). This will further simplify the code, as we have two places where prices are managed now. This is just an idea at this point, I need to make sure this will work as I imagine it.
2. Unify the rules approach across Medusa. Again, this is something that is unrelated to this PR, but this PR does simplify doing that (and eg. the price_list_rules are now more similar to what we have in fulfillment, minus the `operator` field)

Let me know if that makes sense and if you need more info.

sradevski:
@riqwan @olivermrbl I also added a data migration now (see last commit) so it's not a breaking change anymore

olivermrbl:
> The PR is ready as it is, so any follow-up work will just be regular enhancements and improvements.

Cool thanks for sharing. Going through the PR now.

An additional thought while it's top of mind: We should probably get the PriceList UI PR merged first and then ensure that everything works with this PR before we move forward with it.

packages/modules/pricing/src/migrations/.snapshot-medusa-pricing.json
```diff
@@ -599,41 +631,31 @@
...
+          "keyName": "IDX_price_rule_price_id_rule_type_id_unique",
           "columnNames": [
-            "rule_type_id"
+            "price_id"
           ],
           "composite": false,
           "primary": false,
           "unique": false,
-          "expression": "CREATE INDEX IF NOT EXISTS \"IDX_price_set_rule_type_rule_type_id\" ON \"price_set_rule_type\" (rule_type_id) WHERE deleted_at IS NULL"
+          "expression": "CREATE UNIQUE INDEX IF NOT EXISTS \"IDX_price_rule_price_id_rule_type_id_unique\" ON \"price_rule\" (price_id, rule_type_id) WHERE deleted_at IS NULL"
```
riqwan: todo: This index seems to be wrong as it still has the rule_type_id

packages/modules/pricing/src/migrations/Migration20240626133555.ts
```diff
@@ -0,0 +1,67 @@
...
+    this.addSql(
+      "update price_rule set attribute = (SELECT rule_attribute FROM rule_type WHERE rule_type.id = price_rule.rule_type_id);"
+    )
+    this.addSql(
+      "update price_list_rule set value = (SELECT array_to_json(ARRAY(SELECT value FROM price_list_rule_value WHERE price_list_rule_value.price_list_rule_id = price_list_rule.id))::jsonb);"
+    )
+    this.addSql(
+      "update price_list_rule set attribute = (SELECT rule_attribute FROM rule_type WHERE rule_type.id = price_list_rule.rule_type_id);"
+    )
+    /* DATA MIGRATION END */
```
riqwan, line 28: this wasn't too bad 😄

packages/modules/pricing/src/models/price-rule.ts
```diff
@@ -28,18 +26,6 @@ const PriceRuleDeletedAtIndex = createPsqlIndexStatementHelper({
...
-  where: "deleted_at IS NULL",
-})
-
-const PriceRuleRuleTypeIdIndex = createPsqlIndexStatementHelper({
-  tableName: tableName,
-  columns: "rule_type_id",
-  where: "deleted_at IS NULL",
-})
-
 const PriceRulePriceIdIndex = createPsqlIndexStatementHelper({
```
riqwan, line 29: this can be dropped now 🤞🏻

packages/modules/pricing/src/migrations/.snapshot-medusa-pricing.json
```diff
@@ -440,26 +561,26 @@
...
           "type": "text",
           "unsigned": false,
           "autoincrement": false,
           "primary": false,
           "nullable": false,
           "mappedType": "text"
         },
-        "default_priority": {
-          "name": "default_priority",
+        "priority": {
```
riqwan, line 582: q: Is this still present?

riqwan (approved):
Generally about revamps, would be nice to shoot off tinier PRs with clear goals. :smile: Its much easier to reason with and review that way. 

Example, removing the unused services and defunct relationships would've cut this PR down a lot and would've been a no-brainer merge.

packages/modules/pricing/src/migrations/Migration20240626133555.ts
sradevski: Not at all, thanks for suggesting we do this 👍

packages/modules/pricing/src/migrations/.snapshot-medusa-pricing.json
sradevski: Yeah it is, I am not sure if it's used though (at least it wasn't in the calculate prices logic). Should I leave it for now or remove it, WDYT?

packages/modules/pricing/src/migrations/.snapshot-medusa-pricing.json
sradevski: Yup, solved by removing the type id from below

packages/modules/pricing/src/migrations/.snapshot-medusa-pricing.json
riqwan: Just took a look, this was meant to act as a specificity override. i.e. if you wanted a particular price to show up instead of the opinionated one that pricing module spits out by default. 

This looks like a missed feature. I'd suggest leaving this for now and create a ticket to see if we want to even tackle this.

olivermrbl (approved):
Nice work, don't have any comments at this point. Let's get this in

PR #7855, opened 2024-06-27
feat(utils): DML generates foreign key indexes by default

riqwan:
**what:**

- adds default indexes on foreign keys when belongs to relationship is present

RESOLVES CORE-2404

### DIFF
--- a/packages/core/types/src/dml/index.ts
+++ b/packages/core/types/src/dml/index.ts
@@ -106,12 +106,12 @@ export interface EntityConstructor<Props> extends Function {
  */
 export type InferForeignKeys<T> = T extends IDmlEntity<infer Schema>
   ? {
-      [K in keyof Schema as Schema[K] extends RelationshipType<any>
-        ? Schema[K]["type"] extends "belongsTo"
+      [K in keyof Schema as Schema[K] extends { type: infer Type }
+        ? Type extends RelationshipTypes
           ? `${K & string}_id`
           : K
-        : K]: Schema[K] extends RelationshipType<infer R>
-        ? Schema[K]["type"] extends "belongsTo"
+        : K]: Schema[K] extends { type: infer Type }
+        ? Type extends RelationshipTypes
           ? string
           : Schema[K]
         : Schema[K]
@@ -174,8 +174,10 @@ export type InferIndexableProperties<T> = keyof (T extends IDmlEntity<
   infer Schema
 >
   ? {
-      [K in keyof Schema as Schema[K] extends RelationshipType<any>
-        ? never
+      [K in keyof Schema as Schema[K] extends { type: infer Type }
+        ? Type extends RelationshipTypes
+          ? never
+          : K
         : K]: string
     } & InferForeignKeys<T>
   : never)
--- a/packages/core/utils/src/dal/mikro-orm/decorators/foreign-key.ts
+++ b/packages/core/utils/src/dal/mikro-orm/decorators/foreign-key.ts
@@ -1,0 +1,10 @@
+import { MetadataStorage } from "@mikro-orm/core"
+
+export function ForeignKey() {
+  return function (target, propertyName) {
+    const meta = MetadataStorage.getMetadataFromDecorator(target.constructor)
+    const prop = meta.properties[propertyName] || {}
+    prop["isForeignKey"] = true
+    meta.properties[prop.name] = prop
+  }
+}
--- a/packages/core/utils/src/dml/__tests__/entity-builder.spec.ts
+++ b/packages/core/utils/src/dml/__tests__/entity-builder.spec.ts
@@ -2482,6 +2482,7 @@ describe("Entity builder", () => {
           reference: "scalar",
           setter: false,
           type: "string",
+          isForeignKey: true,
         },
         created_at: {
           reference: "scalar",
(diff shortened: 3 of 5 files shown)


packages/core/utils/src/dml/entity.ts
```diff
@@ -122,7 +125,16 @@ export class DmlEntity<Schema extends DMLSchema> implements IDmlEntity<Schema> {
     return this
   }
 
-  indexes(indexes: EntityIndex<Schema, string | QueryCondition>[]) {
+  /**
+   * Adds indexes to be created at during model creation on the DML entity.
+   *
+   * Generated Indexes:
```
sradevski:
```suggestion
   * Default Indexes:
```

packages/core/utils/src/dml/entity.ts
```diff
@@ -122,7 +125,16 @@ export class DmlEntity<Schema extends DMLSchema> implements IDmlEntity<Schema> {
...
+   *  - A foreign key indexes will be added if any belongsTo relationships are present.
```
sradevski:
```suggestion
   *  - Foreign key indexes will be added for all defined belongsTo relationships.
```

packages/core/utils/src/dml/helpers/entity-builder/build-indexes.ts
```diff
@@ -42,3 +42,33 @@ function transformWhere(where?: string): string {
...
+*/
+export function buildForeignKeyIndexes<TSchema extends DMLSchema>(
+  schema: TSchema
+) {
+  const indexes: EntityIndex<TSchema, QueryCondition>[] = []
+
+  for (const [field, property] of Object.entries(schema)) {
+    const parsed = property.parse(field)
+    const isRelationshipType = "type" in parsed
+    const isBelongsTo = isRelationshipType && parsed.type === "belongsTo"
```
sradevski: Isn't the `parsed.type === "belongsTo"` sufficient?

packages/core/utils/src/dml/helpers/entity-builder/build-indexes.ts
```diff
@@ -42,3 +42,33 @@ function transformWhere(where?: string): string {
...
+
+    if (!isBelongsTo) {
+      continue
+    }
+
+    // TODO: infer the keyname from schema
```
sradevski: I guess we need to address this before merging

sradevski (commented):
Nice, just some minor comments. waiting for this to land for my notifications PR :)

packages/core/utils/src/dml/helpers/entity-builder/build-indexes.ts
```diff
@@ -42,3 +42,33 @@ function transformWhere(where?: string): string {
...
+  const indexes: EntityIndex<TSchema, QueryCondition>[] = []
+
+  for (const [field, property] of Object.entries(schema)) {
+    const parsed = property.parse(field)
+    const isRelationshipType = "type" in parsed
+    const isBelongsTo = isRelationshipType && parsed.type === "belongsTo"
+
+    if (!isBelongsTo) {
+      continue
+    }
```
adrien2p:
```suggestion
    if (!BelongsTo.isBelongsTo(property)) {
      continue
    }
```

adrien2p: I ve seen multiple places where we are checking the parsed type, but all the relation modifier have a static util to check if it is one of the relation. If some where we need to know if it is a relation of any kind we can add the same static util in the BaseRelation class like this:
```ts
const IsRelationship = Symbol.for("isRelationship")

/**
 * The BaseRelationship encapsulates the repetitive parts of defining
 * a relationship
 */
export abstract class BaseRelationship<T> implements RelationshipType<T> {
  [IsRelationship]: true = true

  static isRelationship<T>(relation: any): relation is BaseRelationship<T> {
    return !!relation?.[IsRelationship]
  }
```

adrien2p (approved):

packages/core/utils/src/dml/helpers/entity-builder/build-indexes.ts
```diff
@@ -42,3 +42,33 @@ function transformWhere(where?: string): string {
 
   return where
 }
+
+/*
+  When a "belongsTo" relationship is found on the DML schema, we create a default index on the
+  foreign key name of the field.
+*/
+export function buildForeignKeyIndexes<TSchema extends DMLSchema>(
```
adrien2p: todoL: I believe the foreign key indexes should be managed when we transform to mikro orm entity, in this area, but I think we can prevent it from happening at the DML level where the foreign key does not yet really exists.

adrien2p: My thinking is that the way foreign keys are managed will depends on the adapter that will be used. In our mikro orm util, when we create the relation we also create the foreign key complement and this is where I would probably see the index being added as we discussed

packages/core/utils/src/dml/helpers/entity-builder/build-indexes.ts
adrien2p: We have a static method on the relationship to handle this check, based on a symbole string we have on the class to be more robust

packages/core/utils/src/dml/helpers/entity-builder/build-indexes.ts
riqwan: good point. Will move it to mikroorm layer.

PR #7856, opened 2024-06-27
feat(dashboard,types,js-sdk,ui): Add missing Price List features

kasperkristensen:
**What**
- Adds missing features to Price List domain
- Adds `StackedFocusModal` and `StackedDrawer` components that should replace SplitView across the project.
- Add Footer to FocusModal
- Adds missing js-sdk functions and types

**Note**
The DatePickers in the PriceLists forms do not work as intended atm. The component is broken, and needs to be fixed. I am working on a fix, but choose to move that work into a separate branch, to prevent this PR from getting bigger then it already is. Will update once the fixes have been merged.

### DIFF
--- a/integration-tests/modules/__tests__/order/order.spec.ts
+++ b/integration-tests/modules/__tests__/order/order.spec.ts
@@ -290,6 +290,14 @@ medusaIntegrationTestRunner({
               discount_tax_total: 0,
               tax_total: 0,
               original_tax_total: 0,
+              refundable_total: 50,
+              refundable_total_per_unit: 50,
+              fulfilled_total: 0,
+              return_dismissed_total: 0,
+              return_received_total: 0,
+              return_requested_total: 0,
+              shipped_total: 0,
+              write_off_total: 0,
               raw_subtotal: {
                 value: "50",
                 precision: 20,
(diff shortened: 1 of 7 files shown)


olivermrbl:
Epic work @kasperkristensen!

Overall LGTM, I just have a few comments:
- The UI becomes unresponsive when closing the "edit configuration side drawer" without saving
- Can we add client-side validation on the date fields or grey out invalid dates in the picker (maybe this is part of what you will fix in a follow-up PR)? Right now, you can enter a surpassed date in "Starts at". You can also enter an "Ends at" date that is before the "Starts at". This will likely need backend validation too, but it would be a nice thing to catch all of these incorrect date states in the form first if possible
CleanShot 2024-06-28 at 10 13 57
CleanShot 2024-06-28 at 10 13 47
- Accoring to the design these colors should be inverted:
CleanShot 2024-06-28 at 10 13 22
  - Actually, I think we should go through all states and ensure the design aligns. The following is also incorrect according to what is in Figma:
CleanShot 2024-06-28 at 10 17 02
- The customer group availability design is different from Figma, but I assume you have confirmed this with Ludvig?
CleanShot 2024-06-28 at 10 14 20
- Nitpicky, but you cannot clear a date, e.g. the "Ends at"

olivermrbl (approved):

kasperkristensen:
@olivermrbl will look into the UI becoming unresponsive. 

The color of the dates are supposed to be grey when the date is now/has passed and orange when it’s in the future. So the implementation is correct, but the design is off in your image.

Everything else related to dates needs the fix to the DatePicker, so will be fixed later. This also includes validation.

Haven’t cleared the new design for customer groups with Ludvig, but its the latest design for how to enter rules / ConditionBlocks, and we need to be consistent with how a rule looks across admin so I have also used that design here 😄

PR #7857, opened 2024-06-27
chore: totals refundable_total

carlos-r-l-rodrigues:
What:
add refundable total on order items

### DIFF
--- a/packages/core/types/src/order/common.ts
+++ b/packages/core/types/src/order/common.ts
@@ -478,12 +478,12 @@ export interface OrderLineItemTotalsDTO {
   discount_tax_total: BigNumberValue
 
   /**
-   * The total refundable total of the order line item.
+   * The refundable total of the order line item.
    */
   refundable_total: BigNumberValue
 
   /**
-   * The total refundable total per unit of the order line item
+   * The refundable total per unit of the order line item.
    */
 
   refundable_total_per_unit: BigNumberValue
@@ -544,12 +544,12 @@ export interface OrderLineItemTotalsDTO {
   raw_discount_tax_total: BigNumberRawValue
 
   /**
-   * The raw total refundable total of the order line item.
+   * The raw refundable total of the order line item..
    */
   raw_refundable_total: BigNumberRawValue
 
   /**
-   * The raw total refundable total per unit of the order line item
+   * The raw  refundable total per unit of the order line item.
    */
   raw_refundable_total_per_unit: BigNumberRawValue
 }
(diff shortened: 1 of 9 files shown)


packages/core/types/src/order/common.ts
```diff
@@ -477,6 +477,17 @@ export interface OrderLineItemTotalsDTO {
    */
   discount_tax_total: BigNumberValue
 
+  /**
+   * The total refundable total of the order line item.
```
fPolic:
```suggestion
  * The refundable total of the order line item.
```

packages/core/types/src/order/common.ts
```diff
@@ -998,6 +1019,11 @@ export interface OrderDTO {
    */
   original_shipping_tax_total: BigNumberValue
 
+  /**
+   * The refundable total of all order items.
+   */
+  item_refundable_total: BigNumberValue
```
fPolic: suggestion: could this be called just `refundable_total`, this is the amount that is currently refundable on the order right?

packages/core/utils/src/totals/line-item/index.ts
```diff
@@ -66,6 +70,40 @@ export function getLineItemsTotals(
...
 }
 
+function setRefundableTotal(
+  item: GetItemTotalInput,
+  discountTotal: BigNumberInput,
+  totals: GetItemTotalOutput,
+  context: GetLineItemsTotalsContext
+) {
+  const totalReturned = MathBN.sum(
+    item.return_requested_quantity ?? 0,
```
fPolic, line 80: question: should we only consider `return_received_quantity` as a returned quantity

packages/core/utils/src/totals/line-item/index.ts
```diff
@@ -66,6 +70,40 @@ export function getLineItemsTotals(
   return itemsTotals
 }
 
+function setRefundableTotal(
+  item: GetItemTotalInput,
+  discountTotal: BigNumberInput,
+  totals: GetItemTotalOutput,
+  context: GetLineItemsTotalsContext
+) {
+  const totalReturned = MathBN.sum(
```
fPolic:
```suggestion
  const totalReturnedQuantity = MathBN.sum(
```

fPolic (approved):
LGTM! Nice 💪

packages/core/utils/src/totals/line-item/index.ts
carlos-r-l-rodrigues: That quantity decreases when the the item is actually received or dismissed.

packages/core/types/src/order/common.ts
carlos-r-l-rodrigues: not really, because the total refundable is solely dependent on the payments. This is the "virtual" refundable total if we consider all the items.
Maybe this is not even relevant to return to be honest. I will remove it

packages/core/utils/src/totals/line-item/index.ts
fPolic: Ok got it!

PR #7859, opened 2024-06-27
feat: refactor module joiner config and links generation

adrien2p:
**What**
This pr add the ability to auto generate all linkable keys and links from the DML schemas. A new util haave been introduced (ExportModule) which takes care of generating the Links exported by the config module as well as auto generating the joiner config if not provided by the service.

Some modification have been applied to the primary keys management, now the PrimaryKey is a modifier insuring that we can create typings that takes it into consideration

FIXES CORE-2407
FIXES CORE-2408
FIXES CORE-2409
FIXES CORE-2450


packages/core/utils/src/dml/properties/id.ts
```diff
@@ -1,28 +1,39 @@
...
     super()
+
     this.dataType = {
       name: "id",
-      options: { primaryKey: true, ...options },
+      options,
     }
   }
 
   primaryKey(decision: boolean) {
```
olivermrbl: question: completely unrelated, but just noticed: what does `decision` do and how is it used?

adrien2p: it might disapear but I believe it is used under the hood to reset the primary key to non primary key but it does not really make sense and instead use the PrimaryKeyModifier, but this pr is not yet ready and will include many modification that are incoming

thetutlage:
I am wondering if we can break this PR into smaller PRs? Because I suspect that all the work I am about to do with DML will create conflicts in this PR and vice-versa.

adrien2p:
cc @shahednasser , maybe we need to adapt the documentation :)

packages/core/utils/src/dml/properties/id.ts
adrien2p: it has been resolved

shahednasser:
@adrien2p can you clarify exactly what should be adapted? Because the changes seem to be under the hood rather than user-facing.

If you mean anything related to joiner config, we actually removed it from the documentation until the linking mechanism is implemented. Once it's ready I'll document it

adrien2p:
> @adrien2p can you clarify exactly what should be adapted? Because the changes seem to be under the hood rather than user-facing.
> 
> If you mean anything related to joiner config, we actually removed it from the documentation until the linking mechanism is implemented. Once it's ready I'll document it

Got it, yes I was thinking about the usage of the `ExportModule` but we can get back to it later

adrien2p:
@olivermrbl if you want to have a look I think it is the perfect moment :)

carlos-r-l-rodrigues (approved):
🔥

packages/core/utils/src/modules-sdk/types/medusa-service.ts
```diff
@@ -0,0 +1,246 @@
...
+  | "listAndCount"
+  | "delete"
+  | "softDelete"
+  | "restore"
+  | "create"
+  | "update"
+
+export type ModelDTOConfig = {
+  dto: object
+  dml?: DmlEntity<any, any>
```
olivermrbl: nitpicky/suggestion: Can we call this `model`? I am afraid of using the `DML` in the public-facing API in case we change it to be a "product" of its own if that makes sense. DML could evolve to be the name of an implementation detail. Wdyt?
```suggestion
  model?: DmlEntity<any, any>
```

carlos-r-l-rodrigues:
> Got it, yes I was thinking about the usage of the `ExportModule` but we can get back to it later

@shahednasser:

```typescript
export default ExportModule(Modules.CURRENCY, {
  service,
  loaders,
})
```

the index.ts from the module will look like this

packages/core/utils/src/modules-sdk/joiner-config-builder.ts
```diff
@@ -29,13 +43,13 @@ export function defineJoinerConfig(
   {
     alias,
     schema,
-    entityQueryingConfig,
+    dmlObjects,
```
olivermrbl: Goes with my other comment. Can we call this `models`?

packages/modules/promotion/src/joiner-config.ts
```diff
@@ -7,7 +7,7 @@ import {
 import { Campaign, Promotion, PromotionRule } from "@models"
 
 export const joinerConfig = defineJoinerConfig(Modules.PROMOTION, {
-  entityQueryingConfig: [Promotion, Campaign, PromotionRule],
+  dmlObjects: [Promotion, Campaign, PromotionRule],
```
olivermrbl: See other comment

packages/modules/pricing/src/joiner-config.ts
```diff
@@ -7,7 +7,7 @@ import {
 import { Price, PriceList, PriceSet } from "@models"
 
 export const joinerConfig = defineJoinerConfig(Modules.PRICING, {
-  entityQueryingConfig: [PriceSet, PriceList, Price],
+  dmlObjects: [PriceSet, PriceList, Price],
```
olivermrbl: See other comment

shahednasser:
@carlos-r-l-rodrigues oh ok got it, once the change is published I'll check it out and update the docs

packages/modules/file/src/joiner-config.ts
```diff
@@ -6,7 +6,7 @@ import {
 } from "@medusajs/utils"
 
 export const joinerConfig = defineJoinerConfig(Modules.FILE, {
-  entityQueryingConfig: [{ name: "File" }],
+  dmlObjects: [{ name: "File" }],
```
olivermrbl: See other comment

packages/core/utils/src/modules-sdk/export-module.ts
```diff
@@ -0,0 +1,44 @@
...
+
+/**
+ * Build the module export and auto generate the joiner config if needed as well as
+ * return a links object based on the DML objects
+ * @param moduleName
+ * @param service
+ * @param loaders
+ * @constructor
+ */
+export function ExportModule<
```
olivermrbl: suggestion: Should we call this `MedusaModule` so it is similar to `MedusaService`? Imo. we don't need the `Export` keyword to be part of the function definition. That is the action you are taking when using it. 
```suggestion
export function MedusaModule<
```

olivermrbl (commented):
Impressive work man. Left a few initial thoughts. Let me know what you think

packages/core/utils/src/modules-sdk/export-module.ts
carlos-r-l-rodrigues: `MedusaModule` will conflict with something that we already have. If this is chosen, we have to update the current one.
But I like the name. or simply `Module`

```typescript
export default Module(Modules.CURRENCY, {
  service,
  loaders,
})
```

olivermrbl: Ah good point, I think `Module` works just as well

packages/core/utils/src/modules-sdk/types/medusa-service.ts
olivermrbl: Maybe to add some context to this: If we choose to productise the DML, it will likely have a more "commercial" name, similar to how Drizzle is named Drizzle, Prisma is named Prisma, etc. This is what I mean when I say the DML term could become purely an implementation detail. 

Therefore, if we can establish a public-facing API, that won't have to change if we choose to do this, I think that would be nice. Less breaking changes to worry about 😄

shahednasser: I agree with this but for a different reason: we already established data models as `model` so it would be easier to understand for developers (if they end up using it)

adrien2p:
@shahednasser I have also removed the need to provide the second argument of the MedusaService, if needed :)D

packages/core/utils/src/modules-sdk/types/medusa-service.ts
adrien2p: I think we can go with that and eventually settle definately on it later, I ll include those changes

shahednasser:
> I have also removed the need to provide the second argument of the MedusaService

@adrien2p that's fine it wasn't documented.

packages/core/utils/src/modules-sdk/types/medusa-service.ts
adrien2p: yes for sure, we might also need to cleanup what is called entity to model, but we can do that in second time, at least we dont add more of them

packages/core/utils/src/modules-sdk/export-module.ts
adrien2p: I ll rename it then 💪

shahednasser:
@adrien2p should the name passed as a parameter to `Module` be the same used in `medusa-config`? For example, if I have a `hello` module, which is the correct way of exporting the definition:

```ts
// option 1
export default Module("hello", {
  service: HelloModuleService,
})

// option 2
export default Module("helloModuleService", {
  service: HelloModuleService,
})
```

Also, is there a reason we require the module's name? I remember we decided before to infer it

adrien2p:
Indeed @shahednasser for custom modules that does not provide any joiner config, we were taking it from the key. I can do a change to the API of the util in order to make it optional, what do you think?

it would look like this
```ts
export default Module({
  service: HelloModuleService,
})
```

and if you want to provide it like we do in the modules then

```ts
export default Module({
  moduleName: '...'
  service: HelloModuleService,
})
```

shahednasser:
@adrien2p I think that would be great!

adrien2p:
> @adrien2p I think that would be great!

let me add that themn

adrien2p:
@shahednasser should be good, lets wait for the pipeline and ill merge this pr

olivermrbl:
nit: Can we do:
```
export default Module({
  name: '...'
  service: HelloModuleService,
})
```

You are already explicitly exporting a module given the function name

adrien2p:
@olivermrbl everything should be good to merge this one, can you get a last round please :)

PR #7862, opened 2024-06-28
Update meilisearch.md

pavlotsyhanok:
Problem:

The existing documentation for the Meilisearch plugin contains example that, if copy-pasted exactly, will not work.

Solution:

I have updated the Meilisearch plugin configuration example to reflect the correct settings. 

Benefits:

With this updated code snippet, users can simply copy and paste it into their local medusa-config.js file. It will work seamlessly with the standard configuration and is compatible with the latest version of Meilisearch.

Testing:

The updated configuration has been tested and confirmed to work in a live deployment environment on Digital Ocean(Server) and Vercel(Storefront)

image

image


www/apps/docs/content/plugins/search/meilisearch.md
```diff
@@ -127,25 +127,32 @@ const plugins = [
...
-              "title", 
-              "description", 
-              "variant_sku", 
-              "thumbnail", 
+              "thumbnail",
               "handle",
             ],
           },
           primaryKey: "id",
+          transformer: (product) => ({
```
shahednasser, line 147: The `transformer` isn't required, did you have to provide it for the plugin to work?

pavlotsyhanok: Hi @shahednasser yes, I had to provide it, otherwise it wouldn't work.

shahednasser (approved):
Thanks @pavlotsyhanok for this contribution! Can I get you to open an issue with more details on why you had to use the `transformer`?

pavlotsyhanok:
> Thanks @pavlotsyhanok for this contribution! Can I get you to open an issue with more details on why you had to use the `transformer`?

Yes, will do.

SalahAdDin:
> Problem:
> 
> The existing documentation for the Meilisearch plugin contains example that, if copy-pasted exactly, will not work.
> 
> Solution:
> 
> I have updated the Meilisearch plugin configuration example to reflect the correct settings.
> 
> Benefits:
> 
> With this updated code snippet, users can simply copy and paste it into their local medusa-config.js file. It will work seamlessly with the standard configuration and is compatible with the latest version of Meilisearch.
> 
> Testing:
> 
> The updated configuration has been tested and confirmed to work in a live deployment environment on Digital Ocean(Server) and Vercel(Storefront)
> image image

which medusa version is that for?

PR #7866, opened 2024-06-28
fix: Disallow creating duplicate prices

sradevski:
We do two things here:
1. Sort by rules count first, amount later
2. Disallow creating equivalent rules

FIXES CORE-2403

Tests depend on https://github.com/medusajs/medusa/pull/7875

### DIFF
--- a/packages/core/utils/src/dal/mikro-orm/mikro-orm-repository.ts
+++ b/packages/core/utils/src/dal/mikro-orm/mikro-orm-repository.ts
@@ -747,16 +747,22 @@ export function mikroOrmBaseRepositoryFactory<T extends object = object>(
       entityName: string,
       data: any
     ): Record<string, any> & { id: string } {
-      const created = manager.create(entityName, data, {
-        managed: false,
-        persist: false,
-      })
+      // We set the id to undefined to make sure the entity isn't fetched from the entity map if it is an update,
+      // giving us incorrect data for the bignumberdata field (I though managed: false and persist: false would already do this)
+      const created = manager.create(
+        entityName,
+        { ...data, id: undefined },
+        {
+          managed: false,
+          persist: false,
+        }
+      )
 
       const resp = {
         // `create` will omit non-existent fields, but we want to pass the data the user provided through so the correct errors get thrown
         ...data,
         ...(created as any).__helper.__bignumberdata,
-        id: (created as any).id,
+        id: data.id ?? (created as any).id,
       }
 
       // Non-persist relation columns should be removed before we do the upsert.
--- a/packages/modules/pricing/integration-tests/__tests__/services/pricing-module/price-list.spec.ts
+++ b/packages/modules/pricing/integration-tests/__tests__/services/pricing-module/price-list.spec.ts
@@ -927,7 +927,6 @@ moduleIntegrationTestRunner<IPricingModuleService>({
                   currency_code: "EUR",
                 }),
               ]),
-              price_list_rules: [],
             })
           )
         })


packages/modules/pricing/src/services/pricing-module.ts
```diff
@@ -965,38 +942,29 @@ export default class PricingModuleService
...
 
-    const priceListMap = new Map(priceLists.map((p) => [p.id, p]))
+    const existingPrices = priceLists
+      .map((p) => p.prices ?? [])
+      .flat() as PricingTypes.PriceDTO[]
+
+    const pricesToCreate = data
+      .map((addPrice) =>
+        this.normalizePrices(
+          addPrice.prices as any,
```
sradevski: The typings can be improve a lot more, but I decided not to do it as part of this PR

packages/modules/pricing/src/services/pricing-module.ts
```diff
@@ -778,29 +773,11 @@ export default class PricingModuleService
...
-              rules_count: numberOfRules,
-              price_rules: Array.from(rulesDataMap.values()),
-            }
-          })
-
-          entry.prices = pricesData
+          entry.prices = this.normalizePrices(
+            priceListData.prices,
+            [],
+            priceListData.prices.price_set_id
```
riqwan: The price_set_id should come from individual prices in a price list as they can be from multiple price sets. This might not be needed here.

packages/modules/pricing/src/services/pricing-module.ts
```diff
@@ -1295,3 +1221,24 @@ export default class PricingModuleService
...
+    max_quantity: price.max_quantity ?? null,
+    ...("price_rules" in price
+      ? price.price_rules?.reduce((agg, pr) => {
+          agg[pr.attribute] = pr.value
+          return agg
+        }, {})
+      : {}),
+  }).sort(([a], [b]) => a.localeCompare(b))
+
+  return crypto.createHash("md5").update(JSON.stringify(data)).digest("hex")
```
carlos-r-l-rodrigues: https://github.com/medusajs/medusa/blob/develop/packages/core/utils/src/common/simple-hash.ts#L2

sradevski: Nice, changed to it now

packages/modules/pricing/src/repositories/pricing.ts
```diff
@@ -163,6 +163,7 @@ export class PricingRepository
         pl_rules_count: "price.pl_rules_count",
         price_list_type: "price.pl_type",
         price_list_id: "price.price_list_id",
+        all_rules_count: knex.raw("price.rules_count + price.pl_rules_count"),
```
carlos-r-l-rodrigues: don't we need a COALESCE here? "pl_rules_count" could be NULL no?

packages/modules/pricing/integration-tests/__tests__/services/pricing-module/calculate-price.spec.ts
```diff
@@ -1134,7 +1134,7 @@ moduleIntegrationTestRunner<IPricingModuleService>({
               {
                 id: "price-set-PLN",
                 is_calculated_price_price_list: true,
-                calculated_amount: 116,
+                calculated_amount: 232,
```
carlos-r-l-rodrigues, line 802: any idea what happened here?

carlos-r-l-rodrigues (approved):

packages/modules/pricing/integration-tests/__tests__/services/pricing-module/calculate-price.spec.ts
sradevski: The way we were sorting the prices was incorrect, see https://medusacorp.slack.com/archives/C05LKT93XNE/p1719482628598669. As I wrote in the PR description, the PR does two things - fixes the sorting (by adding the rule count as the first sorting key), and also enforcing uniqueness on the rules.

packages/modules/pricing/src/repositories/pricing.ts
sradevski: Both rules count are non-nullable and default to 0, so we should be good

riqwan (approved):

packages/modules/pricing/src/services/pricing-module.ts
sradevski: Yeah you are right, let me try to clean this up a bit more 👍

packages/modules/pricing/src/repositories/pricing.ts
carlos-r-l-rodrigues: At a first glance `price.pl_rules_count` apparently comes from `pl_rules_count: "pl.rules_count"`, and that is a left join with price_list.
that why I asked if that can be null.

adrien2p: it seams that the pl_price_count is indeed retrieved through a left join and it means by default that the right side is optional and therefore if you dont have a matching pl tuple the value will not be present which is why carlos is suggesting to use a coalescing (either in the original sub query or before using it, i would suggest to coalesce in the original select to be always defaulted) as it can be null. The only time it cant be null is when the left join as a match and therefore the default value at least will take place or if you were in the case of an inner join because both side requires to be present to be returned.

Feel free to let us know if we are wrong but could you run a little test query (just one select with one left join) where you select a price that does have a matching price list and let us know what the select returns you, i believe the count will be null 🙏

Maybe we are missing some context and in that case feel free to complement.

sradevski: Ah yes indeed, I actually missed the fact that it can be null due to the left join, nice catch guys 👍

riqwan (approved):
lgtm, should we just have an upsert function and remove the others? Or do we still need the others for some use-cases?

sradevski:
@riqwan I think there is some additional cleanup to be done around the interfaces, but I thought it's outside the scope of this PR, and opened a ticket instead: https://linear.app/medusajs/issue/CORE-2419/normalize-the-interface-of-the-pricing-module

PR #7870, opened 2024-06-28
fix(order): set correct change type for exchanges

shahednasser:
Fix the change type to be `EXCHANGE` instead of `CLAIM` when creating an exchange


sradevski (approved):

carlos-r-l-rodrigues (approved):
Thanks

PR #7875, opened 2024-06-29
fix: Delete before upsert in upsertWithReplace to handle relations with unique constraint

sradevski:
If the same unique constraint key was used in the new data, but it wasn't updating the same entry, the "replace" would fail. Since the "replace" is basically the new state, this should not happen, and instead the data that existed before should be deleted first.


olivermrbl (approved):
LGTM

PR #7881, opened 2024-07-01
chore(order): cancel return

carlos-r-l-rodrigues:
PR 1 of 2

What:
- Cancel return/exchange/claim (Module + Workflow)
- Cancel return endpoint

Next PR:
Missing tests for claims and exchanges
Endpoint tests

### DIFF
--- a/packages/core/core-flows/src/order/workflows/create-return.ts
+++ b/packages/core/core-flows/src/order/workflows/create-return.ts
@@ -242,12 +242,8 @@ function prepareReturnShippingOptionQueryVariables({
     return_shipping?: OrderWorkflow.CreateOrderReturnWorkflowInput["return_shipping"]
   }
 }) {
-  if (input.return_shipping) {
-    return
-  }
-
   const variables = {
-    id: input.return_shipping!.option_id,
+    id: input.return_shipping?.option_id,
     calculated_price: {
       context: {
         currency_code: order.currency_code,


packages/modules/order/src/utils/actions/return-item.ts
```diff
@@ -57,15 +57,15 @@ OrderChangeProcessing.registerActionType(ChangeActionType.RETURN_ITEM, {
     }
 
     const quantityAvailable = MathBN.sub(
-      existing!.detail?.shipped_quantity ?? 0,
+      existing!.detail?.fulfilled_quantity ?? 0,
```
olivermrbl, line 60: comment: Not to be addressed now, but I think this will need to change by introducing a `delivered_quantity`. At this point, you shouldn't even be able to request a return, because the items are still at the warehouse. Canceling should suffice. I created a ticket for it for now.

packages/modules/order/src/utils/actions/reinstate-item.ts
```diff
@@ -0,0 +1,64 @@
+import { MathBN, MedusaError, isDefined } from "@medusajs/utils"
+import { ChangeActionType } from "../action-key"
+import { OrderChangeProcessing } from "../calculate-order-change"
+import { setActionReference } from "../set-action-reference"
+
+OrderChangeProcessing.registerActionType(ChangeActionType.REINSTATE_ITEM, {
```
olivermrbl, line 6: question: What is this case for? When are items reinstated?

packages/modules/order/src/utils/actions/return-item.ts
carlos-r-l-rodrigues: I think this validation, if the item was shipped and delivered should be outside the module, in the workflow. As this involves other module.
That is why I just changed that to check the fulfilled_quantity, which is the bare minimum for the module to support

packages/modules/order/src/utils/actions/reinstate-item.ts
carlos-r-l-rodrigues: When an item that was claimed is cancelled/unclaimed.

packages/modules/order/src/utils/transform-order.ts
```diff
@@ -223,3 +224,35 @@ export function mapRepositoryToOrderModel(config, isRelatedEntity = false) {
 
   return conf
 }
+
+function mapRepositoryToRelatedEntity(config) {
```
olivermrbl, line 236: suggestion: I think a small comment in the code about what this is for would be helpful for the next one seeing it. It was not 100% for me at first glance

packages/modules/order/src/services/actions/cancel-return.ts
```diff
@@ -0,0 +1,98 @@
...
+        "reason.id",
+      ],
+      relations: ["items"],
+    },
+    sharedContext
+  )
+
+  const actions: CreateOrderChangeActionDTO[] = []
+
+  returnOrder.items.forEach((item) => {
```
olivermrbl, line 54: question: Do we need to think about canceling shipping?

packages/core/core-flows/src/order/workflows/cancel-return.ts
```diff
@@ -0,0 +1,83 @@
...
+
+    throwErrorIf(
+      orderReturn_.items!,
+      hasReceived,
+      "Can't cancel a return which has returned items"
+    )
+  }
+)
+
+export const cancelReturnWorkflowId = "cancel-return-return"
```
olivermrbl:
```suggestion
export const cancelReturnWorkflowId = "cancel-return"
```

olivermrbl (approved):

packages/modules/order/src/services/actions/cancel-return.ts
carlos-r-l-rodrigues: yes, absolutely.
Good point.

packages/modules/order/src/utils/transform-order.ts
carlos-r-l-rodrigues: Will add some comments on this entire file and explain these functions.

olivermrbl: Awesome

PR #7886, opened 2024-07-01
fix: Return null for calculated price if no prices are defined

sradevski:
FIXES CORE-2418

### DIFF
--- a/packages/modules/pricing/src/services/pricing-module.ts
+++ b/packages/modules/pricing/src/services/pricing-module.ts
@@ -144,21 +144,20 @@ export default class PricingModuleService
       return priceSets
     }
 
-    const priceSetIds: string[] = []
-    const priceSetMap = new Map()
-    for (const priceSet of priceSets) {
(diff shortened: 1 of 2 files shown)


packages/modules/pricing/src/services/pricing-module.ts
```diff
@@ -182,21 +181,20 @@ export default class PricingModuleService
...
     )
+
+    const calculatedPricesMap = new Map()
     for (const calculatedPrice of calculatedPrices) {
-      const priceSet = priceSetMap.get(calculatedPrice.id)
-      priceSet.calculated_price = calculatedPrice
+      calculatedPricesMap.set(calculatedPrice.id, calculatedPrice)
+    }
+
+    for (const priceSet of priceSets) {
```
sradevski, line 195: We swapped to iterating over the price set so we can set the calculated price explicitly to null (and not undefined).

carlos-r-l-rodrigues (approved):

riqwan (approved):
lgtm!

PR #7889, opened 2024-07-01
fix: Only list price set prices when calling the price set methods

sradevski:
FIXES CORE-2417

### DIFF
--- a/packages/modules/pricing/src/services/pricing-module.ts
+++ b/packages/modules/pricing/src/services/pricing-module.ts
@@ -1328,10 +1328,10 @@ export default class PricingModuleService
     config: FindConfig<PricingTypes.PriceSetDTO> | undefined
   ) {
     return {
-      ...config,
       options: {
         populateWhere: { prices: { price_list_id: null } },
       },
+      ...config,
     }
   }
 }


olivermrbl (approved):
LGTM

packages/modules/pricing/src/services/pricing-module.ts
```diff
@@ -1294,4 +1321,15 @@ export default class PricingModuleService
...
   }
+
+  protected normalizePriceSetConfig(
+    config: FindConfig<PricingTypes.PriceSetDTO> | undefined
+  ) {
+    return {
+      ...config,
+      options: {
+        populateWhere: { prices: { price_list_id: null } },
+      },
```
riqwan, line 1333: q: as a permanent filter? Shouldn't this rather be coming from the http layer for variants?

Reason being that price list prices are also part of the price set.

sradevski: my thought process was that price list prices are owned by the price list, and as such, you wouldn't access them directly through the price set, so I thought of putting it here. However, if we want to keep the module `pure` without such implications we can move it to the http layer, WDYT?

olivermrbl: Are they not always fetched through `listPriceLists` or `retrievePriceList` with a `prices` expansion?

If not, then this is a good point and something to address.

olivermrbl: But in any case, maybe this makes most sense to add at the HTTP layer since having the filter is a somewhat opinionated decision.

sradevski: Ok let's move it to the HTTP layer, I am fine with that

sradevski: Now another problem arises - we currently don't support passing arbitrary options down to the module through remoteQuery. Even if we did, typically all config applies to the entrypoint, not to the expanded relationships.

I think anything that would allow us to do this properly will require quite a bit of rework (I am thinking being able to configure each expanded relationship and the filters applied to it), since we actually never fetch priceSets directly, it's always an expansion of variants, shipping options, etc. Any better suggestions to support this in the HTTP layer?

olivermrbl: In that case, my take would be that it's a fair assumption to say: by default **price set** `list` + `retrieve` exclude **price list** prices, and to fetch **price list** prices you would need to use the dedicated **price list** `list` + `retrieve` methods.

Additionally, we could allow this option to be overwritten if you are consuming the module directly:
```
options: {
  populateWhere: { prices: { price_list_id: null } },
  ...config.options ?? {}
},
```

The solution could be considered temporary until we figure out how to deal with this at the HTTP level. 

Wdyt?

riqwan: They are accessed through price lists, because they have a direct relationship to prices, but so do the price set. 

If we want to make a clear delineation here from a module perspective that price sets will always return non price list prices, i'm not opposed to it. Just felt a bit weird seeing that we're scoping it by default. 😄

sradevski: I think we are quite limited technically now to do anything better, so let me do what Oli suggested (made the options overwritable), and I'll open a ticket to look into this as a use-case that we need to think how to support with `remoteQuery`.

sradevski: Created CORE-2440

riqwan (approved):

PR #7891, opened 2024-07-01
fix(ui,dashboard): Revamp DatePicker component

kasperkristensen:
**What**
- Revamps the DatePicker component. 
- Addresses all issues with broken DatePickers across admin.

**Note**
- Part of this PR is adding a I18nProvider which is used to set the locale that is used for our DatePicker and Calendar components. Per default they use the browser locale. In the current implementation, we are grabbing the locale to use from the language that is picked in the "Profile" section. This means that currently the only possible locale is "en-US", meaning times uses AM/PM. This is likely not what we want, but we need to make a decision on how we want to handle this globally, will create a ticket for it and we can then clean it up later on.
- This PR does not include "presets" or a DateRange picker that were part of the old implementation. Will open tickets to re-add this later on, but since we aren't using it in admin any where it makes sense to address later.
- This PR also bumps and pin every `@radix-ui` dependency in `@medusajs/ui` and `@medusajs/dashboard`. Our different versions were pulling in multiple versions of internal radix dependencies which were breaking Popover and Dialog behaviour across admin. One thing to note is that Radix have started to print warnings for missing Descriptions and Titles in dialogs. We should add these as we go, for better accessibility. Its not an urgent task but something we can add as we clean up admin over the following weeks. 

CLOSES CORE-2382


packages/design-system/ui/src/components/date-picker/date-picker-field.tsx
```diff
@@ -0,0 +1,56 @@
...
+  },
+  defaultVariants: {
+    size: "base",
+  },
+})
+
+const DatePickerField = ({ size = "base", ...props }: DatePickerFieldProps) => {
+  const { locale } = useLocale()
+
+  console.log("locale: ", locale) // locale:  en-US
```
olivermrbl:
```suggestion
```

olivermrbl (approved):
I didn't look at the code too much, because there was a lot going on. But I tested it in the different domains where it has replaced the old date picker and everything worked as expected. Nice work @kasperkristensen

PR #7892, opened 2024-07-01
fix(promotion): tabs + resetting rules logic

riqwan:
what:

- fixes resetting rules logic
- make limit and budget nullable
- name tabs correctly
- add FE validation for spend + tooltip

### DIFF
--- a/packages/modules/promotion/src/models/promotion.ts
+++ b/packages/modules/promotion/src/models/promotion.ts
@@ -51,7 +51,6 @@ export default class Promotion {
     fieldName: "campaign_id",
     nullable: true,
     mapToPk: true,
-    onDelete: "set null",
   })
   campaign_id: string | null = null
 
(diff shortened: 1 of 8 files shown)


packages/modules/promotion/src/models/promotion.ts
```diff
@@ -51,7 +51,6 @@ export default class Promotion {
     fieldName: "campaign_id",
     nullable: true,
     mapToPk: true,
-    onDelete: "set null",
```
sradevski, line 54: Was this the culprit?

sradevski (approved):

packages/modules/promotion/src/models/promotion.ts
riqwan: no, this just wasn't needed. The issue is still a mystery. 😄

PR #7896, opened 2024-07-01
fix(utils): joiner config from medusa project

carlos-r-l-rodrigues:


packages/core/utils/src/modules-sdk/migration-scripts/migration-down.ts
```diff
@@ -58,7 +58,7 @@ export function buildRevertMigrationScript({
       logger?.error(
         `${upperCaseFirst(
           moduleName
-        )} module migration failed to run - Error: ${error}`
+        )} module migration failed to run - Error: ${error.errros ?? error}`
```
olivermrbl, line 61: question: Is this a typo or is it really`errros`? 😄

olivermrbl (approved):

packages/core/utils/src/modules-sdk/migration-scripts/migration-down.ts
carlos-r-l-rodrigues: This is returning an `AggregateError` for some errors. I suspect there is a Promise.reject + new Error somewhere (not in our packages), like connection refused.
So this should solve and display the message at least.

adrien2p (approved):

PR #7900, opened 2024-07-02
Mark keys as primary with explicit method call

thetutlage:
FIXES: CORE-2438

This PR removes the implicit primary key flag from the `id` property of the DML entity and we 
must use `.primaryKey` modifier to mark any property as primary.

As a result: There is only one way to define primary keys. No implicit magic.


packages/core/utils/src/dml/__tests__/entity-builder.spec.ts
```diff
@@ -1554,162 +1553,25 @@ describe("Entity builder", () => {
...
-          primary: true,
-          reference: "scalar",
-          type: "string",
-        },
-        account_id: {
-          columnType: "integer",
-          name: "account_id",
-          nullable: false,
-          primary: true,
-          reference: "scalar",
```
thetutlage, line 1598: I have to delete a lot of code from this test, because we are re-creating/registering the same entity with Mikro ORM (via "User" as the model name).

As a result, Mikro ORM is performing all sorts of meta-data merging under the hood and making it super hard to reason about the assertions. 

So either we break this test into multiple tests, or we should create unique entities in this test.

packages/core/utils/src/dml/helpers/entity-builder/infer-primary-key-properties.ts
```diff
@@ -1,27 +1,9 @@
...
-
-  Model 3:
-    id: model.id()
-    code: model.text().primaryKey() -> composite primary key
-    name: model.text().primaryKey() -> composite primary key
-*/
+/**
+ * The method is used to infer primary keys from the schema
+ */
 export function inferPrimaryKeyProperties<TSchema extends DMLSchema>(
```
thetutlage: I think, we don't need this method anymore. Just still want to confirm with @adrien2p once.

packages/modules/notification/src/models/notification.ts
```diff
@@ -3,7 +3,7 @@ import { NotificationProvider } from "./notification-provider"
 
 // We probably want to have a TTL for each entry, so we don't bloat the DB (and also for GDPR reasons if TTL < 30 days).
 export const Notification = model.define("notification", {
-  id: model.id({ prefix: "noti" }),
+  id: model.id().primaryKey("noti"),
```
sradevski: suggestion: I would have expected the `prefix` to go on the `id` property, not the `primaryKey`. I think it's a bit strange to couple the concept of a primary key with a prefix, WDYT?

sradevski (commented):
Nice cleanup, just one comment on how the prefix is specified

packages/modules/notification/src/models/notification.ts
thetutlage: Ohhh yeah. I thought the prefix is for the index of primary key. My bad. Let me fix it

sradevski (approved):

PR #7903, opened 2024-07-02
Add support for native enums

thetutlage:
FIXES: CORE-2413

This PR adds support for passing native enums to the `model.enum` method.


riqwan (approved):

packages/core/utils/src/dml/__tests__/enum-schema.spec.ts
```diff
@@ -46,4 +46,52 @@ describe("Enum property", () => {
...
+
+    const property = new EnumProperty(Roles)
+    expectTypeOf(property["$dataType"]).toEqualTypeOf<Roles>()
+
+    expect(property.parse("role")).toEqual({
+      fieldName: "role",
+      dataType: {
+        name: "enum",
+        options: {
+          choices: ["ADMIN", "WRITER", "EDITOR", 0, 1, 2],
```
olivermrbl, line 65: question: Do we handle both types of `choices` formats automagically? They seem very different in shape.

olivermrbl (approved):
LGTM

packages/core/utils/src/dml/__tests__/enum-schema.spec.ts
thetutlage: Its a wild usage of enum. I wanted to make sure that we support it properly. But, in reality one will define enums as `key-value` pair.

PR #7912, opened 2024-07-03
Feat(workflows-sdk): step conditional

carlos-r-l-rodrigues:
What:
 - New `when` util to execute a step if a condition is true

```typescript
when({ input, ret }, ({ input, ret }) => {
  return input.runStep && ret.someResponse === "yes"
})
.then(() => step_2({ hey: "hello" }))
```

If the condition is false the step will return undefined and continue the flow.
All steps defined in the `.then` function will only be executed if the evaluated condition is true.

Notes:
* TS Types will be improved
* A bigger refactor allowing properly skipping steps at the transaction orchestrator level will be implemented in another opportunity. That will enable the UI identify steps that were planned to be skipped as part of the execution.


olivermrbl (commented):
Really like this @carlos-r-l-rodrigues, nice work.

I think this API is good. Just sharing a few thoughts, nothing actionable: 

Would it make sense (if possible) to have the conditional `if` helper live by itself instead of chained on a step? Similar to our `transformer`. 

I imagine it would look something like this:
```
const result = if(data, ({ status }) => {
  return status === "auto_capture"
})
.then(() => {
  const captured = capturePayment({ ... })

  anotherStep({ ... })

  return captured
})
```

The suggestion spawned from two thoughts:
- You might want to have more than one step executed if a condition holds true
  - I guess the alternative with the current API would be to create a sub-workflow and then chain `.if` on that
- The API would be closer to a regular if-then control flow

Again, I think the proposed API is already good, so primarily sharing this to hear your thoughts on it. Also, bear in mind, I am not sure my suggestion is technically feasible.

sradevski (approved):
Really nice stuff, this should clean up workflows by a lot!

On the API itself, if possible, I agree with @olivermrbl on having the `if` similar to the transform, as it feels a bit more consistent.

It could even be `if(condition, handler)`, again, similar to the `transform(input, handler)` syntax. Leaving it up to you though

carlos-r-l-rodrigues:
Using a regular if at the moment is not possible because that would require the transaction orchestrator to be mutated during it's execution. As at this point we are just composing the steps of the workflow, we do not have access to any value to compare them in a condition.

Creating it as another function is possible, but not using the reserved keyword `if`. And we also need the inputs as arguments, so I'm not sure if that would improve the readability.

something like this:
```typescript
when({ input, ret }, ({ input, ret }) => {
  return input.runStep && ret.someResponse === "yes"
}, step_2({ hey: "hello" }))
```

sradevski:
@carlos-r-l-rodrigues ah yes, of course 😄 I think `when` or `iff` (I've seen this somewhere) would both work, but I do prefer the `when` when it comes to that discussion. Good point on having to map to actual values, I can see how it can get worse.

I think we can go with what you suggested, and maybe we can revisit later on again, WDYT?

carlos-r-l-rodrigues:
This was the closest I could get considering what we have.

```typescript
 const workflow = createWorkflow("wf",
  function (input: { callSubFlow: boolean }) {
    
    const subWorkflowRes = when({ input }, ({ input }) => {
      return input.callSubFlow
    }).then(() => {
      return subWorkflow.runAsStep({
        input: "hi from outside",
      })
    })

    return step3(subWorkflowRes.result)
  }
)
```
or

```typescript
 const workflow = createWorkflow("wf",
  function (input: { callSubFlow: boolean }) {
    
    when({ input }, ({ input }) => {
      return input.callSubFlow
    })

    const subWorkflowRes = subWorkflow.runAsStep({
      input: "hi from outside",
    })

    endWhen()

    return step3(subWorkflowRes.result)
  }
)

olivermrbl:
I am in favor of #1 here

carlos-r-l-rodrigues:
> I am in favor of #1 here

the `when(input, condition).then(() => { /* steps */})` ?

in any case, internally I will keep the `endWhen` in case someone prefers that way, because if the .then is not called after the when() everything executed after it will be conditional.
and the `.if()` we can decide if we keep it public, or just internal.

olivermrbl:
> the when(input, condition).then(() => { /* steps */}) ?

Yeah, this one. Think we should keep the others private/hidden for now 👍

olivermrbl:
Maybe I would actually argue we should remove the others. I can't see when we would use them, so they would just bloat the codebase no?

Update: Do you have a specific case in mind where these would be preferred over the above one?

carlos-r-l-rodrigues:
> Maybe I would actually argue we should remove the others. I can't see when we would use them, so they would just bloat the codebase no?

I need the ".if" internally because of its scope. But I can just make it private and not expose it.
I will validate if the `.then` is called after the `when` to avoid mistakes.

carlos-r-l-rodrigues:
@olivermrbl added the `when`

olivermrbl:
> I need the ".if" internally because of its scope. But I can just make it private and not expose it.

Ah got it, sounds good then 👍

olivermrbl (approved):
Nice, LGTM!

PR #7915, opened 2024-07-03
fix(dashboard): breaking store page due to default currency

riqwan:
what:

- Remote query sets `take: null` for expand cases to prevent the default limit of 15 from being applied from the module
- safely traverses currency

### DIFF
--- a/packages/admin-next/dashboard/src/routes/store/store-detail/components/store-general-section/store-general-section.tsx
+++ b/packages/admin-next/dashboard/src/routes/store/store-detail/components/store-general-section/store-general-section.tsx
@@ -56,7 +56,7 @@ export const StoreGeneralSection = ({ store }: StoreGeneralSectionProps) => {
         {defaultCurrency ? (
           <div className="flex items-center gap-x-2">
             <Badge size="2xsmall">
-              {defaultCurrency.currency?.code?.toUpperCase()}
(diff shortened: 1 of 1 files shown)


packages/core/modules-sdk/src/remote-query.ts
```diff
@@ -222,6 +223,10 @@ export class RemoteQuery {
       )
     }
 
+    if (ids?.length && !hasPagination) {
+      options.take = null
+    }
```
riqwan, line 228: @carlos-r-l-rodrigues I've added this here as a way to retrieve all the expand relationships to be fetched. 

In this case, store was returning 50 sub resources, but remote query was only fetching 15 which was set at a module level. 

Let me know if there is a better way to approach this fix.

sradevski (commented):
LGTM, but I'll let Carlos approve

packages/core/modules-sdk/src/remote-query.ts
carlos-r-l-rodrigues: When querying by primary keys, the modules should remove this restriction anyway.
But that's ok to have it here.

carlos-r-l-rodrigues: Just double check if that should be a pk on the module you are querying.

carlos-r-l-rodrigues (approved):

PR #7919, opened 2024-07-03
Update stripe payment provider options

sradevski:
FIXES CORE-2283


packages/modules/payment/src/loaders/index.ts
```diff
@@ -1,2 +1 @@
 export * from "./providers"
-export * from "./defaults"
```
sradevski, line 2: Moved the code inside `providers` since it didn't load defaults, but instead it just wrote to the DB. This is more in-line with other loaders.

packages/modules/payment/src/loaders/providers.ts
```diff
@@ -37,4 +43,32 @@ export default async ({
...
+  const providersToCreate: CreatePaymentProviderDTO[] = []
+  for (const id of providersToLoad) {
+    if (loadedProvidersMap.has(id)) {
+      continue
+    }
+
+    providersToCreate.push({ id })
+  }
+
+  await paymentProviderService.create(providersToCreate)
```
riqwan, line 73: you can do an upsert with replace here to remove any providers that no longer exists.

riqwan (approved):

packages/modules/payment/src/loaders/providers.ts
sradevski: I also thought of it, but when I tried it it caused some issues in the platform (eg. if you've selected a default payment provider for the region, and then you remove it), so removing a payment provider will likely have to cascade to other modules as well, so it wasn't the right time :)

PR #7920, opened 2024-07-03
fix: only showcase tooltip when currency is not set

riqwan:


riqwan:
@shahednasser this one should fix the tooltip visibility

shahednasser (approved):
LGTM

PR #7923, opened 2024-07-03
feat: Make the Module require the service name

adrien2p:
**What**
The `Module` util will now require the service name. As part of the work, the Links returned by the `Module` will have a proper type inference of the service name as well as the field the end data will be attached to.

Screenshot 2024-07-03 at 14 29 08
Screenshot 2024-07-03 at 14 37 12

**NOTE**
This pr is preliminary step towards the `defineLink` util

FIXES CORE-2463


adrien2p:
@shahednasser after discussing w @olivermrbl we decided that the service name would become required when using the `Module` util as it is a required value to possess in order to generate the link between modules

shahednasser:
@adrien2p sounds good, will update it in the docs

sradevski (approved):
LGTM!

riqwan (approved):
🥇

PR #7924, opened 2024-07-03
chore: cleanup migrations logs

riqwan:
what:

when running `medusa run migrations`, cleanup logs with short descriptive message.

Before:

Screenshot 2024-07-03 at 14 31 36

After:

Screenshot 2024-07-03 at 14 42 57

### DIFF
--- a/packages/core/utils/src/modules-sdk/migration-scripts/migration-up.ts
+++ b/packages/core/utils/src/modules-sdk/migration-scripts/migration-up.ts
@@ -62,10 +62,6 @@ export function buildMigrationScript({ moduleName, models, pathToMigrations }) {
             pendingMigrations.length
           } migration files executed`
         )
-      } else {
-        logger.info(
-          `${upperCaseFirst(moduleName)} module: No pending migrations`
-        )
       }
     } catch (error) {
       logger.error(
--- a/packages/modules/link-modules/src/migration/index.ts
+++ b/packages/modules/link-modules/src/migration/index.ts
@@ -66,9 +66,10 @@ export function getMigration(
       } else {
         logger.info(`Skipping "${tableName}" migration.`)
       }*/
-      logger.info(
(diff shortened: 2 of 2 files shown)


shahednasser (commented):
Love this! What do you think of only showing the available migrations, and if there are none we just say "there are no pending migrations"? That way if I want to check if my module's migration was executed, I don't have to look for it between pending migration messages.

srindom:
On top of what Shahed says, I also find the link module warnings a bit odd. Does anyone have suggestions for what we can do there?

note: this is not blocking for the PR just a thought

riqwan:
@shahednasser that was my initial goal, but because these migrations are run independently on each module, its hard to make that work without an aggregation layer. Without much required changes the only thing we can do is hide the logs entirely if there aren't any migrations without the "there are no pending migrations" message at the end

shahednasser (approved):
@riqwan got it, maybe this is something to resolve in a separate PR or could be resolve when we work on the migration CLI tool.

For now, this LGTM

riqwan:
With the new cli, there are some plans to listen to mikroorm events directly, that can definitely help structure a much nicer response. 🤞🏻

riqwan:
@carlos-r-l-rodrigues any thoughts on the link module message here?

olivermrbl:
> Does anyone have suggestions for what we can do there?

Don't log anything, if the table already exists. Same for module migrations, as suggested by Shahed.

To me, the logs don't provide any useful information at this point, so we might as well remove them. You only need them if an action is taken or will need to be taken.

riqwan:
This?

Screenshot 2024-07-03 at 15 27 46

olivermrbl:
Yeah, I prefer this when there are no running migrations. Curious to hear what you and others think?

sradevski (approved):
Simple and nice 👍

riqwan:
sgtm, I think the aggregate message will be helpful, but since this is only temporary, lets go with this 🤞🏻

PR #7930, opened 2024-07-03
feat: Flatten the provider config for all modules

sradevski:
FIXES CORE-2460


olivermrbl (approved):
Nice – seeing this, I am convinced this was a good decision. Cleaned everything up and made it more readable. Thx for pushing it through

PR #7931, opened 2024-07-03
feat: define link util

adrien2p:
**What**
Improve the `defineLink` util in order to use our new `Linkable` concept

**NOTE**
Not all modules are exporting there linkables, only the `region` and `currency` for now as they are the only one migrated to DML

FIXES CORE-2447

### DIFF
--- a/packages/admin-next/dashboard/src/routes/store/store-detail/components/store-general-section/store-general-section.tsx
+++ b/packages/admin-next/dashboard/src/routes/store/store-detail/components/store-general-section/store-general-section.tsx
@@ -1,18 +1,18 @@
 import { PencilSquare } from "@medusajs/icons"
+import { AdminStore } from "@medusajs/types"
 import { Badge, Container, Heading, Text } from "@medusajs/ui"
 import { useTranslation } from "react-i18next"
 import { ActionMenu } from "../../../../../components/common/action-menu"
 import { useRegion } from "../../../../../hooks/api/regions"
-import { ExtendedStoreDTO } from "../../../../../types/api-responses"
 
 type StoreGeneralSectionProps = {
-  store: ExtendedStoreDTO
+  store: AdminStore
 }
 
 export const StoreGeneralSection = ({ store }: StoreGeneralSectionProps) => {
   const { t } = useTranslation()
 
-  const { region } = useRegion(store.default_region_id, undefined, {
+  const { region } = useRegion(store.default_region_id!, undefined, {
     enabled: !!store.default_region_id,
   })
 
@@ -56,10 +56,10 @@ export const StoreGeneralSection = ({ store }: StoreGeneralSectionProps) => {
         {defaultCurrency ? (
           <div className="flex items-center gap-x-2">
             <Badge size="2xsmall">
-              {defaultCurrency.currency_code.toUpperCase()}
+              {defaultCurrency.currency_code?.toUpperCase()}
             </Badge>
             <Text size="small" leading="compact">
-              {defaultCurrency.currency.name}
+              {defaultCurrency.currency?.name}
             </Text>
           </div>
         ) : (
--- a/packages/core/modules-sdk/src/remote-query.ts
+++ b/packages/core/modules-sdk/src/remote-query.ts
@@ -87,6 +87,7 @@ export class RemoteQuery {
     select?: string[]
     relations: string[]
     args: JoinerArgument
+    take?: number | null
   } {
     expand = JSON.parse(JSON.stringify(expand))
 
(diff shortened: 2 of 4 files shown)


integration-tests/modules/__tests__/link-modules/define-link.spec.ts
```diff
@@ -0,0 +1,92 @@
...
+
+        const linkDefinition = MedusaModule.getCustomLinks()
+          .map((linkDefinition: any) => {
+            const definition = linkDefinition(MedusaModule.getLoadedModules())
+            return definition.serviceName === link.serviceName && definition
+          })
+          .filter(Boolean)[0]
+
+        expect(link.serviceName).toEqual("currencyCurrencyRegionRegionLink")
+        expect(linkDefinition).toEqual({
```
olivermrbl, line 34: thought: `database` could probably also be autogenerated no?
```
databaseConfig: {
    tableName: [left_source_right_source],
    idPrefix: [first three characters of each source],
  },
```
For example:
```
databaseConfig: {
    tableName: "order_payment_collection",
    idPrefix: "ordpay",
  },
```

adrien2p: everything is autogenerated here, what do you mean? the link module is already providing a default config for it, it is not specified because we do not need to override it here

olivermrbl: Ah sorry, if we are already autogenerating this, forget what I said 😄

packages/core/utils/src/modules-sdk/define-link.ts
```diff
@@ -0,0 +1,314 @@
...
+  linkable: string
+  primaryKey: string
+}
+
+type InputToJson = {
+  toJSON(): InputSource
+}
+
+type InputOptions = {
+  source: InputToJson | InputSource
```
olivermrbl: very nit: Think we should try to come up with a name closer to links, modules, and models. Perhaps, we could do `link` or `model`?

```ts
defineLink({ link: ProductModule.links.product, ...
defineLink({ model: ProductModule.links.product, ...
```

integration-tests/modules/__tests__/link-modules/define-link.spec.ts
```diff
@@ -0,0 +1,92 @@
...
+      let remoteQuery
+
+      beforeAll(async () => {
+        appContainer = getContainer()
+        remoteQuery = appContainer.resolve("remoteQuery")
+      })
+
+      it("should generate a proper link definition", async () => {
+        const currencyLinks = CurrencyModule.links
+        const regionLinks = RegionModule.links as any // TODO fix typings for model.define with config object
```
olivermrbl: comment: Separately, @shahednasser highlighted that `links` could be understood as existing links instead of "something that can be linked", I think that was a good point. Maybe we should also consider renaming this to `linkConfig` or `linkable` or something third.

What do you think?

olivermrbl (commented):
Epic work, will start playing around with this!

integration-tests/modules/__tests__/link-modules/define-link.spec.ts
adrien2p: linker? wdyt?

packages/core/utils/src/modules-sdk/define-link.ts
```diff
@@ -0,0 +1,316 @@
...
+
+    serviceAObj = {
+      key: source.linkable,
+      alias: source.field,
+      primaryKey: source.primaryKey,
+      isList: leftService.isList ?? false,
+      module: source.serviceName,
+    }
+  } else {
+    throw new Error("Invalid value for serviceA config")
```
olivermrbl: Could we do something like this in the error messages?
```suggestion
    throw new Error("Invalid link configuration for ${module}")
```

packages/core/utils/src/modules-sdk/define-link.ts
```diff
@@ -0,0 +1,316 @@
...
+
+    serviceBObj = {
+      key: source.linkable,
+      alias: source.field,
+      primaryKey: source.primaryKey,
+      isList: rightService.isList ?? false,
+      module: source.serviceName,
+    }
+  } else {
+    throw new Error("Invalid value for serviceA config")
```
olivermrbl: Same here

olivermrbl (approved):
LGTM, should we get this in so everyone can start playing around with it?

adrien2p:
Yes definitely, let me just do the changes you commented about and then we can merge 👌

packages/core/utils/src/modules-sdk/define-link.ts
adrien2p: if it is neither of the previous check, then we dont have the module name, should I say first args and second for the other onw

PR #7935, opened 2024-07-03
feat(dashboard,types,js-sdk,ui): Tax Regions UI

kasperkristensen:


olivermrbl:
@kasperkristensen, how is this PR best reviewed? I think going through all files will be a huge task. Maybe there are some areas needing more attention than others.

kasperkristensen:
> @kasperkristensen, how is this PR best reviewed? I think going through all files will be a huge task. Maybe there are some areas needing more attention than others.

I think the critical parts are the ones related to Rules, so the `tax-region-tax-override-*` files, and the `target-*` files in `/common`, since it contains all the hoops and tricks I've had to do to work with tax rate rules.

Other then that, the easiest way to review is properly just to go through each flow in the dashboard, and test it out.

olivermrbl:
Can't delete tax region overrides:

https://github.com/medusajs/medusa/assets/59018053/a3c9638b-44ab-4fd0-916a-30fa3e409112

olivermrbl:
It seems we are missing a translation entry:
CleanShot 2024-07-09 at 18 49 13

olivermrbl (approved):
Epic work, only spotted a few minor things while testing.

PR #7937, opened 2024-07-04
Prettify DML properties when inferred

thetutlage:
FIXES: CORE-2470

Tried a few different things and turns out a one liner helper did the trick 😇

Here are some results.

## Simple entity

Screenshot 2024-07-04 at 10 38 14 AM

## Entity with circular relationship

Screenshot 2024-07-04 at 10 37 47 AM

## Entity with many to many circular relationship

Screenshot 2024-07-04 at 10 37 15 AM


olivermrbl (approved):
This is nice, LGTM. Will let @adrien2p approve :)

adrien2p (approved):
I would suggest to move the prettify to the common types. I can see multiple places wheee it would be useful 🙏

PR #7943, opened 2024-07-04
feat: Add support for managing tax inclusivity

sradevski:
UI / HTTP / Workflows will come in separate PRs

REF CORE-2376

### DIFF
--- a/packages/modules/pricing/src/services/pricing-module.ts
+++ b/packages/modules/pricing/src/services/pricing-module.ts
@@ -256,10 +256,46 @@ export default class PricingModuleService
       pricingContext,
       sharedContext
     )
-    const pricesSetPricesMap = groupBy(
-      results,
-      "price_set_id"
-    ) as unknown as Map<string, PricingTypes.CalculatedPriceSetDTO[]>
+
+    const pricesSetPricesMap = groupBy(results, "price_set_id")
+    const priceIds: string[] = []
+    pricesSetPricesMap.forEach(
+      (prices: PricingTypes.CalculatedPriceSetDTO[], key) => {
+        const priceListPrice = prices.find((p) => p.price_list_id)
+        const defaultPrice = prices?.find((p) => !p.price_list_id)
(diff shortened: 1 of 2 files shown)


packages/modules/pricing/src/services/pricing-module.ts
```diff
@@ -238,67 +255,115 @@ export default class PricingModuleService
...
-
-            original_price: {
-              id: originalPrice?.id || null,
-              price_list_id: originalPrice?.price_list_id || null,
-              price_list_type: originalPrice?.price_list_type || null,
-              min_quantity: parseInt(originalPrice?.min_quantity || "") || null,
-              max_quantity: parseInt(originalPrice?.max_quantity || "") || null,
-            },
+            // We use the price rules to get the right preferences for the price
+            const priceRulesForPrices = await this.priceRuleService_.list(
```
carlos-r-l-rodrigues: Can we bring this .list outside of the loop to avoid multiple db calls?

sradevski: Yes let's do it, it will require some more cleanup so I thought I leave it for now, but I think we should

carlos-r-l-rodrigues (approved):

packages/modules/pricing/src/services/pricing-module.ts
```diff
@@ -641,6 +711,102 @@ export default class PricingModuleService
...
+      | PricingTypes.CreatePricePreferenceDTO[],
+    @MedusaContext() sharedContext: Context = {}
+  ): Promise<PricePreferenceDTO | PricePreferenceDTO[]> {
+    const preferences = await this.pricePreferenceService_.create(
+      data,
+      sharedContext
+    )
+
+    return await this.baseRepository_.serialize<any[]>(preferences)
+  }
```
riqwan, line 738: won't all these methods come from the base service?

sradevski: Ah you are actually right, I actually completely missed that these come implemented by default. I'll clean it up

PR #7944, opened 2024-07-04
Feat/mikro orm based linkable

adrien2p:
**What**
- Generate simple linkable for mikro orm based modules 
- fix module util
- fix joiner config builder
- fix define link relationship extension
- fix migrations not loading custom links


packages/core/utils/src/modules-sdk/joiner-config-builder.ts
```diff
@@ -70,6 +70,10 @@ export function defineJoinerConfig(
 
     while (!stopSearching) {
       const fullPath = getCallerFilePath(index)
+      if (!fullPath) {
+        stopSearching = true
```
sradevski: suggestion: Should we maybe just do `while (true)` and remove the `stopSearching` variable (and use `break` instead)? I think it might clean up the code if we have only 1 way of exiting the loop

sradevski (approved):

packages/core/utils/src/modules-sdk/joiner-config-builder.ts
adrien2p: yes you are right 👍 let me change that

olivermrbl (approved):
🚢 🚢

riqwan (approved):
🚢 🚢 🚀

PR #7947, opened 2024-07-04
fix(product): category tree breaks due to pagination limit

riqwan:
what:

- when children scope beyond the pagination limit, we fail to load all descendants and ancestors
- we fix this by fetching all categories in the tree scoped by the where clause


riqwan:
@sradevski take a look to see if this is correct 🤞🏻

packages/modules/product/src/repositories/product-category.ts
```diff
@@ -158,37 +159,34 @@ export class ProductCategoryRepository extends DALUtils.MikroOrmBaseTreeReposito
...
-    }
+    delete where.id
+    delete where.mpath
+    delete where.parent_category_id
 
-    let allCategories = await manager.find(
-      ProductCategory,
-      whereOptions as MikroFilterQuery<ProductCategory>,
-      findOptions.options as MikroOptions<ProductCategory>
+    const categoriesInTree = JSON.parse(
```
sradevski: I think instead of doing a clone, we can do `Array.from` on the resulting `Collection`

sradevski (approved):
Looks good to me, although I must say that part of the code is a bit difficult to follow 😄

riqwan:
@sradevski didn't you write that? 😄

packages/modules/product/src/repositories/product-category.ts
riqwan: I used `this.serialize` instead. Array.from was throwing some other random errors.

sradevski:
@riqwan I mostly updated the write methods, did very minimal changes to the build hierarchy method. I mean even if I did, it could still be confusing right? 😄

riqwan:
Fair point 😄

PR #7948, opened 2024-07-04
fix(dashboard): order list statuses

fPolic:
**What**
- cleanup order list statuses and display payment statuses

### DIFF
--- a/packages/admin-next/dashboard/src/routes/orders/order-detail/constants.ts
+++ b/packages/admin-next/dashboard/src/routes/orders/order-detail/constants.ts
@@ -4,7 +4,6 @@ const DEFAULT_PROPERTIES = [
   "created_at",
   "canceled_at",
   "email",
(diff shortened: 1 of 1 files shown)


packages/admin-next/dashboard/src/routes/orders/order-list/const.ts
```diff
@@ -4,7 +4,8 @@ const DEFAULT_PROPERTIES = [
   "created_at",
   "email",
   "display_id",
-  // "payment_status", // -> TODO replacement for this
+  "payment_status",
```
sradevski, line 7: I think we have this commented out in the order details as well, can you also check it's working as expected there?

fPolic: there was a TODO which is no longer needed, but currently, we don't show the payment section on the details page since it needs to be revisited

sradevski (approved):

PR #7960, opened 2024-07-04
feat: Add HTTP endpoints and workflows for price preference management

sradevski:
REF CORE-2376

Remaining pieces are adding UI to manage the flag, showing the flag in price editor, plugging it in cart calculations, and https://github.com/medusajs/medusa/pull/7827


integration-tests/http/__tests__/price-preference/admin/price-preference.spec.ts
```diff
@@ -0,0 +1,223 @@
...
+        it("filters price preferences by attribute", async () => {
+          const response = (
+            await api.get(
+              "/admin/price-preferences?attribute=region_id",
+              adminHeaders
+            )
+          ).data.price_preferences
+
+          expect(response).toEqual(
+            expect.arrayContaining([
```
olivermrbl: nit: I think this shouldn't use `arrayContaining` as that could give us a false positive test result:

olivermrbl (approved):
Nice work, clean PR 💪

integration-tests/http/__tests__/price-preference/admin/price-preference.spec.ts
sradevski: Yes, copy-paste mistake. Nice catch 👍

packages/core/core-flows/src/pricing/steps/create-price-preferences.ts
```diff
@@ -0,0 +1,34 @@
...
+    if (!pricePreferences?.length) {
+      return
+    }
+
+    const pricingModule = container.resolve<IPricingModuleService>(
+      ModuleRegistrationName.PRICING
+    )
+
+    await pricingModule.deletePricePreferences(pricePreferences)
+  }
```
riqwan, line 33:
```suggestion
  async (ids, { container }) => {
    if (!ids?.length) {
      return
    }

    const pricingModule = container.resolve<IPricingModuleService>(
      ModuleRegistrationName.PRICING
    )

    await pricingModule.deletePricePreferences(ids)
  }
```

packages/core/core-flows/src/pricing/steps/update-price-preferences.ts
```diff
@@ -0,0 +1,45 @@
...
+  async (prevData, { container }) => {
+    if (!prevData?.length) {
+      return
+    }
+
+    const service = container.resolve<IPricingModuleService>(
+      ModuleRegistrationName.PRICING
+    )
+
+    await service.upsertPricePreferences(prevData)
```
riqwan, line 43: q: shouldn't this be update as well?

riqwan (approved):
🥇

packages/core/core-flows/src/pricing/steps/create-price-preferences.ts
sradevski: damn the automerge was faster 😄 I'll open a new PR now anyway and address this.

packages/core/core-flows/src/pricing/steps/update-price-preferences.ts
sradevski: We pass the prevData as an array of entities, and only the Upsert method accepts such input. `update` is either an ID or a selector, which doesn't work for the reversal (we do this everywhere basically). I do think we might need to revisit the compensation methods in a lot of workflows and test them at the very least

PR #7964, opened 2024-07-05
chore: Migrate modules to use the Module util

adrien2p:
**What**
Apply the `Module` util to each module to handle the export and provide the linkable configuration object.

- Most joiner config that does not have any special config are removed/cleaned up in favor of the autogenerated one or simpler one
- linkable are now available in all modules
- cleaned up the dependencies of the modules

FIXES CORE-2410

### DIFF
--- a/packages/admin-next/dashboard/src/components/common/empty-table-content/empty-table-content.tsx
+++ b/packages/admin-next/dashboard/src/components/common/empty-table-content/empty-table-content.tsx
@@ -3,7 +3,7 @@ import { Button, Text, clx } from "@medusajs/ui"
 import { useTranslation } from "react-i18next"
 import { Link } from "react-router-dom"
 
-type NoResultsProps = {
+export type NoResultsProps = {
   title?: string
   message?: string
   className?: string
--- a/packages/admin-next/dashboard/src/components/layout/shell/shell.tsx
+++ b/packages/admin-next/dashboard/src/components/layout/shell/shell.tsx
@@ -319,11 +319,7 @@ const LoggedInUser = () => {
           <Profile />
           <DropdownMenu.Separator />
           <DropdownMenu.Item asChild>
-            <Link
-              // TODO change link once docs are public
-              to="https://medusa-docs-v2-git-docs-v2-medusajs.vercel.app/"
-              target="_blank"
-            >
+            <Link to="https://docs.medusajs.com/v2" target="_blank">
               <BookOpen className="text-ui-fg-subtle mr-2" />
               Documentation
             </Link>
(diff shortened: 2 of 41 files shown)


packages/modules/auth/integration-tests/__tests__/auth-module-service/index.spec.ts
```diff
@@ -36,6 +37,25 @@ moduleIntegrationTestRunner({
...
+
+        expect(Object.keys(linkable)).toEqual(["authIdentity"])
+
+        linkable.authIdentity.toJSON = undefined
+
+        expect(linkable.authIdentity).toEqual({
+          id: {
+            linkable: "auth_identity_id",
+            primaryKey: "id",
+            serviceName: "auth",
```
olivermrbl, line 53: question: It looks like the `serviceName` is actually the module name. Is that correct @adrien2p, or just a coincidence?

adrien2p: yes exactly it is as we are using the `Modules` enum, I could rename it but it was to be aligned with all the config and mention we have in the joinerConfig, link config etc. The service name wording came from the micro service real

adrien2p: it is also the name we choosen for the utils argument such as joiner config

adrien2p: also, just to clarify something, in the core the Modules enum define the service name, module key (in the medusa config for example) and then we use the ModulesRegistrationName to resolve them because we specified a separate value for that. for a user custom module, the module key is the one used for all the above. So a user can define a const to be exported and use it in the medusa config, various utils or custom config

packages/modules/customer/integration-tests/__tests__/services/customer-module/index.spec.ts
```diff
@@ -1,13 +1,66 @@
...
+import { Module, Modules } from "@medusajs/utils"
+import { CustomerModuleService } from "@services"
 
 jest.setTimeout(30000)
 
 moduleIntegrationTestRunner<ICustomerModuleService>({
   moduleName: Modules.CUSTOMER,
   testSuite: ({ service }) => {
     describe("Customer Module Service", () => {
+      it.only(`should export the appropriate linkable configuration`, () => {
```
carlos-r-l-rodrigues:
```suggestion
      it(`should export the appropriate linkable configuration`, () => {
```

adrien2p: oups

olivermrbl (approved):
Legendary!

PR #7973, opened 2024-07-05
Initial implementation with programmatic APIs

thetutlage:
FIXES: CORE-2479

This implementation wraps the Mikro ORM programmatic API into our own opinionated wrapper that can be used to manage migrations.

Opening the PR for early feedback around the initial design, where the code is stored, the approach to testing. So please feel free to provide feedback :)

### DIFF
--- a/packages/core/utils/src/migrations/integration-tests/__tests__/migrations-revert.spec.ts
+++ b/packages/core/utils/src/migrations/integration-tests/__tests__/migrations-revert.spec.ts
@@ -137,7 +137,7 @@ describe("Revert migrations", () => {
     class CustomTSMigrationGenerator extends TSMigrationGenerator {
       createStatement(sql: string, padLeft: number): string {
         let output = super.createStatement(sql, padLeft)
-        return output.replace("drop table if exists", "drop foo")
+        return output.replace("drop table", "drop foo")
       }
     }
 


packages/core/utils/src/modules-sdk/mikro-orm-cli-config-builder.ts
```diff
@@ -53,17 +51,11 @@ export function defineMikroOrmCliConfig(
...
-    databaseName = options.databaseName
-    // @ts-ignore
-    delete options.databaseName
-  }
+  const databaseName = `medusa-${normalizedModuleName}`
 
   return {
-    clientUrl: `postgres://postgres@localhost/${databaseName}`,
     type: "postgresql",
+    dbName: databaseName,
```
adrien2p, line 58: q: does it build a default client url from just the dbname by default? (in mikro orm)

thetutlage: No idea. I can check it now.

thetutlage: Seems to work fine when no explicit `clientUrl` is provided

packages/core/utils/src/migrations/index.ts
```diff
@@ -0,0 +1,106 @@
...
+      const res = await migrator.up(options)
+      return res
+    } finally {
+      migrator["umzug"].clearListeners()
+      await connection.close(true)
+    }
+  }
+
+  /**
+   * Run migrations for the provided entities
```
olivermrbl, line 84: todo: Only relevant if this is actually what it does. Did you find out what happened in the case of the user reverting several migrations?
```suggestion
   * Revert the most recently executed migration
```

Also, if this is the case, it would be great with a few tests for this.

packages/core/utils/src/migrations/integration-tests/__tests__/migrations-revert.spec.ts
```diff
@@ -0,0 +1,173 @@
...
+  return `Migration${new Date().getTime()}${name ? `_${name}` : ""}`
+}
+
+const pgGodCredentials = {
+  user: DB_USERNAME,
+  password: DB_PASSWORD,
+  host: DB_HOST,
+}
+
+describe("Revert migrations", () => {
```
olivermrbl, line 30: See comment above.

olivermrbl (approved):
Clean PR @thetutlage, nice work. 

Is the next step to integrate this into the CLI?

PR #7974, opened 2024-07-05
fix: Improve get caller file path util

adrien2p:
FIXES CORE-2492

### DIFF
--- a/packages/core/utils/src/common/get-caller-file-path.ts
+++ b/packages/core/utils/src/common/get-caller-file-path.ts
@@ -15,7 +15,7 @@ export function getCallerFilePath(position = 2) {
     if (stack !== null && typeof stack === "object") {
       // stack[0] holds this file
       // stack[1] holds where this function was called
-      // stack[2] holds the file we're interested in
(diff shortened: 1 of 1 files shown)


riqwan (approved):
🥇

thetutlage:
Just curious, where do we use this helper?

adrien2p:
> Just curious, where do we use this helper?

for the joiner config builder, basically since the choice was to make it fully magic and this builder is at different level in the application stack, I needed a way to retrieve the location of the original module caller

adrien2p:
@thetutlage as I was explaining to carlos, since node deprecated the callee and that the require.resolve parent is deprecated as well, I didnt get a choice 😂 if you have anyhting in mind let me know

thetutlage:
@adrien2p Nope. I am not questioning the implementation. Just curious, where do we/in which use cases we look for the callee path?

carlos-r-l-rodrigues (approved):

thetutlage:
> for the joiner config builder, basically since the choice was to make it fully magic and this builder is at different level in the application stack, I needed a way to retrieve the location of the original module caller

Saw this comment now. Cool, makes sense!

PR #7975, opened 2024-07-05
chore: Apply defineMikroOrmCliConfig

adrien2p:
**What**
Ensure all module rely on the same config for the dev db

FIXES CORE-2493


olivermrbl (approved):
Nice clean-up

PR #7982, opened 2024-07-05
fix(vite-plugin-admin): Ensure UI Routes are HMR

kasperkristensen:
**What**
- Fixes an issue that was causing HMR to not work for UI Routes.
- Improves HMR for UI Routes so state is preserved on pages, instead of doing a full page reload on changes.


riqwan (approved):
🥇

sradevski (approved):
Nice 👍

PR #7985, opened 2024-07-05
fix(create-medusa-app): remove glob-related warnings

shahednasser:
On every new installation of `create-medusa-app`, you get a warning related to `glob`'s version.

Since we don't need `glob` anymore (it was used to remove boilerplate files), I've removed it along with its usage. This includes removing the `--no-boilerplate` option.


carlos-r-l-rodrigues (approved):

sradevski (approved):

olivermrbl (approved):
Nice

PR #8001, opened 2024-07-07
Fix product options not getting updated due to missing option id

josetr:
Do note this way of upserting options is still not optimal, since the option values don't keep their identities.

If you update your product options from "sm, md" to "sm, md, lg", all product variants that were using those sm / md options will now have an empty product option value

### DIFF
--- a/packages/modules/product/src/services/product-module-service.ts
+++ b/packages/modules/product/src/services/product-module-service.ts
@@ -1632,6 +1632,7 @@ export default class ProductModuleService
     if (productData.options?.length) {
       ;(productData as any).options = productData.options?.map((option) => {
         return {
+          id: option.id,
           title: option.title,
           values: option.values?.map((value) => {
             return {


riqwan:
Thanks @josetr, would you be able to add a test case for the one you mention in here - `integration-tests/http/__tests__/product/admin/variant.spec.ts`?

@sradevski can you take a look at this one?

sradevski:
I'll have a look in the upcoming few days 👍

sradevski:
@josetr I opened a PR that should fix both issues you mentioned, if it looks good we can close this one.

olivermrbl:
@josetr, even better if you could give [these snapshots](https://github.com/medusajs/medusa/pull/8267#issuecomment-2248188790) a spin and let us know if the issue is resolved :)

josetr:
@sradevski @olivermrbl

The new PR seems to solve both issues that I've mentioned, I'm closing this one

PR #8002, opened 2024-07-07
fix(docs) : Fix wrong model names in the Marketplace recipe

adevinwild:
- Replaced `store_user` that have been copy/paste with the right ones


shahednasser (approved):
LGTM Thanks!

PR #8003, opened 2024-07-07
Feat: Add tax inclusivity to admin

sradevski:
**This PR achieves several things, namely:**
- Add price preferences to SDK
- Add switch to manage tax inclusivity on region
- Show tax-related info on region general section
- Show tax inclusive icon and tooltip in price editors across the UI

**Open points (all of them are non-blocking):**
✅  1. Check designs with Ludvig
✅  2. The price editor has quite a bit of duplication, discuss maybe generalizing and refactoring it a bit (either in this or separate PR)
✅  3. Discuss if we want to stick with this logic of the FE reading the price preferences, or if we want to inject `is_tax_inclusive` to the `currency` and `region` models in the BE

Left to do:
1. Add tax inclusivity management to currency
2. Plug in tax inclusivity preference into checkout/cart
3. Plug tax inclusivity into price calculations (only products for now)

REF CORE-2376

Before (region general section):
Screenshot 2024-07-07 at 15 07 27
After:
Screenshot 2024-07-07 at 15 06 31

Edit Before (edit region):
Screenshot 2024-07-07 at 15 07 31
After:
Screenshot 2024-07-07 at 15 06 36

After (Price inclusivity flag in price editor):
Screenshot 2024-07-07 at 15 07 00

### DIFF
--- a/packages/admin-next/dashboard/src/routes/price-lists/common/hooks/use-price-list-grid-columns.tsx
+++ b/packages/admin-next/dashboard/src/routes/price-lists/common/hooks/use-price-list-grid-columns.tsx
@@ -4,11 +4,10 @@ import { useMemo } from "react"
 import { useTranslation } from "react-i18next"
 
 import { Thumbnail } from "../../../../components/common/thumbnail"
-import { DataGridCurrencyCell } from "../../../../components/data-grid/data-grid-cells/data-grid-currency-cell"
 import { DataGridReadOnlyCell } from "../../../../components/data-grid/data-grid-cells/data-grid-readonly-cell"
 import { createDataGridHelper } from "../../../../components/data-grid/utils"
 import { isProductRow } from "../utils"
-import { IncludesTaxTooltip } from "../../../../components/common/tax-badge/tax-badge"
+import { getPriceColumns } from "../../../../components/data-grid/data-grid-columns/price-columns"
 
 const columnHelper = createDataGridHelper<
   HttpTypes.AdminProduct | HttpTypes.AdminProductVariant
@@ -56,74 +55,22 @@ export const usePriceListGridColumns = ({
         },
         disableHiding: true,
       }),
-      ...currencies.map((currency) => {
-        const preference = pricePreferences.find(
-          (p) =>
-            p.attribute === "currency_code" &&
-            p.value === currency.currency_code
-        )
-
-        return columnHelper.column({
-          id: `currency-price-${currency.currency_code}`,
-          name: `Price ${currency.currency_code.toUpperCase()}`,
-          header: () => (
-            <div className="flex w-full items-center justify-between gap-3">
-              {t("fields.priceTemplate", {
-                regionOrCurrency: currency.currency_code.toUpperCase(),
-              })}
-              <IncludesTaxTooltip includesTax={preference?.is_tax_inclusive} />
-            </div>
-          ),
-
-          cell: (context) => {
-            const entity = context.row.original
-
-            if (isProductRow(entity)) {
-              return <DataGridReadOnlyCell />
-            }
-
-            return (
-              <DataGridCurrencyCell
-                context={context}
-                code={currency.currency_code}
-                field={`products.${entity.product_id}.variants.${entity.id}.currency_prices.${currency.currency_code}.amount`}
-              />
-            )
-          },
-        })
-      }),
-      ...regions.map((region) => {
-        const preference = pricePreferences.find(
-          (p) => p.attribute === "region_id" && p.value === region.id
-        )
-
-        return columnHelper.column({
-          id: `region-price-${region.id}`,
-          name: `Price ${region.name}`,
-          header: () => (
-            <div className="flex w-full items-center justify-between gap-3">
-              {t("fields.priceTemplate", {
-                regionOrCurrency: region.name,
-              })}
-              <IncludesTaxTooltip includesTax={preference?.is_tax_inclusive} />
-            </div>
-          ),
-          cell: (context) => {
-            const entity = context.row.original
-
-            if (isProductRow(entity)) {
-              return <DataGridReadOnlyCell />
-            }
-
-            return (
-              <DataGridCurrencyCell
-                context={context}
-                code={region.currency_code}
-                field={`products.${entity.product_id}.variants.${entity.id}.region_prices.${region.id}.amount`}
-              />
-            )
-          },
-        })
+      ...getPriceColumns({
+        currencies: currencies.map((c) => c.currency_code),
+        regions,
+        pricePreferences,
+        isReadyOnly: (context) => {
+          const entity = context.row.original
+          return isProductRow(entity)
+        },
+        getFieldName: (context, value) => {
+          const entity = context.row.original as any
+          if (context.column.id.startsWith("currency_prices")) {
+            return `products.${entity.product_id}.variants.${entity.id}.currency_prices.${value}.amount`
+          }
+          return `products.${entity.product_id}.variants.${entity.id}.region_prices.${value}.amount`
+        },
+        t,
       }),
     ]
   }, [t, currencies, regions, pricePreferences])
--- a/packages/admin-next/dashboard/src/routes/products/product-create/components/product-create-variants-form/product-create-variants-form.tsx
+++ b/packages/admin-next/dashboard/src/routes/products/product-create/components/product-create-variants-form/product-create-variants-form.tsx
@@ -8,11 +8,10 @@ import { DataGridRoot } from "../../../../../components/data-grid/data-grid-root
 import { DataGridReadOnlyCell } from "../../../../../components/data-grid/data-grid-cells/data-grid-readonly-cell"
 import { DataGridTextCell } from "../../../../../components/data-grid/data-grid-cells/data-grid-text-cell"
 import { createDataGridHelper } from "../../../../../components/data-grid/utils"
-import { DataGridCurrencyCell } from "../../../../../components/data-grid/data-grid-cells/data-grid-currency-cell"
 import { DataGridBooleanCell } from "../../../../../components/data-grid/data-grid-cells/data-grid-boolean-cell"
 import { useRegions } from "../../../../../hooks/api/regions"
-import { IncludesTaxTooltip } from "../../../../../components/common/tax-badge/tax-badge"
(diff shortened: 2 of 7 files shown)


sradevski:
@olivermrbl maybe you have an opinion on the 3rd point, just left it here as a thought. It will make the BE more complicated, but it will simplify the FE (this would be only for reading, so mainly in the pricing tables. Writes would still go directly).

packages/admin-next/dashboard/src/routes/products/product-create/components/product-create-variants-form/product-create-variants-form.tsx
```diff
@@ -167,10 +174,21 @@ const useColumns = ({
...
+        )
+
         return columnHelper.column({
           id: `price_${currency}`,
           name: `Price ${currency.toUpperCase()}`,
-          header: `Price ${currency.toUpperCase()}`,
+          header: () => (
+            <div className="flex w-full items-center justify-between gap-3">
+              {t("fields.priceTemplate", {
+                regionOrCountry: currency,
```
olivermrbl: nit(non-blocking): Is this supposed to be `regionOrCountry`? Seems off given we are passing a `currency`

olivermrbl (commented):
To be honest, I don't find the FE implementation particularly messy, so I am fine going with this in the first iteration. We can revisit later if we find other use cases where having them on the region and currency would be handy.

Everything LGTM, I will just let @kasperkristensen look through the UI changes and approve.

packages/admin-next/dashboard/src/routes/products/product-create/components/product-create-variants-form/product-create-variants-form.tsx
sradevski: No it's wrong, I forgot to update it. I'll change it to currency

packages/admin-next/dashboard/src/hooks/api/price-preferences.tsx
```diff
@@ -0,0 +1,113 @@
...
+import { queryKeysFactory } from "../../lib/query-key-factory"
+
+const PRICE_PREFERENCES_QUERY_KEY = "price-preferences" as const
+export const pricePreferencesQueryKeys = queryKeysFactory(
+  PRICE_PREFERENCES_QUERY_KEY
+)
+
+export const usePricePreference = (
+  id: string,
+  query?: HttpTypes.AdminPricePreferenceListParams,
```
kasperkristensen: Wrong type here

packages/admin-next/dashboard/src/routes/locations/location-service-zone-shipping-option-create/components/create-shipping-options-form/create-shipping-options-prices-form.tsx
```diff
@@ -36,9 +37,12 @@ export const CreateShippingOptionsPricesForm = ({
     limit: 999,
   })
 
+  const { price_preferences: pricePreferences } = usePricePreferences({})
```
kasperkristensen, line 40: In theory you’d might not get all price preferences here. For now I think we should add a very high limit and select only the fields we need to keep the response as small as possible.

packages/admin-next/dashboard/src/routes/price-lists/common/hooks/use-price-list-grid-columns.tsx
```diff
@@ -76,10 +93,21 @@ export const usePriceListGridColumns = ({
...
+        const preference = pricePreferences.find(
+          (p) => p.attribute === "region_id" && p.value === region.id
+        )
+
         return columnHelper.column({
           id: `region-price-${region.id}`,
           name: `Price ${region.name}`,
-          header: `Price ${region.name}`,
+          header: () => (
+            <div className="flex w-full items-center justify-between gap-3">
```
kasperkristensen: Maybe we should make a reusable header component for this. So if we want to change it at some point we can do it from a single place.

packages/admin-next/dashboard/src/routes/regions/region-create/components/create-region-form/create-region-form.tsx
```diff
@@ -91,13 +99,22 @@ export const CreateRegionForm = ({
         automatic_taxes: values.automatic_taxes,
       },
       {
-        onSuccess: ({ region }) => {
-          toast.success(t("general.success"), {
-            description: t("regions.toast.create"),
+        onError: (e) => {
```
kasperkristensen, line 98: Should we not return on error instead of continuing? No matter what we will reach the success state even if both requests fail.

I also think this being managed by two separate endpoint is a bit unfortunate. Even if two different modules are responsible for the data, could we not manage it simultaneously in our core workflows and HTTP layer, so we don’t have to chain requests?

packages/admin-next/dashboard/src/routes/price-lists/common/hooks/use-price-list-grid-columns.tsx
sradevski: @kasperkristensen this refers to my second point in the description - I was gonna only extract the header, but it felt like we can extract a lot more (eg. create a PricingGrid component, or something along those lines), so I left it for now until we discuss it.

WDYT, should we extract the entire pricing data grid?

srindom:
The tax inclusivity indicator should be visible on all columns, but be grayed out for tax exclusive prices and "active" for tax inclusive prices. 

When hovering over the icon it should show "Prices in this column are tax [inclusive/exclusive]".

image

packages/admin-next/dashboard/src/routes/regions/region-create/components/create-region-form/create-region-form.tsx
srindom: +1

packages/admin-next/dashboard/src/routes/locations/location-service-zone-shipping-option-create/components/create-shipping-options-form/create-shipping-options-prices-form.tsx
sradevski: Yeah in practice it shouldn't be more than 10-20, but I'll add some high limit. There aren't actually a lot of fields here, only 5 for now that get selected anyway, so that part should be fine.

packages/admin-next/dashboard/src/routes/regions/region-create/components/create-region-form/create-region-form.tsx
sradevski: That's why I raised the 3rd point, but I think this is a very minor functionality that won't get a lot of interaction, so let's keep it as it is for now. I can create a ticket to revisit this though, WDYT?

I'll check the error handling here.

srindom: If not too cumbersome, I would prefer we move it to the workflow sooner rather than later tbh.

olivermrbl: @sradevski, is this directly related to point 3 though? Maybe I misunderstood. 

Shouldn't we be able to add this as a workflow step on the BE without addressing what you highlighted in the PR description?

sradevski: @olivermrbl Initially I thought of completely hiding the price preferences from the region implementation, both on reads and writes. Then I thought maybe reads is enough (so my ping to you). But I guess what you are alluding to is do the writes but not reads? Or do we do both?

olivermrbl: Yeah, I think the "issue" here is just the fact that we send two requests to update/created tax-inclusivity on regions. Moving this to a step in the create/updateRegion workflows seems reasonable and should be fairly straightforward.

While as the other question is around enriching currencies and regions responses with the price preferences on the BE, so we don't have to do:
```
const preference = pricePreferences.find(
  (p) => p.attribute === "region_id" && p.value === region.id
)
const preference = pricePreferences.find(
  (p) => p.attribute === "currency_code" && p.value === currency.code
)
```

...right?

I would say the writes are the most sensible to address now, so we get rid of the doubled requests.

sradevski: @olivermrbl yes exactly, to get rid of fetching the preferences in the UI, as they will be part of the region/currency (similar to calculated price in a way, but it will come from the HTTP layer, not a module).

Ok cool, let's do the writes then, and we can leave the reads for now.

packages/admin-next/dashboard/src/routes/price-lists/common/hooks/use-price-list-grid-columns.tsx
kasperkristensen: I think the header would be a good start, extracting the table in it self would be ideal. 
 
 They will be the same across the project except for two parameters: The way data is instantiated, and whether the data is displayed purely in relation to the product variants or also the owning product (subrows). If you can abstract that away in a nice way then I think it would be be great, but if its a lot of work, I am not sure if its worth it, since the amount of boilerplate for setting up the grids aren't too much, especially if we make the header re-usable.

sradevski: Ok let me see if it can be extracted nicely, if not I'll just do the header.

sradevski:
@olivermrbl @kasperkristensen @srindom applied all the changes discussed, please have a look.

Screenshot 2024-07-08 at 15 25 09

kasperkristensen (approved):
LGTM!

packages/core/core-flows/src/region/workflows/update-regions.ts
```diff
@@ -1,42 +1,73 @@
...
           return {
             id: region.id,
-            payment_providers: data.data.payment_providers,
+            payment_providers: data.normalizedInput.payment_providers,
           }
         })
       }
     )
 
+    when({ normalizedInput }, (data) => {
```
olivermrbl, line 44: Nice to see this being put to use!

olivermrbl (approved):
LGTM

PR #8021, opened 2024-07-08
feat(dashboard): allocate items

fPolic:
**What**
- add Allocate items flow

---

https://github.com/medusajs/medusa/assets/16856471/8485c8dc-b1f2-4239-bb22-996345d5d2ad


olivermrbl:
@fPolic, is this PR ready for review?

fPolic:
> @fPolic, is this PR ready for review?

I'm almost done but have a few more details to align with the design, will open for review later today or tomorrow

olivermrbl:
Damn, this is looking good. Nice work. 

One thing I noticed: Upon allocation, the status of the items on the order remains unchanged as "Not allocated".

 Is something missing on the BE?

CleanShot 2024-07-10 at 15 37 31

fPolic:
Yup, noticed that after I recorded the video 😅 . It's fixed in [this](https://github.com/medusajs/medusa/pull/8021/commits/a24b1f11a9bea1876b59845fe20fda61ff3e2f58) commit.

olivermrbl:
I think we should not show the items allocation when we don't manage inventory on the variants of the items.

Overall, there are three cases to cover:
- All items need allocation -> Show allocation status on items on the order page and show all items in the items allocation view
- Some items need allocation -> Show allocation status only on managed items on the order page and show only the managed items in the items allocation view
- No items need allocation -> Don't show any allocation-related information at all

Allocation status here refers to "Not allocated", "Allocated", etc.

For the third bullet, the allocated information that needs to be removed are the following highlighted in red:
image

Let me know if this makes sense.

fPolic:
@olivermrbl can you check [this commit](https://github.com/medusajs/medusa/pull/8021/commits/e223be780594b3ef703d8b323c037012b957b66f) for the conditions.
Also, should one of the conditions to show the allocate button be if there are items that are fulfilled but not shipped?

olivermrbl:
> Also, should one of the conditions to show the allocate button be if there are items that are fulfilled but not shipped?

When **all** items are fulfilled, allocation is no longer relevant, as the stock will have been adjusted as part of creating the fulfillment.

So, to summarize, we shouldn't show the "Allocate items" button if the items have been fulfilled. Unless, there are some items that are yet to be fulfilled and we manage inventory for those items.

olivermrbl (approved):
Functionality-wise, LGTM!

PR #8034, opened 2024-07-09
Attempt to fix flaky migration tests

thetutlage:
To be honest, I cannot pin-point where exactly the issue is with these tests. The last two errors were.

- The `database name does not exist`. We have a before each hook creating the DB and I cannot find anything preventing this hook from running.
- The `administrator command, access denied`. Again, no freaking idea how the same database is created and dropped in one run and how its giving privileges error in next run.

So what I have done is.

- Perform cleanup before and after each test.
- Earlier the migrations were programmatically created inside a nested directory, but the file system cleanup function was not removing the top-level parent directory. So I have fixed that.


thetutlage:
If these tests are still flaky, then might have to re-think the testing setup and double check if `pg-god` is having some silent failures under the hood.

riqwan (approved):
worthy shot!

olivermrbl:
@thetutlage, should we try to merge this bad boy?

thetutlage:
Yeah. Let's merge it. If it happens again, then Ill dig deeper

PR #8035, opened 2024-07-09
feat(core-flows,types): cancel order changes workflow

riqwan:
what:

- adds cancel order changes workflow


packages/core/core-flows/src/order/steps/cancel-order-change.ts
```diff
@@ -0,0 +1,38 @@
...
+    })
+
+    await service.cancelOrderChange(data)
+
+    return new StepResponse(void 0, {
+      id: data.id,
+      status: dataBeforeUpdate.status,
+      canceled_at: null,
+      canceled_by: null,
+      metadata: dataBeforeUpdate.metadata,
```
carlos-r-l-rodrigues: do we need the metadata here?
that probably won't be touched if we only update the other fields no?

carlos-r-l-rodrigues (approved):

packages/core/core-flows/src/order/steps/cancel-order-change.ts
riqwan: good catch, made it dynamic based on inputs

PR #8038, opened 2024-07-09
fix(dashboard): JSON view

kasperkristensen:
**What**
- Updates the look of the JSON view

**Note**
- This component, plus many others need a different styling of the scrollbar and thumb. Will follow up with that in a separate PR, properly introducing some component based on `@radix-ui/react-scroll-area`.

image

RESOLVES CORE-2499


packages/admin-next/dashboard/src/components/common/json-view-section/json-view-section.tsx
```diff
@@ -42,86 +46,153 @@ export const JsonViewSection = ({
...
+}
+
+const Copied = ({ style, value }: CopiedProps) => {
+  const [copied, setCopied] = useState(false)
+
+  const handler = (e: MouseEvent<HTMLSpanElement>) => {
+    e.stopPropagation()
+    setCopied(true)
+
+    if (typeof value === "string") {
```
sradevski, line 171: You don't really need this check, `JSON.stringify("string", null, 2)` will just give you "string" anyway

sradevski (approved):
Nice 👍

packages/admin-next/dashboard/src/components/common/json-view-section/json-view-section.tsx
kasperkristensen: Seb wants strings to be copied without quotation marks, and as you say it gives you "string" 😄

kasperkristensen:
@olivermrbl Can you take a look at this on behalf of the UI team, this PR also includes the latest styles from Figma, that I need for other PRs so would be great to get it merged 😄

olivermrbl (approved):

PR #8045, opened 2024-07-09
fix(dashboard): Allow cmd+click on table rows

kasperkristensen:
**What**
- Wraps the inner cells in Links when a `navigateTo` is provided to a `DateTable`. This allows interacting with the rows as a normal link, such as cmd + click to open in a new tab.

RESOLVES CORE-2365


olivermrbl (approved):
Nice!

PR #8049, opened 2024-07-09
fix(dashboard): stock location hook cache invalidation

fPolic:
**What**
- this is a temporary fix for the issue where we get stale stock location after enebling/disabling a pickup/shipping etc.
- remove loaders from the create routes

_**Context**_
This issue occurs in a specific scenario where we visit a page which populates react query cache (e.g. we render useStockLocation(...) hook) and then we go to another page where we invalidate that cache record with some action.
When we go back to the original page **since the useQuery wasn't mounted**, data wasn't refetched but react useQuery hook will return stale data from the cache despite `refetchOnMount` flag being set and `staleTime` being 0 or very small value. As a temporary workaround, in this case, I used `refetchType: "all",` flag in `queryClient.invalidateQueries` which will refetch all endpoints from the cache relevant to that key (even if we don't navigate to the page that renders that data). This might be an internal issue with the query library but since we will probably have more cases like this it would be worth to dig deeper into this issue.

---

FIXES CORE-2194


olivermrbl:
@fPolic, I am good to merge it as is. Have you created a ticket for the further investigation? Don't think we have to spend a lot of time on this right now, but would be good to keep a record of it.

olivermrbl (approved):

fPolic:
@olivermrbl sounds good! I've created CORE-2572 to keep track of this.

PR #8050, opened 2024-07-09
chore: rename return flow

carlos-r-l-rodrigues:
What:
 * Move create return flow to "create complete return"
 * Added `declarationMap` to tsconfig to the IDE correctly navigate the project

### DIFF
--- a/www/packages/docs-ui/tsconfig.json
+++ b/www/packages/docs-ui/tsconfig.json
@@ -13,9 +13,7 @@
     "module": "ESNext",
     "target": "es6",
     "declaration": true,
-    "declarationMap": true,
-    "composite": true,
-    "incremental": true
+    "declarationMap": true
   },
   "include": ["src/**/*", "tailwind.config.js", "postcss.config.js"]
 }


packages/core/core-flows/src/order/workflows/create-complete-return.ts
```diff
@@ -365,10 +365,6 @@ export const createReturnOrderWorkflow = createWorkflow(
         shipping_method: shippingMethodData,
         created_by: input.created_by,
       }),
-      updateOrderTaxLinesStep({
-        order_id: input.order_id,
-        shipping_methods: [shippingMethodData as any], // The types does not seems correct in that step and expect too many things compared to the actual needs
-      }),
```
riqwan, line 371: q: is this being removed on purpose?

riqwan (approved):
one question, but lgtm!

packages/core/core-flows/src/order/workflows/create-complete-return.ts
carlos-r-l-rodrigues: Yes, we shouldn't recalculate taxes for existing line items here.

adrien2p (approved):

shahednasser (approved):

olivermrbl (approved):

PR #8056, opened 2024-07-10
feat(core-flows, order): add workflow to create change order actions

riqwan:
what:

- adds workflow to create change order actions


integration-tests/modules/__tests__/order/workflows/order-change-actions.ts
```diff
@@ -0,0 +1,100 @@
...
+            errors: [error],
+          } = await createOrderChangeActionsWorkflow(container).run({
+            input: [{ order_id: order.id } as any],
+            throwOnError: false,
+          })
+
+          expect(error.error).toEqual(
+            expect.objectContaining({
+              message: expect.stringContaining(
+                `Value for OrderChangeAction.action is required, 'undefined' found`
```
olivermrbl: thought: The error message here doesn't seem to relate to what we are testing for. Can we throw something more descriptive?

olivermrbl (approved):
LGTM, just a single comment

integration-tests/modules/__tests__/order/workflows/order-change-actions.ts
riqwan: ah good catch, changed it!

PR #8058, opened 2024-07-10
fix(dashboard): Add safe zone around link Table rows

kasperkristensen:
**What**
- After adding Table rows as links there was an issue where clicking close to a Checkbox in a row would click the link. This PR adds a safe zone around the Checkboxes, so you don't end up navigating when trying to select a row.

image

RESOLVES CORE-2566


olivermrbl (approved):

olivermrbl:
@kasperkristensen, can you include a fix for [this one](https://linear.app/medusajs/issue/CORE-2566/entire-row-in-tables-should-be-clickable) in this PR as well.

kasperkristensen:
> @kasperkristensen, can you include a fix for [this one](https://linear.app/medusajs/issue/CORE-2566/entire-row-in-tables-should-be-clickable) in this PR as well.

Already part of the PR 👍

olivermrbl:
> Already part of the PR 👍

Ah my bad. Just read the description and thought it was only fixing the issue with a checkbox safe zone

PR #8060, opened 2024-07-10
Feat: Improvements to the migrations CLI and workflow

thetutlage:
FIXES: CORE-2495
FIXES: CORE-2482

Breaking change

This PR revamps the migration commands to work as per the new setup. Here's how the migrations will `run` and `revert` moving forward.

- The `run` command will continue to work as it is. I have to redo the CLI UI for it, but the behavior will remain the same.
- The `revert` command can only be used for specific modules. You can no longer revert migrations without specifying the module name. Here's the reasoning for that.

The goal of the `revert` command is to undo mistakes during development. We should not confuse `revert` with `rollback`, where the later one is used to rollback the database state to a previous version.

Since, the `revert` command intent is highly tied to development and that too within a specific module, you will have to run it against a module as well. There is no global `revert` anymore.

## Changelog

This PR revamps the migrations workflow and the CLI output entirely. Here's a summary of the changes.

- The commands remains the same as `npx medusa migrations run` and `npx medusa migrations revert`
- The CLI output has been changed to be more visually appealing and add separate migrations logs for each module with a line break.

### Breaking changes

The `revert` command has a breaking change that it requires one or more module names to perform the revert. It means, the action revert is not valid at the global/app level. It must be executed on a specific module.

If you are developing a custom module, you can run the `revert` command for it as follows.

```sh
# helloWorld is the moduleName
npx medusa migrations revert helloWorld
```


thetutlage:
Amazed that tests are passing already 🙄

thetutlage:
THE PR IS AVAILABLE FOR REVIEW. BUT, DO NOT MERGE.

packages/core/modules-sdk/src/medusa-app.ts
```diff
@@ -402,10 +403,34 @@ async function MedusaApp_({
     return await remoteQuery.query(query, variables, options)
   }
 
-  const applyMigration = async (linkModuleOptions, revert = false) => {
-    for (const moduleName of Object.keys(allModules)) {
-      const moduleResolution = MedusaModule.getModuleResolutions(moduleName)
+  const applyMigration = async ({
```
adrien2p, line 407: suggestion: As we discussed, we can probably move this one outside the main function wdyt?

packages/core/modules-sdk/src/medusa-app.ts
```diff
@@ -417,55 +442,73 @@ async function MedusaApp_({
...
+        ? { ...linkModuleOrOptions.options }
+        : {
+            ...(linkModuleOrOptions as Partial<ModuleServiceInitializeOptions>),
+          }
+
+    options.database ??= {
+      ...sharedResourcesConfig?.database,
+    }
+
+    await revertLinkModuleMigration(
```
adrien2p, line 510: Q: I suppose we will move that in a later pr right?

packages/core/modules-sdk/src/medusa-module.ts
```diff
@@ -544,6 +545,9 @@ class MedusaModule {
       options,
     })
 
+    const logger_ =
+      container?.resolve("logger", { allowUnregistered: true }) ?? logger
```
adrien2p:
```suggestion
      container?.resolve(ContainerRegistrationKeys.LOGGER, { allowUnregistered: true }) ?? logger
```

packages/core/modules-sdk/src/medusa-module.ts
```diff
@@ -572,6 +577,9 @@ class MedusaModule {
...
```
adrien2p:
```suggestion
      container?.resolve(ContainerRegistrationKeys.LOGGER, { allowUnregistered: true }) ?? logger
```

packages/medusa/src/commands/migrate.ts
```diff
@@ -11,7 +12,9 @@ const main = async function ({ directory }) {
   args.shift()
   args.shift()
 
+  const action = args[0]
   const container = await initializeContainer(directory)
+  container.register("logger", asValue(Logger))
```
adrien2p: todo: the initializeContainer already register the logger, so you should be fine here

packages/medusa/src/commands/migrate.ts
```diff
@@ -20,21 +23,39 @@ const main = async function ({ directory }) {
   const plugins = getResolvedPlugins(directory, configModule, true) || []
   const pluginLinks = await resolvePluginsLinks(plugins, container)
 
-  if (args[0] === "run") {
+  if (action === "run") {
+    Logger.info("Running migrations...\n")
```
adrien2p: Q: Do we need to add a new line since every call is on a new line?

packages/medusa/src/commands/migrate.ts
```diff
@@ -20,21 +23,39 @@ const main = async function ({ directory }) {
...
+
     await migrateMedusaApp({
       configModule,
       linkModules: pluginLinks,
       container,
     })
 
-    Logger.info("Migrations completed.")
+    Logger.info("\nMigrations completed")
```
adrien2p: same question

adrien2p (approved):
💪

packages/core/modules-sdk/src/medusa-app.ts
thetutlage: Yeah. First have to work on a `sync links` implementation and then remove it.

packages/medusa/src/commands/migrate.ts
thetutlage: I want an empty line after the `Running migrations` log to visually break the migration logs from the process start logs. It will be clear once I will post screenshots

packages/medusa/src/commands/migrate.ts
thetutlage: Yup.

thetutlage:
Screenshot 2024-07-11 at 1 02 21 PM

This is how the output looks like. I wish we could use a better logger for the CLI commands. Winston feels more like an app logger than a visually appealing logger

adrien2p:
> Screenshot 2024-07-11 at 1 02 21 PM
> This is how the output looks like. I wish we could use a better logger for the CLI commands. Winston feels more like an app logger than a visually appealing logger

Lovely :)

we might need to remove `service` from all the modules key cc @olivermrbl :)

packages/medusa/src/commands/migrate.ts
thetutlage: Makes sense!

olivermrbl:
> we might need to remove service from all the modules key 

+1, let's create a chore ticket for it 👍

thetutlage:
Here are a few more screenshots

**Error when running revert without module name**

Screenshot 2024-07-11 at 1 22 57 PM

**Reverting migrations from one module**

Ignore the logs from Link module. It is yet to be refactored

Screenshot 2024-07-11 at 1 23 16 PM

**Re-running reverted migrations**

Screenshot 2024-07-11 at 1 23 31 PM

olivermrbl:
Uhh this is sexy @thetutlage

olivermrbl:
Really digging these logs 😍

thetutlage:
@olivermrbl Can be improved a lot, we need a visual logger for CLI commands and ditch Winston. Will get onto refactoring CLI output someday 💪

packages/core/modules-sdk/src/medusa-app.ts
thetutlage: Uses `sharedContainer` and `sharedConfig` from the upper scope. Should we pass them down as parameters?

adrien2p: I think so, just to lighten this function a bit to simplify reading and understanding

srindom:
@thetutlage - in production envs the logger uses a json format. Maybe we want to test that the output isn't filled with these types of outputs:

```
{ "level": "info", "message": "-------------------...." }
```

adrien2p:
> @thetutlage - in production envs the logger uses a json format. Maybe we want to test that the output isn't filled with these types of outputs:
> 
> ```
> { "level": "info", "message": "-------------------...." }
> ```

Do you think it would be problematic to have few spaces at the beginning of the message?

thetutlage:
@srindom Yes, it will be filled with that.

The ideal solution is to have helpers for these sort of outputs that has different behavior when the output is JSON. 

But for now, either we have to remove these lines or apply a filter when these logs are consumed by a logging service.

adrien2p:
Yes and I believe our logger can have a custom handler for message and strip the space on front of it

thetutlage:
revert_error_invalid_module
revert_error_without_module

srindom:
It shouldn't block this PR, but think it would be good for us to look into a way to specify that a log message is just for aesthetic purposes - we could maybe just use the silly log level for it, but let's leave that for another time :)

thetutlage:
Alright. We are ready to merge this PR. I have updated the description explaining the breaking change.

thetutlage:
CC @olivermrbl @adrien2p Ready to be re-reviewed and merged

packages/cli/medusa-cli/src/create-cli.ts
```diff
@@ -170,13 +170,18 @@ function buildLocalCommands(cli, isLocalProject) {
       ),
     })
     .command({
-      command: `migrations [action]`,
+      command: `migrations [action] [modules...]`,
```
adrien2p, line 173: suggestion: is it a common approach to have the spread on the right? otherwise I would suggest `[...modules]` wdyt?

packages/core/utils/src/modules-sdk/migration-scripts/migration-down.ts
```diff
@@ -34,34 +28,30 @@ export function buildRevertMigrationScript({
   > = {}) {
     logger ??= console as unknown as Logger
 
-    const dbData = loadDatabaseConfig(moduleName, options)!
-    const entities = Object.values(models).map((model) => {
-      if (DmlEntity.isDmlEntity(model)) {
-        return toMikroORMEntity(model)
-      }
+    console.log(new Array(TERMINAL_SIZE).join("_"))
+    console.log("")
```
adrien2p: Q: any reason to not use the logger? is it to prevent it from appearing in the json log?

adrien2p (approved):

packages/core/utils/src/modules-sdk/migration-scripts/migration-down.ts
thetutlage: The logger has a prefix of `info` by default and therefore it will show up in the empty line and same for the `-`

packages/cli/medusa-cli/src/create-cli.ts
thetutlage: The yargs parser needs it at the end. So it is not only for the display purposes, it is to make their parse work

olivermrbl:
@thetutlage, upon merging a PR, can I get you to prefix the commit message (and ideally the PR title) with a tag indicating the PR type `feat`, `fix`, `chore`, or `docs`? 💪

thetutlage:
@olivermrbl Sure. Should this one be `refactor`?

olivermrbl:
@thetutlage, we don't have `refactor`, so I think we should go with `feat` for now. It is used to group the changes in the release notes and right now we have the following sections:
- Features
- Bugs
- Chores
- Documentation

thetutlage:
Cool!

olivermrbl:
Very nit, but can you use the lowercased version hehe. Just for consistency 😅

PR #8061, opened 2024-07-10
docs: updated marketplace recipe + added example

shahednasser:
- Moved recipes into its own section of the learning resources
- Updated the marketplace recipe overview
- Added an example implementation of the marketplace recipe

Closes DOCS-791

### DIFF
--- a/www/apps/resources/app/recipes/marketplace/examples/vendors/page.mdx
+++ b/www/apps/resources/app/recipes/marketplace/examples/vendors/page.mdx
@@ -663,7 +663,7 @@ export const POST = async (
 }
 ```
 
-This adds a `POST` API route at `/vendors/products`. It resolves the necessary module main services, and retrieves the sales channels and vendor admin.
+This adds a `POST` API route at `/vendors/products`. It resolves the necessary modules' main services, and retrieves the sales channels and vendor admin.
 
 In the place of the `TODO`, add the following:
 


www/apps/resources/app/recipes/marketplace/examples/vendors/page.mdx
```diff
@@ -0,0 +1,1344 @@
...
+    mappedBy: "admins"
+  })
+})
+
+export default VendorAdmin
+```
+
+This creates a `VendorAdmin` data model, which represents an admin of a vendor.
+
+Then, add a migration file at `src/modules/marketplace/migrations/Migration20240708151444.ts`:
```
olivermrbl: question: Should we not be teaching them about generating migrations instead of providing them with a copy/paste?

It's more error-prone, but if the generated migrations leads to issues, then there is something we need to fix elsewhere.

www/apps/resources/app/recipes/marketplace/examples/vendors/page.mdx
```diff
@@ -0,0 +1,1344 @@
...
+  ): Promise<VendorData[] | VendorData> {
+    const input = (Array.isArray(data) ? data : [data]).map(
+      (vendor) => {
+        if (!vendor.handle) {
+          vendor.handle = vendor.name.toLowerCase().replace(/\s/g, "-")
+        }
+
+        return vendor
+      }
+    )
```
olivermrbl: thought: I understand the idea with the `handle` formatting, but for the sake of the recipe, wouldn't it be cooler to have an empty service? This would make it more obvious that CRUD is coming out of the box.

www/apps/resources/app/recipes/marketplace/examples/vendors/page.mdx
```diff
@@ -0,0 +1,1344 @@
...
+
+```
+
+After that, create the module definition at `src/modules/marketplace/index.ts` with the following content:
+
+```ts title="src/modules/marketplace/index.ts"
+import { Module } from "@medusajs/utils"
+import MarketplaceModuleService from "./service"
+
+export const MARKETPLACE_MODULE = "marketplaceModuleService"
```
olivermrbl, line 158: question: Shouldn't this be `marketPlace`?

www/apps/resources/app/recipes/marketplace/examples/vendors/page.mdx
shahednasser: We can add that. Considering this is tutorial-like, this approach is meant to keep the steps simple. So instead of saying creating the config file -> run the CLI tool -> here's the resulting migration, it just shows the migration.

www/apps/resources/app/recipes/marketplace/examples/vendors/page.mdx
shahednasser: We aligned before on making it a convention for custom module to be `{moduleName}ModuleService`.

www/apps/resources/app/recipes/marketplace/examples/vendors/page.mdx
```diff
@@ -0,0 +1,1344 @@
...
+Create the file `src/api/vendors/route.ts` with the following content:
+
+export const vendorRouteHighlights = [
+  ["10", "schema", "Define the fields expected in the request body."],
+  ["45", "parse", "Validate the request body and, if valid, retrieve it as an object."],
+  ["51", "createVendors", "Create the vendor using the Marketplace Module's main service."],
+  ["54", "createVendorAdminWorkflow", "Execute the workflow created in the first step."],
+]
+
+```ts title="src/api/vendors/route.ts" highlights={vendorRouteHighlights}
```
olivermrbl: thought(non-blocking): This file seems a bit overwhelming to get shown right away. I think we should consider breaking up the different elements of the file in a sort of incremental approach to creating the API route. Will make it more digestible.

www/apps/resources/app/recipes/marketplace/examples/vendors/page.mdx
```diff
@@ -0,0 +1,1344 @@
...
+To add the create product API route, add to the same file the following:
+
+export const createProductsHighlights = [
+  ["15", "CreateProductWorkflowInputDTO", "Accept the same request body as Medusa's Create Product API route"],
+  ["32", "retrieveVendorAdmin", "Retrive the vendor admin to retrieve its vendor's ID."],
+  ["39", "createProductsWorkflow", "Use Medusa's workflow to create a product."],
+  ["50", "create", "Create a link between the created product and the vendor."]
+]
+
+```tsx title="src/api/vendors/products/route.ts" highlights={createProductsHighlights}
```
olivermrbl: Same thought here.

olivermrbl (approved):
Nice work, left a few comments here and there, but not something we need to address now.

www/apps/resources/app/recipes/marketplace/examples/vendors/page.mdx
olivermrbl: That would be the case for the module service, but this should not be the module key. But maybe this is not something to address in this documentation PR. It will likely need some work/renamings in the framework first, and then it can be reflected in the docs afterward.

shahednasser: Cool, from what Adrien mentioned they all should be the same name

olivermrbl: Will create a ticket, so we can address later 👍

www/apps/resources/app/recipes/marketplace/examples/vendors/page.mdx
olivermrbl: Ah I see. I just thought it would be nice to get the workflow to generate migrations under their skin. The next step for them after this Recip would probably be to create more models or modules, and then they wouldn't have a migration to copy over.

Anyway, I'll let it be up to you 👍

shahednasser: Makes sense when you put it that way! I changed the section to generate the migration

PR #8066, opened 2024-07-10
fix: Investigate geo zones with address checks

adrien2p:
need clarification

FIXES CORE-2432


adrien2p:
@olivermrbl should we merge this one?

olivermrbl (approved):

PR #8067, opened 2024-07-10
fix(tax): Add indexes to enforce unique constraint on tax region

olivermrbl:
**What**

Introduce unique index on (country_code) where province_code is null to enforce a unique constraint on `(country_code, province_code)`.

This means there are now two unique indexes on the table:
- `(country_code) where province_code is null`
- `(country_code, province_code)`

**Why**

The existing unique index does not take into account null province codes, so it is possible to create multiple tax regions with the same country_code without a province_code.


riqwan (approved):

srindom (approved):
LGTM

PR #8070, opened 2024-07-10
fix(utils): define link alias + pluralize

carlos-r-l-rodrigues:
What:
 * Fix alias of models that have "compound names": eg "related_product"
 * `deleteCascade` option
 * Package to pluralize english words correctly

### DIFF
--- a/integration-tests/modules/__tests__/link-modules/define-link.spec.ts
+++ b/integration-tests/modules/__tests__/link-modules/define-link.spec.ts
@@ -1,8 +1,9 @@
+import { medusaIntegrationTestRunner } from "medusa-test-utils"
+
 import CurrencyModule from "@medusajs/currency"
 import { MedusaModule } from "@medusajs/modules-sdk"
 import RegionModule from "@medusajs/region"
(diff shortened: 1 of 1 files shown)


olivermrbl (approved):
LGTM

packages/core/utils/package.json
```diff
@@ -40,6 +40,7 @@
     "dotenv": "^16.4.5",
     "jsonwebtoken": "^9.0.2",
     "knex": "2.4.2",
+    "pluralize": "^8.0.0",
```
adrien2p, line 43: Q: what is the impact with our current pluralize type compared to real output?

adrien2p (approved):

packages/core/utils/package.json
carlos-r-l-rodrigues: Worst case scenario I will have to fix tests/Interfaces/Modules where we have wrong plurals.
What we might have to revisit is the TypeScript types that won't be correct

adrien2p: Yes thats definitely my point, the type need to match the function rules, we can check that separately

Issue #8074, opened 2024-07-11
Stripe payment cannot be captured


silenaker:
### Describe the bug

We are using the `@medusajs/payment-stripe` provider for payments. However, after the cart is paid and the order is created, the payment cannot be captured, leading to state inconsistency.

The order status remains "Authorized" in such cases.

<img width="1673" alt="image" src="https://github.com/medusajs/medusa/assets/11684852/8205c510-7a19-4870-b592-2054c098e8b8">

### System information

Medusa version (including plugins): 2.0 preview
Node.js version: 20
Database: postgresql
Operating system: mac os
Browser (if relevant): chrome

### Expected behavior

The following order status is "Captured," as expected.

<img width="1636" alt="image" src="https://github.com/medusajs/medusa/assets/11684852/a6368a34-3db5-48f5-a3a3-15c3c61465f9">

voxsar:
https://github.com/medusajs/medusa/issues/6254

420coupe:
I want to bring this issue back up. As payments are being captured, but statuses are not being updated after payment capture and just being left on authorized. In addition the card_last4 is also not being stored anywhere nor is payment type credit card /visa. Only the payment provider is being logged.

420coupe:
to add some context here, looks like stripe webhook cant find pp_stripe. but payments are being processed on stripe's end. 

![image](https://github.com/user-attachments/assets/fa6c7216-8f8f-4cb2-b003-09477cd4ec9d)


```js
payment_intent.succeeded

Resend

More options
Next retry
in an hour
Response
HTTP status code
502 (Bad Gateway)
error code: 502
Request
{
  "id": ["evt_3Ps5wgK6QjZA3FWk0XqtVyKB"](https://dashboard.stripe.com/test/events/evt_3Ps5wgK6QjZA3FWk0XqtVyKB),
  "object": "event",
  "api_version": "2024-06-20",
  "created": 1724690368,
  "data": {
    "object": {
      "id": ["pi_3Ps5wgK6QjZA3FWk0U1zZ9Vs"](https://dashboard.stripe.com/test/payments/pi_3Ps5wgK6QjZA3FWk0U1zZ9Vs),
      "object": "payment_intent",
      "amount": 267,
      "amount_capturable": 0,
      "amount_details": {
        "tip": {
        }
      },
      "amount_received": 267,
      "application": null,
      "application_fee_amount": null,
      "automatic_payment_methods": {
        "allow_redirects": "always",
[...]

PR #8075, opened 2024-07-11
fix(payment, stripe) stripe payment cannot be captured

silenaker:
#8074


adrien2p (approved):
Sounds good to me, might have to fix some tests or add new one to complete the existing set of tests. I ll let @olivermrbl double check that I haven't missed anything here

packages/modules/payment/src/services/payment-module.ts
```diff
@@ -294,33 +294,42 @@ export default class PaymentModuleService
...
+      paymentSession = await this.createPaymentSession_(
+        paymentCollectionId,
+        input,
+        sharedContext
+      )
+
       const providerSessionSession =
         await this.paymentProviderService_.createSession(input.provider_id, {
-          context: input.context ?? {},
+          context: { ...input.context, resource_id: paymentSession.id },
```
olivermrbl: thought: Should the `resource_id` be overwritable? In case you want to associate the cart instead of the payment session. The Medusa payment session will always have a direct association with the third-party session through its `payment_session.data`.

olivermrbl (approved):
LGTM as well

packages/modules/payment/src/services/payment-module.ts
silenaker: I think we can use a more specific field name to associate the Medusa payment session with the provider session, so as to avoid unintentional overwriting.

olivermrbl: `session_id`?

silenaker: > `session_id`?

it’s ok

silenaker: I removed resource_id and replaced it with session_id. If other resource identifiers like order_id or cart_id are needed, I think they can be added based on future requirements, as currently, only session_id is the most useful.

sradevski (approved):
LGTM! @silenaker can you please also update the tests (as you can see, some are failing) so we can have this merged? Thanks!

silenaker:
> LGTM! @silenaker can you please also update the tests (as you can see, some are failing) so we can have this merged? Thanks!

resolved

carlos-r-l-rodrigues:
Thank you @silenaker

PR #8078, opened 2024-07-11
docs: change event payload handling in subscribers

shahednasser:
Change how event payload is used in a subscriber based on the latest changes.

### DIFF
--- a/www/apps/book/app/advanced-development/events-and-subscribers/data-payload/page.mdx
+++ b/www/apps/book/app/advanced-development/events-and-subscribers/data-payload/page.mdx
@@ -1,3 +1,5 @@
+import { TypeList } from "docs-ui"
+
 export const metadata = {
   title: `${pageNumber} Event Data Payload`,
 }
@@ -10,14 +12,13 @@ In this chapter, you'll learn how subscribers receive an event's data payload.
 
 When events are emitted, they’re emitted with a data payload. 
 
-The object that the subscriber function receives as a parameter has a `data` property, which is the event's data payload.
+The object that the subscriber function receives as a parameter has a `data` property, which is an object holding the event payload with additional context.
 
 For example:
 
 export const highlights = [
   ["7", "", "The event's data payload."],
   ["8", "{ id: string }", "The type of expected data payloads."],
-  ["9", '"data" in data ? data.data.id : data.id', "The payload data is either in `data.data` or directly in `data`."]
 ]
 
 ```ts title="src/subscribers/product-created.ts" highlights={highlights} collapsibleLines="1-5" expandButtonLabel="Show Imports"
@@ -29,7 +30,7 @@ import type {
 export default async function productCreateHandler({
   data,
 }: SubscriberArgs<{ id: string }>) {
-  const productId = "data" in data ? data.data.id : data.id
+  const productId = data.data.id
   console.log(`The product ${productId} was created`)
 }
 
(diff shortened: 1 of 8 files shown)


www/apps/book/app/advanced-development/events-and-subscribers/data-payload/page.mdx
```diff
@@ -29,7 +30,7 @@ import type {
 export default async function productCreateHandler({
   data,
 }: SubscriberArgs<{ id: string }>) {
-  const productId = "data" in data ? data.data.id : data.id
+  const productId = data.data.id
```
olivermrbl, line 33: thought(non-blocking): Still not super happy about the `data.data`. Feels like something is not right. 

@adrien2p, do you have any suggestions here? Can we change the input to the subscriber to `body` instead, so that you would do:
```
export default async function productCreateHandler({
  body,
}: SubscriberArgs) {
  const { data, metadata } = body
  const productId = data.id
```

@shahednasser, please feel free to also pitch in.

shahednasser: `body` sounds ambiguous. we could instead keep `data` as the data payload and pass the `metadata` and `eventName` in a new parameter `context`

olivermrbl: The reason I suggested body is that it is the name of the [type](https://github.com/medusajs/medusa/blob/b3351da264f2da1db11ca2b9420566903de7545b/packages/medusa/src/types/subscribers.ts#L13), but we can change it.

It is worth noting `data` is not the event data payload. It is the entire message content. The actually event data is what we current have at `data.data`. 

But perhaps we should clean up the types here a bit to something along the lines of:
```
export type SubscriberArgs<T = unknown> = {
  event: MessageBody<T> // or something else like `content`
  container: MedusaContainer
  pluginOptions: Record<string, unknown>
}

export type MessageBody<TData = unknown> = {
  name: string
  metadata?: EventMetadata
  data: TData
}
``` 

I think having the `eventName` both in the subscriber arguments and the message content is also confusing. This type refactor would create a clear separation of the concerns. 

Let me know what you think. I am also open to other suggestions.

olivermrbl: What about calling it `event`? 

So you would do:
```
export default async function productCreateHandler({
  event,
}: SubscriberArgs) {
  const { data, metadata } = event
  const productId = data.id
```

shahednasser: > It is worth noting data is not the event data payload. It is the entire message content. The actually event data is what we current have at data.data.

yes I'm aware, I was suggesting splitting it up so we have two properties instead of one `data` property: `data` and `context`.

But I like the `event` option too that you mentioned 👍🏻

olivermrbl: > yes I'm aware, I was suggesting splitting it up so we have two properties instead of one data property: data and context.

Ah I see. 

> But I like the event option too that you mentioned

Let me create a ticket for this and we can tackle it later. Don't want to block this PR 👍

adrien2p: I think I prefer body as it relate to the MessageBody and you are receiving a message

shahednasser: That would require us to change a lot of terminology for users though. The user knows that they're handling an emitted event, they don't know about the underlying work of a message being sent/received.

olivermrbl (approved):

www/apps/book/app/advanced-development/events-and-subscribers/data-payload/page.mdx
adrien2p: yes thats true, should we rename data inside data to payload maybe?

shahednasser: both work for me 👍🏻

olivermrbl: Just linking the [Linear issue](https://linear.app/medusajs/issue/TRI-5/revisit-arguments-passed-to-subscribers) for futher discussions.

PR #8084, opened 2024-07-11
feat(dashboard,medusa): Add updated Metadata Form

kasperkristensen:
**What**
- Adds new Metadata form component.
- Adds the Metadata section as an option to the Page layouts

Skærmbillede 2024-07-11 kl  11 34 06
Skærmbillede 2024-07-11 kl  11 34 33

**Note**
- When Metadata contains non-primitive data, we disable those rows, and show a placeholder value, a tooltip and an alert describing that the row can be edited through the API. I want to add a JSON editor to allow editing these things in admin, but awaiting approval on that.
- This PR only adds the new form to a couple of pages, to keep the PR light, especially since metadata is not implemented correctly in all validators so also needs some changes to the core. This still show some examples of how its used with the new Page layout components. Will follow up with more pages in future PRs. 
- We try to convert the inputs to the best fitting primitive, so if a user types "true" then we save the value as a boolean, "130" as number, "testing" as a string, etc.


srindom:
This looks super nice! One thought (non-blocking, if-minor) could we display the value as JSON in the tooltip when you hover on a disabled row? That way you'd at least be able to see the data. We'd have to consider how this would look for long data, but wanted to hear if you think this even makes sense to do first?

> I want to add a JSON editor to allow editing these things in admin, but awaiting approval on that.

Do you think it's worth the effort to tackle this right now? The non-primitive values would only be there when you do something with these values programmatically so the need for an editor might not be that high anyway.

kasperkristensen:
> This looks super nice! One thought (non-blocking, if-minor) could we display the value as JSON in the tooltip when you hover on a disabled row? That way you'd at least be able to see the data. We'd have to consider how this would look for long data, but wanted to hear if you think this even makes sense to do first?
> 
> > I want to add a JSON editor to allow editing these things in admin, but awaiting approval on that.
> 
> Do you think it's worth the effort to tackle this right now? The non-primitive values would only be there when you do something with these values programmatically so the need for an editor might not be that high anyway.

Re: JSON editor - I don't think it's a high priority thing at all, in general Metadata in admin is properly rather niche. It's more something that would be nice to have at some point down the line. Also nested data structures in V2 seems to be broken atm, see TRI-2. 

I think adding the data as a tooltip could be an okay solution, but you are right that we need to think about how handle cases where it's a large object/array, as Tooltips can't be scrolled.

fPolic (approved):
LGTM! Nice work 👍

olivermrbl:
> I think adding the data as a tooltip could be an okay solution, but you are right that we need to think about how handle cases where it's a large object/array, as Tooltips can't be scrolled.

Would it be too crazy to open the JSON drawer as with any other raw JSON on detail pages? I don't know if this will look weird on top of the modal.

Aside from a possible solution to this use case, we could make this the general approach for displaying any JSON in any context. Again, I am not sure if this will work for all future cases, but having some consistency here could make sense so that the users know what to expect for JSON and how to navigate it.

kasperkristensen:
> > I think adding the data as a tooltip could be an okay solution, but you are right that we need to think about how handle cases where it's a large object/array, as Tooltips can't be scrolled.
> 
> Would it be too crazy to open the JSON drawer as with any other raw JSON on detail pages? I don't know if this will look weird on top of the modal.
> 
> Aside from a possible solution to this use case, we could make this the general approach for displaying any JSON in any context. Again, I am not sure if this will work for all future cases, but having some consistency here could make sense so that the users know what to expect for JSON and how to navigate it.

We could maybe use the new StackedDrawer component, I think this could be pretty clean looks wise. There is just something a little weird in having to open and edit modal, that then opens another modal just to view something. Maybe we should run this by Ludvig, I'll open a ticket for showing disabled metadata fields, and merge this for now, so we can start getting the functionality in on all pages.

olivermrbl:
> Maybe we should run this by Ludvig, I'll open a ticket for showing disabled metadata fields, and merge this for now, so we can start getting the functionality in on all pages.

Sounds good!

PR #8090, opened 2024-07-11
fix: Migration generator and generated migrations

adrien2p:
**What**
Fix the migration generator as well as the generated migrations.
I havent noticed any other places where the same mistake was applied in our other migrations though 🤔 

FIXES: FRMW-2581


thetutlage:
Do we have to fix existing migrations that has `alter column if exists`?

adrien2p:
> Do we have to fix existing migrations that has `alter column if exists`?

as I mentioned in the description, I didn't find any other one, did you find some?

adrien2p:
@thetutlage @olivermrbl do you think the changes to the down should be considered a breaking change? I dont think so but I d like to ask just in case

thetutlage:
@adrien2p Nope. Also, in this case its a bug fix. I don't think that specific migration would have been reverted ever

olivermrbl (approved):

olivermrbl:
Agree with @thetutlage

PR #8096, opened 2024-07-11
fix(utils): error message when isQueryable is possibly not enabled.

shahednasser:
Improve the error message when a link is defined on a module that possibly doesn't have `isQueryable` enabled.


carlos-r-l-rodrigues (approved):
💯

PR #8097, opened 2024-07-11
fix: Handle multiple id prop generation

adrien2p:
throw if multiple id modifier are used

FIXES FRMW-2586

### DIFF
--- a/packages/admin-next/dashboard/src/routes/products/common/variant-pricing-form.tsx
+++ b/packages/admin-next/dashboard/src/routes/products/common/variant-pricing-form.tsx
@@ -78,7 +78,9 @@ export const useVariantPriceGridColumns = ({
           return (
             <ReadonlyCell>
               <div className="flex h-full w-full items-center gap-x-2 overflow-hidden">
-                <span className="truncate">{entity.title}</span>
+                <span title={entity.title} className="truncate">
+                  {entity.title}
+                </span>
               </div>
             </ReadonlyCell>
           )
--- a/packages/admin-next/dashboard/src/routes/promotions/common/edit-rules/components/rule-value-form-field/rule-value-form-field.tsx
+++ b/packages/admin-next/dashboard/src/routes/promotions/common/edit-rules/components/rule-value-form-field/rule-value-form-field.tsx
@@ -84,6 +84,7 @@ export const RuleValueFormField = ({
                   className="bg-ui-bg-base"
                   ref={valuesRef}
                   min={1}
+                  disabled={!fieldRule.attribute}
                 />
               </Form.Control>
               <Form.ErrorMessage />
@@ -97,6 +98,7 @@ export const RuleValueFormField = ({
                   {...field}
                   onChange={onChange}
                   className="bg-ui-bg-base"
+                  disabled={!fieldRule.attribute}
                 />
               </Form.Control>
               <Form.ErrorMessage />
@@ -112,6 +114,7 @@ export const RuleValueFormField = ({
                     Array.isArray(field.value) ? field.value[0] : field.value
                   }
                   onValueChange={onChange}
+                  disabled={!fieldRule.attribute}
                 >
                   <Select.Trigger ref={ref} className="bg-ui-bg-base">
                     <Select.Value placeholder="Select Value" />
@@ -144,6 +147,7 @@ export const RuleValueFormField = ({
                   options={options}
                   onChange={onChange}
                   className="bg-ui-bg-base"
+                  disabled={!fieldRule.attribute}
                 />
               </Form.Control>
 
--- a/packages/admin-next/dashboard/src/routes/promotions/common/edit-rules/components/rules-form-field/rules-form-field.tsx
+++ b/packages/admin-next/dashboard/src/routes/promotions/common/edit-rules/components/rules-form-field/rules-form-field.tsx
@@ -213,7 +213,11 @@ export const RulesFormField = ({
                       return (
                         <Form.Item className="basis-1/2">
                           <Form.Control>
-                            <Select {...field} onValueChange={onChange}>
+                            <Select
+                              {...field}
+                              disabled={!fieldRule.attribute}
+                              onValueChange={onChange}
+                            >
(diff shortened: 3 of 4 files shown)


carlos-r-l-rodrigues:
It looks good, but it is somewhat weird to have 2 ids in the same table.
Should we allow this? @olivermrbl @thetutlage

olivermrbl:
What's happening here, I am not sure?

adrien2p:
I just opened it for discussion tomorrow

adrien2p:
But the current problem is if people are using multiple id type, then we create a text column and a hook to generate an id for that column, the hook was not handled properly if multiple column needs to be autogenerated.

Now the question is do we want to allow it or not, but currently it is the only way to auto generate a ulid for a column even though this issue originated from a miss usage of the api 😅

adrien2p:
BUt in reality id is a bit missleading and the only one that allows you to generate a ulid for a column. Lets discuss it tomorrow guys?

carlos-r-l-rodrigues:
suggestion: `id` should be primary key by default and have the `ulid` generated. And only one could exist in one entity.
And we create another type `uuid` that will only generate the `ulid`

olivermrbl:
What is the use case for multiple autogenerated ulid columns?

adrien2p:
But then to create composite key it gets confusing no? Here primary keys are explicit which also simplify typing checks and computing. I believe just having ulid would be enought

adrien2p:
Just to handle it gracefully, we could also throw instead. Having multiple ulid columns can be useful for example in table with and id and a random key such as api keys but in other context such has generating a project id or anything like this

olivermrbl:
I would probably do something along the lines of what @carlos-r-l-rodrigues suggested, if we want to support this:
```
const someModel = model.define("", {
  id: model.id({ prefix: "mod" }).primaryKey(),
  other_id: model.ulid({ prefix: "oth" })
})
```

But still keep the explicit `.primaryKey()`.

adrien2p:
I can do that having an explicit ulid modifier and throwing if multiple ids are set? But then what is the purpose of id compare to ulid?

olivermrbl:
> But then what is the purpose of id compare to ulid?

Good point. Will just spend some time thinking about this. Let's discuss tomorrow

carlos-r-l-rodrigues:
> What is the use case for multiple autogenerated ulid columns?

I don't have any in mind. just thinking about the feature itself.

I don't have any strong opinion on the primaryKey being explicit. It would just look smaller

```typescript
const someModel = model.define("a", {
  id: model.id({ prefix: "mod" }), // primary key
  other_id: model.ulid({ prefix: "oth" })
})

const otherModel = model.define("b", {
  id: model.uuid({ prefix: "mod" }),
  other_id: model.text().primaryKey()
})

// This is not allowed
const otherModel = model.define("b", {
  id: model.id({ prefix: "mod" }),
  other_id: model.text().primaryKey()
})
```

thetutlage:
Looking at the discussion, at seems like we don't need `id` at all, since it doesn't capture any value.

Maybe we need `model.uuid()` and `model.ulid()` with an explicit `primaryKey` method chain

olivermrbl:
Are we maybe getting ahead of ourselves with this? Correct me if I am wrong, but this was actually not requested, but the idea spawned from misusage of the DML. So perhaps, we could icebox this till we have the use case at hand.

I haven't looked through Discord for requests the past few days, so this is only based on what was mentioned earlier in this discussion. Please push back if we are actually seeing people requesting this.

adrien2p:
> Are we maybe getting ahead of ourselves with this? Correct me if I am wrong, but this was actually not requested, but the idea spawned from misusage of the DML. So perhaps, we could icebox this till we have the use case at hand.
> 
> I haven't looked through Discord for requests the past few days, so this is only based on what was mentioned earlier in this discussion. Please push back if we are actually seeing people requesting this.

We can start with just throwing an error if you have multiple id usage in a single DML, just to prevent those issues from araising.

About usage, there is many cases where it can be useful, for example a table that stores a DNA value and a project id value where the DNA would be a random generated string that is frozen for example. But here I think the issue is not really that, but more what value does the `id` modifier provide to the user, and here it seems that the only purpose of this id modifier is to have an autogenerated value but not really anything else wdyt?

Also, this pr was meant as a fix of an issue raised from miss usage, so I still think we need to fix it

adrien2p:
@olivermrbl should I just do the changes to throw instead when multiple id modifier are used?

olivermrbl:
> But here I think the issue is not really that, but more what value does the id modifier provide to the user, and here it seems that the only purpose of this id modifier is to have an autogenerated value but not really anything else wdyt?

Good point. I guess when we removed implicit primary key generation, we lost the "uniqueness" of the `id` column.

I still like having `id` because it makes it very obvious what developers should use for their id columns. They don't have to worry about using `ulid` vs. `uuid` vs. something else. We make that decision for them.

Could we imagine going back to implicit primary key generation when using `id`?

olivermrbl:
> But here I think the issue is not really that, but more what value does the id modifier provide to the user, and here it seems that the only purpose of this id modifier is to have an autogenerated value but not really anything else wdyt?

Good point. I guess when we removed implicit primary key generation, we lost the “uniqueness” of the `id` column.

I still like having `id` because it makes it very obvious what developers should use for their id columns. They don’t have to worry about using `ulid` vs. `uuid` vs. something else. We make that decision for them.

> should I just do the changes to throw instead when multiple id modifier are used?

Can we start with this and then revisit it in the future if need be? Think it's a little bit too early to start replacing core modifiers. Would be great to see a bit more usage and input on this.

thetutlage:
Are we not going to throw an error when multiple ids exist? And basically allow them?

olivermrbl:
> Are we not going to throw an error when multiple ids exist?

I think we should throw on multiple `id`s now. Sorry if this was not clear in my previous comment 😄

adrien2p:
@thetutlage @olivermrbl I am currently on it, adding tests 😂

olivermrbl (approved):
LGTM!

PR #8103, opened 2024-07-12
feat(migrations): CLI generate command

adrien2p:
**What**
Provide a new CLI command to allow the migrations to be generated in the specified module(s).

```bash
npx medusa migration generate <modules...>
```

Here is an example of the ran command. 
- Generate the migrations for the first time
- Generate the migration without changes
- Apply changes to model
- Generate the migrations again

Screenshot 2024-07-12 at 11 35 10

FIXES FRMW-2480

### DIFF
--- a/packages/core/modules-sdk/src/loaders/utils/load-internal.ts
+++ b/packages/core/modules-sdk/src/loaders/utils/load-internal.ts
@@ -200,6 +200,7 @@ export async function loadModuleMigrations(
 
       const migrationScriptOptions = {
         moduleName: resolution.definition.key,
+        models: moduleResources.models,
         pathToMigrations: join(moduleResources.normalizedPath, "migrations"),
       }
 
--- a/packages/core/utils/src/modules-sdk/migration-scripts/migration-generate.ts
+++ b/packages/core/utils/src/modules-sdk/migration-scripts/migration-generate.ts
@@ -2,6 +2,7 @@ import { LoaderOptions, Logger, ModulesSdkTypes } from "@medusajs/types"
 import { mikroOrmCreateConnection } from "../../dal"
 import { loadDatabaseConfig } from "../load-module-database-config"
 import { Migrations } from "../../migrations"
+import { toMikroOrmEntities } from "../../dml"
 
 const TERMINAL_SIZE = process.stdout.columns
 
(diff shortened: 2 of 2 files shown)


thetutlage (approved):

packages/medusa/src/commands/migrate.ts
```diff
@@ -1,19 +1,50 @@
...
+      `<modules> cannot be specified with the "${action}" action. Please remove the <modules> argument and try again.`
+    )
+    process.exit(1)
+  }
+
+  if (!modules.length && actionsRequiringModules.includes(action)) {
+    Logger.error(
+      "Please provide the modules for which you want to revert migrations"
+    )
+    Logger.error(`For example: "npx medusa migration revert <moduleName>"`)
```
olivermrbl, line 32: suggestion(non-blocking): Should we change `<moduleName>` to an actual module name from our core suite to make it really obvious what they should do?

olivermrbl (approved):
Nice, great work guys!

@adrien2p, if we remove the `Service` suffix from the module key names, as we've talked about previously, it would be:
```
npx medusa migrations generate FooModule
```

...right?

packages/medusa/src/commands/migrate.ts
thetutlage: We display a list of modules after this example.

My fear with saying `npx medusa migration revert user` is that some might think they always have to run the same exact command.

olivermrbl: Yeah, I see. Makes sense. Fine to keep as is 👍

adrien2p:
> Nice, great work guys!
> 
> @adrien2p, if we remove the `Service` suffix from the module key names, as we've talked about previously, it would be:
> 
> ```
> npx medusa migrations generate FooModule
> ```
> 
> ...right?
> 
> I think ideally the module name passed to the command should map to the key used in the `modules` config in `medusa-config.ts`. Wdyt?

@olivermrbl here it is just that the key is named with the suffix Service, but if it was not then it would have been FooModule. But it is actually the key used in the medusa-config that you have to provide :)

Basically, when the user create a module key it is then used across our utils, in the medusa config, and in the container.resolve. It is this same key used in the command

PR #8107, opened 2024-07-12
feat(core-flows): request return on order exchanges and claims

carlos-r-l-rodrigues:
FIXES: CC-92, CC-87

### DIFF
--- a/packages/core/core-flows/src/order/workflows/claim-request-item-return.ts
+++ b/packages/core/core-flows/src/order/workflows/claim-request-item-return.ts
@@ -26,7 +26,7 @@ import {
 } from "../utils/order-validation"
 
 const validationStep = createStep(
-  "request-item-return-validation",
+  "claim-request-item-return-validation",
   async function ({
     order,
     orderChange,
--- a/packages/core/core-flows/src/order/workflows/exchange-request-item-return.ts
+++ b/packages/core/core-flows/src/order/workflows/exchange-request-item-return.ts
@@ -26,7 +26,7 @@ import {
 } from "../utils/order-validation"
 
 const validationStep = createStep(
-  "request-item-return-validation",
+  "exchange-request-item-return-validation",
   async function ({
     order,
     orderChange,


packages/core/core-flows/src/order/utils/order-validation.ts
```diff
@@ -41,15 +41,14 @@ export function throwIfItemsDoesNotExistsInOrder({
   }
 }
 
-export function throwIfReturnIsCancelled({
-  orderReturn,
-}: {
-  orderReturn: ReturnDTO
-}) {
-  if (orderReturn.canceled_at) {
+export function throwIfIsCancelled(
```
olivermrbl, line 44: neat!

packages/core/core-flows/src/order/workflows/exchange-request-item-return.ts
```diff
@@ -0,0 +1,164 @@
...
+import { updateOrderExchangesStep } from "../steps/update-order-exchanges"
+import {
+  throwIfIsCancelled,
+  throwIfItemsDoesNotExistsInOrder,
+  throwIfOrderChangeIsNotActive,
+  throwIfOrderIsCancelled,
+} from "../utils/order-validation"
+
+const validationStep = createStep(
+  "request-item-return-validation",
```
olivermrbl:
```suggestion
  "exchange-request-item-return-validation",
```

packages/core/core-flows/src/order/workflows/claim-request-item-return.ts
```diff
@@ -0,0 +1,163 @@
...
+import { updateOrderClaimsStep } from "../steps/update-order-claims"
+import {
+  throwIfIsCancelled,
+  throwIfItemsDoesNotExistsInOrder,
+  throwIfOrderChangeIsNotActive,
+  throwIfOrderIsCancelled,
+} from "../utils/order-validation"
+
+const validationStep = createStep(
+  "request-item-return-validation",
```
olivermrbl:
```suggestion
  "claim-request-item-return-validation",
```

packages/core/core-flows/src/order/workflows/exchange-request-item-return.ts
```diff
@@ -0,0 +1,164 @@
...
+        entry_point: "return",
+        fields: ["id", "status", "order_id"],
+        variables: { id: orderExchange.return_id },
+        list: false,
+        throw_if_key_not_found: true,
+      }).config({ name: "return-query" }) as ReturnDTO
+    })
+
+    const createdReturn = when({ orderExchange }, ({ orderExchange }) => {
+      return !orderExchange.return_id
```
olivermrbl, line 79: thought(non-blocking): Don't we want the return to always exist at this point? This is requesting to return an item. It seems a bit off to always create the return if it doesn't exist. Same with claims.

Wdyt?

packages/core/core-flows/src/order/workflows/exchange-request-item-return.ts
```diff
@@ -0,0 +1,164 @@
...
+    validationStep({
+      order,
+      items: input.items,
+      orderExchange,
+      orderReturn,
+      orderChange,
+    })
+
+    when({ orderExchange }, ({ orderExchange }) => {
+      return !orderExchange.return_id
```
olivermrbl, line 120: thought: I guess at this point we are guaranteed to have a return, given that we are creating it earlier no?

olivermrbl: Same with claims

olivermrbl (approved):
Overall LGTM

packages/core/core-flows/src/order/workflows/exchange-request-item-return.ts
carlos-r-l-rodrigues: But at which point we are creating a new request for the exchange or the claim?
I thought this would be created implicitly when requesting the return of an item in the claim or the exchange.
Will it be a separated request just for that?

olivermrbl: I think it will be separate requests. So a full exchange lifecycle would look something like the following:
```
POST /admin/exchanges -> initiate change

POST /admin/exchanges/:id/inbound-items -> request return of item one
POST /admin/exchanges/:id/inbound-items -> request return of item two

POST /admin/exchanges/:id/outbound-items -> request addition of item three
POST /admin/exchanges/:id/outbound-items -> request addition of item four

POST /admin/exchanges/:id/inbound-shipping -> add return shipping for returning items
POST /admin/exchanges/:id/outbound-shipping -> add shipping for new items

POST /admin/exchanges/:id/confirm -> confirm the change request
```

Don't mind the API design here, but the overall flow would be something like this.

Let me know if this makes sense.

olivermrbl: Ahh wait, I think I understand what you mean now. I believe we should create the return as part of initiating the exchange request. So on first request:
```
POST /admin/exchanges -> initiate change -> create exchange + return
```

As you can't really have an exchange without a return right?

PR #8108, opened 2024-07-12
feat: Add exchange return shipping

olivermrbl:


packages/core/core-flows/src/order/workflows/create-exchange-return-shipping-method.ts
```diff
@@ -0,0 +1,145 @@
...
+      }) => {
+        const methodPrice =
+          customPrice ?? shippingOption.calculated_price.calculated_amount
+
+        return {
+          action: ChangeActionType.SHIPPING_ADD,
+          reference: "order_shipping_method",
+          reference_id: methodId,
+          amount: methodPrice,
+          details: {
```
olivermrbl, line 136: @carlos-r-l-rodrigues, I noticed the add shipping action of Returns and Exchanges differ. Just want to make sure this is intentional.

The return shipping of exchanges does not have the `details` property to store the IDs of the different associated resources. Instead the properties are top-level in the action:
https://github.com/medusajs/medusa/blob/f6e86f1b99abf23cb3b39f15763bd5b2c63c5c53/packages/modules/order/src/services/actions/create-exchange.ts#L237-L244

carlos-r-l-rodrigues: yes, this is correct.
I believe the ideal thing would be having only specific input for each action, and transform to this complex internal object under the hood.
But this would increase a bit the scope

olivermrbl: So do I put `order_id`, `return_id`, and `claim_id` in the `details` or outside of it?

Shouldn't we be consistent across all types of order changes?

olivermrbl: Right now, my return shipping order change looks as follows:
```
{
  id: "ordcha_1234",
  details: {
    return_id: "ret_1234",
    order_id: "ord_1234"
  }
  return_id: null,
  order_id: null
}
```

And my exchange return shipping order change as follows:
```
{
  id: "ordcha_1234",
  details: {}
  return_id: "ret_1234",
  order_id: "ord_1234"
  exchage_id: "exch_1234"
}
```

carlos-r-l-rodrigues: you can place it only outside details.
I will revisit the other actions and normalise them all.

carlos-r-l-rodrigues (approved):
@olivermrbl this one is ready to review

integration-tests/modules/__tests__/order/workflows/return/create-exchange-return-shipping.spec.ts
```diff
@@ -0,0 +1,126 @@
...
+          })
+
+          const orderChange = result?.[0]
+
+          expect(orderChange).toEqual(
+            expect.objectContaining({
+              id: expect.any(String),
+              reference: "order_shipping_method",
+              reference_id: expect.any(String),
+              details: {
```
olivermrbl, line 111: question: shouldn't these live top-level?

packages/core/core-flows/src/order/workflows/create-exchange-return-shipping-method.ts
```diff
@@ -0,0 +1,145 @@
...
+    }).config({ name: "fetch-shipping-option" })
+
+    const shippingMethodInput = transform(
+      { orderExchange, shippingOptions },
+      (data) => {
+        const option = data.shippingOptions[0]
+
+        return {
+          shipping_option_id: option.id,
+          amount: option.calculated_price.calculated_amount,
```
olivermrbl: suggestion: should try to use `input.custom_price`

packages/core/core-flows/src/order/workflows/create-exchange-return-shipping-method.ts
olivermrbl, line 136: question: according to our previous discussion, these should live top-level and not in details

olivermrbl (commented):
@carlos-r-l-rodrigues, LGTM, but I can't approve because I'm the author. Will you do that 😄

PR #8112, opened 2024-07-12
feat: Add tax inclusivity management to currenices

sradevski:
CLOSES FRMW-2530

This is the last piece of the tax inclusivity work.


carlos-r-l-rodrigues (approved):
LGTM

packages/modules/pricing/src/services/pricing-module.ts
```diff
@@ -841,6 +838,35 @@ export default class PricingModuleService
...
 
+  @InjectTransactionManager("baseRepository_")
+  protected async createPricePreferences_(
+    data: PricingTypes.CreatePricePreferenceDTO[],
+    @MedusaContext() sharedContext: Context = {}
+  ) {
+    const preferences = await this.pricePreferenceService_.create(
+      data.map((d) => ({
+        ...d,
+        is_tax_inclusive: d.is_tax_inclusive ?? false,
```
carlos-r-l-rodrigues, line 849: any particular reason why we need this default value here?

olivermrbl (approved):
LGTM

packages/modules/pricing/src/services/pricing-module.ts
sradevski: If someone passes `null` mikroORM throws instead of using the default value, so we need to normalize it

carlos-r-l-rodrigues: interesting 👍🏻

PR #8113, opened 2024-07-12
chore(pricing): big number support

carlos-r-l-rodrigues:


olivermrbl (approved):
LGTM

packages/core/types/src/pricing/common/price-set.ts
```diff
@@ -143,6 +145,7 @@ export interface CalculatedPriceSet {
    * The amount of the original price, or `null` if there isn't a calculated price.
    */
   original_amount: BigNumberValue | null
+  raw_original_amount: BigNumberRawValue | null
```
sradevski, line 148: question: If `original_amount` is already a bignumber, do we need the raw field as well? Or do we need to change it to `number` if it's indeed a number?

sradevski (approved):
LGTM, just one question on the calculated prices

packages/core/types/src/pricing/common/price-set.ts
carlos-r-l-rodrigues: the `BigNumberValue` can be many things 😆 
`export type BigNumberValue = BigNumberJS | number | string`

the `BigNumberRawValue` is always the object with at least the property "value" in it as a string.

carlos-r-l-rodrigues: but I am not sure if I understood the question correctly

sradevski: What I meant is, if we know that `original_amount` is always a `number` and not a BigNumberJS/string, should we be more specific with the typings and write it as `number`?

carlos-r-l-rodrigues: Oh I see. This is a good your point.
To be honest, we shouldn't even be doing a `parseFloat` here. It has to use what was returned from the entity, which should be a BigNumber.
So whoever is working with the module directly, will always work with BigNumbers.

I tried to be the less disruptive as possible and at least remove the `parseInt`.
So I just added the regular types we have everywhere.
But I will try to remove the parseFloat in another PR and see what will break.

carlos-r-l-rodrigues: Returning a BigNumber and removing the `JSON.parse(JSON.stringify(calculatedPrices))` from the return of `calculatePrices` will require some rework in multiple places.
Will create a ticket to tackle that later.
thanks for raising that.

sradevski: Ok nice, thanks!

PR #8117, opened 2024-07-13
feat: Admin Returns API

olivermrbl:
Opening to solicit feedback

**What**

Adds
- `POST /admin/returns/:id/request-items`
- `POST /admin/returns/:id/shipping-method`
- `POST /admin/returns/:id/request`


olivermrbl:
The PR needs a lot of clean-up of types, return values, more tests and others. Please keep the review isolated to the flow of events. I'll clean it up as soon as we've got the workflows in place.

Aside from the flow of events, there are two larger elements to also discuss:
- ~~How do we go about return fulfillment?~~
- How do we go about refunds?

~~Additionally, one thought I had was that we should create a dedicated order preview endpoint so we can follow our convention for Rest API response types. The `/admin/returns` should not return order previews but the return object. The same goes for `/admin/claims` and `/admin/exchanges/` that should return claims and exchanges respectively.~~

Let me know what you think.

cc @carlos-r-l-rodrigues

olivermrbl:
> How do we go about return fulfillment?

For this one, I could port over what we have in `createAndCompleteReturn` workflow, such that we always create a return fulfillment as part of submitting the request to create a return. But we'll need to decide if this is where we want it to happen.

packages/core/core-flows/src/order/steps/confirm-order-changes.ts
```diff
@@ -0,0 +1,28 @@
+import { OrderChangeDTO } from "@medusajs/types"
+import { ModuleRegistrationName } from "@medusajs/utils"
+import { createStep, StepResponse } from "@medusajs/workflows-sdk"
+
+type ConfirmOrderChangesInput = {
+  orderId: string
+  changes: OrderChangeDTO[]
+}
+
+export const confirmOrderChanges = createStep(
```
olivermrbl, line 10: comment: Can be "bulkified"

packages/core/core-flows/src/order/steps/create-return-items.ts
```diff
@@ -0,0 +1,55 @@
+import { IOrderModuleService, OrderChangeActionDTO } from "@medusajs/types"
+import { ModuleRegistrationName } from "@medusajs/utils"
+import { createStep, StepResponse } from "@medusajs/workflows-sdk"
+
+type CreateReturnItemsInput = {
+  changes: OrderChangeActionDTO[]
+  returnId: string
+}
+
+export const createReturnItems = createStep(
```
olivermrbl, line 10: comment: Can be "bulkified"

packages/medusa/src/api/admin/returns/route.ts
```diff
@@ -41,11 +41,25 @@ export const POST = async (
   res: MedusaResponse
 ) => {
   const input = req.validatedBody as AdminPostReturnsReqSchemaType
```
olivermrbl, line 43: self-review: Move `location_id` from request items API to this one. We only allow creating a return for a single location at a time

packages/core/core-flows/src/order/workflows/confirm-return-request.ts
```diff
@@ -0,0 +1,92 @@
...
+    throwIfIsCancelled(order, "Order")
+    throwIfIsCancelled(orderReturn, "Return")
+    throwIfOrderChangeIsNotActive({ orderChange: orderChanges })
+  }
+)
+
+export const confirmReturnRequestWorkflowId = "confirm-return-request"
+export const confirmReturnRequestWorkflow = createWorkflow(
+  confirmReturnRequestWorkflowId,
+  function (input: { return_id: string }): WorkflowData<string> {
```
olivermrbl: self-review: Create type + add `no_notification` and ensure it is correctly passed through

packages/core/core-flows/src/order/workflows/confirm-return-request.ts
```diff
@@ -0,0 +1,92 @@
...
+    })
+
+    validationStep({ order, orderReturn, orderChanges })
+
+    createReturnItems({ returnId: orderReturn.id, changes: returnItemActions })
+
+    confirmOrderChanges({ changes: orderChanges, orderId: order.id })
+
+    // @ts-ignore
+    return order.id
```
olivermrbl: self-review: Update return type + value

fPolic (approved):

packages/medusa/src/api/admin/returns/[id]/request-items/route.ts
```diff
@@ -0,0 +1,40 @@
...
+
+export const POST = async (
+  req: AuthenticatedMedusaRequest<AdminPostReturnsRequestItemsReqSchemaType>,
+  res: MedusaResponse
+) => {
+  const { id } = req.params
+
+  const remoteQuery = req.scope.resolve(ContainerRegistrationKeys.REMOTE_QUERY)
+
+  await requestItemReturnWorkflow(req.scope).run({
```
Comment (carlos-r-l-rodrigues) [inline packages/medusa/src/api/admin/returns/[id]/request-items/route.ts:20]: I believe we should return the `result` of this workflow, which will be the preview of the Order after the change actions are applied.

Same for the shipping-method

carlos-r-l-rodrigues (approved):

packages/core/types/src/http/return/admin.ts
```diff
@@ -0,0 +1,54 @@
...
+  no_notification?: boolean
+  refund_amount?: number
+  items: BaseReturnItem[]
+}
+
+export interface AdminInitiateReturnRequest {
+  order_id: string
+  location_id?: string
+  description?: string
+  internal_note?: string
```
fPolic, line 29: discussion: we would need to send this info when we confirm a return request right? When we initiate a return request we don't know this info yet and later we can't update this fields.

EDIT: or we add a general update return request endpoint which would update note, location etc.?

packages/medusa/src/api/admin/returns/[id]/request-items/route.ts
Comment (olivermrbl) [reply packages/medusa/src/api/admin/returns/[id]/request-items/route.ts:20]: Do you think we will need anything but the totals preview in the FE? 

Was just discussing the response with Seb, and we thought of doing:
```
res.status(200).json({ return: orderReturn, total_summary: result.summary })
```

...provided the `result` here is the order preview.

Comment (olivermrbl) [reply packages/medusa/src/api/admin/returns/[id]/request-items/route.ts:20]: Do you think we will need anything but the totals preview in the FE? 

Was just discussing the response with Seb, and we thought of doing:
```
res.status(200).json({ return: orderReturn, total_summary: result.summary })
```

...provided the `result` here is the order preview.

packages/core/types/src/http/return/admin.ts
olivermrbl: Yes, that's right – it will be send with the confirmation request. Also just discussed the return reasons with Seb. We will move these from the return level to the return item level, as we also talked about this morning. This is pending design.

olivermrbl: > we add a general update return request endpoint which would update note, location etc.?

Yeah, that's also possible. We will have the following endpoint:
```
POST /admin/returns/:id 

{ location_id?, no_notification?, metadata? }
```

carlos-r-l-rodrigues: Good point.
And based on the message displayed when the location selected doesn't have stock associated with that location, maybe we need an extra endpoint to update the return location and return the result of this stock location check.

olivermrbl:
@fPolic @carlos-r-l-rodrigues, would it make sense to merge this and address the rest in follow up PRs?

fPolic:
> @fPolic @carlos-r-l-rodrigues, would it make sense to merge this and address the rest in follow up PRs?

sounds good 👍

Issue #8102, opened 2024-07-12
[v1.x] - Sales Channels feature is mandatory when retrieving a Product on the Admin API Route ?


adevinwild:
### Describe the bug
When you disable the `sales_channel` feature flag, you keep getting errors on the `[GET] - /admin/products/:id` route.
It's because that route still tries to look for product's sales channels, I remember that at a time a check was made using the FF router, but now it seems that it's [always there](https://github.com/medusajs/medusa/blob/v1.20.7/packages/medusa/src/api/routes/admin/products/get-product.ts#L125C3-L155C4) :
```ts
// [v1.20.7] - src/api/routes/admin/products/get-product.ts

const shouldSetAvailability =
    req.retrieveConfig.relations?.includes("variants")

  if (shouldSetAvailability) {
    let salesChannels

    if (isMedusaV2FlagOn) {
      const remoteQuery = req.scope.resolve("remoteQuery")
      const query = {
        sales_channel: {
          fields: ["id"],
        },
      }
      salesChannels = await remoteQuery(query)
    } else {
      const salesChannelService: SalesChannelService = req.scope.resolve(
        "salesChannelService"
      )
      ;[salesChannels] = await salesChannelService.listAndCount(
        {},
        { select: ["id"] }
      )
    }

    decoratePromises.push(
      productVariantInventoryService.setProductAvailability(
        [product],
        salesChannels.map((salesChannel) => salesChannel.id)
      )
    )
  }
```

### System information

Medusa version (including plugins): v1.20.7

### Steps to reproduce the behavior

1. Disable the `sales_channel` feature flag in the Medusa config
2. Create a product on the Admin UI
3. Error is throwed in the console :

```bash
driverError: error: relation "public.sales_channel" does not exist
```  

### Expected behavior

I expect a check with the FF router to check if the `sales_channel` feature flag is enabled to make sure we don't have this relation error.

sradevski:
@adevinwild hey thanks for the report! Would you be keen on opening a PR to fix this? Thanks!

adevinwild:
> @adevinwild hey thanks for the report! Would you be keen on opening a PR to fix this? Thanks!

Yes!
Will fix this this weekend 🙌

adevinwild:
Here is the related PR 🙌
https://github.com/medusajs/medusa/pull/8120

PR #8120, opened 2024-07-13
fix: Check if the SalesChannel FF is enabled before using it on the /admin/products/:id route

adevinwild:
# Related issue
This PR is related to this issue https://github.com/medusajs/medusa/issues/8102

## What
This PR fix an issue I had for some times now, the issue in question is the fact that following a product creation in the admin ui having disabled FF sales channels, an error is immediately thrown out on the API route.

## Why
By solving this problem, the API route will now be functional and will no longer throw errors in that specific condition, which will also result in an error-free UI following successive product creation and display/fetching. 

## How
The idea is to add a condition using the `FeatureFlagRouter` to check whether the FF is active before using the service dedicated to the FF. By adding this condition, there are no more errors when retrieving a product having disabled this FF.

### DIFF
--- a/packages/medusa/src/api/routes/admin/products/get-product.ts
+++ b/packages/medusa/src/api/routes/admin/products/get-product.ts
@@ -126,7 +126,10 @@ export default async (req, res) => {
   const shouldSetAvailability =
     req.retrieveConfig.relations?.includes("variants")
 
-  if (shouldSetAvailability) {
+  if (
+    shouldSetAvailability &&
+    featureFlagRouter.isFeatureEnabled(SalesChannelFeatureFlag.key)
+  ) {
     let salesChannels
 
     if (isMedusaV2FlagOn) {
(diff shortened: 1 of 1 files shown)


packages/medusa/src/api/routes/admin/products/get-product.ts
```diff
@@ -136,7 +137,9 @@ export default async (req, res) => {
         },
       }
       salesChannels = await remoteQuery(query)
-    } else {
+    } else if (
+      featureFlagRouter.isFeatureEnabled(SalesChannelFeatureFlag.key)
```
adevinwild: Only in that specific case we can continue and use the `salesChannelService`.

carlos-r-l-rodrigues: should `salesChannels` be initialized as `[]` to `salesChannels.map` at line 155 not fail?

adevinwild: You right, I added a new commit to raise the condition at a higher level
( commit https://github.com/medusajs/medusa/pull/8120/commits/8ef558ec67d7ef26c667242b87fcf0c0a6484656)

carlos-r-l-rodrigues (approved):
LGTM
Thanks @adevinwild

PR #8133, opened 2024-07-15
fix: workflow return type inf missing type

adrien2p:


shahednasser (approved):
Tested and works as expected.

PR #8137, opened 2024-07-15
feat(core-flows,medusa,order): remove return shipping

carlos-r-l-rodrigues:
What:
* `DELETE /admin/returs/:id/shipping_methods/:action_id`

FIXES: CC-187


packages/modules/order/src/services/order-module-service.ts
```diff
@@ -2028,7 +2089,7 @@ export default class OrderModuleService<
 
     if (Object.keys(addedShippingMethods).length > 0) {
       const addedShippingDetails = await this.listShippingMethods(
-        { id: Object.keys(addedShippingMethods) },
+        { id: Object.keys(addedShippingMethods), deleted_at: null },
```
olivermrbl: question: Don't we fetch non-deleted rows by default? If not, should we not?

packages/medusa/src/api/admin/returns/[id]/shipping-method/[action_id]/route.ts
```diff
@@ -0,0 +1,25 @@
...
+    req.scope
+  ).run({
+    input: {
+      return_id: id,
+      action_id,
+    },
+  })
+
+  res.json({
+    order: orderPreview,
```
Comment (olivermrbl) [inline packages/medusa/src/api/admin/returns/[id]/shipping-method/[action_id]/route.ts]: suggestion(non-blocking): Add return to response for consistency

packages/core/core-flows/src/order/workflows/remove-return-shipping-method.ts
```diff
@@ -0,0 +1,101 @@
...
+      list: false,
+      throw_if_key_not_found: true,
+    })
+
+    const orderChange: OrderChangeDTO = useRemoteQueryStep({
+      entry_point: "order_change",
+      fields: ["id", "status", "version", "actions.*"],
+      variables: {
+        filters: {
+          order_id: orderReturn.order_id,
```
olivermrbl, line 71: question: Should we add the `return_id` as a filter here?

integration-tests/http/__tests__/returns/returns.spec.ts
```diff
@@ -233,6 +233,22 @@ medusaIntegrationTestRunner({
           })
         )
 
+        // remove shipping method
+        const action_id = result.data.order.shipping_methods[1].actions[0].id
```
olivermrbl, line 274: question: Should we add an assertion in between the deletion and creation to ensure it is deleted correctly?

olivermrbl (approved):
LGTM, a few questions 👍

packages/modules/order/src/services/order-module-service.ts
carlos-r-l-rodrigues: Yes, this is a left over from tests because in some entities the default filter of deleted entities were conflicting with my custom repository and causing an infinity loop in mikro-orm.
I will remove that.

packages/core/core-flows/src/order/workflows/remove-return-shipping-method.ts
carlos-r-l-rodrigues: yes, that is better.

integration-tests/http/__tests__/returns/returns.spec.ts
carlos-r-l-rodrigues: absolutely, I got distracted fixing wierd bugs when this was added and forgot about getting back to it 😆

packages/medusa/src/api/admin/returns/middlewares.ts
```diff
@@ -65,6 +65,16 @@ export const adminReturnRoutesMiddlewares: MiddlewareRoute[] = [
       ),
     ],
   },
+  {
+    method: ["DELETE"],
+    matcher: "/admin/returns/:id/shipping-method/:action_od",
```
olivermrbl: nit:
```suggestion
    matcher: "/admin/returns/:id/shipping-method/:action_id",
```

Issue #8134, opened 2024-07-15
Pull current remote links using useRemoteQueryStep.


Arrowkuu:
### Describe the bug

I wanted to pull currentLinks using 'useRemoteQueryStep' and there is no such option. More informations in screenshots.

### System information

Medusa version (including plugins): v2
Node.js version: -
Database: -
Operating system: -
Browser (if relevant): -

### Steps to reproduce the behavior

-

### Expected behavior

-

### Screenshots

<img src="https://i.imgur.com/cRvyMRU.png" />
<img src="https://i.imgur.com/Wax63Pw.png" />

### Code snippets

-

### Additional context

-

sradevski:
@Arrowkuu can you please check what the name of the database table is for this relationship, and ensure the link has been defined? It sounds like either the link is not created, or you are using the incorrect name to fetch the data.

adrien2p:
@sradevski I ve asked to open this issue to be linked with the pr 👍

adrien2p:
@Arrowkuu this should probably be available in a next preview, I will close this in the meantime and feel free to let us know if something goes wrong 👍

PR #8140, opened 2024-07-16
chore: Allow to fetch remote link using the service name or the alias from the use remote link step

adrien2p:
**What**
The defineLink now also return the alias that can be used with the remote query or the remote query step. Since the remote query util can accept either an entry point or a service we now provide both values and the step supports both.
For usage of internal link, it is possible to import their definition from the core and use the serviceName to be passed to the service property of the step that is now supported

Solve [this](https://github.com/medusajs/medusa/issues/8134) issue


olivermrbl (approved):
LGTM

PR #8143, opened 2024-07-16
breaking: restructure events payload

thetutlage:
FIXES: FRMW-2587
Related to: #8078 

### Breaking change
This PR changes the event payload accepted by the event listeners.

Here's a diff of an event listener accessing event properties from the `SubscriberArgs` type.

```diff
function productCreateHandler({
- eventName,
- data,
+ event,
  container,
  pluginOptions
}: SubscriberArgs) {
-  console.log(eventName)
+  console.log(event.name)
-  console.log(data.data)
+  console.log(event.data)
-  console.log(data.metadata)
+  console.log(event.metadata)
}
```

Also, the `MessageBody` type has been removed in favor of the `Event` type.

```diff
import {
- MessageBody,
+ Event
} from '@medusajs/types'
```


packages/medusa/src/loaders/helpers/subscribers/index.ts
```diff
@@ -193,10 +192,12 @@ export class SubscriberLoader {
...
     for (const e of events) {
-      const subscriber = async (data: MessageBody<T>) => {
+      const subscriber = async (data: T) => {
         return await handler({
-          eventName: e,
-          data,
+          event: {
+            name: e,
+            data: data,
+          },
```
adrien2p: suggestion: should we spread the data? here we will still have `event.data.data` to access the data, if we spready it we would have everyting flat in event which I think is ok wdyt, it would look like `event.data` instead of `event.data.data` ? It would mean that in the subscriber API event becore Event + name

packages/medusa/src/subscribers/payment-webhook.ts
```diff
@@ -8,14 +8,14 @@ type SerializedBuffer = {
...
-  data,
+  event,
   container,
 }: SubscriberArgs<ProviderWebhookPayload>) {
   const paymentService: IPaymentModuleService = container.resolve(
     ModuleRegistrationName.PAYMENT
   )
 
-  const input = "data" in data ? data.data : data
+  const input = "data" in event ? event.data : event
```
olivermrbl: thought: Shouldn't we be able to get rid of this now?

adrien2p: yes, it could have been removed earlier now thinking about it, we always have data in event now

packages/medusa/src/loaders/helpers/subscribers/index.ts
olivermrbl: We should aim to have:
```ts
export function handler({ event, container, ... }) {
  const { data, name } = event
}
```

where `data` is what is passed as `data` in the `eventBusService.emit` call:
```ts
emit({ data: { test: "1234" }, ... })
```

So:
```
const { data, name } = event
console.log(data) // -> { test: "1234" }
```

Correct me if I am wrong here.

adrien2p: yes that is what I am trying to get to with that suggestion, with the current changes you have to do `{ data: { data: innerData }, name }`

olivermrbl: Cool, sounds good. +1 for the suggestion then 😄

thetutlage: I am little confused here. What does subscriber receive here as an argument? It looks like it receives the entire event

adrien2p: currently, data correspond to the MessageBody (Event now), so when the subscriber wrapper populate the event prop with { name, data } it means that data correspond to Event. So data will contain data, metadata etc. The idea here is to flatten that

thetutlage: @adrien2p Does the recent change looks fine?

packages/medusa/src/subscribers/payment-webhook.ts
thetutlage: Done.

packages/medusa/src/loaders/helpers/subscribers/index.ts
adrien2p: sounds good 👍

adrien2p:
just adding an integration test then we can merge

sradevski (approved):

adrien2p:
I think we can merge 👍

shahednasser:
cc @olivermrbl please merge #8149 once this is merged

PR #8145, opened 2024-07-16
feat: add scheduled jobs options

u11d-bartlomiej-galezowski:
**What** - Added controlling of BullMQ job options.
**Why** - For now developer are not able to control job options such as removeOnComplete etc. In the result frequent scheduled job can fill Redis instance by thousands of hashes containing insignificant informations about completed jobs which can slows done Redis instance.
**How** - Added `options` property to `ScheduledJobConfig` which allows to define BullMQ job options for scheduled jobs.
**Test** - It's not easy to test the feature with unit or integration tests. I've tested it by linking the package to existing Medusa project, appling the job options to `ScheduledJobConfig` in `src/jobs/job.ts` such as below:
```
export const config: ScheduledJobConfig = {
  name: "scheduled-job",
  schedule: "*/10 * * * * *",
  data: {},
  options: {
    removeOnComplete: {
      count: 3,
    },
  },
};
```
and monitoring if `JobSchedulerService:scheduled-jobs:queue:completed` ZSET in Redis contains not more then 3 elements. I've tested it also with other options such as `removeOnFailed` - also works correctly.


sradevski:
Hey @u11d-bartlomiej-galezowski thanks for the contribution! One issue with allowing all possible options is that we leak bullmq details, and any major upgrade to bullmq might mean a breaking change for Medusa as well, which is not something we really want.

Probably we want to start with good defaults, and if we see a strong need for configuration is when we will expose a public API for it.

In this case, I would probably go with something like https://docs.bullmq.io/guide/queues/auto-removal-of-jobs#keep-jobs-based-on-their-age that is hard-coded (eg. 24 hours for successful jobs, 72 hours for failed), and see if there is ever a need to have this handled differently.

u11d-bartlomiej-galezowski:
Hey @sradevski thanks for the answer. We can hard code some defaults or add possibility to modify only `removeOnComplete`, `removeOnFailed` and maybe `backoff`, what do you think?

sradevski:
@u11d-bartlomiej-galezowski let's just add the defaults for now, and we can add the possibility to override them later as needed. It's always easier to go from a more restricted to relaxed API than the other way around 😄

u11d-bartlomiej-galezowski:
Alright, all clear, but I suggest different default values. 24 hours for completed jobs sounds good, but for failed jobs I think better will be one week with e.g. 5000 count, in the result BullMQ will keep failed jobs not older than 1 week, and not more than 5000 items. In the result admin or developer will be able to check the 1 week old errors, but by specify count value we will secure against every second cron job.

sradevski:
@u11d-bartlomiej-galezowski yeah those values sounds good to me, it honestly isn't that big of a deal on what we choose as defaults now, so let's go with what you suggested 👍

u11d-bartlomiej-galezowski:
@sradevski I close this PR since the assumptions are different. I've created the new PR - https://github.com/medusajs/medusa/pull/8156.

PR #8146, opened 2024-07-16
core(core-flows, medusa): remove request item return

carlos-r-l-rodrigues:
What:
* `DELETE /admin/returns/:id/request-items/:action_id`
* add missing `deleted_at` columns

FIXES: CC-188

### DIFF
--- a/integration-tests/modules/__tests__/order/order.spec.ts
+++ b/integration-tests/modules/__tests__/order/order.spec.ts
@@ -205,8 +205,9 @@ medusaIntegrationTestRunner({
               metadata: null,
               created_at: expect.any(String),
               updated_at: expect.any(String),
+              deleted_at: null,
               tax_lines: [],
-              adjustments: [
+              adjustments: expect.arrayContaining([
                 {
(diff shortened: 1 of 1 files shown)


packages/modules/order/src/migrations/Migration20240715174100.ts
```diff
@@ -1,6 +1,6 @@
 import { Migration } from "@mikro-orm/migrations"
 
-export class Migration20240715102100 extends Migration {
+export class Migration20240715174100 extends Migration {
```
olivermrbl, line 3: question: should we revert this change?

olivermrbl (approved):
LGTM!

packages/modules/order/src/migrations/Migration20240715174100.ts
carlos-r-l-rodrigues: I think the migration will read the file name instead of the Class. In the previous commit I didn't change the name and it was the same as the previous migration at that time.
Should be fine to keep it

PR #8151, opened 2024-07-16
chore(medusa,core-flows): update request return items and shipping

carlos-r-l-rodrigues:
What:
* `POST /admin/returns/:id/shipping-methods/:action_id`
* `POST /admin/returns/:id/request-items/:action_id`

CLOSES: CC-184, CC-185


packages/core/types/src/workflow/order/request-item-return.ts
```diff
@@ -9,6 +10,15 @@ export interface DeleteRequestItemReturnWorkflowInput {
   action_id: string
 }
 
+export interface UpdateRequestItemReturnWorkflowInput {
+  return_id: string
+  action_id: string
+  data: {
+    quantity?: BigNumberInput
+    internal_note?: string | null
```
fPolic, line 18: question: should we be able to update return reason here?

fPolic (approved):
LGTM 👍

packages/core/types/src/workflow/order/request-item-return.ts
carlos-r-l-rodrigues: yes, good point.
I have to add it to the request item as well

carlos-r-l-rodrigues: I will them it in a next PR because multiple files are changed...

PR #8152, opened 2024-07-16
fix(core-flow): request item return reason

carlos-r-l-rodrigues:
What:
 - Added support for return reason when requesting item return
 - Endpoints returning `items` and `items.reason` details for the `Return` entity


olivermrbl (approved):
LGTM

PR #8153, opened 2024-07-16
chore(core-flows, medusa): cancel request return

carlos-r-l-rodrigues:
What
* Cancel request return `DELETE /admin/returns/:id/request` 
* Use args.entity name to throw when entity is not found from Remote Joiner if available

CLOSES: CC-186


olivermrbl (approved):
Nice LGTM

PR #8157, opened 2024-07-17
feat: Ensure async workflow executions have access to shared container

sradevski:
REF FRMW-2595

### DIFF
--- a/packages/core/types/src/modules-sdk/index.ts
+++ b/packages/core/types/src/modules-sdk/index.ts
@@ -87,8 +87,8 @@ export type ModuleDefinition = {
   isRequired?: boolean
   isQueryable?: boolean // If the module is queryable via Remote Joiner
   dependencies?: string[]
-  // Only to be used in exceptional cases - relying on the shared container will break the isolation of the module
-  passSharedContainer?: boolean
+  /** @internal only used in exceptional cases - relying on the shared contrainer breaks encapsulation */
+  __passSharedContainer?: boolean
   defaultModuleDeclaration:
     | InternalModuleDeclaration
     | ExternalModuleDeclaration
--- a/packages/medusa/src/loaders/index.ts
+++ b/packages/medusa/src/loaders/index.ts
@@ -154,11 +154,14 @@ export default async ({
   const plugins = getResolvedPlugins(rootDirectory, configModule, true) || []
   const pluginLinks = await resolvePluginsLinks(plugins, container)
 
-  const { onApplicationShutdown, onApplicationPrepareShutdown } =
-    await loadMedusaApp({
-      container,
-      linkModules: pluginLinks,
-    })
+  const {
+    onApplicationStart,
+    onApplicationShutdown,
+    onApplicationPrepareShutdown,
+  } = await loadMedusaApp({
+    container,
+    linkModules: pluginLinks,
+  })
 
   await registerWorkflows(plugins)
   const entrypointsShutdown = await loadEntrypoints(
@@ -168,6 +171,7 @@ export default async ({
     rootDirectory
   )
   await createDefaultsWorkflow(container).run()
+  await onApplicationStart()
 
   const shutdown = async () => {
     const pgConnection = container.resolve(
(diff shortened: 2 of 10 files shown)


packages/medusa/src/loaders/index.ts
```diff
@@ -153,21 +153,21 @@ export default async ({
...
   const pluginLinks = await resolvePluginsLinks(plugins, container)
-  await registerWorkflows(plugins)
 
   const { onApplicationShutdown, onApplicationPrepareShutdown } =
     await loadMedusaApp({
       container,
       linkModules: pluginLinks,
     })
 
+  await registerWorkflows(plugins)
```
olivermrbl, line 166: thought: The workflow engine will attempt to retry failed workflows when the module is booting, so this _might_ cause issues because the workflows are not yet registered. Maybe we can test this?

https://github.com/medusajs/medusa/blob/1d40b3cc98faaaf5c2b5ff9601a16f01d970d95e/packages/modules/workflow-engine-redis/src/utils/workflow-orchestrator-storage.ts#L54-L79

olivermrbl (commented):
Overall LGTM, but I'll leave the approval to Carlos and Adrien. I just had one thought, that I'd like your thoughts on

packages/medusa/src/loaders/index.ts
sradevski: Logically speaking, the modules have to be ready before any workflows are registered and run, so I think it makes sense to register the workflows after the app is loaded. But what you say can probably happen, so let me test it and see if we can ensure workflows are retried only after the app has fully loaded.

The reason I moved it is, if you have a custom workflow with a `schedule` (so a scheduled workflow), the workflow engine has to be loaded before the workflow itself is loaded (since it calls `.schedule` on the workflow engine).

carlos-r-l-rodrigues: To prevent this, maybe the workflow engine can start listening to its queue `onApplicationStart`
That will prevent from start picking jobs before the app is fully loaded

packages/core/types/src/modules-sdk/index.ts
```diff
@@ -87,6 +87,8 @@ export type ModuleDefinition = {
   isRequired?: boolean
   isQueryable?: boolean // If the module is queryable via Remote Joiner
   dependencies?: string[]
+  // Only to be used in exceptional cases - relying on the shared container will break the isolation of the module
+  passSharedContainer?: boolean
```
carlos-r-l-rodrigues: I would suggest using a name that seems more like an "internal" prop. eg: `__passAppContainer` and comment it with `/** @internal */`

carlos-r-l-rodrigues: LGTM

would be ideal to start listening to jobs `onApplicationStart`

carlos-r-l-rodrigues (approved):

packages/medusa/src/loaders/index.ts
olivermrbl: +1 – that sounds like a more appropriate approach 👍

sradevski: @carlos-r-l-rodrigues @olivermrbl I was doing exactly that, see the second commit 😄

packages/core/types/src/modules-sdk/index.ts
sradevski: Sounds good 👍

adrien2p (commented):
Sorry for the delay, just had time to review it now. It looks very good to me, just one thing, can we now omit or make optional the container to be provided when calling the run from the workflow engine public api?

adrien2p:
Cc @shahednasser we are not obliged anymore to provide the container when using the workflow engine, which is a first step to later not have to pass the container to workflow either (still need some work such as using the workflow engine if provided to run the workflows under the hood)

shahednasser:
@adrien2p got it, I think once we have the second step (not having to pass the container to a workflow) we can update the docs to just be `helloWorkflow.run` rather than `helloWorkflow(container).run`

sradevski:
@adrien2p https://github.com/medusajs/medusa/pull/8165

PR #8162, opened 2024-07-17
breaking: rework how links database migrations are managed

adrien2p:
FIXES: FRMW-2588

Breaking: rework how links database schema is managed

This PR updates the process/flow to keep the database in sync with the links defined by the user application or the installed packages.

Earlier, we used to `migrate` and `revert` the database for the defined links. We move away from these concepts and instead perform a `sync`. Here's the rationale.

- The links are managed in the code, and we do not expect you to define migrations for them.
- Therefore, the new sync engine will look at your app's current set of links and compute the database state from that.
- The sync engine will create the needed database tables if you define a new link.
- If you remove a link, the sync engine will drop the database table created earlier.
- If you modify a link, the sync engine will alter the database table (which might lead to data loss in some instances).

To conclude, the sync engine puts the database into its final state based on the links defined in your codebase. It keeps track of links inside the `link_module_migrations` table so that it can detect link removals and alterations as well.

## Introducing the `links sync` command
The `npx medusa links sync` command runs the sync engine and updates the database with the links defined in your codebase. The following are some screenshots from the `links sync` command.

### When new tables are created

In the following example, the sync engine detects the database tables that must be created for the links. It creates those tables and displays a log of them.

tables_created

### When existing tables are updated

In the following example, the sync engine detects a change in some existing links and decides to alter the database tables for them. Since these alter commands are safe to execute (no data loss will occur), the sync engine will execute the SQL commands immediately.

tables_updated_with_sql_queries

### When existing tables are updated (but might lead to data loss)

In the following example, the sync engine detects a change in some existing links, which alters the database tables. However, this change might lead to some data loss, so we display a prompt for the user to manually approve the changes.

constent_to_update_the_tables

### When existing tables are deleted

In the following example, the sync engine detects that a pre-existing link has been removed from the codebase and offers to do the database cleanup. A prompt is displayed if the user wants to preserve the table to prevent data loss.

constent_with_deleting


thetutlage:
I see we have this `import { MigrationsExecutionPlanner } from "@medusajs/link-modules"` import inside the `modules-sdk`. I believe, the modules-sdk should not rely on the link module, because it feels like a core sdk is rely on high-level modules as a dependency. Maybe, I am wrong :). WDYT?

packages/modules/link-modules/integration-tests/__tests__/migrations.spec.ts
```diff
@@ -0,0 +1,152 @@
...
+        expect(actionPlan[0]).toEqual({
+          action: "create",
+          linkDescriptor: {
+            fromModule: "User",
+            toModule: "Car",
+            fromModel: "user",
+            toModel: "car",
+          },
+          tableName: "User_user_Car_car",
+          sql: 'create table "User_user_Car_car" ("user_id" varchar(255) not null, "car_id" varchar(255) not null, "id" varchar(255) not null, "created_at" timestamptz(0) not null default CURRENT_TIMESTAMP, "updated_at" timestamptz(0) not null default CURRENT_TIMESTAMP, "deleted_at" timestamptz(0) null, constraint "User_user_Car_car_pkey" primary key ("user_id", "car_id"));\ncreate index "IDX_car_id_-8c9667b4" on "User_user_Car_car" ("car_id");\ncreate index "IDX_id_-8c9667b4" on "User_user_Car_car" ("id");\ncreate index "IDX_user_id_-8c9667b4" on "User_user_Car_car" ("user_id");\ncreate index "IDX_deleted_at_-8c9667b4" on "User_user_Car_car" ("deleted_at");\n\n',
```
adrien2p, line 59: we might need to review the index name generation @carlos-r-l-rodrigues 😂

carlos-r-l-rodrigues: Yeah, it's using the `simpleHash` of the table name.
we have to be aware of the limit of 63 chars to name stuff in Postgres. Maybe place some todos to manage that nicely later

carlos-r-l-rodrigues: if people use very long names for their modules + entities this will cause issues at the moment,

adrien2p: yes for sure, nothing to do now though

olivermrbl (approved):
Overall LGTM, great work @thetutlage @adrien2p 💪 

Do we need to include anything actional in the release notes about this change?

thetutlage:
> Do we need to include anything actional in the release notes about this change? If not, I'll copy over the PR description as is. Think it's good.

No additional actions are needed. We infact account for the use-cases where someone will run this command on an existing project. So we should be all good to go

olivermrbl:
> No additional actions are needed. We infact account for the use-cases where someone will run this command on an existing project. So we should be all good to go

Love it!

PR #8165, opened 2024-07-17
fix: Remove the option to pass a container to workflow engine API

sradevski:


carlos-r-l-rodrigues:
The `container` is optional `?` already. Probably we can keep this no?

sradevski:
> The `container` is optional `?` already. Probably we can keep this no?

@carlos-r-l-rodrigues We chatted with Adrien and said that it's probably better to just remove it until we have a need for it, and add it then (it wasn't used anywhere anyway). Let me know if you think differently

carlos-r-l-rodrigues (approved):

carlos-r-l-rodrigues:
That doesn't change much at the moment. I just prefer to have available options in place, but if needed we just add it back.

PR #8169, opened 2024-07-17
feat: Add support for private files to file module

sradevski:
CLOSES CC-213

### DIFF
--- a/packages/modules/providers/file-s3/integration-tests/__tests__/services.spec.ts
+++ b/packages/modules/providers/file-s3/integration-tests/__tests__/services.spec.ts
@@ -35,45 +35,50 @@ describe.skip("S3 File Plugin", () => {
       }
     )
   })
+  ;(["public", "private"] as const).forEach((access) => {
+    it("uploads, reads, and then deletes a file successfully", async () => {
+      const fileContent = await fs.readFile(fixtureImagePath)
+      const fixtureAsBinary = fileContent.toString("binary")
 
-  it("uploads, reads, and then deletes a file successfully", async () => {
-    const fileContent = await fs.readFile(fixtureImagePath)
-    const fixtureAsBinary = fileContent.toString("binary")
+      const resp = await s3Service.upload({
+        filename: "catphoto.jpg",
+        mimeType: "image/jpeg",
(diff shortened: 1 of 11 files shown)


packages/modules/providers/file-s3/integration-tests/__tests__/services.spec.ts
```diff
@@ -35,45 +35,50 @@ describe.skip("S3 File Plugin", () => {
...
+      expect(signedUrlFile.toString("binary")).toEqual(fixtureAsBinary)
 
-    await s3Service.delete({ fileKey: resp.key })
+      await s3Service.delete({ fileKey: resp.key })
 
-    // TODO: Currently the presignedURL will be returned even if the file doesn't exist. Should we check for existence first?
-    const deletedFileUrl = await s3Service.getPresignedDownloadUrl({
-      fileKey: resp.key,
-    })
+      // TODO: Currently the presignedURL will be returned even if the file doesn't exist. Should we check for existence first?
```
olivermrbl, line 72: It sounds reasonable to check for existence in `getPresignedDownloadUrl` and throw `File not found` or similar, if that is the case. Wdyt?

olivermrbl (approved):
LGTM

packages/modules/providers/file-s3/integration-tests/__tests__/services.spec.ts
sradevski: @olivermrbl Yeah I think it makes sense, the only downside is that it requires one more request, so I was a bit unsure if we want to make the 99% of the requests slower for better error message.

Now that I think about it though, presigned URLs are only valid for private files, which will be a lot less common, so maybe we can add the check. WDYT?

Btw I'll just merge this PR as it is unrelated, and depending on what we decide I'll open a new one.

olivermrbl: Yeah, sounds good to me. No need to rush with the follow-up PR though. Can be tackled later.

sradevski: I'll just open a ticket for now

PR #8170, opened 2024-07-17
fix: product module joiner config

adrien2p:
**what**
Fix product module joiner config missing suffix

FIXES FRMW-2600


sradevski (approved):

adrien2p:
Not to be merged, the method suffix is computed under the hood automatically. Ill rework this pr to provide the true fox

adrien2p:
@shahednasser you can test the next snapshot 👍 
Just to let you know, the field alias being used for the remote query is the snake case version of the entity name (singular/plural depending on isList). Meaning that in your scenario, you want to add the following field: `product_variant.*`. don't know if that part is documented

shahednasser:
@adrien2p I tried that but got an error that `product_variant` property doesn't exist. Would this be fixed with this change?

adrien2p:
> @adrien2p I tried that but got an error that `product_variant` property doesn't exist. Would this be fixed with this change?

yes it will 👍 keep me in touch, lets generate a snap

olivermrbl (approved):

shahednasser (approved):
LGTM issue fixed 🎉

PR #8171, opened 2024-07-17
chore(medusa,core-flows): begin and cancel receive return

carlos-r-l-rodrigues:
What:
* `POST /admin/returns/:id/receive` 
* `DELETE /admin/returns/:id/receive`

CLOSES: CC-189

### DIFF
--- a/packages/medusa/src/api/admin/returns/[id]/receive/route.ts
+++ b/packages/medusa/src/api/admin/returns/[id]/receive/route.ts
@@ -4,6 +4,8 @@ import {
 } from "@medusajs/core-flows"
 import {
   ContainerRegistrationKeys,
+  ModuleRegistrationName,
+  promiseAll,
   remoteQueryObjectFromString,
 } from "@medusajs/utils"
 import {
@@ -17,6 +19,7 @@ export const POST = async (
   res: MedusaResponse
 ) => {
   const remoteQuery = req.scope.resolve(ContainerRegistrationKeys.REMOTE_QUERY)
+  const orderModuleService = req.scope.resolve(ModuleRegistrationName.ORDER)
 
   const { id } = req.params
 
@@ -39,11 +42,14 @@ export const POST = async (
     fields: req.remoteQueryConfig.fields,
   })
 
-  const [orderReturn] = await remoteQuery(queryObject)
+  const [order, orderReturn] = await promiseAll([
+    orderModuleService.retrieveOrder(result.order_id),
+    remoteQuery(queryObject),
+  ])
 
   res.json({
-    order_change: result,
-    return: orderReturn,
+    order,
+    return: orderReturn[0],
   })
 }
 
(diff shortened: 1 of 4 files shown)


packages/core/core-flows/src/order/workflows/return/begin-receive-return.ts
```diff
@@ -0,0 +1,73 @@
...
+        return {
+          change_type: "return" as const,
+          order_id: order.id,
+          return_id: orderReturn.id,
+          created_by: input.created_by,
+          description: input.description,
+          internal_note: input.internal_note,
+        }
+      }
+    )
```
olivermrbl, line 70: question: Can we tell if this change is for receiving the return? It doesn't look different from the begin return request action, but maybe this is intentional, so just asking out of curiosity.

packages/core/core-flows/src/order/workflows/return/cancel-receive-return.ts
```diff
@@ -0,0 +1,69 @@
...
+    }).config({ name: "order-query" })
+
+    const orderChange: OrderChangeDTO = useRemoteQueryStep({
+      entry_point: "order_change",
+      fields: ["id", "status", "version"],
+      variables: {
+        filters: {
+          order_id: orderReturn.order_id,
+          return_id: orderReturn.id,
+          status: [OrderChangeStatus.PENDING, OrderChangeStatus.REQUESTED],
```
olivermrbl, line 59: question: Can a receivable return be pending?

packages/medusa/src/api/admin/returns/[id]/receive/route.ts
```diff
@@ -9,12 +16,52 @@ export const POST = async (
...
+      },
+    },
+    fields: req.remoteQueryConfig.fields,
   })
 
-  res.status(200).json({ return: result })
+  const [orderReturn] = await remoteQuery(queryObject)
+
+  res.json({
+    order_change: result,
```
Comment (olivermrbl) [inline packages/medusa/src/api/admin/returns/[id]/receive/route.ts]: thought: Would it make sense to return the order + return?

olivermrbl (approved):
Overall LGTM, just had a few questions 💪

packages/core/core-flows/src/order/workflows/return/begin-receive-return.ts
carlos-r-l-rodrigues: This type is generic and it is just covering "return", "claim", "exchange" and "edit". It is indeed the same as begin return request, the difference is that the return already exist.

packages/core/core-flows/src/order/workflows/return/cancel-receive-return.ts
carlos-r-l-rodrigues: This is only the status of the OrderChange entity. Not the return itself.
This status is pending while the changes are being done in the UI and they are not yet confirmed.

packages/medusa/src/api/admin/returns/[id]/receive/route.ts
Comment (carlos-r-l-rodrigues) [reply packages/medusa/src/api/admin/returns/[id]/receive/route.ts]: yeah, we can do it.
I will change it

PR #8176, opened 2024-07-18
chore: update ModuleRegistrationName imports to @medusajs/utils across docs

marianocordoba:
### What

Update some wrong imports of `ModuleRegistrationName` from `@medusajs/modules-sdk` to `@medusajs/utils`


carlos-r-l-rodrigues:
Thanks @marianocordoba.
That was part of some cleanup tasks we had and it was done.

PR #8178, opened 2024-07-18
feat: Add product export endpoint and a dummy workflow

sradevski:
CLOSES CC-222


integration-tests/http/__tests__/product/admin/product-export.spec.ts
```diff
@@ -0,0 +1,396 @@
...
+            tags: [{ value: "new-tag" }],
+            type_id: baseType.id,
+          }),
+          adminHeaders
+        )
+      ).data.product
+    })
+
+    describe("POST /admin/products/export", () => {
+      it("should export a csv file containing the expected products", async () => {
```
sradevski, line 74: The tests will be enabled once the workflow is implemented (follow-up PRs)

olivermrbl (approved):

integration-tests/http/__tests__/product/admin/product-export.spec.ts
```diff
@@ -0,0 +1,398 @@
...
+          "/admin/products/export",
+          {},
+          adminHeaders
+        )
+
+        const workflowId = batchJobRes.data.workflow_id
+        expect(workflowId).toBeTruthy()
+
+        // Pull to check the status until it is completed
+        while (true) {
```
carlos-r-l-rodrigues, line 86: General comment from someone that is not aware of all the underlying mechanism and not only valid for the test here:
This looks a good use case for the workflow engine and the async workflow triggers the action and we subscribe to that transaction to handle when it is done.
Even to push these events to the UI would be great.

sradevski: Yeah agreed, I just copy-pasted the tests from the original batch tests, and I will be changing them as I am implementing the exports. I'll try using what you suggested instead of the while loop 👍

PR #8180, opened 2024-07-18
docs: fix module provider definition export

shahednasser:
Fix the module definition of provider modules to export an object with a `services` property.


olivermrbl (approved):
LGTM, perhaps we should consider adding a `ModuleProvider` export or see if we can reuse the existing `Module` export @adrien2p

adrien2p:
> LGTM, perhaps we should consider adding a `ModuleProvider` export or see if we can reuse the existing `Module` export @adrien2p

We cant reuse the same one as they do not export the same thing :) I add a similar issue locally because of fail build but yes we could have an ModuleProvider same as we have Module except that it would be lighter but typed

olivermrbl:
> We cant reuse the same one as they do not export the same thing

What I was thinking was we could introduce a `type` option or similar to `Module`. But I can see the benefit in having a separate util – it makes the usage more obvious for the two cases.

PR #8188, opened 2024-07-18
chore(order): simplify order engine

carlos-r-l-rodrigues:
what:
- remove unused features of order changes calculation engine

### DIFF
--- a/packages/modules/order/src/services/__tests__/util/actions/exchanges.ts
+++ b/packages/modules/order/src/services/__tests__/util/actions/exchanges.ts
@@ -103,7 +103,7 @@ describe("Order Exchange - Actions", function () {
       transaction_total: 0,
       original_order_total: 270,
       current_order_total: 312.5,
-      temporary_difference: -20,
+      temporary_difference: 62.5,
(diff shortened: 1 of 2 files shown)


packages/modules/order/src/services/__tests__/util/actions/exchanges.ts
```diff
@@ -103,9 +103,7 @@ describe("Order Exchange - Actions", function () {
       transaction_total: 0,
       original_order_total: 270,
       current_order_total: 312.5,
-      temporary_difference: 62.5,
-      future_difference: 0,
-      future_temporary_difference: 0,
+      temporary_difference: -20,
```
olivermrbl: question: What happened here? 😄

olivermrbl (approved):
Nice 💪

packages/modules/order/src/services/__tests__/util/actions/exchanges.ts
carlos-r-l-rodrigues: thanks for asking. I removed something and checked the test incorrectly.

PR #8192, opened 2024-07-19
chore: default isQueryable module

adrien2p:
**What**
Make module default isQueryable from the medusa loader as it is the one that decide to make them defaulted by default

FIXES FRMW-2602


adrien2p:
@carlos-r-l-rodrigues wdyt as a first step, I have delegated this to the medusa loader layer so that it stays opininated at load time

adrien2p:
Cc @olivermrbl @shahednasser

shahednasser:
@adrien2p so there's no need to pass `isQueryable` once this is merged, correct?

carlos-r-l-rodrigues:
It looks good. If you can `.catch` the `__joinerConfig.?()` too it would be ideal for external modules where the __joinerConfig will exist but the external call can fail. Then we can display the same message.

adrien2p:
> @adrien2p so there's no need to pass `isQueryable` once this is merged, correct?

yes that's correct

adrien2p:
@carlos-r-l-rodrigues like the last commit?

olivermrbl:
This is super nice!

PR #8194, opened 2024-07-19
fix: normalize path before consuming it while loading models

adrien2p:
**What**
Normalize path before consuming them while auto loading the models to build the joiner config. Makes it run on any os.

Also, now as part of the Module util usage, if no joiner config are provided, the custom one built on the fly will only include the config based on the models used in the medusa service just like we do for the linkable in order to prevent discrepency


adrien2p:
cc @shahednasser

adrien2p:
<strike>@olivermrbl wdyt if instead of loading models magically we now rely on the ones stored on the medusa service when run in the context of the Module util? It wouldn't change the magical discovery of the defineJoinerConfig when consume inside a module which can still auto discover the models. Ill wait for your response before acting on anything </strike>

olivermrbl (approved):

PR #8195, opened 2024-07-19
fix: workflow 'when' should return T or undefined instead of Partial<T>

adrien2p:
**What**
Fix `when` type to return T or undefined instead of partial T


shahednasser (approved):
LGTM

PR #8204, opened 2024-07-20
chore(inventory, core-flows): big number support

carlos-r-l-rodrigues:
What:
- Inventory: big number calculations
- Small fixes with deleted_at
- Small fixes on flows with shipment data and labels

### DIFF
--- a/integration-tests/modules/__tests__/order/workflows/create-shipment.spec.ts
+++ b/integration-tests/modules/__tests__/order/workflows/create-shipment.spec.ts
@@ -418,7 +418,7 @@ medusaIntegrationTestRunner({
           inventoryItem.id,
           [location.id]
         )
-        expect(stockAvailability).toEqual(1)
+        expect(stockAvailability.valueOf()).toEqual(1)
       })
     })
   },


sradevski (approved):

packages/modules/inventory-next/src/migrations/Migration20240719123015.ts
```diff
@@ -0,0 +1,73 @@
+import { Migration } from "@mikro-orm/migrations"
+
+export class Migration20240719123015 extends Migration {
+  async up(): Promise<void> {
```
adrien2p, line 4: Q: when in our modules we manage the migrations, shouldn't we get the snapshot updated as well? such that if another person update the schema and run the command to generate the migration we don't end up with massive SQL to remove/update/create a lot things which was not present in the snap?

adrien2p (approved):

packages/modules/inventory-next/src/migrations/Migration20240719123015.ts
carlos-r-l-rodrigues: yes, I believe there are many out of sync. 
we should also generate a single migration before releasing.

PR #8214, opened 2024-07-22
feat: Add support for exporting products in backend

sradevski:
CLOSES CC-221
CLOSES CC-223
CLOSES CC-224


packages/modules/providers/file-local/src/services/local-file.ts
```diff
@@ -38,9 +38,9 @@ export class LocalFileService extends AbstractFileProviderService {
 
     const fileKey = path.join(
       parsedFilename.dir,
-      // We append "private" to the file key so deletions and presigned URLs can know which folder to look into
```
sradevski, line 41: This was messing up the extensions of the file, so I made it into a prefix.

sradevski:
Example export file:

[exported-products.csv](https://github.com/user-attachments/files/16330956/exported-products.csv)

Let me know if you think we need to do some tweaks to it cc @srindom @olivermrbl

olivermrbl:
> Example export file:
> 
> [exported-products.csv](https://github.com/user-attachments/files/16330956/exported-products.csv)
> 
> Let me know if you think we need to do some tweaks to it cc @srindom @olivermrbl

Overall, I think this looks good. 

Two immediate thoughts:
- Are we dealing with both region and currency prices? If so, what would a region price look like? Is it `Variant Price <region name>`?
- Do we want to include inventory quantities in the export?

sradevski:
> > Example export file:
> > [exported-products.csv](https://github.com/user-attachments/files/16330956/exported-products.csv)
> > Let me know if you think we need to do some tweaks to it cc @srindom @olivermrbl
> 
> Overall, I think this looks good.
> 
> Two immediate thoughts:
> 
> * Are we dealing with both region and currency prices? If so, what would a region price look like? Is it `Variant Price <region name>`?
> * Do we want to include inventory quantities in the export?

@olivermrbl Currently it's the region ID, but we can replace it with the name (I just thought it will make it easier to programatically parse it if we use the ID).

In terms of inventory, I didn't focus too much on the exact fields we render, since all of that logic is isolated in the `normalizeForExport` function, so we can tweak it as we see fit. We can easily add/remove fields as needed, I just wanted to get everything working first. Let me know if you think I should focus on including inventory as well now, or if you want to create a new ticket with the exact format of it and do it in a separate PR

srindom:
Think it looks good! Noticed that the full url to the product images aren't there. Think we should include that.

olivermrbl:
>  Let me know if you think I should focus on including inventory as well now, or if you want to create a new ticket with the exact format of it and do it in a separate PR

Fine to keep this for a separate follow-up PR 👍

sradevski:
> Think it looks good! Noticed that the full url to the product images aren't there. Think we should include that.

@srindom That's just the test data, otherwise it will be the full URL :)

srindom:
> That's just the test data, otherwise it will be the full URL :)

Got it! In that case LGTM ✅

packages/modules/notification/integration-tests/__tests__/notification-module-service/index.spec.ts
```diff
@@ -67,6 +83,29 @@ moduleIntegrationTestRunner({
...
+        const notification = {
+          to: "admin@medusa.com",
+          template: "some-template",
+          channel: "email",
+          data: {},
+        }
+
+        const result = await service.createNotifications(notification)
+
+        console.log(eventBusEmitSpy.mock.calls)
```
olivermrbl:
```suggestion
```

packages/core/core-flows/src/product/workflows/export-products.ts
```diff
@@ -1,18 +1,49 @@
...
   exportProductsWorkflowId,
   (
     input: WorkflowData<WorkflowTypes.ProductWorkflow.ExportProductsDTO>
   ): WorkflowData<void> => {
-    const products = useRemoteQueryStep({
-      entry_point: "product",
-      fields: input.select,
-      variables: input.filter,
-      list: true,
+    const products = getAllProductsStep(input)
```
olivermrbl: question: Not sure if this is very relevant now, but did you consider a long-running workflow for the export?

olivermrbl (approved):
LGTM!

packages/core/core-flows/src/product/workflows/export-products.ts
sradevski: Ah yes indeed, I already was thinking about it in that way (which is why I was waiting for the notification in the tests), but didn't make it to run in the background. Let me change it.

sradevski: @olivermrbl done!

PR #8215, opened 2024-07-22
fix(create-medusa-app): run links sync during setup

shahednasser:
Run `links sync` command along with running migrations

### DIFF
--- a/.github/workflows/test-cli-with-database.yml
+++ b/.github/workflows/test-cli-with-database.yml
@@ -60,7 +60,7 @@ jobs:
         working-directory: ../cli-test
 
       - name: Run migrations
-        run: medusa migrations run
+        run: medusa migrations run && medusa links sync
         working-directory: ../cli-test
 
       - name: Create admin user


olivermrbl (approved):

packages/cli/create-medusa-app/src/utils/prepare-project.ts
```diff
@@ -138,7 +138,7 @@ export default async ({
     await processManager.runProcess({
       process: async () => {
         const proc = await execute(
-          ["npx medusa migrations run", npxOptions],
+          ["npx medusa migrations run && npx medusa links sync", npxOptions],
```
adrien2p, line 141: Q: is there a reason why there is many if with the exact same condition? :D also, should we display a new box saying that we are syncing the links?

adrien2p (approved):

packages/cli/create-medusa-app/src/utils/prepare-project.ts
shahednasser: > is there a reason why there is many if with the exact same condition? :D

We had other conditions related to seeding in the condition after it, but then we removed them as it no longer applied with the onboarding, so the conditions became the same now 😄 I'll open another PR to refactor.

> should we display a new box saying that we are syncing the links?

I'm not sure that's necessary really, but I'll look into it as part of the refactor PR 👍🏻

adrien2p: should we make it concurrent @shahednasser ?

PR #8221, opened 2024-07-22
chore(framework): Initial commit

adrien2p:
**What**
- Initiate the framework package (which is just a place to move things around for now)
- move the config loader and related resources as well as the `ConfigModule` type
- Create a ConfigManager singleton which prepare and store the config (later can be stored entirely in the container) and allow for easier test override
- re export the logger from the framework
- replace medusa config loader with the framework one
- `build` run type check on tests as well but `prepublishOnly` will not fail on build if tests are typed broken

FIXES FRMW-2607
FIXES FRMW-2609
FIXES FRMW-2614
FIXES FRMW-2618


adrien2p:
This pr include experimental test of `nodenext` resolution to use sub path exports/imports. If the test fail and create more issues then we will have to post pone the usage of sub paths and rely on classic approach

integration-tests/modules/package.json
```diff
@@ -5,8 +5,8 @@
   "license": "MIT",
   "private": true,
   "scripts": {
-    "test:integration": "jest --silent=false --no-cache --maxWorkers=50% --bail --detectOpenHandles --forceExit --logHeapUsage",
-    "test:integration:chunk": "jest --silent --no-cache --bail --maxWorkers=50% --forceExit --testPathPattern=$(echo $CHUNKS | jq -r \".[${CHUNK}] | .[]\")",
+    "test:integration": "NODE_OPTIONS=--experimental-vm-modules jest --silent=false --no-cache --maxWorkers=50% --bail --detectOpenHandles --forceExit --logHeapUsage",
```
thetutlage, line 8: Just curious about the role of `experimental-vm-modules` here

thetutlage:
So far it looks good to me!

integration-tests/modules/package.json
adrien2p: Jest complaining about esm modules https://jestjs.io/docs/ecmascript-modules

packages/framework/framework/src/config/config.ts
```diff
@@ -0,0 +1,157 @@
+import { ConfigModule } from "./types"
+import { deepCopy, isDefined } from "@medusajs/utils"
+import { Logger } from "@medusajs/types"
+
+export class ConfigManager {
+  /**
+   * A flag to secify if we are in product or not, determine weither an error would be critical and thrown or just logged as a warning in developement
```
olivermrbl:
```suggestion
   * A flag to secify if we are in production or not, determine whether an error would be critical and thrown or just logged as a warning in developement
```

packages/framework/framework/src/config/config.ts
```diff
@@ -0,0 +1,157 @@
...
+  ): ConfigModule["projectConfig"] {
+    const outputConfig: ConfigModule["projectConfig"] = deepCopy(projectConfig)
+
+    if (!outputConfig?.redisUrl) {
+      console.log(`redisUrl not found. A fake redis instance will be used.`)
+    }
+
+    outputConfig.http = this.buildHttpConfig(projectConfig)
+
+    let workedMode = outputConfig?.workerMode!
```
olivermrbl: nit:
```suggestion
    let workerMode = outputConfig?.workerMode!
```

olivermrbl (approved):
LGTM, nice work!

PR #8222, opened 2024-07-22
chore(core-flows): adjust inventory when return is received

carlos-r-l-rodrigues:
What:
* Adjust inventory when receiving item return
* Fix stock check adding to cart with updated quantity
* Fix Order module indexes


integration-tests/http/__tests__/returns/returns.spec.ts
```diff
@@ -14,11 +19,30 @@ medusaIntegrationTestRunner({
...
     beforeEach(async () => {
       const container = getContainer()
       await createAdminUser(dbConnection, adminHeaders, container)
 
+      const inventoryModule = container.resolve(
+        ModuleRegistrationName.INVENTORY
+      )
+      const productModule = container.resolve(ModuleRegistrationName.PRODUCT)
+
+      const [product] = await productModule.createProducts([
```
olivermrbl: nit(non-blocking): Use `POST /admin/products`

integration-tests/http/__tests__/returns/returns.spec.ts
```diff
@@ -123,6 +148,50 @@ medusaIntegrationTestRunner({
         )
       ).data.fulfillment_set
 
+      inventoryItem = await inventoryModule.createInventoryItems({
```
olivermrbl: nit(non-blocking): Use API endpoints

integration-tests/modules/__tests__/order/workflows/create-fulfillment.spec.ts
```diff
@@ -351,6 +352,9 @@ medusaIntegrationTestRunner({
         )
 
         const order = await createOrderFixture({ container, product, location })
+        const itemWithInventory = order.items!.find(
+          (o) => o.variant_sku === "test-variant"
```
olivermrbl: nit(non-blocking): Extract "test-variant" to const

packages/modules/order/src/migrations/Migration20240604100512.ts
```diff
@@ -32,17 +32,17 @@ export class Migration20240604100512 extends Migration {
     CREATE INDEX IF NOT EXISTS "IDX_order_transaction_return_id" ON "order_transaction" (
```
olivermrbl, line 32: question: nice catch, should the changes to indexes come in a separate migration?

packages/core/core-flows/src/order/workflows/return/confirm-receive-return-request.ts
```diff
@@ -40,13 +48,95 @@ const validationStep = createStep(
...
+          variant_id: inventory_items.variant_id,
+          inventory_item_id: inventoryItemId,
+          required_quantity: inventory_items.required_quantity,
+        })
+      }
+    }
+  )
+
+  if (hasManagedInventory && !hasStockLocation) {
+    // TODO: how to bypass this error?
```
olivermrbl: question: This looks OK at first glance. What were your thoughts here?

packages/core/core-flows/src/order/workflows/return/confirm-receive-return-request.ts
```diff
@@ -40,13 +48,95 @@ const validationStep = createStep(
   }
 )
 
+function prepareInventoryUpdate({ orderReturn, returnedQuantityMap }) {
```
olivermrbl, line 52: nit: Can I get you to add a comment describing the logic in this function? I think it's a bit hard to follow, so some context might help for the next reader.

packages/core/core-flows/src/order/workflows/return/confirm-receive-return-request.ts
```diff
@@ -81,45 +171,84 @@ export const confirmReturnReceiveWorkflow = createWorkflow(
...
+              )?.item?.variant_id
 
-        itemUpdates[itemId] = {
-          id: itemMap[itemId],
-          received_quantity: act.details!.quantity,
+              if (!variantId) {
+                return
+              }
+
+              const cur = returnedQuantityMap[variantId] ?? 0
```
olivermrbl: nit(non-blocking): Rename `cur` to something more sensible

packages/modules/order/src/migrations/Migration20240219102530.ts
```diff
@@ -115,37 +115,37 @@ export class Migration20240219102530 extends Migration {
       CREATE INDEX IF NOT EXISTS "IDX_order_display_id" ON "order" (
```
adrien2p, line 115: Q: when handling migration changes, do you use the command or do you do it manually? Cause if the snapshot are not up to date we will never be able to use the commands again :D so many at some point we need to be sure that all modules snapshots are up to data wdyt?

carlos-r-l-rodrigues: I always do it manually.
yes, I think we have to have just 1 single initial migration file per module with the final snapshot updated before releasing them in v2.

packages/modules/order/src/migrations/Migration20240604100512.ts
carlos-r-l-rodrigues: These are fine to live here because it will only impact performance and not the functionality.
And I believe before releasing we should have just 1 single migration file for each module.

packages/core/core-flows/src/order/workflows/return/confirm-receive-return-request.ts
carlos-r-l-rodrigues: Just thinking if someone get stuck because of this condition, if they could somehow bypass it. But maybe they need to update the location of the return and then confirm this.

packages/core/core-flows/src/order/workflows/return/confirm-receive-return-request.ts
carlos-r-l-rodrigues: I added a couple of comments along the function.
I hope that helps a bit

olivermrbl: This is great, thanks!

olivermrbl (approved):

PR #8223, opened 2024-07-22
docs: added digital product recipe example

shahednasser:
- Added an example for the digital product recipe
- Fixes to the digital product recipe (overview)

Closes DOCS-792


olivermrbl (approved):
Noiiice

PR #8224, opened 2024-07-22
feat: Implement notifications feed

sradevski:
Designs: https://www.figma.com/design/z3aUuOVWUKmdHH0ofmMpEV/Web-app-3.0?node-id=10-50&t=9k6K9k7oJh5tIi09-0

Screenshot 2024-07-22 at 17 02 10

CLOSES CC-219

### DIFF
--- a/packages/admin-next/dashboard/src/components/common/infinite-list/infinite-list.tsx
+++ b/packages/admin-next/dashboard/src/components/common/infinite-list/infinite-list.tsx
@@ -53,13 +53,6 @@ export const InfiniteList = <
     },
     getPreviousPageParam: (firstPage) => {
       const moreItemsExist = firstPage.offset !== 0
-      console.log(
-        firstPage,
-        moreItemsExist
-          ? Math.max(firstPage.offset - firstPage.limit, 0)
-          : undefined
-      )
-
       return moreItemsExist
         ? Math.max(firstPage.offset - firstPage.limit, 0)
         : undefined
@@ -80,10 +73,6 @@ export const InfiniteList = <
       return
     }
 
-    // Clear the old observers
-    startObserver.current?.disconnect()
-    endObserver.current?.disconnect()
-
     // Define the new observers after we stop fetching
     if (!isFetching) {
       // Define the new observers after paginating
@@ -103,6 +92,12 @@ export const InfiniteList = <
       startObserver.current?.observe(parentRef.current!.firstChild as Element)
       endObserver.current?.observe(parentRef.current!.lastChild as Element)
     }
+
+    // Clear the old observers
+    return () => {
+      startObserver.current?.disconnect()
+      endObserver.current?.disconnect()
+    }
   }, [
     fetchNextPage,
     fetchPreviousPage,
@@ -119,7 +114,7 @@ export const InfiniteList = <
   }, [error])
 
   if (isPending) {
-    return <Skeleton className="h-[148x] w-full rounded-lg" />
+    return <Skeleton className="h-[148px] w-full rounded-lg" />
   }
 
   return (
--- a/packages/admin-next/dashboard/src/components/layout/notifications/notifications.tsx
+++ b/packages/admin-next/dashboard/src/components/layout/notifications/notifications.tsx
@@ -11,6 +11,7 @@ import { formatDistance } from "date-fns"
 import { Divider } from "../../common/divider"
 import { InfiniteList } from "../../common/infinite-list"
 import { sdk } from "../../../lib/client"
+import { notificationQueryKeys } from "../../../hooks/api"
 
 interface NotificationData {
   title: string
@@ -66,7 +67,7 @@ export const Notifications = () => {
             HttpTypes.AdminNotificationListParams
           >
             responseKey="notifications"
-            queryKey={["notifications"]}
+            queryKey={notificationQueryKeys.all}
             queryFn={(params) => sdk.admin.notification.list(params)}
             queryOptions={{ enabled: open }}
             renderItem={(notification) => {
@@ -98,40 +99,40 @@ const Notification = ({
 
   return (
     <>
-      <div className="align-center mt-2 flex flex-row justify-center px-5">
-        <div className="text-ui-fg-muted mr-3">
+      <div className="flex items-start justify-center gap-3 border-b p-6">
+        <div className="text-ui-fg-muted flex size-5 items-center justify-center">
           <InformationCircleSolid />
         </div>
-        <div className="w-full">
-          {/* We set the line height to 1 so the icon and text are aligned */}
-          <div className="align-center mb-1 flex flex-row justify-between">
-            <Label weight="plus" className="leading-4">
-              {data.title}
-            </Label>
-            <Text
-              as={"span"}
-              className="text-ui-fg-subtle leading-4"
-              size="small"
-            >
-              {formatDistance(notification.created_at, new Date(), {
-                addSuffix: true,
-              })}
-            </Text>
+        <div className="flex w-full flex-col gap-y-3">
+          <div>
+            <div className="align-center flex flex-row justify-between">
+              <Text size="small" leading="compact" weight="plus">
+                {data.title}
+              </Text>
+              <Text
+                as={"span"}
+                className="text-ui-fg-subtle"
+                size="small"
+                leading="compact"
+                weight="plus"
+              >
+                {formatDistance(notification.created_at, new Date(), {
+                  addSuffix: true,
+                })}
+              </Text>
+            </div>
+            {!!data.description && (
+              <Text
+                className="text-ui-fg-subtle whitespace-pre-line"
+                size="small"
+              >
+                {data.description}
+              </Text>
+            )}
           </div>
-
-          {!!data.description && (
-            <Text
-              className="text-ui-fg-subtle whitespace-pre-line"
-              size="small"
-            >
-              {data.description}
-            </Text>
-          )}
           <NotificationFile file={data.file} />
         </div>
       </div>
-
-      <Divider className="my-4" />
     </>
   )
 }
@@ -142,20 +143,16 @@ const NotificationFile = ({ file }: { file: NotificationData["file"] }) => {
   }
 
   return (
-    <div className="my-3 flex flex-col">
-      <div className="shadow-elevation-card-rest bg-ui-bg-component transition-fg rounded-md px-4 py-2">
-        <div className="flex items-center gap-4">
-          <div className="flex w-full flex-row items-center justify-between">
-            <Text size="small" leading="compact">
-              {file?.filename ?? file.url}
-            </Text>
-            <IconButton variant="transparent" className="ml-2">
-              <a href={file.url} download={file.filename ?? `${Date.now()}`}>
-                <ArrowDownTray />
-              </a>
-            </IconButton>
-          </div>
-        </div>
+    <div className="shadow-elevation-card-rest bg-ui-bg-component transition-fg rounded-md px-3 py-2">
+      <div className="flex w-full flex-row items-center justify-between gap-2">
+        <Text size="small" leading="compact">
+          {file?.filename ?? file.url}
+        </Text>
+        <IconButton variant="transparent" asChild>
+          <a href={file.url} download={file.filename ?? `${Date.now()}`}>
+            <ArrowDownTray />
+          </a>
+        </IconButton>
       </div>
     </div>
   )


packages/admin-next/dashboard/src/components/layout/notifications/notifications.tsx
```diff
@@ -1,9 +1,35 @@
...
+    mimeType?: string
+  }
+}
 
 export const Notifications = () => {
   const [open, setOpen] = useState(false)
+  const { t } = useTranslation()
+
+  // TODO: Make it infinitely scrollable
+  const { notifications } = useNotifications(
```
sradevski: @kasperkristensen I see we have an infinite scroll for the combobox, is there also some other component/list that is infinitely scrollable already?

packages/admin-next/dashboard/src/components/layout/notifications/notifications.tsx
```diff
@@ -31,10 +57,93 @@ export const Notifications = () => {
...
+              size="small"
+            >
+              {data.description}
+            </Text>
+          )}
+          <NotificationFile file={data.file} />
+        </div>
+      </div>
+
+      <Divider className="my-4" />
```
sradevski: @kasperkristensen since the drawer body has padding, the Divider is not full width (I can force it into being so, but I was wondering what's the pattern we've followed in other places)

packages/admin-next/dashboard/src/components/layout/notifications/notifications.tsx
```diff
@@ -31,10 +57,93 @@ export const Notifications = () => {
       </Drawer.Trigger>
       <Drawer.Content>
         <Drawer.Header>
-          <Heading>Notifications</Heading>
+          <Heading>{t("notifications.domain")}</Heading>
         </Drawer.Header>
-        <Drawer.Body>Notifications will go here</Drawer.Body>
+        <Drawer.Body className="overflow-y-auto">
```
kasperkristensen: You can remove the padding here, by setting `p-0` or `py-0 px-0` (can't remember what class is used to set the padding on the component in UI, and you need to override the specific one(s)).

packages/admin-next/dashboard/src/components/layout/notifications/notifications.tsx
kasperkristensen: We don't have a component for it, it's more a matter of setting up the `useInfiniteQuery` from `@tanstack/react-query`. You could properly re-use the `useComboboxData` hook, and just not make use of the search related values.

packages/admin-next/dashboard/src/components/layout/notifications/notifications.tsx
```diff
@@ -31,10 +57,93 @@ export const Notifications = () => {
       </Drawer.Trigger>
       <Drawer.Content>
         <Drawer.Header>
-          <Heading>Notifications</Heading>
+          <Heading>{t("notifications.domain")}</Heading>
```
kasperkristensen:
```suggestion
          <Drawer.Title asChild>
              <Heading>{t("notifications.domain")}</Heading>
          </Drawer.Title>
          <Drawer.Description className="sr-only">{t("some.translated.aria.description")}</Drawer.Description>
```

packages/admin-next/dashboard/src/components/layout/notifications/notifications.tsx
kasperkristensen: Answered this another comment, but all classNames can be overwritten

sradevski: Ok cool, I wasn't sure if we want to do such overrides, but if we're good with it I'll do it 👍

kasperkristensen: Thats the intended way to use `@medusajs/ui`. The components implement the common design, but we merge classNames, so when needed you can alter it to fit your use case 👍

sradevski:
@kasperkristensen I created an infinite list component, please have a look (modified the default arguments and the notification size to make it easier to demonstrate).


https://github.com/user-attachments/assets/5f19b987-195b-4ade-a617-a096c483e5ce

packages/admin-next/dashboard/src/components/common/infinite-list/infinite-list.tsx
```diff
@@ -0,0 +1,138 @@
...
+  ])
+
+  useEffect(() => {
+    if (error) {
+      toast.error(error.message)
+    }
+  }, [error])
+
+  if (isPending) {
+    return <Skeleton className="h-[148x] w-full rounded-lg" />
```
kasperkristensen:
```suggestion
    return <Skeleton className="h-[148px] w-full rounded-lg" />
```

packages/admin-next/dashboard/src/components/common/infinite-list/infinite-list.tsx
```diff
@@ -0,0 +1,138 @@
...
+  return (
+    <div ref={parentRef}>
+      {items &&
+        items.map((item) => <div key={item.id}>{renderItem(item)}</div>)}
+
+      {(isFetching || !hasNextPage) && (
+        <div className="my-4 flex flex-col items-center justify-center">
+          {isFetching && <Spinner className="animate-spin" />}
+          {!hasNextPage && <p className="m-2">{t("general.noMoreData")}</p>}
+        </div>
```
kasperkristensen, line 129: Don't think there is any need to display that there is no more data, if you reached the end, then it's implicit there is no more data to show. If you think we should keep it, we should at least style it so it fits in properly.

What we properly do want is a "no data to display" state, which this is also currently used for, but it looks off. Properly need ping Ludvig for a design.

packages/admin-next/dashboard/src/components/common/infinite-list/infinite-list.tsx
```diff
@@ -0,0 +1,138 @@
...
+
+  return (
+    <div ref={parentRef}>
+      {items &&
+        items.map((item) => <div key={item.id}>{renderItem(item)}</div>)}
+
+      {(isFetching || !hasNextPage) && (
+        <div className="my-4 flex flex-col items-center justify-center">
+          {isFetching && <Spinner className="animate-spin" />}
+          {!hasNextPage && <p className="m-2">{t("general.noMoreData")}</p>}
```
fPolic, line 128: todo: we need to show placeholder as well if no data exists, now this renders for empty state as well

packages/admin-next/dashboard/src/components/common/infinite-list/infinite-list.tsx
```diff
@@ -0,0 +1,138 @@
...
+    },
+    initialPageParam: 0,
+    maxPages: 5,
+    getNextPageParam: (lastPage) => {
+      const moreItemsExist = lastPage.count > lastPage.offset + lastPage.limit
+      return moreItemsExist ? lastPage.offset + lastPage.limit : undefined
+    },
+    getPreviousPageParam: (firstPage) => {
+      const moreItemsExist = firstPage.offset !== 0
+      console.log(
```
fPolic: todo: rm

packages/admin-next/dashboard/src/components/layout/notifications/notifications.tsx
```diff
@@ -31,10 +52,111 @@ export const Notifications = () => {
...
         </Drawer.Header>
-        <Drawer.Body>Notifications will go here</Drawer.Body>
+        <Drawer.Body className="overflow-y-auto px-0">
+          <InfiniteList<
+            HttpTypes.AdminNotificationListResponse,
+            HttpTypes.AdminNotification,
+            HttpTypes.AdminNotificationListParams
+          >
+            responseKey="notifications"
+            queryKey={["notifications"]}
```
fPolic: suggestion: `notificationQueryKeys.all` or `NOTIFICATION_QUERY_KEY `

packages/admin-next/dashboard/src/components/common/infinite-list/infinite-list.tsx
kasperkristensen: rm

packages/admin-next/dashboard/src/components/layout/notifications/notifications.tsx
fPolic: should we make dividers full width as per design?

packages/admin-next/dashboard/src/components/common/infinite-list/infinite-list.tsx
```diff
@@ -0,0 +1,138 @@
...
+  const endObserver = useRef<IntersectionObserver>()
+
+  useEffect(() => {
+    if (isPending) {
+      return
+    }
+
+    // Clear the old observers
+    startObserver.current?.disconnect()
+    endObserver.current?.disconnect()
```
fPolic: suggestion: we should do this in a function that is returned from the effect as per react convention

fPolic (commented):
Looks good overall, left a few comments

packages/admin-next/dashboard/src/components/layout/notifications/notifications.tsx
sradevski: @fPolic they are now, sorry i didn't update the screenshot

packages/admin-next/dashboard/src/components/common/infinite-list/infinite-list.tsx
srindom: +1 - have asked Ludvig for a design

packages/admin-next/dashboard/src/components/layout/notifications/notifications.tsx
srindom: > should we make dividers full width as per design?

yes :D

packages/admin-next/dashboard/src/components/common/infinite-list/infinite-list.tsx
srindom: btw. would suggest that this doesn't need to block we can make it a separate task

packages/admin-next/dashboard/src/components/layout/notifications/notifications.tsx
```diff
@@ -31,10 +52,111 @@ export const Notifications = () => {
...
+              <a href={file.url} download={file.filename ?? `${Date.now()}`}>
+                <ArrowDownTray />
+              </a>
+            </IconButton>
+          </div>
+        </div>
+      </div>
+    </div>
+  )
+}
```
kasperkristensen: The CSS is a bit off compared to Figma

```suggestion
const NotificationFile = ({ file }: { file: NotificationData["file"] }) => {
  if (!file?.url) {
    return null
  }

  return (
    <div className="shadow-elevation-card-rest bg-ui-bg-component transition-fg rounded-md px-3 py-2">
      <div className="flex w-full flex-row items-center justify-between gap-2">
        <Text size="small" leading="compact">
          {file?.filename ?? file.url}
        </Text>
        <IconButton variant="transparent" asChild>
          <a href={file.url} download={file.filename ?? `${Date.now()}`}>
            <ArrowDownTray />
          </a>
        </IconButton>
      </div>
    </div>
  )
}
```

packages/admin-next/dashboard/src/components/layout/notifications/notifications.tsx
```diff
@@ -31,10 +52,111 @@ export const Notifications = () => {
...
+            </Text>
+          )}
+          <NotificationFile file={data.file} />
+        </div>
+      </div>
+
+      <Divider className="my-4" />
+    </>
+  )
+}
```
kasperkristensen: CSS is a bit of compared to Figma:

```suggestion
const Notification = ({
  notification,
}: {
  notification: HttpTypes.AdminNotification
}) => {
  const data = notification.data as unknown as NotificationData | undefined

  // We need at least the title to render a notification in the feed
  if (!data?.title) {
    return null
  }

  return (
    <div className="flex items-start justify-center gap-3 border-b p-6">
      <div className="text-ui-fg-muted flex size-5 items-center justify-center">
        <InformationCircleSolid />
      </div>
      <div className="flex w-full flex-col gap-y-3">
        <div>
          <div className="align-center flex flex-row justify-between">
            <Text size="small" leading="compact" weight="plus">
              {data.title}
            </Text>
            <Text
              as={"span"}
              className="text-ui-fg-subtle"
              size="small"
              leading="compact"
              weight="plus"
            >
              {formatDistance(notification.created_at, new Date(), {
                addSuffix: true,
              })}
            </Text>
          </div>
          {!!data.description && (
            <Text
              className="text-ui-fg-subtle whitespace-pre-line"
              size="small"
            >
              {data.description}
            </Text>
          )}
        </div>
        <NotificationFile file={data.file} />
      </div>
    </div>
  )
}
```

packages/admin-next/dashboard/src/components/common/infinite-list/infinite-list.tsx
sradevski: This particular disconnect has to happen before we create the new ones, but I added a new empty useEffect to clean up on unmount. WDYT?

packages/admin-next/dashboard/src/components/common/infinite-list/infinite-list.tsx
sradevski: We can definitely do a better job at the design (will ask Ludvig), let me see what he thinks. I think being explicit makes it clear that there is no more data vs. maybe users thinking the request is just slow.

sradevski:
@fPolic @kasperkristensen @srindom  opened a new ticket for the empty state/no more data, and resolved all other comments.

packages/admin-next/dashboard/src/components/common/infinite-list/infinite-list.tsx
kasperkristensen: But isn't that apparent from the spinner being shown when loading new data? 

Just think it looks weird that when you have just a few notifications, you can constantly see:

[Notification]
[No more data]

packages/admin-next/dashboard/src/components/common/infinite-list/infinite-list.tsx
fPolic: But the new effect isn't necessary, all we need is to move the return function from your new effect to this effect and everything will work as expected.

On the initial pass, there are no observers so we don't need cleanup before. After first, pass a cleanup function is registered and React will call it before it runs the effect for the second time and so on.

fPolic (approved):

sradevski:
@fPolic ah yes true, let me update it 👍

packages/admin-next/dashboard/src/components/common/infinite-list/infinite-list.tsx
sradevski: Most of the time people won't see the spinner, since the API is quite fast, so they might not even know that a spinner shows up. Anyway, it's a really minor thing, so I'll let Ludvig decide however he wants it handled, I am happy either way.

PR #8231, opened 2024-07-23
fix(admin): Allow `backend` to pass through in development

### LINKED ISSUE #8218: ERR_CONNECTION_REFUSED, after npx create-medusa-app@latest on REMOTE_IP:7001
There is a bug, when I am using the default settings `npx create-medusa-app@latest`, which starts the dev, I can't access the store in REMOTE_IP:7001, -> `ERR_CONNECTION_REFUSED`
 I need to change `localhost` with `0.0.0.0` and I can access it, but then, I see a CORS problem. This is with every EC2 instance, that Ive tried - Ubuntu and Amazon Linux 2023. The CORS problem is for host `localhost` and there is no sence, because on the client, `localhost` resolves to the client, but not to remote.
In my local machine with Windows, it works fine - it opens `http://localhost:7001/invite?token=...`, but in remote, after ``npx create-medusa-app@latest` -> REMOTE_IP:7001, -> `ERR_CONNECTION_REFUSED`

- The settings are, what they are by default, when you run `npx create-medusa-app@latest`
- NodeJS v18.20.3

kasperkristensen:
Fixes #8218

### DIFF
--- a/.changeset/strange-knives-check.md
+++ b/.changeset/strange-knives-check.md
@@ -1,0 +1,5 @@
+---
+"@medusajs/admin": patch
+---
+
+fix(admin): Allow `backend` arg to pass through in development


olivermrbl (approved):

olivermrbl:
@kasperkristensen, have you given this a test-spin? I want to include it in the next release :)

PR #8232, opened 2024-07-23
feat: Add skeleton for supporting product imports in BE

sradevski:
REF CC-51


packages/medusa/src/api/admin/products/middlewares.ts
```diff
@@ -28,6 +28,12 @@ import {
   AdminUpdateProductVariant,
   AdminUpdateVariantInventoryItem,
 } from "./validators"
+import multer from "multer"
+
+// TODO: For now we keep the files in memory, as that's how they get passed to the workflows
+// This will need revisiting once we are closer to prod-ready v2, since with workflows and potentially
+// services on other machines using streams is not as simple as it used to be.
```
olivermrbl, line 35: comment: If you haven't already, should we create a ticket for this so we don't forget about it

olivermrbl (approved):
LGTM!

PR #8241, opened 2024-07-23
chore: Make scheduled job tests reliable

sradevski:
Testing against a real service is a pain, so we can't do better unless we mock/fake some of the components

### DIFF
--- a/packages/modules/workflow-engine-redis/integration-tests/__fixtures__/workflow_scheduled.ts
+++ b/packages/modules/workflow-engine-redis/integration-tests/__fixtures__/workflow_scheduled.ts
@@ -5,8 +5,13 @@ import {
   StepResponse,
 } from "@medusajs/workflows-sdk"
 
-export const createScheduled = (name: string, schedule?: SchedulerOptions) => {
+export const createScheduled = (
+  name: string,
+  next: () => void,
+  schedule?: SchedulerOptions
+) => {
   const workflowScheduledStepInvoke = jest.fn((input, { container }) => {
+    next()
     return new StepResponse({
       testValue: container.resolve("test-value"),
     })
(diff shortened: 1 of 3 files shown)


carlos-r-l-rodrigues:
I believe instead of setting the timeout to wait, you could have a counter and only check the expect after that.
This way we avoid any issue if there is any delay executing the job because it is busy, etc...

```
function ExternalPromise(num) {
  let externalResolve;
  let externalReject;
  let counter = 0
  const promise = new Promise((resolve, reject) => {
    externalResolve = resolve;
    externalReject = reject;
  });

  return {
    next: () => {
       if(++counter === num) {
          externalResolve()
       }
    },
    promise,
    resolve: externalResolve,
    reject: externalReject
  };
}

it("should execute a scheduled workflow", async () => {
   const wait = ExternalPromise(3)
   const spy = createScheduled("standard", wait)
   await wait.promise
   expect(spy).toHaveBeenCalledTimes(3)
   WorkflowManager.unregister("standard")
})
```
and in `createScheduled` you execute `wait.next()`

sradevski:
> I believe instead of setting the timeout to wait, you could have a counter and only check the expect after that. This way we avoid any issue if there is any delay executing the job because it is busy, etc...
> 
> ```
> function ExternalPromise(num) {
>   let externalResolve;
>   let externalReject;
>   let counter = 0
>   const promise = new Promise((resolve, reject) => {
>     externalResolve = resolve;
>     externalReject = reject;
>   });
> 
>   return {
>     next: () => {
>        if(++counter === num) {
>           externalResolve()
>        }
>     },
>     promise,
>     resolve: externalResolve,
>     reject: externalReject
>   };
> }
> 
> it("should execute a scheduled workflow", async () => {
>    const wait = ExternalPromise(3)
>    const spy = createScheduled("standard", wait)
>    await wait.promise
>    expect(spy).toHaveBeenCalledTimes(3)
>    WorkflowManager.unregister("standard")
> })
> ```
> 
> and in `createScheduled` you execute `wait.next()`

@carlos-r-l-rodrigues nice, I did something similar for waiting for the events, but I didn't think of using the same approach here 👍  Let me try it out

sradevski:
@carlos-r-l-rodrigues done, much better

carlos-r-l-rodrigues:
Nice... it looks good.

carlos-r-l-rodrigues (approved):

PR #8244, opened 2024-07-23
fix: `isList` on field alias in link configuration

adrien2p:
**What**
Fixes the link configurations from the link modules

FIXES TRI-62


carlos-r-l-rodrigues (approved):

sradevski (approved):

olivermrbl:
@adrien2p, looks like we might need to update a few tests. Could I get you to take a look, so we can get it merged? 💪

PR #8247, opened 2024-07-23
fix(types,order): fixes to order module's service and data models

shahednasser:
- Fix the return type of some methods in the `IOrderModuleService`.
- Remove duplicate signatures for `deleteOrderChanges`.
- Fix typo in `soft-remove` in the `OrderChange` data model.

### DIFF
--- a/packages/core/types/src/order/service.ts
+++ b/packages/core/types/src/order/service.ts
@@ -278,7 +278,10 @@ export interface IOrderModuleService extends IModuleService {
    * ```
    *
    */
-  updateOrders(data: UpdateOrderDTO[]): Promise<OrderDTO[]>
+  updateOrders(
+    data: UpdateOrderDTO[],
+    sharedContext?: Context
+  ): Promise<OrderDTO[]>
 
   /**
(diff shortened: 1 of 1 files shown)


packages/core/types/src/order/service.ts
```diff
@@ -1978,7 +1942,8 @@ export interface IOrderModuleService extends IModuleService {
   ): Promise<OrderExchangeDTO[]>
 
   updateOrderExchanges(
-    data: UpdateOrderExchangeWithSelectorDTO[]
+    data: UpdateOrderExchangeWithSelectorDTO[],
+    sharedContext?: Context
```
shahednasser, line 1972: @carlos-r-l-rodrigues I noticed many `update` and `create` methods are missing the context. I wasn't sure if that was purpose. Is it ok to add them?

carlos-r-l-rodrigues: Hi Shahed, yes please.
these are just wrong copy/pasta

carlos-r-l-rodrigues (approved):
Thank you @shahednasser.

PR #8250, opened 2024-07-23
fix(order): remove nested entities

carlos-r-l-rodrigues:


fPolic (approved):

packages/modules/order/src/utils/transform-order.ts
```diff
@@ -99,10 +100,15 @@ export function formatOrder(
 }
 
 function cleanNestedRelations(obj) {
-  delete obj.order
-  delete obj.return
-  delete obj.claim
-  delete obj.exchange
+  if (!isObject(obj)) {
```
adrien2p, line 103: shouldn't we use omitDeep? :)

carlos-r-l-rodrigues: I have never seen this util before 🤣 
Will keep that in mind to cleanup more nestesd things returned from Mikro-ORM.

PR #8254, opened 2024-07-24
feat: Extending API routes validators

thetutlage:
FIXES: FRMW-2617

In this PR we add support for extending the API routes validators. This requires a few changes at different layers of the `medusa` package.

## First update
Currently, I have updated the `validateAndTransformBody` body to read a custom (user defined) schema from the `req.bodyValidator` property and merge it with the original validator of the route.

However, in some cases, the original validator can be a `ZodEffect` (result of `zod.refine` returns `ZodEffect`) and you cannot merge new schemas to a `ZodEffect`. Therefore, this needs some changes with existing validators as well.

Any validator that was previously a `ZodEffect` is now converted to a callback function which accepts a custom validation schema, performs a merge under the hood and then applies the `refine`.

Let me know if anyone of you have a suggestion for a better implementation.


olivermrbl (approved):
LGTM, imo. the current approach wrt ZodEffects is fine. We haven't used refinements much, so don't think we have to spend too much time coming up with a very sophisticated solution right now.

adrien2p (approved):
LGTM

for the schema, would `.and | .intersection` have the same effect has merge in general?

packages/medusa/src/api/utils/validate-body.ts
```diff
@@ -4,15 +4,29 @@ import { MedusaRequest, MedusaResponse } from "../../types/routing"
...
   res: MedusaResponse,
   next: NextFunction
 ) => Promise<void> {
   return async (req: MedusaRequest, _: MedusaResponse, next: NextFunction) => {
     try {
-      req.validatedBody = await zodValidator(zodSchema, req.body)
+      let schema: z.ZodObject<any, any> | z.ZodEffects<any, any>
+
+      if (typeof zodSchema === "function") {
+        schema = zodSchema(req.bodyValidator)
```
sradevski: req.bodyValidator can be undefined right?

packages/medusa/src/types/routing.ts
```diff
@@ -61,6 +62,16 @@ export interface MedusaRequest<Body = unknown>
    * A generic context object that can be used across the request lifecycle
    */
   context?: Record<string, any>
+  /**
+   * A custom validator for the request body that will be merged with the original
+   * validator of the route
+   */
+  bodyValidator?: ZodObject<any, any>
```
sradevski: NIT: The naming maybe can be better, as like this is sounds like it's the only validator, unless you read the description. I am raising more to see what you think, as I don't have great ideas, `extendedBodySchema` and `extendedQuerySchema` come to mind.

sradevski (approved):
Nice, I think this works! just a couple of minor comments

packages/medusa/src/api/utils/validate-body.ts
thetutlage: Yes it can be!

packages/medusa/src/types/routing.ts
thetutlage: Yeah. I agree, the `bodyValidator` seems like the final validator to be applied and not merged.

We intentionally decided to not use the keyword `schema` because it might mean a query schema, a db schema and so on.

So maybe, we can rename the property to `validatorsToMerge.body` and `validatorsToMerge.queryParams`. Wdyt?

thetutlage: Something like this 

```ts
req.validatorsToMerge = {
  body: zod.object(),
  queryParams: zod.object(),
}
```

adrien2p: I was fine with the current name as it can evolve later, eventually I was thinking of something like:
```ts
type ExtendedValidator = ZodObject<any, any> | {
  schema: ZodObject<any, any>,
  strategy?: 'merge' | 'override' // default being merge
}

req.extendedValidator = {}
```

but we can argue that we don't need that right now and can leave that for the future

olivermrbl: very-nit and totally non-blocking: How about? Would bring their name very close to what they are 😄 
```
validatorExtensions.body
validatorExtensions.params
```
or (as mentioned by @adrien2p )
```
extendedValidators.body
extendedValidators.params
```

thetutlage: Went with `extendedValidators` suggestion

thetutlage:
After this PR, we can essentially use a middleware to assign `req.extendedValidators.body` and `req.extendedValidators.queryParams` to merge custom zod schemas with the default validator of a route.

However, another PR is needed that will act as a wrapper for creating this middleware by exposing a high-level API to define these schemas.

PR #8258, opened 2024-07-24
chore(framework): Continue to move loaders to framework

adrien2p:
**What**
More move and cleanup

FIXES FRMW-2603
FIXES FRMW-2608
FIXES FRMW-2610
FIXES FRMW-2611

### DIFF
--- a/packages/framework/framework/package.json
+++ b/packages/framework/framework/package.json
@@ -66,7 +66,7 @@
   },
   "dependencies": {
     "@medusajs/medusa-cli": "^1.3.22",
-    "@medusajs/utils": "^1.11.10",
+    "@medusajs/utils": "^1.11.9",
     "awilix": "^8.0.0",
     "cookie-parser": "^1.4.6",
     "express": "^4.18.2",
--- a/packages/core/medusa-test-utils/package.json
+++ b/packages/core/medusa-test-utils/package.json
@@ -26,6 +26,7 @@
   "devDependencies": {
     "@medusajs/framework": "^0.0.1",
     "@medusajs/types": "^1.11.16",
+    "@medusajs/utils": "^1.11.9",
     "cross-env": "^5.2.1",
     "jest": "^29.7.0",
     "rimraf": "^3.0.2",
--- a/yarn.lock
+++ b/yarn.lock
@@ -4610,7 +4610,7 @@ __metadata:
   dependencies:
     "@medusajs/medusa-cli": ^1.3.22
     "@medusajs/types": ^1.11.16
-    "@medusajs/utils": ^1.11.10
+    "@medusajs/utils": ^1.11.9
     "@types/express": ^4.17.17
     awilix: ^8.0.0
     cookie-parser: ^1.4.6
@@ -5284,13 +5284,6 @@ __metadata:
   languageName: unknown
   linkType: soft
 
-"@medusajs/types@npm:^1.11.17":
-  version: 1.11.17
-  resolution: "@medusajs/types@npm:1.11.17"
-  checksum: 544099b215cd4de87682c2154696ff4849704f74290f15e3e383e6cf503a335767b399b9fe2a381ffa4405fa2ec7ed911a8a5eecff6c7f10d5ef95af802064b6
-  languageName: node
-  linkType: hard
-
 "@medusajs/ui-preset@1.1.3, @medusajs/ui-preset@^1.1.3, @medusajs/ui-preset@workspace:packages/design-system/ui-preset":
   version: 0.0.0-use.local
   resolution: "@medusajs/ui-preset@workspace:packages/design-system/ui-preset"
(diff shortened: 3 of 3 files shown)


packages/medusa/package.json
```diff
@@ -45,7 +45,7 @@
     "@inquirer/checkbox": "^2.3.11",
     "@medusajs/admin-sdk": "0.0.1",
     "@medusajs/core-flows": "^0.0.9",
-    "@medusajs/framework": "^0.0.1-preview-20240724081425",
+    "@medusajs/framework": "workspace:^",
```
adrien2p: I ll have a disussion with @olivermrbl on this issue I am having at the moment

adrien2p:
@olivermrbl once this pr is ok (meaning only the tests relying on the CLI/starter fail) I will revert the version to preview, then everything will fail. Should we still merge after that change?

packages/medusa/package.json
adrien2p: Only apply this change before merging

```suggestion
    "@medusajs/framework": "^0.0.1-preview-20240724081425",
```

integration-tests/http/package.json
```diff
@@ -5,7 +5,7 @@
   "license": "MIT",
   "private": true,
   "scripts": {
-    "test:integration": "NODE_OPTIONS=--experimental-vm-modules jest --no-cache --maxWorkers=50% --bail --detectOpenHandles --forceExit --logHeapUsage",
+    "test:integration": "NODE_OPTIONS=--experimental-vm-modules jest --no-cache --maxWorkers=50% --bail --detectOpenHandles --forceExit --logHeapUsage -- __tests__/product/store/product.spec.ts",
```
sradevski: I guess this isn't intentional :) If it was about the flaky test, it should be fixed now, but let me know if it happens again.

sradevski (approved):
LGTM!

adrien2p:
@olivermrbl should we apply the comment changes and merge? Wdyt? I already have another branch from this one for the next part

integration-tests/http/package.json
adrien2p: Definitely not 👌 thanks I missed it in my review

olivermrbl (approved):

olivermrbl:
@adrien2p, let's do it!

adrien2p:
@olivermrbl just merged the revert of the version, I expect the tests to fail but I think it is temporary right?

olivermrbl:
@adrien2p, I believe the lock-file should also be updated before we merge right?

adrien2p:
> @adrien2p, I believe the lock-file should also be updated before we merge right?

Yes thats correct

olivermrbl:
@adrien2p, just playing around with making the framework-versions work. I'll revert my commits if I can't solve it.

adrien2p:
Got it 👍 no problem at all let me know now how it goes and the solution

olivermrbl:
> let me know now how it goes and the solution

Do you think the latest changes could work?

adrien2p:
@olivermrbl I believe after this last commit we should be able to finally merge it

adrien2p:
@olivermrbl should we merge as it is?

packages/framework/framework/package.json
```diff
@@ -42,25 +48,31 @@
...
     "tsc-alias": "^1.8.6",
     "typescript": "^5.1.6",
     "vite": "^5.2.11"
   },
   "dependencies": {
-    "@medusajs/medusa-cli": "^1.3.23",
-    "@medusajs/utils": "^1.12.0-preview-20240724081425",
-    "awilix": "^8.0.0"
+    "@medusajs/medusa-cli": "^1.3.22",
+    "@medusajs/utils": "^1.12.0-snapshot-20240728140928",
```
adrien2p: @olivermrbl should i move that back to preview before we merge?

olivermrbl: Can you move them to use their latest versions with a caret (`^`). I think then we can merge

adrien2p: let me try, last time it was failing for me 😂

adrien2p: Looks like this breaks everything 🤒

adrien2p: let me try something

olivermrbl: ~~Think the wrong version of utils is installed right now. There is no 1.11.10~~

adrien2p: Yes there is one https://www.npmjs.com/package/@medusajs/utils/v/1.11.10

adrien2p: Otherwise the install would have failed with 404 and not reaching the build

adrien2p: Now I have another question, how come 1.11.10 does not have the latest changes when it is the latest version and i have to downgrade ?

olivermrbl: Yes, I noticed this. Edited my comment :)

olivermrbl: > Now I have another question, how come 1.11.10 does not have the latest changes when it is the latest version and i have to downgrade ?

This is what spawned my comment initially. I was wondering this too. My immediate thought was that this is because 1.11.10 has been released as part of v1.x, and that branch doesn't have everything from `develop`.

adrien2p: ahah makes sense 👍

olivermrbl: Pretty sure this is actually what has been causing issues all along.

adrien2p: Yes probably, and i ve missed it for so long 😣

olivermrbl: Let's slap an automerge on this, shall we?

adrien2p: done 😍

PR #8261, opened 2024-07-24
chore(framework): update deps

adrien2p:
**What**
Update deps of medusa and framework to not rely on workspace

### DIFF
--- a/packages/medusa/src/api/admin/campaigns/validators.ts
+++ b/packages/medusa/src/api/admin/campaigns/validators.ts
@@ -26,7 +26,7 @@ export const AdminGetCampaignsParams = createFindParams({
   )
   .strict()
 
-export const CreateCampaignBudget = z
+const CreateCampaignBudget = z
   .object({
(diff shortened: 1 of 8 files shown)


packages/medusa/src/loaders/pg-connection.ts
```diff
@@ -1,4 +1,4 @@
-import { ConfigModule } from "@medusajs/types"
+import { ConfigModule } from "@medusajs/framework"
```
shahednasser, line 1: Maybe unrelated to the PR, but why is `ConfigModule` in `framework` and not `types`?

shahednasser (approved):

packages/medusa/src/loaders/pg-connection.ts
adrien2p: For now I ve done that but both still exists. This will be better decided later on when we have a proper separation between what the framework provide and what medusa will do with it. Also, this loader will probably move to the framework as well so I thought it is fine for now

PR #8263, opened 2024-07-24
feat: add extend router helper

thetutlage:
FIXES: FRMW-2621

The goal of the `extendRouter` helper is to provide a unified API for extending routes. Extending routes is a bucket of extensions which includes:

- Defining custom middleware for a route
- Defining extendedValidators for a route
- Defining custom bodyParser config for a route.

More options can be added later.

---

## The `middlewares.ts` file has to go away

If you follow the `v2` documentation https://docs.medusajs.com/v2/advanced-development/api-routes/middlewares, the `src/api/middlewares.ts` file is recommended to define a custom set of middleware.

Now since we are merging the concept of defining `middlewares` and `extendedValidators` in one place. We will have to rename this file to something else and also update the loader to load the new file.

Any recommendations for the new file name? Let me start by suggesting `src/api/extend-router.ts`.


shahednasser (commented):
Thought: I'm not very fond of the name `extendRouter`, just because the developer isn't working with any router concept directly. Maybe `routesConfig`? And the file's name can be `config.ts` or `route-config.ts`. Wdyt?

Also, should we update the docs to use this new file right after this is merged or is there more work to be done?

adrien2p (approved):

adrien2p:
we can indeed align on the naming but from an implementation pov LGTM

thetutlage:
Yeah, not a big fan of the `extendRouter` name myself. 

The issue with `routesConfig` is that its not an actionable name. Usually, when you will look for builder functions in the wild, you will notice they are named after performing some action.

For example: `defineConfig`, `useRouter` and so on.

So yeah, maybe we can brainstorm a little bit more and see if we can find some names. For the context.

- The helper does not define any routes. It just extends the existing routes.
- The extension nature is also very generic. Middleware, validators, custom config all can be added via this helper.

@shahednasser Yes. The loaders have to be updated as well. I will do that in separate PR and ping you for the docs change.

shahednasser:
@thetutlage fair point. Maybe `defineRouteConfig` then? or `defineRoutingConfig`

thetutlage:
Makes sense. Should we finalize `routes-config.ts` for the file name and `defineRoutesConfig` for the helper?

I am using plural form for the `routes`, since you will be configuring multiple routes in one go.

shahednasser:
Sounds good to me, but would love to hear others thoughts too

PR #8265, opened 2024-07-24
chore(utils): Update internal service update input validation

adrien2p:
**What**
Add more validation to the update input arguments to catch wrong usage 

FIXES TRI-64

### DIFF
--- a/packages/core/utils/src/modules-sdk/medusa-internal-service.ts
+++ b/packages/core/utils/src/modules-sdk/medusa-internal-service.ts
@@ -246,12 +246,39 @@ export function MedusaInternalService<
       selectorAndData: SelectorAndData[],
       sharedContext?: Context
     ): Promise<InferEntityType<TEntity>[]>
+    update(
+      idOrSelector: string | SelectorAndData["selector"],
+      data: any,
+      sharedContext?: Context
+    ): Promise<InferEntityType<TEntity>[]>
 
     @InjectTransactionManager(shouldForceTransaction, propertyRepositoryName)
     async update(
       input: any | any[] | SelectorAndData | SelectorAndData[],
+      maybeDataOrContext?: any,
       @MedusaContext() sharedContext: Context = {}
     ): Promise<InferEntityType<TEntity> | InferEntityType<TEntity>[]> {
+      if (arguments.length === 2) {
+        Object.assign(sharedContext, maybeDataOrContext)
+      } else if (arguments.length === 3) {
+        const input_ = {
+          selector: {},
+          data: {} as any,
+        }
+
+        if (isString(input)) {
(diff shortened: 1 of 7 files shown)


adrien2p:
I am also trying to improve the base API in a separate pr @olivermrbl, though, I ll have to ts expect error in many services because they override the base method with a different API which is not supposed to happen, but I can ignore them for now until the full modules API is settled and in any case those override normalize their input to be consumed by the base methods

packages/core/utils/src/modules-sdk/medusa-internal-service.ts
```diff
@@ -252,12 +252,22 @@ export function MedusaInternalService<
       input: any | any[] | SelectorAndData | SelectorAndData[],
       @MedusaContext() sharedContext: Context = {}
     ): Promise<InferEntityType<TEntity> | InferEntityType<TEntity>[]> {
-      if (!isDefined(input) || (Array.isArray(input) && input.length === 0)) {
+      if (
```
carlos-r-l-rodrigues, line 282: can we use `isPresent` here?

adrien2p: yes you are right 👍 much better, I ll add that in

packages/core/utils/src/modules-sdk/medusa-internal-service.ts
```diff
@@ -252,12 +252,22 @@ export function MedusaInternalService<
       input: any | any[] | SelectorAndData | SelectorAndData[],
```
olivermrbl, line 257: thought: @shahednasser, do we have a reference for the internal module services? If not, would it make sense to have such that we can provide a link in the error message?

shahednasser: Not a reference, but we have examples here: https://docs.medusajs.com/v2/advanced-development/modules/service-factory#generated-methods

We can generate a reference if we find it necessary, but the reason I didn't is I imagine the typings being confusing. We can alternatively have a document that provides more advanced / realistic examples

adrien2p:
@olivermrbl I will stop this pr for now as it is still documented. But it require a much broader work on aligning the different API's the modules are defining VS what the internal service is provided. There is too much discrepancies for now and I d like to finalise my other tasks. wdyt?

thought, I ll keep this pr open for reference in case I get back to this one later

olivermrbl:
> @olivermrbl I will stop this pr for now as it is still documented. But it require a much broader work on aligning the different API's the modules are defining VS what the internal service is provided. There is too much discrepancies for now and I d like to finalise my other tasks. wdyt?
> 
> thought, I ll keep this pr open for reference in case I get back to this one later

Yeah, let's do that for now. We can pick it up as part of our polishing phase.

PR #8266, opened 2024-07-24
feat: Add basic support for importing products

sradevski:
We are still missing variant pricing updates support, and async confirmation endpoint and step, but there are separate tickets for those.


olivermrbl (approved):
Overall LGTM, nice work. Looking forward to start testing this. Think it will be easier for detecting potentials bugs and/or shortcomings of the import 👍

PR #8271, opened 2024-07-24
chore(core-flows,medusa): order claims

carlos-r-l-rodrigues:
PR 1 of 2

What:
 * `POST /admin/claims`
 * `POST /admin/claims/:id/claim-items`
 * `POST /admin/claims/:id/claim-items/:action_id`
 * `DELETE /admin/claims/:id/claim-items/:action_id`
 * `POST /admin/claims/:id/inbound/items`
 * `POST /admin/claims/:id/inbound/items/:action_id`
 * `DELETE /admin/claims/:id/inbound/items/:action_id`
 * `POST /admin/claims/:id/inbound/shipping-method`
 * `POST /admin/claims/:id/inbound/shipping-method/:action_id`
 * `DELETE /admin/claims/:id/inbound/shipping-method/:action_id`
 * `POST /admin/claims/:id/outbound/items`
 * `POST /admin/claims/:id/outbound/items/:action_id`
 * `DELETE /admin/claims/:id/outbound/items/:action_id`
 * `POST /admin/claims/:id/outbound/shipping-method`
 * `POST /admin/claims/:id/outbound/shipping-method/:action_id`
 * `DELETE /admin/claims/:id/outbound/shipping-method/:action_id`
 
 FIXES: CC-75, CC-86, CC-85, CC-84, CC-83

### DIFF
--- a/packages/core/core-flows/src/order/steps/update-returns.ts
+++ b/packages/core/core-flows/src/order/steps/update-returns.ts
@@ -5,7 +5,7 @@ import {
 } from "@medusajs/utils"
 import { StepResponse, createStep } from "@medusajs/workflows-sdk"
 
-export const updateReturnsStepId = "update-return-items"
+export const updateReturnsStepId = "update-returns"
 export const updateReturnsStep = createStep(
   updateReturnsStepId,
   async (data: UpdateReturnDTO[], { container }) => {


packages/medusa/src/api/admin/claims/route.ts
```diff
@@ -0,0 +1,74 @@
...
+      id: result.claim_id,
+      filters: {
+        ...req.filterableFields,
+      },
+    },
+    fields: req.remoteQueryConfig.fields,
+  })
+
+  const [order, orderClaim] = await promiseAll([
+    orderModuleService.retrieveOrder(result.order_id),
```
olivermrbl, line 66: question: Could we imagine the remote query supporting combined queries, such that you could fetch order + claim in this case? 

This is not something we need to think about now, I was just curious if this had struck your mind previously.

packages/medusa/src/api/admin/claims/middlewares.ts
```diff
@@ -0,0 +1,229 @@
...
+      validateAndTransformBody(AdminPostClaimsConfirmRequestReqSchema),
+      validateAndTransformQuery(
+        AdminGetOrdersOrderParams,
+        QueryConfig.retrieveTransformQueryConfig
+      ),
+    ],
+  },
+  {
+    method: ["DELETE"],
+    matcher: "/admin/claims/:id/request",
```
olivermrbl, line 226: nit: Typo + I don't think the route implementation is included
```suggestion
    matcher: "/admin/claims/:id",
```

packages/core/core-flows/src/order/workflows/return/confirm-receive-return-request.ts
```diff
@@ -232,9 +238,26 @@ export const confirmReturnReceiveWorkflow = createWorkflow(
...
+        const hasReceivedAllItems = retItems.every((item) => {
+          const received = itemUpdates[item.item_id]
+            ? itemUpdates[item.item_id].received_quantity
+            : item.received_quantity
+
+          return MathBN.eq(received, item.quantity)
+        })
+        const updateReturnData = hasReceivedAllItems
+          ? { status: ReturnStatus.RECEIVED, received_at: new Date() }
+          : { status: ReturnStatus.PARTIALLY_RECEIVED }
```
olivermrbl, line 250: comment: This might change if we choose to remove partial receival. Just adding a comment in case we reach a conclusion before the PR is merged so we can get it updated.

packages/core/core-flows/src/order/workflows/claim/update-claim-add-item.ts
```diff
@@ -0,0 +1,118 @@
...
+        `No request to add item for claim ${input.claim_id} in order change ${orderChange.id}`
+      )
+    } else if (associatedAction.action !== ChangeActionType.ITEM_ADD) {
+      throw new Error(`Action ${associatedAction.id} is not adding an item`)
+    }
+  }
+)
+
+export const updateClaimAddItemWorkflowId = "update-claim-add-item"
+export const updateClaimAddItemWorkflow = createWorkflow(
```
olivermrbl, line 60: question: Maybe this is obvious and I am just missing a bit of context, but how is this workflow different from `updateClaimItemWorkflow`? It looks like only the validation differs. Is that intentional?

olivermrbl (commented):
Overall LGTM, nice work @carlos-r-l-rodrigues!

It was quite a big PR, so I didn't go through it line-by-line. Would maybe suggest we either use this PR to build the claim flows, or we merge it as is and address bugs/changes in follow-up PRs. 

What do you think? 

cc @fPolic

packages/medusa/src/api/admin/claims/middlewares.ts
```diff
@@ -0,0 +1,229 @@
...
+  AdminPostClaimsAddItemsReqSchema,
+  AdminPostClaimsConfirmRequestReqSchema,
+  AdminPostClaimsRequestItemsActionReqSchema,
+  AdminPostClaimsRequestReturnItemsReqSchema,
+  AdminPostClaimsShippingActionReqSchema,
+  AdminPostClaimsShippingReqSchema,
+  AdminPostOrderClaimsReqSchema,
+} from "./validators"
+
+export const adminClaimRoutesMiddlewares: MiddlewareRoute[] = [
```
fPolic, line 17: todo: export routes to index

packages/medusa/src/api/admin/claims/middlewares.ts
fPolic: Also `POST /admin/claims/:id/request` isn't included. (Disregard if it will be added in part 2 :D).

fPolic (approved):
I'm fine with merging this now not to block further work and we can open Linear thread if there will be potential things to address while building FE flow.

carlos-r-l-rodrigues:
I should have added the context on the PR description, that this is mostly a skeleton for claims operations.
The following PR will contain all the tests and potentially specific behaviors for claims.

packages/medusa/src/api/admin/claims/route.ts
carlos-r-l-rodrigues: Yes, the Remote Joiner could resolve multiple queries in parallel if this is what you mean. Here is even simpler because it is a regular relation, but there are multiple flows where a single query could be used.
When we have all the flows done, I will take the time to cleanup all the repetition we have in most workflows where the validations are essencially the same thing and transform into single queries.

packages/medusa/src/api/admin/claims/middlewares.ts
carlos-r-l-rodrigues: Claims will have a DELETE method to cancel the "draft claim" but it also needs a DELETE to cancel the claim after it is created and confirmed.

packages/core/core-flows/src/order/workflows/return/confirm-receive-return-request.ts
carlos-r-l-rodrigues: I'm just bringing what we already have done for the complete return flow
`packages/modules/order/src/services/actions/create-return.ts`

If you guys decide to remove this feature, I can remove even other things like the received_quantity from the return model, etc.
I personally wouldn't remove features that could be used.

packages/core/core-flows/src/order/workflows/claim/update-claim-add-item.ts
carlos-r-l-rodrigues: It is very likely the same.
At this point all these post purchase flows have much more repetition than I'm comfortable maintaining. After a cleanup I expect more generic flows that can be reused and many ones renamed or removed.

PR #8274, opened 2024-07-25
breaking: remove loading middlewares file in favor of routes-config files

thetutlage:
Fixes: FRMW-2622

This PR includes a breaking change in which we will not load the `middlewares` file anymore from the user project and the `@medusajs/medusa` package. Instead, we look for `routes-config` file.

Since, the exports of both the files are same, we do not have to change much, except now I am favoring the `default export` over a named `config export`. Lemme know if you think default export is not needed.

## Upgrading projects

You must rename the `src/api/middlewares.ts` file to `src/api/routes-config.ts` file. And make the following changes to it.

```diff
- import { MiddlewareConfig } from '@medusajs/medusa'
+ import { defineRoutesConfig } from '@medusajs/medusa'

- export const config: MiddlewareConfig = {
-   routes: []
- }
+ export default defineRoutesConfig([
+
+ ])
```

The `defineRoutesConfig` method accepts an overload of arguments. 

- Either, you can pass it an array of routes. 
- Or, you may pass an object with `routes` property and the `errorHandler` properties.

```ts
defineRoutesConfig([
   // array of routes
])

defineRoutesConfig({
  errorHandler: () => {},
  routes: [
    // array of routes
  ]
})
```


thetutlage:
@shahednasser After this PR gets merged, we will have to update the docs. Related issue. https://linear.app/medusajs/issue/FRMW-2623/update-documentation-to-use-new-routes-configuration-setup

packages/medusa/src/utils/define-routes-config.ts
```diff
@@ -32,8 +32,16 @@ export function defineRoutesConfig<
       next: MedusaNextFunction
     ) => any)[]
   }
->(routes: Route[]): MiddlewaresConfig {
+>(
+  config:
+    | Route[]
+    | { routes?: Route[]; errorHandler?: MiddlewaresConfig["errorHandler"] }
```
olivermrbl, line 38: thought(non-blocking): I wonder if we should rethink the `errorHandler`, so we don't have to jam it into the API like this.

IMO, the `defineRoutesConfig` would ideally only accept `Route[]` as a top-level argument.

The `errorHandler` is used to either:
- Opt-out of our default error handler -> `errorHandler: false`
- Inject a custom `errorHandler` -> `errorHandler: () => { ... }`

Both features are not frequently requested, so we could optimize for a cleaner API. Perhaps, it can be part of a second argument to the function?

What do you think?

olivermrbl: On second thought, I am not sure if a second `options` (or similar) argument is that much better. Would love to hear what you think about this!

packages/medusa/src/utils/define-routes-config.ts
```diff
@@ -32,8 +32,16 @@ export function defineRoutesConfig<
       next: MedusaNextFunction
     ) => any)[]
   }
->(routes: Route[]): MiddlewaresConfig {
+>(
```
adrien2p, line 35: Suggestion : what about doing something similar to the defineLink and adding a symbol to the returned object such that when we import the file in the loader we dont have to check default or not and just look for that symbol in Object.values(...)?

Let me know what you think

adrien2p (approved):

packages/medusa/src/utils/define-routes-config.ts
adrien2p: IMO it is fine for now, you can either pass routes only or an object that allows you to provide an error handler

thetutlage: Is this a global error handler? If so, then we might decouple it from this `defineRoutesConfig` helper. And maybe export it as a standalone value. Something like:

```ts
export const routesConfig = defineRoutesConfig()
export const errorHandler = defineErrorHandler()
```

Not 100% confident with the above suggestion. But, if we will need a few more config values for the API server, then it might make sense to have a server config file for these sort of things.

packages/medusa/src/utils/define-routes-config.ts
thetutlage: But that means you can export the config via any named export. To be honest, feels a bit weak, since in this case the `default export` kind of acts as a Symbol + ensures consistency too.

adrien2p: I am fine with either way. It was more a suggestion to not rely on default and more be able to let the user choose if he want to default export it or just export a named const for any reason (like carrying meaning)

adrien2p: Maybe defineLink was another thing because the user is more likely to reimport and consume it and having a named export sounds better

packages/medusa/src/utils/define-routes-config.ts
shahednasser: I agree with Oli that we should not have two different signatures for `defineRoutesConfig`. I think passing a second `options` argument is a good approach, since it won't be frequently used anyway.

packages/medusa/src/utils/define-routes-config.ts
thetutlage: > let the user choose if he want to default export it or just export a named

Yeah, that's the thing. I am not sure if the freedom to choose the export names is something we need. If this becomes a conscious choice across the entire codebase, then yeah it might make sense to go with your suggestion. :)

packages/medusa/src/utils/define-routes-config.ts
thetutlage: @shahednasser I am thinking more from a fundamental point of view. Is assigning an `errorHandler` a route specific config? 

Since, we have the freedom to rethink/break APIs right now. Maybe we can give the `errorHandler` its own place :)

shahednasser: Fine with the `defineErrorHandler` approach too, my only thinking is that as Oli mentioned this feature isn't requested as much.

thetutlage: Alright. Let's see if @olivermrbl has more to add here!

thetutlage:
Going to merge this PR as it is for now! We can always extract the `errorHandler` to a separate argument or a config later :)

PR #8280, opened 2024-07-25
refactor: revert to using middlewares.ts file

thetutlage:
Reverts: 8274

Reverting the changes made in https://github.com/medusajs/medusa/pull/8274, except we will continue to use a builder function to define the config.

@srindom Do we want to go back to named exported as well. I changed to using `export default`.


srindom (approved):
LGTM

adrien2p (approved):

thetutlage:
@shahednasser Here's an example of how the final file will look like. The path remains the same as `src/api/middlewares.ts`.

```ts
import { defineMiddlewares } from '@medusajs/medusa'

export default defineMiddlewares({
  routes: [
    // routes as usual
  ],
})
```

PR #8281, opened 2024-07-25
feat: Add support for product export in UI

sradevski:
I've keep the PR simple, so it is missing column selection (we should discuss if we event want it now or in the polish stage) and push notifications (similarly, needs discussion if we want to tackle it right away).

One thing we need to improve (outside of the scope here) is to preserve query params on `handleSuccess` call, so filters and such is not lost if you eg. go to `create` and go back.

CLOSES CC-234

### DIFF
--- a/integration-tests/http/__tests__/product/admin/product-export.spec.ts
+++ b/integration-tests/http/__tests__/product/admin/product-export.spec.ts
@@ -12,7 +12,8 @@ import { ModuleRegistrationName } from "@medusajs/utils"
 jest.setTimeout(50000)
 
 const compareCSVs = async (filePath, expectedFilePath) => {
-  let fileContent = await fs.readFile(filePath, { encoding: "utf-8" })
+  const asLocalPath = filePath.replace("http://localhost:9000", process.cwd())
+  let fileContent = await fs.readFile(asLocalPath, { encoding: "utf-8" })
   let fixturesContent = await fs.readFile(expectedFilePath, {
     encoding: "utf-8",
   })
--- a/packages/modules/providers/file-local/src/services/local-file.ts
+++ b/packages/modules/providers/file-local/src/services/local-file.ts
@@ -77,7 +77,6 @@ export class LocalFileService extends AbstractFileProviderService {
     return
   }
 
-  // For private files, we simply return the file path, which can then be loaded manually by the backend.
   // The local file provider doesn't support presigned URLs for private files (i.e files not placed in /static).
   async getPresignedDownloadUrl(
     file: FileTypes.ProviderGetFileDTO
@@ -94,10 +93,6 @@ export class LocalFileService extends AbstractFileProviderService {
         MedusaError.Types.NOT_FOUND,
         `File with key ${file.fileKey} not found`
       )
-    }
-
-    if (isPrivate) {
-      return filePath
     }
 
     return this.getUploadFileUrl(file.fileKey)


packages/admin-next/dashboard/src/components/table/data-table/data-table-filter/date-filter.tsx
```diff
@@ -118,123 +119,136 @@ export const DateFilter = ({
...
-                "filters_menu_content"
-              ) {
-                e.preventDefault()
+      <DateDisplay
+        label={label}
+        value={displayValue}
+        onRemove={handleRemove}
+        readonly={readonly}
+      />
+      {!readonly && (
```
sradevski, line 128: These are just indentation changes, no actual code changed other than the readonly check.

packages/admin-next/dashboard/src/components/table/data-table/data-table-filter/data-table-filter.tsx
```diff
@@ -41,10 +41,15 @@ export type Filter = {
...
 
-export const DataTableFilter = ({ filters, prefix }: DataTableFilterProps) => {
+export const DataTableFilter = ({
+  filters,
+  readonly,
+  prefix,
+}: DataTableFilterProps) => {
   const { t } = useTranslation()
   const [searchParams] = useSearchParams()
   const [open, setOpen] = useState(false)
```
sradevski, line 55: Question: Can't we rely solely on the query parameters for the `activeFilters` state? I feel like it should be possible, and we can remove some code around maintaining their state (so basically we have getInitialFilters be the state, and that gets memoized with `location.search` passed to it)

olivermrbl:
@sradevski, this doesn't seem to work-out-of-the box with the local file provider – did you use S3? 

The exported CSV file is put into the `/private` folder, which is not served by the server as far as I can see. To be downloadable from admin, this should be placed in static no?

Update: Perhaps the issue is also related to the fact that the generated URL of the exported file points to a path on the local file system and not one on the server. Not sure if our Vite application can deal with this?

sradevski:
@olivermrbl That behavior is intentional actually. The problem with using the local file provider for the exports is that we can't have security controls (so, no support for private files) if we put it in static, as anyone would be able to read the file, so I create a `private` folder where private files would go (and you would need direct access to the machine to fetch them).

Let me know if you think this should behave differently, it's super easy to change it.

olivermrbl:
> The problem with using the local file provider for the exports is that we can't have security controls (so, no support for private files) if we put it in static, as anyone would be able to read the file

Given the local file provider is only used for development, wouldn't this be OK?

sradevski:
> > The problem with using the local file provider for the exports is that we can't have security controls (so, no support for private files) if we put it in static, as anyone would be able to read the file
> 
> Given the local file provider is only used for development, wouldn't this be OK?

I've heard people use it in production as well for some small sites, so I thought to err on the safe side. If we are OK with it since people really shouldn't use it in prod, then we can def. put it in `static`. WDYT?

You can actually configure if you want it in static or private currently (you just need to set `privateUploadDir_` to the same URL as `uploadDir_`), but it defaults to private. Maybe we can default to `static` instead?

olivermrbl:
> You can actually configure if you want it in static or private currently (you just need to set privateUploadDir_ to the same URL as uploadDir_), but it defaults to private. Maybe we can default to static instead?

Originally (in V1), the provider was supposed to be for development only, and IMO, this should still be the recommendation from our side. With that in mind, I would suggest we optimize for this use case such that we can guarantee a good first experience across all our features. 

Then in the rare (my hunch; how often have you heard this being used in prod?) case, we can guide developers to a secure configuration with the local provider.

Wdyt?

sradevski:
@olivermrbl it should definitely be only for development, but I've seen at least 2 conversations on Discord, and one from a local agency (and I obviously told them to use S3 😄).

Then let me use the same, static directory for private files as well, and leave a comment that if they really wanted to they can change it through the config, WDYT?

olivermrbl:
> Then let me use the same, static directory for private files as well, and leave a comment that if they really wanted to they can change it through the config, WDYT?

Sounds good!

olivermrbl:
@sradevski, the static folder is correctly used now for the file uploads. However, we still have the issue with the admin dashboard attempting to download from the local file system. This leads to the download containing the `index.html` because that's the fallback in case the requested file cannot be found.

olivermrbl:
@kasperkristensen, could you give this a review? 🙏

sradevski:
@olivermrbl I missed removing one check (see last commit), should be good now.

olivermrbl (approved):
LGTM!

olivermrbl:
@sradevski, works nicely – good one 👍

PR #8284, opened 2024-07-25
feat:Add confirmation step to product import

sradevski:
CLOSES CC-251


carlos-r-l-rodrigues (approved):
should we also have a `/decline`?

sradevski:
> should we also have a `/decline`?

@carlos-r-l-rodrigues we can just leave the workflow expire, I don't think we need an explicit action, WDYT? We can even make the timeout smaller (eg. 10 min)

carlos-r-l-rodrigues:
Sure, it is alright. I was just thinking about cleaning up uploaded files, and all that. But no big deal for now.

sradevski:
@carlos-r-l-rodrigues in this case there is no uploaded file, it's all in memory, since we have to process the file before we return (so we can tell them how many will be created, updated)

carlos-r-l-rodrigues:
I see... well, I believe soon we'll need to improve how workflows work with big inputs.
This has potential to cause some issues when using large files, but this is a discussion for another time.

sradevski:
Yeah I agree, I was thinking about it. Potentially we'll need to keep the state in S3/files instead for large contexts. I did try to execute the import and parsing in one step, so at least we don't keep too much context around. Let's see once we get to a point where this is a practical issue (which I am sure will happen sooner than later 😄 )

PR #8287, opened 2024-07-25
[wip] feat: Create return fulfillment

olivermrbl:


packages/core/core-flows/src/order/workflows/return/confirm-return-request.ts
```diff
@@ -85,11 +158,38 @@ export const confirmReturnRequestWorkflow = createWorkflow(
...
+    )
+
+    const fulfillment = createReturnFulfillmentStep(fulfillmentData)
+
+    const link = transform(
+      { order_id: order.id, fulfillment: fulfillment },
+      (data) => {
+        return [
+          {
+            [Modules.ORDER]: { order_id: data.order_id },
```
olivermrbl, line 184: self-review: Will change this to link to the return entity instead

PR #8291, opened 2024-07-26
feat(core-flows, fulfillment): validate shipping options to stock locations

riqwan:
what:

- validate shipping options to stock locations

RESOLVES CC-232

### DIFF
--- a/integration-tests/http/__tests__/returns/returns.spec.ts
+++ b/integration-tests/http/__tests__/returns/returns.spec.ts
@@ -216,6 +216,14 @@ medusaIntegrationTestRunner({
           },
         },
         {
+          [Modules.STOCK_LOCATION]: {
+            stock_location_id: location.id,
+          },
+          [Modules.FULFILLMENT]: {
+            fulfillment_provider_id: "manual_test-provider",
+          },
+        },
+        {
           [Modules.SALES_CHANNEL]: {
             sales_channel_id: "test",
           },
--- a/integration-tests/modules/__tests__/cart/store/cart.workflows.spec.ts
+++ b/integration-tests/modules/__tests__/cart/store/cart.workflows.spec.ts
@@ -1480,6 +1480,36 @@ medusaIntegrationTestRunner({
         })
 
         it("should add shipping method to cart", async () => {
+          const stockLocation = (
+            await api.post(
+              `/admin/stock-locations`,
+              { name: "test location" },
+              adminHeaders
+            )
+          ).data.stock_location
+
+          await remoteLink.create([
+            {
+              [Modules.STOCK_LOCATION]: {
+                stock_location_id: stockLocation.id,
+              },
+              [Modules.FULFILLMENT]: {
+                fulfillment_set_id: fulfillmentSet.id,
+              },
+            },
+          ])
+
+          await remoteLink.create([
+            {
+              [Modules.STOCK_LOCATION]: {
+                stock_location_id: stockLocation.id,
+              },
+              [Modules.FULFILLMENT]: {
+                fulfillment_provider_id: "manual_test-provider",
+              },
+            },
+          ])
+
           const shippingOption = await fulfillmentModule.createShippingOptions({
             name: "Test shipping option",
             service_zone_id: fulfillmentSet.service_zones[0].id,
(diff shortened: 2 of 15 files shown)


integration-tests/http/__tests__/shipping-option/admin/shipping-option.spec.ts
```diff
@@ -331,6 +406,73 @@ medusaIntegrationTestRunner({
             })
           )
         })
+
+        it.only("should throw an error when provider does not belong to service location", async () => {
```
sradevski:
```suggestion
        it("should throw an error when provider does not belong to service location", async () => {
```

integration-tests/http/__tests__/shipping-option/admin/shipping-option.spec.ts
```diff
@@ -359,6 +501,22 @@ medusaIntegrationTestRunner({
             rules: [shippingOptionRule],
           }
 
+          const remoteLink = appContainer.resolve(
+            ContainerRegistrationKeys.REMOTE_LINK
+          )
+
+          // TODO: move this to an endpoint when available
+          await remoteLink.create([
```
sradevski, line 509: suggestion: Maybe create a helper func so we have one place to update when the endpoint is available

packages/core/core-flows/src/fulfillment/steps/validate-fulfillment-providers.ts
```diff
@@ -0,0 +1,117 @@
...
+  "validate-fulfillment-providers-step"
+export const validateFulfillmentProvidersStep = createStep(
+  validateFulfillmentProvidersStepId,
+  async (input: FulfillmentProviderValidationInput[], { container }) => {
+    const dataToValidate: {
+      service_zone_id: string
+      provider_id: string
+    }[] = []
+    const shippingOptionIds = input.map((d) => d.id).filter(Boolean) as string[]
+    const fulfillmentService = container.resolve(
```
sradevski, line 25: Super NIT: maybe move `shippingOptionsIds` below `fulfillmentService`, it's a bit easier to read 😄. Maybe even passing the map to `id: input.map...` would be fine as well, WDYT?

packages/core/core-flows/src/fulfillment/steps/validate-fulfillment-providers.ts
```diff
@@ -0,0 +1,117 @@
...
+    )
+    const shippingOptions = await fulfillmentService.listShippingOptions({
+      id: shippingOptionIds,
+    })
+
+    const shippingOptionsMap = new Map<string, ShippingOptionDTO>(
+      shippingOptions.map((so) => [so.id, so])
+    )
+
+    const invalidProviders: string[] = []
```
sradevski: Similarly small - move the declaration of `invalidProviders` further down where it is used

packages/core/core-flows/src/fulfillment/steps/validate-fulfillment-providers.ts
```diff
@@ -0,0 +1,117 @@
...
+          provider_id:
+            data.provider_id! || existingShippingOption?.provider_id!,
+        })
+
+        continue
+      }
+
+      if (data.service_zone_id && data.provider_id) {
+        dataToValidate.push({
+          service_zone_id: data.service_zone_id!,
```
sradevski: I would have imagined the `!` to not be necessary here?

sradevski (approved):
LGTM! Welcome back btw 😄

integration-tests/http/__tests__/shipping-option/admin/shipping-option.spec.ts
riqwan: I've already replaced these with the endpoint in the new branch i'm working on 🤞🏻

olivermrbl (approved):
LGTM, nice work. 

A minor thing on UI-side (for a follow-up PR): I believe we also need to filter out selectable providers in the UI when creating and updating shipping options. We should only show the providers that are associated with the location for which you are currently manage shipping options.

riqwan:
@olivermrbl I've created a ticket for it in the project 🤞🏻

riqwan:
@olivermrbl this PR is essentially a breaking change for shipping options, do we have a way to communicate this?

PR #8292, opened 2024-07-26
docs: added a service factory reference

shahednasser:
Added a service factory reference of the generated methods and how to use them.

Since the types in the codebase are a bit complex, a generated reference won't provide a valuable result.

Instead, this reference is created manually and aims to cover common use cases, providing examples for them.

Closes DOCS-844


olivermrbl (approved):
This is great @shahednasser – essentially, what I talked about the other day. 

Minor nit and totally non-blocking: We could change `customModuleService` to `postModuleService` now that we are working with a `post` model.

shahednasser:
@olivermrbl got point! Will make the change 👍🏻

PR #8293, opened 2024-07-26
Cart API should also return billing_address.province by default

josetr:
Shouldn't we also return `shipping_address.company` / `billing_address.company` by default?


sradevski (approved):
LGTM, thanks! I think the default fields will get revised as we get closer to the public v2 release, but yes, company can also be included.

josetr:
I've added the company fields. I hope you don't mind.

By the way, is there any info regarding adding back the ability to update draft orders?

https://github.com/medusajs/medusa/blob/develop/packages/medusa/src/api/admin/draft-orders/%5Bid%5D/route.ts

I've been trying to wait patiently, tho I don't know how important it is for you guys right now...

sradevski:
@josetr we are actively working on orders, currently on exchanges, claims and returns. We'll be adding missing functionalities in the next few weeks hopefully

josetr:
Alright, thanks for the info and merging the PR.

PR #8296, opened 2024-07-26
docs-util: add docblock generator for models built with DML

shahednasser:
Add to the docblock generator tool a new `dml` mode that generates JSON files for data models created using DML. These JSON files hold the description of each property, avoiding to stuff the actual model source code with TSDocs.

An example of the generated JSON output for the Region Module:

```json
{
	"Region": {
		"filePath": "packages/modules/region/src/models/region.ts",
		"properties": {
			"id": "The ID of the region.",
			"name": "The name of the region",
			"currency_code": "The currency code of the region",
			"automatic_taxes": "Whether the region automatic taxes.",
			"countries": "The associated countries.",
			"metadata": "Holds custom data in key-value pairs."
		}
	},
	"Country": {
		"filePath": "packages/modules/region/src/models/country.ts",
		"properties": {
			"iso_2": "The iso 2 of the country",
			"iso_3": "The iso 3 of the country",
			"num_code": "The num code of the country",
			"name": "The name of the country",
			"display_name": "The display name of the country",
			"region": "The associated region.",
			"metadata": "Holds custom data in key-value pairs."
		}
	}
}
```

(The process of updating the descriptions to be clearer is manual, similar to other types of docs that we generate)

These generated files will then be utilized by a Typedoc plugin to attach the descriptions to the data models and their properties when generating the references. This will be part of another PR.

## Usage

In the `www/utils` directory, run the following command:

```bash
yarn generate:dml
```

It generates the files under the `www/utils/generated/dml-output` directory.

## Changes in PR

1. Add `dml` kind generator that detects a data model in the source code to generate the JSON object for it. It can both create and update the JSON objects.
2. Add a `clean:dml` command that cleans up the generated DML JSON files in case a data model is deleted.
3. Add a new action that generates the DML once a day and on release.

Closes DOCS-849

### DIFF
--- a/www/utils/turbo.json
+++ b/www/utils/turbo.json
@@ -20,6 +20,11 @@
       "dependsOn": [
         "^build"
       ]
+    },
+    "generate:dml": {
+      "dependsOn": [
+        "^build"
+      ]
     }
   }
 }


shahednasser:

olivermrbl (approved):

PR #8298, opened 2024-07-26
feat: Add support for product imports

sradevski:
Depends on #8281 (will retarget to develop once that is merged)

Designs: https://www.figma.com/design/z3aUuOVWUKmdHH0ofmMpEV/Web-app-3.0?node-id=6612-67477&t=FQjXXVUyAbMvFhfh-0

CLOSES CC-253


packages/admin-next/dashboard/src/components/common/file-preview/file-preview.tsx
```diff
@@ -0,0 +1,130 @@
...
+              <ArrowDownTray />
+            </a>
+          </IconButton>
+        )}
+      </div>
+    </div>
+  )
+}
+
+const FileThumbnail = () => {
```
sradevski, line 60: This is not really an icon, so I wasn't sure how to handle it. I think having it defined here is fine, but let me know if another place would be better.

packages/admin-next/dashboard/src/routes/products/product-import/product-import.tsx
```diff
@@ -15,7 +21,119 @@ export const ProductImport = () => {
...
+          {t("products.import.template.title")}
+        </Heading>
+        <Text size="small" className="text-ui-fg-subtle">
+          {t("products.import.template.description")}
+        </Text>
+        <div className="mt-4">
+          <FilePreview
+            filename={"medusa-template-product-list.csv"}
+            // TODO: Where would the template file be stored?
+            url={"https://example.com"}
```
sradevski: @olivermrbl where did we store the template for the imports before?

olivermrbl: We constructed and downloaded the file on the fly:
https://github.com/medusajs/medusa/blob/v1.x/packages/admin-ui/ui/src/domain/products/batch-job/download-template.ts

sradevski: Ok cool, let's do that, seems like the simpler option anyway.

sradevski: Done

riqwan (approved):
lgtm 🤞🏻

PR #8302, opened 2024-07-26
feat: update return request

olivermrbl:
What
- Add `updateReturnWorkflow`
- Add `POST /admin/returns/:id`
- Add `onChange` handler for `location_id` in return request flow in admin

Closes CC-183

### DIFF
--- a/packages/medusa/src/api/admin/returns/[id]/route.ts
+++ b/packages/medusa/src/api/admin/returns/[id]/route.ts
@@ -1,3 +1,4 @@
+import { updateReturnWorkflow } from "@medusajs/core-flows"
 import {
   ContainerRegistrationKeys,
   remoteQueryObjectFromString,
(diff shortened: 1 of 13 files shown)


packages/medusa/src/api/admin/returns/[id]/route.ts
```diff
@@ -32,3 +34,34 @@ export const GET = async (
     return: orderReturn,
   })
 }
+
+export const POST = async (
```
Comment (carlos-r-l-rodrigues) [inline packages/medusa/src/api/admin/returns/[id]/route.ts:38]: I thought we had this endpoint hehe.

carlos-r-l-rodrigues (approved):

carlos-r-l-rodrigues:
@olivermrbl can I merge that to make the changes in the other PR [8304](https://github.com/medusajs/medusa/pull/8304)?

PR #8303, opened 2024-07-26
feat(workflows-sdk): createWorkflowWithHooks

adrien2p:
In this proposal, we are providing one approach to support hooks and another tested approach which are

**Approach 1: new util**

```ts
type MyWorkflowHooks = {
  beforeCreate: (id: string) => string
}

const workflowWithHooks = createWorkflowWithHooks(
  (useHook: UseHook<MyWorkflowHooks>) => {
    return createWorkflow(
      "sub-workflow",
      function (input: WorkflowData<{ outsideWorkflowData: string }>) {
        step1()
        step3()
        const res = useHook("beforeCreate", "id")
        return step2({ filters: { id: input.outsideWorkflowData } })
      }
    )
  }
)

// And here is how to assign a new handler to the hook. Not yet steps
workflowWithHooks.beforeCreate((id) => {
  return "test"
})
```

**Approach 2: using existing util**
In my opinion this one seems less good than the previous one but I ll share it

```ts
type MyWorkflowHooks = {
  beforeCreate: (id: string) => string
}

createWorkflow(
  "sub-workflow",
  function (input: WorkflowData<{ outsideWorkflowData: string }>, { useHook }: { useHook: UseHook<MyWorkflowHooks> }) {
    step1()
    step3()
    const res = useHook("beforeCreate", "id")
    return step2({ filters: { id: input.outsideWorkflowData } })
  }
)

// And here is how to assign a new handler to the hook. Not yet steps
workflowWithHooks.beforeCreate((id) => {
  return "test"
})
```

This gives the same result as the proposal in this PR

But I am not sure this is more readable and usable than having a proper util 🤔 also, it is very likely than the workflows SDK will be re written in a better way in order to make it easier to enrich the feature and maintaining it.

for my own ref, here is the changes the second example possible:

```ts
export function createWorkflow<TData, TResult, THooks extends Hooks>(
  /**
   * The name of the workflow or its configuration.
   */
  nameOrConfig: string | ({ name: string } & TransactionModelOptions),
  /**
   * The constructor function that is executed when the `run` method in {@link ReturnWorkflow} is used.
   * The function can't be an arrow function or an asynchronus function. It also can't directly manipulate data.
   * You'll have to use the {@link transform} function if you need to directly manipulate data.
   */
  composer: (
    input: WorkflowData<TData>,
    { useHook }: { useHook: UseHook<THooks> }
  ) =>
    | void
    | WorkflowData<TResult>
    | {
        [K in keyof TResult]:
          | WorkflowData<TResult[K]>
          | WorkflowDataProperties<TResult[K]>
          | TResult[K]
      }
): ReturnWorkflow<TData, TResult> & DefineHooks<THooks> { /*...*/ }
```

### DIFF
--- a/packages/core/workflows-sdk/src/utils/_playground.ts
+++ b/packages/core/workflows-sdk/src/utils/_playground.ts
@@ -1,8 +1,10 @@
 import {
   createStep,
   createWorkflow,
+  createWorkflowWithHooks,
   StepResponse,
   transform,
+  UseHook,
   when,
   WorkflowData,
 } from "./composer"
@@ -42,9 +44,9 @@ const workflow2 = createWorkflow("workflow", function () {
     console.log(step3Res)
   })
 
-  const workflowRes = workflow.asStep({ outsideWorkflowData: step1Res.step1 })
+  /*const workflowRes = workflow.runAsStep({ outsideWorkflowData: step1Res.step1 })*/
 
-  return workflowRes
+  return step3Res
 })
 
 workflow2()
@@ -52,6 +54,28 @@ workflow2()
   .then((res) => {
     console.log(res.result)
   })
+
+type MyWorkflowHooks = {
+  beforeCreate: (id: string) => string
+}
+
+const workflowWithHooks = createWorkflowWithHooks(
+  (useHook: UseHook<MyWorkflowHooks>) => {
+    return createWorkflow(
+      "sub-workflow",
+      function (input: WorkflowData<{ outsideWorkflowData: string }>) {
+        step1()
+        step3()
+        const res = useHook("beforeCreate", "id")
+        return step2({ filters: { id: input.outsideWorkflowData } })
+      }
+    )
+  }
+)
+
+workflowWithHooks.beforeCreate((id) => {
+  return "test"
+})
 
 /*const step1 = createStep("step1", async (input: {}, context) => {
   return new StepResponse({ step1: ["step1"] })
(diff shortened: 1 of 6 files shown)


packages/core/workflows-sdk/src/utils/_playground.ts
```diff
@@ -53,6 +55,28 @@ workflow2()
     console.log(res.result)
   })
 
+type MyWorkflowHooks = {
+  beforeCreate: (id: string) => string
+}
+
+const workflowWithHooks = createWorkflowWithHooks(
```
adrien2p, line 62: hint: We are able here to maintain the chain of type coming from the inner workflow, which means the inputs are still typed etc

packages/core/workflows-sdk/src/utils/_playground.ts
```diff
@@ -53,6 +55,28 @@ workflow2()
...
+  (useHook: UseHook<MyWorkflowHooks>) => {
```
adrien2p, line 63: hint: This allows us to wrap the hook util under the hook and it fully typed, the first argument being the hook name coming from the type define above `MyWorkflowHooks ` and the argument is also typed based on the above type `MyWorkflowHooks`

packages/core/workflows-sdk/src/utils/_playground.ts
```diff
@@ -53,6 +55,28 @@ workflow2()
...
+        step1()
+        step3()
+        const res = useHook("beforeCreate", "id")
+        return step2({ filters: { id: input.outsideWorkflowData } })
+      }
+    )
+  }
+)
+
+workflowWithHooks.beforeCreate((id) => {
```
adrien2p, line 76: hint: here the handler is also fully typed while maintaining the original workflow types for the arguments

adrien2p:
Close in favour of a separate pr

PR #8304, opened 2024-07-26
chore(core-flows): return fulfillment link

carlos-r-l-rodrigues:


packages/core/core-flows/src/order/workflows/return/confirm-return-request.ts
```diff
@@ -36,6 +47,69 @@ const validationStep = createStep(
...
+      line_item_id: i.item_id,
+      quantity: i.quantity,
+      return_quantity: i.quantity,
+      title: orderItem.variant_title ?? orderItem.title,
+      sku: orderItem.variant_sku || "",
+      barcode: orderItem.variant_barcode || "",
+    } as FulfillmentWorkflow.CreateFulfillmentItemWorkflowDTO
+  })
+
+  const locationId =
```
olivermrbl, line 86: With #8302 merged, we can rely on `return.location_id` here

packages/modules/link-modules/src/definitions/order-return-fulfillment.ts
```diff
@@ -0,0 +1,67 @@
...
+      args: {
+        methodSuffix: "Fulfillments",
+      },
+    },
+  ],
+  extends: [
+    {
+      serviceName: Modules.ORDER,
+      fieldAlias: {
+        return_fulfillments: {
```
olivermrbl, line 44: thought: I wasn't expecting returns to have multiple fulfillments. What's your thinking?

carlos-r-l-rodrigues: That's just how it is desined on the backend.
in theory it could be split in multiple shipments, even if we have many places where multiple shipping options are not handled properly.
Should I just remove the `isList`?

olivermrbl: Got it. Let's just keep it as is for now 👍

carlos-r-l-rodrigues: current there is an issue expanding this link, I will investigate it later and possibly push a fix in another PR.

packages/core/core-flows/src/order/workflows/return/confirm-return-request.ts
carlos-r-l-rodrigues: I will be able to remove a request after the changes are applied, but here I can keep like this as we are also using the address

packages/modules/link-modules/src/definitions/order-return-fulfillment.ts
carlos-r-l-rodrigues: @olivermrbl At the end, we need the shipping option from the shipping method created, so the location_id of the return at this point is not super relevant.
Unless we want to validate that maybe?
if the shipping_option of the shipping method is in a different location_id of the return

olivermrbl (approved):
LGTM

PR #8312, opened 2024-07-27
Chore/order claims 2

carlos-r-l-rodrigues:
What:
- Begin Order Claim
- Add return items
- Claim items
- Add additional items
- Request Order Claim

Missing:
- complete http tests
- cancel claim request
- cancel claim

### DIFF
--- a/packages/admin-next/dashboard/src/components/inputs/country-select/country-select.tsx
+++ b/packages/admin-next/dashboard/src/components/inputs/country-select/country-select.tsx
@@ -12,56 +12,67 @@ import { countries } from "../../../lib/data/countries"
 
 export const CountrySelect = forwardRef<
   HTMLSelectElement,
-  ComponentPropsWithoutRef<"select"> & { placeholder?: string }
->(({ className, disabled, placeholder, ...props }, ref) => {
-  const { t } = useTranslation()
-  const innerRef = useRef<HTMLSelectElement>(null)
+  ComponentPropsWithoutRef<"select"> & {
+    placeholder?: string
+    value?: string
+    defaultValue?: string
+  }
+>(
+  (
+    { className, disabled, placeholder, value, defaultValue, ...props },
+    ref
+  ) => {
+    const { t } = useTranslation()
+    const innerRef = useRef<HTMLSelectElement>(null)
 
-  useImperativeHandle(ref, () => innerRef.current as HTMLSelectElement)
+    useImperativeHandle(ref, () => innerRef.current as HTMLSelectElement)
 
-  const isPlaceholder = innerRef.current?.value === ""
+    const isPlaceholder = innerRef.current?.value === ""
 
-  return (
-    <div className="relative">
-      <TrianglesMini
-        className={clx(
-          "text-ui-fg-muted transition-fg pointer-events-none absolute right-2 top-1/2 -translate-y-1/2",
-          {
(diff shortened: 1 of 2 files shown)


integration-tests/http/__tests__/claims/claims.spec.ts
```diff
@@ -0,0 +1,528 @@
...
+              quantity: 1,
+            },
+          ],
+        },
+        adminHeaders
+      )
+    })
+
+    describe("Returns lifecycle", () => {
+      it.only("Full flow with 2 orders", async () => {
```
olivermrbl, line 387:
```suggestion
    describe("Claims lifecycle", () => {
      it("Full flow with 2 orders", async () => {
```

packages/core/core-flows/src/order/steps/create-claim-items.ts
```diff
@@ -0,0 +1,59 @@
...
+}
+
+export const createOrderClaimItemsStep = createStep(
+  "create-claim-items",
+  async (input: CreateOrderClaimItemsInput, { container }) => {
+    const orderModuleService = container.resolve<IOrderModuleService>(
+      ModuleRegistrationName.ORDER
+    )
+
+    const claimItems = input.changes.map((item) => {
```
olivermrbl, line 17: question: Should we put this in a transformer in the workflow?

packages/core/core-flows/src/order/workflows/claim/confirm-claim-request.ts
```diff
@@ -0,0 +1,361 @@
...
+    orderClaim: OrderClaimDTO
+    orderChange: OrderChangeDTO
+  }) {
+    throwIfIsCancelled(order, "Order")
+    throwIfIsCancelled(orderClaim, "Claim")
+    throwIfOrderChangeIsNotActive({ orderChange })
+  }
+)
+
+function prepareFulfillmentData({
```
olivermrbl, line 50: thought: Think this is 1:1 with the other `prepareFulfillmentData` – if so, we could extract to a utility

packages/core/core-flows/src/order/workflows/return/confirm-return-request.ts
```diff
@@ -111,6 +99,31 @@ function prepareFulfillmentData({
...
+
+    for (const action of modifiedShippingMethod_.actions) {
+      if (
+        action.action === ChangeActionType.SHIPPING_ADD &&
+        action.return_id === orderReturn.id
+      ) {
+        returnShippingMethod = shippingMethod
+        break
+      }
+    }
```
olivermrbl, line 122: question: If there is only one, we can just use `.find()` on the actions. Think that's a bit cleaner no?

packages/medusa/src/api/admin/claims/[id]/inbound/items/[action_id]/route.ts
```diff
@@ -10,6 +10,7 @@ import {
   AuthenticatedMedusaRequest,
```
Comment (olivermrbl) [inline packages/medusa/src/api/admin/claims/[id]/inbound/items/[action_id]/route.ts:10]: question: Shouldn't we return the claim as well in all the `/claims` Rest APIs? It seems we are doing it in some but not all.

olivermrbl (approved):
Overall LGTM

carlos-r-l-rodrigues:
@olivermrbl , I'm addressing you comments and a few more issues in the next PR in a few minutes.

PR #8316, opened 2024-07-28
feat:Allow updating prices through product update workflow

sradevski:
CLOSES CC-250

### DIFF
--- a/packages/core/core-flows/src/product/workflows/update-products.ts
+++ b/packages/core/core-flows/src/product/workflows/update-products.ts
@@ -67,7 +67,7 @@ function prepareUpdateProductInput({
     update: {
       ...input.update,
       sales_channels: undefined,
-      variants: input.update.variants?.map((v) => ({
+      variants: input.update?.variants?.map((v) => ({
         ...v,
         prices: undefined,
       })),
--- a/packages/core/core-flows/src/product/workflows/upsert-variant-prices.ts
+++ b/packages/core/core-flows/src/product/workflows/upsert-variant-prices.ts
@@ -1,4 +1,8 @@
-import { CreateMoneyAmountDTO, ProductTypes } from "@medusajs/types"
+import {
+  CreatePricesDTO,
+  UpdatePricesDTO,
+  CreatePriceSetDTO,
+} from "@medusajs/types"
 import { Modules, arrayDifference } from "@medusajs/utils"
 import {
   WorkflowData,
@@ -13,7 +17,7 @@ type WorkflowInput = {
   variantPrices: {
     variant_id: string
     product_id: string
-    prices?: CreateMoneyAmountDTO[]
+    prices?: (CreatePricesDTO | UpdatePricesDTO)[]
   }[]
   previousVariantIds: string[]
 }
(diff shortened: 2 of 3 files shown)


packages/core/core-flows/src/product/steps/wait-confirmation-product-import.ts
```diff
@@ -7,7 +7,7 @@ export const waitConfirmationProductImportStep = createStep(
     name: waitConfirmationProductImportStepId,
     async: true,
     // After an hour we want to timeout and cancel the import so we don't have orphaned workflows
-    timeout: 60 * 60 * 1 * 1000,
```
sradevski, line 10: I thought this is in ms, but it's in seconds

packages/core/core-flows/src/product/workflows/upsert-variant-prices.ts
```diff
@@ -0,0 +1,114 @@
...
+        data.input.variantPrices.map((v) => v.variant_id)
+      )
+    })
+
+    removeRemoteLinkStep({
+      [Modules.PRODUCT]: { variant_id: removedVariantIds },
+    }).config({ name: "remove-variant-link-step" })
+
+    const { newVariants, existingVariants } = transform({ input }, (data) => {
+      const previousMap = new Map(
```
riqwan: sugg: you can use a set here instead

packages/core/core-flows/src/product/workflows/upsert-variant-prices.ts
```diff
@@ -0,0 +1,114 @@
...
+} from "@medusajs/workflows-sdk"
+import { removeRemoteLinkStep, useRemoteQueryStep } from "../../common"
+import { createPriceSetsStep, updatePriceSetsStep } from "../../pricing"
+import { createVariantPricingLinkStep } from "../steps"
+
+type WorkflowInput = {
+  variantPrices: {
+    variant_id: string
+    product_id: string
+    prices?: CreateMoneyAmountDTO[]
```
riqwan: q: shouldn't this be Create or UpdateMoneyAmountDTO?

sugg: use `CreatePricesDTO` instead, we should wipe the money amount types eventually.

packages/core/core-flows/src/product/workflows/update-products.ts
```diff
@@ -148,7 +212,7 @@ export const updateProductsWorkflow = createWorkflow(
   (
     input: WorkflowData<WorkflowInput>
   ): WorkflowData<ProductTypes.ProductDTO[]> => {
-    // TODO: Delete price sets for removed variants
+    const previousVariantIds = getVariantIdsForProductsStep(input)
```
riqwan, line 215: sugg: you can use `useRemoteQueryStep` directly here and a transform to get the ids

riqwan (approved):
Added some comments, overall lgtm! 🤞🏻

packages/core/core-flows/src/product/workflows/update-products.ts
olivermrbl: +1, if possible

sradevski: I was gonna do it as well, but it will require a transform and useRemoteQueryStep, and I found creating a new step just cleaner overall (it's a big workflow, so it will clutter it more). WDYT?

sradevski: I'll just have this merged as-is so it doesn't linger for too long, but if you think we should still go with useRemoteQueryStep I'll just handle it in a separate PR

PR #8334, opened 2024-07-29
feat(orchestration,workflows-sdk): Skip step

carlos-r-l-rodrigues:
What:
 - Transaction Orchestrator `skipStep`
 - new response object for workflows-sdk: `StepResponse.skip()`


adrien2p (approved):

adrien2p:
As discussed we can add it to the WHEN when it is false to have the info it has been skipped

adrien2p:
CC @thetutlage we can return `StepResponse.skip` from our placeholder step once this pr is merged 👍 
ref: feat/workflow-hooks

thetutlage:
What's the purpose of skip? Is it to ignore the return value or to create a dummy step without any invokeFn?

adrien2p:
> What's the purpose of skip? Is it to ignore the return value or to create a dummy step without any invokeFn?

Mostly to have its real state and being able to display it properly. Instead of marking it as run and success but nothing really ran. Same when you apply the when conditional util

olivermrbl (approved):

PR #8336, opened 2024-07-29
fix(dashboard,types,js-sdk): Cleanup `settings/store`

kasperkristensen:
**What**
- Cleans up Store domain of all TS errors
- Adds layout component to Store domain
- Adds currencies types and js-sdk methods
- Fixes a bug that caused Table rows to render incorrectly when takings up more then the viewport height.


sradevski (approved):
LGTM 👍

PR #8341, opened 2024-07-29
fix: Added more tests and fixed a couple of issues with product import

sradevski:
The git tree has deepened a bit, but I expect to merge all of these soon (so please review 😄 )
CLOSES CC-249


packages/core/core-flows/src/product/workflows/import-products.ts
```diff
@@ -30,6 +30,23 @@ export const importProductsWorkflow = createWorkflow(
 
     waitConfirmationProductImportStep()
 
+    // Q: Can we somehow access the error from the step that threw here? Or in a compensate step at least?
```
olivermrbl, line 33: cc @carlos-r-l-rodrigues

packages/core/core-flows/src/product/helpers/normalize-for-import.ts
```diff
@@ -148,10 +151,19 @@ const normalizeVariantForImport = (
           { currency_code: priceKey.toLowerCase(), amount: normalizedValue },
         ]
       } else {
+        const region = regionsMap.get(priceKey)
+        if (!region) {
+          throw new MedusaError(
+            MedusaError.Types.INVALID_DATA,
+            `Region with ID ${priceKey} not found`
+          )
+        }
```
olivermrbl, line 160: question: Do we want to change this to region name now, or would you rather wait for a follow-up PR?

packages/core/core-flows/src/notification/steps/notify-on-failure.ts
```diff
@@ -0,0 +1,36 @@
...
+
+export const notifyOnFailureStepId = "notify-on-failure"
+export const notifyOnFailureStep = createStep(
+  notifyOnFailureStepId,
+  async (data: NotifyOnFailureStepInput) => {
+    return new StepResponse(void 0, data)
+  },
+
+  async (data, { container }) => {
+    console.log("HERE!!!")
```
olivermrbl: todo: The good ol' 😄 
```suggestion
```

sradevski: Ahaha classic 🤦 we should probably add a console.log linter...

packages/core/core-flows/src/product/helpers/normalize-for-import.ts
sradevski: I already created a separate ticket, so in a follow up PR

packages/core/core-flows/src/product/workflows/import-products.ts
sradevski: Merged it but this is still an open question (created a Linear ticket to track it)

carlos-r-l-rodrigues: I'm not sure I follow here what is needed here. Something can fail but the workflow will continue to be executed?

At the moment using this workflows api we are not exposing anything to access other steps responses/errors.
[Here is the context](https://github.com/medusajs/medusa/blob/24c105f28837a1190da35da1335b6b3cedc8eed0/packages/core/workflows-sdk/src/utils/composer/create-step.ts#L130-L143) we are proving to the step, but main invoke argument `stepArguments` has everything.

@adrien2p @olivermrbl I'm fine with exposing more properties in this context, like the transaction and the invoke and compensation responses. At least people could find their way if needed. Or we think about exposing compensation responses, transaction errors or transaction/step states/status.

adrien2p: we could eventually wrap it under another getter, something like `getErrorFromStep(stepName)` and so on if we want to make it easy for them to discover and find those kind of information?

PR #8346, opened 2024-07-30
breaking: implement workflow hooks (first iteration)

thetutlage:
Fixes: FRMW-2629

This PR implements the workflow hooks proposal as defined in the [linked](https://www.notion.so/medusajs/RFC-workflow-hooks-fa18841459a74bbc916a293092a3e487) RFC.

---

## ChangeLog

This PR adds a new helper, `createHook`, to the `workflows-sdk`. The `createHook` helper exposes a hook from a workflow. Later (after the workflow has been composed), the workflow consumers can bind a handler to the hook to run custom logic.

> [!NOTE]
> As of now, you can only register one hook handler, and the workflow orchestrator ignores the return value.

### Exposing hook via `createHook`

```ts
import {
  createStep,
  createHook,
  createWorkflow,
  WorkflowResponse
} from '@medusajs/workflows-sdk'

const createProductStep = createStep('createProduct', function () {
  // business logic for "createProduct"
})

const workflow = createWorkflow('name', function (input) {
  createProductStep(input)
  const productCreatedHook = createHook('productCreated', { productId: input.id })

  return new WorkflowResponse(input, {
    hooks: [productCreatedHook]
  })
})
```

**Points to note**

- Unlike the `createStep` function, the `createHook` method is called within the workflow composition callback.
- You must return the created hooks from the composition callback. Returning of hooks is needed for the TypeScript engine to provide intellisense when registering a handler for the hook.
- Hooks are executed in the same position as they are defined within the composition callback

### Registering the hook handler
The workflow user must register a hook handler to run custom logic within a workflow. They can do that as follows.

```ts
workflow.hooks.productCreated(() => {
  // run custom business logic
})
```

**Points to note**

- The hook handler behaves similarly to a workflow step. It can run async code, will be retried, and can also have compensation behavior (defined as the 2nd parameter)
- There can only be one handler for a hook. If you need multiple handlers, you should create another workflow and register that as the hook handler (not supported yet).
- The return value of the hook handler is ignored in this first iteration.

### Introducing the `WorkflowResponse` class and breaking changes

The introduction of hooks has changed the return value of the `createWorkflow` composition callback. Now, we must return both the workflow results and the configured hooks.

Instead of manually composing the return value, you can use the `WorkflowResponse` class to construct the current response. The `WorkflowResponse` class accepts the following parameters.

- The first parameter is the result of the workflow.
- The second parameter (optional) is a config object with configured hooks.


packages/core/core-flows/src/sales-channel/workflows/create-sales-channels.ts
```diff
@@ -1,13 +1,21 @@
...
   createSalesChannelsWorkflowId,
-  (input: WorkflowData<WorkflowInput>): WorkflowData<SalesChannelDTO[]> => {
-    return createSalesChannelsStep({ data: input.salesChannelsData })
+  (
+    input: WorkflowData<WorkflowInput>
+  ): WorkflowResponse<WorkflowData<SalesChannelDTO[]>> => {
+    return new WorkflowResponse(
+      createSalesChannelsStep({ data: input.salesChannelsData })
+    )
   }
```
thetutlage: This is how all Workflow return values will look like. @adrien2p @olivermrbl @srindom

adrien2p: yes I believe so, I would just split it in two steps but no strong opinion
```ts
const salesChannel = createSalesChannelsStep({ data: input.salesChannelsData })
return new WorkflowResponse(salesChannel)
```

thetutlage: Lol. I prefer wrapping for simple use-case :). But yeah, no strong opinions either

thetutlage: Looping in @shahednasser as well for the docs.

integration-tests/modules/__tests__/promotion/admin/campaigns.spec.ts
```diff
@@ -86,7 +86,7 @@ const adminHeaders = {
 medusaIntegrationTestRunner({
   env,
   testSuite: ({ dbConnection, getContainer, api }) => {
-    describe("Admin Campaigns API", () => {
+    describe.only("Admin Campaigns API", () => {
```
adrien2p: todo: rm

packages/core/core-flows/src/definition/cart/workflows/complete-cart.ts
```diff
@@ -20,7 +21,7 @@ import { confirmVariantInventoryWorkflow } from "./confirm-variant-inventory"
 export const completeCartWorkflowId = "complete-cart"
 export const completeCartWorkflow = createWorkflow(
   completeCartWorkflowId,
-  (input: WorkflowData<any>): WorkflowData<OrderDTO> => {
+  (input: WorkflowData<any>): WorkflowResponse<WorkflowData<OrderDTO>> => {
```
adrien2p: Q: should we wrap WorkflowData as part or WorkflowResponse such that you don't have to provide both always?

packages/core/workflows-sdk/src/utils/composer/type.ts
```diff
@@ -73,18 +85,20 @@ export type WorkflowData<T = unknown> = (T extends Array<infer Item>
...
   handlers: WorkflowHandler
   stepBinder: <TOutput = unknown>(
     fn: StepFunctionResult
   ) => WorkflowData<TOutput>
-  hookBinder: <TOutput = unknown>(
-    name: string,
-    fn: Function
-  ) => WorkflowData<TOutput>
+  // TODO change the typings
+  hookBinder: (name: string, fn: Function) => any
```
adrien2p: suggestion: should we update the types here? should return void for example

packages/core/workflows-sdk/src/utils/composer/hook.ts
```diff
@@ -107,8 +107,8 @@ export function hook<TOutput>(
     ] as CreateWorkflowComposerContext
   ).hookBinder
 
-  return hookBinder(name, function (context) {
-    return {
+  return hookBinder(name, function () {
+    /*return {
```
adrien2p: todo: we can probably rm this file entirely and also update the tests

packages/core/workflows-sdk/src/utils/composer/helpers/resolve-value.ts
```diff
@@ -5,12 +5,14 @@ async function resolveProperty(property, transactionContext) {
 
   if (property?.__type === OrchestrationUtils.SymbolInputReference) {
     return transactionContext.payload
+  } else if (
```
adrien2p, line 8: nit: wdyt to move to switch case to make it easier to read?

packages/core/workflows-sdk/src/utils/composer/helpers/create-step-handler.ts
```diff
@@ -0,0 +1,83 @@
...
+        idempotencyKey,
+        attempt: metadata.attempt,
+        container: stepArguments.container,
+        metadata,
+        eventGroupId:
+          stepArguments.transaction.getFlow()?.metadata?.eventGroupId ??
+          stepArguments.context!.eventGroupId,
+        transactionId: stepArguments.context!.transactionId,
+        context: stepArguments.context!,
+      }
```
adrien2p, line 52: suggestion: wdyt if you extract that to a separate function (e.g stepContextBuilder) which takes the action and stepArguments and build the step context, then we can use it here and in the compensate as well and be sure that only one place is building this context

packages/core/workflows-sdk/src/utils/composer/helpers/create-step-handler.ts
```diff
@@ -0,0 +1,83 @@
+import {CreateWorkflowComposerContext, StepExecutionContext} from "../type"
+import {WorkflowStepHandlerArguments} from "@medusajs/orchestration"
+import {resolveValue} from "./resolve-value"
+import {StepResponse} from "./step-response"
+import {deepCopy, OrchestrationUtils} from "@medusajs/utils"
+
+export function createStepHandler(
+  this: CreateWorkflowComposerContext,
+  { stepName, input, invokeFn, compensateFn }
```
adrien2p: suggestion: I havent added the types here but we could wdyt?

packages/core/workflows-sdk/src/utils/composer/create-workflow.ts
```diff
@@ -169,22 +180,9 @@ export function createWorkflow<
...
-        return
-      }
-
-      context.hooksCallback_[hook].push(fn)
-      shouldRegisterHookHandler = false
-    }
+  for (const hook of context.hooks_.declared) {
+    mainFlow["hooks"] ??= {}
+    mainFlow["hooks"][hook] = context.hooksCallback_[hook].bind(context)
   }
```
adrien2p: nit: we can probably move to `mainFlow.hooks` now that we have added the type, it was temporary 😅

packages/core/workflows-sdk/src/utils/composer/create-workflow.ts
```diff
@@ -19,6 +19,16 @@ import {
 
 global[OrchestrationUtils.SymbolMedusaWorkflowComposerContext] = null
 
+/**
+ * Workflow response class encapsulates the return value of a workflow
+ */
+export class WorkflowResponse<TData, THooks = []> {
```
adrien2p: suggestion: should we move that to its own file like the step response?

packages/core/core-flows/src/definition/cart/workflows/complete-cart.ts
thetutlage: Not sure. It seems like workflow return value can be a key-value pair with `WorkflowData` as the value. Referencing this code here. https://github.com/medusajs/medusa/pull/8346/files/1a54187c953e06b2b80cc7bc87ada3ea04ac0b05#diff-11ee48fe88c22283d687f2e85c805ade2ef634594e710516e603a2038fe42a29R98-R103

But if return value always have to be `WorkflowData`, then we can do as you suggested

packages/core/workflows-sdk/src/utils/composer/type.ts
thetutlage: Sure

packages/core/workflows-sdk/src/utils/composer/helpers/create-step-handler.ts
thetutlage: I will leave the existing logic cleanup to a separate PR :)

packages/core/workflows-sdk/src/utils/composer/helpers/create-step-handler.ts
thetutlage: Done

packages/core/workflows-sdk/src/utils/composer/create-workflow.ts
thetutlage: Done

packages/core/workflows-sdk/src/utils/composer/create-workflow.ts
thetutlage: It looks like `mainFlow` does not have hooks on it. It is set as type `MainExportedWorkflow`

packages/core/core-flows/src/definition/cart/workflows/complete-cart.ts
adrien2p: if you move that to a specific type, it includes all the cases
```ts
WorkflowData<TResult>
    | {
        [K in keyof TResult]:
          | WorkflowData<TResult[K]>
          | WorkflowDataProperties<TResult[K]>
          | TResult[K]
      }
```

that way when you pass the type it will allow you to return the workflow data with normal data or just normal data etc

packages/core/workflows-sdk/src/utils/composer/helpers/create-step-handler.ts
adrien2p: ahah got it 👍

packages/core/workflows-sdk/src/utils/composer/create-workflow.ts
adrien2p: humm, you are right 👍

packages/core/workflows-sdk/src/utils/composer/hook.ts
thetutlage: Done

packages/core/core-flows/src/definition/cart/workflows/complete-cart.ts
thetutlage: Done

packages/core/workflows-sdk/src/utils/composer/create-workflow.ts
thetutlage: Done

thetutlage:
@olivermrbl @srindom The PR is ready for the final review. Maybe you can go through the changelog once (in the PR description) to see if everything looks as expected.

packages/core/workflows-sdk/src/utils/_playground.ts
```diff
@@ -45,13 +54,13 @@ const workflow2 = createWorkflow("workflow", function () {
...
   return workflowRes
-})
+})*/
 
-workflow2()
-  .run({})
-  .then((res) => {
-    console.log(res.result)
-  })
+// workflow()
```
olivermrbl, line 59: question: Intentionally commented out?

packages/core/workflows-sdk/src/helper/workflow-export.ts
```diff
@@ -139,7 +139,7 @@ function createContextualWorkflowRunner<
     return {
       errors,
       transaction,
-      result,
+      result: result,
```
olivermrbl: nit(non-blocking): Not needed afaict

packages/core/workflows-sdk/src/utils/composer/__tests__/compose.ts
```diff
@@ -927,33 +923,35 @@ describe("Workflow composer", function () {
...
       const { result: final } = await workflow().run({
         input: workflowInput,
       })
 
-      expect(warn).toHaveBeenCalledTimes(1)
       expect(final).toEqual({
         id: "id_123",
-        prod: "product_1**",
-        var: "variant_2**",
+        // prod: "product_1**",
```
olivermrbl: question: Same question

packages/core/workflows-sdk/src/utils/composer/__tests__/compose.ts
```diff
@@ -1911,33 +1900,36 @@ describe("Workflow composer", function () {
...
```
olivermrbl: question: Same question

packages/core/workflows-sdk/src/utils/composer/create-hook.ts
```diff
@@ -0,0 +1,74 @@
...
+export type Hook<Name extends string, Input> = {
+  __type: typeof OrchestrationUtils.SymbolWorkflowHook
+  name: Name
+  /**
+   * By prefixing a key with a space, we remove it from the
+   * intellisense of TypeScript. This is needed because
+   * input is not set at runtime. It is a type-only
+   * property to infer input data type of a hook
+   */
+  " input": Input
```
olivermrbl, line 18: neat

olivermrbl (approved):
LGTM, nice work 🚢

packages/core/workflows-sdk/src/utils/_playground.ts
adrien2p: yes it is just a playground and I needed clarity :D, in that file you can do whatever you want, you just run it manually

packages/core/workflows-sdk/src/helper/workflow-export.ts
thetutlage: Yup.

packages/core/workflows-sdk/src/utils/composer/__tests__/compose.ts
thetutlage: Ohh yeah. Wanted to talk about it with @adrien2p once, because the old hooks were able return new data that gets merged with the output in the test.

packages/core/workflows-sdk/src/helper/workflow-export.ts
thetutlage: Done

adrien2p (approved):

PR #8347, opened 2024-07-30
chore(framework): Move and improve subscriber loader

adrien2p:
**What**
Move `SubscriberLoader` and improve implementation

FIXES FRMW-2635


packages/framework/framework/src/subscribers/types.ts
```diff
@@ -0,0 +1,16 @@
+import { Event, MedusaContainer } from "@medusajs/types"
```
olivermrbl, line 1: question: Do we keep types internal to the package?

olivermrbl (approved):
Nice LGTM!

packages/framework/framework/src/subscribers/types.ts
adrien2p: The types will be part of a completely separate work where we will identify what needs to be part of the framework and what will not. Unless obvious typings 😅 otherwise it can create some issues and breaking changes that could be too early

adrien2p: same from what I am currently doing, we are moving to get them out of the way, but we might decide to maybe have a more generic way of managing resources and then provide a front facing opinionated consume way for medusa in order to split what is fully opinionated and what needs to be generic, but that will be later discussions 👍 So you can imagine that a front facing with more constraint is wrapping the underlying API to restrict things while the framework is more open, It is just first thought but nothing has been discussed or whatever, just throwing things out loud for now but as you can imagine there is some later discussion that will happen

packages/framework/framework/package.json
```diff
@@ -32,6 +32,12 @@
       "import": "./dist/database/index.js",
       "require": "./dist/database/index.js",
       "node": "./dist/database/index.js"
+    },
+    "./subscribers": {
+      "types": "./dist/subscribers/index.d.ts",
+      "import": "./dist/subscribers/index.js",
+      "require": "./dist/subscribers/index.js",
+      "node": "./dist/subscribers/index.js"
```
thetutlage: NIT. Do we need explicit mappings for all the types? Because the following will do the same thing

```json
{
  "./subscribers": "./dist/subscribers/index.js"
}
```

adrien2p: not necessarely, I wasn't sure what if we would like to split later, but I can reduce it for now and we can come back to it at a later point

integration-tests/modules/package.json
```diff
@@ -5,7 +5,7 @@
   "license": "MIT",
   "private": true,
   "scripts": {
-    "test:integration": "NODE_OPTIONS=--experimental-vm-modules jest --silent=false --no-cache --maxWorkers=50% --bail --detectOpenHandles --forceExit --logHeapUsage",
+    "test:integration": "NODE_OPTIONS=--experimental-vm-modules jest --no-cache --maxWorkers=50% --bail --detectOpenHandles --forceExit --logHeapUsage -- __tests__/event-bus/index.spec.ts",
```
adrien2p: todo:revert

olivermrbl (approved):
LGTM, should we merge?

PR #8355, opened 2024-07-30
chore(core-flows,modules-sdk): add TSDocs to helper steps

shahednasser:
Add TSDocs to helper steps and related types (to later generate a reference)


olivermrbl (approved):
Nice!

Issue #8348, opened 2024-07-30
Sales channels error when disabled in feature flags


BeyondOT:
### Describe the bug

Error when trying to create a price_list when sales_channels are disabled.

::1 - - [30/Jul/2024:09:36:22 +0000] "GET /admin/customer-groups?limit=4&offset=0&expand=customers&q= HTTP/1.1" 304 - "http://localhost:7001/" "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/127.0.0.0 Safari/537.36 Edg/127.0.0.0"
error:   Relation with property path sales_channels in entity was not found.
TypeORMError: Relation with property path sales_channels in entity was not found.

::1 - - [30/Jul/2024:09:36:31 +0000] "GET /admin/products?limit=20&offset=0&expand=variants%2Csales_channels%2Ccollection&q= HTTP/1.1" 500 86 "http://localhost:7001/" "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/127.0.0.0 Safari/537.36 Edg/127.0.0.0"
error:   Relation with property path sales_channels in entity was not found.
TypeORMError: Relation with property path sales_channels in entity was not found.

### System information

Medusa version (including plugins): 1.20.9
Node.js version: 20.16.0
Database: Postgresql
Operating system: Windows 11
Browser (if relevant): Microsoft Edge

### Steps to reproduce the behavior

1. Disable sales_channels in feature flags.
2. Go to Price list in admin and click create.
3. Enter the title and description and click next, the first step should work well.
4. An error should occur in this step because the products can't be retrieved

### Expected behavior

To be able to create the price list by having the products retrieved properly. This is the behavior that happens when sales_channels are not disabled.

### Screenshots

If applicable, add screenshots to help explain your problem

### Code snippets

If applicable, add code samples to help explain your problem

### Additional context

Add any other context about the problem here

adevinwild:
I think the problem is similar to my last PR (https://github.com/medusajs/medusa/issues/8102), I fixed it for the products but not for the rest sorry!
I'll take care of it ASAP 🙌

sradevski:
I'll close the issue as the fix was merged, should be available in the next release.

PR #8357, opened 2024-07-30
fix: Check if the SalesChannels FF is enabled on `/admin/products`

adevinwild:
# Related issue
This PR is related to this issue https://github.com/medusajs/medusa/issues/8348
and follows on from the previous PR which was merge (https://github.com/medusajs/medusa/pull/8120)

## What
This PR make sure to check if the SalesChannels FF is enabled before using the sales channels relations with the products

## Why
By solving this problem, the API route will now be functional and will no longer throw errors in that specific condition (Having SalesChannels FF disabled), which will also result in an error-free UI.

### DIFF
--- a/packages/medusa/src/api/routes/admin/products/list-products.ts
+++ b/packages/medusa/src/api/routes/admin/products/list-products.ts
@@ -9,7 +9,7 @@ import {
 import { listProducts } from "../../../../utils"
 
 import { IInventoryService } from "@medusajs/types"
-import { MedusaV2Flag } from "@medusajs/utils"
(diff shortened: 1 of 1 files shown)


adevinwild (commented):
Here, we're just using the `FeatureFlagRouter` to check if the `SalesChannelFeatureFlag` is enabled before doing anything related to the concept of `SalesChannels`

carlos-r-l-rodrigues (approved):
LGTM

riqwan (approved):
🔥

Issue #8158, opened 2024-07-17
Incorrect Popover z-index in Edit Variant modal


Dominic-Preap:
### Describe the bug

Popover of Edit Variant is behind number input.

### System information

Medusa version (including plugins):  `@medusajs/admin": "^7.1.14"`
Node.js version: 20.9.0
Database: Postgresql 
Operating system: Windows 10
Browser (if relevant): Chrome

### Steps to reproduce the behavior

1. Go to Product detail
2. Click Edit Variant in Variant section
3. Click '...' for Edit Variant to see

### Expected behavior

Popover should be in front of input.

### Screenshots

![image](https://github.com/user-attachments/assets/736b26b2-e29b-4ac1-83d4-60d51dac77f7)

### Code snippets

N/A

### Additional context

N/A

srindom:
Thanks for opening this issue Dominic - would you be up for opening a PR with a fix?

srindom:
Thanks for opening this issue @Dominic-Preap - would you be up for working on a PR to fix this?

Dominic-Preap:
@srindom sure, I'll check contribution guide and make a PR.

srindom:
@Dominic-Preap - how did it go with getting started on this? Happy to help if you are bumping into challenges?

Dominic-Preap:
I'm so sorry,  I've been busy lately 😔

adevinwild:
Hey! 
I had a little bit of time to fix it, It should be ok now 🙌

kasperkristensen:
Fixed by @adevinwild in #8358

PR #8358, opened 2024-07-30
fix(admin-ui): Fixed DropdownMenu inside a Modal zIndex issues

adevinwild:
# Related issue
This PR is related to this issue https://github.com/medusajs/medusa/issues/8158

## What
This PR make sure to render the `DropdownMenu` in a React `Portal` to solves any issues when you have a `DropdownMenu` inside a `Modal`

- This is a common issue with RadixUI the solution is to use a Portal to make sure the DropdownMenu is rendered on the `document.body` instead of the same container
- Updated the zIndex from `z-30` to `z-50`

image


kasperkristensen (approved):
LGTM, thanks @adevinwild

PR #8368, opened 2024-07-31
feat(dashboard,js-sdk,types): add ability to capture payment from order page

riqwan:
what:

- adds ability to capture payment from order page

bugs found:

- when capturing payment of one order, other orders statuses get affected from "captured" to "partially captured". Will investigate this separately. 

https://github.com/user-attachments/assets/0a1beac2-74fc-4803-8528-8de5913964d4


packages/admin-next/dashboard/src/routes/orders/order-detail/components/order-payment-section/order-payment-section.tsx
```diff
@@ -4,45 +4,62 @@ import {
...
     </Container>
   )
 }
 
 const Header = ({ order }: { order: Order }) => {
   const { t } = useTranslation()
 
-  const hasCapturedPayment = order.payments.some((p) => !!p.captured_at)
+  const hasCapturedPayment = order.payment_collections.some(
+    (p) => !!p.captured_at
```
fPolic: seems that `PaymentCollection` doesn't have this field ATM since we haven't define what it means for a PC to be captured because it can have multiple payments and refunds

packages/core/js-sdk/src/admin/payment.ts
```diff
@@ -0,0 +1,20 @@
...
+import { Client } from "../client"
+import { ClientHeaders } from "../types"
+
+export class Payment {
+  private client: Client
+  constructor(client: Client) {
+    this.client = client
+  }
+
+  async capture(id: string, headers?: ClientHeaders) {
```
fPolic: todo: we should add body: `{amount?: number}` to match the API

packages/admin-next/dashboard/src/routes/orders/order-detail/components/order-payment-section/order-payment-section.tsx
riqwan: good catch, removed this based on our discussions.

packages/core/js-sdk/src/admin/payment.ts
riqwan: nice, added them!

fPolic (approved):
LGTM!

PR #8370, opened 2024-07-31
feat(dashboard, types, js-sdk): Claims client, hooks and types

fPolic:


olivermrbl (approved):
LGTM!

A quick question: What is the difference between:
- `POST /admin/claims/:id/claim-items`
- `POST /admin/claims/:id/inbound/items`
- `POST /admin/claims/:id/outbound/items`

And know what the difference is between inbound and outbound. But not 100% sure, I am fully aware of what the first endpoint is used for.

fPolic:
> A quick question: What is the difference between:
> 
> * `POST /admin/claims/:id/claim-items`
> * `POST /admin/claims/:id/inbound/items` and `POST /admin/claims/:id/outbound/items`

I was about to ask @carlos-r-l-rodrigues the same things 😄

carlos-r-l-rodrigues:
`claim-items` flags the claimed items,
`outbound` are the additional items added to the order and
`inbound` are the returned items

olivermrbl:
I would have expected claim + inbound to be performed in the same operation. The items you return as part of a claim are the items that are flagged. What's the idea behind separating the two?

carlos-r-l-rodrigues:
We can claim a missing item, can't we?
The idea was just that something can be claimed, but not necessarily returned.
But if this is not needed I can change it.

PR #8385, opened 2024-08-01
feat(dashboard,types,js-sdk,payment): ability to refund payment in order page

riqwan:
what:

- adds ability to refund payment in order page

**observations:**

1. when creating a single payment, we validate that the refund is not higher than the payment amount. This however does not account for sending multiple refunds. You can send multiple refunds that gross greater than the payment amount. 
2. In the designs, we have refund reasons, note and send notification. None of these have been implemented in the backend yet. 
3. Currency input breaks on changing input. Will take a look at this separately. In this PR, its a simple number input. 

https://github.com/user-attachments/assets/0ba9df14-0beb-4830-9591-9ae135794729


olivermrbl:
Looks like the refund entry deviates from the design:

**Video**
CleanShot 2024-08-01 at 13 12 08

**Design**
CleanShot 2024-08-01 at 13 12 47

riqwan:
@olivermrbl I've moved the date and fixed the icon. The issue with the badge and the note icon will still persists as I've written in the observations. Those are not done in the backend, i'm unsure if they're on purpose or not. 


Screenshot 2024-08-01 at 17 06 34

packages/admin-next/dashboard/src/lib/client/payments.ts
```diff
@@ -1,5 +1,13 @@
-import { getRequest } from "./common"
+import { HttpTypes } from "@medusajs/types"
 import { PaymentProvidersListRes } from "../../types/api-responses"
+import { getRequest } from "./common"
+
+async function retrievePayment(id: string, query?: Record<string, any>) {
```
olivermrbl: suggestion: Should we migrate this to the `js-sdk` while we are at it?

olivermrbl (approved):
LGTM, are we tracking the observations in the PR description?

packages/admin-next/dashboard/src/lib/client/payments.ts
riqwan: yes def! force of habit 😄

PR #8386, opened 2024-08-01
feat: Add support for sorting export headers

sradevski:
I just put the most important fields in front, and the rest are sorted alphabetically. We can adjust this as needed.

CLOSES CC-272


olivermrbl (approved):
LGTM

PR #8387, opened 2024-08-01
docs: added a note about admin customizations as arrow functions

shahednasser:
Added a note mentioning the limitation for creating admin widgets and routes, that they should be arrow functions.


shahednasser:

olivermrbl (approved):

PR #8390, opened 2024-08-01
feat: Show notice on new notifications

sradevski:
CLOSES CC-255

The PR adds identifiers for when there are new notifications. Currently it will change the bell icon, and it will show a small dot next to each unread notification.

Screenshot 2024-08-01 at 11 24 34
Screenshot 2024-08-01 at 11 24 43
Screenshot 2024-08-01 at 11 28 34

### DIFF
--- a/packages/admin-next/dashboard/src/components/layout/notifications/notifications.tsx
+++ b/packages/admin-next/dashboard/src/components/layout/notifications/notifications.tsx
@@ -1,6 +1,7 @@
 import {
   BellAlert,
   BellAlertDone,
+  CircleFilledSolid,
   InformationCircleSolid,
 } from "@medusajs/icons"
 import { Drawer, Heading, IconButton, Text } from "@medusajs/ui"
@@ -10,7 +11,7 @@ import { HttpTypes } from "@medusajs/types"
 import { formatDistance } from "date-fns"
 import { InfiniteList } from "../../common/infinite-list"
 import { sdk } from "../../../lib/client"
-import { notificationQueryKeys } from "../../../hooks/api"
+import { notificationQueryKeys, useNotifications } from "../../../hooks/api"
 import { TFunction } from "i18next"
 import { FilePreview } from "../../common/file-preview"
 
@@ -24,9 +25,17 @@ interface NotificationData {
   }
 }
 
+const LAST_READ_NOTIFICATION_KEY = "notificationsLastReadAt"
+
 export const Notifications = () => {
-  const [open, setOpen] = useState(false)
   const { t } = useTranslation()
+  const [open, setOpen] = useState(false)
+  const [hasUnread, setHasUnread] = useUnreadNotifications()
+  // This is used to show the unread icon on the notification when the drawer is open,
+  // so it should lag behind the local storage data and should only be reset on close
+  const [lastReadAt, setLastReadAt] = useState(
+    localStorage.getItem(LAST_READ_NOTIFICATION_KEY)
+  )
 
   useEffect(() => {
     const onKeyDown = (e: KeyboardEvent) => {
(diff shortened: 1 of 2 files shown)


packages/admin-next/dashboard/src/components/layout/notifications/notifications.tsx
```diff
@@ -42,14 +51,25 @@ export const Notifications = () => {
     }
   }, [])
 
+  const handleOnOpen = (isOpen: boolean) => {
```
olivermrbl, line 54:
```suggestion
  const handleOnOpen = (shouldOpen: boolean) => {
```

packages/admin-next/dashboard/src/components/layout/notifications/notifications.tsx
```diff
@@ -161,3 +195,31 @@ const NotificationsEmptyState = ({ t }: { t: TFunction }) => {
     </div>
   )
 }
+
+const useUnreadNotifications = () => {
+  const [hasUnread, setHasUnread] = useState(false)
+  const { notifications } = useNotifications(
+    { limit: 1, offset: 0, fields: "created_at" },
+    { refetchInterval: 3000 }
```
olivermrbl, line 203: thought: Can we add some sort of exponential backoff that resets on page refreshes?

olivermrbl (approved):
LGTM!

packages/admin-next/dashboard/src/components/layout/notifications/notifications.tsx
sradevski: Hey Oli, what are we trying to achieve with that? Are you afraid that we are polling too frequently?

Adding exponential backoff would mean that the longer you are on the page, the longer it will take to have cause-and-effect (requested an export -> a notification is shown for it), which will probably make the UX worse, but maybe you have some other idea in mind

olivermrbl: > what are we trying to achieve with that? Are you afraid that we are polling too frequently?

Not afraid of excessive polling. Just think it looks a bit nasty to poll every third second when there are no notifications.

> Adding exponential backoff would mean that the longer you are on the page, the longer it will take to have cause-and-effect (requested an export -> a notification is shown for it)

The idea would be to reset it on page refresh. I imagine in the vast majority of cases the notification would come within 10 seconds (maybe even less), so the exponential backoff wouldn't be an issue. 

In any case, this is nitpicking and non-blocking – just thought I'd bring it up to hear your thoughts.

sradevski: I think we definitely could, but considering there will rarely be more than 1-2 people using the admin at the same time, that's less than 1 request per second, compared to eg. just loading the products which is an order of magnitude heavier operation, I'd rather just keep it simple.

Ideally we don't even poll when the time comes to handle notifications properly 😄

adrien2p: Ahah yes and we already support sse as we shown during the poc for workflow notification. But yeah long term polling should be removed imo

olivermrbl: Yes, SSE (or something similar) would definitely be the end goal, but we decided to keep it simple in the first iteration

carlos-r-l-rodrigues: Just to let my comment here as well to emphasize the importance of having this sse/fetch-stream in place to subscribe to workflows. 
This will be somewhat a big differential for all kind of flows to create a near real time notification system for the admin. And extensible for customizations as it can handle all sorts of flows.

PR #8391, opened 2024-08-01
feat: Add `requested_at` + `open` status to return

olivermrbl:
Closes CC-289

### What 
- Add `requested_at` column on `return` model
- Augment return status enum with an `open` status

Upon initiating a return, the status will be `open`. Upon confirming the return request, we set the status to `requested` and `requested_at`.

### Why
We want to distinguish initiated returns from requested ones in the UI.


sradevski (approved):
LGTM, just one very minor thing:
At first sight it wasn't clear to me what the difference between `requested` and `open` is, so maybe it makes sense to have `requested`, `confirmed`, ... based on the PR description. Let me know what you think.

olivermrbl:
The multi-request return flow we have is as follows:
- You open a return request. This means you can add items, shipping, etc. to it.
- You confirm the return request. This closes the return for modification. 
- You receive the return request.

And the status (as currently implemented) follows this flow as such: `open` -> `requested` -> `received`.

I would be OK with changing the statuses to `created` -> `confirmed` -> `received` or something like that. But I don't want this to be a blocker. We need this pretty urgently for the UI, so perhaps we can address this in a follow-up PR as we polish things?

carlos-r-l-rodrigues (approved):
LGTM.

PR #8394, opened 2024-08-01
feat: Use tag ids instead of values wherever possible

sradevski:
CLOSES CC-284


packages/admin-next/dashboard/src/hooks/table/query/use-product-table-query.tsx
```diff
@@ -54,7 +54,7 @@ export const useProductTableQuery = ({
     collection_id: collection_id?.split(","),
     is_giftcard: is_giftcard ? is_giftcard === "true" : undefined,
     order: order,
-    tags: tags ? { value: tags.split(",") } : undefined,
```
sradevski, line 57: This was wrong, the tags here were IDs, not values

packages/core/core-flows/src/product/helpers/normalize-for-import.ts
```diff
@@ -108,10 +116,15 @@ const normalizeProductForImport = (
     }
 
     if (normalizedKey.startsWith("product_tag_")) {
-      response["tags"] = [
-        ...(response["tags"] || []),
-        { value: normalizedValue },
-      ]
+      const tag = tagsMap.get(normalizedValue)
+      if (!tag) {
```
sradevski, line 120: I would be ok for the import to first create all tags and then create the products if we want that behavior (as a separate PR though).

packages/admin-next/dashboard/src/hooks/table/query/use-product-table-query.tsx
olivermrbl: Nice catch

olivermrbl (approved):
LGTM!

PR #8396, opened 2024-08-01
feat(core-flows,order,medusa): exchange endpoints

carlos-r-l-rodrigues:
Order Exchange: PR 2:n

What:
- Order Exchange endpoints
- Claims and Exchange worfklows fixes 
- Small fixes on Order module action validations

Next PR will contain inventory management of claims and exchanges.

### DIFF
--- a/packages/modules/order/src/migrations/Migration20240801085921.ts
+++ b/packages/modules/order/src/migrations/Migration20240801085921.ts
@@ -3,10 +3,24 @@ import { Migration } from "@mikro-orm/migrations"
 export class Migration20240801085921 extends Migration {
   async up(): Promise<void> {
     this.addSql(`
-      ALTER TABLE "return"
-        ADD COLUMN IF NOT exists "requested_at" timestamptz NULL;
+        ALTER TABLE "return"
+          ADD COLUMN IF NOT exists "requested_at" timestamptz NULL;
 
-      ALTER TYPE return_status_enum ADD VALUE 'open';
+        
+        ALTER TABLE "return" ALTER COLUMN "status" DROP DEFAULT;
+        ALTER TABLE "return" ALTER COLUMN "status" TYPE text USING "status"::text;
+        DROP TYPE return_status_enum;
+
+        CREATE TYPE return_status_enum AS ENUM (
+          'open',
+          'requested',
+          'received',
+          'partially_received',
+          'canceled'
+        );
+        
+        ALTER TABLE "return" ALTER COLUMN "status" TYPE return_status_enum USING "status"::text::return_status_enum;
+        ALTER TABLE "return" ALTER COLUMN "status" SET DEFAULT 'open';        
       `)
   }
 }


packages/core/core-flows/src/order/workflows/exchange/remove-exchange-item-action.ts
```diff
@@ -0,0 +1,97 @@
...
+        `No item exchange found for exchange ${input.exchange_id} in order change ${orderChange.id}`
+      )
+    } else if (associatedAction.action !== ChangeActionType.ITEM_ADD) {
+      throw new Error(`Action ${associatedAction.id} is not adding an item`)
+    }
+  }
+)
+
+export const removeItemExchangeActionWorkflowId = "remove-item-exchange-action"
+export const removeItemExchangeActionWorkflow = createWorkflow(
```
olivermrbl, line 57: would love to get these steps to be reusable soon 😄

packages/medusa/src/api/admin/exchanges/[id]/inbound/items/[action_id]/route.ts
```diff
@@ -0,0 +1,96 @@
+import {
```
Comment (olivermrbl) [inline packages/medusa/src/api/admin/exchanges/[id]/inbound/items/[action_id]/route.ts:1]: question: Why do we fetch and return `returns` in these endpoints when they are for exchanges?

packages/medusa/src/api/admin/exchanges/[id]/inbound/items/route.ts
```diff
@@ -0,0 +1,60 @@
+import { orderExchangeRequestItemReturnWorkflow } from "@medusajs/core-flows"
```
Comment (olivermrbl) [inline packages/medusa/src/api/admin/exchanges/[id]/inbound/items/route.ts:1]: same question here

packages/medusa/src/api/admin/exchanges/[id]/inbound/shipping-method/[action_id]/route.ts
```diff
@@ -0,0 +1,110 @@
+import {
```
Comment (olivermrbl) [inline packages/medusa/src/api/admin/exchanges/[id]/inbound/shipping-method/[action_id]/route.ts:1]: Same question here, will not comment more on this anymore if there are other similar endpoints.

olivermrbl (approved):
Lots of stuff in this PR. Can we try to keep them smaller going forward because it's a bit hard to review? 

I left a few comment, but I would be OK getting this merged so it doesn't block any UI implementation work.

packages/medusa/src/api/admin/exchanges/[id]/inbound/items/[action_id]/route.ts
Comment (carlos-r-l-rodrigues) [reply packages/medusa/src/api/admin/exchanges/[id]/inbound/items/[action_id]/route.ts:1]: In practice, inbounds are returns

packages/core/core-flows/src/order/workflows/exchange/remove-exchange-item-action.ts
carlos-r-l-rodrigues: I will work to clean these similar steps everywhere in Order flows

PR #8405, opened 2024-08-02
feature: introduce additional_data to the product endpoints

thetutlage:
Fixes: FRMW-2647, FRMW-2648

First approach of introducing and consuming additional data.

- Update the `AdminCreateProduct` to now include `additional_data` as part of the payload. 
- Next we have to update the validator for this endpoint to do two things. First, optionally allow `additional_data` as an arbitrary object. But then also, use merge the user-defined validator (if exists). This change might impact some other parts of the codebase.
- Finally, we destructure `additional_data` and rest of the properties and pass them to the workflow as needed.

### DIFF
--- a/packages/modules/payment/integration-tests/__tests__/services/payment-module/index.spec.ts
+++ b/packages/modules/payment/integration-tests/__tests__/services/payment-module/index.spec.ts
@@ -1,12 +1,12 @@
 import { IPaymentModuleService } from "@medusajs/types"
 import { Module, Modules, promiseAll } from "@medusajs/utils"
+import { PaymentModuleService } from "@services"
 import { moduleIntegrationTestRunner } from "medusa-test-utils"
 import {
   createPaymentCollections,
   createPayments,
   createPaymentSessions,
 } from "../../../__fixtures__"
-import { PaymentModuleService } from "@services"
 
 jest.setTimeout(30000)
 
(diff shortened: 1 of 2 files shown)


packages/medusa/package.json
```diff
@@ -39,7 +39,7 @@
     "prepublishOnly": "cross-env NODE_ENV=production tsc --build",
     "build": "rimraf dist && tsc --build",
     "serve": "node dist/app.js",
-    "test": "jest --silent --bail --maxWorkers=50% --forceExit"
+    "test": "jest --silent=false --bail --maxWorkers=50% --forceExit"
```
olivermrbl:
```suggestion
    "test": "jest --silent --bail --maxWorkers=50% --forceExit"
```

packages/medusa/src/api/admin/products/route.ts
```diff
@@ -30,11 +30,17 @@ export const GET = async (
 }
 
 export const POST = async (
-  req: AuthenticatedMedusaRequest<HttpTypes.AdminCreateProduct>,
+  req: AuthenticatedMedusaRequest<
+    HttpTypes.AdminCreateProduct & {
+      additional_data?: Record<string, unknown>
+    }
```
olivermrbl: nit(non-blocking): We might want to create a dedicated type instead of inline at some point - wdyt?

olivermrbl (approved):
🔥

packages/medusa/src/api/admin/products/route.ts
carlos-r-l-rodrigues: +1

carlos-r-l-rodrigues (approved):

packages/medusa/src/api/admin/products/route.ts
thetutlage: Sure. Let me do that now, instead of leaving it to later

PR #8406, opened 2024-08-02
fix: Check for existence of modules when seeding default data

sradevski:


adrien2p (approved):
LGTM, eventually I would add allowUnregistered on the resolve just for extra precaution :)

riqwan (approved):
lgtm, whats the use-case for this one?

sradevski:
@riqwan the default workflow runs on startup, but in platform I am not using the sales channels and stores modules, so it crashes.

PR #8407, opened 2024-08-02
chore(framework): Allow multiple source dir for subcribers loader

adrien2p:
RESOLVE FRMW-2640

**What**
Allow the loader to receive multiple source paths like other loaders now

### DIFF
--- a/packages/core/core-flows/src/defaults/steps/create-default-store.ts
+++ b/packages/core/core-flows/src/defaults/steps/create-default-store.ts
@@ -13,6 +13,10 @@ export const createDefaultStoreStep = createStep(
   async (data: CreateDefaultStoreStepInput, { container }) => {
     const storeService = container.resolve(ModuleRegistrationName.STORE)
 
+    if (!storeService) {
+      return new StepResponse(void 0)
+    }
+
     let shouldDelete = false
     let [store] = await storeService.listStores({}, { take: 1 })
 
--- a/packages/core/core-flows/src/sales-channel/steps/create-default-sales-channel.ts
+++ b/packages/core/core-flows/src/sales-channel/steps/create-default-sales-channel.ts
@@ -17,6 +17,10 @@ export const createDefaultSalesChannelStep = createStep(
       ModuleRegistrationName.SALES_CHANNEL
     )
 
+    if (!salesChannelService) {
+      return new StepResponse(void 0)
+    }
+
     let shouldDelete = false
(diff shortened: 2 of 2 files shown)


packages/framework/framework/src/subscribers/subscriber-loader.ts
```diff
@@ -213,20 +216,20 @@ export class SubscriberLoader {
...
-    } catch (err) {
-      logger.debug(`No subscriber directory found in ${this.#sourceDir}`)
-    }
+    const normalizeSourcePaths = Array.isArray(this.#sourceDir)
+      ? this.#sourceDir
+      : [this.#sourceDir]
+    const promises = normalizeSourcePaths.map(async (sourcePath) => {
+      try {
+        await access(sourcePath)
+      } catch {
```
riqwan, line 225: Sugg: let's throw here instead? Much better experience knowing something explicitly screwed up than it getting swallowed

riqwan (approved):
Lgtm!

packages/framework/framework/src/subscribers/subscriber-loader.ts
adrien2p: It is more if you have a plugin that does not provide any subscribers, then the dir wont exists but it is expected since you dont provide any. It would throw mater though if we cant properly load a subscriber

adrien2p: So i believe it would create more confusion to throw at this level, probably a user would wonder why it throw that it cant load the subscriber while I don't provide any subscribers and don't want to provide any

adrien2p: eventually I can add info log to say that it will be skipped for that source? wdyt?

riqwan: I thought this config was exposed to the user, looks like its not. You're right, this doesn't require a throw. 🤞🏻

sradevski: I think adding the log is sufficient, WDYT Riqwan?

sradevski (approved):
👍

packages/framework/framework/src/subscribers/subscriber-loader.ts
adrien2p: I ve added them, let me know what you think guys

sradevski (approved):
Much better having the logs 👍

PR #8411, opened 2024-08-02
fix(framework/config): Properly resolve env variables

adrien2p:
**What**
fix env variable resolution


thetutlage:
Is it that the file was imported before the `NODE_ENV` was set?

adrien2p:
> Is it that the file was imported before the `NODE_ENV` was set?

Yes very likely die to the import at the top, but i wasn't expecting that because it gets instanciated mater but probably the evaluation lead to this

PR #8418, opened 2024-08-02
feat(dashboard): BulkEditor Boolean cell behaviour

kasperkristensen:
**What**
- Part two of bulk editor improvements, mainly focussed on adding special behaviours for the Boolean cell.

**Changes to the Boolean cell**
- Boolean cells no longer require a double click or "Enter" key press before being interactive.
- Boolean cells now invert when focused and the user hits "Space".
- If a range of boolean cells are selected hitting "Space" will do the following:
  - If all fields are either true or false then invert the values.
  - If some fields are true and others are false, then set all fields to true.
- Hitting "Enter" on a Boolean cell will invert its value, and then move focus to the next row. This is not the same behaviour, as you would find in a spreadsheet like Google Sheets. Their hitting enter shows the underlying text value, e.g. "FALSE". We don't have a underlying text value, that we allow users to edit, so instead I opted for the described behaviour, which allows the user to toggle cells in quick succession.

**Changes to scroll behaviour**
- We now attempt to scroll to the next row/column if possible. This overscroll provides a more smooth experience, as it allows the virtualizer to load the next couple of rows/columns even if the cell we are currently scrolling to is at the edge of the screen.

**Changes to how selection and dragSelection is calculated**
- The `useEffect`'s that were used to update the selection / drag selection on updates to `anchor`, `rangeEnd`, and `dragEnd` have been removed in favour of utils that a cell can call to check if it is in either selection.

**Clean up of query selector use**
- A new hook `useGridQueryTool` has been added that exposes a couple of utils for getting a cell container or cell input HTML element.

**Other**
- Adds `prop-types` dependency to project. This should prevent an error being thrown about the dependency missing when running the dashboard in development. This is might temporary, as the current FocusTrap solution is not perfect and needs to be improved. Not sure yet, if the desired effect can be achieved with `focus-trap-react`, but if not we will remove the dependency and `prop-types`.

Closes CC-268


kasperkristensen:
@shahednasser I have added the `prop-types` dependency to `dashboard`, so when this gets merged we should check if we can remove it from the starter 👍

olivermrbl:
I accidentally dragged horizontally and got this error:
CleanShot 2024-08-06 at 09 57 56

kasperkristensen:
> I accidentally dragged horizontally and got this error: CleanShot 2024-08-06 at 09 57 56

Should be fixed now 👍

olivermrbl (approved):

PR #8448, opened 2024-08-05
feat(core-flows): manage inventory reservation of claims and exchanges

carlos-r-l-rodrigues:


packages/core/core-flows/src/definition/cart/workflows/confirm-variant-inventory.ts
```diff
@@ -110,7 +111,11 @@ export const confirmVariantInventoryWorkflow = createWorkflow(
       return { items }
     })
 
-    confirmInventoryStep(confirmInventoryInput)
+    when({ input }, ({ input }) => {
+      return !input.skipInventoryCheck
+    }).then(() => {
+      confirmInventoryStep(confirmInventoryInput)
+    })
```
olivermrbl, line 118: question: Does it make sense for this check to live inside the `confirmVariantInventoryWorkflow`? Shouldn't it be outside of it when you consume it?

olivermrbl (approved):
LGTM

packages/core/core-flows/src/definition/cart/workflows/confirm-variant-inventory.ts
carlos-r-l-rodrigues: Well, this is a good point.
I think it makes sense to separate in a util the portion to format the input to confirm/reserve inventory, from the inventory check.
I will open a follow up PR organizing that.

PR #8452, opened 2024-08-05
feat(create-medusa-app, medusa): check Node.js version before installation.

shahednasser:
Check before installation with `create-medusa-app` and `medusa new` whether the current node version is >= 20. If not, exit with an error.

### DIFF
--- a/packages/cli/create-medusa-app/src/commands/create.ts
+++ b/packages/cli/create-medusa-app/src/commands/create.ts
@@ -26,7 +26,7 @@ import {
   installNextjsStarter,
   startNextjsStarter,
 } from "../utils/nextjs-utils.js"
-import { getNodeVersion } from "@medusajs/utils"
+import { getNodeVersion, MIN_SUPPORTED_NODE_VERSION } from "@medusajs/utils"
 
 const slugify = slugifyType.default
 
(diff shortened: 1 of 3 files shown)


packages/cli/create-medusa-app/src/commands/create.ts
```diff
@@ -53,6 +53,13 @@ export default async ({
   withNextjsStarter = false,
   verbose = false,
 }: CreateOptions) => {
+  const nodeVersion = getNodeVersion()
+  if (nodeVersion < 20) {
+    logMessage({
+      message: `Medusa requires at least v20 of Node.js. You're using v${nodeVersion}. Please install at least v20 and try again: https://nodejs.org/en/download`,
+      type: "error"
+    })
```
riqwan, line 61: q: should we exit here as well?

riqwan (approved):
🥇

packages/core/utils/src/common/get-node-version.ts
```diff
@@ -0,0 +1,5 @@
+export function getNodeVersion(): number {
+  const [major] = process.versions.node.split('.').map(Number)
+
+  return major
+}
```
riqwan: super nit: 
```suggestion
}

export const MIN_SUPPORTED_NODE_VERSION = 20

```

olivermrbl (approved):
Other than @riqwan's comment, LGTM :)

packages/cli/create-medusa-app/src/commands/create.ts
shahednasser: This one exits automatically when type is `error`.

PR #8454, opened 2024-08-05
fix(medusa): plus sign to space

carlos-r-l-rodrigues:
Http clients will transform `+` in the URL to `%20` (space)

### DIFF
--- a/packages/modules/order/src/migrations/Migration20240801085921.ts
+++ b/packages/modules/order/src/migrations/Migration20240801085921.ts
@@ -3,10 +3,24 @@ import { Migration } from "@mikro-orm/migrations"
 export class Migration20240801085921 extends Migration {
   async up(): Promise<void> {
     this.addSql(`
-      ALTER TABLE "return"
-        ADD COLUMN IF NOT exists "requested_at" timestamptz NULL;
+        ALTER TABLE "return"
+          ADD COLUMN IF NOT exists "requested_at" timestamptz NULL;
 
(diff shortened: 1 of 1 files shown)


packages/medusa/src/utils/get-query-config.ts
```diff
@@ -60,15 +60,17 @@ export function prepareListQuery<T extends RequestQueryFields, TEntity>(
...
       })
+
     if (shouldReplaceDefaultFields) {
-      allFields = new Set(customFields.map((f) => f.replace(/^[+-]/, "")))
+      allFields = new Set(customFields.map((f) => f.replace(/^[+ -]/, "")))
     } else {
       customFields.forEach((field) => {
-        if (field.startsWith("+")) {
-          allFields.add(field.replace(/^\+/, ""))
+        if (field.startsWith("+") || field.startsWith(" ")) {
```
riqwan, line 72: Q: why are we checking for empty space here?

riqwan (approved):
nice!

packages/medusa/src/utils/get-query-config.ts
carlos-r-l-rodrigues: because most clients will transform `+` into and empty space, and that is what Medusa app will receive.

olivermrbl (approved):
LGTM, perhaps we need to rethink the + sign some day, but I think we can leave that for when/if we see people getting issues with it

Issue #8455, opened 2024-08-05
PayPal payment session not being updated after cart is updated


Arsenalist:
Node: 18
Medusa 1.20.9,
PayPal plugin: 6.0.4
OS: MacOS

Steps to reproduce:

- Add 1 item to cart for $10
- Create PayPal payment session, it will be for $10
- Click PayPal button to see popup prompt to pay for $10
- Close popup
- Change quantity of item in cart to 2, so total is $20
- Click PayPal button again, it will still say $10 when it should say $20

The call to PayPal API is happening OK in `PayPalProvider.updatePayment()` but the PayPal order ID does not reflect the new value.

In PayPal console, the PATCH request which is supposed to update the amount to $20 has no data associated with it:

<img width="779" alt="image" src="https://github.com/user-attachments/assets/98147c61-b265-4f90-96f6-cae3bb68b88c">

This makes sense as it's calling patch order in the paypal-sdk file and that doesn't send any data (even though data is passed into it):

```
  async patchOrder(orderId: string, data?: PatchOrder[]): Promise<void> {
    const url = PaypalApiPath.PATCH_ORDER.replace("{id}", orderId)
    return await this.httpClient_.request({ url, method: "PATCH" })
  }
```

Perhaps the fix is to actually pass the data into the call?

Arsenalist:
PR which fixes this: [https://github.com/medusajs/medusa/pull/8456](https://github.com/medusajs/medusa/pull/8456)

srindom:
Thanks for the fix! 💪

ximenadev:
hi @srindom, I'm having this issue and looking at the code into `/medusa-payment-paypal/dist/core/paypal-sdk.js` in [npm package](https://www.npmjs.com/package/medusa-payment-paypal?activeTab=code) I do not the the fix @Arsenalist made. 

is there some way to fix this?

ximenadev:
Hi @srindom @carlos-r-l-rodrigues , having a similar bug, but instead of editing the quantity for the item I get it after adding another item or a shipping method

looking [into the code](https://www.npmjs.com/package/medusa-payment-paypal?activeTab=code) I think this update wasn't published

carlos-r-l-rodrigues:
@olivermrbl https://github.com/medusajs/medusa/pull/8455 is missing a changeset and wasn't included in the release.

Thanks @ximenadev for the heads up

olivermrbl:
~~Nice catch, will publish a new release later today.~~

[...]

PR #8456, opened 2024-08-05
Pass in data to PATCH call so cart amount is updated on quantity change

Arsenalist:
Fixes https://github.com/medusajs/medusa/issues/8455


carlos-r-l-rodrigues (approved):
LGTM

PR #8457, opened 2024-08-05
chore(medusa,core-flows): cart payment collection link

carlos-r-l-rodrigues:


integration-tests/modules/__tests__/cart/store/cart.workflows.spec.ts
```diff
@@ -1280,15 +1269,15 @@ medusaIntegrationTestRunner({
 
           const paymentCollection =
             await paymentModule.createPaymentCollections({
-              amount: 5000,
+              amount: 5001,
```
carlos-r-l-rodrigues, line 1272: Changed here to test when the value has changed.
if the amount and currency_code are the same, the payment collection isn't updated and the session isn't deleted.

packages/core/core-flows/src/definition/cart/workflows/add-to-cart.ts
```diff
@@ -115,10 +115,15 @@ export const addToCartWorkflow = createWorkflow(
     parallelize(
       refreshCartShippingMethodsStep({ cart }),
       updateTaxLinesStep({ cart_or_cart_id: input.cart.id, items }),
-      refreshCartPromotionsStep({ id: input.cart.id }),
-      refreshPaymentCollectionForCartStep({ cart_id: input.cart.id })
+      refreshCartPromotionsStep({ id: input.cart.id })
     )
 
+    refreshPaymentCollectionForCartWorkflow.runAsStep({
```
carlos-r-l-rodrigues, line 122: moved outside the parallel step as operations there can affect Cart totals.

packages/core/core-flows/src/definition/cart/workflows/refresh-payment-collection.ts
```diff
@@ -44,47 +23,66 @@ export const refreshPaymentCollectionForCartWorkflow = createWorkflow(
...
-        selector: { id: data.cart.payment_collection.id },
-        update: {
-          amount: data.cart.total,
-          currency_code: data.cart.currency_code,
-        },
-      }
-    })
+      return true
+    }).then(() => {
+      const deletePaymentSessionInput = transform(
```
riqwan, line 53: q: when payment value is equal, but payment session's payment provider is different, this wouldn't delete existing sessions right? That was an issue we addressed with this one

packages/core/core-flows/src/definition/cart/workflows/update-line-item-in-cart.ts
```diff
@@ -92,10 +92,13 @@ export const updateLineItemInCartWorkflow = createWorkflow(
 
     parallelize(
       refreshCartShippingMethodsStep({ cart }),
```
riqwan: Not related, but promotions step is kinda dependent on this completing first as we use the shipping methods as a part of the context. Parallelising this would lead to a temporary wrong result.

packages/core/types/src/cart/workflows.ts
```diff
@@ -93,9 +93,6 @@ export interface UpdateCartWorkflowInputDTO {
 
 export interface CreatePaymentCollectionForCartWorkflowInputDTO {
   cart_id: string
-  region_id: string
```
riqwan, line 96: note: you'd also have to update the starter to account for these

riqwan (approved):
Added some comments, overall lgtm! Would recco a second pair of eyes as well.

packages/core/core-flows/src/definition/cart/workflows/refresh-payment-collection.ts
carlos-r-l-rodrigues: From what I can understand, the only way to select a payment provider is by calling `/store/payment-collections/:id/payment-session` and that will trigger `createPaymentSessionsWorkflow` which will delete existing payment sessions.

packages/core/core-flows/src/definition/cart/workflows/update-line-item-in-cart.ts
carlos-r-l-rodrigues: yeah, good point.
moved them to outside the parallel as well.

packages/core/types/src/cart/workflows.ts
carlos-r-l-rodrigues: just had to adjust the `js-sdk` here.

olivermrbl (approved):
Nice!

PR #8463, opened 2024-08-06
docs-util: added workflows typedoc plugin

shahednasser:
Add a typedoc plugin that, when it finds a workflow, extracts its steps (differentiating between steps, workflows as steps, and hooks) and attaches them as documents to the workflow's reflection.

In a next PR, this attached information is used in the markdown theme generating the browsable workflow reference.

Closes DOCS-861


olivermrbl (approved):
Uhh, sounds nice – like where this is going

PR #8471, opened 2024-08-06
fix(core-flows): copy line items attribute

carlos-r-l-rodrigues:
removing `requires_shipping` from variant won't be part of this PR

### DIFF
--- a/packages/core/core-flows/src/definition/cart/workflows/complete-cart.ts
+++ b/packages/core/core-flows/src/definition/cart/workflows/complete-cart.ts
@@ -92,9 +92,9 @@ export const completeCartWorkflow = createWorkflow(
           unitPrice: item.raw_unit_price ?? item.unit_price,
           isTaxInclusive: item.is_tax_inclusive,
           quantity: item.raw_quantity ?? item.quantity,
-          metadata: item?.metadata ?? {},
-          taxLines: item.tax_lines || [],
-          adjustments: item.adjustments || [],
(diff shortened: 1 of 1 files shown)


packages/core/core-flows/src/definition/cart/workflows/complete-cart.ts
```diff
@@ -80,11 +81,32 @@ export const completeCartWorkflow = createWorkflow(
...
+        })
+      })
+
+      const shippinMethods = JSON.parse(
+        JSON.stringify(cart.shipping_methods ?? []),
+        (_, value) => {
+          delete value.id
+          return value
+        }
+      )
```
olivermrbl: nit: typo
```suggestion
      const shippingMethods = JSON.parse(
        JSON.stringify(cart.shipping_methods ?? []),
        (_, value) => {
          delete value.id
          return value
        }
      )
```

suggestion: The stringify -> parse is a bit hard to read. Would maybe prefer a loop here. Also, didn't this work previously? What happened

olivermrbl (approved):

packages/core/core-flows/src/definition/cart/workflows/complete-cart.ts
carlos-r-l-rodrigues: Just changed to make it explicit.
It was working, but not removing the `id`s the entities were reusing the same id from the cart.

PR #8472, opened 2024-08-06
fix(core-flows): Allow return requests with no shipping

olivermrbl:
Closes CC-265


packages/core/core-flows/src/order/workflows/return/confirm-return-request.ts
```diff
@@ -106,10 +106,6 @@ function prepareFulfillmentData({
 }
 
 function extractReturnShippingOptionId({ orderPreview, orderReturn }) {
-  if (!orderPreview.shipping_methods?.length) {
```
olivermrbl, line 109: We can't rely on this check, as the preview also contains the regular shipping methods of the order.

carlos-r-l-rodrigues: this is just skipping if nothing exist. not important, we could just modify the loop as `of orderPreview.shipping_methods ?? []`
Inside the loop, we check for the `actions` on each method and get the correct one.

olivermrbl: Yeah, I can do that before I merge 👍

packages/modules/order/src/services/actions/receive-return.ts
```diff
@@ -50,6 +50,9 @@ function updateReturnItems(returnEntry, items) {
       const data = items.find((i) => i.details.reference_id === item.item_id)
       if (!data) return
 
+      console.log("ITEM: ", item)
```
carlos-r-l-rodrigues, line 53: @olivermrbl Oups...

olivermrbl: Damn my bad :'((

PR #8483, opened 2024-08-07
chore(core-flows): util to format inventory input

carlos-r-l-rodrigues:


packages/core/core-flows/src/order/workflows/claim/confirm-claim-request.ts
```diff
@@ -310,14 +310,16 @@ export const confirmClaimRequestWorkflow = createWorkflow(
         }
       })
 
-      const formatedInventoryItems = confirmVariantInventoryWorkflow.runAsStep({
-        input: {
-          skipInventoryCheck: true,
-          sales_channel_id: (claim as any).order.sales_channel_id,
-          variants,
-          items,
+      const formatedInventoryItems = transform(
```
olivermrbl, line 313: question: Am I missing something – AFAICT we are no longer confirming the inventory, only preparing the items for it? It looks like we've replaced the sub-workflow with a transformer

carlos-r-l-rodrigues: In these places where we reserve the inventory we don't have to check it before, because the module itself is doing this exact same check before reserving it.
And the `confirmVariantInventoryWorkflow` is used in multiple other flows mutating the Cart.

olivermrbl: Ahh got it – in that case nice 😄

olivermrbl (approved):

PR #8488, opened 2024-08-07
feat(adshboard,types,medusa): enable adding notes/return reason to inbound claims

riqwan:
what:

- adds toast messages to failed actions
- add ability to create and update notes / return reason
- fixes types in claims components

RESOLVES CC-297

https://github.com/user-attachments/assets/a3a14384-49e5-4d20-a9ef-5379b1122740

### DIFF
--- a/packages/admin-next/dashboard/src/routes/orders/order-create-claim/components/claim-create-form/claim-create-form.tsx
+++ b/packages/admin-next/dashboard/src/routes/orders/order-create-claim/components/claim-create-form/claim-create-form.tsx
@@ -366,11 +366,8 @@ export const ClaimCreateForm = ({
           })
         )
       )
-        .filter((it) => it?.variant)
+        .filter((it) => !!it?.variant)
         .forEach((item) => {
-          if (!item) {
-            return
-          }
 
           const { variant } = item
           const levels = variant.inventory[0]?.location_levels


packages/admin-next/dashboard/src/routes/orders/order-create-claim/components/claim-create-form/claim-create-form.tsx
```diff
@@ -161,14 +166,18 @@ export const ClaimCreateForm = ({
...
-        inbound_items: previewItems.map((i) => ({
-          item_id: i.id,
-          quantity: i.detail.return_requested_quantity,
-          note: i.actions?.find((a) => a.action === "RETURN_ITEM")
-            ?.internal_note,
-          reason_id: i.actions?.find((a) => a.action === "RETURN_ITEM")?.details
-            ?.reason_id,
-        })),
+        inbound_items: previewItems.map((i) => {
+          const returnAction = i.actions?.find(
```
olivermrbl, line 170: thought: Should we check if we found a `returnAction` here before sending it to the backend, or does it not matter?

packages/admin-next/dashboard/src/routes/orders/order-create-claim/components/claim-create-form/claim-create-form.tsx
```diff
@@ -341,6 +368,10 @@ export const ClaimCreateForm = ({
       )
         .filter((it) => it?.variant)
         .forEach((item) => {
+          if (!item) {
+            return
+          }
```
olivermrbl:
```suggestion
        .filter((it) => !!it?.variant)
        .forEach((item) => {
```

olivermrbl (approved):
Nice work!

packages/admin-next/dashboard/src/routes/orders/order-create-claim/components/claim-create-form/claim-create-form.tsx
riqwan: The return action will always be present here. Right now in the previewItems array, there are only inbound items as we can only create those atm. When I get to creating other types of actions, I will have to change the logic of the previewItems here and extract inboundItems from there. At that point, this logic here will be removed.

olivermrbl: Got it, thx 👍

PR #8504, opened 2024-08-08
chore(core-flows): [1] export types and types, add basic TSDocs

shahednasser:
This PR exports types and steps from the core-flows package, and adds simple TSDocs for workflows / steps. This is essential for the workflows reference.

PR 1/n

### DIFF
--- a/packages/core/core-flows/src/customer/workflows/update-customers.ts
+++ b/packages/core/core-flows/src/customer/workflows/update-customers.ts
@@ -11,7 +11,7 @@ import {
 } from "@medusajs/workflows-sdk"
 import { updateCustomersStep } from "../steps"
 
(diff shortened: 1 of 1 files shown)


packages/core/core-flows/src/customer/workflows/update-customers.ts
```diff
@@ -11,17 +11,18 @@ import {
 } from "@medusajs/workflows-sdk"
 import { updateCustomersStep } from "../steps"
 
-type UpdateCustomersStepInput = {
+type UpdateCustomersWorkflowInput = {
```
carlos-r-l-rodrigues:
```suggestion
export type UpdateCustomersWorkflowInput = {
```

PR #8511, opened 2024-08-08
feat(medusa): filter shipping options by location

riqwan:
what:

- filters shipping options by location ID

RESOLVES CC-328


packages/medusa/src/api/utils/maybe-apply-link-filter.ts
```diff
@@ -37,21 +38,70 @@ export function maybeApplyLinkFilter({
...
+        another: []
+      }
+    }
+  }
+*/
+function transformFilterableFields(filterableFields: Record<string, unknown>) {
+  const result = {}
+
+  for (const key in filterableFields) {
+    if (filterableFields.hasOwnProperty(key)) {
```
carlos-r-l-rodrigues, line 82:
```suggestion
  for (const key of Object.keys(filterableFields)) {
```

packages/medusa/src/api/utils/maybe-apply-link-filter.ts
```diff
@@ -37,21 +38,70 @@ export function maybeApplyLinkFilter({
...
+          }
+
+          current = current[part]
+        }
+      }
+    }
+  }
+
+  return result
+}
```
carlos-r-l-rodrigues:
```suggestion
function transformFilterableFields(filterableFields: Record<string, unknown>) {
  const result = {};
  for (const key of Object.keys(filterableFields)) {
    const value = filterableFields[key];
    const keys = key.split(".");
    let current = result;

    // Iterate over the keys, creating nested objects as needed
    for (let i = 0; i < keys.length; i++) {
      const part = keys[i];
      current[part] ??= {};

      if (i === keys.length - 1) {
        // If its the last key, assign the value
        current[part] = value;
        break;
      }

      current = current[part];
    }
  }

  return result;
}

```

carlos-r-l-rodrigues (commented):
LGTM

carlos-r-l-rodrigues (approved):

PR #8513, opened 2024-08-08
chore(order): item update quantity

carlos-r-l-rodrigues:


packages/modules/order/src/utils/actions/item-update.ts
```diff
@@ -0,0 +1,69 @@
...
+      action.details.quantity,
+      existing.detail?.fulfilled_quantity
+    )
+
+    if (lower) {
+      throw new MedusaError(
+        MedusaError.Types.INVALID_DATA,
+        `Item ${refId} has already been fulfilled and quantity cannot be lower than ${existing.detail?.fulfilled_quantity}.`
+      )
+    }
```
olivermrbl, line 67: I am not sure I understand this check + error message, but might just be missing some context.

What we want to ensure is that you can only update the unfulfilled quantity of an item right? 

So there are a couple of different cases to solve for:

**Item One**
- quantity: 3
- fulfilled_quantity: 0
You can update the quantity to anything, even 0 (i.e. remove it). 

**Item Two**
- quantity: 3
- fulfilled_quantity: 3
You can increase the quantity by as much as you want, but you cannot decrease the quantity because it is fully fulfilled.

**Item Three**
- quantity: 3
- fulfilled_quantity: 2
You can increase the quantity by as much as you want, but only decrease the quantity by 1. The "decreasable" quantity is `quantity - fulfilled_quantity`. 

It is not clear to me if we are solving for all three cases here. Can you confirm or deny this?

carlos-r-l-rodrigues: That is it. All these 3 examples are true.
This action "ITEM_UPDATE" is to set an arbitrary number in the quantity, different from "ITEM_ADD" and "ITEM_REMOVE".

So, we need to compare if the final quantity set is equal or bigger than what was fulfilled. Otherwise, we throw the error.

olivermrbl: Okay cool, sounds good 👍

olivermrbl (approved):

PR #8516, opened 2024-08-08
feat: create return reason

christiananese:
Closes CC-305

### DIFF
--- a/packages/admin-next/dashboard/src/routes/return-reasons/return-reason-list/components/return-reason-list-table/return-reason-list-table.tsx
+++ b/packages/admin-next/dashboard/src/routes/return-reasons/return-reason-list/components/return-reason-list-table/return-reason-list-table.tsx
@@ -64,6 +64,7 @@ export const ReturnReasonListTable = () => {
         isLoading={isPending}
         columns={columns}
         pageSize={PAGE_SIZE}
+        noHeader={true}
         pagination
         search
       />


packages/admin-next/dashboard/src/components/table/data-table/data-table-root/data-table-root.tsx
```diff
@@ -52,6 +52,10 @@ export interface DataTableRootProps<TData> {
    * Whether the table is empty due to no results from the active query
    */
   noResults?: boolean
+  /**
+   * Whether to display the tables header
```
christiananese, line 56: TBH I'm not a big fan of "negative" naming for props, but wanted to stay consistent with other props in this component

olivermrbl:
We might want to have a max-width on the column and truncate the text – wdyt?

CleanShot 2024-08-08 at 21 12 37@2x

olivermrbl:
We also need to increase the padding of the content in the row to follow the design:
CleanShot 2024-08-08 at 21 15 59@2x

olivermrbl:
We're missing a subtitle on the table:

**Currently:**
CleanShot 2024-08-08 at 21 19 30@2x

**Design:**
CleanShot 2024-08-08 at 21 20 03@2x

packages/admin-next/dashboard/src/providers/keybind-provider/hooks.tsx
```diff
@@ -264,6 +264,14 @@ export const useGlobalShortcuts = () => {
       type: "settingShortcut",
       to: "/settings/locations",
     },
+    {
+      keys: {
+        Mac: ["G", ",", "M"],
```
olivermrbl, line 269: question: Would this be more fitting? Typically, the keyboard shortcut is coupled with the resource name
```suggestion
        Mac: ["G", ",", "R"],
```

olivermrbl (approved):
Nice, LGTM

PR #8533, opened 2024-08-09
feat(types,dashboard): ability to set shipping methods for claim

riqwan:
what:

- ability to set shipping methods for a claim

Note: Introducing this led to a bug being found - deleting a outbound shipping option throws an error as the trail of order_change_id breaks somewhere in the chain, I will create a ticket to address that. 

https://github.com/user-attachments/assets/bcbb996e-31a7-45ee-940e-b74187bed27f

RESOLVES CC-321
RESOLVES CC-293


olivermrbl (approved):
Nice, LGTM!

PR #8535, opened 2024-08-09
chore: dependecies reorg

carlos-r-l-rodrigues:
What:
 - Removes `Symbol.from` from DML types
 - Removes unnecessary dependencies from package.json
 - Exports `MEDUSA_CLI_PATH` from `@medusajs/framework`
 - Removes references to `workspaces:ˆ`
 - `@medusajs/types` as devDependency only
 - common packages that are in `@medusajs/utils` are `peerDependencies` elsewhere

* Note: 
`"@medusajs/types"` was added temporarily as a dependency in `utils`. A next PR adding `import type` everywhere will change hundreds/thousands of files


sradevski (approved):

carlos-r-l-rodrigues:
This will require some deeper investigation of dependency issues on Medusa projects.

olivermrbl:
This looks like a nice clean-up – it is ready for a review?

packages/core/types/src/dml/index.ts
```diff
@@ -33,7 +28,7 @@ export interface IDmlEntity<
   Schema extends DMLSchema,
   Config extends IDmlEntityConfig
 > {
-  [IsDmlEntity]: true
+  __isDmlEntity: boolean
```
thetutlage: We were using symbols since they don't leak in the autocomplete of TypeScript.

carlos-r-l-rodrigues: I couldn't find any good solution to type it and use the symbol coming from elsewhere.
any suggestion?
we can't import the same symbol from `@medusajs/types`

carlos-r-l-rodrigues: I can remove from typings, and just use the symbol here and potentially ignore TS if needed

packages/medusa/package.json
```diff
@@ -41,15 +41,15 @@
     "serve": "node dist/app.js",
     "test": "jest --silent --bail --maxWorkers=50% --forceExit"
   },
+  "peerDependencies": {
+    "@medusajs/medusa-cli": "*"
+  },
```
srindom: thought: would it make sense to make `@medusajs/medusa-cli` a dependency of `@medusajs/framework`. we want the bin to always be available when the framework package is installed so think it'd be fair.

srindom: I see now that it already is a dependency in @meduasjs/framework so is the peerDependency necessary?

carlos-r-l-rodrigues: It is like that. `@medusajs/medusa-cli` is a dependency of `@medusa/framework`, but just a peerDependency at `@medusajs/medusa`, that already have the framework as a dependency.

carlos-r-l-rodrigues: At this moment it wouldn't be required because I removed the `require("@medusajs/medusa-cli")` from medusa, but maybe we'll need some other things exported from the cli later, keeping it as a peerDependency at least we have no risk of requiring a different version of it here

srindom: Alright 👍

.changeset/curly-buses-allow.md
```diff
@@ -0,0 +1,48 @@
+---
```
olivermrbl: question: Should we remove this again?

olivermrbl (commented):
Overall LGTM, what will be the recommended approach for linking packages locally? Think this will need to be documented as a replacement for `medusa-dev-cli`

.changeset/curly-buses-allow.md
carlos-r-l-rodrigues: But without the changeset, how it will update the dependencies to use the correct packages?
When the changeset wasn't available, the `test-cli-with-database` was failing using old versions of dependencies

carlos-r-l-rodrigues:
> what will be the recommended approach for linking packages locally

I believe we are all using `"resolutions"` in the package.json to link local packages.

.changeset/curly-buses-allow.md
olivermrbl: Changesets does not really know anything about the actual versions used. It only looks at what version bumps are specified across all files in `.changesets/` and replaces the versions accordingly. 

With this change, we will have two changesets for all packages that both contain a `patch` bump. This won't really do anything except tell changeset a second time, that versions should be replaced with the next patch version in the range.

olivermrbl:
> I believe we are all using "resolutions" in the package.json to link local packages.

Yup, I do too. Just want to make sure we can give developers regardless of package-manager the same experience that `medusa-dev-cli` did. 

For starters, we can document the resolutions approach with yarn and follow up with overrides for npm (or a different strategy here). 

cc @shahednasser

olivermrbl (approved):
LGTM

olivermrbl:
Feel free to merge when tests are passing

.changeset/curly-buses-allow.md
carlos-r-l-rodrigues: What I mean is, how `test-cli-with-database` will get the correct version from the PRs to test the project if it is set to use the version "preview"?
My tests were all failing using old packages, not containing my updates, until I added the changeset.
So, maybe the issue is somewhere else, and I got the impression that something was reading the changeset to invalidate some sort of cache, etc

olivermrbl: > What I mean is, how test-cli-with-database will get the correct version from the PRs to test the project if it is set to use the version "preview"?

I guess we will have to change the way that action is set up and executed. Until now, it has not relied on published versions, but instead used `medusa-dev-cli` i.e. a local registry.

Now, with the dev CLI removed, we will have to use a published version, which is not happening in the updated workflow as far as I can see. It needs to integrate a publishing step. 

Unless we can set up linking (e.g. with resolutions) in the action too?

packages/modules/providers/payment-stripe/package.json
```diff
@@ -25,17 +25,17 @@
...
     "@types/stripe": "^8.0.417",
-    "awilix": "^8.0.1",
     "cross-env": "^5.2.1",
     "jest": "^29.7.0",
     "rimraf": "^5.0.1",
     "typescript": "^5.3.3"
   },
   "peerDependencies": {
-    "@medusajs/medusa": "^1.12.0"
+    "@medusajs/medusa": "^1.12.0",
```
olivermrbl: question: Can we remove this one as well?

olivermrbl (commented):
Very nice clean-up!

olivermrbl (approved):

packages/modules/providers/payment-stripe/package.json
carlos-r-l-rodrigues: Good catch. we don't need that anymore in plugins/providers.

PR #8552, opened 2024-08-12
feat(create-medusa-app): prompt for database name with credentials

### LINKED ISSUE #8494: create-medusa-app should not make assumptions about database name
## Bug report
`npx create-medusa-app` will ask for a **Postgres username** and a **Postgres password** , it never asks for a postgres database. it assumes the username is the database and requires there to be a database table that has the same name as the database user. this should not be required, and or assumed. either ask for the database or dont require the database name to match the username, espcially since none of that matters anyway, because the `npx create-medusa-app` creates its own database with the name medusa-xxxx

shahednasser:
If the default credentials can't connect to the database, prompt for database name along with user and password credentials.

Close #8494


riqwan (approved):
🏅

PR #8553, opened 2024-08-12
feat(core-flows): introduce a generic create entities step

riqwan:
what:

We end up doing the same things for create steps without gaining anything and same is true for other CRUD entities. We have created N files for N entities to CRUD. PoC, if others agree, will create for other ops as well. 

We can potentially remove around 50 files from just create alone.


packages/core/core-flows/src/order/workflows/create-order-change-actions.ts
```diff
@@ -18,6 +19,13 @@ export const createOrderChangeActionsWorkflow = createWorkflow(
   (
     input: WorkflowData<CreateOrderChangeActionDTO[]>
   ): WorkflowResponse<OrderChangeActionDTO[]> => {
-    return new WorkflowResponse(createOrderChangeActionsStep(input))
+    const orderChangeActions = createEntitiesStep({
+      moduleRegistrationName: ModuleRegistrationName.ORDER,
+      invokeMethod: "addOrderAction",
```
sradevski, line 24: I think this is working well. Another alternative we could also have the following args instead `(moduleName, modelName, data)`. It is a bit less reusable, but a bit closer to how we autogenerate methods for modules.

I am not suggesting it's better, just wanted to put it out there to see what others think.

packages/core/core-flows/src/common/steps/create-entities.ts
```diff
@@ -0,0 +1,50 @@
...
+  compensateMethod: string
+  entityIdentifier?: string
+  data: any[]
+}
+
+export const createEntitiesStepId = "create-entities-step"
+/**
+ * This step creates entities for any given module or resource
+ */
+export const createEntitiesStep = createStep(
```
carlos-r-l-rodrigues, line 15: I think it is ok to have this generic step, but I believe the ultimate solution will be modules auto generating their methods as steps.
That would cleanup a lot of steps we have and custom modules would also have their methods as steps by default.

carlos-r-l-rodrigues (approved):

PR #8564, opened 2024-08-12
chore(medusa,types): [7] Add request types to API routes

shahednasser:
Add request types as type arguments in API routes. This also includes adding / rearranging types under the `types` package.

PR 7/n


packages/medusa/src/api/admin/exchanges/[id]/cancel/route.ts
```diff
@@ -4,10 +4,11 @@ import {
   MedusaResponse,
 } from "../../../../../types/routing"
 import { AdminPostCancelExchangeReqSchemaType } from "../../validators"
+import { HttpTypes } from "@medusajs/types"
 
 export const POST = async (
   req: AuthenticatedMedusaRequest<AdminPostCancelExchangeReqSchemaType>,
-  res: MedusaResponse
+  res: MedusaResponse<HttpTypes.AdminExchangeResponse>
```
Comment (carlos-r-l-rodrigues) [inline packages/medusa/src/api/admin/exchanges/[id]/cancel/route.ts:11]: @shahednasser at the moment the Claim's cancel and Exchange's cancel workflows are returning void.
You can force cast the res.json here, but I will add later a remoteQuery call to return the correct type after the workflow is executed

carlos-r-l-rodrigues (approved):

PR #8567, opened 2024-08-12
chore(medusa,types): [10] Add request types to API routes

shahednasser:
Add request types as type arguments in API routes. This also includes adding / rearranging types under the `types` package.

PR 10/n


carlos-r-l-rodrigues (approved):
Thank you for organizing all that 🥇

PR #8582, opened 2024-08-13
Run migrations and links sync after loading the app

thetutlage:


packages/core/medusa-test-utils/src/medusa-test-runner.ts
```diff
@@ -172,6 +163,16 @@ export function medusaIntegrationTestRunner({
       throw error
     }
 
+    try {
+      console.log(`Running migrations and syncing links ${dbName}`)
+      dbUtils.pgConnection_ = await initDb({
+        env,
+      })
```
riqwan: q: before this, the db doesn't exist? Doesn't the bootstrapApp function need a pg connection to run the defaults workflow?

thetutlage: Ohh yeah. Actually, have to split this into two parts.

- Init DB before the app starts.
- Run migrations and sync links after the app starts. Because, we need these operations to pick migrations and links from the app too

riqwan (approved):
🥇

packages/core/medusa-test-utils/src/medusa-test-runner.ts
```diff
@@ -172,6 +179,14 @@ export function medusaIntegrationTestRunner({
       throw error
     }
 
+    try {
+      console.log(`Migrating database ${dbName}`)
+      await migrateDatabase()
```
riqwan: Just realized another problem - without the migrations the defaults will not run 😄

thetutlage: Yeah. Realized that too. It seems like, the defaults needs the migrations, but then app migrations are skipped without loading the app.

sradevski (approved):

PR #8590, opened 2024-08-14
feat(core-flows,dashboard): add fixes to allow only outbound or inbound claims

riqwan:
what:

- allows completing claim with only inbound items
- allows completing claim with only outbound items
- validates against creating claims when an active change order is present


packages/admin-next/dashboard/src/routes/orders/order-detail/components/order-summary-section/order-summary-section.tsx
```diff
@@ -406,6 +406,17 @@ const ReturnBreakdown = ({
 const Total = ({ order }: { order: AdminOrder }) => {
   const { t } = useTranslation()
 
+  const totalPaid = order.payment_collections.reduce(
```
olivermrbl: thought: Wonder if we should be getting this calculated from the backend?

packages/core/core-flows/src/order/utils/order-validation.ts
```diff
@@ -53,6 +53,15 @@ export function throwIfIsCancelled(
   }
 }
 
+export function throwIfPresent(obj: unknown, type: string) {
```
olivermrbl: question: Can't we have multiple order changes present at the same time? But just not allowed in a few select states, e.g. an open but not requested return.

olivermrbl (approved):
LGTM

packages/admin-next/dashboard/src/routes/orders/order-detail/components/order-summary-section/order-summary-section.tsx
riqwan: as in, setting this on the order totals?

packages/core/core-flows/src/order/utils/order-validation.ts
riqwan: Can we have multiple active order changes? cc @carlos-r-l-rodrigues

carlos-r-l-rodrigues: No, we can only have 1 active order change at time.
They are considered active when the status is "pending" or "requested" which means that the client must accept or decline that order change.

carlos-r-l-rodrigues: But we already have this check everywhere in a different shape.
and the module itself has this verification.

packages/admin-next/dashboard/src/routes/orders/order-detail/components/order-summary-section/order-summary-section.tsx
carlos-r-l-rodrigues: The order should be returning the total paid from its "transactions", which is inside the module.
It will be inside `order.summary`

packages/core/core-flows/src/order/utils/order-validation.ts
riqwan: I still ended up with a duplicate change order here in this workflow & caused all sorts of ruckus, thats why I added this validation.

packages/admin-next/dashboard/src/routes/orders/order-detail/components/order-summary-section/order-summary-section.tsx
carlos-r-l-rodrigues: https://github.com/medusajs/medusa/pull/8591

I adjusted it to only add the transaction on payment capture, and not after the authorization.

packages/core/core-flows/src/order/utils/order-validation.ts
carlos-r-l-rodrigues: We'll have to investigate this one more. the root cause doesn't seem to be this validation, as the module is handling this.

riqwan: removed this, can come back to this later! 🤞🏻

packages/admin-next/dashboard/src/routes/orders/order-create-claim/components/claim-create-form/claim-outbound-section.tsx
```diff
@@ -191,10 +191,18 @@ export const ClaimOutboundSection = ({
 
       return action && !!!action?.return?.id
     })
-
+    console.log("outboundShippingMethods -- ", outboundShippingMethods)
```
carlos-r-l-rodrigues: rm

carlos-r-l-rodrigues (approved):

PR #8595, opened 2024-08-14
docs: remove steps from workflows reference overview

shahednasser:
Steps aren't important to highlight in the Medusa Workflows Reference overview page, so this PR removes them from that page.


shahednasser:

olivermrbl (approved):

PR #8603, opened 2024-08-15
fix(payment): Partial refunds

olivermrbl:
**What**
- Pass refund amount to payment provider instead of full payment amount

**Why**
- Without it, payments would always be fully refunded regardless of the amount specified

Closes CC-336


riqwan (approved):
nice! How did you come across this? 😄

olivermrbl:
> nice! How did you come across this? 😄

Currently setting up an end-to-end store with 3rd party providers installed and noticed this happened with Stripe payments

PR #8617, opened 2024-08-15
feat(medusa,core-flows,types): API to create payment collections for order

riqwan:
what:

- adds an API to create a payment collection for order


packages/core/core-flows/src/order/workflows/create-order-payment-collection.ts
```diff
@@ -0,0 +1,135 @@
...
+        )
+
+        if (MathBN.lte(pendingPayment, 0)) {
+          throw new MedusaError(
+            MedusaError.Types.NOT_ALLOWED,
+            `Cannot create a payment collection for amount less than 0`
+          )
+        }
+
+        if (input.amount && MathBN.gt(input.amount, pendingPayment)) {
```
carlos-r-l-rodrigues, line 96: Maybe we don't need this limitation.
one could pre-charge the order 😄

packages/core/core-flows/src/order/workflows/create-order-payment-collection.ts
```diff
@@ -0,0 +1,135 @@
...
+      list: false,
+    }).config({ name: "payment-collection-query" })
+
+    throwIfActivePaymentCollectionExists({ paymentCollection })
+
+    const paymentCollectionData = transform(
+      { order, input },
+      ({ order, input }) => {
+        const pendingPayment = MathBN.sub(
+          order.summary.current_order_total,
```
carlos-r-l-rodrigues: here the value is serialized. you can read them from `raw_` to keep the precision

carlos-r-l-rodrigues (approved):

olivermrbl (approved):
LGTM

packages/core/core-flows/src/order/workflows/create-order-payment-collection.ts
riqwan: These are just guard rails until we remove them 😄

PR #8628, opened 2024-08-16
docs: redesign code highlights + add default code titles

shahednasser:
- change the colors for highlights by user, line highlights, and token highlights.
- Add default title for code blocks if none provided.
- Show a different icon at the beginning of a code block if text is a bash script. While I tested this out in multiple pages this may lead to some inconsistencies found in deeper pages but we can fix them as we find them

Closes DOCS-868, DOCS-904, DOCS-901

Preview: https://docs-v2-git-docs-code-highlights-medusajs.vercel.app/v2

### DIFF
--- a/www/packages/docs-ui/src/components/CodeBlock/index.tsx
+++ b/www/packages/docs-ui/src/components/CodeBlock/index.tsx
@@ -356,7 +356,9 @@ export const CodeBlock = ({
                     !codeTitle.length &&
                       (!noCopy || !noReport) &&
                       "xs:max-w-[83%]",
-                    noLineNumbers && "pl-docs_0.75",
+                    (noLineNumbers ||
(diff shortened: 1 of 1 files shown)


olivermrbl:
The curl snippet looks off, would probably also have the `>`  here:

CleanShot 2024-08-18 at 17 01 48

shahednasser:
@olivermrbl nice catch! will fix it

shahednasser:
@olivermrbl I added some padding to one-liner code snippets which we don't show line numbers for. I didn't add the terminal prefix to `curl` commands because for multi-line ones it'll become hacky (for curl commands you only show the `>` prefix for the first line, whereas for other terminal code snippets you'd show the `>` for every line).

olivermrbl:
@shahednasser, makes sense. Looks good now 👍

PR #8629, opened 2024-08-16
fix: Remove the endpoint to create a user

sradevski:


carlos-r-l-rodrigues (approved):

haf:
Why did you approve and merge a change that breaks the tests for the same endpoint that you removed?

carlos-r-l-rodrigues:
@haf the tests removed should break in this PR, we will investigate why they didn't.

PR #8630, opened 2024-08-16
feat(dashboard,core-flows,js-sdk,types,link-modules,payment): ability to copy payment link

riqwan:
what: 

- enables a button to create a payment link when a payment delta is present
- api to delete order payment collection
- adds a pending amount to payment collections

Note: Not the happiest with the decision on when to create a payment collection and when not to. The code should programatically create or delete payment collections currently to generate the right collection for the payment delta. Adding a more specific flow to create and manage a payment collection will help reduce this burden from the code path and onto CX/merchant.

Another issue I found is that the payment collection status doesn't get updated when payment is complete as it still gets stuck to "authorized" state

https://github.com/user-attachments/assets/037a10f9-3621-43c2-94ba-1ada4b0a041b

### DIFF
--- a/packages/core/medusa-test-utils/src/database.ts
+++ b/packages/core/medusa-test-utils/src/database.ts
@@ -30,6 +30,9 @@ export function getMikroOrmConfig({
     entities: Object.values(mikroOrmEntities),
     schema: schema ?? process.env.MEDUSA_DB_SCHEMA,
     debug: false,
+    pool: {
+      min: 2,
+    },
     migrations: {
       pathTs: pathToMigrations,
       silent: true,
--- a/packages/core/types/src/workflows-sdk/service.ts
+++ b/packages/core/types/src/workflows-sdk/service.ts
@@ -11,6 +11,7 @@ type FlowRunOptions<TData = unknown> = {
   context?: Context
   resultFrom?: string | string[] | Symbol
   throwOnError?: boolean
+  logOnError?: boolean
   events?: Record<string, Function>
 }
 
--- a/packages/core/utils/src/dal/mikro-orm/mikro-orm-create-connection.ts
+++ b/packages/core/utils/src/dal/mikro-orm/mikro-orm-create-connection.ts
@@ -93,6 +93,9 @@ export async function mikroOrmCreateConnection(
     schemaGenerator: {
       disableForeignKeys: false,
     },
-    pool: database.pool as any,
+    pool: {
+      min: 2,
+      ...database.pool,
+    },
   })
 }
--- a/packages/core/utils/src/modules-sdk/create-pg-connection.ts
+++ b/packages/core/utils/src/modules-sdk/create-pg-connection.ts
@@ -27,7 +27,7 @@ export function createPgConnection(options: Options) {
     pool: {
       // https://knexjs.org/guide/#pool
       ...(pool ?? {}),
-      min: (pool?.min as number) ?? 0,
+      min: (pool?.min as number) ?? 2,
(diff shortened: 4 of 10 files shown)


packages/modules/payment/src/models/payment-collection.ts
```diff
@@ -103,12 +103,12 @@ export default class PaymentCollection {
   payment_providers = new Collection<Rel<PaymentProvider>>(this)
 
   @OneToMany(() => PaymentSession, (ps) => ps.payment_collection, {
-    cascade: [Cascade.PERSIST, "soft-remove"] as any,
+    cascade: [Cascade.PERSIST] as any,
```
riqwan, line 106: unfortunately had to remove this from the cascade as its causing an ambiguity in the SQL generated as both payment session and payment have the same column called "payment_collection_id". The MikroORM util does not specify which columns to generate, thereby dumping both columns in the query causing this issue.

olivermrbl:
Just placed a new order. Should the "Copy payment link" be there when there is an active payment missing capture?

CleanShot 2024-08-18 at 17 29 44

cc @carlos-r-l-rodrigues

olivermrbl:
Upon capturing the payment, the "Copy payment link" button should disappear. However, it currently requires a full page refresh. Even though the amounts correctly change.

To reproduce:
- Place an order
- Capture payment
- Notice the payment collection still rendering
- Refresh page
- Notice the payment collection no longer rendering

CleanShot 2024-08-18 at 17 42 50

packages/admin-next/dashboard/src/routes/orders/order-detail/components/order-summary-section/order-summary-section.tsx
```diff
@@ -96,45 +100,61 @@ export const OrderSummarySection = ({ order }: OrderSummarySectionProps) => {
     return false
   }, [reservations])
 
+  // TODO: We need a way to link payment collections to a change order to
```
olivermrbl, line 103: thought: Could the link have an additional column `type` or something like that?

packages/core/core-flows/src/order/workflows/delete-order-payment-collection.ts
```diff
@@ -0,0 +1,47 @@
...
+      entry_point: "payment_collection",
+      fields: ["id", "status"],
+      variables: { id: input.id },
+      throw_if_key_not_found: true,
+      list: false,
+    }).config({ name: "payment-collection-query" })
+
+    throwUnlessStatusIsNotPaid({ paymentCollection })
+
+    removeRemoteLinkStep({
```
olivermrbl, line 43: question: Shouldn't we delete the actual payment collection as well? 🤔

riqwan: The link has cascade enabled in this PR that should take care of deleting the payment collection

Screenshot 2024-08-19 at 20 03 39

packages/admin-next/dashboard/src/routes/orders/order-detail/components/order-summary-section/order-summary-section.tsx
riqwan: yes, we can do that

packages/core/core-flows/src/order/workflows/delete-order-payment-collection.ts
olivermrbl: Ah, I missed that – that's neat!

PR #8631, opened 2024-08-16
docs: improve prep scripts in resources + modification to husky hooks

shahednasser:
- Improve the prep scripts of the resources project by only generating the files / slug map for changed files
- Change husky hooks to run `prep` in the `www` monorepo if there are changes in the docs.

> Note: since there seems to be a problem with the current husky hook, I added a check before running the current actions that ensures that are changes under `pacakges`. From my understanding, nobody in the team has husky set up, so this is mainly for my benefit to avoid issues when I'm using it for the docs 😄

> Another note: there's an overlap between this PR and PR #8623 . Once both PRs are merged I'll open another with some refactoring.


shahednasser:
@olivermrbl can you take a look at this PR?

PR #8651, opened 2024-08-19
docs: redesign footer

shahednasser:
- Redesign footer showing feedback, pagination, and edit button components
- Redesign the feedback and edit button components
- Fix for wide layout on `lg` screens and other general fixes

Closes DOCS-858

Preview: https://docs-v2-meo3uaonj-medusajs.vercel.app/v2


olivermrbl:
Should we make the table a little bit wider? There are many weird line breaks in the tables and it doesn't look super nice:
CleanShot 2024-08-20 at 12 37 59

Or alternatively, break the lines more naturally, e.g. at a dot or space.

shahednasser:
@olivermrbl I believe the comment is meant for the tables PR right? 😄

> Or alternatively, break the lines more naturally, e.g. at a dot or space.

The main problem with this approach is that the inline code snippets are long, so they end up either overflowing the table or with a scroll.

Will relay the feedback to Ludvig and see what we can come up with

PR #8654, opened 2024-08-19
feat(dashboard): Hitting escape restores previous value

kasperkristensen:
**What**
- Hitting "esc" while editing a DataGrid cell, will now restore the initial value.
- Cleanup of component, adds built-in loading state.

Resolves CC-267


olivermrbl (approved):
Works! Also, nice clean-up :)

PR #8666, opened 2024-08-19
feat(workflows-sdk): log on error

carlos-r-l-rodrigues:
What:
* `logOnError` option when executing a workflow.
* When `true`, when the workflow is finished, it will log all the errors that occured in the workflow.


sradevski (approved):
Nice!

PR #8674, opened 2024-08-20
feat(dashboard): Wrap each route in an ErrorBoundary

kasperkristensen:
**What**
- Updates the copy of the different error types
- Wraps each route (+ custom routes) in an ErrorBoundary to preserve app layout on error (sidebar and topbar)

Skærmbillede 2024-08-20 kl  12 35 53

**Note**
If the user goes to a route that does not exist at all, e.g. `/some-weird-url`, then we have no way of knowing if the user is inside of a context where we can render the sidebar and topbar (as they require the user to be authenticated). So in this case we still show an ErrorBoundary where the two aren't included (see second picture), and include a button that takes the user to "/", which depending on whether the user is logged in will take them to "/login" or "/orders".

image

Resolves CC-248


olivermrbl (approved):
> So in this case we still show an ErrorBoundary where the two aren't included (see second picture), and include a button that takes the user to "/", which depending on whether the user is logged in will take them to "/login" or "/orders".

Nice, sounds like a good solution.

PR #8675, opened 2024-08-20
feat: create auth provider identity

christiananese:


sradevski (approved):
LGTM!

packages/modules/auth/src/services/auth-module.ts
```diff
@@ -121,6 +121,36 @@ export default class AuthModuleService
...
+  ): Promise<AuthTypes.ProviderIdentityDTO | AuthTypes.ProviderIdentityDTO[]> {
+    const providerIdentities = await this.providerIdentityService_.create(
+      data,
+      sharedContext
+    )
+
+    return await this.baseRepository_.serialize<
+      AuthTypes.ProviderIdentityDTO[]
+    >(providerIdentities, {
+      populate: true,
```
sradevski, line 150: IIRC you don't need populate, as it will do it by default.

packages/modules/auth/src/services/auth-module.ts
```diff
@@ -121,6 +121,36 @@ export default class AuthModuleService
...
+      | AuthTypes.CreateProviderIdentityDTO,
+    @MedusaContext() sharedContext: Context = {}
+  ): Promise<AuthTypes.ProviderIdentityDTO | AuthTypes.ProviderIdentityDTO[]> {
+    const providerIdentities = await this.providerIdentityService_.create(
+      data,
+      sharedContext
+    )
+
+    return await this.baseRepository_.serialize<
+      AuthTypes.ProviderIdentityDTO[]
```
sradevski, line 148: This should be 
```suggestion
      AuthTypes.ProviderIdentityDTO | AuthTypes.ProviderIdentityDTO[]
```

packages/modules/auth/src/services/auth-module.ts
christiananese: Yes you are right

PR #8679, opened 2024-08-20
feat(dashboard,core-flows,js-sdk,types): ability to mark payment as paid

riqwan:
what:

- adds ability to mark payment collection as paid

https://github.com/user-attachments/assets/4297db53-5638-4ec9-b084-9faadc2995aa

### DIFF
--- a/integration-tests/http/__tests__/claims/claims.spec.ts
+++ b/integration-tests/http/__tests__/claims/claims.spec.ts
@@ -789,8 +789,6 @@ medusaIntegrationTestRunner({
             )
           ).data.payment_collection
 
-          console.log("finalPaymentCollection -- ", finalPaymentCollection)
-
           expect(finalPaymentCollection).toEqual(
             expect.objectContaining({
               currency_code: "usd",


packages/core/core-flows/src/order/workflows/mark-payment-collection-as-paid.ts
```diff
@@ -0,0 +1,78 @@
...
+
+    const payment = authorizePaymentSessionStep({
+      id: paymentSession.id,
+      context: { order_id: input.order_id },
+    })
+
+    capturePaymentWorkflow.runAsStep({
+      input: {
+        payment_id: payment.id,
+        captured_by: "req.auth_context.actor_id",
```
carlos-r-l-rodrigues: it should come from the input?

carlos-r-l-rodrigues (approved):

packages/core/core-flows/src/order/workflows/mark-payment-collection-as-paid.ts
riqwan: good catch

riqwan: pushed

olivermrbl:
You can only mark system payments as paid, correct?

riqwan:
@olivermrbl yes, thats correct. Did I miss something?

packages/admin-next/dashboard/src/routes/orders/order-detail/components/order-summary-section/order-summary-section.tsx
```diff
@@ -104,10 +110,54 @@ export const OrderSummarySection = ({ order }: OrderSummarySectionProps) => {
     (pc) => pc.status === "not_paid"
```
olivermrbl, line 110: question: Can we rely on this check? couldn't there be multiple unpaid collections?

olivermrbl:
> @olivermrbl yes, thats correct. Did I miss something?

No no, I am just trying to wrap my head around how this will work with other payment collections on the order 😄

riqwan:
> > @olivermrbl yes, thats correct. Did I miss something?
> 
> No no, I am just trying to wrap my head around how this will work with other payment collections on the order 😄

There's no way to get in this situation through the UI as of yet. So not possible right now. But when we link the payment collection to the order change, we can get rid of this check

packages/admin-next/dashboard/src/routes/orders/order-detail/components/order-summary-section/order-summary-section.tsx
riqwan: There's no way to get in this situation through the UI as of yet. So not possible right now. But when we link the payment collection to the order change, we can get rid of this check

PR #8683, opened 2024-08-20
feat: Separate registration from authentication in auth domain

olivermrbl:
### Context

Our current flow for signing in and signing up is causing confusion externally and internally. To understand why this might be the case, we will briefly outline how we currently sign in and up. In both cases, we fire a request to an endpoint to authenticate with credentials that match our chosen provider – in the example below, we will use the email-password provider. 

The request will look like this:

```ts
POST /auth/user/emailpass
{ "email": "oli@medusajs.com", password: "supersecret" }
-> { token: "ey..." }
```

You’ll get back a JWT token, which can be used to authenticate subsequent requests.

As mentioned, the above endpoint is used to both sign in and sign up. The behavior in the request tries to deal with both scenarios and respond accordingly, which has proven to be the root cause of the confusion:

- If the email does exist but the password is incorrect, we will respond with 401
- If the email does exist and the password is correct, we will respond with 200
- If the email does not exist, we will respond with 200
    - Because we assume you wish to register a new identity for authentication
    - Subsequent requests fail with a 401 because no associated actor (e.g. user) exists for the AuthIdentity

The two former flows are sensible; however, the latter is less so – especially when used in the context of **signing in**. The main concern raised is that you receive a 200 when attempting to sign in with a non-existing email. The common expectation is to receive a 401.

This PR attempts to eliminate this confusion and allow for a more explicit intent behind the requests.

### What

🚧 Breaking change

**Introduce endpoint `/auth/[scope]/[provider]/register`**

This PR adds an endpoint specifically for registering new AuthIdentities, thereby separating registration from authentication at the Rest API level. This change will only be used by providers that require registration.

**Introduce method `register` in auth provider and auth module**

This PR introduces a new method `register` to the auth provider and auth module interfaces. The idea is the same as with the endpoint, i.e. separate the two actions. Again, will only be used by providers that require registration.

For SSO providers, like Google, identity registration is less relevant because, in some sense, the identity already exists at the third-party provider. In a scenario like this, we go through the existing authentication flow using `POST /auth/[scope]/[provider]` and `authenticate`. 

**Examples**

To illustrate the different flows, here are examples using email-password and Google:

Sign up with email-password:
```ts
POST /admin/invites -> admin creates invite
POST /auth/user/emailpass/register -> user registers identity
POST /admin/invites/accept -> invite is accepted passing the invite + auth token 
```

Sign in with email-password:
```ts
POST /auth/user/emailpass -> authenticate with email-password
GET /admin/users/me -> get authenticated user
```

Sign up with Google:
```ts
POST /auth/user/google -> redirects to Google auth
POST /auth/user/google/callback -> Google hits callback URL, authenticates, and responds with user
POST /admin/invites/accept -> invite is accepted passing the invite + auth token 
```

Sign up with Google:
```ts
POST /auth/user/google -> redirects to Google auth
POST /auth/user/google/callback -> Google hits callback URL, authenticates, and responds with user
GET /admin/users/me -> get authenticated user
```

Sign up as customer with email-password:
```ts
POST /auth/customer/emailpass/register -> customer registers identity
POST /store/customers -> customer is created
```

Sign in with email-password:
```ts
POST /auth/customer/emailpass -> authenticate customer with email-password
```


sradevski (approved):
LGTM! I think this flow makes a lot more sense, using the same endpoint for both operations was bound to confuse people 👍

olivermrbl:
Alright, it seems we will be moving forward with this approach. Will clean up the PR and make it ready for review.

sradevski (approved):
LGTM code-wise as well 👍

olivermrbl:
As per discussion with @shahednasser , we'll merge this tomorrow afternoon. We need some additional docs changes.

420coupe:
Does this resolve the issue of not being able to create an account after making a purchase?

ie guest user checks out with email 420coupe@gmail.com. If user tries to register using 420coupe@gmail.com after the purchase 

image

olivermrbl:
@420coupe, what endpoint are you hitting here?

420coupe:
> @420coupe, what endpoint are you hitting here?

image

PR #8687, opened 2024-08-21
fix(dashboard): bust order preview cache to reset fulfilled quantity

riqwan:
what:

- busts order cache to reset fulfilled quantity

RESOLVES CC-325

### DIFF
--- a/www/apps/api-reference/components/Tags/Operation/CodeSection/Responses/Sample/index.tsx
+++ b/www/apps/api-reference/components/Tags/Operation/CodeSection/Responses/Sample/index.tsx
@@ -51,12 +51,14 @@ const TagsOperationCodeSectionResponsesSample = ({
             </select>
           )}
           {selectedExample && (
-            <CodeBlock
-              source={selectedExample.content}
-              lang={getLanguageFromMedia(Object.keys(response.content)[0])}
-              collapsed={true}
-              className="mt-2 mb-0"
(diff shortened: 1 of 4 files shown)


packages/admin-next/dashboard/src/hooks/api/fulfillment.tsx
```diff
@@ -35,7 +35,7 @@ export const useCancelFulfillment = (
     onSuccess: (data: any, variables: any, context: any) => {
       queryClient.invalidateQueries({ queryKey: fulfillmentsQueryKeys.lists() })
       queryClient.invalidateQueries({
-        queryKey: ordersQueryKeys.details(),
+        queryKey: ordersQueryKeys.all,
```
fPolic: There is no need to invalidate everything on `orders` if we just want to refresh preview.
You can pass `orderId` as a param in the hook and call `ordersQueryKeys.preview(id)`.

riqwan: good push, I'll do this for all the other ones i've done so far as well.

riqwan: @fPolic just changed. Invalidating just the `.detail(id)` isn't invalidating it correctly though.

fPolic (approved):

PR #8699, opened 2024-08-21
chore: Remove `prepublishOnly` script

olivermrbl:
The script is redundant because we always run a separate build step in our release pipelines.


packages/core/medusa-test-utils/package.json
```diff
@@ -12,7 +12,6 @@
     "node": ">=16"
   },
   "scripts": {
-    "prepublishOnly": "cross-env NODE_ENV=production tsc --build",
     "prepare": "cross-env NODE_ENV=production yarn run build",
```
adrien2p: Should we rm this as well?

carlos-r-l-rodrigues (approved):

packages/modules/providers/auth-emailpass/package.json
```diff
@@ -17,10 +17,9 @@
   "author": "Medusa",
   "license": "MIT",
   "scripts": {
-    "prepublishOnly": "cross-env NODE_ENV=production tsc --build",
     "test": "jest --passWithNoTests src",
     "test:integration": "jest --forceExit -- integration-tests/**/__tests__/**/*.spec.ts",
-    "build": "rimraf dist && tsc -p ./tsconfig.json",
+    "build": "rimraf dist && tsc --build ./tsconfig.json",
```
adrien2p: Suggestion: The path is not needed here as it is the default

packages/modules/providers/auth-google/package.json
```diff
@@ -17,10 +17,9 @@
   "author": "Medusa",
   "license": "MIT",
   "scripts": {
-    "prepublishOnly": "cross-env NODE_ENV=production tsc --build",
     "test": "jest --passWithNoTests src",
     "test:integration": "jest --forceExit -- integration-tests/**/__tests__/**/*.spec.ts",
-    "build": "rimraf dist && tsc -p ./tsconfig.json",
+    "build": "rimraf dist && tsc --build ./tsconfig.json",
```
adrien2p: Suggestion : same where it is specified and not different that the default config file name

sradevski (approved):

PR #8705, opened 2024-08-21
feat(core-flows): order edit request

carlos-r-l-rodrigues:
What:
- `admin/order-edits/:id/request`
- Reserve item quantity for updated items


packages/core/core-flows/src/order/workflows/update-order-changes.ts
```diff
@@ -0,0 +1,20 @@
...
+  WorkflowResponse,
+  createWorkflow,
+} from "@medusajs/workflows-sdk"
+import { updateOrderChangesStep } from "../steps"
+
+export const updateOrderChangesWorkflowId = "update-order-change"
+/**
+ * This workflow updates one or more order changes.
+ */
+export const updateOrderChangesWorkflow = createWorkflow(
```
riqwan, line 13: q: is this used anywhere?

riqwan (approved):
lgtm!

packages/core/core-flows/src/order/workflows/update-order-changes.ts
carlos-r-l-rodrigues: not really, I added it because I'm used to do it haha.
I will keep it anyway

PR #8706, opened 2024-08-21
docs: updates following authentication flow changes

shahednasser:
- Add a new Auth tag for both admin and store OAS
- Add new auth routes for both admin and store OAS. These are added manually as they're made specific for `user` and `customer` actor type
- Generate the OAS for the `api-reference` project
- Update existing Auth Module's docs to provide the general approach around using the authentication route and main service methods


shahednasser:
@olivermrbl if the authentication flow is released we need to get this PR in

olivermrbl (approved):

PR #8712, opened 2024-08-21
feat(medusa,types,core-flows): apply created_by values - claims, exchanges, returns, fulfillment

riqwan:
what:

- Applies created_by values - claims, exchanges, returns, fulfillment

RESOLVES CC-96


olivermrbl (approved):
LGTM 🚢

PR #8713, opened 2024-08-21
docs: fix link href for Cart module

ahrbil:
The cart module in the commerce modules list on the resources page points to the product module instead of the cart module.

Thank you, medusa team, for the amazing product, this is a small fix because I spent much time on the docs and use the shortcuts in this list. 🙏


shahednasser (approved):
Nice catch thanks!

PR #8717, opened 2024-08-22
feat: add missing crud to provider identity service

christiananese:


packages/modules/auth/src/services/auth-module.ts
```diff
@@ -149,6 +149,35 @@ export default class AuthModuleService
     >(providerIdentities)
   }
 
+  updateProviderIdentites(
```
carlos-r-l-rodrigues, line 152: This method is auto generated. When we don't overwrite it with some custom logic and just call the service, we don't need to add them here.
And I believe there is actually a typo here, it should be `updateProviderIdentities`

olivermrbl: Oh gosh, I missed the typo in the review. Nice catch

PR #8720, opened 2024-08-22
feat(utils): use dotenv-expand to allow variables within env

shahednasser:
Use [dotenv-expand](https://github.com/motdotla/dotenv-expand) in `loadEnv` to allow using variables within another in `.env` like this:

```
DB_NAME=medusa-1UH1
DATABASE_URL=postgres://postgres@localhost/$DB_NAME
```


carlos-r-l-rodrigues (approved):
LGTM

PR #8727, opened 2024-08-22
feat(create-medusa-app): set database name to project name

shahednasser:
- Set database name to `medusa-[project-name]`
- In case database already exists, prompt the user to enter a name for the database
- Change the environment variables to have a `DB_NAME` variable, and a `DATABASE_URL` that references `DB_NAME`.

### DIFF
--- a/packages/cli/create-medusa-app/src/utils/prepare-project.ts
+++ b/packages/cli/create-medusa-app/src/utils/prepare-project.ts
@@ -159,7 +159,6 @@ export default async ({
             )
             errorOccurred = migrations.rowCount == 0
           } catch (e) {
-            console.log(e)
             // avoid error thrown if the migrations table
             // doesn't exist
             errorOccurred = true


packages/cli/create-medusa-app/src/utils/create-db.ts
```diff
@@ -16,6 +16,12 @@ export default async function createDb({ client, db }: CreateDbOptions) {
   await client.query(`CREATE DATABASE "${db}"`)
 }
 
+async function doesDbExist (client: pg.Client, dbName: string): Promise<boolean> {
+  const result = await client.query(`SELECT datname FROM pg_catalog.pg_database WHERE datname='${dbName}';`)
+
+  return !!result.rowCount
+}
+
```
thetutlage, line 24: Does this query work fine? Because from my experience, the PG client throws an exception before running the query (in case the database is missing). 

This is how I handled it within the medusa monorepo.

https://github.com/medusajs/medusa/pull/8718/files#diff-0cd2ae0bc94a653c4b2dff2733a3dd0a2efea552d2ae9ca6a36b1fe28c0be600R96-R106

shahednasser: Worked fine in my testing 😄 I'll generate a snapshot and maybe you can give it a go?

thetutlage: Sure. Maybe in my case I was relying on the Mikro ORM so that is performing some checks.

shahednasser: @thetutlage seems like the snapshot isn't working. Can you go to `packages/cli/create-medusa-app` and run `yarn dev` to test it out? You can also use the `--directory-path` option to create the project in a different directory. For example, `yarn dev --directory-path ~/test`.

thetutlage: Sure. Lemme do it now!

thetutlage:
CleanShot 2024-08-23 at 15 38 12@2x

An unrelated issue. I am not able to press `Ctrl + C` to exist out of the command :)

packages/cli/create-medusa-app/src/utils/create-db.ts
thetutlage: Yes. That works. So in my case the issue is from the Mikro ORM then

shahednasser:
> I am not able to press Ctrl + C to exist out of the command :)

Yea this is related to `medusa-telemetry` which I'm not really aware why it blocks the execution 😅

thetutlage (approved):

PR #8729, opened 2024-08-22
feat(notification-sendgrid): include ability to handle attachments

420coupe:
feat(notification-sendgrid): include ability to handle attachments array if passed to dynamicTemplateData.attachments

attachments array objects must contain the following: 

```ts
interface Attachment {
  content: string
  filename: string
  type?: string
  disposition?: string
  content_id?: string
}```

### DIFF
--- a/packages/modules/providers/notification-sendgrid/src/services/sendgrid.ts
+++ b/packages/modules/providers/notification-sendgrid/src/services/sendgrid.ts
@@ -64,9 +64,15 @@ export class SendgridNotificationService extends AbstractNotificationProviderSer
         }))
       : undefined
 
-    const message = {
+    const data = notification.data || {}
+    const from =
+      typeof data.from === "string" && data.from.trim() !== ""
+        ? data.from
+        : this.config_.from
+
+    const message: sendgrid.MailDataRequired = {
       to: notification.to,
-      from: this.config_.from,
+      from: from,
       templateId: notification.template,
       dynamicTemplateData: notification.data as
         | { [key: string]: any }
--- a/packages/modules/providers/notification-sendgrid/src/services/sendgrid.ts
+++ b/packages/modules/providers/notification-sendgrid/src/services/sendgrid.ts
@@ -18,14 +18,6 @@ interface SendgridServiceConfig {
   from: string
 }
 
-interface Attachment {
-  content: string
-  filename: string
-  type?: string
-  disposition?: string
-  content_id?: string
-}
-
 export class SendgridNotificationService extends AbstractNotificationProviderService {
   protected config_: SendgridServiceConfig
   protected logger_: Logger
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

**Review by carlos-r-l-rodrigues — APPROVED** [review]
LGTM

@sradevski can you review this one?

**Comment by 420coupe** [comment]
updated docs to include attachments example.

**Comment by 420coupe** [comment]
Another update, include ability to change ``from`` email. From email must be verified as a single sender.

File: packages/modules/providers/notification-sendgrid/src/services/sendgrid.ts [hunk 1]
```diff
@@ -18,6 +18,14 @@ interface SendgridServiceConfig {
   from: string
 }
 
+interface Attachment {
+  content: string
+  filename: string
+  type?: string
```
sradevski: Let's call it `content_type` so it's clearer what it is. That should also remove the need for a comment in the code.

packages/modules/providers/notification-sendgrid/src/services/sendgrid.ts
```diff
@@ -18,6 +18,14 @@ interface SendgridServiceConfig {
...
+  disposition?: string
+  content_id?: string
```
sradevski: This one can be called `id`, as it is the attachment ID essentially.

packages/modules/providers/notification-sendgrid/src/services/sendgrid.ts
```diff
@@ -46,13 +54,30 @@ export class SendgridNotificationService extends AbstractNotificationProviderSer
...
+          type: attachment.type, // MIME type (e.g., 'application/pdf')
+          disposition: attachment.disposition ?? "attachment", // Default to 'attachment'
+          content_id: attachment.content_id ?? undefined, // Optional: unique identifier for inline attachments
+        }))
+      : undefined
+
+    const data = notification.data || {}
+    const from =
+      typeof data.from === "string" && data.from.trim() !== ""
+        ? data.from
```
sradevski: I am thinking we might want to add `from` as part of the `NotificationDTO` model (but optional). I can see it being useful in other channels as well, so it's better if it is a first-class citizen.

sradevski (commented):
Looks pretty good, just some minor comments 👍

packages/modules/providers/notification-sendgrid/src/services/sendgrid.ts
420coupe: content_id is the required var name from sendgrid

https://www.twilio.com/docs/sendgrid/api-reference/mail-send/mail-send#request-body

packages/modules/providers/notification-sendgrid/src/services/sendgrid.ts
420coupe: calling sendgrid defined variables. 

https://www.twilio.com/docs/sendgrid/api-reference/mail-send/mail-send#request-body

packages/modules/providers/notification-sendgrid/src/services/sendgrid.ts
420coupe: This was my thought, but i didnt want to add a breaking change. If we're moving towards making it a first-class i'll drop this piece.

sradevski: It's not a breaking change since it will be optional, so let's go for it 👍

packages/modules/providers/notification-sendgrid/src/services/sendgrid.ts
sradevski: I was only thinking we can name it `id` internally, and remap it to `content_id`, but it's a NIT one.

packages/modules/providers/notification-sendgrid/src/services/sendgrid.ts
sradevski: Same logic as below, as I am thinking we can reuse the attachment type for other providers as well, so we can name it as we think makes sense, and remap it in each provider.

420coupe:
@sradevski ok made em as optional first-class vars. 

tested both with attachments and without. 

~~actually hold. Re-testing attachments.~~

was on me didnt update my local path. everything works as intended. 

Updated docs as well.

sradevski (approved):

sradevski (approved):

www/apps/resources/app/architectural-modules/notification/sendgrid/page.mdx
```diff
@@ -172,6 +184,8 @@ In this subscriber:
 - By specifying the `email` channel, the SendGrid Notification Module Provider is used to send the notification.
 - The `template` property of the `create` method's parameter specifies the ID of the template defined in SendGrid.
 - The `data` property allows you to pass data to the template in SendGrid.
+- The `attachments` optional property allows you to pass attachments to the template in SendGrid.
+- The `from` optional property allows you to pass a single sender verified email, default is `SENDGRID_FROM` env variable.
```
shahednasser:
```suggestion
- The `from` optional property allows you to pass a single sender-verified email. If not provided, the value of the `from` configuration of the module is used.
```

shahednasser (changes requested):

shahednasser (approved):
LGTM

PR #8741, opened 2024-08-23
feat: add env editor utility to edit update .env files

thetutlage:
Fixes: FRMW-2673

We are planning to add the `db:create` command which will read the contents of the `.env` file and also modify/update it. 

The `EnvEditor` utility is the first step towards modifying the `.env` files. I have made future proof by also update the `.env.template` file, so that it stays in-sync with the `.env` file.

Also, both the `.env` and the `.env.template` files will be created by this utility if they do not exists


riqwan (approved):
nice!

carlos-r-l-rodrigues (approved):

olivermrbl (approved):

PR #8744, opened 2024-08-23
chore: dependencies reorg

carlos-r-l-rodrigues:


olivermrbl (approved):

packages/core/modules-sdk/package.json
```diff
@@ -36,6 +35,7 @@
     "resolve-cwd": "^3.0.0"
   },
   "peerDependencies": {
+    "@mikro-orm/core": "*",
```
adrien2p, line 38: Q: shouldn't we use fixed versions of mikro orm packages to not end up having multiple versions and leading to unsuccessful instance of checks under the hood?

adrien2p (approved):
nice clean up :)

packages/core/modules-sdk/package.json
carlos-r-l-rodrigues: I added it as an optional peer dependency everywhere, so none package manager will install it. And only `@medusajs/utils` will have the version fixed.
Same for `awilix`

PR #8751, opened 2024-08-24
docs(ui): update 3.4 Modules & Services v2 documentation

WEEFAA:
## What

Explicitly define the requirement for module to work.

Service Class must be created under the `service.ts` file at the root of the module.

## Reference
https://docs.medusajs.com/v2/basics/modules-directory-structure#servicets
> A module must have a main service. It's created in the service.ts file at the root of your module directory as explained in a [previous chapter](https://docs.medusajs.com/v2/basics/modules-and-services).

It is mentioned in `3.6 Modules and Directory structure` that is must be explicitly named `service.ts` at the root directory of the module. But in the actual page (`3.4 Modules and Services`), it wasn't mentioned. 

## Changes
- explicitly mentioned and reference the `service.ts` requirement of `3.6 Modules and Directory structure` in `3.4 Modules and Services`

Avoid confusion.

I subconsciously created the service in a completely different file (still at the root of the my custom module), thinking of my own convention.


shahednasser (approved):
Thanks for your contribution!

PR #8760, opened 2024-08-26
feature: add db:create command

thetutlage:
FIXES: FRMW-2664

This PR introduces a new command called `db:create`. The command is meant to create the database (if it is missing) and update the `.env` file after the operation has been successful.

> [!IMPORTANT]
> This command does not boot the application nor read the `medusa-config.js` file. The `db:create` command should work regardless of the state of the application. Otherwise, we might run into issues where a loader wants to establish the DB connection, but the database does not exist.

## How it works?
This is how the command works under the hood.

- The command reads the `.env` file using the `EnvEditor` and looks for `DATABASE_URL` and `DB_NAME` environment variables.
- The command will exit if the `DATABASE_URL` is missing. We need the `DATABASE_URL` to make the database connection.
- The user is prompted to specify the database name with a default value. The default value is set to the `DB_NAME` environment variable and falls back to the project directory name.
- Finally, taking all this info. The following steps are performed.
   - Establish a connection with the database server.
   - Check if the database already exists
   - If not, create the database
   - Update the `.env` file.

## Newly introduced utilities

The following utilities have been introduced to the `@medusajs/utils` package.

### EnvEditor
It reads the `.env` file as a text file and exposes the API to edit the values inside it and persist them on the disk.

### parseConnectionString
We need to parse the PG connection string and convert it into an object to safely remove the database name from the `DATABASE_URL` value. We need to remove the database name because we want to establish the connection without selecting this database since it might not exist in the first place.

### createClient
Creates the PostgreSQL client instance (Not the Mikro ORM)

### createDb
Creates the database

### dbExists
Check if the database exists.

### DIFF
--- a/packages/cli/medusa-cli/src/create-cli.ts
+++ b/packages/cli/medusa-cli/src/create-cli.ts
@@ -126,8 +126,8 @@ function buildLocalCommands(cli, isLocalProject) {
       handler: handlerP(newStarter),
     })
     .command({
-      command: `db:create`,
-      desc: `Manage migrations from the core and your own project`,
+      command: "db:create",
+      desc: "Create the database used by your application",
       builder: (builder) => {
         builder.option("db", {
           type: "string",
@@ -137,7 +137,7 @@ function buildLocalCommands(cli, isLocalProject) {
           type: "boolean",
           default: true,
           describe:
-            "Skip prompts and use the default value for the database name",
+            "Display prompts. Use --no-interactive flag to run the command without prompts",
         })
       },
       handler: handlerP(
--- a/packages/cli/medusa-cli/src/create-cli.ts
+++ b/packages/cli/medusa-cli/src/create-cli.ts
@@ -126,6 +126,28 @@ function buildLocalCommands(cli, isLocalProject) {
       handler: handlerP(newStarter),
     })
     .command({
+      command: `db:create`,
+      desc: `Manage migrations from the core and your own project`,
+      builder: (builder) => {
+        builder.option("db", {
+          type: "string",
+          describe: "Specify the name of the database you want to create",
+        })
+        builder.option("interactive", {
+          type: "boolean",
+          default: true,
+          describe:
+            "Skip prompts and use the default value for the database name",
+        })
+      },
+      handler: handlerP(
+        getCommandHandler("db/create", (args, cmd) => {
+          process.env.NODE_ENV = process.env.NODE_ENV || `development`
+          return cmd(args)
+        })
+      ),
+    })
+    .command({
       command: `telemetry`,
       describe: `Enable or disable collection of anonymous usage data.`,
       builder: (yargs) =>
--- a/packages/core/utils/package.json
+++ b/packages/core/utils/package.json
@@ -49,7 +49,7 @@
     "prepublishOnly": "cross-env NODE_ENV=production tsc --build",
     "build": "rimraf dist && tsc --build",
     "watch": "tsc --build --watch",
-    "test": "jest --silent --bail --maxWorkers=50% --forceExit --testPathIgnorePatterns='/integration-tests/' -- src/**/__tests__/**/*.spec.ts",
+    "test": "jest --silent --bail --maxWorkers=50% --forceExit --testPathIgnorePatterns='/integration-tests/' -- src/**/__tests__/**/*.ts",
     "test:integration": "jest --silent --bail --runInBand --forceExit -- src/**/integration-tests/__tests__/**/*.ts"
   }
 }
--- a/packages/core/utils/package.json
+++ b/packages/core/utils/package.json
@@ -40,6 +40,8 @@
     "dotenv": "^16.4.5",
     "dotenv-expand": "^11.0.6",
     "jsonwebtoken": "^9.0.2",
+    "pg": "^8.12.0",
+    "pg-connection-string": "^2.6.4",
     "pluralize": "^8.0.0",
     "ulid": "^2.3.0"
   },
@@ -47,7 +49,7 @@
     "prepublishOnly": "cross-env NODE_ENV=production tsc --build",
     "build": "rimraf dist && tsc --build",
     "watch": "tsc --build --watch",
-    "test": "jest --silent --bail --maxWorkers=50% --forceExit --testPathIgnorePatterns='/integration-tests/' -- src/**/__tests__/**/*.ts",
+    "test": "jest --silent=false --bail --maxWorkers=50% --forceExit --testPathIgnorePatterns='/integration-tests/' -- src/**/__tests__/**/pg_utils.spec.ts",
     "test:integration": "jest --silent --bail --runInBand --forceExit -- src/**/integration-tests/__tests__/**/*.ts"
   }
 }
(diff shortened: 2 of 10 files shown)


packages/cli/medusa-cli/src/create-cli.ts
```diff
@@ -125,6 +125,28 @@ function buildLocalCommands(cli, isLocalProject) {
       desc: `Create a new Medusa project.`,
       handler: handlerP(newStarter),
     })
+    .command({
+      command: `db:create`,
+      desc: `Manage migrations from the core and your own project`,
```
olivermrbl: todo: Update the description to reflect the new command

packages/medusa/src/commands/db/create.ts
```diff
@@ -0,0 +1,117 @@
...
+      } catch (error) {
+        if (error.name === "ExitPromptError") {
+          process.exit()
+        }
+        throw error
+      }
+    } else {
+      dbName = defaultValue
+    }
+  }
```
adrien2p, line 63:
```suggestion
let dbName = db ?? envEditor.get("DB_NAME") ?? `medusa-${slugify(basename(directory))}`

if (interactive && !db) {
  try {
    dbName = await input({
      message: "Enter the database name",
      default: defaultValue,
      required: true,
    })
  } catch (error) {
    if (error.name === "ExitPromptError") {
      process.exit()
    }
    throw error
  }
}
```

adrien2p: this is just a suggestion, I found it more easy to read wdyt? feel free to discard :)

packages/core/utils/package.json
```diff
@@ -40,14 +40,16 @@
...
+    "pg-connection-string": "^2.6.4",
     "pluralize": "^8.0.0",
     "ulid": "^2.3.0"
   },
   "scripts": {
     "prepublishOnly": "cross-env NODE_ENV=production tsc --build",
     "build": "rimraf dist && tsc --build",
     "watch": "tsc --build --watch",
-    "test": "jest --silent --bail --maxWorkers=50% --forceExit --testPathIgnorePatterns='/integration-tests/' -- src/**/__tests__/**/*.ts",
+    "test": "jest --silent=false --bail --maxWorkers=50% --forceExit --testPathIgnorePatterns='/integration-tests/' -- src/**/__tests__/**/pg_utils.spec.ts",
```
olivermrbl:
```suggestion
    "test": "jest --silent --bail --maxWorkers=50% --forceExit --testPathIgnorePatterns='/integration-tests/' -- src/**/__tests__/**/*.ts",
```

packages/medusa/src/commands/db/create.ts
```diff
@@ -0,0 +1,117 @@
...
+    port: connectionOptions.port ? Number(connectionOptions.port) : undefined,
+    user: connectionOptions.user,
+    password: connectionOptions.password,
+  })
+
+  let hasEstablishedConnection = false
+  try {
+    await client.connect()
+    hasEstablishedConnection = true
+    logger.info("Connection established with the database")
```
adrien2p: suggestion: `Connection established with the database ${dbName}`

packages/medusa/src/commands/db/create.ts
```diff
@@ -0,0 +1,117 @@
...
+  } catch (error) {
+    process.exitCode = 1
+    hasEstablishedConnection = false
+    logger.error(
+      "Unable to establish database connection because of the following error"
```
adrien2p, line 87:
```suggestion
      "Unable to establish database connection"
```

packages/medusa/src/commands/db/create.ts
```diff
@@ -0,0 +1,117 @@
...
+    )
+    logger.error(error)
+  }
+
+  if (!hasEstablishedConnection) {
+    return
+  }
```
adrien2p: suggestion: I believe it is only possible to fall into that case if we go through the catch above, would it make sense to have the return in the catch directly and remove this check?

olivermrbl (approved):
Clean PR, LGTM

packages/cli/medusa-cli/src/create-cli.ts
adrien2p: Q: I am not sure the description is right here, could you confirm please?

packages/cli/medusa-cli/src/create-cli.ts
```diff
@@ -125,6 +125,28 @@ function buildLocalCommands(cli, isLocalProject) {
...
+      builder: (builder) => {
+        builder.option("db", {
+          type: "string",
+          describe: "Specify the name of the database you want to create",
+        })
+        builder.option("interactive", {
+          type: "boolean",
+          default: true,
+          describe:
+            "Skip prompts and use the default value for the database name",
```
adrien2p: suggestion: I feel like the description should be the otherway around wdyt?

packages/cli/medusa-cli/src/create-cli.ts
thetutlage: haha. Missed it :)

packages/core/utils/package.json
thetutlage: As always. I mess up things in the package.json file :)

packages/medusa/src/commands/db/create.ts
thetutlage: Yeah, makes sense!

packages/cli/medusa-cli/src/create-cli.ts
thetutlage: Since the default value for this prompt is `true`. You will instead use `--no-interactive` to hide the prompt.

packages/cli/medusa-cli/src/create-cli.ts
thetutlage: Yeah, its wrong

packages/medusa/src/commands/db/create.ts
thetutlage: Ohh yeah!

packages/medusa/src/commands/db/create.ts
thetutlage: I think telling the following error belongs to this line will be a little helpful

packages/cli/medusa-cli/src/create-cli.ts
adrien2p: But if you show the help it will show you interactive skip the prompts no?

packages/medusa/src/commands/db/create.ts
adrien2p: As you prefer ☺️ i just felt like since the error is straight below it was enough

packages/medusa/src/commands/db/create.ts
thetutlage: I think I find the existing flow of code a little better especially the way I read it. But again, I understand its a subjective thing.

packages/cli/medusa-cli/src/create-cli.ts
thetutlage: Hmmm. You are right

adrien2p (approved):

packages/cli/medusa-cli/src/create-cli.ts
thetutlage: What do you think of `Display prompts. Use --no-interactive flag to run the command without prompts`

adrien2p: I think it would be nicer 👍

PR #8765, opened 2024-08-26
fix(orchestration, workflow-sdk): Local workflow separated orchestrator

adrien2p:
FIXES TRI-174

**What**
Currently, every time a Local workflow is being instantiated, it will grab the global workflow definition including the orchestrator instance. This leads to issues when we have concurrent running workflows which all register their event listeners to this single orchestrator instance which can lead to exhausting the listerners. 
With this fix, every local workflow will have a copy of the global workflow definition plus a new instance (cloned) of the orchestrator meaning that from now on, every local workflow will have its own orchestrator.


packages/core/workflows-sdk/src/helper/workflow-export.ts
```diff
@@ -9,9 +9,9 @@ import {
 import { Context, LoadedModule, MedusaContainer } from "@medusajs/types"
 import {
   ContainerRegistrationKeys,
+  isPresent,
```
carlos-r-l-rodrigues, line 12: what kind of sorting does your IDE do? 🤣

carlos-r-l-rodrigues (approved):
LGTM

packages/core/workflows-sdk/src/helper/workflow-export.ts
adrien2p: I believe it is alphabetically sorting 😂

carlos-r-l-rodrigues: case-insensitive

adrien2p:
@olivermrbl can I merge this one?

olivermrbl:
Sure

packages/core/workflows-sdk/src/helper/workflow-export.ts
adrien2p: Yes just alphabetical 😅 i can have a look later if i can refine

PR #8766, opened 2024-08-26
fix: Add metadata to property to product category data model

Alexnortung:


sradevski (approved):
LGTM 👍

packages/modules/product/src/models/product-category.ts
```diff
@@ -112,6 +112,9 @@ class ProductCategory {
   @Property({ columnType: "timestamptz", nullable: true })
   deleted_at?: Date
 
+  @Property({ columnType: "jsonb", nullable: true })
+  metadata?: Record<string, unknown> | null
```
riqwan, line 116:
```suggestion
  metadata: Record<string, unknown> | null = null
```

olivermrbl:
Thanks for contribution @Alexnortung

packages/modules/product/src/models/product-category.ts
riqwan: You also need to add this to the types:

```
metadata?: Record<string, unknown> | null
```
 file: `https://github.com/medusajs/medusa/blob/develop/packages/core/types/src/http/product-category/common.ts#L5`

riqwan (approved):
one comment, but lgtm!

riqwan:
Sorry, bit late to the party 😄

packages/modules/product/src/models/product-category.ts
sradevski: Yeah we mentioned it in Slack but Oli merged already 😄

olivermrbl: Sorry must have missed that.

PR #8767, opened 2024-08-26
chore: Treat internal event differently, primarely do not display info logs for those events

adrien2p:
FIXES TRI-179

**What**
Add support for `internal` options as part of the emitted events. The internal can be treated differently from the classic events. As a first step, we are treating internal event in a way where we do not display info log about them such as the number of subscribers for a particular internal event. Though, we do still log warning in case someone still find out about an event and get an error from the handler


olivermrbl (commented):
Question: Since `@EmitEvents` decorator will likely only be used module services, would it be reasonable to default to `internal: true`?

adrien2p:
> Question: Since `@EmitEvents` decorator will likely only be used module services, would it be reasonable to default to `internal: true`? 

Well, originally i went with approximately what you are suggesting and then thought that a user could use the decorator and not mark it as internal and it would make the dx weird to have to set it to false every time you use the decorator. But from my understanding this is temporary and also maybe some logs can be removed as it is currently not really aligned across the event bus modules. My takes would be to specify it on our side only until we reach a point of certainty and then see what we do and in the meantime we dont change the users dx

olivermrbl:
My thinking was to not allow users to use module events for the time being. I feel like it's better to hide the concept entirely until we know how to deal with them. Having the `internal: true` as a default would mean they would never stumble into these events.

olivermrbl (approved):
LGTM

adrien2p:
@olivermrbl do you want to have a last review before merging ?

packages/modules/event-bus-local/src/services/event-bus-local.ts
```diff
@@ -54,7 +54,7 @@ export default class LocalEventBusService extends AbstractEventBusModuleService
         eventData.name
       )
 
-      if (!eventData.options?.internal) {
```
olivermrbl: what's the case here?

adrien2p: Options can be passed to the emit or as part of the message, so we check both

packages/modules/event-bus-local/src/services/event-bus-local.ts
```diff
@@ -48,9 +54,11 @@ export default class LocalEventBusService extends AbstractEventBusModuleService
         eventData.name
       )
 
-      this.logger_?.info(
-        `Processing ${eventData.name} which has ${eventListenersCount} subscribers`
-      )
+      if (options.internal || !eventData.options?.internal) {
```
adrien2p: I need to fix this

olivermrbl:
Let's merge this when we are ready to publish new release notes. Would like to highlight these changes, as it could otherwise be a bit disruptive for end-users.

PR #8774, opened 2024-08-27
docs: correct auth url

420coupe:
Place correct url for auth module.


shahednasser:
Nice catch! Thank you for your contribution! But this is actually resolved by another open PR (#8706) so will close this

PR #8775, opened 2024-08-27
feature: add sync links command

thetutlage:
Fixes: FRMW-2666

This PR introduces the `db:sync-links` command with two additional flags.

- The `--execute-safe` flag can be used to run only safe SQL commands without displaying any prompts. Helpful for CI
- The `--execute-all` flag can be used to run all SQL commands without displaying any prompts. Helpful for CI

The existing `links sync` command is still there, which internally invokes this new command. But the old command does not support the new flags.


packages/cli/medusa-cli/src/create-cli.ts
```diff
@@ -146,6 +146,26 @@ function buildLocalCommands(cli, isLocalProject) {
         })
       ),
     })
+    .command({
+      command: "db:sync-links",
+      desc: "Sync database schema with the links defined by your application and Medusa core",
+      builder: (builder) => {
+        builder.option("execute-all", {
+          type: "boolean",
+          describe: "Skip prompts and execute all (including unsafe) actions",
```
adrien2p, line 151: suggestion: same as for the db:create, I would expect it to be the opposite of skip by default, otherwise in the help it will also show `execute-all: skip...`

packages/cli/medusa-cli/src/create-cli.ts
```diff
@@ -146,6 +146,26 @@ function buildLocalCommands(cli, isLocalProject) {
...
+        })
+        builder.option("execute-safe", {
+          type: "boolean",
+          describe: "Skip prompts and execute only safe actions",
```
adrien2p, line 155: suggestion: same as the previous one

packages/medusa/src/commands/utils/index.ts
```diff
@@ -0,0 +1,21 @@
+import { logger } from "@medusajs/framework"
+import { MedusaContainer } from "@medusajs/types"
+import { ContainerRegistrationKeys } from "@medusajs/utils"
+
+export async function ensureDbExists(container: MedusaContainer) {
+  const pgConnection = container.resolve(
```
adrien2p, line 6: Q: if the database does not exists, the pg connection loader would have fail to create the pg_connection right? in that case, does this make sense and in that case should we do that before initializing the container in the sync links file?

thetutlage: I think creating the connection (ie the knex instance) does not check with the database server unless you issue a query.

packages/cli/medusa-cli/src/create-cli.ts
thetutlage: Did not get. 

The flag is used when you want to execute all SQL actions and skip prompts. So you will run the command as `npx medusa db:sync-links --execute-all`

packages/medusa/src/commands/utils/index.ts
adrien2p: you are right

packages/cli/medusa-cli/src/create-cli.ts
adrien2p: oh sorry, so execute all and execute safe will always skip the prompt, my bad 😅

adrien2p (approved):

olivermrbl (approved):
LGTM!

PR #8798, opened 2024-08-27
feat(dashboard): DataGrid improvements [4/4]

kasperkristensen:
**What**
- Changes cell state strategy from tracking from lazy to eager. This has required some changes to the API of the DataGrid component, and createDataGridColumnHelper function.
- Displays error messages in both affected cells and their rows. The row indicator also provides an option to quickly jump to an error.
- Allows the user to hide all rows and columns that don't have errors, to help quickly get an overview of the errors in a large grid.
- The first column of a DataGrid is now pinned, making it easier for a user to tell which entity they are editing.
- Fixes and improvements to column visibility menu.
- Adds a shortcuts modal that explains the different available keyboard commands.
- Updates `@tanstack/react-table` to the latest version.

Resolves CC-269


olivermrbl (approved):
Legend @kasperkristensen, tested it and works very nicely 🚢

PR #8813, opened 2024-08-27
fix(utils): DML hasOne - belongsTo not behaving correctly

### LINKED ISSUE #8768: Single "belongsTo" causes error in digital products example
I am struggling with the problem for already couple of days and after a lot of debugging, unfortunately my current suspcious that there is some kind of the bug.

I am following instruction: https://docs.medusajs.com/v2/resources/recipes/digital-products/examples/standard

But in my scenario, I would like to have only single relationship between `digitalProduct` and its child (in my case it is called `digitalProductReference` instead of media as in example). Here are my models:

```ts
const DigitalProduct = model.define("digital_product", {
  id: model.id().primaryKey(),
  name: model.text(),
  reference: model.hasOne(() => DigitalProductReference, {
    mappedBy: "digitalProduct"
  }),
  orders: model.manyToMany(() => DigitalProductOrder, {
    mappedBy: "products",
  }),
})
.cascades({
  delete: ["reference"],
})

export default DigitalProduct
```
```ts
const DigitalProductReference = model.define("digital_product_reference", {
  id: model.id().primaryKey(),
  type: model.enum(ReferenceType),
  licenceKey: model.text().nullable(),
  url: model.text().nullable(),
  digitalProduct: model.belongsTo(() => DigitalProduct, {
    mappedBy: "reference",
  }),
})

export default DigitalProductReference
```

And when I am trying to create entry in database, here is the error:
```
error:   insert into "public"."digital_product_reference" ("id", "type", "licence_key", "digital_product_id", "created_at", "updated_at") values ('01J62VKPNTTJH3PTP4BP26XYJS', 'licence_key', '456', '01J62VKPNV9FCCM61SS6DR8JMD', '2024-08-24T18:50:35.579Z', '2024-08-24T18:50:35.579Z', '01J62VKPNV9FCCM61SS6DR8JMD') returning "id", "created_at", "updated_at" - INSERT has more expressions than target columns
{
```

The example works properly with [array] of digitalProductReference (or digitalProductMedias) in `mappedBy`, but it fails as above in case of `model.hasOne` relationship.

for analysis purposes, here are the examples of `changeSets` which are going into database:
**working example**:
```
ChangeSet<DigitalProductReference> {
    entity: DigitalProductReference {
      digital_product_id: '01J62V9GN6R8KNQG75PV57X5QZ',
      id: '01J62V9GNMB00QRAR08P5SESMH',
      digitalProduct: [(DigitalProduct)],
      created_at: 2024-08-24T18:45:01.748Z,
      updated_at: 2024-08-24T18:45:01.748Z
    },
    type: 'create',
    payload: {
      id: '01J62V9GNMB00QRAR08P5SESMH',
      digital_product_id: '01J62V9GN6R8KNQG75PV57X5QZ',
      created_at: 2024-08-24T18:45:01.748Z,
      updated_at: 2024-08-24T18:45:01.748Z
    },
    name: 'DigitalProductReference',
    rootName: 'DigitalProductReference',
    collection: 'digital_product_reference',
    schema: 'public',
    originalEntity: undefined
  }
```

**not working example:**
```
ChangeSet<DigitalProductReference> {
    entity: DigitalProductReference {
      type: 'licence_key',
      licenceKey: '456',
      id: '01J62VKPNTTJH3PTP4BP26XYJS',
      digitalProduct: [DigitalProduct],
      digital_product_id: '01J62VKPNV9FCCM61SS6DR8JMD',
      created_at: 2024-08-24T18:50:35.579Z,
      updated_at: 2024-08-24T18:50:35.579Z
    },
    type: 'create',
    payload: {
      id: '01J62VKPNTTJH3PTP4BP26XYJS',
      type: 'licence_key',
      licenceKey: '456',
      digitalProduct: '01J62VKPNV9FCCM61SS6DR8JMD',
      created_at: 2024-08-24T18:50:35.579Z,
      updated_at: 2024-08-24T18:50:35.579Z,
      digital_product_id: '01J62VKPNV9FCCM61SS6DR8JMD'
    },
    name: 'DigitalProductReference',
    rootName: 'DigitalProductReference',
    collection: 'digital_product_reference',
    schema: 'public',
    originalEntity: undefined
  }
```

As you can see, in not working example, there is additional field `digitalProduct` in `payload` which probably causes errors as there are more columns than in database.

adrien2p:
FIXES FRMW-2676

**What**
ref: https://discord.com/channels/876835651130097704/1023889804544458752/threads/1276979858781503528

Currently, when providing the following

```ts
const user1 = manager.create(User, {
  username: "User 1",
  team: {
    name: "Team 1",
  },
})
```

It would result in an error inserting into the database because the foreign key will be sent twice as part of the insert, one for the relation and one for the foreign key that both relate to the foreign key property.

To fix that and allow both approaches (providing the entity to cascade persist or just providing the foreign key in case of another side nullable relation) we need to handle it a bit differently.

now both approaches would be valid. the entities for the example might not be the best ones but it is just to illustrate
option 1 - we create both the user and the team:
```ts
const user1 = manager.create(User, {
  username: "User 1",
  team: {
    name: "Team 1",
  },
})
```
option 2 - the team already exists (for example the previous user have been detached from the team but we kept the team alive and assign a new user to that team) :
```ts
const user1 = manager.create(User, {
  username: "User 1",
  team_id: team.id
})
```


adrien2p:
@thetutlage could you have a look to this one please, let me know if the description is not good enough and we can discuss further if you want

thetutlage:
What happens when I do this.

```ts
const user1 = manager.create(User, {
  username: "User 1",
  team: {
    id: 1,
    name: "Team 1",
  },
})
```

Also, it seems like we are supporting two ways to associate relationships. One is by passing a tree like structure and other one is by using the Foreign keys. 

Maybe we should check this internally once, because last time @srindom mentioned that we want our relationships API/behavior to be close to the SQL and hence in that case supporting tree like structure might not be something we want to do.

adrien2p:
> What happens when I do this.
> 
> 
> 
> ```ts
> 
> const user1 = manager.create(User, {
> 
>   username: "User 1",
> 
>   team: {
> 
>     id: 1,
> 
>     name: "Team 1",
> 
>   },
> 
> })
> 
> ```
> 
> 
> 
> Also, it seems like we are supporting two ways to associate relationships. One is by passing a tree like structure and other one is by using the Foreign keys. 
> 
> 
> 
> Maybe we should check this internally once, because last time @srindom mentioned that we want our relationships API/behavior to be close to the SQL and hence in that case supporting tree like structure might not be something we want to do.

Well, it seems reasonable to be able to pass the relationship to be also be inserted in the scenario of cascading the one to one. At least the way it was originally implemented allows it as the default configuration for the one to one cascade was not specified and therefore default to persist. In the case for example of user - information details you would likely want to pass the tree instead of creating the information details with a nullable user and then create the user with the information details right? Ill wait for seb answer and do the appropriate changes toward msg the final decision 👍

carlos-r-l-rodrigues (approved):

olivermrbl (approved):
I am fine with this 👍

PR #8818, opened 2024-08-27
fix(core-flows,order): return damaged items

carlos-r-l-rodrigues:
CLOSES: CC-363

### DIFF
--- a/integration-tests/http/__tests__/returns/returns.spec.ts
+++ b/integration-tests/http/__tests__/returns/returns.spec.ts
@@ -973,6 +973,7 @@ medusaIntegrationTestRunner({
               items: [
                 expect.objectContaining({
                   received_quantity: 2,
+                  damaged_quantity: 1,
                 }),
               ],
             })
--- a/packages/core/core-flows/src/order/workflows/return/confirm-receive-return-request.ts
+++ b/packages/core/core-flows/src/order/workflows/return/confirm-receive-return-request.ts
@@ -237,12 +237,24 @@ export const confirmReturnReceiveWorkflow = createWorkflow(
               itemUpdates[itemId].received_quantity,
               act.details!.quantity as BigNumberInput
             )
+
+            if (act.action === ChangeActionType.RECEIVE_DAMAGED_RETURN_ITEM) {
+              itemUpdates[itemId].damaged_quantity = MathBN.add(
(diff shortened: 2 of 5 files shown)


integration-tests/http/__tests__/returns/returns.spec.ts
```diff
@@ -973,6 +973,7 @@ medusaIntegrationTestRunner({
               items: [
                 expect.objectContaining({
                   received_quantity: 2,
+                  damaged_quantity: 1,
```
adrien2p, line 976: suggestion: it sounds like this test is checking a lot of things and could grow more. should it be breaking down for more clarity about the different return?

adrien2p (approved):

integration-tests/http/__tests__/returns/returns.spec.ts
fPolic, line 976: question: so in this case, the user receives 3 items: 2 regular and one damaged? If so should we adjust the condition in confirm receive return to count damaged items when setting return status to partialy/fully received?

riqwan: its testing a flow end to end, would be good to keep this as one test as each and every step before this affects the changes here and future changes

riqwan: q: should we also test for inventory changes to ensure this is reflected correctly there?

riqwan (approved):
lgtm!

integration-tests/http/__tests__/returns/returns.spec.ts
carlos-r-l-rodrigues: The `received_quantity` will be always the total received, the regular ones and the damaged ones.
If 3 items are returned, and 1 is damaged, the object will contain:
```
{
  received_quantity: 3,
  damaged_quantity: 1,
}
```

adrien2p: make sense, so you can always base on received quantity to not go over it for example 👍

carlos-r-l-rodrigues: This is already been checked right below this.
That "damaged_quantity" counter is only to display things correctly in the UI, but the rules and inventory management haven't changed, they are fine.

PR #8831, opened 2024-08-28
fix(order): run migration before setting not null

riqwan:
what:

- runs the migration to set not null values before setting not null
- updating existing migration is safe in this case as for those who this is broken will not have a migration timestamp set in the db and for those that didn't, it ran successffully without issues.

RESOLVES CC-400


packages/modules/order/src/migrations/Migration20240827133639.ts
```diff
@@ -3,7 +3,15 @@ import { Migration } from "@mikro-orm/migrations"
 export class Migration20240827133639 extends Migration {
   async up(): Promise<void> {
     this.addSql(
-      'alter table if exists "return_item" add column if not exists "damaged_quantity" numeric not null default 0, add column if not exists "raw_damaged_quantity" jsonb not null;'
+      'alter table if exists "return_item" add column if not exists "damaged_quantity" numeric not null default 0, add column if not exists "raw_damaged_quantity" jsonb;'
+    )
+
+    this.addSql(
+      `UPDATE "return_item" SET raw_damaged_quantity = '{"value": "0", "precision": 20}'::jsonb;`
```
riqwan, line 10: q: do we want to set this as a default value? or let the app code handle it?

olivermrbl: Hmm, my immediate thought was that this should not be part of the migration. How have we dealt with this elsewhere?

fPolic:
https://github.com/medusajs/medusa/pull/8832 😁

packages/modules/order/src/migrations/Migration20240827133639.ts
riqwan: I don't think we've had to run null migrations on big number js fields with data before.

packages/modules/order/src/migrations/Migration20240827133639.ts
```diff
@@ -3,7 +3,15 @@ import { Migration } from "@mikro-orm/migrations"
 export class Migration20240827133639 extends Migration {
   async up(): Promise<void> {
     this.addSql(
-      'alter table if exists "return_item" add column if not exists "damaged_quantity" numeric not null default 0, add column if not exists "raw_damaged_quantity" jsonb not null;'
+      'alter table if exists "return_item" add column if not exists "damaged_quantity" numeric not null default 0, add column if not exists "raw_damaged_quantity" jsonb;'
```
fPolic, line 6: q: this will also fail if there is existing data in the table right? we need to set default for `raw_damaged_quantity` here
EDIT: nvm you removed `not null`

packages/modules/order/src/migrations/Migration20240827133639.ts
adrien2p: I would say we let the app manages it, I don't think we have any default for other big numbers

adrien2p (approved):

olivermrbl:
Shall we merge?

riqwan:
Let's do it

PR #8834, opened 2024-08-28
fix: Validate boolean query params

olivermrbl:
Closes CC-70

### DIFF
--- a/integration-tests/http/__tests__/product/admin/product.spec.ts
+++ b/integration-tests/http/__tests__/product/admin/product.spec.ts
@@ -1092,6 +1092,58 @@ medusaIntegrationTestRunner({
             }),
           ])
         })
+
+        it("should get product variants filtered by manage_inventory", async () => {
+          const payload = {
+            title: "Test product - 1",
+            handle: "test-1",
+            variants: [
+              {
+                title: "Custom inventory 1",
+                prices: [{ currency_code: "usd", amount: 100 }],
+                manage_inventory: true,
+                inventory_items: [],
+              },
+              {
+                title: "Custom inventory 2",
+                prices: [{ currency_code: "usd", amount: 100 }],
+                manage_inventory: false,
+              },
+            ],
+          }
+
+          const product = (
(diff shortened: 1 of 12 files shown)


integration-tests/http/__tests__/product/admin/product.spec.ts
```diff
@@ -1092,6 +1092,58 @@ medusaIntegrationTestRunner({
             }),
           ])
         })
+
+        it.only("should get product variants filtered by manage_inventory", async () => {
```
sradevski:
```suggestion
        it("should get product variants filtered by manage_inventory", async () => {
```

packages/medusa/src/api/utils/common-validators/common.ts
```diff
@@ -32,7 +32,30 @@ const optionalBooleanMapper = new Map([
...
+ * Validates that a value is a boolean when it is passed as a string.
+ */
+export const booleanString = () =>
+  z
+    .union([z.boolean(), z.string()])
+    .refine((value) => {
+      if (typeof value === "boolean") {
+        return true
+      }
+      return value.toLowerCase() === "true" || value.toLowerCase() === "false"
```
sradevski:
```suggestion
      return ["true", "false"].includes(value.toString().toLowerCase())
```

packages/medusa/src/api/utils/common-validators/common.ts
```diff
@@ -32,7 +32,30 @@ const optionalBooleanMapper = new Map([
...
+    })
+    .transform((value) => {
+      if (typeof value === "boolean") {
+        return value
+      }
+      return value.toLowerCase() === "true"
```
sradevski:
```suggestion
      return value.toString().toLowerCase() === "true"
```

sradevski (approved):

integration-tests/http/__tests__/product/admin/product.spec.ts
olivermrbl: Damn

sradevski (approved):

PR #8841, opened 2024-08-28
fix(core-flows): use unit price of item in create cart and add to cart flows

shahednasser:
Use the `unit_price` property of an item in the `addToCart` and `createCart` workflows.


olivermrbl:
Think it's fine to pass it to the workflow, so perhaps we should update the steps to accept it?

shahednasser:
@olivermrbl updated the approach 👍🏻

olivermrbl (approved):
LGTM

PR #8843, opened 2024-08-28
chore(medusa): Re enable plugin loading

adrien2p:
### Summary

This PR re-enables the ability to load resources from external packages through the `plugins` property in the `medusa-config`.

**Resolves:** FRMW-2678

#### What’s Changed

- **Plugin Loading:**
  - The `plugins` property in `medusa-config` now supports loading resources from external packages from `node modules`, `plugins` directory or any other relative path outside of the project.
  - When loading plugins:
    - If the plugin is specified as a relative path and does not exist in the `plugins` directory of the local project, the system will resolve it using the relative path and will load the resources from either its `dist` or `src` directory, depending on how the project is started.
    - If the plugin is specified as a relative path and exist in the `plugins` directory of the local project, the system will resolve it using the relative path and will load the resources the directory itself
    - If the plugin is specified as an absolute path and/or is part of the node_modules we will load the resources from its `dist`

#### Note:
- This update distinguishes between local development plugins and external packages based on their location and path resolution.

### Testing

I created and tested a plugin in a local dummy project using the following structure:

- **Source code:** [src.zip](https://github.com/user-attachments/files/16782016/src.zip)

The test plugin successfully loaded and executed the following components:

- **API:** Added a store route that triggers an event and an admin route.
- **Subscribers:** Registered a subscriber that listens to the event emitted by the route.
- **Jobs:** Scheduled a job that executes every second.
- **Links:** Registered a new link.
- **Workflows:** Registered a new workflow.

**Configuration Used:**
```ts
  plugins: [{
    resolve: '../plugin-1'
  }],
  /*...*/
})


olivermrbl (approved):
Nice LGTM

adrien2p:
@olivermrbl shall we merge?

olivermrbl:
Yeah, think that should be fine right?

adrien2p:
yeah it shouldnt be a problem

PR #8845, opened 2024-08-28
fix(order): order change references

carlos-r-l-rodrigues:


integration-tests/http/__tests__/claims/claims.spec.ts
```diff
@@ -427,11 +427,11 @@ medusaIntegrationTestRunner({
 
           expect(orderResult.summary).toEqual(
             expect.objectContaining({
-              paid_total: 0,
+              paid_total: 61,
```
carlos-r-l-rodrigues, line 430: The reason for these adjustments is that this test was merged with a `.only`.

fPolic (approved):
LGTM

PR #8848, opened 2024-08-28
fix(dashboard) RMAs shipping pricing

fPolic:
**What**
- fix selecting inbound/outbound and custom shipping prices for claims and exchanges
- enable unsettling custom shipping price (note: temporarily we will show error popup from BD until custom price reset is added to BD)


olivermrbl (approved):

packages/admin-next/dashboard/src/routes/orders/order-create-claim/components/claim-create-form/claim-create-form.tsx
```diff
@@ -774,22 +774,27 @@ export const ClaimCreateForm = ({
...
                                 actionId = a.id
                               }
                             }
                           }
                         })
 
+                        const customPrice =
+                          customShippingAmount === ""
+                            ? null
+                            : parseInt(customShippingAmount)
```
carlos-r-l-rodrigues, line 791: should it be `parseFloat` here?

fPolic: yup, good catch, will push a fix

PR #8851, opened 2024-08-28
chore: Workflow execution JS-SDK

olivermrbl:


packages/core/types/src/http/workflow-execution/admin/entities.ts
```diff
@@ -1,11 +1,92 @@
+export type TransactionStepStatus =
```
olivermrbl, line 1: I got most of these types from admin, but not sure where to check to validate if they are correct

sradevski (approved):
LGTM! For the typings, I guess you already did this, but we can maybe check `workflow-engine-*` and check the response types there?

olivermrbl:
> LGTM! For the typings, I guess you already did this, but we can maybe check `workflow-engine-*` and check the response types there?

Yeah, I tried to cross-check, but many of the types in those packages are just `Record<string, unknown`s. I'll take one more pass at this and then merge.

packages/core/types/src/http/workflow-execution/admin/entities.ts
olivermrbl, line 1: @carlos-r-l-rodrigues, perhaps you could glance over these and let me know if something is off.

fPolic (approved):

packages/core/types/src/http/workflow-execution/admin/entities.ts
carlos-r-l-rodrigues: It looks great. Thanks!

PR #8852, opened 2024-08-28
fix(dashboard, js-sdk, types, medusa): separate between delete response with and without parent

shahednasser:
- Separate the previous `DeleteResponse` to `DeleteResponse` and `DeleteResponseWithParent`, as not every API route's delete response returns a parent. This ensures more accurate types shown in OAS / documentation.
- Use `DeleteResponse` or `DeleteResponseWithParent` in response API routes based on what they return
- Remove direct usage of `DeleteResponse` in API route, and instead create a type in the `type` package specific for the route / domain.
- Use the new types in the `js-sdk` and `dashboard`.


olivermrbl:
> Remove direct usage of DeleteResponse in API route, and instead create a type in the type package specific for the route / domain.

Is this to get a name for the reference? The generic type is quite nifty ~~and we are using it a lot in the js-sdk.~~

Nevermind, I commented before I checked the code 😄

olivermrbl (approved):
Nice clean-up!

PR #8853, opened 2024-08-28
feat: Add support for providers to validate their options at loading time

adrien2p:
RESOLVES FRMW-2679

**What**
- Add new `static validateOptions` method to all providers abstract class with a default implementation (do nothing).
- This pr add the Implementation of the validation to the different providers that currently validate their options.
- The provider loader once it has discovered the providers classes and before registering it in the container, will call the static validation method to validate the options and throw otherwise.


olivermrbl (dismissed):
Very nice!

olivermrbl (approved):
Now, I can do it! 😄

adrien2p:
> Now, I can do it! 😄 

Perfect timing now 👌

carlos-r-l-rodrigues (approved):

PR #8877, opened 2024-08-29
chore(types, medusa): [1/n] add missing query type arguments

shahednasser:

1/n


packages/core/types/src/http/invite/admin/queries.ts
```diff
@@ -0,0 +1,10 @@
+import { BaseFilterable, OperatorMap } from "../../../dal"
+
+export interface AdminGetInvitesParams extends BaseFilterable<AdminGetInvitesParams> {
```
olivermrbl: suggestion: Do we need `FindParams` here?

packages/core/types/src/http/order/admin/queries.ts
```diff
@@ -1,3 +1,6 @@
-import { BaseOrderFilters } from "../common";
+import { BaseFilterable } from "../../../dal";
 
-export interface AdminOrderFilters extends BaseOrderFilters {}
\ No newline at end of file
+export interface AdminOrderFilters extends BaseFilterable<AdminOrderFilters> {
```
olivermrbl: Same here.

olivermrbl (approved):
LGTM

PR #8879, opened 2024-08-29
chore(types, medusa): [2/n] add missing query type arguments

shahednasser:

2/n


packages/core/types/src/http/payment/admin/queries.ts
```diff
@@ -16,13 +16,20 @@ export interface AdminPaymentCollectionFilters
   extends BasePaymentCollectionFilters {}
 export interface AdminPaymentSessionFilters extends BasePaymentSessionFilters {}
 
-export interface AdminPaymentFilters extends BasePaymentFilters {}
+export interface AdminPaymentFilters extends BasePaymentFilters {
```
olivermrbl: suggestion :`FindParams`

olivermrbl: And on the other filters in the file too

olivermrbl (approved):
LGTM

PR #8880, opened 2024-08-29
chore(types, medusa): [3/3] add missing query type arguments

shahednasser:

3/3


packages/core/types/src/http/tax-rate/admin/queries.ts
```diff
@@ -1,16 +1,16 @@
...
+    q?: string
+    tax_region_id?: string | string[] | OperatorMap<string | string[]>
+    is_default?: string
+    service_zone_id?: string
+    shipping_profile_id?: string
+    provider_id?: string
+    shipping_option_type_id?: string
+    created_at?: string | OperatorMap<string>
+    updated_at?: string | OperatorMap<string>
+    deleted_at?: string | OperatorMap<string>
```
olivermrbl: thought: Should we remove string here?

olivermrbl (approved):
LGTM

packages/core/types/src/http/tax-rate/admin/queries.ts
shahednasser: Good point 👍🏻

PR #8896, opened 2024-08-30
fix: Customer registration

olivermrbl:
Closes CC-409


packages/core/core-flows/src/customer/workflows/create-customer-account.ts
```diff
@@ -19,8 +20,12 @@ export const createCustomerAccountWorkflowId = "create-customer-account"
  */
 export const createCustomerAccountWorkflow = createWorkflow(
   createCustomerAccountWorkflowId,
-  (input: WorkflowData<CreateCustomerAccountWorkflowInput>): WorkflowResponse<CustomerDTO> => {
-    const customers = createCustomersStep([input.customersData])
+  (
+    input: WorkflowData<CreateCustomerAccountWorkflowInput>
+  ): WorkflowResponse<CustomerDTO> => {
+    const customerData = validateCustomerAccountCreation(input)
```
sradevski: I am thinking this validation should happen inside the module (and it will implicitly due to the index), so maybe all we need to do is inside the module catch the error and make it more user-friendly? I am fine with having this validation but it is:
1. prone to race conditions
2. additional code we need to maintain

packages/core/core-flows/src/customer/steps/validate-customer-account-creation.ts
```diff
@@ -0,0 +1,49 @@
...
+
+      if (!hasExistingAccount && !input.authIdentityId) {
+        throw new MedusaError(
+          MedusaError.Types.DUPLICATE_ERROR,
+          "Guest customer with this email already exists"
+        )
+      }
+    }
+
+    input.customerData.has_account = !!input.authIdentityId
```
sradevski: It's a bit misleading that the validation also modifies the input, maybe we do this in a separate step?

sradevski (commented):
LGTM, let me know what you think about the validation comment

packages/core/core-flows/src/customer/workflows/create-customer-account.ts
olivermrbl: Yeah, I thought about adding it to the module, however it becomes a bit more tricky because we support bulk actions. Do you suggest we throw on the first error we catch? This is a pattern we haven't really implemented elsewhere, if I am not mistaken.

It is worth mentioning, this workflow is purpose-fully built for a single customer creation, which is why I deemed it a good fit for the logic.

packages/core/core-flows/src/customer/steps/validate-customer-account-creation.ts
olivermrbl: Yeah, can do!

packages/core/core-flows/src/customer/workflows/create-customer-account.ts
sradevski: I suggested that because it just feels like it's better suited to be part of the module, since the module holds the "has_account" flag. I think throwing on first error in a bulk operation and reverting is probably a good pattern in general, but we can go with this for now and revisit later.

In general, I think it's good to at least delegate validation to the modules for things that the module owns, but we don't need to dwell on it for too long IMO

sradevski (approved):

packages/core/core-flows/src/customer/workflows/create-customer-account.ts
olivermrbl: I'll keep this as is for now, and we can revisit in the future. Now, we have the error handling, so it will be pretty easy to refactor.

PR #8897, opened 2024-08-30
fix: handle case where product to be updated does not exist

thetutlage:
Fixes: FRMW-2680

Checking the product to exist before calling the workflow to update it.


packages/medusa/src/api/admin/products/[id]/route.ts
```diff
@@ -46,6 +57,20 @@ export const POST = async (
...
 
+  /**
+   * Check 2
+   *
+   * In order to prevent ourselves from reading the `results[0].id`, we
+   * should first check if the workflow returned an array with items
+   * or not.
+   *
+   * If not, should we consider that the product was not found? Or
+   * how do we handle this case.
```
Comment (adrien2p) [inline packages/medusa/src/api/admin/products/[id]/route.ts]: I would say yes, if nothing is returned we can take it like nothing happen

packages/medusa/src/api/admin/products/[id]/route.ts
```diff
@@ -38,6 +38,17 @@ export const POST = async (
...
 
+  /**
+   * Check 1
+   *
+   * Check if the product exists with the id or not before calling the workflow.
+   * Please look at the Check 2 as well. If we have Check 2, then do we need
+   * Check 1?
+   */
+  if (!(await refetchEntity("product", req.params.id, req.scope, ["id"]))) {
+    throw new MedusaError(MedusaError.Types.NOT_FOUND, "Product not found")
```
Comment (adrien2p) [inline packages/medusa/src/api/admin/products/[id]/route.ts]: Sounds good to me, would just change the message to `Product with id ${req.params.id} not found`

Comment (thetutlage) [reply packages/medusa/src/api/admin/products/[id]/route.ts]: 👍

packages/medusa/src/api/admin/products/[id]/route.ts
Comment (thetutlage) [reply packages/medusa/src/api/admin/products/[id]/route.ts]: 👍

Comment (adrien2p) [reply packages/medusa/src/api/admin/products/[id]/route.ts]: but here in theory it would not be possible to be in this scenario right?

Comment (thetutlage) [reply packages/medusa/src/api/admin/products/[id]/route.ts]: I think it can be if workflow applies filters when trying to find the product and those same filters are not applied by the "Case 1" check

Comment (olivermrbl) [reply packages/medusa/src/api/admin/products/[id]/route.ts]: Not sure I would have this check. `Product not found` is a bit misleading because the product does in fact exist, as verified in check 1. 

And if the workflow returns an empty result in this scenario, I would say something is wrong with the workflow.

Wdyt?

Comment (thetutlage) [reply packages/medusa/src/api/admin/products/[id]/route.ts]: Alright, if we think that as long as the product with id exists, the workflow should return a result, we can remove the second check

adrien2p (approved):

packages/medusa/src/api/admin/products/[id]/route.ts
```diff
@@ -38,6 +38,16 @@ export const POST = async (
 ) => {
   const { additional_data, ...update } = req.validatedBody
 
+  /**
+   * Check if the product exists with the id or not before calling the workflow.
+   */
+  if (!(await refetchEntity("product", req.params.id, req.scope, ["id"]))) {
```
Comment (sradevski) [inline packages/medusa/src/api/admin/products/[id]/route.ts]: NIT: Might be a bit cleaner if we move the refetch to not be inline with the `if`

sradevski (approved):

adrien2p (approved):
I do agree with stev on the comment, LGTM otherwise

packages/medusa/src/api/admin/products/[id]/route.ts
Comment (thetutlage) [reply packages/medusa/src/api/admin/products/[id]/route.ts]: Sure

thetutlage:
@olivermrbl Should I go ahead and update all the `UPDATE` routes? The issue persists with almost every route with the same or a slightly different behavior.

thetutlage:
Added the same check to all problematic routes.

packages/medusa/src/api/admin/campaigns/[id]/route.ts
```diff
@@ -36,6 +36,16 @@ export const POST = async (
   req: AuthenticatedMedusaRequest<AdminUpdateCampaignType & AdditionalData>,
   res: MedusaResponse<HttpTypes.AdminCampaignResponse>
 ) => {
+  const existingCampaign = await refetchCampaign(req.params.id, req.scope, [
```
Comment (riqwan) [inline packages/medusa/src/api/admin/campaigns/[id]/route.ts:39]: non-blocking suggestion:

These refetch*() methods underneath call the `refetchEntity` method, which calls the remoteQuery.

The find or throw scenario will be standard across all atomic requests, might be worth adding a flag to the fetchEntity method that passes to the remoteQuery's `throwIfKeyNotFound`, which should take care of the error messages as well

riqwan (approved):

packages/medusa/src/api/admin/campaigns/[id]/route.ts
Comment (thetutlage) [reply packages/medusa/src/api/admin/campaigns/[id]/route.ts:39]: Had a word for the same with @carlos-r-l-rodrigues this morning. The `throwIfKeyNotFound` doesn't work when the `id` is supplied in the `filters` object. Which seems to be the case with these helpers

Comment (riqwan) [reply packages/medusa/src/api/admin/campaigns/[id]/route.ts:39]: I see. If you decide to go this way, you can also safely override `refetchEntity` as filters are only needed for the list method `refetchEntities`.

Comment (thetutlage) [reply packages/medusa/src/api/admin/campaigns/[id]/route.ts:39]: Yeah thinking not to change too much in all the layers right now! Maybe, we can discuss all this one day and then do the cleanup :)

PR #8900, opened 2024-08-30
feat(notification): Handle long running transaction and add status support

adrien2p:
RESOLVES FRMW-2110
RESOLVES FRMW-2095

**What**
- Fix notification module to not retain transactions open while sending notification
- Add support for notification status [success, pending, failure]


packages/modules/notification/src/services/notification-module-service.ts
```diff
@@ -108,57 +109,114 @@ export default class NotificationModuleService
...
-            MedusaError.Types.NOT_FOUND,
-            `Could not find a notification provider for channel: ${entry.channel}`
-          )
-        }
+        const notificationsToProcess = data.filter(
+          (entry) =>
+            !entry.idempotency_key ||
+            !existsMap.has(entry.idempotency_key) ||
+            (existsMap.has(entry.idempotency_key) &&
+              [NotificationStatus.PENDING, NotificationStatus.FAILURE].includes(
```
sradevski: We probably shouldn't process the notification if it is in pending, since another process might already be processing it, so let's just compare against failures

carlos-r-l-rodrigues (approved):

packages/modules/notification/src/services/notification-module-service.ts
```diff
@@ -108,57 +109,114 @@ export default class NotificationModuleService
...
+          await this.notificationProviderService_.getProviderForChannels(
+            channels
           )
-        }
 
-        const res = await this.notificationProviderService_.send(
-          provider,
-          entry
+        // Create the notifications to be sent to prevent concurrent actions listing the same notifications
+        const normalizedNotificationsToProcess = notificationsToProcess.map(
```
sradevski, line 148: suggestion:
Instead of `alreadyExists` flag, should we just set the `idempotencyKey` as part of the returned object, and then below in the `create` call just do `filter(e => entry.idempotency_key && existsMap.has(entry.idempotency_key)`? I think it's a bit easier to follow, but it might just be me.

Feel free to go with whatever you think is better.

packages/modules/notification/src/services/notification-module-service.ts
```diff
@@ -108,57 +109,114 @@ export default class NotificationModuleService
...
+            notificationToUpdate.push(entry.data)
+
+            const errorMessage = !provider
+              ? `Could not find a notification provider for channel: ${entry.data.channel}`
+              : `Notification provider ${provider.id} is not enabled. To enable it, configure it as a provider in the notification module options.`
+
+            throw new MedusaError(MedusaError.Types.NOT_FOUND, errorMessage)
+          }
+
+          const res = await this.notificationProviderService_.send(
```
sradevski: Should we have a catch here and set the status to failure?

adrien2p: We already update all the status in the finally block for both success and failure and we do not have to re throw or anything as it is handled automatically

adrien2p: Let me know if you still think that i am missing something. But i ve added tests including both success and failure in order to validate the behavior

sradevski: We do the update there, but we never set the status to `Failure` in case the `.send` fails right? The test only covers the scenario where the provider isn't enabled/doesn't exist. Let me know if I misunderstand the code though

adrien2p: Ouch no you are completely right, ill fix that and add a test as well 👌

sradevski (approved):
Nice 👍

olivermrbl (approved):

PR #8906, opened 2024-08-30
feat: added totals tests for end 2 end RMA flow

riqwan:
what:

I've added some specs and comments in the specs for where I think totals are incorrect. 

Lets get this test merged in as the status quo and proceed to fix these issues one by one.

Additionally, this also removes the temporary_difference as its not something we use. 

RESOLVES CC-346


integration-tests/http/__tests__/order/admin/rma-flows.spec.ts
```diff
@@ -0,0 +1,801 @@
...
+          .data.order
+
+        // Totals summary after payment has been marked as paid
+        expect(orderResult.summary).toEqual(
+          expect.objectContaining({
+            paid_total: 131,
+            difference_sum: 70,
+            refunded_total: 10,
+            transaction_total: 121,
+            // Q: Should the pending difference be changed here? Its been refunded, so nothing is pending per-ce
```
riqwan: @carlos-r-l-rodrigues

integration-tests/http/__tests__/order/admin/rma-flows.spec.ts
```diff
@@ -0,0 +1,801 @@
...
+
+        // Totals summary after all
+        expect(orderResult.summary).toEqual(
+          expect.objectContaining({
+            paid_total: 131,
+            difference_sum: -20,
+            refunded_total: 10,
+            transaction_total: 121,
+            // TODO: This looks wrong
+            pending_difference: -10,
```
riqwan: @carlos-r-l-rodrigues If a refund is made outside of payment collections created during an RMA flow, the pending difference is always accounting for that refund

integration-tests/http/__tests__/order/admin/rma-flows.spec.ts
```diff
@@ -0,0 +1,801 @@
...
+            refunded_total: 10,
+            transaction_total: 121,
+            // TODO: This looks wrong
+            pending_difference: 15,
+            // TODO: This looks wrong
+            current_order_total: 136,
+            // TODO: This looks wrong
+            original_order_total: 146,
+            // TODO: This looks wrong
+            temporary_difference: 15,
```
riqwan: @carlos-r-l-rodrigues when I create 2 claims and none are fulfilled, the totals don't seem to be correct at all.

integration-tests/http/__tests__/order/admin/rma-flows.spec.ts
riqwan: The transactions probably needs to account for these rogue refund payments

riqwan:
cc @carlos-r-l-rodrigues

carlos-r-l-rodrigues (approved):

fPolic (approved):

PR #8912, opened 2024-08-30
feat(dashboard): allow custom shopping prices for claims/exchanges

fPolic:
**What**
- add editing custom shipping price for claims and exchanges

---

CLOSES CC-369


olivermrbl (commented):
Overall LGTM.

A couple of comments:
- Do we want to have a dedicated custom price removal func?
  - Or do we just expect the merchant to re-enter the original price, if they regret setting a custom one?
- Should we default to the price of the shipping? Otherwise, if you click to set custom price and remove focus, it will be set to 0. I don't think that's a good default behavior.

fPolic:
> Do we want to have a dedicated custom price removal func? Or do we just expect the merchant to re-enter the original price, if they regret setting a custom one?

Yes, but it requires changes on the BD, I opened a ticket for that (CC-402). I believe it's in. progress.

> Should we default to the price of the shipping? Otherwise, if you click to set custom price and remove focus, it will be set to 0. I don't think that's a good default behavior.

Yup good point. I will add a fix.

olivermrbl (approved):
LGTM

PR #8927, opened 2024-09-01
docs: update setup instructions to include cloud PostgreSQL option wi…

navjeevan-alone:
…th --db-url

I faced challenges connecting to PostgreSQL locally while setting up Medusa. Although I had a cloud-based PostgreSQL instance, it took a couple of hours to figure out how to connect it properly. Including the option to directly add a database URL using the --db-url flag will make the setup process easier for users with cloud-based databases.

### DIFF
--- a/www/apps/docs/content/starters/nextjs-medusa-starter.mdx
+++ b/www/apps/docs/content/starters/nextjs-medusa-starter.mdx
@@ -43,9 +43,26 @@ You can use the following command to install a full ecommerce store that include
 npx create-medusa-app@latest --with-nextjs-starter
 ```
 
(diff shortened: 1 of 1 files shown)


shahednasser (changes requested):
Thank you for your contribution @navjeevan-alone ! Can you instead just add a note like:

```
:::note

The setup process requires PostgreSQL installed and running locally. Alternatively, you can pass a `--db-url` option to the `create-medusa-app` command. Refer to the [create-medusa-app](../create-medusa-app.mdx) documentation for more details on prerequisites, steps, and troubleshooting.

:::
```

PR #8937, opened 2024-09-02
fix(core-flows): Handle variant creation duplicate inventory item ids

adrien2p:
FIXES CC-407

**What**
- Prevent the admin product create variant kit management to allow choosing duplicated inventory item as part of a variant inventory kit
- Prevent the workflow from accepting duplicated variant inventory item

https://github.com/user-attachments/assets/42b548cd-3010-4088-9a24-817c69a6a414


packages/core/core-flows/src/product/workflows/create-product-variants.ts
```diff
@@ -50,6 +54,41 @@ const buildLink = (
...
+    const duplicatedInventoryItemIds = inventoryItemIds.filter(
+      (id, index) => inventoryItemIds.indexOf(id) !== index
+    )
+
+    if (duplicatedInventoryItemIds.length) {
+      errors.push(
+        `Variant with id ${
+          variantData.variantId
+        } has duplicate inventory item ids ${duplicatedInventoryItemIds.join(
+          ", "
```
olivermrbl: suggestion(non-blocking): Simplify error message. Don't think the IDs help here, so if this bubbles up to admin, then it would be a more readable message
```suggestion
        `Cannot associate duplicate inventory items to a variant`
```

olivermrbl: Also, would maybe argue we can just throw on the first error we find, but I'll let you decide.

olivermrbl (approved):
LGTM, nice

packages/core/core-flows/src/product/workflows/create-product-variants.ts
adrien2p: The idea is the following, since the api can be consumed outside of the admin, if you send multiple variants with duplicated inventory items, you get the list of all variants to change instead of receiving one, change it, send them back and get another error. If thats ok for you ill keep it, but ill include your previous comment anyway

olivermrbl: Yeah, I am aware why we are usually doing this. Fine to keep it as is 👍

PR #8943, opened 2024-09-02
fix(dashboard): Product create price columns race condition

kasperkristensen:
Since the issue was not reproducible 100% of the time, I am not completely sure it's fixed with these changes, but haven't been able to reproduce it since. Would be great if reviewers would also test it out.

Resolves CC-283


olivermrbl:
Tested it and could not reproduce. From a logical standpoint, I also think with these changes it shouldn't happen. Let's get this in and see

olivermrbl (approved):

PR #8944, opened 2024-09-02
feat(core-flows,types): Refunds can only be performed when order is imbalanced

riqwan:
what:

Due to the fact that any adhoc refunds cause an imbalance in the order totals, we constrain the refunds to only be performed when there is an outstanding balance in the order

RESOLVES CC-412


packages/core/core-flows/src/payment/workflows/refund-payment.ts
```diff
@@ -11,6 +12,41 @@ import { emitEventStep, useRemoteQueryStep } from "../../common"
 import { addOrderTransactionStep } from "../../order/steps/add-order-transaction"
 import { refundPaymentStep } from "../steps/refund-payment"
 
+/**
+ * This step validates that the refund is valid for the order
+ */
+export const validateRefundStep = createStep(
+  "begin-claim-order-validation",
```
olivermrbl:
```suggestion
  "validate-refund-step",
```

packages/core/core-flows/src/payment/workflows/refund-payment.ts
```diff
@@ -11,6 +12,41 @@ import { emitEventStep, useRemoteQueryStep } from "../../common"
...
+    payment,
+    amount,
+  }: {
+    order: OrderDTO
+    payment: PaymentDTO
+    amount?: BigNumberInput
+  }) {
+    const pendingDifference = order.summary?.raw_pending_difference!
+
+    if (MathBN.gte(pendingDifference, 0)) {
```
olivermrbl, line 31: question: Can the pending different be greater than in this scenario? Wouldn't it always be 0 or am I reading this wrong?

packages/core/core-flows/src/payment/workflows/refund-payment.ts
```diff
@@ -11,6 +12,41 @@ import { emitEventStep, useRemoteQueryStep } from "../../common"
...
+      )
+    }
+
+    const amountPending = MathBN.mult(pendingDifference, -1)
+    const amountToRefund = amount ?? payment.raw_amount ?? payment.amount
+
+    if (MathBN.gt(amountToRefund, amountPending)) {
+      throw new MedusaError(
+        MedusaError.Types.INVALID_DATA,
+        `amount to refund cannot be greater than pending difference - ${amountPending}`
```
olivermrbl: nit: capitalize
suggestion: rephrase
```suggestion
        `Cannot refund more than pending difference - ${amountPending}`
```

olivermrbl (approved):
Overall LGTM, should be good to merge, but would be good to have @carlos-r-l-rodrigues eyes on it as well

packages/core/core-flows/src/payment/workflows/refund-payment.ts
riqwan: yes, it can be a positive number as well in case there is pending payment from the customer side

olivermrbl: Got it!

carlos-r-l-rodrigues (approved):

PR #8961, opened 2024-09-03
fix(medusa): fix complete order API route to remove duplicate params

shahednasser:
Remove the `order_id` request body param from the complete order API route and use the `id` path parameter instead.

### DIFF
--- a/packages/medusa/src/api/admin/orders/validators.ts
+++ b/packages/medusa/src/api/admin/orders/validators.ts
@@ -52,11 +52,7 @@ export const AdminArchiveOrder = z.object({
 })
 export type AdminArchiveOrderType = z.infer<typeof AdminArchiveOrder>
 
-export type AdminCompleteOrderType = z.infer<typeof CompleteOrder>
-export const CompleteOrder = z.object({
-  order_id: z.string(),
-})
-export const AdminCompleteOrder = WithAdditionalData(CompleteOrder)
+export const AdminCompleteOrder = WithAdditionalData(z.object({}))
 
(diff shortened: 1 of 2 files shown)


riqwan (approved):

sradevski (approved):

packages/medusa/src/api/admin/orders/validators.ts
```diff
@@ -52,11 +52,7 @@ export const AdminArchiveOrder = z.object({
 })
 export type AdminArchiveOrderType = z.infer<typeof AdminArchiveOrder>
 
-export type AdminCompleteOrderType = z.infer<typeof CompleteOrder>
-export const CompleteOrder = z.object({
-  order_id: z.string(),
-})
-export const AdminCompleteOrder = WithAdditionalData(CompleteOrder)
+export const AdminCompleteOrder = WithAdditionalData(z.object({}))
```
sradevski, line 55: Just FYI, this is a breaking change as now the validation will fail if you pass `order_id` if I am not mistaken

shahednasser: True, @olivermrbl how should we label this as a breaking change?

riqwan: we don't use this anywhere yet and doubt any existing users will as well

olivermrbl: Yeah, should be OK. I'll make a minor note of it in the release notes 👍

PR #8962, opened 2024-09-03
feat: Reset password

olivermrbl:
**Generate reset password token:**
- `POST /auth/user/emailpass/reset-password { identifier: "oli@oli.com" }`
- Execute `generateResetPasswordToken` workflow
  - Call `generateToken`
  - Emit event with token
- Respond with 201 CREATED

**Reset password:**
- `POST /auth/user/emailpass/update?token="..." { email: "oli@oli.com", password: "***" }`
- Validate token in middleware
- Call `authModule.updateProvider`
- Call `authProvider.update`
  - Hash new password
  - Update provider identity with new hashed password
- Respond with 200


integration-tests/http/__tests__/auth/admin/auth.spec.ts
```diff
@@ -8,120 +9,208 @@ jest.setTimeout(30000)
...
-    it("Invite + registration + authentication flow", async () => {
-      // Create invite
-      const { token: inviteToken } = (
-        await api.post(
-          "/admin/invites",
-          { email: "newadmin@medusa.js" },
-          adminHeaders
-        )
-      ).data.invite
+    describe("Full authentication lifecycle", () => {
```
olivermrbl, line 22: These are existing tests

integration-tests/http/__tests__/auth/admin/auth.spec.ts
```diff
@@ -8,120 +9,208 @@ jest.setTimeout(30000)
...
-        .catch((e) => e)
-
-      expect(unAuthedRequest.response.status).toEqual(401)
     })
 
-    it("should respond with 401 on register, if email already exists", async () => {
-      const signup = await api
-        .post("/auth/user/emailpass/register", {
-          email: "admin@medusa.js",
+    describe("Reset password flows", () => {
```
olivermrbl, line 138: These are new tests

packages/core/core-flows/src/auth/steps/reset-password.ts
```diff
@@ -0,0 +1,47 @@
...
+export const resetPasswordStepId = "reset-password-step"
+export const resetPasswordStep = createStep(
+  resetPasswordStepId,
+  async (
+    input: {
+      provider: string
+      entityId: string
+      password: string
+      providerIdentity: ProviderIdentityDTO
+    },
```
olivermrbl: todo: create type

packages/core/core-flows/src/auth/steps/validate-reset-password-token.ts
```diff
@@ -0,0 +1,32 @@
...
+}
+
+export const validateResetPasswordTokenStep = createStep(
+  "validate-reset-password-token",
+  async (input: {
+    token: string
+    entityId: string
+    providerIdentityId: string
+    passwordHash: string
+  }) => {
```
olivermrbl: todo: create type

packages/core/core-flows/src/auth/workflows/reset-password.ts
```diff
@@ -0,0 +1,64 @@
...
+import {
+  createWorkflow,
+  transform,
+  WorkflowResponse,
+} from "@medusajs/workflows-sdk"
+import { useRemoteQueryStep } from "../../common"
+import { resetPasswordStep } from "../steps/reset-password"
+import { validateResetPasswordTokenStep } from "../steps/validate-reset-password-token"
+
+type WorkflowInput = {
```
olivermrbl: suggestion: move to types

packages/core/utils/src/auth/abstract-auth-provider.ts
```diff
@@ -238,6 +240,24 @@ export abstract class AbstractAuthModuleProvider implements IAuthProvider {
     )
   }
 
+  resetPassword(
+    data: ResetPasswordInput,
+    authIdentityProviderService: AuthIdentityProviderService
+  ): Promise<AuthIdentityDTO> {
+    throw new Error(
+      `Method 'recover' not implemented for provider ${this.provider}`
```
olivermrbl: todo: update message

packages/core/utils/src/core-flows/events.ts
```diff
@@ -10,6 +10,17 @@ export const OrderWorkflowEvents = {
   COMPLETED: "order.completed",
 }
 
+export const UserWorkflowEvents = {
+  CREATED: "user.created",
+  UPDATED: "user.updated",
+  DELETED: "user.deleted",
+  PASSWORD_RESET: "user.password_reset",
```
olivermrbl: todo: remove

packages/medusa/src/api/auth/[actor_type]/[auth_provider]/reset-password/route.ts
```diff
@@ -0,0 +1,31 @@
...
+  const { result } = await resetPasswordWorkflow(req.scope).run({
+    input: {
+      provider: auth_provider,
+      token: req.validatedBody.token,
+      password: req.validatedBody.password,
+      entityId: req.validatedBody.entity_id,
+    },
+  })
+
+  console.log("RESULT: ", result)
```
Comment (olivermrbl) [inline packages/medusa/src/api/auth/[actor_type]/[auth_provider]/reset-password/route.ts]: todo: remove

packages/medusa/src/api/auth/[actor_type]/[auth_provider]/reset-password/route.ts
```diff
@@ -0,0 +1,31 @@
...
+import { z } from "zod"
+import { MedusaRequest, MedusaResponse } from "../../../../../types/routing"
+
+export const AdminResetPasswordReq = z.object({
+  token: z.string(),
+  password: z.string(),
+  entity_id: z.string(),
+})
+
+export type AdminResetPasswordReqType = z.infer<typeof AdminResetPasswordReq>
```
Comment (olivermrbl) [inline packages/medusa/src/api/auth/[actor_type]/[auth_provider]/reset-password/route.ts]: todo: move to validators

packages/medusa/src/api/utils/unless-path.ts
```diff
@@ -1,15 +1,16 @@
+import { MiddlewareFunction } from "@medusajs/framework"
 import { NextFunction } from "express"
 import { MedusaRequest, MedusaResponse } from "../../types/routing"
-import { MiddlewareFunction } from "@medusajs/framework"
 
 // Due to how our route loader works, where we load all middlewares before routes, ambiguous routes end up having all middlewares on different routes executed before the route handler is.
 // This function allows us to skip middlewares for particular routes, so we can temporarily solve this without completely breaking the route loader for everyone.
 export const unlessPath =
-  (onPath: RegExp, middleware: MiddlewareFunction) =>
+  (onPath: RegExp, middleware: MiddlewareFunction | MiddlewareFunction[]) =>
```
olivermrbl: todo: revert

packages/modules/providers/auth-emailpass/src/services/emailpass.ts
```diff
@@ -19,6 +21,8 @@ type InjectedDependencies = {
 
 interface LocalServiceConfig extends EmailPassAuthProviderOptions {}
 
+// 15 minutes in milliseconds
```
olivermrbl: todo: correct comment

packages/core/core-flows/src/auth/steps/validate-reset-password-token.ts
```diff
@@ -0,0 +1,32 @@
...
+    const isExpired = decoded.exp < Math.floor(Date.now() / 1000)
+
+    const verified = jwt.verify(input.token, input.passwordHash) as TokenPayload
+    const isInvalid =
+      verified.entity_id !== input.entityId ||
+      verified.provider_identity_id !== input.providerIdentityId
+
+    if (!verified || isExpired || isInvalid) {
+      throw new MedusaError(MedusaError.Types.INVALID_DATA, `Invalid token`)
+    }
```
olivermrbl: suggestion: This validation logic could live in the auth providers as well

integration-tests/http/__tests__/auth/admin/auth.spec.ts
```diff
@@ -8,120 +9,208 @@ jest.setTimeout(30000)
...
+          "test@medusa-commerce.com"
+        )
 
-    it("should respond with 401 on sign in, if email does not exist", async () => {
-      const signup = await api
-        .post("/auth/user/emailpass", {
-          email: "john@doe.com",
-          password: "secret_password",
+        const response = await api.post("/auth/user/emailpass/reset-password", {
+          entity_id: "test@medusa-commerce.com",
```
riqwan: sugg: This can be potentially eliminated as we can pick this up from the token

or if this is necessary, a test to ensure that given a specific token, other user's account can't be updated

packages/core/types/src/auth/provider.ts
```diff
@@ -34,4 +41,12 @@ export interface IAuthProvider {
     data: AuthenticationInput,
     authIdentityProviderService: AuthIdentityProviderService
   ): Promise<AuthenticationResponse>
+  resetPassword(
```
riqwan: q: Aside from emailpass, won't all other auth use oauth? In that case, does this make sense for this to exist as a provider signature?

edit: From the method, it looks like its just doing a update on the identity.

packages/modules/providers/auth-emailpass/src/services/emailpass.ts
```diff
@@ -35,14 +39,97 @@ export class EmailPassAuthService extends AbstractAuthModuleProvider {
...
   }
 
-  protected async createAuthIdentity({ email, password, authIdentityService }) {
+  protected async hashPassword(password: string) {
     const hashConfig = this.config_.hashConfig ?? { logN: 15, r: 8, p: 1 }
     const passwordHash = await Scrypt.kdf(password, hashConfig)
+    return passwordHash.toString("base64")
+  }
+
+  async generateResetPasswordToken(
```
riqwan: q: should this be a generic temporary auth token generator? That way we can also reuse this for things like "magic link login"

riqwan (commented):
looks good, added some questions/suggestions

packages/core/types/src/auth/provider.ts
olivermrbl: There are other password providers, like phone-pass. However, you correctly point out that this method only does an update, so will be removed from the interface.

packages/modules/providers/auth-emailpass/src/services/emailpass.ts
olivermrbl: Yup, currently working on this – @sradevski also pointed this out in Slack.

riqwan: nice!

packages/core/core-flows/src/auth/steps/set-auth-app-metadata.ts
```diff
@@ -58,7 +58,7 @@ export const setAuthAppMetadataStep = createStep(
       delete appMetadata[key]
     }
 
-    await service.updateAuthIdentites({
+    await service.updateAuthIdentities({
```
sradevski, line 61: Wow that misspell went for a long time without anyone noticing 😄

packages/medusa/src/api/auth/middlewares.ts
```diff
@@ -15,21 +17,31 @@ export const authRoutesMiddlewares: MiddlewareRoute[] = [
   {
     method: ["POST"],
     matcher: "/auth/:actor_type/:auth_provider/callback",
-    middlewares: [],
+    middlewares: [validateScopeProviderAssociation()],
```
sradevski, line 20: Nice!

packages/core/core-flows/src/auth/workflows/generate-reset-password-token.ts
```diff
@@ -0,0 +1,62 @@
...
+          )
+        }
+
+        const token = generateJwtToken(
+          {
+            entity_id: input.entityId,
+            provider: input.provider,
+          },
+          {
+            // Ensures the token can only be used once per requested reset
```
sradevski: I think this opens up another set of attacks, which might be more concerning than someone being able to use the same token twice.

This basically makes the password hashing pointless. Imagine this scenario - You leak DB data on auth identities. Now, an attacker can now generate a JWT for any user, and reset their password, without having access to their email, nor knowing what their password is. This is equivalent to having the password in plaintext.

I would suggest we sign the token with the JWT secret - the reasons are:
1. Leaking DB data doesn't expose how your tokens are signed
2. In the case of leaking the JWT secret, rotating it will invalidate every single token.

I believe this results in a more secure setup.


Now, we can support token invalidation but I think for that it's better if we store that information in storage (eg. Redis with expiry)

sradevski (commented):
Very nice, just one security concern, otherwise LGTM

packages/core/core-flows/src/auth/workflows/generate-reset-password-token.ts
olivermrbl: > Now, an attacker can now generate a JWT for any user, and reset their password, without having access to their email, nor knowing what their password is. This is equivalent to having the password in plaintext.

Not sure I agree, or I might not be following. The token would not be available in the Rest API response, so the only way to obtain the token would be through whatever subscription mechanism is setup, e.g. emails.

**Update:** Yes, if the attacker has access to, in this case, emails then this would be equivalent to sending it in plaintext.

If we are concerned about using the password hash, we could just use a substring of it, e.g. the first 6 characters.

olivermrbl: Having said that, I am fine using the JWT secret. This was my initial approach (if you look at previous commits), and is a very easy change to the PR. I am not expert on the subject matter, so if your experience tells you this is a better approach, we should do that.

sradevski: We discussed this in Slack a bit more, and decided to go with the JWT secret for now. We need to tackle invalidation for both invite links and standard tokens as a separate work.

sradevski (approved):
Really nice one, I was someone asking if we'll support resets just today, so timely as well 😄

sradevski:
@olivermrbl I'll merge this one so I can do the rebase conflicts

PR #8964, opened 2024-09-03
fix(medusa): change the request type of removing products from price list

shahednasser:
Remove the `add` parameter in request body of `/price-lists/:id/products` to avoid confusion


olivermrbl (approved):
Nice

PR #8966, opened 2024-09-03
fix(medusa): fix update promotion's request type

shahednasser:
Fix the validator of the update promotion API route to match the update workflow input's type.

### DIFF
--- a/integration-tests/http/__tests__/promotions/admin/promotions.spec.ts
+++ b/integration-tests/http/__tests__/promotions/admin/promotions.spec.ts
@@ -495,24 +495,6 @@ medusaIntegrationTestRunner({
           )
         })
 
-        it("should throw an error when both campaign and campaign_id params are passed", async () => {
-          const { response } = await api
-            .post(
-              `/admin/promotions/${promotion.id}`,
-              {
-                campaign: { name: "test campaign" },
-                campaign_id: "test",
-              },
-              adminHeaders
-            )
-            .catch((e) => e)
-
-          expect(response.status).toEqual(400)
-          expect(response.data.message).toContain(
-            `Invalid request: Field 'campaign, campaign_identifier' is required`
-          )
-        })
-
         it("should update a promotion successfully", async () => {
           const response = await api.post(
             `/admin/promotions/${promotion.id}`,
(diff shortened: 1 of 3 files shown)


olivermrbl (approved):

packages/medusa/src/api/admin/promotions/validators.ts
```diff
@@ -125,15 +125,14 @@ export type AdminUpdateApplicationMethodType = z.infer<
 >
 export const AdminUpdateApplicationMethod = z
   .object({
+    id: z.string(),
```
riqwan: todo: this will break the admin promotion update in dashboard as we don't pass the id in there and this change will require it now

riqwan (approved):

packages/medusa/src/api/admin/promotions/validators.ts
shahednasser: Should I address it in this PR?

integration-tests/http/__tests__/promotions/admin/promotions.spec.ts
```diff
@@ -495,24 +495,6 @@ medusaIntegrationTestRunner({
           )
         })
 
-        it("should throw an error when both campaign and campaign_id params are passed", async () => {
```
shahednasser, line 498: @riqwan found this test and removed it as it doesn't apply anymore, but is it possible we need the `campaign`?

riqwan: We don't need it anymore. We built the APIs before the UI was ready and basically tried to handle all cases. This case turned out not to be used.

packages/medusa/src/api/admin/promotions/validators.ts
riqwan: I think so, because it will break promotion update functionalities in admin.

An easier way would be to remove this and [update this instead](https://github.com/medusajs/medusa/blob/develop/packages/core/types/src/promotion/common/promotion.ts#L141) to:

```
application_method?: Omit<UpdateApplicationMethodDTO, "id">
```

shahednasser: But would this update the application method without its ID?

riqwan: yes, its a 1:1 relationship. The module picks it up from the promotion id that's already passed to its parent. This is only used as a subresource an never referenced directly

shahednasser: cool, updated it

PR #8968, opened 2024-09-03
fix(order,utils): fix outstanding amount stuck on long orders

riqwan:
what:

When update requests cross the default limit, it doesn't update any records beyond it. This PR fixes the issue generally on the abstract internal service and additionally for the order module when sub-resources cross the default limit. 

RESOLVES CC-367

### DIFF
--- a/packages/core/utils/src/modules-sdk/medusa-internal-service.ts
+++ b/packages/core/utils/src/modules-sdk/medusa-internal-service.ts
@@ -2,9 +2,9 @@ import {
   BaseFilterable,
   Context,
   FilterQuery,
-  FilterQuery as InternalFilterQuery,
   FindConfig,
   InferEntityType,
+  FilterQuery as InternalFilterQuery,
   ModulesSdkTypes,
   PerformedActions,
   UpsertWithReplaceConfig,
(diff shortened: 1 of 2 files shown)


packages/core/utils/src/modules-sdk/medusa-internal-service.ts
```diff
@@ -304,7 +304,7 @@ export function MedusaInternalService<
       if (keySelectorForDataOnly.$or.length) {
         const entitiesToUpdate = await this.list(
           keySelectorForDataOnly,
-          {},
+          { take: null },
```
riqwan, line 307: cc. @adrien2p just so you are aware

carlos-r-l-rodrigues (approved):

packages/core/utils/src/modules-sdk/medusa-internal-service.ts
srindom: Can't we pass the selector to mikro orm instead of listing first?

adrien2p: I thought we fixed that long time ago 😅

PR #8972, opened 2024-09-03
feat(dashboard): update create fulfillment UI part 1

fPolic:
**What**
- update Create fulfillment modal according to the design

**Note**
- in a followup I will add support for inventory kits

**Question**
- should we support overriding shipping method as per design?

---

Before:
Screenshot 2024-09-03 at 17 57 18

After:

Screenshot 2024-09-03 at 17 51 26


olivermrbl:
> should we support overriding shipping method as per design?

Don't think this is important near term 👍

olivermrbl (approved):
Looks and works nicely

PR #8973, opened 2024-09-03
fix(dashboard): Tab behaviour in DataGrid

kasperkristensen:
**What**
- Fixes "Tab" behaviour in DataGrid component.
- Removes FocusTrap dependency, and prop-types dependency as it is no longer needed now that we have removed FocusTrap.
- Adds hotkeys for moving the focus outside of the grid. `Ctrl + Alt + ,` to move focus to before the grid, and `Ctrl + Alt + .` to move it to after the grid.

Resolves CC-414, CC-397


olivermrbl (commented):
Tab behavior works nicely, but I couldn't get the other new keys to work with the instructions provided in the PR description 🤔

olivermrbl (commented):
Anyway, I don't think that should block the PR necessarily

kasperkristensen:
@olivermrbl should be good now

olivermrbl (approved):
LGTM

PR #8974, opened 2024-09-03
fix(core-flows, types, medusa): fix batch delete types in workflows and routes

shahednasser:
- Fix the returned data of batch delete steps / workflows for deleted records to just be the array of IDs.
- Add a new type `BatchResponse` for API routes to have a different shape for deleted records
- Update batch delete helpers to return the expected shape for API routes

### DIFF
--- a/packages/core/core-flows/src/fulfillment/steps/delete-shipping-option-rules.ts
+++ b/packages/core/core-flows/src/fulfillment/steps/delete-shipping-option-rules.ts
@@ -34,7 +34,7 @@ export const deleteShippingOptionRulesStep = createStep(
     await fulfillmentModule.deleteShippingOptionRules(ids)
 
     return new StepResponse({
-      ids: ids ?? [],
+      ids,
       object: "shipping_option_rule",
       deleted: true,
     }, shippingOptionRules)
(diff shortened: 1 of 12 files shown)


packages/core/core-flows/src/fulfillment/steps/delete-shipping-option-rules.ts
```diff
@@ -33,7 +33,11 @@ export const deleteShippingOptionRulesStep = createStep(
 
     await fulfillmentModule.deleteShippingOptionRules(ids)
 
-    return new StepResponse(null, shippingOptionRules)
+    return new StepResponse({
+      ids,
+      object: "shipping_option_rule",
+      deleted: true,
```
olivermrbl: thought: think this should just be the IDs ... this shape is very specific to how we want to respond in the HTTP layer

shahednasser: I took this shape from other workflows, and it's also the expected response type for deleting the shipping option rule and multiple similar API routes.

I'll update the return types then of these workflows / API routes

olivermrbl: The return value should have this shape in the API Routes, but not in the workflows

shahednasser: Yes I meant to update the API routes for the changes in the workflows

olivermrbl (approved):
LGTM

Issue #8976, opened 2024-09-03
Scheduled jobs print an empty "error:" after every successful execution


matteovava:
### Describe the bug

I'm following the v2 guide about settings [scheduled jobs](https://docs.medusajs.com/v2/basics/scheduled-jobs) and i've noticed that after setting a scheduled job the app print an empty "error:" in the console every time the job runs.
I've tried a basic debug and the message seems to be printed from the file [workflow-export.ts in the function attachOnFinishReleaseEvents at line 530](https://github.com/medusajs/medusa/blob/58c78a7f62aee6082f9620d670ae4f2b426178eb/packages/core/workflows-sdk/src/helper/workflow-export.ts#L530). Seems that is missing a check on allWorkflowErrors var in order to skip when there are no error.

### System information

Medusa version (including plugins): preview (v2.0.7)
Node.js version: 20.17.0
Database: PostgreSQL 16
Operating system: MacOS Sonoma
Browser (if relevant):

### Steps to reproduce the behavior

1. Create a new project
2. Create a scheduled jobs like the one suggested [here](https://docs.medusajs.com/v2/basics/scheduled-jobs)
3. After 1 minute it will print the message followed by a "error:"

### Screenshots

Error message:
![Screenshot 2024-09-03 at 22 38 11](https://github.com/user-attachments/assets/7196c6fb-61f6-4f50-815d-efb4108ee2b8)

Variables values in workflow-export.ts
![Screenshot 2024-09-04 at 00 01 41](https://github.com/user-attachments/assets/69f9b24f-d574-49d3-8f02-bdf719cfb7f3)

PR #8978, opened 2024-09-03
fix: Workflow always print "error:" even if there are none

matteovava:
Fix issue [https://github.com/medusajs/medusa/issues/8976](https://github.com/medusajs/medusa/issues/8976)


packages/core/workflows-sdk/src/helper/workflow-export.ts
```diff
@@ -527,7 +527,7 @@ function attachOnFinishReleaseEvents(
         )
         .join(EOL + separator + EOL)
 
-      logger.error(allWorkflowErrors)
+      if (allWorkflowErrors) logger.error(allWorkflowErrors)
```
sradevski:
```suggestion
      if (allWorkflowErrors) {
        logger.error(allWorkflowErrors)
      }
```

sradevski (commented):
Nice, thanks!  Let's just reformat it to be more consistent (we need to check why our linting/prettier setup didn't catch this)

packages/core/workflows-sdk/src/helper/workflow-export.ts
olivermrbl: Should we maybe just commit this ourselves and get it merged?

sradevski: Let's merge it, I'll reformat in a separate PR

matteovava: Hi, sorry i've seen it only now and accepted the commit suggestion

carlos-r-l-rodrigues (approved):
Thanks @matteoxplo

PR #8980, opened 2024-09-04
feat: Add github authentication provider

sradevski:
There is few breaking changes in this PR, namely:

1. The Google options got renamed to be more consistent (`clientID` -> `clientId`, `callbackURL` -> `callbackUrl`)
2. The `successRedirectUrl` option was removed from google provider
3. The auth callback handling changed - It used to redirect to `successRedirectUrl` with the `access_token` as a query parameter, and the `callbackUrl` would point to the backend. The flow now would be to set `callbackUrl` to a FE path, and within that path call `sdk.auth.callback()` to login similarly to how emailpass login works.

### DIFF
--- a/packages/core/types/src/auth/providers/google.ts
+++ b/packages/core/types/src/auth/providers/google.ts
@@ -2,5 +2,4 @@ export interface GoogleAuthProviderOptions {
   clientId: string
   clientSecret: string
   callbackUrl: string
-  successRedirectUrl?: string
 }
--- a/packages/modules/auth/src/services/auth-module.ts
+++ b/packages/modules/auth/src/services/auth-module.ts
@@ -282,6 +282,74 @@ export default class AuthModuleService
           createdAuthIdentity
         )
       },
+      update: async (
+        entity_id: string,
+        data: {
+          provider_metadata?: Record<string, unknown>
+          user_metadata?: Record<string, unknown>
+        }
+      ) => {
+        const authIdentities = await this.authIdentityService_.list(
+          {
+            provider_identities: {
+              entity_id,
+              provider,
+            },
+          },
+          {
+            relations: ["provider_identities"],
+          }
+        )
+
+        if (!authIdentities.length) {
+          throw new MedusaError(
+            MedusaError.Types.NOT_FOUND,
+            `AuthIdentity with entity_id "${entity_id}" not found`
+          )
+        }
+
+        if (authIdentities.length > 1) {
+          throw new MedusaError(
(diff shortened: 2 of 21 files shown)


packages/core/types/src/auth/providers/google.ts
```diff
@@ -1,6 +1,6 @@
 export interface GoogleAuthProviderOptions {
-  clientID: string
+  clientId: string
```
sradevski, line 2: This is a breaking change but I thought it might be better to do it now than later. I don't think many people have used google auth anyway.

packages/modules/providers/auth-github/src/services/github.ts
```diff
@@ -0,0 +1,204 @@
...
+    providerMetadata: {
+      access_token: string
+      refresh_token: string
+      access_token_expires_at: string
+      refresh_token_expires_at: string
+    },
+    authIdentityService: AuthIdentityProviderService
+  ) {
+    if (!providerMetadata?.access_token) {
+      return { success: false, error: "No ID found" }
```
olivermrbl:
```suggestion
      return { success: false, error: "Missing access token" }
```

packages/modules/auth/src/services/auth-module.ts
```diff
@@ -282,6 +282,74 @@ export default class AuthModuleService
           createdAuthIdentity
         )
       },
+      update: async (
```
olivermrbl, line 285: thought(non-blocking): Would it be more appropriate to call this `updateProviderIdentity`?

The other two methods, `retrieve` and `create`, works on `AuthIdentities`, while as this one works on `ProviderIdentities`.

olivermrbl (approved):
This looks good to me, nice work.

packages/modules/auth/src/services/auth-module.ts
sradevski: The `create` also operates on a provider identity, but I think I can rename all of them to be more concrete.

sradevski: Ah well I guess we pass provider identity data, but it creates an auth identity. Let's keep the methods as-is, since it looks a bit strange to rename `update` where the input is almost the same as the `create`, WDYT?

olivermrbl: > Let's keep retrieve and create as-is and only rename the provider identity one so it's not another breaking change.

Yeah, this is what I had in mind 👍

PR #8988, opened 2024-09-04
feat(admin-sdk,admin-bundler,admin-shared,medusa): Restructure admin packages

kasperkristensen:
**What**
- Renames /admin-next -> /admin
- Renames @medusajs/admin-sdk -> @medusajs/admin-bundler
- Creates a new package called @medusajs/admin-sdk that will hold all tooling relevant to creating admin extensions. This is currently `defineRouteConfig` and `defineWidgetConfig`, but will eventually also export methods for adding custom fields, register translation, etc. 
  - cc: @shahednasser we should update the examples in the docs so these functions are imported from `@medusajs/admin-sdk`. People will also need to install the package in their project, as it's no longer a transient dependency.
  - cc: @olivermrbl we might want to publish a changelog when this is merged, as it is a breaking change, and will require people to import the `defineXConfig` from the new package instead of `@medusajs/admin-shared`.
- Updates CODEOWNERS so /admin packages does not require a review from the UI team.


olivermrbl:
My bad, I had one PR with admin changes on automerge!

olivermrbl:
@shahednasser, I have paused preview releasing for now, so we can get this PR merged and the documentation updated. Do you think you can prioritise this tomorrow morning?

shahednasser:
Sorry missed this one, yes should be an easy fix, will open a PR first thing tomorrow.

> People will also need to install the package in their project, as it's no longer a transient dependency.

We should install it in the starter as well. Will take care of that once the change is published.

olivermrbl (approved):
LGTM

PR #8991, opened 2024-09-04
feat(types,medusa): add acknowledgement typing

shahednasser:
- Add acknowledgement typing based on the returned type by the redis and in-memory workflow engines
- Add a new http type for the `/admin/workflows-executions/:workflow_id/run` API route, which returns the acknowledgement.

### DIFF
--- a/packages/core/types/src/workflows-sdk/service.ts
+++ b/packages/core/types/src/workflows-sdk/service.ts
@@ -15,6 +15,11 @@ type FlowRunOptions<TData = unknown> = {
   events?: Record<string, Function>
 }
 
+export type Acknowledgement = {
+  workflowId: string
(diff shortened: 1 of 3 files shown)


packages/core/types/src/workflows-sdk/service.ts
```diff
@@ -15,6 +15,11 @@ type FlowRunOptions<TData = unknown> = {
   events?: Record<string, Function>
 }
 
+export type Acknowledgement = {
+  workflowId: string
+  transactionId: string
```
olivermrbl, line 20: ideally, for consistency, these would be snake cased in the HTTP response

olivermrbl (approved):

PR #8994, opened 2024-09-04
fix(types,medusa): fix query types for some store routes

shahednasser:
- Match address filters to validator
- Move query type from `/store/orders/:id` to `/store/orders`.
- Fix `/store/payment-providers` filters to only allow region_id (since that's the only filter used in the route).
- Pass query parameters of `/store/product-categories/:id` to the refetchEntity
- Match category filters between type and validator.
- Match the product's filter type to validator
- Remove region filters that don't match the validator
- Rename the `code` filter to `currency_code` in region filter types / validators in store and admin routes


olivermrbl (approved):
Nice clean-up

PR #8995, opened 2024-09-04
feat: Application types generation from project GQL schema's

adrien2p:
RESOLVES FRMW-2685
RESOLVES FRMW-2686

This PR introduces a workflow for generating types for the modules an application is using within the `.medusa` directory in the root of the app.

The main purpose is to make the RemoteQuery (soon to be `query`) have type-information about the available endpoints, their fields and the return values.

The types are generated by converting the DML models of modules to GraphQL schema and then generating TypeScript types from them.


packages/core/utils/src/common/__tests__/deep-merge.ts
```diff
@@ -0,0 +1,46 @@
+import { deepMerge } from "../deep-merge"
```
carlos-r-l-rodrigues, line 1: Ended up not using this, but it might be a good addition to the utils

packages/cli/medusa-cli/src/create-cli.ts
```diff
@@ -6,12 +6,17 @@ import resolveCwd from "resolve-cwd"
...
 
 const yargs = require(`yargs`)
 
 const handlerP =
   (fn) =>
   (...args) => {
+    const directory = path.resolve(`.`)
+
+    const projectInfo = { directory }
+    const newArgs = { ...argv, ...projectInfo }
```
thetutlage: All this can be removed right?

packages/cli/medusa-cli/src/create-cli.ts
```diff
@@ -59,6 +64,15 @@ function buildLocalCommands(cli, isLocalProject) {
...
   cli
+    .command({
+      command: "types",
+      handler: handlerP(
+        getCommandHandler("gql-to-ts", (args, cmd) => {
+          process.env.NODE_ENV = process.env.NODE_ENV || `development`
+          return cmd(args)
+        })
+      ),
+    })
```
thetutlage: Should we remove this command for now and just rely on dev-server to generate the needed types.

Let's see if people have any custom use-cases for manual generation?

packages/modules/payment/src/schema/index.ts
```diff
@@ -0,0 +1,111 @@
...
+enum PaymentCollectionStatus {
+  not_paid
+  awaiting
+  authorized
+  partially_authorized
+  canceled
+}
+
+enum PaymentSessionStatus {
+  authorized
```
thetutlage, line 11: Should we remove these manually defined GQL schema?

thetutlage (changes requested):
Looks pretty neat for the most part, just a few things

packages/cli/medusa-cli/src/create-cli.ts
adrien2p: Yes 👌

packages/cli/medusa-cli/src/create-cli.ts
adrien2p: Yes i was thinking of that too, lets do that

packages/modules/payment/src/schema/index.ts
adrien2p: It was to have a more concrete test, though, we can keep them for now and rm them when they are migrated to dml

thetutlage: But that means we will have to maintain them with the changes done inside the entities as long as we haven't fully migrated

adrien2p: Yes thats true. I let you decide 👍

thetutlage: Cool. Let's wait and see what @carlos-r-l-rodrigues thinks

carlos-r-l-rodrigues: I believe we should keep them to have more data to work with and test things in a more realistic scenario.
But I agree, if schemas are changed we have one more place to update them.

carlos-r-l-rodrigues (approved):

packages/modules/payment/src/schema/index.ts
thetutlage: Cool. Let's keep them for now

thetutlage (approved):

PR #9004, opened 2024-09-04
feat(product): product option value methods

carlos-r-l-rodrigues:


packages/core/types/src/product/service.ts
```diff
@@ -1366,6 +1370,270 @@ export interface IProductModuleService extends IModuleService {
...
+   * This method is used to create a product option.
+   *
+   * @param {CreateProductOptionValueDTO} data - The product option to be created.
+   * @param {Context} sharedContext - A context used to share resources, such as transaction manager, between the application and the module.
+   * @returns {Promise<ProductOptionValueDTO>} The created product option.
+   *
+   * @example
+   * const option = await productModuleService.createProductOptionValues({
+   *   value: "Blue",
+   *   product_id: "prod_123",
```
sradevski: This should be `option_id` right?

sradevski (approved):

packages/core/types/src/product/service.ts
carlos-r-l-rodrigues: yes, it is. thanks.

PR #9009, opened 2024-09-05
chore: fix incorrect oas property name

shahednasser:
Fix incorrect oas property name breaking the build


shahednasser:

olivermrbl (approved):

PR #9012, opened 2024-09-05
fix: wrap ModuleImplementations so that all methods are async

Alexnortung:
As discussed on slack: modules are wrapped in a proxy which makes all of its methods async. Currently if you try to do declaration merging on th `ModuleImplementations` type, then you can end up getting the wrong types, since non-async methods will not be converted to async on a type level.

This PR fixes that by wrapping `ModuleImplementations` in a type that makes all non-async methods async.

CLOSES: FRMW-2687

### DIFF
--- a/packages/framework/framework/src/container.ts
+++ b/packages/framework/framework/src/container.ts
@@ -1,5 +1,6 @@
 import { createMedusaContainer } from "@medusajs/utils"
 import { AwilixContainer, ResolveOptions } from "awilix"
+import { TransformObjectMethodToAsync } from "@medusajs/types";
 
 /**
  * The following interface acts as a bucket that other modules or the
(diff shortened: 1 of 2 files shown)


packages/framework/framework/src/container.ts
```diff
@@ -7,11 +7,17 @@ import { AwilixContainer, ResolveOptions } from "awilix"
  */
 export interface ModuleImplementations {}
 
+type ToAsyncMethods<T extends object> = {
+  [K in keyof T]: T[K] extends (...args: infer A) => infer R
+    ? T[K] extends (...args: A) => Promise<infer PR> ? T[K] : (...args: A) => Promise<R>
+    : T[K]
+}
+
```
adrien2p:
```suggestion
type TransformObjectMethodToAsync<T extends object> = {
  [K in keyof T]: T[K] extends (...args: infer A) => infer R
    ? (...args: A) => Promise<Awaited<R>>
    : T[K] extends object ? TransformObjectMethodToAsync<T[K]> : T[K];
}
```

Very nice, I would just simplify it a bit and also take into consideration nested objects, here you want to change the method of the interfaces of ModuleImplementations for which the props are interfaces

adrien2p: In regards to its location, I would eventually move it to core/types/src/common and maybe name it differently like the proposal above? cc @olivermrbl

olivermrbl: @adrien2p, think that makes sense. Should we get this merged?

adrien2p: It is good to me 👌

adrien2p (approved):
LGTM, thank you for your contribution 🥂

olivermrbl:
Only unrelated tests failing due to authentication issues, so merging this now.

PR #9013, opened 2024-09-05
feat: Add support for refreshing JWT tokens

sradevski:
This is useful in few scenarios, namely:
1. When doing oauth, after creating a user, we want to immediately refresh the JWT token so it has the actor ID populated
2. You don't want people to have to re-login every day, so you can implement a token refresh eg. every time your site is visited, as long as the `expiresIn` hasn't passed.

CLOSES CC-425

### DIFF
--- a/packages/core/utils/src/auth/token.ts
+++ b/packages/core/utils/src/auth/token.ts
@@ -1,12 +1,20 @@
 import jwt from "jsonwebtoken"
+import { MedusaError } from "../common"
 
 export const generateJwtToken = (
   tokenPayload: Record<string, unknown>,
   jwtConfig: {
-    secret: string
-    expiresIn: string
+    secret: string | undefined
+    expiresIn: string | undefined
   }
 ) => {
+  if (!jwtConfig.secret || !jwtConfig.expiresIn) {
+    throw new MedusaError(
+      MedusaError.Types.INVALID_ARGUMENT,
+      "JWT secret and expiresIn must be provided when generating a token"
+    )
+  }
+
   return jwt.sign(tokenPayload, jwtConfig.secret, {
     expiresIn: jwtConfig.expiresIn,
   })
--- a/packages/medusa/src/api/auth/token/refresh/route.ts
+++ b/packages/medusa/src/api/auth/token/refresh/route.ts
@@ -1,0 +1,40 @@
+import { IAuthModuleService } from "@medusajs/types"
+import {
+  AuthenticatedMedusaRequest,
+  MedusaResponse,
+} from "../../../../types/routing"
+import {
+  ContainerRegistrationKeys,
+  ModuleRegistrationName,
+} from "@medusajs/utils"
+import { generateJwtTokenForAuthIdentity } from "../../utils/generate-jwt-token"
+
+// Retrieve a newly generated JWT token. All checks that the existing token is valid already happen in the auth middleware.
+// The token will include the actor ID, even if the token used to refresh didn't have one.
+// Note: We probably want to disallow refreshes if the password changes, and require reauth.
+export const POST = async (
+  req: AuthenticatedMedusaRequest,
+  res: MedusaResponse
+) => {
+  const service: IAuthModuleService = req.scope.resolve(
+    ModuleRegistrationName.AUTH
+  )
+
+  const authIdentity = await service.retrieveAuthIdentity(
+    req.auth_context.auth_identity_id
+  )
+
+  const { http } = req.scope.resolve(
+    ContainerRegistrationKeys.CONFIG_MODULE
+  ).projectConfig
+
+  const token = generateJwtTokenForAuthIdentity(
+    { authIdentity, actorType: req.auth_context.actor_type },
+    {
+      secret: http.jwtSecret,
+      expiresIn: http.jwtExpiresIn,
+    }
+  )
+
+  return res.json({ token })
+}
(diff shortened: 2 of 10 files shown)


sradevski:
@olivermrbl I think this will cause some merge conflicts with your reset password PR, apologies for that :)

riqwan:
q: if a token is refreshed 3 times, will all 3 tokens be valid if they're not expired yet?

shahednasser:
@sradevski does this add a last step in the third-party authentication flow to send a request to `/auth/token/refresh`?

packages/medusa/src/api/auth/token/refresh/route.ts
```diff
@@ -0,0 +1,40 @@
...
+} from "../../../../types/routing"
+import {
+  ContainerRegistrationKeys,
+  ModuleRegistrationName,
+} from "@medusajs/utils"
+import { generateJwtTokenForAuthIdentity } from "../../utils/generate-jwt-token"
+
+// Retrieve a newly generated JWT token. All checks that the existing token is valid already happen in the auth middleware.
+// The token will include the actor ID, even if the token used to refresh didn't have one.
+// Note: We probably want to disallow refreshes if the password changes, and require reauth.
```
riqwan, line 14: I'll put this scenario in here, will leave it upto you to decide if its worth to tackle or not:

If a JWT token is compromised, anyone with that JWT can essentially create unlimited tokens here that will be valid using this endpoint here. We don't currently seem to have an invalidation strategy in place to prevent this.

packages/medusa/src/api/auth/utils/generate-jwt-token.ts
```diff
@@ -1,8 +1,15 @@
...
-  { authIdentity, actorType },
-  { secret, expiresIn }
+  {
+    authIdentity,
+    actorType,
+  }: { authIdentity: AuthIdentityDTO; actorType: string },
+  {
+    secret,
+    expiresIn,
+  }: { secret: string | undefined; expiresIn: string | undefined }
```
riqwan, line 12:
```suggestion
  }: { secret?: string; expiresIn?: string }
```

packages/core/utils/src/auth/token.ts
```diff
@@ -1,12 +1,20 @@
 import jwt from "jsonwebtoken"
+import { MedusaError } from "../common"
 
 export const generateJwtToken = (
   tokenPayload: Record<string, unknown>,
   jwtConfig: {
-    secret: string
-    expiresIn: string
+    secret: string | undefined
+    expiresIn: string | undefined
```
riqwan, line 8:
```suggestion
    secret?: string
    expiresIn?: string
```

riqwan (commented):
Added one question in there for a scenario

packages/medusa/src/api/auth/utils/generate-jwt-token.ts
sradevski: They are not optional, but they can be undefined, so I think semantically it's better to use `undefined`

packages/core/utils/src/auth/token.ts
sradevski: Same as above

packages/medusa/src/api/auth/token/refresh/route.ts
sradevski: Yeah that's the nature of JWT tokens, and is the reason why they should be kept relatively short-lived. Although it might seem dangerous, losing JWT tokens is not something that happens that easily, so the practical risks are not as high as people make them sound.

When it comes to creating unlimited tokens, you can probably do enough damage with 1 as you can do with unlimited tokens, so it doesn't make the severity any higher.

Now, there are few strategies to deal with how you use JWT tokens, such as issuing longer-lived refresh tokens, or supporting invalidation, but I don't think it is something we have to focus on anytime soon. The only thing we might want to support is have a flag on the auth_identity to not allow for refreshes (eg. when you change your password), but even that is something that is for later on.

You could always use session storage if you cannot accept that risk, this is just for the cases where you want to use JWTs.

sradevski:
> @sradevski does this add a last step in the third-party authentication flow to send a request to `/auth/token/refresh`?

Yes @shahednasser , the code will look like this:

```
const token = await sdk.auth.callback(code..)
const user = await sdk.user.create(...., {token})
await sdk.auth.refresh()
```

(obviously it's pseudocode-ish)

packages/medusa/src/api/auth/token/refresh/route.ts
riqwan: > Although it might seem dangerous, losing JWT tokens is not something that happens that easily, so the practical risks are not as high as people make them sound.

This can be said about any security key 😄 

> When it comes to creating unlimited tokens, you can probably do enough damage with 1 as you can do with unlimited tokens, so it doesn't make the severity any higher.

Not exactly, you can blacklist one using a simple list where a human can simply add it. When you have a system that can generate unlimited tokens, its harder to blacklist with a simple manual intervention.

Either way, will resolve this as I just wanted to surface this.

sradevski: Yes indeed, security keys that are not handled manually (eg. API keys) have lower chance of being stolen. Not to say it doesn't happen, but it's all about tradeoffs. 

TBH, If you have a simple server, using sessions is probably easier to handle, so you can go with that. Once you get to a more distributed setup, session-less handling can be better (although if you add token invalidation, you get closer to session-based auth), so it's just about having either option available and making an informed decision.

You can always rotate the JWT secret (which is a bit of a big-bang approach), so there are other mechanisms to "invalidate" tokens, but I get your point 😄

olivermrbl (approved):

olivermrbl:
Nice work, no immediate comments 👍

PR #9014, opened 2024-09-05
fix(utils,medusa,order,cart): fix totals when promotions are included

riqwan:
what:

- Fixes totals when adjustments are included
- Fixes original_* totals
- Fixes shipping_* totals
- Adds discount_subtotal
- Fixes totals when tax inclusive

RESOLVES CC-418
RESOLVES CC-417


integration-tests/modules/__tests__/cart/store/carts.spec.ts
```diff
@@ -2078,12 +2078,12 @@ medusaIntegrationTestRunner({
...
-                  tax_total: 5.364,
-                  total: 94.764,
+                  tax_total: 5.4,
+                  total: 95.4,
                   subtotal: 100,
                   original_total: 106,
-                  discount_total: 10.6,
-                  discount_tax_total: 0.636,
+                  discount_total: 10,
+                  discount_tax_total: 0,
```
srindom: Are we sure this shouldn't be:

```
discount_subtotal: 10
discount_tax_total: 0.6
discount_total: 10.6
```

srindom:
Regarding:

> when the item is tax inclusive and an adjustment is present in the item, we calculate tax of the item as tax_item_total = taxOn({ taxableAmount: (item.unit_price * item.qty) - item.discount })

for percentage promotions this is alright, but for fixed promotions we are going to have to either 
1. Enforce that fixed promotions are always tax exclusive (this would also mean having to always narrow fixed promos to only one country)
2. Add a tax inclusivity switch to the amount you enter when creating a promotion.

carlos-r-l-rodrigues (approved):

PR #9018, opened 2024-09-05
feat(product): Allow URL-Encoded special characters in product handle

v0eak:
Changed the regex of the validation function, which checks if the product handle is valid.
It now allows for URL encoded special characters (e.g. ★ => %E2%98%85)

Best practice is to use the `encodeURIComponent()` function to encode the handle that shall be created.

It allows for handles like this:
`some-handle-%E2%98%85`
`some-%E2%98%85-handle`
`some%E2%98%85handle`
`%E2%98%85-handle`
`%E2%98%85handle`


v0eak:
does not work yet, requires modifications to core medusa

PR #9023, opened 2024-09-05
chore(utils): update auth provider TSDocs + fix errors stopping references build

shahednasser:
- Updated the TSDocs of the auth provider
- Update the Node version in the docs-util as the outdated version was causing TypeScript error
- Add `@ts-ignore` for a line that was causing a typescript error. If there's a better fix here let me know


packages/core/utils/src/auth/abstract-auth-provider.ts
```diff
@@ -244,8 +293,13 @@ export abstract class AbstractAuthModuleProvider implements IAuthProvider {
    * In an authentication flow that requires performing an action with a third-party service, such as login
    * with a social account, the {@link authenticate} method is called first.
    *
-   * Then, the third-party service redirects to the Medusa application's validate callback API route.
-   * That route uses this method to authenticate the user.
+   * Then, the third-party service redirects to a frontend URL passing it a `code` query parameter.
+   * The frontend should then send a request to the Medusa application's validate callback API route, passing it the code.
+   * That route uses this method to verify the callback's code.
+   * 
+   * If the callback is verified successfully, the provider creates an auth identity for the user.
```
sradevski: To add: If there is already a user present, the user and provider metadata are updated.

sradevski (approved):

packages/core/utils/src/auth/abstract-auth-provider.ts
shahednasser: @sradevski do you mean that if an auth identity already exist, the provider should update the user and provider metadata? is there a specific reason for this (to add it to the doc)

sradevski: The reason is that the response carries access tokens and the user metadata might have been updated (eg. their first name or avatar), so we want to update them in the DB as well

olivermrbl (approved):

PR #9029, opened 2024-09-06
feat(create-medusa-app): add publishable API key environment variable to Next.js storefront

shahednasser:
When installing with Next.js starter, retrieve the publishable API key seeded and add it to the environment variable in the `.env.local` or `.env.template` file


srindom (approved):
lgtm

PR #9031, opened 2024-09-06
feature: introduce types for query.graph method

thetutlage:


packages/core/modules-sdk/src/medusa-app.ts
```diff
@@ -54,6 +54,7 @@ declare module "@medusajs/types" {
     [ContainerRegistrationKeys.CONFIG_MODULE]: ConfigModule
     [ContainerRegistrationKeys.PG_CONNECTION]: Knex<any>
     [ContainerRegistrationKeys.REMOTE_QUERY]: RemoteQueryFunction
```
thetutlage, line 56: Currently we put the `query` inside the container as `REMOTE_QUERY`. Not sure if there as any way to rename it.

Maybe we don't document it anymore and use the new registration key `QUERY`?

packages/core/modules-sdk/src/medusa-app.ts
```diff
@@ -420,53 +421,24 @@ async function MedusaApp_({
...
-   * @param queryConfig
-   * @param options
-   */
-  async function query<const TEntry extends string>(
-    queryConfig: RemoteQueryObjectConfig<TEntry>,
+  const query: RemoteQueryFunction = async (
+    queryOptions:
+      | RemoteQueryObjectConfig<any>
+      | RemoteQueryObjectFromStringResult<any>
+      | RemoteJoinerQuery,
```
thetutlage, line 428: Decided to type the query function and remove overloads from the implementation. This shouldn't impact the typings. 

But happy to discuss if you think otherwise

packages/core/modules-sdk/src/medusa-app.ts
```diff
@@ -478,20 +450,26 @@ async function MedusaApp_({
...
   }
 
+  /**
+   * Graph function uses the remoteQuery under the hood and
+   * returns a result set
+   */
+  query.graph = async function (queryOptions, options) {
+    const normalizedQuery = remoteQueryObjectFromString(queryOptions).__value
+    return await remoteQuery.query(normalizedQuery, undefined, options)
+  }
```
thetutlage: This is the new implementation

packages/core/types/src/modules-sdk/remote-query.ts
```diff
@@ -12,6 +14,29 @@ export type RemoteQueryFunctionReturnPagination = {
...
+ * QueryGraphFunction is a wrapper on top of remoteQuery
+ * that simplifies the input it accepts and returns
+ * a normalized/consistent output.
+ */
+export type QueryGraphFunction = {
+  <const TEntry extends string>(
+    queryConfig: RemoteQueryObjectConfig<TEntry>,
+    options?: RemoteJoinerOptions
+  ): Promise<Prettify<GraphResultSet<TEntry>>>
+}
```
thetutlage: These are the types for the new implementation

packages/core/modules-sdk/src/medusa-app.ts
carlos-r-l-rodrigues: I believe we can just add a note that this will be removed, and a few releases later we can remove this key

carlos-r-l-rodrigues: That can be done later, as we have to replace it everywhere in the core

thetutlage: Sure!

packages/core/types/src/modules-sdk/__tests__/query.spec.ts
```diff
@@ -0,0 +1,33 @@
...
+  describe("Infer via queryConfig", () => {
+    it("should infer return type of a known entry", async () => {
+      const graph = (() => {}) as unknown as QueryGraphFunction
+      const result = await graph({
+        entryPoint: "product",
+        fields: ["handle", "id"],
+      })
+
+      expectTypeOf(result).toEqualTypeOf<{
+        rows: FixtureEntryPoints["product"][]
```
srindom: Think `rows` is an awkward name when we are this many abstraction layers from the database. The more generic `data` would IMO be better.

thetutlage: Any suggestions for the naming?

packages/core/types/src/modules-sdk/remote-query.ts
```diff
@@ -12,6 +14,29 @@ export type RemoteQueryFunctionReturnPagination = {
...
 
+/**
+ * The GraphResultSet presents a typed output for the
+ * result returned by the underlying remote query
+ */
+export type GraphResultSet<TEntry extends string> = {
+  data: TEntry extends keyof RemoteQueryEntryPoints
+    ? RemoteQueryEntryPoints[TEntry][]
+    : any[]
+  metadata?: any
```
adrien2p: Q: Cant we use the pagination type we have created for the remote query?

adrien2p (approved):

adrien2p (approved):

PR #9038, opened 2024-09-06
feat(dashboard) modal search autofocus

fPolic:
**What**
- autofocus search input on route modal that have tables

---

CLOSES CC-132


kasperkristensen (approved):
LGTM, although the behaviour is not consistent at least on my end. Sometimes it correctly auto focuses and sometimes it does not. But I am thinking that might be an issue with our modals, so feel free to merge this as is 👍

riqwan (approved):

fPolic:
> LGTM, although the behaviour is not consistent at least on my end. Sometimes it correctly auto focuses and sometimes it does not. But I am thinking that might be an issue with our modals, so feel free to merge this as is 👍

Which modals are you getting inconsistent behaviour on? I'll merge since it worked on modals I tested but will look into it more.

PR #9041, opened 2024-09-06
fix: Update auth `app_metadata` when deleting users + customers

olivermrbl:
Closes CC-431


packages/medusa/src/api/admin/users/[id]/route.ts
```diff
@@ -71,10 +74,10 @@ export const DELETE = async (
   res: MedusaResponse<HttpTypes.AdminUserDeleteResponse>
 ) => {
   const { id } = req.params
```
Comment (sradevski) [inline packages/medusa/src/api/admin/users/[id]/route.ts:76]: We need to check here if the `id` matches the token `actor_id`, otherwise anyone can delete any customer account. When we have roles this might be ok (eg. user A ca delete user B if user A is an admin), but I am not sure if we want to allow that for now, WDYT?

packages/medusa/src/api/admin/customers/[id]/route.ts
```diff
@@ -68,10 +68,11 @@ export const DELETE = async (
   res: MedusaResponse<HttpTypes.AdminCustomerDeleteResponse>
 ) => {
   const id = req.params.id
```
Comment (sradevski) [inline packages/medusa/src/api/admin/customers/[id]/route.ts:70]: We should check that `id` and `actor_id` are same here

packages/core/core-flows/src/customer/workflows/remove-customer-account.ts
```diff
@@ -0,0 +1,75 @@
...
+        },
+      })
+
+      const authIdentity = transform(
+        { authIdentities },
+        ({ authIdentities }) => {
+          const authIdentity = authIdentities[0]
+
+          if (!authIdentity) {
+            throw new Error("Auth identity not found")
```
adrien2p: Suggestion: should we throw a MedusaError Invalide data or unexpected state? those errors are watched by the global error handler

adrien2p (approved):
LGTM, not much except stev comments

packages/core/core-flows/src/customer/workflows/remove-customer-account.ts
olivermrbl: Yes, nice catch. Will update!

packages/medusa/src/api/admin/customers/[id]/route.ts
Comment (olivermrbl) [reply packages/medusa/src/api/admin/customers/[id]/route.ts:70]: Are you sure? This is on the admin route. An admin user should be able to delete customers.

packages/medusa/src/api/admin/users/[id]/route.ts
Comment (olivermrbl) [reply packages/medusa/src/api/admin/users/[id]/route.ts:76]: I think both options are not great actually.

Without the ability to delete users, you can't remove an unwanted admin user. 
With the ability to delete users, the unwanted admin user can delete you.

Comment (olivermrbl) [reply packages/medusa/src/api/admin/users/[id]/route.ts:76]: Updated + added a test 👍

packages/medusa/src/api/admin/customers/[id]/route.ts
Comment (sradevski) [reply packages/medusa/src/api/admin/customers/[id]/route.ts:70]: Ah yes, you are right! I didn't pay attention to where this handler is located

packages/medusa/src/api/admin/users/[id]/route.ts
Comment (sradevski) [reply packages/medusa/src/api/admin/users/[id]/route.ts:76]: Yeah indeed, I guess for now the one that has access to the DB has the ultimate authority (so they can remove someone manually), but the same goes for restoring access when someone removes you, so there is no right or wrong here, I leave it up to you to decide which one you think is better.

sradevski (approved):

PR #9043, opened 2024-09-06
fix(types, medusa): request / response types clean-up

shahednasser:
More clean-up of request / response types


olivermrbl (approved):
Nice

PR #9045, opened 2024-09-06
chore: gql code gen as dev dependency

carlos-r-l-rodrigues:


packages/core/modules-sdk/src/utils/gql-schema-to-types.ts
```diff
@@ -1,9 +1,8 @@
-import { MedusaModule } from "../medusa-module"
+import { codegen } from "@graphql-codegen/core"
```
olivermrbl, line 1: Looks like the CLI pipeline is failing because this is needed at runtime and the package ships with the ones we've now moved to dev dependencies

carlos-r-l-rodrigues: doesn't it install dev dependencies in this step?
This tooling shouldn't be part of dependencies, it is a dev dependency.

thetutlage: @carlos-r-l-rodrigues We need it in dependencies because they are used when the user starts their application.

The development dependencies of the package won't be installed in the user app and hence this util will fail.

carlos-r-l-rodrigues: We should just skip it in production.
The test that failed is a "development server". The dependency should be there installed.

carlos-r-l-rodrigues: That has to be a medusa-cli dev dependency and we skip generating the schema in production.
But it will fail because it gets the medusa-cli from `@preview` to run the GH pipeline.
@olivermrbl

olivermrbl: That's right, I forgot about that. 

Can you confirm everything works as expected with the snapshot? Then, we should be good to merge

carlos-r-l-rodrigues: It worked fine in a Medusa project using the snapshot from `/medusa` and `/medusa-cli`

olivermrbl (approved):

packages/core/modules-sdk/src/utils/gql-schema-to-types.ts
thetutlage: Are we saying that these dependencies will be installed by the user app?

adrien2p: I think they should only be dev deps, the start command currently import the util to generate the types during development, once the user will deploy his app, they wont have any needs for types there. I would say that maybe instead of importing the util statically at the top which lead to importing codegen, we might want to import it dynamically + execute the type generation only for dev command (provide a flag or environment) and not in production and it should solve no? Wdyt?

carlos-r-l-rodrigues: That is what the PR is doing. Dynamic import and excluding the schema gen when in production.
@thetutlage adding it as a dev dependency of the cli worked, I didn't add it to my Medusa project. And the cli is a dependency of the project.

adrien2p: Yes sorry, the code looks good, i was more focused on on the discussion and missed the code itself 😅

thetutlage: Okay, I think we should get on a call whenever possible. Because, I am personally a bit lost :)

As per me. If you have a package (let's say medusa-cli) in this case. It's `devDependencies` won't be installed when someone installs this package. 

Maybe I am missing something here (quite possible), so happy to understand it together

carlos-r-l-rodrigues: @thetutlage is right. The `cli` devDependency won't be installed in a Medusa project. My local tests are probably mixing versions and that's why it works.
I'm closing this PR for now, and reopen it when update the starter.

PR #9046, opened 2024-09-07
fix(orchestration): filter out undefined items

carlos-r-l-rodrigues:
What:
* filter `undefined` values so the remote joiner stops resolving relations if there are no more items


olivermrbl (approved):
LGTM

PR #9055, opened 2024-09-09
fix: Allow specifying PORT for tests through environment

sradevski:
This is useful when you want to use something else other than axios/api function. In my case, I want to use EventSource, so I need to know the port of the server.

### DIFF
--- a/packages/core/medusa-test-utils/src/medusa-test-runner-utils/bootstrap-app.ts
+++ b/packages/core/medusa-test-utils/src/medusa-test-runner-utils/bootstrap-app.ts
@@ -19,7 +19,7 @@ async function bootstrapApp({
     expressApp: app,
   })
 
-  const PORT = await getPort()
+  const PORT = process.env.PORT ? parseInt(process.env.PORT) : await getPort()
 
   return {
     shutdown,


packages/core/medusa-test-utils/src/medusa-test-runner-utils/bootstrap-app.ts
```diff
@@ -19,7 +19,7 @@ async function bootstrapApp({
     expressApp: app,
   })
 
-  const PORT = await getPort()
+  const PORT = process.env.PORT ? parseInt(process.env.PORT) : await getPort()
```
adrien2p, line 22: Q: would it be useful to be able to provide an optional port as part of the `medusaIntegrationTestRunner` input args? and you could pass ({ port: process.env.PORT })

adrien2p (approved):

packages/core/medusa-test-utils/src/medusa-test-runner-utils/bootstrap-app.ts
sradevski: I was thinking like this you won't need to add an additional argument, and you can do
```
medusaTestRunner({env: {PORT: 3000}})
```

But I am happy to do it either way, let me know which one you prefer.

adrien2p: I think we can keep it as is, just wanted your opinion :)

PR #9058, opened 2024-09-09
breaking: turn off Mikro-ORM casing change with DML properties

thetutlage:
FIXES: FRMW-2688

This PR turns off the Mikro-ORM casing change behavior for the properties defined using DML. The idea is to have predictable naming conventions throughout the stack without performing any normalization. As a result **"What you define is what you get"**.

However, this approach will lead to a breaking change in existing applications using DML with `camelCase` property names. Because earlier, Mikro-ORM converted these property names to `snake_case` and created the database tables accordingly.

After this PR gets merged, we will turn off Mikro-ORM's default behavior, so you will have to rename properties within the DML to `snake_case`. If you prefer `camelCase,` you will have to drop existing tables, re-generate migrations, and run them.

We understand that this is a destructive change. But it will help us implement predictable naming conventions everywhere in your applications. The sooner the better.


riqwan (approved):
I thought this was the default behaviour. lgtm!

riqwan:
It might be worth checking if the migration generator catches this and performs a rename instead of a drop and create. If so, the breaking change could be handled gracefully.

thetutlage:
> It might be worth checking if the migration generator catches this and performs a rename instead of a drop and create. If so, the breaking change could be handled gracefully.

Yup, lemme give it a try.

adrien2p (approved):

adrien2p:
yes we need to validate that all the migration generation and custom migration patch we have are working, but I believe they should not be impacted by it

thetutlage:
CleanShot 2024-09-09 at 18 48 23@2x

@riqwan Columns are dropped. Which kind of makes sense, there is no way for Mikro-ORM to know that `my_email` is now `myEmail`.

olivermrbl (approved):

PR #9062, opened 2024-09-09
fix(medusa-test-utils): Run with modules without models

adrien2p:
RESOLVES FRMW-2698

**What**
- Allow to test module without models without issues
- Dynamically find from where to load the models
  - will look if src/models - dist/models or models exists from either the `resolve` or `process.cwd()` and load the models from there


sradevski (approved):
LGTM, I'll give it a try soon.

FYI, there is the file module that already passed a dummy model to circumvent the issue, maybe we can clean that up now (and also test this change with it)

adrien2p:
> LGTM, I'll give it a try soon.
> 
> FYI, there is the file module that already passed a dummy model to circumvent the issue, maybe we can clean that up now (and also test this change with it)

I ve updated the file integration and it works 🚀

packages/core/medusa-test-utils/src/module-test-runner.ts
```diff
@@ -43,8 +43,16 @@ export function moduleIntegrationTestRunner<TService = any>({
 
   process.env.LOG_LEVEL = "error"
 
-  moduleModels ??= Object.values(require(`${process.cwd()}/src/models`))
-  moduleModels = toMikroOrmEntities(moduleModels)
+  try {
+    moduleModels ??= Object.values(require(`${process.cwd()}/src/models`))
```
srindom: thought: this path should probably be dynamic depending on whether we are using the test runner within a module pkg or a medusa project.

For example, if I use this in my Medusa project I get:

I am getting this:

```
❯ jest -- src/modules/car/__tests__/index.ts
 FAIL  src/modules/car/__tests__/index.ts
  ● Test suite failed to run

    Cannot find module '/Users/sebastianrindom/Developer/sebs-test-store/src/models' from 'node_modules/medusa-test-utils/dist/module-test-runner.js'
```

adrien2p: should be fixed, I ll generate a snap

adrien2p: no sorry I ve been too fast, I ll get back to it

adrien2p: Just tested with the following and it works

Screenshot 2024-09-09 at 18 39 21

Screenshot 2024-09-09 at 18 39 46

adrien2p: Ill fix the tests tomorrow

adrien2p:
@srindom @sradevski I ll generate a snap for you if you want to give it a try before merging, but I ve tested on my side and it was working

sradevski:
@adrien2p there is some issue with peer dependencies when installing the snapshot, let's just get it merged and if there are still issues (I don't expect any TBH), then we can look into the fix into a separate PR, WDYT?

adrien2p:
> @adrien2p there is some issue with peer dependencies when installing the snapshot, let's just get it merged and if there are still issues (I don't expect any TBH), then we can look into the fix into a separate PR, WDYT?

let me just quickly check something?

on another note we discussed a solution to finally truly fix dev deps and optional peer deps for dev env, we can chat later on it

adrien2p:
@sradevski for now, I ve moved them all to deps instead, since this package will always be a dev deps in a project, it shouldn't be a problem I would say, we can try like that and see, wdyt?

sradevski:
Let me try it out

sradevski:
@adrien2p all good, we can merge it!

adrien2p:
🚀 I just have an issue with the API key tests 🤔

adrien2p:
@sradevski I believe it will be mergeable now, the deps should be fixed

PR #9063, opened 2024-09-09
fix(dashboard): rq cache and pending flag

fPolic:
**What**
- fix issues from stale data on modal pages

---

FIXES CC-137


olivermrbl (approved):
Nice, LGTM

PR #9065, opened 2024-09-09
feat: Add support for fetch streaming to js SDK

sradevski:
CLOSES CC-441

### DIFF
--- a/packages/core/js-sdk/src/client.ts
+++ b/packages/core/js-sdk/src/client.ts
@@ -1,5 +1,13 @@
 import qs from "qs"
-import { ClientFetch, Config, FetchArgs, FetchInput, Logger } from "./types"
+import { events } from "fetch-event-stream"
+import {
+  ClientFetch,
+  Config,
+  FetchArgs,
+  FetchInput,
+  FetchStreamResponse,
+  Logger,
+} from "./types"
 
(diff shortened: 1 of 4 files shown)


sradevski:
@olivermrbl this is a very basic version, and we can enhance it as needed going forward.

packages/core/js-sdk/src/client.ts
```diff
@@ -1,5 +1,13 @@
 import qs from "qs"
-import { ClientFetch, Config, FetchArgs, FetchInput, Logger } from "./types"
+import { events } from "fetch-event-stream"
```
olivermrbl, line 2: comment(non-blocking): looks like this package is only a single file, so we could include it ourselves, if we don't want another dep

olivermrbl (approved):
Neat, looks good to me!

packages/core/js-sdk/src/client.ts
sradevski: Yeah it is, I also thought about copying it over, but I don't see the harm in having it as a dependency TBH so I thought of leaving it as-is. Let me know if you think we should get rid of it, and just copy it over

sradevski: I'll get this merged as I want to use it in platform already but if you think we should remove the dependency I'll do it in a separate PR

PR #9068, opened 2024-09-09
feat(utils,types,framework,medusa): store endpoints should require publishable key

riqwan:
what:

- requires the store apis to always have a valid publishable key
- updates specs to now conform to this
- updates js-sdk to include publishable keys for store endpoints

The medusa starter already creates a publishable key, nothing to change there. 

Related:

- adds publishable keys to storefront - https://github.com/medusajs/nextjs-starter-medusa/pull/365

Some open questions:

- Should we throw a 401 for publishable key? Since its not really authenticating anything, it feels wrong - I've went ahead with a 400 and not_allowed. Will switch to 401 if there's consensus.
- If the publishable key is present and no sales channels are configured, should we still default to the store sales channel?
- Shall we merge and publish all the changes for this (backend starter, sdk and core) at one go? Or do we need to do a phased rollout here?

RESOLVES CC-98


packages/core/js-sdk/src/store/index.ts
```diff
@@ -5,13 +5,29 @@ import {
...
   private client: Client
+  private config: Config
 
-  constructor(client: Client) {
+  constructor(client: Client, config: Config) {
     this.client = client
+    this.config = config
+  }
+
+  private withDefaultStoreHeaders(clientHeaders?: ClientHeaders) {
```
sradevski: The `client` has a `globalHeaders` field, so we can just set the `x-publishable-api-key` header there instead.

Basically, in the the `class Medusa` constructor just do
```
if (config.publishableKey) {
  config.globalHeaders['x-...'] = config.publishableKey
}
```

riqwan: won't that add the publishable key to all endpoints instead of just the store?

sradevski: It would, but it doesn't do any harm if you do, and you would rarely (if ever) call all of them from the same SDK instance.

But if you prefer this approach, I am happy to keep it as-is, it's just a bit more boilerplate, that's all

riqwan: No strong opinion on this, will keep this open along with the other questions from the PR and change if there's consensus.

packages/medusa/src/api/utils/middlewares/products/filter-by-valid-sales-channels.ts
```diff
@@ -1,77 +1,49 @@
...
-      "default_sales_channel_id",
-    ])
-
-    if (!store) {
-      return next(
-        new MedusaError(MedusaError.Types.INVALID_DATA, `Store not found`)
+    return next(
+      new MedusaError(
+        MedusaError.Types.INVALID_DATA,
+        `Publishable key needs to have a sales channel configured`
```
fPolic, line 45: question: why don't we use the default sales channel here anymore?

fPolic (approved):
LGTM!

packages/medusa/src/api/utils/middlewares/products/filter-by-valid-sales-channels.ts
riqwan: This is an open question. Now that we require publishable keys on all store endpoints, should we push the developers to setup their publishable keys more explicitly than falling back to the default sales channel? wdyt?

fPolic: I think that fallback to the default is more convenient and that's why we have the default channel but no strong opinion there.

packages/medusa/src/utils/middlewares/ensure-publishable-api-key.ts
```diff
@@ -0,0 +1,68 @@
...
+} from "@medusajs/framework"
+import {
+  ApiKeyType,
+  isPresent,
+  MedusaError,
+  PUBLISHABLE_KEY_HEADER,
+} from "@medusajs/utils"
+import { refetchEntity } from "../../api/utils/refetch-entity"
+
+export function ensurePublishableApiKey() {
```
olivermrbl, line 14: suggestion: Albeit not something we typically do, here it might make sense to throw a more guiding error message, because it is otherwise keeping them from getting started. Something like:
```
Publishable API key required in the request header: {PUBLISHABLE_KEY_HEADER}.
You can manage your keys in settings in the dashboard.
```

Wdyt?

packages/core/js-sdk/src/store/index.ts
olivermrbl: Without a global header, external developers would have to pass this header themselves if they extend the SDK with more store endpoints right?

riqwan: yes, thats true. Reverted this!

packages/medusa/src/utils/middlewares/ensure-publishable-api-key.ts
riqwan: Much better, changed

packages/core/js-sdk/src/store/index.ts
olivermrbl: This would maybe be an argument for configuring this as a global header, as @sradevski suggested. Wdyt?

I didn't mean to suggest we revert the changes entirely, since now we don't populate the header if the SDK is configured with the key.

riqwan: I checked it, we already do add to the default headers. This change wasn't necessary

riqwan: Screenshot 2024-09-11 at 11 01 08

olivermrbl: Got it, nice!

olivermrbl:
@riqwan, ready to merge?

riqwan:
Lets do it! 🎉

PR #9074, opened 2024-09-10
fix: JSON stringify DML json property default value

thetutlage:
FIXES: FRMW-2690


packages/core/utils/src/dml/helpers/entity-builder/define-property.ts
```diff
@@ -214,6 +214,27 @@ export function defineProperty(
...
+      fieldName: field.fieldName,
+      /**
+       * MikroORM does not ignore undefined values for default when generating
+       * the database schema SQL. Conditionally add it here to prevent undefined
+       * from being set as default value in SQL.
+       */
+      ...(isDefined(field.defaultValue) && {
+        default: JSON.stringify(field.defaultValue),
+      }),
+    })(MikroORMEntity.prototype, field.fieldName)
```
adrien2p, line 234:
```suggestion
const propertyConfig: PropertyOptions<any> = {  
  columnType: "jsonb",
  type: "any",
  nullable: field.nullable,
  fieldName: field.fieldName,
}

if (isDefined(field.defaultValue)) {
  propertyConfig.default = JSON.stringify(field.defaultValue)
}

Property(propertyConfig)(MikroORMEntity.prototype, field.fieldName)
```

I feel like it would be easier to read that way, wdyt?

adrien2p (approved):

packages/core/utils/src/dml/helpers/entity-builder/define-property.ts
thetutlage: We have same inline ternary else-where for other properties as well. So it decided to keep it uniformed everywhere.

PR #9084, opened 2024-09-10
chore: joiner config entity property

carlos-r-l-rodrigues:
What:
* promote entity as a property of joiner config relashionships
* defineLink - forward arguments on field aliases and small fixes

Breaking Change:
* rename Customer module Address related methods to CustomerAddress (e.g: listCustomerAddress)


adrien2p (approved):
LGTM, do you think we should make the entity required? It will always refers to a type in the graphql in theory (even though not in the db) wdyt?

carlos-r-l-rodrigues:
> LGTM, do you think we should make the entity required? It will always refers to a type in the graphql in theory (even though not in the db) wdyt?

I believe we can keep it optional, that way it is still possible to link a Module to other directly. Suppose module X have 10 entities containing field `z_id`, 1 single link from X to module Z would be enough.

Issue #9075, opened 2024-09-10
Data Model /Service Factory error generating manyToMany relationships


yuning-zhang:
### Describe the bug

While generating manyToMany relationship between two models, retrieved relationship is not correct (using retrieve or list). This is possible due to ORM switching the select statement from different tables.

### System information

Medusa version (including plugins): v2.0.8-preview
Node.js version: v20.12.2
Database: postgres (PostgreSQL) 14.12 (Homebrew)
Operating system: macOS 14.6
Browser (if relevant):

### Steps to reproduce the behavior

Example repo of this bug https://github.com/yuning-zhang/medusa-bug

1. Set up this models
```
const Role = model.define("role", {
  id: model.id().primaryKey(),
  name: model.text().unique(),
  employees: model.manyToMany(() => Employee),
});

const Employee = model.define("employee", {
  id: model.id().primaryKey(),
  name: model.text().nullable(),
  first_name: model.text().nullable(),
  last_name: model.text().nullable(),
  email: model.text().unique(),
  roles: model.manyToMany(() => Role),
});
```

2. Generate migrations and run
3. Set up test route to generate example data
```
// src/api/company/test/route.ts

import { AuthenticatedMedusaRequest, MedusaResponse } from "@medusajs/medusa";
import B2BModuleService from "../../../modules/b2b/service";
import { B2B_MODULE } from "../../../modules/b2b";

type RequestBody = {};

export const POST = async (
  req: AuthenticatedMedusaRequest<RequestBody>,
  res: MedusaResponse
) => {
  const b2bModuleService: B2BModuleService = req.scope.resolve(B2B_MODULE);

  const role = await b2bModuleService.createRoles({
    name: "test_role",
  });

  const employee = await b2bModuleService.createEmployees({
    name: "Hello World",
    first_name: "Hello",
    last_name: "World",
    email: "test@test.com",
    roles: [role.id],
  });

  const fullEmployee = await b2bModuleService.retrieveEmployee(employee.id, {
    relations: ["roles"],
  });

  res.json({ role, employee, fullEmployee });
};

```

4. Send request to test route
```
curl -X POST 'http://localhost:9000/company/test' \
-H 'Content-Type: application/json' \
--data-raw '{}'
```

5. Received wrong response with roles not correct within employee. As you can see, role and employee are created correctly. But to retrieve the employee again with the roles relationship, the role id is actually the employee id, and the role name is the employee name
```
{
    -----------correct role created-------------
    "role": {
[...]

PR #9085, opened 2024-09-10
fix(dml): Throw when many to many are not configured properly

adrien2p:
**What**
When both sides of a many to many relationship do not define the mapped by option, it leads to misconfigured relations and a malformed SQL query. (ref. https://github.com/medusajs/medusa/issues/9075)

- When both mapped by are not defined, infer look up to try to identify the missing configuration if any
- Disallow defining many to many only on one side


packages/core/utils/src/dml/integration-tests/__tests__/many-to-many.spec.ts
```diff
@@ -186,4 +186,40 @@ describe("manyToMany - manyToMany", () => {
       ]),
     })
   })
+
+  it(`should fail to load the dml's if both side of the relation are missing the mappedBy options`, () => {
+    const team = model.define("team", {
+      id: model.id().primaryKey(),
+      name: model.text(),
+      users: model.manyToMany(() => user, {
```
riqwan, line 194: q: the `user` model doesn't have a `teams.manyToMany`. Is the inverse optional?

riqwan (approved):
lgtm!

packages/core/utils/src/dml/integration-tests/__tests__/many-to-many.spec.ts
adrien2p: It has the `squads` on the `user` model which is the one that refers to the `team`, now I haven't named it `teams` but it is the relation that is important. So you can add mappedby on the `squads` to `user` or on `users` to `squads`. does that make sense?

I get why squads can be misleading cause it could also refers to the pivot 😂 right?

riqwan: Ah, I see. I missed that. So then follow up question, if the inverse is not present, does that throw an error? i.e. if the user.squads relationship did not exist or is that allowed?

packages/core/utils/src/dml/helpers/entity-builder/define-relationship.ts
```diff
@@ -280,6 +280,40 @@ export function defineManyToManyRelationship(
         `${MikroORMEntity.name}.${relationship.name}`
       ] = true
     }
+  } else {
+    /**
+     * Since we dont have the information about the other side of the
+     * relationship, we will try to find by finding all the other side many to many that refers to the current entity.
+     * If there is any, we will try to find if at least on of them has a mappedBy.
```
sradevski:
```suggestion
     * If there is any, we will try to find if at least one of them has a mappedBy.
```

packages/core/utils/src/dml/helpers/entity-builder/define-relationship.ts
```diff
@@ -280,6 +280,40 @@ export function defineManyToManyRelationship(
         `${MikroORMEntity.name}.${relationship.name}`
       ] = true
     }
+  } else {
+    /**
+     * Since we dont have the information about the other side of the
+     * relationship, we will try to find by finding all the other side many to many that refers to the current entity.
```
sradevski:
```suggestion
     * relationship, we will try to find all the other side many to many that refers to the current entity.
```

sradevski (approved):
Nice!

packages/core/utils/src/dml/integration-tests/__tests__/many-to-many.spec.ts
adrien2p: No you can't but, indeed, we don't check that, mikro orm wont complain if you do that, but you won't be able to have a real many to many and the result of querying will be missing the other side relation even though you can specify it as part of the populate, let me add that.

basically, it is the case where you define a many to many only one one side without custom pivot and without mappedBy

adrien2p: I ve updated the check, you can have another look

riqwan: Nice, looks good man! 🔥

adrien2p: thanks to you for mentionning the missing part 😄

adrien2p:
@olivermrbl should we consider this fix a breaking change?

olivermrbl:
I can include a comment about this in the release notes 👍

olivermrbl:
Can we merge?

adrien2p:
> Can we merge? 

Yes we can

PR #9086, opened 2024-09-10
feat(core-flows,types,medusa): validate deleting location level when quantities exist

riqwan:
what:

- adds validation in workflow to prevent deleting a location level when reserved or stocked quantity exists
- disabled delete button when quantities exist
- consolidate delete workflows

Screenshot 2024-09-10 at 16 39 02

RESOLVES CC-120


sradevski (approved):

fPolic (approved):
LGTM

PR #9094, opened 2024-09-11
fix: `instanceof MedusaError` does not work

### LINKED ISSUE #9088: [bug] `instanceof MedusaError` does not work
### Describe the bug

When trying to check error type using the `instanceof` or `error.constructor.name`, MedusaErrors won't identify as `MedusaError`.

### System information

Medusa version (including plugins): 1.20
Node.js version: 22/20
Database: Postgres
Operating system: Linux

### Expected behavior

`error instaceof MedusaError === true`

### Additional context

`this.name = this.constructor.name;` or any other value would fix this.

This issue also persists in v2, if I am not mistaken.

McTom234:
Closes #9088

As stated in the issue and [this block article](https://davidwalsh.name/detect-error-type-javascript), adding one line of code, fixes the `instanceof` method.

# What

Set the name of the `MedusaError` type.

# Why

Developers can differ between `MedusaError`s and other errors easily.

I already have an idea/suggestion to improve the default error handler with this new possibility (see below).

# Testing

Somewhere in your code, throw a `MedusaError` and a generic `Error`. Try this with and without this fix. For both, get the `instanceof Error` and `instanceof MedusaError`.

Before the fix, both return true for Error. After, the medusa error will additionally return true for the `MedusaError`.

Because of that, no breaking changes were made.

# Further possibilities

The current behavior of the `errorHandler` for generic/non-medusa-errors is to return a `500` with message `An unknown error occurred.`
This is cool in most cases. Some developers might prefer to get the actual error message returned, if it is not a `MedusaError`. The `instanceof` can be used in that case. This configuration could be added via the medusa config or via the parameters for the default `errorHandler` function.

This shall not be part of this PR and just an insight on why this might be useful.


.changeset/kind-fireants-obey.md
```diff
@@ -0,0 +1,15 @@
...
+
+# Why?
+
+You can now identify the error type using the `error.constructor.name` or `error instanceof` method.
+
+# How to update your code?
+
+Initially, probably no change is required.
+In the future you might find yourself in a situation, where handling a `MedusaError` is different from other errors.
+
```
riqwan: we've temporarily removed adding changesets, can you remove this from the PR?

McTom234: yeah, although then please update the contributing guideline :wink:

riqwan: haha, touché!

riqwan (approved):
LGTM, some specs are failing!

McTom234:
@riqwan Oh, I see, at least one test fails, maybe some more. The default name of the error will change. As you can test locally, the `instanceof` will remain functional.

As you test for the name of the error, do you rely on that somewhere in the code or should I just update the integration tests?

riqwan:
> As you test for the name of the error, do you rely on that somewhere in the code or should I just update the integration tests?

I think this is a safe change, you just need to update the test.

We handled this issue differently ourselves - [ref](https://github.com/medusajs/medusa/blob/develop/packages/core/utils/src/dml/entity.ts#L86-L88). I don't know if that will be translatable here. Will loop in @adrien2p for a second opinion.

packages/core/utils/src/common/errors.ts
```diff
@@ -46,6 +46,8 @@ export class MedusaError extends Error {
   constructor(type: string, message: string, code?: string, ...params: any) {
     super(...params)
 
+    this.name = this.constructor.name
```
adrien2p: I would suggest to do the following to follow our practices and be consistent with the rest of the code base

```ts
export class MedusaError extends Error {
  // we do not use Symbol because it is not serializable
  __isMedusaError = true

  public type: string
  public message: string
  public code?: string
  public date: Date
  public static Types = MedusaErrorTypes
  public static Codes = MedusaErrorCodes

  static isMedusaError(error: any): error is MedusaError {
    return !!error.__isMedusaError
  }
```

and then when checking somewhere you can do `MedusaError.isMedusaError(err)`

wdyt?

McTom234: You have your point :+1: 

In that case, I would add a hint in the docs as well?

riqwan: Sure, open that as a different PR as it requires different reviewers.

McTom234:
@riqwan I applied the changes. Could you run the pipelines again?

McTom234:
Seems good to go @riqwan :smile:

riqwan:
@McTom234 legend! Thanks for the contribution!

PR #9096, opened 2024-09-11
fix: dropdown-content not scrollable

### LINKED ISSUE #8743: [UI] DropdownMenu content not scrollable
### Describe the bug

The content of a DropdownMenu cannot be scrolled.

### System information

Medusa version (@medusajs/ui): 3.0.1

1. Create a dropdown menu as instructed in docs.
2. Fill it with many items (or lower your screen height).
3. Cannot access all items.

### Expected behavior

Scroll to see all elements.

### Code snippets

I think the problem is here: https://github.com/medusajs/medusa/blob/531285c1e91bba7be9854131646aadbaf1fd5b7d/packages/design-system/ui/src/components/dropdown-menu/dropdown-menu.tsx#L110

Is there a reason for the `overflow-hidden` (which is quite strong :sweat_smile:)? Otherwise, just replace it with a x-specific one...

McTom234:
Fixes #8743

I think all details can be found in the issue.

### DIFF
--- a/.changeset/pink-paws-cry.md
+++ b/.changeset/pink-paws-cry.md
@@ -1,0 +1,6 @@
+---
+"@medusajs/ui": minor
+---
+
+fix dropdown menu not scrollable
+
--- a/packages/design-system/ui/src/components/dropdown-menu/dropdown-menu.tsx
+++ b/packages/design-system/ui/src/components/dropdown-menu/dropdown-menu.tsx
@@ -56,7 +56,7 @@ const SubMenuTrigger = React.forwardRef<
     {...props}
   >
     {children}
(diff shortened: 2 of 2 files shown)


kasperkristensen:
Hi @McTom234, thanks for the PR. 

The DropdownMenu isn't intended to be scrollable. The purpose of the DropdownMenu in the context of our design system, is for small action menus, like the ones you find in the top bar of MacOS.

So I am closing this PR, as I don't want the component to allow breaking with our styling guidelines by default. If you have a case where you want to have a dropdown that allows overflowing then you can simply add styles to support that. All components and subcomponents accept a `className` prop, where you can pass `overflow-y-auto` which will allow the component to scroll. In your case you would add the class to Content and SubContent.

PR #9097, opened 2024-09-11
fix(core-flows): item id in reservations

fPolic:
**What**
- on cart completion, when creating reservations, use line item id instead of cart item id
- NOTE: inventory reservation is now done after we create the order

---

FIXES CC-448


carlos-r-l-rodrigues (approved):
LGTM

srindom:
thought(non-blocking): could it make sense to reserve in the same way we are doing it now, but then move the reservation from the cart line item to the order line item as the final step? I don't have strong opinions on this just thinking that it would be beneficial to terminate the workflow early if the reservation fails to happen.

(this could be something to consider at some other point, just wanted to put the idea out there)

fPolic:
> thought(non-blocking): could it make sense to reserve in the same way we are doing it now, but then move the reservation from the cart line item to the order line item as the final step? I don't have strong opinions on this just thinking that it would be beneficial to terminate the workflow early if the reservation fails to happen.
> 
> (this could be something to consider at some other point, just wanted to put the idea out there)

I was thinking of taking this route initially but we don't have a link from which we would know to match order_line_item <> cart_line_item.

PR #9099, opened 2024-09-11
docs: update store cURL examples in OAS

shahednasser:
Update the store cURL examples in OAS / generated specs in API reference to add the PAK header


riqwan (approved):
🥇

olivermrbl (approved):

PR #9103, opened 2024-09-11
chore: Register Query in container

adrien2p:
RESOLVES FRMW-2702

**What**
Register and make the Query available


olivermrbl (approved):
LGTM

PR #9106, opened 2024-09-11
feat(core-flows,dashboard,types,medusa): delete shipping methods when all inbound/outbound items are deleted

riqwan:
what:

- delete shipping methods when all inbound/outbound items are deleted
  - claims / exchanges / returns
- updates frontend to update the forms when ^ happens
- resolved a couple of bugs + types along the way

RESOLVES CC-439


fPolic (approved):
Looks good!

PR #9109, opened 2024-09-11
feat: introduce a thin wrapper on top of OpenTelemetry

thetutlage:
Fixes: CC-442

This PR is the first step towards integration telemetry tracing inside Medusa core. Here we introduce a thin wrapper on top of Open Telemetry.


packages/framework/framework/src/telemetry/index.ts
```diff
@@ -0,0 +1,79 @@
...
+   */
+  getOTTracer() {
+    return this.#otTracer
+  }
+
+  /**
+   * Trace a function call. Using this method will create a
+   * child scope for the invocations within the callback.
+   */
+  trace<F extends (span: Span) => unknown>(
```
sradevski, line 39: Not actionable, just talking out loud - having the tracer take a callback means that all functions will have additional indentation, which makes readability a bit worse.

Now the good thing is, you will rarely need to trace manually, so it's probably not an issue. Also, I can't think of a better interface for it that won't introduce another confusion (eg. having to manually do `span.end` at the end of a function).

sradevski (approved):
It's a great start 👍

packages/framework/framework/src/telemetry/index.ts
thetutlage: The issue comes directly from OpenTelemetry SDK. In order to create parent/child relationship between spans, you have to use `tracer.startActiveSpan` method, which accepts a callback.

There is also `tracer.startSpan` method, but there isn't any way to then assign a parent span to it. For example:

```ts
const parentSpan = tracer.startSpan('parent')

// There is no way for me to tell the child span to be part of the parent span
const childSpan = tracer.startSpan('child')
```

Whereas in the following example:

```ts
tracer.startActiveSpan('parent', (parentSpan) => {
  tracer.startActiveSpan('child', (childSpan) => {
    // childSpan is automatically child of parentSpan
  })
})
```

sradevski: Yeah completely understandable, I was just talking out loud, I know there isn't a lot we can do about it :)

PR #9002, opened 2024-09-04
fix(promotion): handle promotion buy X get X scenario

riqwan:
what:

When a buy get promotion is created where the target and the buy item is the same, we don't handle this scenario correctly as we would end up applying the promotion on the same items that cause it to qualify for the promotion.

This PR removes the qualified items from the buy side when applying promotions on the target.

RESOLVES CC-416

PR #9112, opened 2024-09-12
fix: OrderDTO add first class discount_subtotal var

420coupe:
Include discount_subtotal as a first class var for OrderDTO

mentioned in PR #9002


riqwan (approved):
Nice, thanks @420coupe!

PR #9116, opened 2024-09-12
feature: add telemetry to the HTTP layer

thetutlage:
FIXES: CC-443

Alright. So this PR allows us to enable telemetry inside the Medusa API and have auto instrumentation for HTTP requests, middleware (with some caveats) and the route handler. You just have to drop the following lines of code inside the `medusa-config.js` file and everything will be working.

```ts
// medusa-config.js
const { loadEnv, defineConfig } = require('@medusajs/utils')
const { ZipkinExporter } = require('@opentelemetry/exporter-zipkin')
const { instrumentHttpLayer, registerOtel } = require("@medusajs/medusa")

registerOtel({
  serviceName: 'medusajs',
  exporter: new ZipkinExporter({
    serviceName: 'medusajs',
  })
})
instrumentHttpLayer()

// Rest of the file goes here
loadEnv(process.env.NODE_ENV, process.cwd())
```

The `ZipkinExporter` in the above example can be replaced with other exporter supported by the OpenTelemetry Node.js SDK. Lemme know if you need any help setting up a different service. 

## Current state of things.

- Every HTTP request is instrumented automatically and acts as the parent span.
- All middleware are instrumented automatically. However, there are some caveats with middleware (covered in the next section).
- All route handlers are instrumented automatically.

### Caveat 1

Currently middleware inside our codebase are anonymous functions, hence they all appear the same in the tracing UI (as shown in the following screenshot)

CleanShot 2024-09-12 at 16 19 14@2x

In order to improve this situation. We have to provide names to middleware. I have done that for a few and will rename the rest as well.

**Screenshot after names have been applied**
CleanShot 2024-09-12 at 16 29 13@2x

### Caveat 2

Some of middleware internally calls additional middleware. For example: The `unlessPath` middleware https://github.com/medusajs/medusa/blob/develop/packages/medusa/src/api/utils/unless-path.ts#L7-L15. Now the internal middleware cannot be detected/captured at the boot time, hence the instrumentation cannot wrap it. However, we can still wrap them at runtime, but it will require modifications within the source of the middleware. So let's discuss if we want to go that route right now or not. 

---

@sradevski Please give it a try and let me know if things needs improvements at the HTTP layer (I am sure they will). Also, frontend -> backend tracing is not done yet. But that needs a 2liner change. I was thinking to jump on a call with you after we have telemetry in the frontend and wrap it up together.

---

Special thanks to @adrien2p and @carlos-r-l-rodrigues for brainstorming a few ideas and pairing with me on this :)

---

## Just for fun

A big trace from products create call

CleanShot 2024-09-12 at 17 05 27@2x

CC @srindom @olivermrbl


packages/medusa/package.json
```diff
@@ -22,6 +22,11 @@
   "license": "MIT",
   "devDependencies": {
     "@medusajs/types": "^1.11.16",
+    "@opentelemetry/instrumentation": "^0.53.0",
+    "@opentelemetry/instrumentation-pg": "^0.44.0",
+    "@opentelemetry/resources": "^1.26.0",
+    "@opentelemetry/sdk-node": "^0.53.0",
+    "@opentelemetry/sdk-trace-node": "^1.26.0",
```
thetutlage, line 29: Peer dependencies on the starter. Required if the user wants to enable telemetry.

packages/medusa/src/instrumentation/index.ts
```diff
@@ -82,16 +89,28 @@ export function instrumentHttpLayer() {
...
+              reject(error)
+            } else {
+              span.end()
+              resolve()
+            }
+          })
+        })
       })
+        .catch(next)
+        .then(next)
```
adrien2p: Would it be more clear with this?

```ts
const next_ = (err?: any) => {
  if (err) {
    span.setStatus({
      code: SpanStatusCode.ERROR,
      message: error.message || "Failed",
    })
  }
  span.end()
  next(err)
}

return handler(req, res, next_)
```

thetutlage: Yeah. Can move it out.

srindom:
I'd like to see the side-effects moved out of `medusa-config` and into a root level `instrumentation.[t|j]s` file ideally with an exported function similar to next.js's approach. I know there's not much difference in behavior of this right now, but when we get around to some more advanced bundling it will be beneficial to have these things separated.

packages/medusa/src/instrumentation/index.ts
adrien2p: But also is the nested promise required? Without it it should work since next_ will be called at the right time always by the handler

thetutlage:
@srindom Sure!

packages/medusa/src/instrumentation/index.ts
thetutlage: Yeah, the promise is required. I can explain this over a call. Its because of some surprising behavior from OpenTelemetry

sradevski (approved):
Looks good!

Just one comment on the initialization - does it make sense to have something like this:

```
projectConfig: {
   tracing: {
      middleware: true,
      routeHandler: true
      workflow: true
      step: false
      ...
   }
}
```

And the rest of the initialization happens inside `medusajs/medusa` instead of the config?

The medusa-config.js has been very config-driven so far, so it might make sense to preserve that approach for the tracing.

Btw this is definitely not a blocker, nor do we need to do anything about it, just sharing my thoughts

thetutlage:
@sradevski Ohh yes, 100%. It will be nice to have these knobs to enable/disable instrumentation in different layers. I will keep the suggestion in mind and add it once we touch this part of the implementation.

adrien2p (approved):

PR #9119, opened 2024-09-12
feat: run nested async workflows

carlos-r-l-rodrigues:


packages/core/orchestration/src/transaction/transaction-orchestrator.ts
```diff
@@ -783,6 +783,10 @@ export class TransactionOrchestrator extends EventEmitter {
               })
           )
         } else {
+          console.log(flow.modelId, "BEGIN EXECUTION", {
+            metadata: flow.metadata,
+          })
```
adrien2p: todo: rm :)

adrien2p (approved):

adrien2p:
should we add some tests to validate the event releasing with nested workflows?

packages/modules/workflow-engine-redis/integration-tests/__tests__/index.spec.ts
```diff
@@ -35,10 +35,9 @@ moduleIntegrationTestRunner<IWorkflowEngineService>({
   },
   testSuite: ({ service: workflowOrcModule, medusaApp }) => {
     describe("Workflow Orchestrator module", function () {
-      const afterEach_ = async () => {
+      const afterEach_ = async () => {}
```
sradevski: Very NIT: We could just remove this

packages/modules/workflow-engine-redis/src/services/workflow-orchestrator.ts
```diff
@@ -198,21 +222,35 @@ export class WorkflowOrchestratorService {
...
-    // TODO: temporary
+    const hasFinished = ret.transaction.hasFinished()
+    const metadata = ret.transaction.getFlow().metadata
+    const { parentStepIdempotencyKey } = metadata ?? {}
+    const hasFailed = [
+      TransactionState.REVERTED,
+      TransactionState.FAILED,
+    ].includes(ret.transaction.getFlow().state)
+
+    // TODO: temporary - Create an Aknowledgement type
```
sradevski: I think Shahed already created one, but I am not 100% certain

sradevski (approved):
Nice!

packages/core/orchestration/src/transaction/types.ts
```diff
@@ -237,6 +242,7 @@ export type TransactionFlow = {
   transactionId: string
   metadata?: {
     eventGroupId?: string
+    parentIdempotencyKey?: string
```
adrien2p, line 245: suggestion: should we store the parent workflow id, I am thinking for example for the UI and for the telemetry that way we don't have to infer it from the idempotency key which in theory can be provided by the consumer. does that make sense?

adrien2p (approved):
LGTM, we will just have to keep an eye on bull :D

adrien2p:
This will probably fix import/export as well as I just saw that it was relying on async run as step inside a non async workflow 😂 cc @sradevski I think we have been lucky with this one right? (do we test non happy path for it?)

sradevski:
@adrien2p yeah could be that we haven't caught it, there are few tests but it's not extensively tested.

packages/core/orchestration/src/transaction/types.ts
carlos-r-l-rodrigues: I will let this for another iteration where we can potentially have it as an array.
I believe when we focus in the UI we can tackle this and decide which approach is better.

PR #9123, opened 2024-09-12
fix(types, medusa): fix http type of exchanges and claims

shahednasser:
Remove the extending of the order's type from exchange / claim types in http types


riqwan (approved):
🥇

PR #9128, opened 2024-09-13
feature: add tracing to remote query

thetutlage:
Fixes: FRMW-2705

This PR adds tracing to the remote query. You can enable it by adding the following lines of code to the `instrumentation.js` file.

```ts
import { instrumentRemoteQuery } from '@medusajs/medusa'

export function register() {
  // ... rest of the setup
  instrumentRemoteQuery()
}
```


packages/core/modules-sdk/src/remote-query/query.ts
```diff
@@ -20,6 +20,26 @@ import {
...
+    queryOptions: RemoteQueryObjectConfig<any>
+  ) => Promise<any>
+
+  /**
+   * Method to wrap execution of the remoteQuery overload function
+   * for instrumentation
+   */
+  static traceRemoteQuery?: (
+    queryFn: () => Promise<any>,
+    queryOptions:
```
adrien2p, line 37: suggestion: should we rename the options to config? The options are used for other things

packages/core/modules-sdk/src/remote-query/query.ts
```diff
@@ -76,6 +96,13 @@ export class Query {
     }
 
     const config = this.#unwrapQueryConfig(queryOptions)
+    if (Query.traceRemoteQuery) {
+      return await Query.traceRemoteQuery(
+        async () => this.#remoteQuery.query(config, undefined, options),
+        queryOptions
```
adrien2p, line 114: todo: you might want to pass the config instead of the options here, or both eventually

packages/core/modules-sdk/src/remote-query/query.ts
```diff
@@ -98,11 +125,27 @@ export class Query {
...
+
+    /**
+     * When traceGraphQuery method is defined, we will wrap the implementation
+     * inside a callback and provide the method to the traceGraphQuery
+     */
+    if (Query.traceGraphQuery) {
+      response = await Query.traceGraphQuery(
+        async () =>
+          this.#remoteQuery.query(normalizedQuery, undefined, options),
+        queryOptions
```
adrien2p, line 152: todo: same as the previous one

packages/medusa/src/instrumentation/index.ts
```diff
@@ -117,6 +117,46 @@ export function instrumentHttpLayer() {
...
+        span.setAttributes({
+          "query.fields": queryOptions.fields,
+        })
+        return await queryFn().finally(() => span.end())
+      }
+    )
+  }
+
+  Query.traceRemoteQuery = async function (queryFn, queryOptions) {
+    const traceIdentifier =
```
adrien2p, line 153: todo: here you are manipulating the options like it was the config, maybe we need both as mentioned above :) also, the config is supposed to be unwrapped when it reaches the remote query

packages/core/modules-sdk/src/remote-query/query.ts
thetutlage: The config has flags like `throwIfUndefined`. I think, they are not useful for tracing or do you think otherwise?

adrien2p (approved):

PR #9137, opened 2024-09-15
chore: remove container registration name

carlos-r-l-rodrigues:
What:
* Deprecated `ModuleRegistrationName` in favor of `Modules` enum
* Modules are always registered in the container using their `key`
* Modules' key name normalized to PascalCase and `Service` suffix was removed (breaking change if module is resolved using strings)


adrien2p (approved):
LGTM, we might need a release note for that :D

sradevski (approved):
Really nice cleanup! Q: Are we also cleaning up the `ContainerRegistrationKeys`?

carlos-r-l-rodrigues:
> Really nice cleanup! Q: Are we also cleaning up the `ContainerRegistrationKeys`?

Yeah, I think we need. I haven't done in this PR because I dont have any good naming suggestion.

carlos-r-l-rodrigues:
@olivermrbl, let me know if we can merge this one.

olivermrbl:
Yes, all good

olivermrbl:
Maybe `ContainerRegistrationKeys` could be split into something like `Resources` and `Tools`:
```
const query = container.resolve(Tools.QUERY)
const logger = container.resolve(Tools.LOGGER)
const linker = container.resolve(Tools.REMOTE_LINK) // might want to rename this at some point?

const pgConnection = container.resolve(Resources.PG_CONNECTION)
const config = container.resolve(Resources.CONFIG)
```

Hard to find a single term that encapsulates all, because of how different these all are.

PR #9140, opened 2024-09-16
feat: add tracing to workflow steps

thetutlage:
Fixes: CC-444

Here's the full trace for creating a new product.

CleanShot 2024-09-16 at 12 25 55@2x


packages/medusa/src/loaders/index.ts
```diff
@@ -164,6 +163,8 @@ export default async ({
     expressApp,
     rootDirectory
   )
+
+  const { createDefaultsWorkflow } = await import("@medusajs/core-flows")
```
thetutlage, line 167: Importing `core-flows` package has a side-effect that it registers workflows and steps as soon as the package is imported.

Therefore, the instrumentation code is not able to apply tracing to core flows, since they are registered before the instrumentation code triggers.

sradevski, line 167: Maybe we need to revisit this decision of calling `createDefaultsWorkflow` when starting up the app, it is probably better suited to be run as part of `seed`. It is incomplete in what it creates anyway (eg. it doesn't create a shipping profile, which is required to create a shipping option).

cc @olivermrbl WDYT?

packages/medusa/src/instrumentation/index.ts
```diff
@@ -21,7 +22,7 @@ function shouldExcludeResource(resource: string) {
 }
 
 /**
- * Instrumenting the first touch point of the HTTP layer to report traces to
+ * Instrumen the first touch point of the HTTP layer to report traces to
```
sradevski:
```suggestion
 * Instrument the first touch point of the HTTP layer to report traces to
```

sradevski (approved):
Nice one!

packages/medusa/src/loaders/index.ts
olivermrbl: Yeah, think it should be fine to remove this in favor of a defaults seed script. Lemme create a ticket for this

adrien2p: yeah I dont think it should be part of this pr but yes agree 👍

adrien2p: this should be fixed with our last changes 👍

adrien2p (approved):
Maybe we can add some ts doc on the traces functions wdyt?

carlos-r-l-rodrigues:
Something is affecting the tests in this PR where it can't find the table.
As if it is using a different database other than the one created by `moduleIntegrationTestRunner`
It can be reproduced locally, on the integration tests of `workflow-engine-inmemory`.

PR #9149, opened 2024-09-16
fix(product): check if duplicate variant exists

fPolic:
**What**
- validate that the combination of options provided on variant create and update doesn't exist on another variant
- add check that variants has the same number of option values as product has options
- update tests

**Note**
- this covers only variant create/update, we need to add a similar check when creating and updating variants with products but this will be added in a separate PR

---

FIXES FRMW-2707


integration-tests/http/__tests__/product/admin/product.spec.ts
```diff
@@ -2641,33 +2641,6 @@ medusaIntegrationTestRunner({
           ).toEqual("Updated variant")
         })
 
-        it("removes options not present in update", async () => {
```
fPolic, line 2644: removed since a variant cannot have a subset of options set

packages/modules/product/src/services/product-module-service.ts
```diff
@@ -1759,4 +1801,35 @@ export default class ProductModuleService
...
+          variantData.product_id! !== v.product_id ||
+          !variantData.options?.length
+        ) {
+          return false
+        }
+
+        return (variantData.options as unknown as { id: string }[])!.every(
+          ({ id: optionValueId }) =>
+            !!v.options.find((vo) => vo.id === optionValueId)
+        )
```
olivermrbl: nit(non-blocking): Can we improve readability here, think we try to jam a lot of things into a one-liner here 😄

olivermrbl (approved):
LGTM

srindom:
@fPolic - why was this closed?

srindom:
NVM - see its in favor of https://github.com/medusajs/medusa/pull/9171

fPolic:
Yes moved everything to #9171

PR #9151, opened 2024-09-16
fix(link-modules): table name

carlos-r-l-rodrigues:
FIXES: FRMW-2706

### DIFF
--- a/packages/modules/link-modules/src/migration/index.ts
+++ b/packages/modules/link-modules/src/migration/index.ts
@@ -6,15 +6,15 @@ import {
   PlannerActionLinkDescriptor,
 } from "@medusajs/types"
 
-import { generateEntity } from "../utils"
-import { EntitySchema, MikroORM } from "@mikro-orm/core"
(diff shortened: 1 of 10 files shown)


packages/modules/link-modules/src/migration/index.ts
```diff
@@ -125,10 +125,10 @@ export class MigrationsExecutionPlanner implements ILinkMigrationsPlanner {
     `
         )
     )
-      .map(({ table_name }) => table_name)
+      .map(({ table_name }) => table_name.toLowerCase())
       .filter((tableName) =>
         this.#linksEntities.some(
-          ({ entity }) => entity.meta.collection === tableName
+          ({ entity }) => entity.meta.collection.toLowerCase() === tableName
```
olivermrbl: If the tables were initially created as lower case, I guess we don't need these

olivermrbl (approved):
LGTM, let's get this in

packages/modules/link-modules/src/migration/index.ts
carlos-r-l-rodrigues: true

PR #9158, opened 2024-09-17
feat(modules-sdk): Parse filters based on loaded modules graph

adrien2p:
> [!IMPORTANT]  
> Mikro orm has a bug when applying `Filters` to relations, it is not correctly propagated (see [issue](https://github.com/mikro-orm/mikro-orm/discussions/4840)) and the pr that fixes this problem is part of v6 [here](https://github.com/mikro-orm/mikro-orm/pull/5063/files#diff-a62cd78c400d99253a3c0fb5a9bf26e1e63e0deb511059d28ba51014b58242ddR2032). So for now it is a known limitation that we currently have. This issue has been discovered as part of that pr but is in fact already existing

**What**
Parse query filters and assign them correctly.

Along side having the filters of the Query API typed, it is now possible to filter cross modules (graphql like results) using this kind of typed filters

```ts
    await query.graph({
          entity: "product",
          fields: ["id", "title", "variants.*", "variants.prices.*"],
          filters: {
            id: "string",
            variants: {
              sku: {
                $eq: "string",
              },
              prices: {
                amount: {
                  $lt: 50,
                },
              },
            },
          }
        },
      )
```


carlos-r-l-rodrigues (approved):

olivermrbl (approved):
Nice work! 🔥 

Lots of stuff going on – think the easiest way to review it is to put it to use. With that in mind, I think we should merge it and get people to start playing around with it.

PR #9164, opened 2024-09-17
fix: migrate old links tables before performing a sync

thetutlage:
This PR addresses the issue as follows.

- Do not convert table names to lowercase in the code. Instead rely on `entity.meta.collection` to provide the name with correct casing.
- Migrate old tables, which had different name before #9151 to the new name.


carlos-r-l-rodrigues (approved):

adrien2p (approved):
Sounds good 👌

PR #9165, opened 2024-09-17
docs: updates to digital product recipe

shahednasser:
- Create workflow for fulfilling digital products that sends a notification + create a shipment to mark the fulfillment as fulfilled / shipped.
- Update the subscriber handling the `digital_product_order.created` event to use the workflow
- Add section on using the local notification module provider for testing purposes.

### DIFF
--- a/www/apps/resources/app/recipes/digital-products/examples/standard/page.mdx
+++ b/www/apps/resources/app/recipes/digital-products/examples/standard/page.mdx
@@ -1819,13 +1819,12 @@ In a later step, you’ll add an API route to allow customers to view and downlo
 
 ## Step 12: Fulfill Digital Order Workflow
 
-In this step, you'll create a workflow that fulfills a digital order. Later, you'll execute this workflow in a subscriber.
+In this step, you'll create a workflow that fulfills a digital order by sending a notification to the customer. Later, you'll execute this workflow in a subscriber.
 
 The workflow has the following steps:
 
 1. Retrieve the digital product order's details. For this, you'll use the `useRemoteQueryStep` imported from `@medusajs/core-flows`.
 2. Send a notification to the customer with the digital products to download.
-3. Create a shipment for the fulfillment to indicate that the item is fulfilled. For this step, you'll use from the `@medusajs/core-flows` package the `createOrderShipmentWorkflow` as a step.
 
 So, you only need to implement the second step.
 
@@ -1940,9 +1939,8 @@ You use the `createNotifications` method of the Notification Module's main servi
 Create the workflow in the file `src/workflows/fulfill-digital-order/index.ts`:
 
 export const fulfillWorkflowHighlights = [
-  ["18", "useRemoteQueryStep", "Retrieve the digital product order's details."],
-  ["27", "sendDigitalOrderNotificationStep", "Send a notification to the customer."],
-  ["41", "createOrderShipmentWorkflow", "Create a shipment for the order's fulfillment to mark as fulfilled."]
+  ["17", "useRemoteQueryStep", "Retrieve the digital product order's details."],
+  ["36", "sendDigitalOrderNotificationStep", "Send a notification to the customer."]
 ]
 
 ```ts title="src/workflows/fulfill-digital-order/index.ts" highlights={fulfillWorkflowHighlights} collapsibleLines="1-10" expandMoreLabel="Show Imports"
@@ -1951,8 +1949,7 @@ import {
   WorkflowResponse
 } from "@medusajs/workflows-sdk"
 import {
-  useRemoteQueryStep,
-  createOrderShipmentWorkflow
+  useRemoteQueryStep
 } from "@medusajs/core-flows"
 import { sendDigitalOrderNotificationStep } from "./steps/send-digital-order-notification"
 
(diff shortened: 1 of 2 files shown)


www/apps/resources/app/recipes/digital-products/examples/standard/page.mdx
```diff
@@ -1817,79 +1817,81 @@ In a later step, you’ll add an API route to allow customers to view and downlo
...
+The workflow has the following steps:
 
-export const subscriberHighlight = [
-  ["20", "notificationModuleService", "Resolve the Notification Module's service to use it later to send a notification."],
-  ["22", "fileModuleService", "Resolve the File Module's service to use it later to retrieve a media's URL."],
-  ["26", "query", "Run the query to retrieve the digital product order."]
-]
+1. Retrieve the digital product order's details. For this, you'll use the `useRemoteQueryStep` imported from `@medusajs/core-flows`.
+2. Send a notification to the customer with the digital products to download.
+3. Create a shipment for the fulfillment to indicate that the item is fulfilled. For this step, you'll use from the `@medusajs/core-flows` package the `createOrderShipmentWorkflow` as a step.
```
olivermrbl: thought: I think we should consider removing the third step here. Digital products are not shipped and should ideally be created with a `requires_shipping: false` option, which can be used in the storefront to build the appropriate checkout flow. The item processing flow for digital products should end after fulfillment.

420coupe:
Was the recipe changed otherwise?

Was having an issue when uploading, had to add multer on the api path for upload. Now i'm having an issue with ``req.validatedBody`` coming back as undefined. specifically this piece of the code below. 

```ts
type CreateRequestBody = z.infer<typeof createDigitalProductsSchema>;

export const POST = async (
  req: AuthenticatedMedusaRequest<CreateRequestBody>,
  res: MedusaResponse
) => {
  const { result } = await createDigitalProductWorkflow(req.scope).run({
    input: {
      digital_product: {
        name: req.validatedBody.name,
        medias: req.validatedBody.medias.map((media) => ({
          fileId: media.file_id,
          mimeType: media.mime_type,
          ...media,
        })) as Omit<CreateDigitalProductMediaInput, "digital_product_id">[],
      },
      product: req.validatedBody.product,
    },
  });

  res.json({
    digital_product: result.digital_product,
  });
};
```

Not sure why middleware.ts wasn't applying the multer, but Modified api to use multer, now uploads work, but it dies on verificationBody as mentioned above

```ts
import multer from "multer";
import { AuthenticatedMedusaRequest, MedusaResponse } from "@medusajs/medusa";
import { uploadFilesWorkflow } from "@medusajs/core-flows";
import { MedusaError } from "@medusajs/utils";

const upload = multer({ storage: multer.memoryStorage() });

export const uploadMiddleware = upload.array("files");

export const POST = async (
  res: MedusaResponse
) => {
  uploadMiddleware(req as any, res as any, async (err: any) => {
    if (err) {
      console.error("Error handling file upload:", err);
      return res.status(500).json({ error: "File upload failed" });
    }

    const access = req.params.type === "main" ? "private" : "public";
    const input = req.files as Express.Multer.File[];

    if (!input?.length) {
      throw new MedusaError(
        "No files were uploaded"
      );
    }

    const { result } = await uploadFilesWorkflow(req.scope).run({
      input: {
        files: input?.map((f) => ({
          filename: f.originalname,
          mimeType: f.mimetype,
          content: f.buffer.toString("binary"),
          access,
        })),
      },
    });

    res.status(200).json({ files: result });
  });
};
````

www/apps/resources/app/recipes/digital-products/examples/standard/page.mdx
shahednasser: @olivermrbl ah makes sense, where would `requires_shipping: false` be passed? when the shipping method is set on the cart?

shahednasser:
> Was having an issue when uploading, had to add multer on the api path for upload. 

@420coupe This is already part of the recipe.

> Not sure why middleware.ts wasn't applying the multer, but Modified api to use multer, now uploads work, but it dies on verificationBody as mentioned above

there was an issue with middlewares in a preview version we released yesterday. Can you update and try again?

www/apps/resources/app/recipes/digital-products/examples/standard/page.mdx
shahednasser: I just checked and that's the property of an inventory item, but digital products don't have inventory items (unless that's the better approach?)

420coupe:
> > Was having an issue when uploading, had to add multer on the api path for upload.
> 
> @420coupe This is already part of the recipe.
> 
> > Not sure why middleware.ts wasn't applying the multer, but Modified api to use multer, now uploads work, but it dies on verificationBody as mentioned above
> 
> there was an issue with middlewares in a preview version we released yesterday. Can you update and try again?

Tyvm, been chasing my tail all day trying to figure out why. 


found another small bug, when deleting a digital product from the product section, it doesn't delete the digital product and then causes an error when going to digital products on adminui

420coupe:
also you never attach the digital product in the notification.

shahednasser:
> also you never attach the digital product in the notification.

@420coupe  We instead retrieve the URLs of all digital products and pass them in the notification's payload. You're free to do it differently if that's your use case.

www/apps/resources/app/recipes/digital-products/examples/standard/page.mdx
olivermrbl: > but digital products don't have inventory items

Hmm, I guess that depends on the case. For example, you might have a limited amount of licenses. Conversely, it is also very likely that you have an unlimited amount of licenses 😄 

For the recipe, my take would be that we should not have the shipment step. This is likely also how we will build the digital products starter (when we get around to it).

shahednasser: Makes sense! Will remove it then

olivermrbl (approved):
LGTM, nice!

PR #9171, opened 2024-09-18
feat(core-flows, product): options checks on product create/update

fPolic:
**What**
- validate that variants are unique with respect to options on product update/create and variant update/create
- validate that the product has options upon creation
- ensure variants have the same number of option values as the product has options
- admin error handling
- update tests

---

FIXES FRMW-2707 CC-556


carlos-r-l-rodrigues (approved):

packages/core/core-flows/src/product/workflows/create-products.ts
```diff
@@ -4,19 +4,48 @@ import {
...
+    for (const product of products) {
+      if (!product.options?.length) {
+        throw new MedusaError(
+          MedusaError.Types.INVALID_DATA,
+          `Product options are not provided for ${product.title}`
+        )
+      }
+    }
+  }
+)
```
adrien2p:
```suggestion
const missingOptionsProductTitles = products
  .filter((product) => !product.options?.length)
  .map((product) => product.title)

if (missingOptionsProductTitles.length) {
  throw new MedusaError(
    MedusaError.Types.INVALID_DATA,
    `Product options are not provided for [${missingOptionsProductTitles.join(
      ", "
    )}]`
  )
}
```

adrien2p: it would prevent having many products with wrong options and only receiving the product error one by one :)

packages/modules/product/integration-tests/__tests__/product-module-service/product-variants.spec.ts
```diff
@@ -401,11 +373,131 @@ moduleIntegrationTestRunner<IProductModuleService>({
...
+                    value: "yellow",
+                  }),
                 ]),
               }),
             ])
           )
         })
+
+        it("should throw if there is an existing variant with same options combination", async () => {
+          jest.clearAllMocks()
```
adrien2p: todo: this should be in an afterEach wdyt?

packages/modules/product/integration-tests/__tests__/product-module-service/product-variants.spec.ts
```diff
@@ -401,11 +373,131 @@ moduleIntegrationTestRunner<IProductModuleService>({
...
+          try {
+            await service.updateProductVariants(
+              productFour.variants.find((v) => v.title === "new variant 2")!.id,
+              {
+                options: { size: "small", color: "red" },
+              } as UpdateProductVariantDTO
+            )
+          } catch (e) {
+            error = e
+          }
```
adrien2p:
```suggestion
         const error = await service.updateProductVariants(
  productFour.variants.find((v) => v.title === "new variant 2")!.id,
  {
    options: { size: "small", color: "red" },
  } as UpdateProductVariantDTO
).catch((err) => err);
```

we usually just do this :)

packages/modules/product/src/services/product-module-service.ts
```diff
@@ -209,9 +208,25 @@ export default class ProductModuleService
       sharedContext
     )
 
+    const variants = await this.productVariantService_.list(
+      {
+        product_id: [...new Set<string>(data.map((v) => v.product_id!))],
+      },
+      {
+        take: null,
```
adrien2p: todo: this is not needed

packages/modules/product/src/services/product-module-service.ts
```diff
@@ -328,6 +343,13 @@ export default class ProductModuleService
       {},
       sharedContext
     )
+
+    const allVariants = await this.productVariantService_.list(
+      { product_id: variants.map((v) => v.product_id) },
+      { take: null, relations: ["options"] },
```
adrien2p: todo: rm take: null everywhere :)

packages/modules/product/src/services/product-module-service.ts
```diff
@@ -1604,6 +1642,31 @@ export default class ProductModuleService
...
+        options.forEach((option) => {
+          if (!variant.options?.[option.title]) {
+            throw new MedusaError(
+              MedusaError.Types.INVALID_DATA,
+              `Variant "${variant.title}" doesn't have "${option.title}" option provided.`
+            )
+          }
+        })
+      })
+    }
```
```suggestion
if (options?.length) {
  const missingOptionsVariantTitles = productData.variants
    .filter((variant) => {
      return !options.every((option) => variant.options?.[option.title])
    })
    .map((variant) => variant.title)
  
  if (missingOptionsVariantTitles.length) {
    throw new MedusaError(
      `Product ${productData.title} has variants with missing options: [${missingOptionsVariantTitles.join(
        ", "
      )}]`
    )
  }
}

for the same reason as the previous one
```

PR #9172, opened 2024-09-18
fix(utils): wrong initialization of collection

adrien2p:
**What**
When a collection needs to be initialized, use the classic init function instead of trying to improve a little bit this area in favor of something that we know work all the time


thetutlage:
Was this not relevant?

adrien2p:
I think something happened in my pr local changes in the transpiled files in mikro orm when I tried to backport the fix from v6 cause there is in fact no issues 😂

PR #9177, opened 2024-09-18
fix(workflow-sdk): Use the correct event group ID from the flow metadata

adrien2p:
**What**
Fix wrong event group id used in the on finish wrapper when releasing events from workflow


adrien2p:
@sradevski this is the proper fix, I can explain you why

sradevski (approved):

PR #9179, opened 2024-09-18
chore(): Handle medusa service base methods events

adrien2p:
**What**
Properly handle medusa service event for non overridden methods. Also, provide useful utils to centralise event name building as well as message aggregation with inference from the medusa service itself. Accessible from derived service.


carlos-r-l-rodrigues (approved):

adrien2p:
@olivermrbl do you want to throw an eye?

olivermrbl (approved):

PR #9182, opened 2024-09-18
docs: improvements + additions to workflow docs

shahednasser:
Improve existing workflow docs + add missing docs


www/apps/book/app/advanced-development/api-routes/additional-data/page.mdx
```diff
@@ -0,0 +1,164 @@
+export const metadata = {
```
olivermrbl, line 3: thought: Given we currently have so few API Routes that accept additional data, I think we should do one of two things:
1. Add the list of all API Routes and reference it here
2. Hide this for the time being

The reason I suggest 2. is that we need admin extensions before this is really useful.

olivermrbl (approved):
Nice work, added a single comment but otherwise LGTM

www/apps/book/app/advanced-development/api-routes/additional-data/page.mdx
shahednasser: > Add the list of all API Routes and reference it here

Cool idea, I'll do this manually for now but probably worth looking into generating this info (can look into it at another point)

> Hide this documentation for the time being

We already use this in the new Customizations chapter 😄 What do you think of adding a note that says you'll be benefit more from this once we add admin extensions soon? (or do we not want to mention it?) Either way, I believe it's generally fine to keep.

PR #9189, opened 2024-09-19
fix(types): fixes to HTTP types

shahednasser:
- Change `countries` in `AdminRegion` to be of type `AdminRegionCountry`
- Change `inventory_item`'s type in `AdminReturn` to `AdminInventoryItem`
- Use `ReturnStatus` enum as type of `status` in `BaseReturn`.
- Add `@ignore` tsdocs to FE-related fields in the `BaseRuleAttributeOptions` type since we don't want to show it in the API reference
- Change the `operator` type in `AdminShippingOptionRule`
- Remove `id` from `BaseRuleValueOptions` because the API route doesn't return it
- Remove incorrect properties of a tax region

### DIFF
--- a/packages/core/types/src/http/promotion/common.ts
+++ b/packages/core/types/src/http/promotion/common.ts
@@ -50,9 +50,22 @@ export interface BaseRuleAttributeOptions {
   id: string
   value: string
   label: string
+  /**
+   * @ignore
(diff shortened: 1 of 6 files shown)


olivermrbl (approved):
neat

PR #9191, opened 2024-09-19
feat(order,utils): adds a type column to order

riqwan:
what:

- adds a type column to order to differentiate different kinds of orders that can be generated - eg: draft orders


olivermrbl (commented):
We already have these two columns on the order model: 
```
is_draft_order: boolean = false
status: OrderStatus = "draft"
```

Would it be possible to reuse those instead of adding a new type?

If not, we should probably consider removing at least one of them.

riqwan:
Ah my bad, didn't see that one. Those should be good enough for the quote use-case. Will close this 🍻

PR #9192, opened 2024-09-19
fix(medusa): list shipping option rules on store endpoint

fPolic:
**What**
- return shipping option rules when listing cart options


olivermrbl (approved):
LGTM, left a comment on the PR to the Next.js starter

PR #9193, opened 2024-09-19
feat(index): Provide a similar API to Query

adrien2p:
**What**
Align the index engine API to be similar to the Query API

## Example

```ts
        // Benefit from the same level of typing like the remote query

        const { data, metadata } = await indexEngine.query<'product'>({
          fields: [
            "product.*",
            "product.variants.*",
            "product.variants.prices.*",
          ],
          filters: {
            product: {
              variants: {
                prices: {
                  amount: { $gt: 50 },
                },
              },
            },
          },
          pagination: {
            order: {
              product: {
                variants: {
                  prices: {
                    amount: "DESC",
                  },
                },
              },
            },
          },
        })
```


packages/modules/index/src/utils/flatten-object-keys.ts
```diff
@@ -0,0 +1,59 @@
...
+ * @param where
+ */
+export function flattenObjectKeys(where: Record<string, any>) {
+  const result: Record<string, any> = {}
+
+  function flatten(obj: Record<string, any>, path?: string) {
+    for (const key in obj) {
+      const isOperatorMap = !!OPERATOR_MAP[key]
+
+      if (!isOperatorMap) {
```
carlos-r-l-rodrigues:
```suggestion
      if (isOperatorMap) {
        result[path ?? key] = obj
        continur
      }
      
      // ...
```

packages/core/types/src/index/query-config/query-input-config.ts
```diff
@@ -0,0 +1,22 @@
...
+    ? string[]
+    : ObjectToIndexFields<
+        IndexServiceEntryPoints[TEntry & keyof IndexServiceEntryPoints]
+      >[]
+  filters?: IndexFilters<TEntry>
+  joinFilters?: IndexFilters<TEntry>
+  pagination?: {
+    skip?: number
+    take?: number
+    orderBy?: IndexOrderBy<TEntry>
```
olivermrbl: question: Wasn't this just `order` in `remoteQuery`?

adrien2p: my bad, let me chnge that

packages/core/types/src/index/service.ts
```diff
@@ -1,6 +1,11 @@
 import { IModuleService } from "../modules-sdk"
+import { IndexQueryConfig } from "./query-config"
 
 export interface IIndexService extends IModuleService {
-  query(...args): Promise<any>
-  queryAndCount(...args): Promise<any>
+  query<const TEntry extends string>(
+    config: IndexQueryConfig<TEntry>
+  ): Promise<any[]>
+  queryAndCount<const TEntry extends string, Test = IndexQueryConfig<TEntry>>(
```
olivermrbl: question: Can we kill this `queryAndCount` and return the same shape as `query.graph`?

```
const { data, metadata } = ...
``` 

Count would be part of metadata. Wdyt?

adrien2p: Yes we can def do that 👍 i dont see any issues

packages/modules/index/package.json
```diff
@@ -23,9 +23,9 @@
   "scripts": {
     "watch": "tsc --build --watch",
     "watch:test": "tsc --build tsconfig.spec.json --watch",
-    "build": "rimraf dist && tsc --noEmit && tsc -p ./tsconfig.build.json && tsc-alias -p ./tsconfig.build.json",
+    "build": "rimraf dist && tsc -p ./tsconfig.spec.json --noEmit && tsc -p ./tsconfig.build.json && tsc-alias -p ./tsconfig.build.json",
     "test": "jest --passWithNoTests ./src",
-    "test:integration": "jest --runInBand --forceExit -- integration-tests/**/__tests__/**/*.ts",
+    "test:integration": "jest --runInBand --forceExit -- integration-tests/__tests__/index-engine-module.spec.ts",
```
olivermrbl: todo: rm

packages/modules/index/src/services/index-module-service.ts
```diff
@@ -93,20 +94,23 @@ export default class IndexModuleService implements IndexTypes.IIndexService {
...
+  ) {
+    return await this.storageProvider_.query(config)
   }
 
-  async queryAndCount(...args) {
-    return await this.storageProvider_.queryAndCount.apply(
-      this.storageProvider_,
-      args
-    )
+  async queryAndCount<const TEntry extends string>(
```
olivermrbl: todo(if-possible): Remove in favor of unified return shape as discussed above

packages/modules/index/src/types/index.ts
```diff
@@ -119,9 +120,13 @@ export interface StorageProvider {
 
   onApplicationStart?(): Promise<void>
 
-  query(...args): unknown
+  query<const TEntry extends string>(
+    config: IndexTypes.IndexQueryConfig<TEntry>
+  ): Promise<any[]>
 
-  queryAndCount(...args): unknown
+  queryAndCount<const TEntry extends string>(
```
olivermrbl: same

olivermrbl (approved):
Nice work, few comments 💪

packages/modules/index/src/services/postgres-provider.ts
```diff
@@ -353,6 +227,91 @@ export class PostgresProvider {
...
+    const { take, skip, order: inputOrderBy = {} } = config.pagination ?? {}
+
+    const select = normalizeFieldsSelection(fields)
+    const where = flattenObjectKeys(filters)
+    const joinWhere = flattenObjectKeys(joinFilters)
+    const orderBy = flattenObjectKeys(inputOrderBy)
+
+    const { manager } = sharedContext as { manager: SqlEntityManager }
+    let hasPagination = false
+    if (typeof take === "number" || typeof skip === "number") {
```
carlos-r-l-rodrigues: I believe we can have a `take` that is not a pagination.
I would only consider to paginate if we have `skip`.

carlos-r-l-rodrigues (approved):
🔥 🔥 🔥

PR #9205, opened 2024-09-19
chore(orchestrator): remote joiner using entitymap

carlos-r-l-rodrigues:
What:
Entity discovery using the entity map from graphql schema
This enables the same module having multiple links to another module, using the same alias in sub entities.

ex: product.translation, product.variant.translation, product.categories.translation


adrien2p (approved):
LGTM, one minor suggestion, can we make the different functions take an object as argument with the optional one being optional. That way it is named and we dont have to provide the undefined explicitly as well. Wdyt?

Later on, it can also be an improvement base where we always pass object with the same shape (property orders) no matter the values and we can benefit from v8 function optimisation in the retrieval path. But we can discuss that later ☺️

carlos-r-l-rodrigues:
> LGTM, one minor suggestion, 

yeah, it's time to overhaul this entire class.

adrien2p:
> > LGTM, one minor suggestion, 
> 
> 
> 
> yeah, it's time to overhaul this entire class.

We will get to that phase 😉

carlos-r-l-rodrigues:
> We will get to that phase 😉

i've changed most function parameters to objects. That will help better organize things in the future.

PR #9211, opened 2024-09-20
fix(framework): throw when middleware fails to register

riqwan:
what:

When middlewares fails to get registered, we swallow the error and skip registering the middlewares. This PR changes it to throw instead. 

An error I encountered - I imported a function from an invalid path. Instead of the server throwing an error, the routes continued to be functional without going through the middleware.

As I did not have a logger setup, I didn't get the error log message that something went wrong. 

Middlewares are pretty important piece of the route, I'd expect it to make a hard fail when it fails. wdyt?

Error after removing catch:
Screenshot 2024-09-20 at 14 55 36

### DIFF
--- a/packages/framework/framework/src/http/router.ts
+++ b/packages/framework/framework/src/http/router.ts
@@ -10,6 +10,9 @@ import {
 } from "express"
 import { readdir } from "fs/promises"
 import { extname, join, parse, sep } from "path"
+import { configManager } from "../config"
+import { logger } from "../logger"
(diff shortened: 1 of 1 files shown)


olivermrbl (approved):
I think this makes sense I can't immediately think of any reasons for not throwing. 

Curios if others have opinions on this.

carlos-r-l-rodrigues (approved):

adrien2p (approved):
Makes total sense 🙌

PR #9231, opened 2024-09-22
fix: Idempotent cart completion

olivermrbl:
What
- Store result of cart-completion workflow for three days by default
  - This enables the built-in idempotency mechanism to kick-in, provided the same transaction ID is used on workflow executions
- Return order from cart-completion workflow if the cart has already been completed
  - In case transaction ID is not used on workflow executions, we still only want to complete a cart once


packages/core/core-flows/src/cart/workflows/complete-cart.ts
```diff
@@ -34,36 +31,54 @@ export type CompleteCartWorkflowInput = {
...
 
+export const THREE_DAYS = 60 * 60 * 24 * 3
+
 export const completeCartWorkflowId = "complete-cart"
 /**
  * This workflow completes a cart.
  */
 export const completeCartWorkflow = createWorkflow(
-  completeCartWorkflowId,
+  {
```
olivermrbl, line 48: The only things that have changed in this file are:
- Workflow configuration
- Check if order exists before completing cart

The rest is indentation as a result of the `when-then` addition

carlos-r-l-rodrigues (approved):
Once the option of running workflows with transaction_id + idempotency_key is available, this on can be reverted.

olivermrbl:
> Once the option of running workflows with transaction_id + idempotency_key is available, this on can be reverted.

The workflow configuration you mean? I think the check for order existence still makes sense, as you could have developers consume the workflow without passing the transaction_id + idempotency_key.

carlos-r-l-rodrigues:
> The workflow configuration you mean? I think the check for order existence still makes sense, as you could have developers consume the workflow without passing the transaction_id + idempotency_key.

The `transaction_id` will always return the last execution. This is the behavior you want. Only in case something has failed and you want to force a new execution using the same transaction_id, you would provide an `idempotency_key` + `transaction_id`

olivermrbl:
> > The workflow configuration you mean? I think the check for order existence still makes sense, as you could have developers consume the workflow without passing the transaction_id + idempotency_key.
> 
> 
> 
> The `transaction_id` will always return the last execution. This is the behavior you want. Only in case something has failed and you want to force a new execution using the same transaction_id, you would provide an `idempotency_key` + `transaction_id`
> 
> 

Yes, I understand. But what I mean is, in the case where the transaction_id is not provided upon executing the workflow, you would still want to return early if the cart has already been completed right?

carlos-r-l-rodrigues:
> Yes, I understand. But what I mean is, in the case where the transaction_id is not provided upon executing the workflow, you would still want to return early if the cart has already been completed right?

That defeats a bit the the purpose of the workflow, and at the end the concept becomes the `when`. However to me the functionality in this case, the transaction_id used must be the cart_id, to guarantee that this is executed only once, and in case this has to be explicitly retried, another idempotency_key should be provided.
Does that make sense?

olivermrbl:
I think we are talking past each other. 

I understand that ideally you would pass the `cart_id` as the `transaction_id` upon executing the workflow, but we can't guarantee that right? 

For example, someone can import it and use it in their own subscriber.

carlos-r-l-rodrigues:
yeah, and in that case this check should be performed elsewhere, but that's alright.
we can keep it like this.

adrien2p (approved):

PR #9234, opened 2024-09-23
feat: Cleanup utils package

thetutlage:
Fixes: FRMW-2712


packages/core/utils/package.json
```diff
@@ -3,35 +3,40 @@
   "version": "1.11.9",
   "description": "Medusa utilities functions shared by Medusa core and Modules",
   "main": "dist/index.js",
-  "types": "dist/index.d.ts",
+  "exports": {
+    ".": "./dist/index.js"
+  },
```
thetutlage, line 8: Adding exports everywhere, later we can try to add more subpaths.

packages/core/utils/package.json
```diff
@@ -3,35 +3,40 @@
...
   "repository": {
     "type": "git",
     "url": "https://github.com/medusajs/medusa",
     "directory": "packages/utils"
   },
   "engines": {
-    "node": ">=16"
+    "node": ">=20.0.0"
```
thetutlage, line 15: Targeting Node.js >= `20.0.0` in all the packages

packages/core/utils/package.json
```diff
@@ -3,35 +3,40 @@
...
   },
   "publishConfig": {
     "access": "public"
   },
   "files": [
-    "dist"
+    "dist",
+    "!dist/**/__tests__",
+    "!dist/**/__mocks__",
+    "!dist/**/__fixtures__"
```
thetutlage, line 24: Now that I am type checking the tests as well using the TypeScript compiler they will be bundled in the output.

However, we can ignore them from the publish list of `files` that makes their way to npm.

Another alternative is to have a different `tsconfig.json` file for the tests. However, the only downside of doing that is the code editors not reading that file.

packages/core/utils/package.json
```diff
@@ -3,35 +3,40 @@
...
-    "expect-type": "^0.19.0",
-    "express": "^4.18.2",
+    "expect-type": "^0.20.0",
+    "express": "^4.21.0",
     "jest": "^29.7.0",
     "rimraf": "^5.0.1",
-    "ts-jest": "^29.1.1",
-    "typescript": "^5.1.6"
+    "ts-jest": "^29.2.5",
+    "typescript": "^5.6.2"
```
thetutlage, line 36: Going to use the latest TypeScript version everywhere.

packages/core/utils/package.json
```diff
@@ -3,35 +3,40 @@
...
   },
   "dependencies": {
-    "@medusajs/types": "^1.11.16",
+    "@medusajs/types": "workspace:^",
```
thetutlage, line 39: Monorepo packages should be linked via `workspace` identifier only.

packages/core/utils/src/common/__tests__/partition-array.spec.ts
```diff
@@ -1,4 +1,4 @@
-import { partitionArray } from "../../../dist"
+import { partitionArray } from "../../index"
```
thetutlage, line 1: The source code imports should not look for files in the builder output. This was caught after type checking test files

packages/core/utils/src/common/build-query.ts
```diff
@@ -42,7 +41,7 @@ function buildRelationsOrSelect(collection: string[]): Selects | Relations {
  * }
  * @param orderBy
  */
-export function buildOrder<T>(orderBy: { [k: string]: "ASC" | "DESC" }): Order {
+export function buildOrder(orderBy: { [k: string]: "ASC" | "DESC" }): Order {
```
thetutlage, line 44: The generic was not used.

packages/core/utils/src/event-bus/__tests__/message-aggregator.spec.ts
```diff
@@ -18,6 +19,7 @@ describe("MessageAggregator", function () {
       data: { id: 999 },
     })
     aggregator.save({
+      // @ts-expect-error
```
thetutlage: As per types, there is no property called `eventName` in `Event` type. For now I have suppressed the type errors. But I think we should check why there is a type mismatch.

packages/core/utils/tsconfig.json
```diff
@@ -5,24 +5,24 @@
...
     "experimentalDecorators": true,
     "sourceMap": true,
     "noImplicitReturns": true,
+    "forceConsistentCasingInFileNames": true,
     "strictNullChecks": true,
     "strictFunctionTypes": true,
     "noImplicitThis": true,
     "allowJs": true,
-    "skipLibCheck": true,
-    "downlevelIteration": true
```
thetutlage, line 20: Removed `downlevelIteration` since we do not want to compile code to oldest versions of JavaScript engines. 

The `tsconfig.json` file will output code in `ES2021` format, which are supported Node 20. Infact, Node 20 supports many more latest features of ES.

https://node.green/

olivermrbl:
@thetutlage, nit: can I get you to use `fix/`, `chore/`, or `/feat` as branch prefix? Our labels are automatically added based on the branch name

thetutlage:
@olivermrbl Sure will do! It's just I have a habit of using `refactor` for changes that doesn't impact the public usage. Should I use `feat` for these sort of changes as well?

olivermrbl:
Hmm, yeah feat or chore would be fine here, I think

thetutlage:
New PR with correct branch name. https://github.com/medusajs/medusa/pull/9238

PR #9243, opened 2024-09-23
fix(user): check if user account with email already exist on invite create

fPolic:
**What**
- when creating an invite, validate that provided email is not already used for existing user account

---

FIXES CC-513


olivermrbl (approved):
LGTM

PR #9254, opened 2024-09-23
feat(utils): define read only link

carlos-r-l-rodrigues:


adrien2p (approved):

SalahAdDin:
I don't get, what is this for?

carlos-r-l-rodrigues:
> I don't get, what is this for?

read-only links can link modules when there isn't a pivot table involved.

Linking your CustomModule with Product Variant:
```typescript
defineReadOnlyLink(
  { serviceName: "CustomModule" },
  "variant",
  "variant_id",
  ProductModule.linkable.productVariant.id
);
```

now, when there is a `variant_id` property, you can ask to resolve the property `variant`, linked to the Product module.

suppose your `CustomModule`'s method `list` returns:
```
[
  {
    id: "123",
    something: "custom stuff",
    metadata: {
      deep: {
        custom: "stuff",
        variant_id: "variant_01J742X0S8S1FMMDQN0V7312SB",
      }
    }
  }
];
```

you can query to expand the product variant like this:
```typescript
 await query({
  entryPoint: "custom_module",
  fields: [
    "id",
    "something",
    "metadata.deep.variant.*",
  ],
});
```

SalahAdDin:
And you will be unable to edit such a link, right?

packages/core/utils/src/modules-sdk/define-link.ts
```diff
@@ -375,3 +427,74 @@ ${serviceBObj.module}: {
...
+ * @param rightService
+ * @param readOnlyLinkOptions
+ */
+export function defineReadOnlyLink(
+  leftService: DefineReadOnlyLinkInputSource,
+  property: string,
+  foreignKey: string,
+  rightService: DefineReadOnlyLinkInputSource,
+  readOnlyLinkOptions?: ReadOnlyOptions
+): void {
```
olivermrbl: thought: 

I have two suggestions to this:
- We should consider passing an object to make the purpose of each argument more explicit
- We should try to get as close to the `defineLink` API as possible

The latter more important than the former.

Perhaps, something like the following:
```
defineReadOnlyLink(
  {
    linkable: CustomModule.linkable.customModel,
    field: "variant_id"
  },
  ProductModule.linkable.productVariant,
)
```
```
defineLink(
  {
    linkable: CustomModule.linkable.customModel,
    field: "variant_id"
  },
  linkable: ProductModule.linkable.variant,
  {
    readOnly: true,
  }
)
```

Or something like that. We can discuss a more specific API, but just wanted to hear your thoughts on this first.

carlos-r-l-rodrigues: I think we can use `defineLink` and the option `readOnly: true`.
Note that "linkable" might not exist in 1 or both sides, so the objects can be passed and what will be required is that the first argument requires `field`, and the second `alias`.

something like this:

```typescript
defineLink(
  {
    serviceName: "CustomModule",
    field: "variant_id",
  },
  ProductModule.linkable.productVariant.id,
  {
    readOnly: true,
  }
);

defineLink(
  {
    serviceName: "CustomModule",
    field: "variant_id",
  },
  {
    ...ProductModule.linkable.productVariant.id,
    alias: "my_variant",
  },
  {
    readOnly: true,
  }
);
```

olivermrbl: How do you know what data model from `CustomModule` the read only link applies to here?

```
defineLink(
  {
    serviceName: "CustomModule",
    field: "variant_id",
  }
  ...
```

Let's say it contains both a:
- `CustomDataModel`
- `AnotherCustomDataModel`

One or both of them have a `variant_id`.

carlos-r-l-rodrigues: when no entity is provided, it will always default to the method `.list` from the module. If one specific data model is the target, the `entity` has to be set with the `serviceName`. Using the DML and `linkable`, this is transparent, but Custom modules can be anything, not only a module created with DML.

olivermrbl: > when no entity is provided, it will always default to the method .list from the module.

Hmm I see, tbh. I still prefer explicitness in this scenario, so you would have to specify the exact method intended to be used, e.g. `list` or `listSomething` instead of just the module name.

After all, in all our models, there will be no such thing as a default `list` method. They will all be suffixed with the data model name. 

Anyway, as long as some variation of what I suggested initially is possible, we can always decide how we want to document your described type of usage in our documentation when appropriate.

Right now, you should be able to do:
```
defineLink(
  {
    linkable: CustomModule.myModel,
    field: "variant_id",
  },
  ProductModule.linkable.productVariant.id,
  {
    readOnly: true,
  }
);
```

Wdyt?

carlos-r-l-rodrigues: This example is supported.
The first argument the "FROM" argument. You can link your module, which would consider any entity that contains `field` to resolve, or you specify the entity of your module.

The second argument is the "TO" argument, where you can also point to the module itself, or to an specific entity of the module.

We have to support this in the internals, regardless of what is exposed as helpers. I don't see the "linkable" property existing for external modules, so it will require other approaches.

olivermrbl: > This example is supported.

Great, then this LGTM!

olivermrbl (approved):

Issue #9247, opened 2024-09-23
InventoryService takes only first 15 items


noubase:
## Bug report

InventoryService takes only first 15 items

### Describe the bug

If you query the InventoryService and expect more than 15 records, you will only receive the first 15.
This behavior causes an issue in the `ensureInventoryLevels` method within the `InventoryModuleService`, as it incorrectly calculates the `missing` array.

### System information

Medusa version (including plugins): 2.0.9-preview
Node.js version: v20.15.1
Database: PostgreSQL 16.3
Operating system: MacOS Version 14.6.1
Browser (if relevant):

### Steps to reproduce the behavior

```
const sku = variants.map(v => v.sku); // 53 elements
const inventoryItems = await inventoryService.listInventoryItems({
  sku
}, {take: sku.length});

const inventoryItemsWithoutTake = await inventoryService.listInventoryItems({
  sku
});
```

### Expected behavior

The `inventoryItems` array has a length of 53, and `inventoryItemsWithoutTake` should have the same length. However, it currently contains only the first 15 elements.

noubase:
`packages/modules/inventory-next/src/services/inventory-module.ts`

lines 120 ... 128

```
const inventoryLevels = await this.listInventoryLevels(
      {
        $or: [
          { id: idData.filter(({ id }) => !!id).map((e) => e.id) },
          ...itemLocationData,
        ],
      },
      {},
      context
    )
```    
In this case, `itemLocationData` contains 53 elements, but `inventoryLevels` still only includes the first 15.

noubase:
![image](https://github.com/user-attachments/assets/e2086111-fd5c-417f-a3b0-40f389f49bd1)

`packages/core/utils/src/modules-sdk/build-query.ts`

As I understand it, this line is causing the issue. 
While I can control how many items I want to retrieve in my code, the `ensureInventoryLevels` method will still block me regardless.

adrien2p:
@noubase thanks for reporting it, I ve open a pr that fixes this hidden constraint and many other issues. You don't have to do anything on your side and when it will be released you should be good 👍

PR #9257, opened 2024-09-23
chore(): Remove default limit from the build query

adrien2p:
TRI-281

**What**
Remove the repository layer default limit which lead to miss understanding and issues when working with deep relations etc. Instead, rely on higher level default limit provided to the module (e.g the default limit in our API).


olivermrbl:
Just adding @srindom's [comment](https://linear.app/medusajs/issue/TRI-281/inventoryservice-takes-only-first-15-items#comment-b19994ed), so we won't proceed without addressing it

adrien2p:
> Just adding @srindom's [comment](https://linear.app/medusajs/issue/TRI-281/inventoryservice-takes-only-first-15-items#comment-b19994ed), so we won't proceed without addressing it

@olivermrbl I dont understand, this pr addresses this issues as well many other issues it creates

olivermrbl:
> > Just adding @srindom's [comment](https://linear.app/medusajs/issue/TRI-281/inventoryservice-takes-only-first-15-items#comment-b19994ed), so we won't proceed without addressing it
> 
> @olivermrbl I dont understand, this pr addresses this issues as well many other issues it creates

What I was trying to say was I just want to make sure we address the comment before moving forward

adrien2p:
@olivermrbl it should be good now :)

carlos-r-l-rodrigues (approved):

olivermrbl:
@adrien2p, will just give @srindom a chance to get back to your comment in Linear :)

adrien2p:
sure @olivermrbl just to add a bit more context to it.

We currently have the default limit managed through the API and sent to the module query and then to the modules. Currently, the issue is that we have hidden conditions at the repository level to also manage a default limit, those conditions built up in  with time and became hard to use and are completely hidden. When consuming a module or a query or a repository inside a module, you always have to pass take: null in order to not have this default repository limit applied. This has caused many issues during the modules implementation and keeps coming back.

The main idea here is to lighten this repository layer, it does not manage any default limit and give back this behaviour to the consumer. In our case the default limit that we already apply in our end points for example

srindom:
LGTM 👍

srindom:
Can you update https://github.com/medusajs/medusa/issues/9247 to explain what should be done (if anything)

adrien2p:
@srindom done I ve informed him, also he doesnt have to do anything on his side

PR #9271, opened 2024-09-24
chore: Convert to TS

olivermrbl:
Build-output was missing TS-files, which has resulted in "corrupted" published packages.

Fixing this now to get a new version out.


shahednasser (approved):
LGTM

PR #9273, opened 2024-09-24
fix(types, medusa): fixes to HTTP types

shahednasser:
### types

- Remove `is_internal` and `is_active` from `StoreProductCategory`
- Add missing status, reference, and action enums from `BaseOrderChange` and `BaseOrderChangeAction`
- Remove unused totals from `BaseOrderSummary`.
- Fix references to base types from admin / store types
- Other general fixes

### medusa

- Use HTTP types for create / update return reason routes
- Remove type arguments from the calculate taxes route since it's a `POST` API route but the type argument for the select query parameters only.
- Add `metadata` to the list of cart's fields.
- Add type casting for order previews in response


carlos-r-l-rodrigues (approved):
🥇

PR #9275, opened 2024-09-24
fix(orchestration): Throw if not exists using filters

adrien2p:
**What**
Search for filters in the args.filters if not found at the top level args.


packages/core/orchestration/src/joiner/remote-joiner.ts
```diff
@@ -1239,3 +1233,59 @@ export class RemoteJoiner {
...
+    const filters = (primaryKeyArg = queryObj.args?.find(
+      (arg) => arg.name === "filters"
+    )?.value)
+    if (filters) {
+      const primaryKeyFilter = Object.keys(filters).find((key) => {
+        return serviceConfig.primaryKeys.includes(key)
+      })
+
+      if (primaryKeyFilter) {
+        console.log(primaryKeyFilter, filters)
```
carlos-r-l-rodrigues:
```suggestion
```

carlos-r-l-rodrigues (approved):

olivermrbl:
I feel like this is a bit contradictory with always returning the same shape regardless of input:
```
const { data, metadata } = await ...
```

I want to challenge if these options even make sense any more. I would have expected the query to return an empty array instead of throwing. Similar to GQL and our list methods.

Of cause it is convenient, but I don't mind forcing using to always have to check for the empty result set.

Curious to hear what you think.

adrien2p:
> I feel like this is a bit contradictory with always returning the same shape regardless of input:
> 
> ```
> const { data, metadata } = await ...
> ```
> 
> I want to challenge if these options even make sense any more. I would have expected the query to return an empty array instead of throwing. Similar to GQL and our list methods.
> 
> Of cause it is convenient, but I don't mind forcing using to always have to check for the empty result set.
> 
> Curious to hear what you think.

It does exactly what you are expecting, if the query does not find anything it returns an empty array.

Now the options are shortcut like we have with `findOrFail` | `findOneOrFail` in Orm, or `retrieve` in our modules, it was to preserves us to always have to check after every retrieve. The idea was that since the query always call the list method, you loose that feature and instead it is backported to the remote joiner.

Now, if you do not specify the options, it will just return an empty array, but at least you keep the ability to either do the check yourself or ask the query to do the check for you. We also have another option to check if the relation have been returned or not and throw in that case, which gives the same capabilities, either you want to check it yourself and you can or you can ask the joiner to check it for you and throw

olivermrbl:
> It does exactly what you are expecting, if the query does not find anything it returns an empty array.

> Now the options are shortcut like we have with findOrFail | findOneOrFail in Orm, or retrieve in our modules, it was to preserves us to always have to check after every retrieve. The idea was that since the query always call the list method, you loose that feature and instead it is backported to the remote joiner.

> Now, if you do not specify the options, it will just return an empty array, but at least you keep the ability to either do the check yourself or ask the query to do the check for you. We also have another option to check if the relation have been returned or not and throw in that case, which gives the same capabilities, either you want to check it yourself and you can or you can ask the joiner to check it for you and throw

As I said, I understand what the option does. I am challenging if it makes sense to have now that we are simplifying the API of the Query engine. Just because we can do something, doesn't mean we always should. That said, this is not a strong opinion, so I am willing to let it stay. 

Does this option work for any key?

adrien2p:
> > It does exactly what you are expecting, if the query does not find anything it returns an empty array.
> 
> > Now the options are shortcut like we have with findOrFail | findOneOrFail in Orm, or retrieve in our modules, it was to preserves us to always have to check after every retrieve. The idea was that since the query always call the list method, you loose that feature and instead it is backported to the remote joiner.
> 
> > Now, if you do not specify the options, it will just return an empty array, but at least you keep the ability to either do the check yourself or ask the query to do the check for you. We also have another option to check if the relation have been returned or not and throw in that case, which gives the same capabilities, either you want to check it yourself and you can or you can ask the joiner to check it for you and throw
> 
> As I said, I understand what the option does. I am challenging if it makes sense to have now that we are simplifying the API of the Query engine. Just because we can do something, doesn't mean we always should. That said, this is not a strong opinion, so I am willing to let it stay.
> 
> Does this option work for any key?

I understand your opinion and I fine with any of the approach, I just thought this was nice to have when you always have to check. But what I like more is that you have to opt-in to enable it, so by default we can just let the user handle that himself. This option only works with the modules exported primary keys. so if you filter by any of the primary keys configured in the module joiner config then it will work

carlos-r-l-rodrigues:
If these options are removed, all the endpoints and workflows that uses it will have to perform the check and throw the error.

It works for any entity and it can be used as well for relations, so if none is found it stops resolving and aggregating data.

adrien2p:
should we open a ticket for discussion eventually if we want and try to not expand the scope more before the release? also maybe since it is a separate discussion from that pr maybe we can unblock this pr which fix an issue in our current usage.

I ve created a ticket @olivermrbl so that the discussion can continue, in the meantime if we could merge this one to resolve the bug :)

olivermrbl:
Alright, let's keep it as is for now

PR #9279, opened 2024-09-24
fix: Tax region + rates clean up

olivermrbl:
What
- Require `code` on Tax Rates
- Update dashboard to account for non-nullable code on Tax Rates

- Include `automatic_taxes` in API Route response

Closes CC-524 CC-525


packages/modules/tax/src/migrations/Migration20240924114005.ts
```diff
@@ -0,0 +1,19 @@
+import { Migration } from "@mikro-orm/migrations"
+
+export class Migration20240924114005 extends Migration {
+  async up(): Promise<void> {
+    this.addSql(
+      'alter table if exists "tax_rate" alter column "code" set not null;'
```
carlos-r-l-rodrigues: if this wasn't required before and there is any row where `code is null`, this will fail.
If the value is always there, just disregard the comment.

carlos-r-l-rodrigues (approved):

packages/modules/tax/src/migrations/Migration20240924114005.ts
olivermrbl: Nice catch, I'll include a backfilling strategy

PR #9281, opened 2024-09-24
fix(dashboard): use svg component instead of static png for logo

fPolic:


sradevski (approved):

sradevski:
I was just about to raise it that the logo is broken now 😄

PR #9308, opened 2024-09-25
fix(pricing): price set update deletes price list prices

fPolic:
**Why**
- price set update uses `upsertWithReplace` so price list prices are removed since we "ignore" them in regular price set operations and use price list methods to manage them

**What**
- preserve price list prices when updating price sets

---

FIXES CC-516


packages/modules/pricing/src/services/pricing-module.ts
```diff
@@ -515,6 +515,11 @@ export default class PricingModuleService
     // We can make the `insert` inside upsertWithReplace do an `upsert` instead to avoid this
     const normalizedData = await this.normalizeUpdateData(data)
 
+    const priceListPrices = await this.priceService_.list({
+      price_set_id: normalizedData.map(({ id }) => id),
+      price_list_id: { $ne: null },
+    })
+
```
srindom, line 522: thought(non-blocking): I wonder if this is the right approach to take since a price list price can potentially have a massive amount of prices so the performance of this could be bad.

It might be something to come back to later and have a more specialized endpoint to update/create individual prices without removing existing ones.

srindom (approved):
LGTM

packages/modules/pricing/src/services/pricing-module.ts
srindom: Realizing now that this is not fetching all prices from the price list but only prices for the price set that belongs to a price list. 

I.e., the number of prices here is not going to be the number of prices in the price list but rather how many times a product variant shows up in a price list - which will probably not be a number that balloons to thousands or even hundreds, so this should be fine 👍

PR #9312, opened 2024-09-25
fix(dashboard): allow to unset currency cell value

fPolic:
**What**
- unset datagrid currency cell on delete press instead of setting it to 0
- consolidate pricing editors validations
- fix PL edit pricing schema

---

FIXES CC-529


packages/admin/dashboard/src/routes/locations/location-service-zone-shipping-option-create/components/create-shipping-options-form/create-shipping-options-form.tsx
srindom, line 1: q(non-blocking): should a similar thing happen on editing shipping option prices?

The behavior seems to be correct when testing it but wanted to put it here to be sure.

srindom (approved):
LGTM!

packages/admin/dashboard/src/routes/locations/location-service-zone-shipping-option-create/components/create-shipping-options-form/create-shipping-options-form.tsx
fPolic: Good point, will add the same check there for consistency.

fPolic: Nvm, we had the check there before, should be consistent now across all pricing editors!

PR #9315, opened 2024-09-26
fix(dashboard): set/unset automatic taxes on a region

fPolic:
**What**
- allow toggling automatic taxes on a region
- fix cache and query issues

---

FIXES CC-520


olivermrbl:
already solved in #9279 but we can merge this first, it's cool

olivermrbl (approved):

PR #9322, opened 2024-09-26
feat(medusa,fulfillment): pass stock location data to fulfillment provider  

srindom:
**What**
- Fetches the stock location's details when creating a fulfillment and return fulfillment.
- Passes the data to the fulfillment module, which in turn passes it to the fulfillment provider.

**Why**
- When creating a fulfillment in a multi-location setup the fulfillment provider will need to know where the package is being sent from (so the shipping service can pick it up). 
- Previously, we didn't pass anything but the location id to the fulfillment provider. Because the fulfillment provider can't have a dependency on the stock location module this was not sufficient. 
- This change ensures there is enough data passed to the fulfillment provider to build integrations properly.


olivermrbl (approved):
LGTM

PR #9323, opened 2024-09-26
fix(types): fix order DTO types

shahednasser:
- Fix claim / exchange / return extending an order (we previously addressed this in the http types)
- add `@ignore` for all `raw_*` fields to hide them from references.

### DIFF
--- a/packages/core/types/src/order/common.ts
+++ b/packages/core/types/src/order/common.ts
@@ -1828,7 +1828,7 @@ export interface OrderClaimDTO {
   /**
    * The version of the order when the claim is applied.
    */
-  order_version: string
+  order_version: number
 
   /**
    * The claim's display ID.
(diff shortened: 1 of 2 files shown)


packages/core/types/src/order/common.ts
```diff
@@ -1613,25 +1868,79 @@ export interface OrderClaimDTO
...
-   * The items to be returned from the customer.
+   * The associated order.
    */
-  return_items: any[]
+  order?: OrderDTO
+
+  /**
+   * The version of the order when the exchange is applied.
+   */
+  order_version: string
```
fPolic: q: should this be a number?

fPolic (approved):
LGTM w/ one question.

packages/core/types/src/order/common.ts
shahednasser: Nice catch! WIll fix it 👍🏻

PR #9334, opened 2024-09-26
fix(types, medusa): http types fixes

shahednasser:
- Change `StoreCollection` to reference `StoreProduct`
- remove `created_by` from the returned customer fields in store routes (both in http types and query config in medusa)
- remove `version` from the returned order fields in store query config
- remove `transactions` from `StoreOrder`
- remove `no_notification` and `order_version` from `StoreReturn`
- Use http types in promotion rule routes
- Fix filters for product and variant store routes
- Remove request type argument from routes that don't accept paramters
- Fix `AdminDraftOrderListResponse` not having an array response type
- Fix store payload types for cart
- Other store / admin scoped fixes in http types


olivermrbl (approved):
Nice clean-up!

PR #9338, opened 2024-09-26
feat(dashboard,admin-vite-plugin,admin-bundler,admin-sdk): Rework admin extensions and introduce custom fields API

kasperkristensen:


packages/admin/admin-sdk/src/config/utils.ts
```diff
@@ -35,3 +40,77 @@ export function defineWidgetConfig(config: WidgetConfig) {
...
+
+/**
+ * Define a custom fields configuration.
+ *
+ * @param config The custom fields configuration.
+ * @returns The custom fields configuration.
+ *
+ * @experimental This API is experimental and may change in the future.
+ */
+export function unstable_defineCustomFieldsConfig<
```
olivermrbl, line 52: Until we know with certainty what approach we want to take with the custom fields, let's just remove the export of these two utilities, so they are not accessible to end-users and keep all the code.

It has become obvious that a more standalone Custom Fields API on the backend is worth dedicating time to post the 2.0 release, and this will likely influence the API shape of the admin utilities. What we provide now might solve more advanced use cases, but for simple field additions, it's a lot of work, code, and files when building it end to end. With that in mind, we don't want our users to start experimenting with this too much, even though it's denoted as unstable. 

I know you've put a lot of effort into this PR, but it feels wrong to publicly release and shout about a feature, we might want to rethink within a couple of months. Is that OK with you?

olivermrbl (approved):
Strong work @kasperkristensen, I left a comment regarding the form extensions 💪

packages/admin/admin-sdk/src/config/utils.ts
kasperkristensen: I think this is fine, I will just remove the export but keep all the code around, so we can pick it up again post V2 👍

PR #9341, opened 2024-09-26
fix: Bind abort controller to abort in sdk

sradevski:


fPolic (approved):

christiananese (approved):
LGTM 🚀

olivermrbl (approved):
LGTM, just out of curiousity, in what scenario did the previous implementation cause issues?

sradevski:
@olivermrbl you want to abort the stream when navigating away from a page (so as a return function to a useEffect hook), or when the server sends a disconnect event, so the UI doesn't try to reconnect. As part of the return func in useEffect it wouldn't be able to resolve the `this` correctly

PR #9343, opened 2024-09-26
fix(core-flows): shipping options for cart

carlos-r-l-rodrigues:
FIXES: CC-536


packages/core/core-flows/src/cart/workflows/list-shipping-options-for-cart.ts
```diff
@@ -21,83 +21,103 @@ export const listShippingOptionsForCartWorkflow = createWorkflow(
...
+        "type.id",
+        "type.label",
+        "type.description",
+        "type.code",
+
+        "provider.id",
+        "provider.is_enabled",
+
+        "calculated_price.calculated_amount",
+        "calculated_price.is_calculated_price_tax_inclusive",
```
fPolic: we need  "rules.attribute" and "rules.value in here for the storefront

packages/core/core-flows/src/cart/workflows/list-shipping-options-for-cart.ts
```diff
@@ -21,83 +21,103 @@ export const listShippingOptionsForCartWorkflow = createWorkflow(
...
+      ],
+      variables: {
+        context: {
+          is_return: "false",
```
fPolic: we want return options (that are enabled in store) as well on the storefront for customer returns
i.e. we want all options that are store enabled in the response

carlos-r-l-rodrigues: but I believe this should be an option in the query string no?
here it is the context, and if we set to `true` it will only bring the ones that are return.
I can remove it to return both.

fPolic: yes exactly we just need to remove `is_return` rule filter

> but I believe this should be an option in the query string no?

also sounds good 👍

packages/core/core-flows/src/cart/workflows/list-shipping-options-for-cart.ts
carlos-r-l-rodrigues: do we need the original price as well? or only the calculated and its rules?

adrien2p (approved):

packages/core/core-flows/src/cart/workflows/list-shipping-options-for-cart.ts
fPolic: `option.amount` is used now at the storefront, should we change that?

fPolic (approved):

PR #9355, opened 2024-09-27
fix(admin-ui): remove unsafe access to index

kasperkristensen:


packages/admin-ui/webpack.config.dev.ts
```diff
@@ -34,7 +34,7 @@ const getDevServerConfig = () => {
     ...devConfig,
     ...{
       devServer: {
-        port: 7001,
+        port: 7000,
```
carlos-r-l-rodrigues: is it intentional?

packages/admin-ui/ui/src/domain/orders/details/mark-shipped/index.tsx
```diff
@@ -75,7 +73,9 @@ const MarkShippedModal: React.FC<MarkShippedModalProps> = ({
       fulfillment.claim_order_id || fulfillment.swap_id || fulfillment.order_id
     const [type] = resourceId.split("_")
 
-    const tracking_numbers = data.tracking_numbers.map((tn) => tn.value)
+    const tracking_numbers = data.tracking_numbers
+      .map((tn) => tn.value)
```
carlos-r-l-rodrigues:
```suggestion
      .map((tn) => tn?.value)
```
not sure why the cause is, but if this is returning `undefined` as an array, we don't break

carlos-r-l-rodrigues (approved):

packages/admin-ui/ui/src/domain/orders/details/mark-shipped/index.tsx
kasperkristensen: Good catch, I don't know why this happening either in theory tn should never be undefined. But this works without having to spend to much time refactoring

PR #9360, opened 2024-09-28
fix: force string compare

srindom:
fixes CC-536


adrien2p (approved):

srindom:
@adrien2p - do you know if this change is sufficient? Just thinking that if some of the rule values are not stored as strings maybe this will act in an unpredictable way?

adrien2p:
> @adrien2p - do you know if this change is sufficient? Just thinking that if some of the rule values are not stored as strings maybe this will act in an unpredictable way? 

The only question I had in my head when looking at it was about boolean. But they are tested in the fulfillment modules integration tests and therefore it looks like it still works

PR #9361, opened 2024-09-28
fix(medusa,dashboard): don't send price updates for deleted regions/currencies

srindom:
**What**
- Fixes an issue where the admin dashboard would send region prices for deleted regions.
- Also, includes the implementation for `GET /admin/shipping-options/:id` and its corresponding SDK function.

**Why**
- When a region price for a deleted region was sent to the backend it would result in the insert hitting a not null constraint on the currency_code for prices. To avoid this the dashboard should not send region prices for deleted regions. 

**Additional context**
- Prices for deleted regions should ideally not be returned when fetching shipping option prices. However, we don't yet have a mechanism for cleaning up region prices after a region is deleted.

Fixes CC-540


packages/admin/dashboard/src/routes/locations/location-service-zone-shipping-option-pricing/location-service-zone-shipping-option-pricing.tsx
```diff
@@ -1,25 +1,17 @@
...
 
 export function LocationServiceZoneShippingOptionPricing() {
   const { so_id, location_id } = useParams()
 
-  const { shipping_options, isPending, isFetching, isError, error } =
-    useShippingOptions({
-      // TODO: change this when GET option by id endpoint is implemented
-      id: [so_id!],
+  const { shipping_option: shippingOption, isError, error } =
+    useShippingOption(so_id!, {
```
adrien2p: Q: instead of forcing the so_id to be present should we check and narrow it instead to prevent bugs from raising ?

packages/medusa/src/api/admin/shipping-options/[id]/route.ts
```diff
@@ -8,7 +8,31 @@ import {
...
+  AdminGetShippingOptionParamsType,
+  AdminUpdateShippingOptionType,
+} from "../validators"
+import { MedusaError } from "@medusajs/framework/utils"
+
+export const GET = async (
+  req: AuthenticatedMedusaRequest<AdminGetShippingOptionParamsType>,
+  res: MedusaResponse<HttpTypes.AdminShippingOptionResponse>
+) => {
+  const shippingOption = await refetchShippingOption(
```
Comment (olivermrbl) [inline packages/medusa/src/api/admin/shipping-options/[id]/route.ts:21]: nit(non-blocking): Use `query.graph`

olivermrbl (approved):

fPolic (approved):

packages/medusa/src/api/admin/shipping-options/[id]/route.ts
Comment (srindom) [reply packages/medusa/src/api/admin/shipping-options/[id]/route.ts:21]: updated the refetch method to use query.graph

PR #9364, opened 2024-09-28
fix: validate regions exist for shipping option price update

srindom:
**What**
- Adds a step to `updateShippingOptionsWorkflow` and `createShippingOptionsWorkflow` that validates if the region prices being updated have corresponding regions configured.

**Why**
- Previously, if you tried to send a region price update for a region that had been deleted the backend would throw an error when attempting to insert the region price. The error comes from a not-null constraint in the db, but it is better to validate that the regions we are trying to create prices for exist. 

Fixes CC-542

### DIFF
--- a/integration-tests/modules/__tests__/shipping-options/workflows/update-shipping-options.ts
+++ b/integration-tests/modules/__tests__/shipping-options/workflows/update-shipping-options.ts
@@ -257,6 +257,133 @@ medusaIntegrationTestRunner({
         )
       })
 
+      it("should fail on non-existent region update shipping options", async () => {
+        const regionService = container.resolve(
+          Modules.REGION
+        ) as IRegionModuleService
+
+        const [region] = await regionService.createRegions([
+          {
+            name: "Test region",
+            currency_code: "eur",
+            countries: ["fr"],
+          },
+        ])
+
+        const shippingOptionData: FulfillmentWorkflow.CreateShippingOptionsWorkflowInput =
+          {
+            name: "Test shipping option",
+            price_type: "flat",
+            service_zone_id: serviceZone.id,
+            shipping_profile_id: shippingProfile.id,
+            provider_id,
+            type: {
+              code: "manual-type",
+              label: "Manual Type",
+              description: "Manual Type Description",
+            },
+            prices: [
+              {
+                currency_code: "usd",
+                amount: 10,
+              },
+              {
+                region_id: region.id,
+                amount: 100,
+              },
+              {
+                currency_code: "dkk",
+                amount: 1000,
+              },
+            ],
+            rules: [
+              {
+                attribute: "total",
+                operator: RuleOperator.EQ,
+                value: "100",
+              },
+            ],
+          }
+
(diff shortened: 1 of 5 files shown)


integration-tests/modules/__tests__/shipping-options/workflows/update-shipping-options.ts
```diff
@@ -257,6 +257,133 @@ medusaIntegrationTestRunner({
...
+
+        const { result } = await createShippingOptionsWorkflow(container).run({
+          input: [shippingOptionData],
+        })
+
+        const remoteQuery = container.resolve(
+          ContainerRegistrationKeys.REMOTE_QUERY
+        )
+
+        let remoteQueryObject = remoteQueryObjectFromString({
```
adrien2p, line 316: Q: nothing to do but should we start using the query graph api instead when writing new things?

packages/core/core-flows/src/fulfillment/steps/validate-shipping-option-prices.ts
```diff
@@ -0,0 +1,53 @@
...
+    options: {
+      prices?: FulfillmentWorkflow.UpdateShippingOptionsWorkflowInput["prices"]
+    }[],
+    { container }
+  ) => {
+    const allPrices = options.flatMap((option) => {
+      if (!option.prices) {
+        return []
+      }
+      return option.prices
```
adrien2p: Sug: return option.prices ?? []

packages/core/core-flows/src/fulfillment/steps/validate-shipping-option-prices.ts
```diff
@@ -0,0 +1,53 @@
...
+    const regionIdSet = new Set<string>()
+
+    allPrices.forEach((p) => {
+      if ("region_id" in p && p.region_id) {
+        regionIdSet.add(p.region_id)
+      }
+    })
+
+    const regionService = container.resolve(Modules.REGION)
+    const regionList = await regionService.listRegions({
```
adrien2p, line 34: Sug: we should probably check the length before calling the service just as a best practice, i think we are handling it under the hood but if a module is replaced it might not always be the case wdyt? Also, it removes a call to the service and we can return early if there is nothing to check

adrien2p (approved):

integration-tests/modules/__tests__/shipping-options/workflows/update-shipping-options.ts
srindom: Yes happy to do that - do you have an example implementation somewhere that I can reference

adrien2p: https://github.com/medusajs/medusa/pull/9061
Should give some examples to look at. Let me know how it goes

srindom: Will probably merge this for now - still happy to do a separate PR with the graph api to try it out though if you can share an example :)

PR #9367, opened 2024-09-29
fix: Update tax lines on cart when region changes

olivermrbl:
**What**
- Removes broken cart retrieval step `retrieveWithCartLinks` in favor of `useRemoteQueryStep` in `updateTaxLinesWorkflows`

**Why**
- Filters variables in the step were passed with an incorrect shape
  - I removed the step, since it's only used once and the behavior is identical to the generic remote query step

**Note**
- Because the filters were passed incorrectly, the, now-removed, step has always returned the first cart in the database. As a result all tax calculations so far have been based on whatever the shape of that cart has. It basically ignores all input to this workflow


srindom (approved):
LGTM! important fix 👀

packages/core/core-flows/src/cart/workflows/add-shipping-method-to-cart.ts
```diff
@@ -112,8 +112,8 @@ export const addShippingMethodToWorkflow = createWorkflow(
 
     updateTaxLinesWorkflow.runAsStep({
       input: {
-        cart_or_cart_id: input.cart_id,
-        shipping_methods: shippingMethodsToAdd,
+        cartId: input.cart_id,
+        shippingMethods: shippingMethodsToAdd,
```
adrien2p: nit(non blocking): havent we settle a convention for input prop naming to be snake cases?

packages/core/core-flows/src/cart/workflows/update-tax-lines.ts
```diff
@@ -74,18 +70,21 @@ export const updateTaxLinesWorkflowId = "update-tax-lines"
 export const updateTaxLinesWorkflow = createWorkflow(
   updateTaxLinesWorkflowId,
   (input: WorkflowData<UpdateTaxLinesWorkflowInput>): WorkflowData<void> => {
-    const cart = retrieveCartWithLinksStep({
-      cart_or_cart_id: input.cart_or_cart_id,
+    const cart = useRemoteQueryStep({
```
adrien2p, line 73: Q: should we migrate this step to use query.graph API? currently I think we are deprecating this API so maybe it makes sense to move on on the step wdyt? I could do that in a separate pr of course

adrien2p (approved):

packages/core/core-flows/src/cart/workflows/add-shipping-method-to-cart.ts
olivermrbl: Oh yeah, my bad. Will update

packages/core/core-flows/src/cart/workflows/update-tax-lines.ts
olivermrbl: Have a new `useQueryStep` in a separate PR. Will wait until that is merged

adrien2p: Yes if course 👌was just thinking about missing pieces for the release

ricardolopesdev:
@olivermrbl you need to remove the tax in cart/product when the country selected don't have any tax associated.
I test this latest version of Tax Module and work fine when you select one country with tax defined, but if you change to one country don't have tax, the product/cart keep the latest tax applied to product. If you change to another country with tax associated the calculation it's done and it's correct. Please check this issue.
Thanks.

PR #9368, opened 2024-09-29
fix(medusa): Instrumentation resolution

adrien2p:


thetutlage (approved):

packages/medusa/src/commands/start.ts
```diff
@@ -24,7 +25,7 @@ const CRON_SCHEDULE = EVERY_SIXTH_HOUR
  */
 async function registerInstrumentation(directory: string) {
   try {
-    const instrumentation = await import(
+    const instrumentation = await dynamicImport(
```
olivermrbl, line 28: suggestion: can we add a test so this doesn't happen again?

adrien2p: I ll add on the second pr ill make

adrien2p: Cause it will make more sense than testing the import module from nodejs 😊 the next one will be more granular

olivermrbl: Makes sense, don't think that is our responsibility 😄

PR #9369, opened 2024-09-29
fix: Handle region updates on cart (1/n)

olivermrbl:
**What**

On cart creation:
- If region only has one country -> create cart with country code

On cart updates:
- If shipping address country code is provided in input ->
  - If cart region doesn't include that country -> throw
  - If cart includes the country -> update shipping address
- If region is provided in input and is different from the one currently on the cart -> 
  - If there is a shipping address on the cart -> clear the address
    - If the region only has one country -> set country code of address
  - If there is not a shipping address on the cart ->
    - If the region only has one country -> set country code of address

Closes CC-545


integration-tests/modules/__tests__/cart/store/carts.spec.ts
```diff
@@ -316,6 +316,49 @@ medusaIntegrationTestRunner({
           )
```
olivermrbl, line 316: These are all tests of the HTTP layer, so I'll move these to `integration-tests/http` before merging. Keeping them as is, because the diff would otherwise be off

integration-tests/http/__tests__/product/store/product.spec.ts
```diff
@@ -1642,7 +1642,7 @@ medusaIntegrationTestRunner({
           products[0].variants[0].calculated_price.calculated_amount_without_tax.toFixed(
             1
           )
-        ).toEqual("40.9")
+        ).toEqual("45.0")
```
olivermrbl, line 1645: This is intentional. The country with code `it` is a tax-inclusive region

packages/core/core-flows/src/cart/workflows/create-carts.ts
```diff
@@ -97,8 +98,30 @@ export const createCartWorkflow = createWorkflow(
       context: pricingContext,
     })
 
+    // If there is only one country in the region, we prepare a shipping address with that country's code.
+    //   This is useful for other operations, such as tax line calculations.
+    // That is if the shipping address is not provided in the input.
+    const shippingAddress = when({ region, input }, (data) => {
```
sradevski: Q: Can't we do this default shipping address definition inside the transform below? It might be a bit cleaner even if you don't need to create a new step, WDYT?

packages/core/core-flows/src/cart/workflows/update-cart.ts
```diff
@@ -35,35 +31,93 @@ export const updateCartWorkflowId = "update-cart"
...
+          region_id: data.region?.id, // This is either the region from the input or the region from the cart or null
+        }
+
+        // When the region is updated, we do a few things:
+        // - If the shipping address country code is provided, we need to make sure the country is in the region
+        //   - If not, we throw
+        // - If the shipping address is not provided
+        //   - Clear the shipping address if the region has more than one country
+        //   - Set the country code if the region has only one country
+        const regionIsNew = !!(
```
sradevski: I don't think you need the double negation here

packages/medusa/src/api/store/products/helpers.ts
```diff
@@ -64,6 +64,8 @@ export const wrapProductsWithTaxPrices = async <T>(
     return
   }
 
+  console.log("Tax context: ", req.taxContext)
```
sradevski: Probably not needed 😄

packages/core/core-flows/src/cart/workflows/update-cart.ts
```diff
@@ -35,35 +31,93 @@ export const updateCartWorkflowId = "update-cart"
...
-        if (isPresent(updateCartData.region_id)) {
-          if (!data.region) {
+        const data_ = {
+          ...updateCartData,
+          currency_code: data.region?.currency_code,
+          region_id: data.region?.id, // This is either the region from the input or the region from the cart or null
+        }
+
+        // When the region is updated, we do a few things:
+        // - If the shipping address country code is provided, we need to make sure the country is in the region
```
sradevski: It was a bit difficult for me to understand why we are doing these steps (I did understand it after spending a bit of time on it). Maybe it's better to describe what we try to achieve, as the "what" part can be deduced from the code itself

packages/core/core-flows/src/cart/workflows/create-carts.ts
olivermrbl: Yeah, I think we can do that. This was the end-state of a few earlier iterations and after going deep into a rabbit-hole of the `when-then`, so I never got around to reconsidering this 😄

packages/core/core-flows/src/cart/workflows/update-cart.ts
```diff
@@ -35,35 +31,86 @@ export const updateCartWorkflowId = "update-cart"
 export const updateCartWorkflow = createWorkflow(
   updateCartWorkflowId,
   (input: WorkflowData<UpdateCartWorkflowInputDTO & AdditionalData>) => {
-    const [salesChannel, region, customerData] = parallelize(
+    const cartToUpdate = useRemoteQueryStep({
+      entry_point: "cart",
+      variables: { id: input.id },
+      fields: ["id", "shipping_address.*", "region.*"],
```
carlos-r-l-rodrigues:
```suggestion
      fields: ["id", "shipping_address.*", "region.*", "region.countries.*"],
```

packages/core/core-flows/src/cart/workflows/update-cart.ts
```diff
@@ -35,35 +31,86 @@ export const updateCartWorkflowId = "update-cart"
...
+      list: false,
+      throw_if_key_not_found: true,
+    }).config({ name: "get-cart" })
+
+    const regionId = transform(
+      { input, cartToUpdate },
+      (data) => data.input.region_id || data.cartToUpdate.region_id
+    )
```
carlos-r-l-rodrigues:
```suggestion
```

packages/core/core-flows/src/cart/workflows/update-cart.ts
```diff
@@ -35,35 +31,86 @@ export const updateCartWorkflowId = "update-cart"
...
+    }).then(() => {
+      return useRemoteQueryStep({
+        entry_point: "region",
+        variables: { id: regionId },
+        fields: ["id", "countries.*", "currency_code", "name"],
+        list: false,
+        throw_if_key_not_found: true,
+      }).config({ name: "get-region" })
+    })
+
```
carlos-r-l-rodrigues:
```suggestion
    const regionFetch = when({ input }, ( { input }) => {
      return !!input.region_id
    }).then(() => {
      return useRemoteQueryStep({
        entry_point: "region",
        variables: { id: input.region_id },
        fields: ["id", "countries.*", "currency_code", "name"],
        list: false,
        throw_if_key_not_found: true,
      }).config({ name: "get-region" })
    })

    const region = transform({cartToUpdate, regionFetch}, (data) => {
      return data.regionFetch ?? data.cartToUpdate.region
    })
```

integration-tests/modules/__tests__/cart/store/carts.spec.ts
```diff
@@ -784,12 +823,13 @@ medusaIntegrationTestRunner({
...
         })
 
-        it("should update a cart's region, sales channel, customer data and tax lines", async () => {
+        it.only("should update a cart's region, sales channel, customer data and tax lines", async () => {
```
carlos-r-l-rodrigues:
```suggestion
        it("should update a cart's region, sales channel, customer data and tax lines", async () => {
```

integration-tests/modules/__tests__/cart/store/carts.spec.ts
```diff
@@ -316,6 +316,49 @@ medusaIntegrationTestRunner({
           )
         })
 
+        it("should create cart with shipping address country code when there is only one country assigned to the region", async () => {
+          const region = await regionModule.createRegions({
+            name: "US",
+            currency_code: "usd",
+            countries: ["us"],
+          })
```
riqwan, line 324:
```suggestion
          const region = (await api.post(
            `/admin/regions`,
            {
              name: "US",
              currency_code: "usd",
              countries: ["us"],
            },
            adminHeaders
          )).data.region
```

integration-tests/modules/__tests__/cart/store/carts.spec.ts
```diff
@@ -759,13 +796,15 @@ medusaIntegrationTestRunner({
             })
           )
 
-          await cartModule.updateCarts(cart.id, {
-            region_id: region.id,
+          const region = await regionModule.createRegions({
+            name: "US",
+            currency_code: "usd",
+            automatic_taxes: false,
           })
```
riqwan, line 803:
```suggestion
          const region = (await api.post(
            `/admin/regions`,
            {
              name: "US",
              currency_code: "usd",
              automatic_taxes: false
            },
            adminHeaders
          )).data.region
```

packages/core/core-flows/src/cart/workflows/create-carts.ts
```diff
@@ -119,6 +116,15 @@ export const createCartWorkflow = createWorkflow(
...
         }
 
+        // If there is only one country in the region, we prepare a shipping address with that country's code.
+        if (!data.input.shipping_address) {
+          if (data.region.countries.length === 1) {
+            data_.shipping_address = {
+              country_code: data.region.countries[0].iso_2,
+            }
+          }
+        }
```
carlos-r-l-rodrigues:
```suggestion
        if (!data.input.shipping_address && data.region.countries.length === 1) {
          data_.shipping_address = {
            country_code: data.region.countries[0].iso_2,
          }
        }
```

carlos-r-l-rodrigues (approved):

riqwan (approved):
noice!

PR #9380, opened 2024-09-30
fix(dashboard): categories add products

fPolic:
**What**
- fix listing/adding products in the add category products flow

---

FIXES CC-523


sradevski (approved):

olivermrbl (commented):
We seem to be missing some cache invalidation. If I open the add products modal, add products, save, and then do the same again, the products I just added are not shown as "disabled" in the table.

fPolic:
Nice catch - should be fixed!

olivermrbl (approved):

PR #9381, opened 2024-09-30
fix(product): fix product variant update price with min_quantity and max_quantity

BeyondOT:
## What : 
The issue is updating the variant price with min_quantity and max_quantity  always returned null for both these attributes. Which is not the expected behaviour based on the admin api reference. https://docs.medusajs.com/api/admin#products_postproductsproductvariantsvariant

## How : 
Make sure to include all the attributes when updating the variant price.

## Testing: 
Update variant price through api including min_quantity and max_quantity.

### DIFF
--- a/integration-tests/api/__tests__/admin/product.js
+++ b/integration-tests/api/__tests__/admin/product.js
@@ -2072,6 +2072,50 @@ medusaIntegrationTestRunner({
           )
         })
 
+        it("successfully updates a variant's default prices by changing an existing price (currency_code) with min_quantity and max_quantity", async () => {
+          const data = {
+            prices: [
+              {
+                currency_code: "usd",
+                amount: 1500,
+                min_quantity: 1,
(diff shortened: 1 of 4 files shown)


adrien2p (approved):

BeyondOT:
I just added a commit that fixes the unit test error.

The price update fix was made by spreading the price object in money amount so the test should expect the currency code to be there as well if it was included in the price object.

carlos-r-l-rodrigues (approved):

BeyondOT:
Hi @adrien2p, this is my first pull request, and it's been open for over two weeks now. The tests have passed, and it has already been approved. Could you let me know if there's anything preventing it from being merged?

riqwan (approved):
nice work! 💪🏻

riqwan:
Apologies for the delay @BeyondOT, merging this in now! 💪🏻

PR #9384, opened 2024-09-30
feat: Add `useQueryStep`

adrien2p:
**What**
Provide a new query step as a future replacement of the remote query step


packages/core/core-flows/src/common/steps/use-query-graph.ts
```diff
@@ -0,0 +1,37 @@
...
+} from "@medusajs/workflows-sdk"
+import { ContainerRegistrationKeys } from "@medusajs/utils"
+
+type EntryStepInput<TEntry extends string> = RemoteQueryInput<TEntry> & {
+  options?: RemoteJoinerOptions
+}
+
+const useQueryGraphStepId = "use-query-graph-step"
+
+const step = createStep(
```
olivermrbl, line 16: nice! is this typed as well?

adrien2p: it will yes, I just need a bit more time to test it

olivermrbl:
@adrien2p, is this ready to merge? Would be nice to get in, so we can start using it.

olivermrbl (approved):

PR #9387, opened 2024-09-30
fix: allow customer_id in UpdateCart

VariableVic:
Allows `customer_id` to be present when updating a cart.


riqwan (approved):

srindom (changes requested):
I'd expect to see:

1. A test
2. Some considerations around edge cases - e.g. update has both customer_id and email. Which one dictates the customer on the cart?

assume this is for the b2b starter - is it maybe better to add a custom endpoint?

olivermrbl:
@VariableVic, can I get you to address the comments, so we can move forward with this?

VariableVic:
@srindom 

The issue is that carts are not properly linked to existing customers with an account because we can't pass in a customer_id for logged in customers. Instead, a new anonymous customer is created with the same email, and linked to the cart.

AFAICT the [workflow](https://github.com/medusajs/medusa/blob/develop/packages/core/core-flows/src/cart/workflows/update-cart.ts#L45) expects a customer_id to be present for customers with an account. When passing in an email, the workflow will only look for customers that don't have an account (`has_account`: false)

> update has both customer_id and email.

This is taken care of in [findOrCreateCustomerStep](https://github.com/medusajs/medusa/blob/develop/packages/core/core-flows/src/cart/steps/find-or-create-customer.ts#L25) - the workflow accounts for receiving both email and customer id.

> A test

I have no experience writing these tests. Maybe @riqwan can help out?

I don't think this is a B2B starter specific issue, that's why we decided to open the PR.

riqwan:
Whats the expected outcome here?

If the `customer_id` is passed to the workflow, we always prioritize the email from the customer record and ignore any email coming into the api. Is this missing just a test?

Only case I can think of is when you run 2 requests:

1. Update cart with customer_id
2. Update cart with email

In this case ^ do we update the customer's email, but keep the customer_id? and vice versa, do we override the email?

Alternatively, we remove `customer_id` from the storefront API and then pass `customer_id` to all cart workflows when the customer is logged in. i.e to handle the case of when the user goes about adding items to cart and then logging in after.

VariableVic:
The expected outcome is that we can update a cart for a logged in customer. This is currently impossible.

I think that when a customer is logged in, the customer_id and email from the customer record should always be used on the cart.

adrien2p (approved):
should we add some tests?

olivermrbl:
Discussed internally, and we will not support `customer_id` in the payload. Customer ID should always be set by being logged in OR by email. Note for email: It will only work if the customer does not have an account, as you would otherwise be able to impersonate users in the checkout flow.

PR #9390, opened 2024-09-30
fix(medusa): Replace default http tracer

adrien2p:
**What**
the tracer was assigned to the exports instead of the start object

### DIFF
--- a/packages/core/framework/src/http/router.ts
+++ b/packages/core/framework/src/http/router.ts
@@ -180,7 +180,7 @@ function getBodyParserMiddleware(args?: ParserConfigArgs) {
 
 // TODO this router would need a proper rework, but it is out of scope right now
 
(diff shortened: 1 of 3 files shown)


carlos-r-l-rodrigues (approved):

sradevski (approved):

adrien2p:
DO NOT MERGE YET :D

thetutlage:
@adrien2p May I know what exactly we changed here?

adrien2p:
> @adrien2p May I know what exactly we changed here?

The instrumentation wasn't loading and the spans wasn't grouped for some reason. For the http adapter the issue was that we required the module and the function to replace was on start which is default exported but we attached it on the exports instead of the start which lead to always rely on the noop impl. For the rest it was cleanup while investigating

PR #9412, opened 2024-10-01
fix(dashboard): location details loader

fPolic:
**What**
- fix location loader returning stale data if SL details are fetched with another field params

**Why**
- if we visit the order details page and visit location details afterwards, delivery and pickup will appear disabled because a new query with `fulfillment_sets` is not fetched


olivermrbl (approved):
nice catch

PR #9415, opened 2024-10-01
fix(dashboard, fulfilment): fulfilment providers enabled check

fPolic:
**What**
- hide disabled fulfilment providers on the admin
- check if the fulfilment provider has an identifier when loading providers

---

FIXES CC-549


olivermrbl (approved):
Do we need to add the same two checks for payment providers?

fPolic:
@olivermrbl added!

PR #9422, opened 2024-10-02
fix: Export all classes and types from the SDK

sradevski:
Extending the SDK is a common pattern, so having access to all underlying types and classes is important for creating a wrapper.


christiananese (commented):
🚀

christiananese (approved):

PR #9426, opened 2024-10-02
feature: Add MikroORM CLI wrapper to bypass hardcoded module system

thetutlage:
FIXES: FRMW-2727

MikroORM (with version 5.9) has [hardcoded the TypeScript module](https://github.com/mikro-orm/mikro-orm/blob/5.x/packages/core/src/utils/ConfigurationLoader.ts#L138-L139) system to `commonjs`, which makes it incompatible with the module system we are using, ie `Node16`.

So, in order to continue using the Mikro ORM CLI within our modules, we will have to monkey-patch the block of code responsible for configuring `ts-node`. However, the monkey-patching must be done before their CLI gets booted.

As a result of this, we have to create a wrapper CLI on top of Mikro ORM CLI that performs the following steps.

- Monkey-patch the relevant code
- Register Mikro ORM CLI as the second step.

Due do this, we will have to use this new wrapper CLI within the modules, which is exposed as `medusa-db`. Maybe, `medusa-db` is not a great name, so please send your suggestions.


packages/core/framework/src/mikro-orm-cli/bin.ts
```diff
@@ -0,0 +1,61 @@
+/**
+ * Customer wrapper on top of MikroORM CLI to override the issue
```
adrien2p: custom you mean right? :joy:

packages/modules/order/bin.js
```diff
@@ -0,0 +1,45 @@
+const path = require("path")
```
adrien2p: Q: is it normal to have it in the module order?

packages/core/framework/src/mikro-orm-cli/bin.ts
thetutlage: 😅 Yes

packages/modules/order/bin.js
thetutlage: Ohh. Have to delete this file. Created for expirement

adrien2p (approved):

packages/core/framework/package.json
```diff
@@ -10,6 +10,9 @@
     "!dist/**/__mocks__",
     "!dist/**/__fixtures__"
   ],
+  "bin": {
+    "medusa-db": "./dist/mikro-orm-cli/bin.js"
```
olivermrbl: thought: since this is specific to MikroORM, maybe we could call it `medusa-mikro-orm`

That said, maybe this is not super important, as this would not be used by end-users right?

olivermrbl (approved):
LGTM

packages/core/framework/package.json
thetutlage: Yeah, we can call it `medusa-mikro-orm` too. I think, the end-users will use it if they develop custom modules. Because, we do not have any dedicated CLI for that

olivermrbl: > I think, the end-users will use it if they develop custom modules

Not in their own project right? We do have a CLI for managing migrations here afaik. 

Maybe we are talking about two different use cases.

thetutlage: > Not in their own project right? We do have a CLI for managing migrations here afaik.

Yeah, the projects (aka starter kits) are all good, since they use the `npx medusa` CLI.

But custom modules (as standalone packages) will be created as we create them.

olivermrbl: Cool, that makes sense 👍

packages/modules/order/tsconfig.json
```diff
@@ -8,5 +8,8 @@
       "@types": ["./src/types"],
       "@utils": ["./src/utils"]
     }
+  },
+  "ts-node": {
+    "transpileOnly": true
   }
```
riqwan: q: shouldn't this be on the base config?

riqwan (approved):
Just tested this, thanks man! 💪🏻

packages/modules/order/tsconfig.json
thetutlage: Needs to be removed altogether. Slipped from some custom testing

PR #9427, opened 2024-10-02
fix(workflows-sdk): when then return value

carlos-r-l-rodrigues:
FIXES: FRMW-2731

### DIFF
--- a/packages/core/orchestration/src/transaction/orchestrator-builder.ts
+++ b/packages/core/orchestration/src/transaction/orchestrator-builder.ts
@@ -34,7 +34,6 @@ export class OrchestratorBuilder {
     this.updateDepths(this.steps, {}, 1, -1)
     return this
   }
-  a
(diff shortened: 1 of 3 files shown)


packages/core/orchestration/src/transaction/orchestrator-builder.ts
```diff
@@ -22,16 +22,19 @@ export class OrchestratorBuilder {
...
+              (steps!.action ? steps : steps!.next) as InternalStep
+            )
           )
         : undefined,
     }
 
     this.updateDepths(this.steps, {}, 1, -1)
     return this
   }
+  a
```
adrien2p: 😂 it got lost

olivermrbl (approved):
Nice @carlos-r-l-rodrigues 💪

adrien2p (approved):

PR #9429, opened 2024-10-02
feat(framework,medusa): Ensure publishable key middleware is set for all store endpoints

riqwan:
what:

- publishable key middleware is applied to all store routes on a router level

RESOLVES CC-519


packages/core/modules-sdk/src/remote-query/parse-filters.ts
```diff
@@ -33,23 +29,23 @@ export function parseAndAssignFilters(
   const joinerConfigs = MedusaModule.getAllJoinerConfigs()
 
   for (const [filterKey, filterValue] of Object.entries(filters)) {
-    let entryAlias!: JoinerServiceConfigAlias
+    // let entryAlias!: JoinerServiceConfigAlias
```
riqwan: These changes will be cleaned up in Adrien's PR

packages/core/modules-sdk/src/remote-query/parse-filters.ts
```diff
@@ -33,23 +29,23 @@ export function parseAndAssignFilters(
...
+    // entryAlias = alias
     entryJoinerConfig = joinerConfig
 
-    const entryEntity = entitiesMap[entryAlias.entity!]
-    if (!entryEntity) {
-      throw new Error(
-        `Entity ${entryAlias.entity} not found in the public schema of the joiner config from ${entryJoinerConfig.serviceName}`
-      )
-    }
+    // const entryEntity = entitiesMap[entryAlias.entity!]
```
sradevski: Is this intentional?

riqwan: yes, the latest rebase from develop will have a comment in there

packages/core/framework/src/http/middlewares/ensure-publishable-api-key.ts
```diff
@@ -0,0 +1,77 @@
...
+
+  if (!isPresent(publishableApiKey)) {
+    try {
+      throw new MedusaError(
+        MedusaError.Types.NOT_ALLOWED,
+        `Publishable API key required in the request header: ${PUBLISHABLE_KEY_HEADER}. You can manage your keys in settings in the dashboard.`
+      )
+    } catch (e) {
+      return next(e)
+    }
```
adrien2p, line 30: Q: Do you remember why we are not doing 😂 
```ts
const error = new MedusaError(...)
return next(error)
```

adrien2p (approved):

packages/core/framework/src/http/middlewares/ensure-publishable-api-key.ts
riqwan: I tried this and the middleware got stuck 🤷🏻

PR #9431, opened 2024-10-02
docs: Fix typo. Use instead yuse

zaidrashid:
Fix typo. Use instead of use. I found this while reading through the documents here https://docs.medusajs.com/v2/resources/references/auth/provider#validatecallback


shahednasser (approved):
Nice catch thanks!

PR #9447, opened 2024-10-03
fix: If country on cart has no tax region, clear tax lines

olivermrbl:
**What**

If the country on the shipping address changes to a country without an associated tax region, we clear the tax lines on shipping methods and line items.


packages/core/core-flows/src/cart/steps/set-tax-lines-for-items.ts
```diff
@@ -35,18 +35,15 @@ export const setTaxLinesForItemsStep = createStep(
...
     const shippingTaxLinesData =
       normalizeShippingTaxLinesForCart(shipping_tax_lines)
-    const setShippingTaxLinesPromise = shippingTaxLinesData.length
-      ? await cartService.setShippingMethodTaxLines(
-          cart.id,
-          shippingTaxLinesData
-        )
-      : 0
+    const setShippingTaxLinesPromise =
+      await cartService.setShippingMethodTaxLines(cart.id, shippingTaxLinesData)
```
olivermrbl: In the case of no tax region for the provided country, `shippingTaxLinesData` and `itemsTaxLinesData` will be empty arrays and that should be reflected on the cart.

fPolic (approved):

carlos-r-l-rodrigues (approved):

packages/core/core-flows/src/cart/steps/set-tax-lines-for-items.ts
riqwan: No longer a promise, can be removed from the promise array.

```suggestion
    await cartService.setShippingMethodTaxLines(cart.id, shippingTaxLinesData)
```

riqwan (approved):
lgtm!

packages/core/core-flows/src/cart/steps/set-tax-lines-for-items.ts
olivermrbl: I guess none of the promises are relevant. Please check latest commit.

riqwan (approved):

PR #9455, opened 2024-10-03
chore: workflow internals improvementss

adrien2p:
**What**
- Now `transformer` results are temporarily stored in the distributed transaction for its life time
- `deep-copy` will not run on proxies which prevents running proxies for every property increasing the CPU usage and blocking the event loop
- refactored the `context` creation for steps to be in a separate function and shared for compensate and invoke
- Improve `event bus` to only emitting events when the instance is `server` only
- `when/then` result resolution fix
- improve and simplify the workflow `proxify` util
- Remove data copy from `transform` to prevent duplicated copy of an already copied object
- Workflow engine `storage` remove duplicated copy of the data before saving
- parallelize steps that can be in the `cart` area


packages/modules/event-bus-local/src/services/event-bus-local.ts
```diff
@@ -21,18 +22,24 @@ eventEmitter.setMaxListeners(Infinity)
...
+    moduleDeclaration: InternalModuleDeclaration
+  ) {
     // @ts-ignore
     // eslint-disable-next-line prefer-rest-params
     super(...arguments)
 
     this.logger_ = logger
     this.eventEmitter_ = eventEmitter
     this.groupedEventsMap_ = new Map()
+    this.#isWorkerMode = moduleDeclaration.worker_mode !== "server"
```
sradevski: I think we should also emit events if the mode is `shared` I believe?

sradevski (commented):
I was just checking for changes to make sure it doesn't affect platform, I think this one issue might (but let me know if I am wrong here)

packages/modules/event-bus-local/src/services/event-bus-local.ts
olivermrbl: +1, I think this should be `moduleDeclaration.worker_mode === "worker"` – otherwise you won't be able to process events in a setup where one instance serves both purposes.

carlos-r-l-rodrigues: We emit them, we just don't let subscribers on instances that are `server` only.
@olivermrbl the default one is `shared`, so this should work fine.

adrien2p: yes I am not sure I get it too, we just prevent the subscribers from registering themselves if you are in server mode only. if you are in shared mode it will just work as it does today. It prevents smaller project to be required to have both server and worker separately

olivermrbl: Just reverting back here, the behavior seems correct - my bad -  I just got tricked by the naming at first glance. Maybe we can say `shouldRegisterSubscribers` or `shouldSubscribe` because that is this flag's purpose here?

adrien2p: Well, it can be used for any purpose that would need to know if we are in worker mode or not. If someone build his own event bus modules (which will happen) they might also use it for something else. So in the base class the information is are we in worker mode, and then it happens that we use it for our subscribers management

adrien2p: Though, we can create another variable with the name should subscribe if that makes it clearer 😅

olivermrbl (approved):
LGMT, nice work 💪

PR #9456, opened 2024-10-03
fix(core-flows): update cart promotion data usage

adrien2p:
**What**
The nullish coalescing can't really work since we are using proxies under the hood which is present

### DIFF
--- a/packages/core/core-flows/src/cart/workflows/update-cart-promotions.ts
+++ b/packages/core/core-flows/src/cart/workflows/update-cart-promotions.ts
@@ -1,8 +1,9 @@
 import { PromotionActions } from "@medusajs/framework/utils"
 import {
-  WorkflowData,
   createWorkflow,
   parallelize,
+  transform,
+  WorkflowData,
 } from "@medusajs/framework/workflows-sdk"
 import { useRemoteQueryStep } from "../../common"
 import {
@@ -42,10 +43,18 @@ export const updateCartPromotionsWorkflow = createWorkflow(
       list: false,
     })
 
+    const promo_codes = transform({ input }, (data) => {
+      return (data.input.promo_codes || []) as string[]
+    })
+
+    const action = transform({ input }, (data) => {
+      return data.input.action || PromotionActions.ADD
+    })
+
     const promotionCodesToApply = getPromotionCodesToApply({
       cart: cart,
-      promo_codes: input.promo_codes ?? [],
-      action: input.action || PromotionActions.ADD,
+      promo_codes,
+      action: action as PromotionActions,
     })
 
     const actions = getActionsToComputeFromPromotionsStep({


adrien2p:
@shahednasser did we documented that? The fact that when you want to use data in workflows you are not really manipulating them but rather manipulating proxies under the hood, so nullishcoalescing and ternary like that can't work. Instead users should rely on transformers

shahednasser:
No not aware of this. I can document it tomorrow

olivermrbl (approved):
I suspect we might be doing this more places, I recently found a similar pattern in a cart workflow

adrien2p:
> I suspect we might be doing this more places, I recently found a similar pattern in a cart workflow

Humm, but I was sure we share that internally and the same about data manipulation in general in workflow. It is just a composer basically to organize steps and stuff in the order they should be executed and we allow through utils to have places to manipulate the data to be ingested by steps and sub workflows

adrien2p:
> No not aware of this. I can document it tomorrow

Basically the composer function only serves to compose a flow with a certain order of execution and configuration. Then we have utils like transform which allow to manipulate data if needed and then those are resolved when we reach a strp or workflow consuming it.

Cc @carlos-r-l-rodrigues we might write again about it wdyt?

riqwan (approved):
I know this might be a long stretch, but can we detect if we're doing this within the proxies and throw an error? 

In hindsight, I knew this wasn't correct, but because I'm writing in "JS", it was just intuitive to continue along. 😅

adrien2p:
> I know this might be a long stretch, but can we detect if we're doing this within the proxies and throw an error? 
> 
> 
> 
> In hindsight, I knew this wasn't correct, but because I'm writing in "JS", it was just intuitive to continue along. 😅 

🤣 unfortunately there is no way for us to know unless we do a static analysis of the composer function. The proxies are around the object so that when you access a prop it generates a transformer for you instead.

packages/core/core-flows/src/cart/workflows/update-cart-promotions.ts
```diff
@@ -42,10 +43,18 @@ export const updateCartPromotionsWorkflow = createWorkflow(
...
+
+    const action = transform({ input }, (data) => {
+      return data.input.action || PromotionActions.ADD
+    })
+
     const promotionCodesToApply = getPromotionCodesToApply({
       cart: cart,
-      promo_codes: input.promo_codes ?? [],
-      action: input.action || PromotionActions.ADD,
+      promo_codes,
```
olivermrbl, line 56: on another note, this could probably also be handled by the step itself, so that it is more resuable

adrien2p: Yes 100% but it is also handled I believe

PR #9465, opened 2024-10-04
fix: Remove extra saving on serialization which breaks the chain

adrien2p:
**What**
The extra serialization check hapen to break the serialization chain by reusing already serialized entities when they have been serialized from a different parents sequence


adrien2p:
Cc @shahednasser @olivermrbl

olivermrbl (approved):

PR #9466, opened 2024-10-04
docs: fix imports from http

shahednasser:
- Change `@medusajs/framework/http` imports to just `@medusajs/framework` (both work)
- Fix logger imports


adrien2p:
Question: isn't it better to import from the subpath instead of the route where everything is mixed? that way you know the domain of what you are using

PR #9473, opened 2024-10-04
chore(types): index module type

carlos-r-l-rodrigues:


packages/core/types/src/index.ts
```diff
@@ -9,12 +9,14 @@ export * from "./common"
...
 export * from "./customer"
 export * from "./dal"
+export * from "./dml"
 export * from "./event-bus"
 export * from "./feature-flag"
 export * from "./file"
 export * from "./file-service"
 export * from "./fulfillment"
 export * from "./http"
+export * from "./index_data"
```
olivermrbl: question: Why does this use `_` in the file name? 🤔

olivermrbl (approved):

packages/core/types/src/index.ts
carlos-r-l-rodrigues: my bad. let me rename it.

PR #9474, opened 2024-10-04
fix(payment): correct import of MedusaError

Veskel01:
What

Fixes a bug related to incorrect import of @medusajs/utils:

`Could not resolve module: Payment. Error: Cannot find module '@medusajs/utils'`

Require stack:
- ...\node_modules\@medusajs\payment\dist\loaders\providers.js


olivermrbl (approved):
LGTM, thanks!

carlos-r-l-rodrigues (approved):

PR #9479, opened 2024-10-04
update CreateOrderFulfillmentWorkflowInput

420coupe:
Update CreateOrderFulfillmentWorkflowInput to have an optional ``requires_shipping?: boolean`` for creation of fulfillment for digital products.


420coupe:
reason for this is to be able to createFulfillmentWorkFlows that do not require shipping, so have to modify the input to accept the optional variable

carlos-r-l-rodrigues (approved):

PR #9481, opened 2024-10-06
docs: fix import path in the workflow basic documentation

damien-thiesson:
The import is three levels up, but, the `src/api/workflow/route.ts` file is only two levels down from `src/workflow`.

Screenshot 2024-10-06 at 6 09 38 PM


shahednasser (approved):
Nice catch thanks! 🙏

PR #9496, opened 2024-10-08
breaking: Standalone builds

thetutlage:
Fixes: FRMW-2742

In this PR, we fix the build output of the backend source code, which eliminates a lot of magic between the development and production environments.

Right now, we only compile the source files from the `src` directory and write them within the `dist` directory.

**Here's how the `src` directory with a custom module looks like**

```
src
├── modules
│   └── hello
│       ├── index.ts
```

**Here's the build output**

```
dist
├── modules
│   └── hello
│       ├── index.js
```

Let's imagine a file at the root of your project (maybe the `medusa-config.js` file) that wants to import the `modules/hello/index` file. How can we ensure that the import will work in both the development and production environments?

If we write the import targeting the `src` directory, it will break in production because it should target the `dist` directory.

## Solution
The solution is to compile everything within the project and mimic the file structure in the build output, not just the `src` directory.

**Here's how the fixed output should look like**

```
dist
├── src
│  ├── modules
│  │   └── hello
│  │       ├── index.js
├── medusa-config.js
├── yarn.lock
├── package.json
```

If you notice carefully, we also have `medusa-config.js`, `yarn.lock`, and `package.json` within the `dist` directory. We do so to create a standalone built application, something you can copy/paste to your server and run without relying on the original source code.

- This results in small containers since you are not copying unnecessary files.
- Clear distinction between the development and the production code. If you want to run the production server, then `cd` into the `dist` directory and run it from there.

## Changes in the PR

- Breaking: Remove the `dist` and `build` folders. Instead, write them production artefacts within the `.medusa` directory as `.medusa/admin` and `.medusa/server`.
- Breaking: Change the output of the `.medusa/server` folder to mimic the root project structure.
- Refactor: Remove `Symbol.for("ts-node.register.instance")]` check to find from where to load the source code.
- Refactor: Use `tsc` for creating the production build. This ensures we respect `tsconfig` settings when creating the build and also perform type-checking.


packages/core/utils/src/common/normalize-import-path-with-source.ts
```diff
@@ -16,9 +16,15 @@ export function normalizeImportPathWithSource(
...
-      : "dist"
+    /**
+     * If someone is using the correct path pointing to the "src" directory
+     * then we are all good. Otherwise we will point to the "src" directory.
+     *
+     * In case of the production output. The app should be executed from within
+     * the "./build" directory and the "./build" directory will have the
+     * "./src" directory inside it.
+     */
+    let sourceDir = normalizePath.startsWith("./src") ? "./" : "./src"
```
olivermrbl, line 21: q: shouldn't this always be `src`?

packages/core/utils/src/common/normalize-import-path-with-source.ts
```diff
@@ -16,9 +16,15 @@ export function normalizeImportPathWithSource(
    * happen under the hood.
```
olivermrbl: todo: we should clean up the comments in this util

olivermrbl (commented):
Nice work so far!

packages/core/utils/src/common/normalize-import-path-with-source.ts
thetutlage: In the existing projects, the module resolve string looks as `modules/hello`, basically not including the `src` prefix. In this check, I am making sure that we do not break those existing projects in development. 

Lemme know if its still not clear, we can do a huddle

olivermrbl: Ah yes ofc, that's nice.

packages/core/types/src/common/config-module.ts
```diff
@@ -48,7 +48,7 @@ export interface AdminOptions {
    * })
    * ```
    */
-  path?: `/${string}`
+  path: `/${string}`
```
kasperkristensen, line 51: Won't this result in `path` being required if you want to set any admin options:

```
export default defineConfig({
  admin: {
     path: "/app", <-- This is now required even though I just want to use the default value
   }
})

File: packages/core/types/src/common/config-module.ts [hunk 4]
```diff
@@ -63,7 +63,7 @@ export interface AdminOptions {
    * })
    * ```
    */
-  outDir?: string
+  outDir: string
```
kasperkristensen, line 66: Same for this

packages/core/types/src/common/config-module.ts
thetutlage: Ohh yeah. Thanks for flagging that.

We basically want to get rid of the defaults we have in multiple places and instead put them within the `defineConfig` and that means the output of `defineConfig` should contain these properties.

But as you mentioned, that makes it required when you define `admin` property. So lemme make it a `DeepPartial` when receiving the input.

thetutlage: Should be fixed by this commit. https://github.com/medusajs/medusa/pull/9496/commits/25cd88a087430f8e0754c2dd1209355de7ce724d

adrien2p (approved):

adrien2p:
should we add a note on the changes to make to the medusa-config for the users, now they can specify `resolve: './src/modules/myModule`

thetutlage:
> should we add a note on the changes to make to the medusa-config for the users, now they can specify resolve: './src/modules/myModule'

Yeah. But I have kept it backward compatible too :)

adrien2p:
> > should we add a note on the changes to make to the medusa-config for the users, now they can specify resolve: './src/modules/myModule'
> 
> Yeah. But I have kept it backward compatible too :)

Maybe we can add a note just so that we don't keep it for too long :)

thetutlage:
All seems to be good. Let's merge it

olivermrbl:
Another question (maybe this is already dealt with): Can we do so that I can execute `medusa start` from the root of the project, and then it starts the production server?

thetutlage:
@olivermrbl Is there any reason you want to do that?

joekendal:
for some reason i can no longer directly load path urls in the browser, it has to be navigated to manually from /. this is a serious bug

PR #9497, opened 2024-10-08
fix(order): undo order change

carlos-r-l-rodrigues:
FIXES: CC-573

### DIFF
--- a/packages/modules/order/src/services/order-module-service.ts
+++ b/packages/modules/order/src/services/order-module-service.ts
@@ -2335,6 +2335,7 @@ export default class OrderModuleService<
   ): Promise<void> {
     const currentVersion = order.version
 
+    const updatePromises: Promise<any>[] = []
     // Order Changes
     const orderChanges = await this.orderChangeService_.list(
       {
@@ -2353,7 +2354,9 @@ export default class OrderModuleService<
       }
     })
 
-    await this.orderChangeService_.update(orderChangesIds, sharedContext)
+    updatePromises.push(
+      this.orderChangeService_.update(orderChangesIds, sharedContext)
+    )
 
     // Order Changes Actions
     const orderChangesActions = await this.orderChangeActionService_.list(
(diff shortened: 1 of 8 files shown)


packages/modules/order/src/services/order-module-service.ts
```diff
@@ -2302,6 +2302,137 @@ export default class OrderModuleService<
...
+        version: currentVersion,
+      },
+      { select: ["id", "version"] },
+      sharedContext
+    )
+
+    const orderChangesIds = orderChanges.map((change) => {
+      return {
+        id: change.id,
+        status: OrderChangeStatus.PENDING,
```
fPolic: q: why are we setting it to `PENDING`?

In a scenario I tested, the order edit is `REQUESTED` and when I try to force confirm which fails due to low stock quantity the edit is reverted to pending instead of requested status. It's not a big deal from the admin standpoint since users can just continue the edit and rerequest again but just curious if we can revert the status as well?

carlos-r-l-rodrigues: I haven't thought about it, I believe to reset this, we'll need the previous state of it.
I will rework that part.

packages/modules/order/src/services/order-module-service.ts
```diff
@@ -2302,6 +2303,132 @@ export default class OrderModuleService<
...
+        selector: {
+          id: order.id,
+        },
+        data: {
+          version: order.version - 1,
+        },
+      },
+      sharedContext
+    )
+  }
```
adrien2p, line 2445: Q: is there any of the update that could be concurrently done or do they rely on each other?

adrien2p (approved):

packages/modules/order/src/services/order-module-service.ts
carlos-r-l-rodrigues: All of them can be in parallel or most of them. Same with revertLastChange method.
I will update it, even though mikro-orm is doing that under the hood anyway.

adrien2p: yeah I was considering that mikro orm does it fo us, but just from an implementation pov I prefer to be explicit in case for some reason a flush happen in the middle or tomorrow people base their impl on that while being on an external module using another tool that does not do it for them, that way they can see best practices :)

fPolic (approved):
LGTM - nice!

PR #9502, opened 2024-10-08
fix(core-flows, dashboard): inventory kit reservations

fPolic:
**What**
- fix `prepareConfirmInventory` to account for inventory kit items
  - _note: this step is reused in the complete cart and all RMA flows_ 
- properly remove reservations for items that are removed from the order edit
- invalidate inventory/reservations cache when order edit is confirmed

---

https://github.com/user-attachments/assets/f12e9198-0718-4c08-bd81-efc536eca146

---

FIXES CC-565

### DIFF
--- a/packages/core/core-flows/src/cart/utils/prepare-confirm-inventory-input.ts
+++ b/packages/core/core-flows/src/cart/utils/prepare-confirm-inventory-input.ts
@@ -28,7 +28,7 @@ interface ConfirmInventoryItem {
   inventory_item_id: string
   required_quantity: number
   allow_backorder: boolean
-  quantity: number
+  quantity: BigNumberInput
   location_ids: string[]
 }
 
@@ -160,7 +160,7 @@ const formatInventoryInput = ({
         inventory_item_id: variantInventoryItem.inventory_item_id,
         required_quantity: variantInventoryItem.required_quantity,
         allow_backorder: !!variant.allow_backorder,
-        quantity: item.quantity as number, // TODO: update type to BigNumberInput
+        quantity: item.quantity,
         location_ids: location_ids,
       })
     )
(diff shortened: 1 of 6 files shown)


christiananese (approved):
👍

packages/core/core-flows/src/cart/utils/prepare-confirm-inventory-input.ts
```diff
@@ -143,25 +143,27 @@ const formatInventoryInput = ({
...
-      quantity: item.quantity as number, // TODO: update type to BigNumberInput
-      location_ids: location_ids,
-    })
+    variantInventoryItems.forEach((variantInventoryItem) =>
+      itemsToConfirm.push({
+        id: item.id,
+        inventory_item_id: variantInventoryItem.inventory_item_id,
+        required_quantity: variantInventoryItem.required_quantity,
+        allow_backorder: !!variant.allow_backorder,
+        quantity: item.quantity as number, // TODO: update type to BigNumberInput
```
riqwan: You can do this:

```suggestion
        quantity: MathBn.convert(item.quantity),
```

riqwan (approved):

packages/core/core-flows/src/order/workflows/order-edit/confirm-order-edit-request.ts
```diff
@@ -133,8 +133,14 @@ export const confirmOrderEditRequestWorkflow = createWorkflow(
       throw_if_key_not_found: true,
     }).config({ name: "order-items-query" })
 
-    const lineItemIds = transform({ orderItems }, (data) =>
-      data.orderItems.items.map(({ id }) => id)
+    const lineItemIds = transform(
+      { orderItems, previousOrderItems: order.items },
+
+      (data) => {
+        const previousItemIds = data.previousOrderItems!.map(({ id }) => id) // items that have been removed with the change
```
adrien2p: suggestion: just for certainty instead of doing `previousOrderItems!` would we do `previousOrderItems?` just to not end up in a place where it is not present?

adrien2p (approved):

PR #9504, opened 2024-10-08
feat(dashboard): Add global search

kasperkristensen:
**What**
- Adds the ability to do global searches from cmd + k in the admin.
- The solution is temporary, until we have a proper search API.

**Note**
I have deviated a bit from the design, due to the constraints of this temporary solution:
- We don't have nested items, such as showing variants under a product (don't think having a proper search API will make this any easier, and not entirely sure how we would handle this for cases where a query returns multiple products, which is the only case that is designed)
- I have added a "Load {{count}} more" button instead of doing infinite scrolling, I am assuming the later is the intended behaviour based on the design file, but with 20+ sources of data changing so often it was resulting in some weird behaviours, so settled for the simpler approach for this temporary solution.
- Removed the "Details" label on search results as it seemed a bit repetitive
- I haven't added icons for the different types of search results, as there are only a couple of examples in the design doc, and I wasn't sure what to pick for all the different types of results. If we want to add icons, then I think it's something we can add when we revisit this later, but think its fine to omit, as each group of results is labeled, so they are easy to tell apart.

Resolves CC-574


olivermrbl (approved):
LGTM, tested and works as expected.

One alternative to this approach is that we move the queries to the backend and introduce a `POST /admin/index?` endpoint. It wouldn't save us for any DB queries, but it would eliminate the 23 network requests and hide the inelegancy behind a single endpoint.

I am fine keeping this as is, but thought I'd share.

packages/admin/dashboard/src/components/search/use-search-results.tsx
```diff
@@ -0,0 +1,750 @@
...
+import { useGlobalShortcuts } from "../../providers/keybind-provider/hooks"
+import { DynamicSearchResult, SearchArea } from "./types"
+
+type UseSearchProps = {
+  q?: string
+  limit: number
+  area?: SearchArea
+}
+
+export const useSearchResults = ({
```
fPolic, line 39: I noticed that sometimes results are shown only while typing i.e. a fetch is running. If I type an email from an order, results are shown only while the loader is visible and when fetching is done "no results" is rendered.

fPolic (approved):
LGTM - clean!

kasperkristensen:
> LGTM, tested and works as expected.
> 
> One alternative to this approach is that we move the queries to the backend and introduce a `POST /admin/index?` endpoint. It wouldn't save us for any DB queries, but it would eliminate the 23 network requests and hide the inelegancy behind a single endpoint.
> 
> I am fine keeping this as is, but thought I'd share.

We could do that, but wouldn't that just set us up for introducing a breaking change in a not to distant future? I imagine that once we have our index engine, then we would introduce this exact endpoint.

olivermrbl:
> > LGTM, tested and works as expected.
> > One alternative to this approach is that we move the queries to the backend and introduce a `POST /admin/index?` endpoint. It wouldn't save us for any DB queries, but it would eliminate the 23 network requests and hide the inelegancy behind a single endpoint.
> > I am fine keeping this as is, but thought I'd share.
> 
> We could do that, but wouldn't that just set us up for introducing a breaking change in a not to distant future? I imagine that once we have our index engine, then we would introduce this exact endpoint.

The idea would be to introduce an endpoint that will later on be used for the index engine. 

So it would be something like:

**Search across all resources**
```
POST /admin/search
{ q: "shirt" }
```

**Search for specific resources**
```
POST /admin/search
{ q: "shirt", scopes: ["product", "product_variant"] }
```

And the API Route would iterate over the `scopes` (if defined) and perform a `.graph` request for each and return a grouped response to the client. 

However, as I write this out, I am wondering if that will cause issues with pagination.

As I said, I am fine keeping this PR as is. If we see issue reports with the current approach, we can revisit and otherwise wait until we can properly think through a dedicated search API.

PR #9507, opened 2024-10-09
fix(workflows-sdk):transaction id inheritence

adrien2p:
**What**
Fix the broken inheritence of the parent workflow transaction id


carlos-r-l-rodrigues (approved):

olivermrbl (approved):

carlos-r-l-rodrigues:
This has caused something to run infinitely, maybe a test has set a way too big timeout and it is preventing jest to fail.
This change can also potentially cause issues if we have a recursive workflow, where a flow calls itself. Having the same transactionId will have unexpected behavior.

adrien2p:
> This has caused something to run infinitely, maybe a test has set a way too big timeout and it is preventing jest to fail. This change can also potentially cause issues if we have a recursive workflow, where a flow calls itself. Having the same transactionId will have unexpected behavior.

wouldn't circular workflow execution be an anti pattern though? it would create an infinite loop as far as I can see

PR #9511, opened 2024-10-09
fix: hover states on filters and chip groups

srindom:
### What

**Hovering a button now changes background of entire "x"-area**
https://github.com/user-attachments/assets/b77e7436-c3c1-4bfc-a930-efbe24fd15dc
Previously only a small square behind the x would hold a hover state.

**Hover states on filter displays**
https://github.com/user-attachments/assets/a1c0a52b-28d9-46b9-a389-ce87ca942fa3

A few modifications:
1. The popover trigger is now only active on the value segment of a filter group.
2. Only the value segment of a filter group is clickable and has a hover state.
3. Similar to the chip group the entire "close" area segment of the group is active.
4. Introduced a Popover Anchor so that when a filter is applied the popover will first show from the beginning of the filter group. The next time it will show from the value segment of the filter group.

Fixes CC-534
Fixes CC-108


fPolic (approved):
LGTM

PR #9519, opened 2024-10-10
docs: Fix link to the correct AuthIdentity model's page

zaidrashid:
I think there is a bug on the link to the `AuthIdentity ` model's page. The page containing the issue can be found [here](https://docs.medusajs.com/v2/resources/commerce-modules/auth/auth-identity-and-actor-types#what-is-an-auth-identity)


shahednasser (approved):
Nice catch thanks!

PR #9520, opened 2024-10-10
fix(dashboard): cleanup Payments section

fPolic:
**What**
- fix content breaking
- use a shorter payment ID in the UI

---

FIXES CC-572


packages/admin/dashboard/src/routes/orders/order-detail/components/order-payment-section/order-payment-section.tsx
```diff
@@ -176,14 +176,14 @@ const Payment = ({
     payment.captured_at ? ["Captured", "green"] : ["Pending", "orange"]
   ) as [string, "green" | "orange"]
 
-  const cleanId = payment.id.replace("pay_", "")
+  const cleanId = `#${payment.id.replace("pay_", "").slice(-7)}`
```
olivermrbl: suggestion: Not confident this is the right solution. Perhaps we can revert and just omit the ID from the sentence, so it will be "Payment is ready to be captured" and we keep the ID as stored in the DB. I also don't think we should remove `pay_`.

fPolic: In general, displaying entities that don't have a title is a bit more challenging, especially on mobile. With RMA flows we introduced this pattern of displaying entity + abbr. id: `Exhange #7asdfad`, `Claim #5sdf8sd` etc. We could reuse that for similar scenarios such as this one and with addition of adding "copy on click" or display the entire id on hover so we still provide the id from the DB to the user, wdyt?

adrien2p: I do agree with oli, here we kind of loose an information right?

olivermrbl: I would prefer to have the entire ID (with `pay_`) of the payment here. Ideally, we change the exchange and claim IDs to be display IDs, but we can do t hat further down the line

fPolic: Alright, will push a change

srindom: I am actually pretty on board with this initial approach taken here - see it as similar to git commit hashes. In GitHub you typically only see the first N characters and then you can hover/click to view more/copy. 

Similarly, we here take the random part of the ULID as a display value, but still make the rest of the id available.

olivermrbl: > In GitHub you typically only see the first N characters and then you can hover/click to view more/copy.

I would agree with you if our IDs didn't have prefixes. But IMO, in this case, we are omitting key information about the entity with which the ID is associated and robbing ourselves for a neat feature. It's not a big deal nor is this a strong opinion, but I just like the Stripe-approach better.

We can keep it as is 👍

olivermrbl:
> We could reuse that for similar scenarios such as this one and with addition of adding "copy on click" or display the entire id on hover so we still provide the id from the DB to the user, wdyt?

@fPolic, should we add this before we merge?

packages/admin/dashboard/src/routes/orders/order-detail/components/order-payment-section/order-payment-section.tsx
thetutlage: Actually with Git, the short hash is useful in itself. Not sure if the same with true with our IDs

fPolic:
@oli sorry forgot about todo. I will add that in.

fPolic:
https://github.com/user-attachments/assets/eeaea20b-b383-401b-8fbd-04c44f0b08fc

@olivermrbl wdyt?

olivermrbl:
@fPolic, should we add a micro-interaction to indicate you can copy it?

See pub keys below:

https://github.com/user-attachments/assets/e9d94411-d3ae-4281-8465-7510da403ac3


---

Update: I realise you can't see my cursor, but it shows as a pointer. I was not referring to the Copy tooltip. 😄

olivermrbl (approved):
LGTM

PR #9521, opened 2024-10-10
breaking: remove deprecated commands and code

thetutlage:
Fixes: FRMW-2736

This PR removes deprecated commands and code from the Medusa monorepo. Here's a quick summary of it. Please, feel free to review this PR thoroughly to ensure I am not breaking something by mistake :)

### Commands removal

- `seed`: There was no implementation for the `seed` command in the first place.
- `migrations [action]`: Removed in favor of `db:migrate`, `db:rollback` and `db:generate`.
- `links [action]`: Removed in favor of `db:sync-links`.
- `start-cluster`: Removed `start-cluster` command in favor of the `--cluster` flag on the start command.

### Config options removal

- Remove config option `databaseType` as it is not used anywhere.
- Remove config option `httpCompression` in favor of `compression` property.
- Remove `allowedFields` in favor of `allowed` property from the `MedusaRequest`.
- Remove `expand` property in favor of `allowed` property from the `RequestQueryFields`.
- Remove `defaultFields` and `defaultRelations` in favor of `defaults` property from the `RequestQueryFields`.
- Remove `allowedFields` and `allowedRelations` in favor of `allowed` property from the `RequestQueryFields`.

### Events build changes

- Remove `eventsEnum` in favor of `eventName` property accepted by the `moduleEventBuilderFactory`. Also updated all the usages to use `eventName`.

### ModelDTO changes

- Remove unused properties `singular` and `plural` from `ModelDTOConfig`.

### Models template changes

- Remove unused properties `create`, `update`, `singular`, and `plural` from `ModelsConfigTemplate`.

### Zod validator changes

- Remove unused `OptionalBooleanValidator` in favor of `booleanString`.


packages/cli/medusa-cli/src/create-cli.ts
```diff
@@ -422,42 +362,6 @@ function buildLocalCommands(cli, isLocalProject) {
         })
       ),
     })
-    .command({
```
adrien2p, line 425: todo: as discussed this should be added back for users self hosting on a single machine to benefit from their resources while increasing capabilities 👍

packages/core/framework/src/http/types.ts
```diff
@@ -123,12 +123,7 @@ export interface MedusaRequest<Body = unknown>
...
-  /**
-   * An array of fields and relations that are allowed to be queried, this can be set by the
-   * consumer as part of a middleware and it will take precedence over the defaultAllowedFields
-   * @deprecated use `allowed` instead
-   */
-  allowedFields?: string[]
+
   /**
    * An array of fields and relations that are allowed to be queried, this can be set by the
    * consumer as part of a middleware and it will take precedence over the defaultAllowedFields set
```
adrien2p: suggestion: we might need to update the description slightly, and maybe mention req.allowed explicitly wdyt?

packages/core/framework/src/http/utils/get-query-config.ts
```diff
@@ -29,27 +28,15 @@ export function prepareListQuery<T extends RequestQueryFields, TEntity>(
...
-    allowedFields = [],
-    allowedRelations = [],
-    defaultRelations = [],
-    isList,
-  } = queryConfig
-
-  allowedFields = allowed.length ? allowed : allowedFields
-  defaultFields = defaults.length ? defaults : defaultFields
+  const { order, fields, limit = 50, offset = 0 } = validated
+  let { allowed = [], defaults = [], defaultLimit, isList } = queryConfig
```
adrien2p: todo: the defaultLimit seems to never be used because we have a default value on limit, I would suggest to inverse the default value and set it to the default limit instead

packages/core/framework/src/http/utils/get-query-config.ts
```diff
@@ -174,8 +136,8 @@ export function prepareListQuery<T extends RequestQueryFields, TEntity>(
     }
 
     if (
-      queryConfig?.allowedFields?.length &&
-      !queryConfig?.allowedFields.includes(orderField)
+      queryConfig?.allowed?.length &&
+      !queryConfig?.allowed.includes(orderField)
```
adrien2p:
```suggestion
     allowed.length &&
      !allowed.includes(orderField)
```

packages/core/utils/src/modules-sdk/medusa-service.ts
```diff
@@ -464,7 +464,7 @@ export function MedusaService<
     }: {
       action: string
       object: string
-      eventName?: string
+      eventName: string
```
adrien2p: todo: optional

adrien2p:
Should we include the following
- packages/core/utils/src/modules-sdk/definition.ts

packages/cli/medusa-cli/src/create-cli.ts
riqwan: Suggestion to group the commands:

`medusa server:dev`
`medusa server:start`
`medusa server:start --cluster`

Or just:

`medusa server:start --dev --cluster`

packages/core/framework/src/http/types.ts
```diff
@@ -123,12 +123,7 @@ export interface MedusaRequest<Body = unknown>
    */
   filterableFields: Record<string, unknown>
   includes?: Record<string, boolean>
-  /**
-   * An array of fields and relations that are allowed to be queried, this can be set by the
-   * consumer as part of a middleware and it will take precedence over the defaultAllowedFields
-   * @deprecated use `allowed` instead
-   */
-  allowedFields?: string[]
+
```
riqwan, line 126: sugg: 

While you're here, you can also remove these from the request:

```
allowedProperties, listConfig, retreiveConfig, validatedQuery, validatedQuery (potentially)
```

packages/core/framework/src/http/utils/get-query-config.ts
riqwan: can drop the `allowed.length` as its defaulting to an array

riqwan (approved):

packages/cli/medusa-cli/src/create-cli.ts
thetutlage: Indeed that makes sense!

packages/core/framework/src/http/utils/get-query-config.ts
thetutlage: Done

packages/core/framework/src/http/utils/get-query-config.ts
thetutlage: Done

packages/core/framework/src/http/utils/get-query-config.ts
adrien2p: no we shouldnt drop the length as otherwise it will never include the order fields when no allowed are provided, instead when nothing is provided we don't want to check it

packages/core/framework/src/http/types.ts
thetutlage: What are the alternative properties for these?

thetutlage:
> Should we include the following
> 
> * packages/core/utils/src/modules-sdk/definition.ts

@adrien2p  Include where?

packages/core/framework/src/http/types.ts
thetutlage: Not sure I get it, which comment should I update?

packages/core/framework/src/http/types.ts
riqwan: `listConfig, retreiveConfig` => `fields`
`validatedQuery` => `filterableFields`

`allowedProperties` don't know what this is. We don't seem to be using them anywhere

packages/core/utils/src/modules-sdk/medusa-service.ts
thetutlage: Done

packages/core/framework/src/http/types.ts
adrien2p: this one where we mention among other things the `defaultAllowedFields` no?
```text
An array of fields and relations that are allowed to be queried, this can be set by the
   * consumer as part of a middleware and it will take precedence over the defaultAllowedFields set
   ```

File: packages/core/framework/src/http/types.ts (hunk 6 above)
Comment (adrien2p) [reply packages/core/framework/src/http/types.ts:126]: it was to configure what can be returned in v1, it was used to clean up the response before sending it back to the customer I think 😂

**Comment by adrien2p** [comment]
> > Should we include the following
> > 
> > * packages/core/utils/src/modules-sdk/definition.ts
> 
> @adrien2p Include where?

oh no worries I think you did it already to remove the deprecated inventory and stock

File: packages/core/framework/src/http/types.ts (hunk 2 above)
Comment (thetutlage) [reply packages/core/framework/src/http/types.ts]: Done

File: packages/core/framework/src/http/types.ts (hunk 6 above)
Comment (thetutlage) [reply packages/core/framework/src/http/types.ts:126]: I think, we will have to do a huddle for this, because I cannot find any sibling property named `fields` that has the similar shape as `listConfig`. Maybe I am looking at the wrong place :)

Comment (riqwan) [reply packages/core/framework/src/http/types.ts:126]: Sure, happy to. Also, feel free to move on, this isn't important to address in this PR.

In v1, we used `listConfig` and `retrieveConfig` to access what is `remoteQueryConfig.fields` today. The `remoteQueryConfig.*` object makes the previous 2 config's redundant as its the v2 version of `listConfig` and `retrieveConfig`.

File: packages/medusa/src/commands/start.ts [hunk 7]
```diff
@@ -133,7 +142,54 @@ async function start({ port, directory, types }) {
     }
   }
 
-  await internalStart()
+  /**
+   * When the cluster flag is used we will start the process in
+   * cluster mode
+   */
+  if ("cluster" in args) {
```
thetutlage, line 149: @adrien2p and @riqwan . Added support for cluster within the `start` command.

```sh
# starts in cluster mode with Max cpus
npx medusa start --cluster 

# starts in cluster mode with defined cpus
npx medusa start --cluster=2 
```

packages/cli/medusa-cli/src/create-cli.ts
thetutlage: Done. But did not move the commands to a namespace as that might require further discussions

packages/medusa/src/commands/start.ts
riqwan: nice! 🔥 looks good!

packages/core/framework/src/http/utils/get-query-config.ts
thetutlage: Wow! I wasted 2hours on this. Why did I not read your comment 🤦‍♂️

packages/cli/medusa-cli/src/create-cli.ts
```diff
@@ -397,6 +336,12 @@ function buildLocalCommands(cli, isLocalProject) {
             describe: process.env.PORT
               ? `Set port. Defaults to ${process.env.PORT} (set by env.PORT) (otherwise defaults ${defaultPort})`
               : `Set port. Defaults to ${defaultPort}`,
+          })
+          .option(`cluster`, {
```
olivermrbl, line 340: nice!

olivermrbl (approved):
Very, very nice

carlos-r-l-rodrigues (approved):
LGTM.

not related to this PR, but the `singular` `plural` properties at some point will have to be added somewhere, as we auto generate methods based on them, non english names won't work, and non standard english plurals won't work properly.

PR #9522, opened 2024-10-10
chore(): Allow to register modules through array

adrien2p:
**What**
Currently, in the medusa-config, the modules are expected to be registered in an object. While this API is maintain backward compatible, it is now possible to register them using an array. The idea being to have a better consistency between plugins and modules registration.

```diff
// before

-modules: {
-  GithubModuleService: {
-    resolve: "./modules/github",
-      options: {
-      apiKey: "test",
-    },
-  },
-},
+modules: [
+ {
+    // key: string // an optional key property to force the module key
+    // disable: boolean // to enable disabling core modules
+    resolve:  "./modules/github",
+    options: {
+      apiKey: "test",
+    },
+  },
+],
```

**Note**
The object API remain backward compatible but is being deprecated

The underlying API remains unchanged, only the user-facing API of the config is being changed. From that, we manipulate the config to conform to the underlying expectation.

**Breaking changes**

The import location changed for the following utils
```diff
- import { MODULE_PACKAGE_NAMES } from "@medusajs/framework/modules-sdk
+ import { MODULE_PACKAGE_NAMES } from "@medusajs/framework/utils
```

The value has changed for the following const.

```diff
- export const Modules = {
-   AUTH: "Auth",
-   CACHE: "Cache",
-   CART: "Cart",
-   CUSTOMER: "Customer",
-   EVENT_BUS: "EventBus",
-   INVENTORY: "Inventory",
-   LINK: "LinkModules",
-   PAYMENT: "Payment",
-   PRICING: "Pricing",
-   PRODUCT: "Product",
-   PROMOTION: "Promotion",
-   SALES_CHANNEL: "SalesChannel",
-   TAX: "Tax",
-   FULFILLMENT: "Fulfillment",
-   STOCK_LOCATION: "StockLocation",
-   USER: "User",
-   WORKFLOW_ENGINE: "Workflows",
-   REGION: "Region",
-   ORDER: "Order",
-   API_KEY: "ApiKey",
-   STORE: "Store",
-   CURRENCY: "Currency",
-   FILE: "File",
-   NOTIFICATION: "Notification",
-   INDEX: "Index",
- } as const
+ export const Modules = {
+   AUTH: "auth",
+   CACHE: "cache",
+   CART: "cart",
+   CUSTOMER: "customer",
+   EVENT_BUS: "event_bus",
+   INVENTORY: "inventory",
+   LINK: "link_modules",
+   PAYMENT: "payment",
+   PRICING: "pricing",
+   PRODUCT: "product",
+   PROMOTION: "promotion",
+   SALES_CHANNEL: "sales_channel",
+   TAX: "tax",
+   FULFILLMENT: "fulfillment",
+   STOCK_LOCATION: "stock_location",
+   USER: "user",
+   WORKFLOW_ENGINE: "workflows",
+   REGION: "region",
+   ORDER: "order",
+   API_KEY: "api_key",
+   STORE: "store",
+   CURRENCY: "currency",
+   FILE: "file",
+   NOTIFICATION: "notification",
+   INDEX: "index",
+ } as const
```

⚠️ for the above, it will change the event name for event using the module key as the source

**Resolution**

someone author a module like so
```ts
export const MyCustomModuleKey = 'myCustomService'
export default Module(MyCustomModuleKey, { ... } )
```

as the consumer you will register it like this
```ts
{
  modules: [
    { resolve: 'my-custom-module' }
  ]
}
```

and then you want to resolve it

```ts
const myCustomModule = req.scope.resolve(MyCustomModuleKey)

// or

const myCustomModule = req.scope.resolve('myCustomService')
```

now, the module registration `key` default to the joiner config service name unless a specific key is set during the module registration in the medusa config


olivermrbl:
Really like this guys!

How do we deal with module registration keys? Do we default to the one used in the module export?

adrien2p:
> Really like this guys!
> 
> How do we deal with module registration keys? Do we default to the one used in the module export?

yes by default, we grab the joiner config service name, which is populated from the module key used when consuming the `Module` util. so most of the time, the registration key in the container will be the same as the service name in the joiner config unless If the user provide the key prop, then we will use this value instead to register in the container

olivermrbl:
> > Really like this guys!
> > How do we deal with module registration keys? Do we default to the one used in the module export?
> 
> yes by default, we grab the joiner config service name, which is populated from the module key used when consuming the `Module` util. so most of the time, the registration key in the container will be the same as the service name in the joiner config unless If the user provide the key prop, then we will use this value instead to register in the container

Very nice. 

One suggestion I have (that we've talked about before) is to use the `Modules.[MODULE_NAME]` as registration key, so that you would register and resolve it like so:
```
// in module
export default Module(Modules.PRODUCT, {
  service: ProductModuleService,
})

// in e.g. API route
const productModule = req.scope.resolve(Modules.PRODUCT)
```

What do you think about that?

olivermrbl:
That would also mean, you can use strings like so:
```
const productModule = req.scope.resolve("product")
```

adrien2p:
Yes @olivermrbl  thats what it is doing, did I miss something?

lets' take this example:

someone author a module like so
```ts
export const MyCustomModuleKey = 'myCustomService'
export default Module(MyCustomModuleKey, { ... } )
```

as the consumer you will do 
```ts
{
  modules: [
    { resolve: 'my-custom-module' }
  ]
}
```

and then you want to resolve it

```ts
const myCustomModule = req.scope.resolve(MyCustomModuleKey)

// or

const myCustomModule = req.scope.resolve('myCustomService')
```

Also, this pr in that sense helps align the resolution key and joiner co fig service name, which later will always be the same to reduce confusion

adrien2p:
Maybe i will update the description to add an explanation about resolution cases 😅

olivermrbl:
> Yes @olivermrbl thats what it is doing, did I miss something?
> 
> lets' take this example:
> 
> someone author a module like so
> 
> ```ts
> export const MyCustomModuleKey = 'myCustomService'
> export default Module(MyCustomModuleKey, { ... } )
> ```
> 
> as the consumer you will do
> 
> ```ts
> {
>   modules: [
>     { resolve: 'my-custom-module' }
>   ]
> }
> ```
> 
> and then you want to resolve it
> 
> ```ts
> const myCustomModule = req.scope.resolve(MyCustomModuleKey)
> 
> // or
> 
> const myCustomModule = req.scope.resolve('myCustomService')
> ```

Sorry, I actually didn't know we had changed this already (lots of stuff shipping 🚢 ). 

What I would like to propose to add to this is that we camelCase (or snake_case) **our core module keys** and make that the recommended approach for defining module exports in custom modules:
```
const productModule = req.scope.resolve("product")
const stockLocation = req.scope.resolve("stockLocation") // or req.scope.resolve("stock_location")
```

I have rarely seen PascalCase for these use cases, and it feels a bit off. If our documentation uses an example like the one you linked above, there will be an inconsistency between our modules' registration keys and the custom modules' registration keys.

--

On argument for snake_case is that we use it to create links and it would therefore follow that naming convention:
```ts
const link = await remoteLink.create({
  product: { product_id: "prod_1234" },
  sales_channel: { sales_channel_id: "sc_1234" }
})
```

As well as events:
```
`product.product_variant.created`
```

I believe we are currently performing some casing changes when emitting events, and with this change, we could eliminate that.

Let me know if there are any cases where a change of casing wouldn't make sense. Happy to discuss!

---

**Update**: We were talking past each other, discussion resolved internally and PR updated.

adrien2p:
We dont have it yet, it is included in that pr 😊 the casing you are referring too is link to entity and properties, not sure the way the user want to name their service should be constrained the same way though, also, wouldn't it create confusion to have product for both the service name and the entity? When you resolve product, are you resolving a module or an entity? I know the answer but wouldn't be the kind of question people would ask if everything is named the same?

Also, for the event i am not sure that we change the casing for the first segment which is the serviceName, it remains the same actually

adrien2p:
@olivermrbl I ve updated all the description, do see something I would have missed by any chance?

adrien2p:
@olivermrbl should we merge this one?

olivermrbl (approved):
LGTM, nice!

PR #9523, opened 2024-10-10
fix: js-sdk compilation error

joekendal:


sradevski (approved):

packages/core/js-sdk/src/admin/order.ts
```diff
@@ -4,7 +4,7 @@ import {
   PaginatedResponse,
   SelectParams,
 } from "@medusajs/types"
-import { AdminOrderChangesResponse } from "@medusajs/types/src/http/order/admin/responses"
+import { AdminOrderChangesResponse } from "@medusajs/types"
```
fPolic, line 7: should be moved to the import statement above since now we have duplicate import from "@medusajs/types"

PR #9524, opened 2024-10-10
feat(locking): Locking module

carlos-r-l-rodrigues:
**What**
- Locking Module to manage concurrency
- Default `in-memory` provider


packages/core/modules-sdk/src/medusa-app.ts
```diff
@@ -84,7 +84,7 @@ export async function loadModules(args: {
     sharedResourcesConfig,
     migrationOnly = false,
     loaderOnly = false,
-    workerMode = "server" as ModuleBootstrapOptions["workerMode"],
+    workerMode = "shared" as ModuleBootstrapOptions["workerMode"],
```
olivermrbl, line 87: q: What happened here?

packages/core/utils/src/modules-sdk/definition.ts
```diff
@@ -24,6 +24,7 @@ export const Modules = {
   FILE: "File",
   NOTIFICATION: "Notification",
   INDEX: "Index",
+  LOCKING: "Locking",
```
olivermrbl: todo: Should be `locking`
```suggestion
  LOCKING: "locking",
```

packages/modules/locking/src/loaders/providers.ts
```diff
@@ -0,0 +1,85 @@
...
+      }
+      container.register(LockingDefaultProvider, asValue(provider.id))
+    }
+  }
+
+  if (!hasDefaultProvider) {
+    logger.warn(
+      `No default Locking provider explicit defined, using "${container.resolve(
+        LockingDefaultProvider
+      )}" as default.`
```
olivermrbl:
```suggestion
      `No default locking provider defined. Using "${container.resolve(
        LockingDefaultProvider
      )}" as default.`
```

packages/modules/locking/src/types/index.ts
```diff
@@ -0,0 +1,40 @@
+import {
+  IEventBusModuleService,
+  Logger,
+  ModuleProviderExports,
+  ModuleServiceInitializeOptions,
+} from "@medusajs/framework/types"
+
+export type InitializeModuleInjectableDependencies = {
+  logger?: Logger
+  EventBus?: IEventBusModuleService
```
olivermrbl:
```suggestion
  eventBus?: IEventBusModuleService
```

packages/modules/locking/src/providers/in-memory.ts
```diff
@@ -0,0 +1,176 @@
...
+      if (lock.ownerId === ownerId) {
+        if (expire) {
+          lock.expiration = now + expire * 1000
+          this.locks.set(key, lock)
+        }
+
+        continue
+      }
+
+      if (lock.currentPromise && args?.awaitQueue) {
```
olivermrbl, line 112: q: can you briefly explain this `awaitQueue`?

olivermrbl (approved):
LGTM, a few comments. Can you add the module to the large changeset we have?

packages/core/modules-sdk/src/medusa-app.ts
carlos-r-l-rodrigues: this is supposed to be "shared" by default.
we have already other places setting this as "shared" if not provided, it is just to align with the real default.

packages/modules/locking/src/providers/in-memory.ts
carlos-r-l-rodrigues: the methods `execute` and `acquire` use the same mechanism. But the `execute` has to `awaitQueue` to be executed if the lock is acquired by another process, whereas the `acquire` will fail if the lock is not acquired.

olivermrbl: Got it thanks!

PR #9528, opened 2024-10-10
fix(core-flows, link-module): product <> inventory delete cascades

fPolic:
**What**
- remove cascade delete of inventory items on product delete
- implement inventory deletion in product/variant delete workflows with checks:
  - product/variant cannot be deleted if there are reservations associated with their inventory items
  - inventory item will be cascade deleted if it's not used by other variants (that are not being deleted in the current flow)

---

FIXES CC-581 CC-582


olivermrbl (approved):
LGTM 💪

packages/core/core-flows/src/inventory/workflows/delete-inventory-items.ts
```diff
@@ -15,6 +16,17 @@ export const deleteInventoryItemWorkflowId = "delete-inventory-item-workflow"
...
   (input: WorkflowData<string[]>): WorkflowResponse<string[]> => {
+    const inventoryItemsToDelete = useRemoteQueryStep({
+      entry_point: "inventory",
+      fields: ["id", "reserved_quantity"],
+      variables: {
+        id: input,
+      },
+      list: true,
+    })
+
```
srindom: suggestion: `useQueryGraphStep` instead?

packages/core/core-flows/src/product/workflows/delete-products.ts
```diff
@@ -27,6 +32,47 @@ export const deleteProductsWorkflow = createWorkflow(
         .map((variant) => variant.id)
     })
 
+    const variantsWithInventory = useRemoteQueryStep({
```
srindom: suggestion: `useQueryGraphStep`

integration-tests/http/__tests__/inventory/admin/inventory.spec.ts
```diff
@@ -782,7 +782,7 @@ medusaIntegrationTestRunner({
       })
 
       describe("DELETE /admin/inventory-items/:id", () => {
-        it("should remove associated levels and reservations when deleting an inventory item", async () => {
+        it("fails to remove inventory item with reservations", async () => {
```
adrien2p: nit: should fail to ...

packages/core/core-flows/src/inventory/steps/delete-inventory-items.ts
```diff
@@ -1,6 +1,27 @@
...
+  validateVariantInventoryStepId,
+  async (data: ValidateInventoryDeleteStepInput) => {
+    for (const inventoryItem of data.inventory_items) {
+      if (MathBN.gt(inventoryItem.reserved_quantity, 0)) {
+        throw new MedusaError(
+          MedusaError.Types.INVALID_DATA,
+          `Cannot remove inventory item: ${inventoryItem.id} which has reservations.`
+        )
+      }
+    }
```
adrien2p:
```suggestion
const nonDeletable = data.inventory_items.filter((inventoryItem) => {
  return MathBN.gt(inventoryItem.reserved_quantity, 0)
})

if (nonDeletable.length) {
  throw new MedusaError(
    MedusaError.Types.INVALID_DATA, `Cannot remove inventory items: [${nonDeletable.map((inventoryItem) => inventoryItem.id).join(', ')}] which have reservations.`
  )
}
```

packages/core/core-flows/src/product/workflows/delete-product-variants.ts
```diff
@@ -25,6 +30,47 @@ export const deleteProductVariantsWorkflow = createWorkflow(
...
+
+          for (const inventoryItem of variant.inventory) {
+            if (inventoryItem.variants.every((v) => variantsMap.has(v.id))) {
+              toDeleteIds.push(inventoryItem.id)
+            }
+          }
+        }
+
+        return Array.from(new Set(toDeleteIds))
+      }
```
adrien2p:
```suggestion
const toDeleteInventoryItemIds = transform(
  { variants: variantsWithInventory },
  (data) => {
    const variantsMap = new Set(data.variants.map((v) => v.id))

    const toDeleteIds: Set<string> = new Set()
    
    data.variants.forEach((variant) => {
      if (!variant.manage_inventory) {
        return
      }

      for (const inventoryItem of variant.inventory) {
        if (inventoryItem.variants.every((v) => variantsMap.has(v.id))) {
          toDeleteIds.add(inventoryItem.id)
        }
      }
    })

    return Array.from(toDeleteIds)
  }
)
```

packages/core/core-flows/src/product/workflows/delete-products.ts
```diff
@@ -27,6 +32,47 @@ export const deleteProductsWorkflow = createWorkflow(
...
+      variables: {
+        id: variantsToBeDeleted,
+      },
+      list: true,
+    })
+
+    const toDeleteInventoryItemIds = transform(
+      { variants: variantsWithInventory },
+      (data) => {
+        const variantsMap = new Map(data.variants.map((v) => [v.id, true]))
```
adrien2p: suggestion: same as the previous one

packages/core/core-flows/src/product/workflows/delete-products.ts
fPolic: @adrien2p what would be correct replacement for this with `useQueryGraphStep`?
Getting an empty array with this:
```ts
    const { data: variantsWithInventory } = useQueryGraphStep({
      entity: "variants",
      fields: [
        "id",
        "manage_inventory",
        "inventory.id",
        "inventory.variants.id",
      ],
      filters: {
        id: input.ids,
      },
    })
```

adrien2p: you can't destructure the result of the step because this is not the run time execution here but the composition time :)

srindom (approved):
LGTM!

PR #9532, opened 2024-10-11
fix: do not pass additional_data to service

thetutlage:
Fixes: FRMW-2743

This PR extracts and removes the `additional_data` from the workflow input before calling the steps and hence the `additional_data` is not passed down to the service layer.

However, this bug has made me discover one inconsistency in the input structure of certain workflows.

✅ **Following is the input structure of the `updateProductsWorkflow`**. Here, we accept the `products` and the `additional_data` as two separate top-level properties.

```ts
// Shape
export type UpdateProductsWorkflowInputSelector = {
  selector: ProductTypes.FilterableProductProps
  update: Omit<ProductTypes.UpdateProductDTO, "variants"> & {
    sales_channels?: { id: string }[]
    variants?: UpdateProductVariantWorkflowInputDTO[]
  }
} & AdditionalData

// Calling the workflow
const { result } = await updateProductsWorkflow(req.scope).run({
  input: {
    selector: { id: req.params.id },
    update,
    additional_data,
  },
})
```

❌ **Following in the input structure of the `updateCartWorflow`**. In this case, we are accepting the cart properties at the top-level, hence the `additional_data` is merged within those properties, increasing the chance of passing it down to the service layer by mistake.

```ts
// Shape
WorkflowData<UpdateCartWorkflowInputDTO & AdditionalData>

// Calling the workflow
await workflow.run({
  input: {
    ...req.validatedBody // Additional data is part of the validatedBody,
    id: req.params.id,
  },
})
```

Ideally, the input of `updateCartWorkflow` should look as follows.

```ts
WorkflowData<{ cart: UpdateCartWorkflowInputDTO } & AdditionalData>
```


carlos-r-l-rodrigues (approved):

packages/core/core-flows/src/cart/workflows/update-cart.ts
```diff
@@ -68,7 +68,11 @@ export const updateCartWorkflow = createWorkflow(
     const cartInput = transform(
       { input, region, customerData, salesChannel, cartToUpdate },
       (data) => {
-        const { promo_codes, ...updateCartData } = data.input
+        const {
+          promo_codes,
+          additional_data: _,
```
adrien2p, line 73: Q: any reason we create a new var `_` to store additional data here? wouldn't just destructure be enough here?

adrien2p (approved):

packages/core/core-flows/src/cart/workflows/update-cart.ts
olivermrbl: +1

olivermrbl (approved):

packages/core/core-flows/src/cart/workflows/update-cart.ts
carlos-r-l-rodrigues: That just removes additional_data property from updateCartData variable.
Elegant solution 😎

thetutlage: If you don't put `_` it will complain that the unstructured property is unused. We have disabled `noUnusedParameters` in our `tsconfig.json` so it might not be obvious right away.

https://www.typescriptlang.org/play/?allowUnusedLabels=true&noUnusedLocals=true&noUnusedParameters=true#code/KYDwDg9gTgLgBAYwgOwM7wJbLAV3gXjgG8AoORAQ1gC5iBfAGjLgoBNWMYMUKAbAfVYUYFWkTokJJJGnhFKsBi3aduyPoOEU4dOISy4Y0qjCA

adrien2p: I dont think it is true when you have a spread operator after the destructured properties since it can only happen at the end and therefore it can be expected that you have unused variables when you dont want them to be part of the aggregate

thetutlage: Ahuh, interesting. I didn't know that :)

PR #9534, opened 2024-10-11
fix(utils): Cascade soft deletion management

adrien2p:
**What**
Fix broken soft deletion depending on the object received


olivermrbl:
What was the broken scenario?

adrien2p:
> What was the broken scenario?

in some cases depending on the object, entity is not a function which was breaking the soft deletion entirely. Now it works in all cases. Also, since we haven't migrated our entire core to DML we didn't catch it and we didn't had tests on the DML for this. So I ve added it so that DML test the soft deletion itself as well

adrien2p:
@christiananese have you been able to give it a spin?

riqwan (approved):

olivermrbl (approved):

PR #9535, opened 2024-10-11
fix(dashboard): reservation inventory details

fPolic:
**What**
- display data in inventory details section of the reservation details page

---

FIXES CC-595

### DIFF
--- a/packages/admin/dashboard/src/routes/reservations/reservation-detail/reservation-detail.tsx
+++ b/packages/admin/dashboard/src/routes/reservations/reservation-detail/reservation-detail.tsx
@@ -7,6 +7,7 @@ import { useReservationItem } from "../../../hooks/api/reservations"
 import { InventoryItemGeneralSection } from "../../inventory/inventory-detail/components/inventory-item-general-section"
 import { ReservationGeneralSection } from "./components/reservation-general-section"
 import { reservationItemLoader } from "./loader"
+import { useInventoryItem } from "../../../hooks/api"
 
 export const ReservationDetail = () => {
   const { id } = useParams()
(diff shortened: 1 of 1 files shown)


packages/admin/dashboard/src/routes/reservations/reservation-detail/reservation-detail.tsx
```diff
@@ -23,6 +24,9 @@ export const ReservationDetail = () => {
     }
   )
 
+  // TEMP: fetch directly since the fields are not populated with reservation call
+  const { inventory_item } = useInventoryItem(reservation.inventory_item?.id!)
```
olivermrbl: nit: check if ID exists

olivermrbl (approved):

packages/admin/dashboard/src/routes/reservations/reservation-detail/reservation-detail.tsx
```diff
@@ -23,6 +24,9 @@ export const ReservationDetail = () => {
     }
   )
 
+  // TEMP: fetch directly since the fields are not populated with reservation call
```
srindom, line 27: q: where would it be more correct to fetch the item details?

fPolic: what endpoint would you suggest, this is fetching inventory item

srindom: I am asking wrt. the comment you put here - what would be the optimal approach to make this not `TEMP`? Is it just adding the `inventory_item` field when fetching the reservation?

If so we should maybe either include that in this PR or create a separate task so that can be fixed

fPolic: > Is it just adding the inventory_item field when fetching the reservation?

Yup, ideally we would want that but with this approach computed properties such as `reserved_quantity` are not being returned. I'll open a ticket.

PR #9537, opened 2024-10-11
fix(dashboard,ui): Fixes to Combobox and CategoryCombobox

kasperkristensen:
**What**
- Fixes the Combobox to keep the width of the content constant.
- Brings CategoryCombobox inline with the other Combobox component
- Adds keyboard navigation to the CategoryCombobox: You can now navigate options using ArrowUp and ArrowDown, and if an option has children you can use ArrowRight to see the children options.
- Add "outline-none" to the Drawer component to stop it from flashing whenever focus is dropped.
- Removes a dependency that was added to the UI package by mistake

Resolves CC-155


srindom (approved):
LGTM ✅

PR #9542, opened 2024-10-12
fix: Add shipping method data validation

olivermrbl:
Closes CC-538

### DIFF
--- a/packages/core/core-flows/src/cart/workflows/add-shipping-method-to-cart.ts
+++ b/packages/core/core-flows/src/cart/workflows/add-shipping-method-to-cart.ts
@@ -12,6 +12,7 @@ import {
   validateCartShippingOptionsStep,
 } from "../steps"
 import { validateCartStep } from "../steps/validate-cart"
+import { validateAndReturnShippingMethodsDataStep } from "../steps/validate-shipping-methods-data"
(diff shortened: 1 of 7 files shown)


packages/core/core-flows/src/cart/workflows/add-shipping-method-to-cart.ts
```diff
@@ -83,13 +107,17 @@ export const addShippingMethodToWorkflow = createWorkflow(
             )
           }
 
+          const methodData = data.validatedMethodData?.find((methodData) => {
+            return methodData?.[option.id]
+          })
```
riqwan, line 112: q: If this bit is not valid, should we return early or throw than trying to proceed?

riqwan (approved):
one question, otherwise lgtm!

fPolic (approved):

packages/core/core-flows/src/cart/workflows/add-shipping-method-to-cart.ts
olivermrbl: If the fulfillment provider doesn't return anything (maybe there is no validation), then I'd rather proceed and default to an empty object than throw

PR #9544, opened 2024-10-13
feat(providers): locking redis

carlos-r-l-rodrigues:
What:
 * Providers loading like modules, auto discovering models, migrations, etc
 * Locking module provider - Redis
 * Create reservation step locking to manage inventory

New util added to export a provider: `ModuleProvider`

```
export default ModuleProvider(Modules.LOCKING, {
  services,
  loaders,
})
```

Module providers are registered on the Module's container using the name `__providers__` + `provider_key_name`. It is available on Module's loaders.

Postgres: https://github.com/medusajs/medusa/pull/9545

**:warning: Note**
In case of `too many connection` issue, the first thing to check (especially when running many processes) is the `pool` configuration in `medusa-config.ts`. e.g:
```ts
projectConfig: {
  databaseDriverOptions: {
      pool: {
        max: 8, // This number can also vary depending on the max connection configuration of your db or bouncer or whatever tech you are using and configuring
      },
    },
}
```

**Convention**
With the current approach to load providers, we need to validate the convention being the following one:
- module options must has the `providers` keys as the option being an array and must contain 
  - `resolve`
  - `id` (optional) only used if the provider does not have an identifier (@carlos-r-l-rodrigues wdyt of that, seems we need to validate something here)
  - the provider can have an `identifier` static property that would be used to register the provider into the module container under `__providers__${identifier | id}`


olivermrbl (commented):
This is very exciting @carlos-r-l-rodrigues, did you already run the same tests as without it?

carlos-r-l-rodrigues:
https://github.com/medusajs/medusa/pull/9545
Postgres implementation

packages/modules/providers/locking-redis/src/services/redis-lock.ts
```diff
@@ -0,0 +1,269 @@
...
+
+    const ownerId = args?.ownerId ?? "*"
+    const awaitQueue = args?.awaitQueue ?? false
+
+    const acquirePromises = keys.map(async (key) => {
+      const errMessage = `Failed to acquire lock for key "${key}"`
+      const keyName = this.getKeyName(key)
+
+      const acquireLock = async () => {
+        while (true) {
```
olivermrbl, line 172: thought: For good measure, should we have a max retries config?

packages/modules/providers/locking-redis/src/services/redis-lock.ts
```diff
@@ -0,0 +1,269 @@
...
+    },
+    cancellationToken?: { cancelled: boolean }
+  ): Promise<void> {
+    keys = Array.isArray(keys) ? keys : [keys]
+
+    const timeout = Math.max(args?.expire ?? this.waitLockingTimeout, 1)
+    const timeoutSeconds = Number.isNaN(timeout) ? 1 : timeout
+    let retryTimes = 0
+
+    const ownerId = args?.ownerId ?? "*"
```
olivermrbl, line 164: q: Shouldn't this be random when not provided to avoid "imitating owners" or am I missing something?

packages/modules/providers/locking-redis/src/services/redis-lock.ts
```diff
@@ -0,0 +1,269 @@
...
+      return result === 1
+    })
+
+    const results = await promiseAll(releasePromises)
+
+    return results.every((released) => released)
+  }
+
+  async releaseAll(args?: { ownerId?: string | null }): Promise<void> {
+    const ownerId = args?.ownerId ?? "*"
```
olivermrbl, line 231: Same question here.

packages/modules/providers/locking-redis/src/services/redis-lock.ts
olivermrbl: Or alternatively, a required argument

olivermrbl (approved):
Left a few questions, but generally LGTM. Nice work!

packages/modules/providers/locking-redis/src/services/redis-lock.ts
carlos-r-l-rodrigues: This number of retries is tricky because it will greatly depend on the time to wait for the locking and the interval of the retry.
We are increasing the time for the next retry to not retry constantly if there are many process trying to to lock it.
As we always have a timeout set, and this scenario is an edge case I believe this is fine.

packages/modules/providers/locking-redis/src/services/redis-lock.ts
carlos-r-l-rodrigues: Good point, I changed things and the last version is not correct.
It should be "*" or a default value, however in the case the key is present, it should throw instead of extending the ttl.
If we generate a random ownerId, the locking won't be able to be released.
I will fix this in the other providers as well.

packages/modules/providers/locking-redis/src/services/redis-lock.ts
olivermrbl: Got it, makes sense.

packages/modules/providers/locking-redis/src/services/redis-lock.ts
carlos-r-l-rodrigues: the `acquire` method can be used for long locks that have `undefined` expiration time.
I realised as well that this is not covered in this provider.

packages/modules/providers/locking-redis/src/services/redis-lock.ts
```diff
@@ -0,0 +1,269 @@
+import { MedusaService, promiseAll } from "@medusajs/framework/utils"
+import { ILockingProvider } from "@medusajs/types"
+import { RedisCacheModuleOptions } from "@types"
+import { Redis } from "ioredis"
+import { setTimeout } from "node:timers/promises"
+
+export class RedisLockingProvider
+  extends MedusaService({})
```
olivermrbl: q: Should this extend `MedusaService`? I actually only thought that was relevant for modules with data models.

carlos-r-l-rodrigues: good catch... removing it here

olivermrbl:
We might need to draft up some docs on this new `ModuleProvider` utility

cc @shahednasser

PR #9549, opened 2024-10-14
fix(dashboard): Adds routes for handling tax overrides for provinces

kasperkristensen:
**What**
- Adds routes for creating and editing tax overrides for provinces.
- Fixes an issue where color-schema, was locked to the users OS settings. E.g. if the users OS was in light mode, but the admin in darkmode, then the scrollbars would be light mode.

Resolves CC-548


srindom (approved):
LGTM!

fPolic (approved):

adrien2p (approved):

olivermrbl (approved):

PR #9553, opened 2024-10-14
fix(dashboard): Show progress on tabs in create form

kasperkristensen:
**What**
- Shows progress on tabs in create promotion form. This is a bit of a band-aid fix, the promotion form and domain needs some proper cleanup, but this will work for now.
- Adds missing translation of "Method" field.

Resolves CC-113


fPolic (approved):

srindom (approved):
LGTM!

PR #9560, opened 2024-10-14
fix: Link migration descriptor case changes and hash computation

adrien2p:
**What**
The module service name case has changed and the hash generation was performed on the non lower cased version of it while after the hash generation everything is based on the lower case version of the generated table name form the service names leading to different hash computation. This pr update the link migration table to adjust the to/from module value with its new value as well as generating the hash from the lower case version of the computed table name


packages/modules/link-modules/src/migration/index.ts
```diff
@@ -332,24 +332,30 @@ export class MigrationsExecutionPlanner implements ILinkMigrationsPlanner {
...
             trackedTable.link_descriptor.toModel &&
-          entity.linkDescriptor.fromModule ===
-            trackedTable.link_descriptor.fromModule &&
-          entity.linkDescriptor.toModule ===
-            trackedTable.link_descriptor.toModule
+          entity.linkDescriptor.fromModule.toLowerCase() ===
+            trackedTable.link_descriptor.fromModule.toLowerCase() &&
+          entity.linkDescriptor.toModule.toLowerCase() ===
+            trackedTable.link_descriptor.toModule.toLowerCase()
         )
```
thetutlage, line 345: Do we want to do `toLowerCase` or convert the module name to `snake_case`? I am not 100% sure either

carlos-r-l-rodrigues (approved):

PR #9564, opened 2024-10-14
chore: add compare_at_unit_price when price list price is retrieved

riqwan:
What:

- sets compare_at_unit_price when price list sale price is present
- optionally sets compare_at_unit_price when setting custom price in change order
- refresh cart items upon cart updates

RESOLVES CORE-599
RESOLVES CORE-557


SalahAdDin:
> _ No description provided. _

what about a full description?

packages/core/core-flows/src/cart/steps/get-line-item-actions.ts
```diff
@@ -48,15 +48,18 @@ export const getLineItemActionsStep = createStep(
       if (existingItem && metadataMatches) {
         const quantity = MathBN.sum(
           existingItem.quantity as number,
-          item.quantity || 1
-        ).toNumber()
+          item.quantity ?? 1
```
adrien2p, line 51: ahah nice catch, otherwise you cant provide 0 :)

packages/core/core-flows/src/cart/steps/get-line-item-actions.ts
```diff
@@ -48,15 +48,18 @@ export const getLineItemActionsStep = createStep(
...
+        )
 
         itemsToUpdate.push({
           selector: { id: existingItem.id },
           data: {
             id: existingItem.id,
             quantity: quantity,
             variant_id: item.variant_id!,
+            unit_price: item.unit_price || existingItem.unit_price,
```
adrien2p: Q: should we use nullishcoalescing as well?

packages/core/core-flows/src/cart/steps/get-line-item-actions.ts
```diff
@@ -48,15 +48,18 @@ export const getLineItemActionsStep = createStep(
...
+            compare_at_unit_price:
```
adrien2p, line 61: Q: should we use nullishcoalescing as well?

packages/core/core-flows/src/cart/workflows/complete-cart.ts
```diff
@@ -88,14 +88,14 @@ export const completeCartWorkflow = createWorkflow(
...
-          variant_id: item.variant_id,
-          quantity: item.quantity,
+        data.cart?.items?.forEach((item) => {
+          allItems.push({
+            id: item.id,
+            variant_id: item.variant_id,
+            quantity: item.quantity,
+          })
+          allVariants.push(item.variant)
         })
```
adrien2p: nit

```suggestion
const allVariants: any[] = []
const allItems = data.cart?.items?.map((item) => {
  allVariants.push(item.variant)
  return {
    id: item.id,
    variant_id: item.variant_id,
    quantity: item.quantity,
  }
})
```

adrien2p: is it possible to have duplicated variant ids here?

packages/core/core-flows/src/cart/workflows/complete-cart.ts
```diff
@@ -109,7 +109,9 @@ export const completeCartWorkflow = createWorkflow(
           return prepareLineItemData({
             item,
             variant: item.variant,
-            unitPrice: item.raw_unit_price ?? item.unit_price,
+            unitPrice: item.raw_unit_price || item.unit_price,
+            compareAtUnitPrice:
+              item.raw_compare_at_unit_price || item.compare_at_unit_price,
```
adrien2p: Q: Since we move to nullish to OR does that mean that 0 is not a valid value?

packages/core/core-flows/src/cart/workflows/refresh-cart-items.ts
```diff
@@ -0,0 +1,131 @@
...
+import { prepareLineItemData } from "../utils/prepare-line-item-data"
+import { refreshPaymentCollectionForCartWorkflow } from "./refresh-payment-collection"
+import { updateCartPromotionsWorkflow } from "./update-cart-promotions"
+import { updateTaxLinesWorkflow } from "./update-tax-lines"
+
+export const refreshCartItemsWorkflowId = "refresh-cart-items"
+/**
+ * This workflow refreshes a cart's items
+ */
+export const refreshCartItemsWorkflow = createWorkflow(
```
adrien2p, line 24: discussion: With that new workflow would that mean that in the future we could provide a light add to cart and eventually call the refresh to get all the right taxes and promotions etc?

cc @carlos-r-l-rodrigues what we talked about ahah

packages/core/core-flows/src/cart/workflows/refresh-cart-items.ts
```diff
@@ -0,0 +1,131 @@
...
+        promo_codes: cartPromoCodes,
+        action: PromotionActions.REPLACE,
+      },
+    })
+
+    refreshPaymentCollectionForCartWorkflow.runAsStep({
+      input: { cart_id: cart.id },
+    })
+
+    return new WorkflowResponse(refetchedCart)
```
adrien2p, line 129: suggestion: should we really return the cart here knowing that the totals etc eventully wont be correct at this point

packages/core/core-flows/src/cart/workflows/update-cart.ts
```diff
@@ -133,35 +129,10 @@ export const updateCartWorkflow = createWorkflow(
...
-        promo_codes: input.promo_codes,
-        action: PromotionActions.REPLACE,
-      },
-    })
-
-    refreshPaymentCollectionForCartWorkflow.runAsStep({
-      input: {
-        cart_id: input.id,
-      },
+    const cart = refreshCartItemsWorkflow.runAsStep({
```
adrien2p, line 134: that answer my previous question ahah 😄

packages/core/core-flows/src/cart/workflows/refresh-cart-items.ts
riqwan: Yes, that should be possible. This would require some clever frontend state management for it to work accurately

packages/core/core-flows/src/cart/workflows/refresh-cart-items.ts
riqwan: yeah I wanted to skip this as well, but its necessary for the caller workflow. This whole flow needs a bit of optimization.

packages/core/core-flows/src/cart/steps/get-line-item-actions.ts
riqwan: converted them all. Lets see if that passes the specs. 🥂

packages/core/core-flows/src/cart/workflows/complete-cart.ts
riqwan: Yes, possible. Converted it to an object.

packages/core/core-flows/src/cart/steps/get-line-item-actions.ts
riqwan: Looks good! all 🟢

packages/core/core-flows/src/cart/workflows/refresh-cart-items.ts
adrien2p: The idea would be to let the option to the front end to for example only show the number of items you have in your cart (which is the case most of the time) and in that case you don't need to always perform the calculation and related update. Then when the customer goes to the cart it could be the place you refresh it. Also, we could have safety system on our side, for example if you complete a cart we check that it is fresh or we refresh it always or something like that. But the main idea would be to remove that bottleneck for most cases but still have the ability to refresh for each add to cart depending on your needs

packages/core/core-flows/src/cart/workflows/refresh-cart-items.ts
adrien2p: it is used for the hook right? but in that case the hook will received a non fully updated cart, we just need to be clear about it right?

adrien2p (approved):

packages/core/core-flows/src/cart/workflows/refresh-cart-items.ts
riqwan: Yes, but I think that approach is wrong. The hooks should ideally be refetching the cart and we should only pass IDs into the hooks and never work with an entire deep object that always has the potential to mutate in multiple ways or go stale if async.

We've already fixed so many bugs internally because of issues on relying on an object passed to a step that at some point changed and broke all the subsequent workflows.

PR #9580, opened 2024-10-15
breaking: rename package names to be consistent and under @medusajs scope

thetutlage:
Fixes: FRMW-2725, FRMW-2726, FRMW-2689

In this PR we rename the following packages.

- `medusa-test-utils` to `@medusajs/test-utils`
- `medusa-telemetry` to `@medusajs/telemetry`
- `@medusajs/medusa-cli` to `@medusajs/cli`
- `@medusajs/inventory-next` to `@medusajs/inventory`
- `@medusajs/stock-location-next` to `@medusajs/stock-location`
- `@medusajs/file-local-next` to `@medusajs/file-local`

Feel free to let me know if any other packages must be renamed.


carlos-r-l-rodrigues:
we'll need to update the CLI tests [here](https://github.com/medusajs/medusa/blob/develop/.github/workflows/test-cli-with-database.yml#L50) as well

thetutlage:
@carlos-r-l-rodrigues 

Will it work? Because the package doesn't exist in the registry?

carlos-r-l-rodrigues:
> @carlos-r-l-rodrigues
> 
> Will it work? Because the package doesn't exist in the registry?

No, not in this PR. so we need to decide to either have a subsequent PR just modifying this after it is released to not break the rest, or we break this one and force merge it

olivermrbl (approved):
This LGTM, nice work!

PR #9582, opened 2024-10-15
chore: remove internal module resources option

carlos-r-l-rodrigues:
What:
* removes resouces type "shared" or "isolated" from internal modules.
* modules can have an isolated database connection by providing a database config as part of their options on `medusa-config`

CLOSES: FRMW-2593


olivermrbl (approved):
Nice

PR #9584, opened 2024-10-15
chore: Update admin build/serve configuration

adrien2p:
**Breaking changes**

The `outDir` has been deprecated and wont be used anymore, instead all the path are computed internally following these rules
- if admin is not `disabled` and the `build` command is run without the `--admin-only` flag, then the admin output dir will be `.medusa/server/public/admin` and it will be served from that same location from the medusa instance.
- if admin is not `disabled` and the `build` command is run with the `--admin-only` flag, then the admin output dir will be `.medusa/admin` with the purpose of deploying the admin separately. ⚠️ (expect to receive a warning log)
- if the admin is `disabled` and the `build` command is run with the `--admin-only` flag, then fallback to rule number 2

| admin enabled  | medusa build --admin-only  | output dir |
|---|---|---|
| true  | true  | `.medusa/admin` ⚠️ (expect to receive a warning log) |
| true  | false  | `.medusa/server/public/admin`  |
| false  | true  | `.medusa/admin`  |
| false  | false  | none |

```diff
// medusa-config.ts

{
  // ...
  admin: {
-    outDir: 'some/path'
  }
}
```

cc @kasperkristensen @sradevski @olivermrbl


packages/core/types/src/common/config-module.ts
```diff
@@ -63,7 +65,7 @@ export interface AdminOptions {
    * })
    * ```
    */
-  outDir: string
+  outDir?: string
```
sradevski: Should we remove it already? I doubt many use it, and it's better to do it now than later WDYT?

sradevski (approved):
Nice!

packages/core/types/src/common/config-module.ts
adrien2p: Yes I am good with that, let me check

kasperkristensen:
> * if admin is not `disabled` and the `build` command is run with the `--admin-only` flag, then the admin output dir will be `.medusa/admin` with the purpose of deploying the admin separately

This case kind of doesn't make sense, you would never want to run the build command with --admin-only if you haven't disabled it as well, because in that case you end up with two admins. 

I think its fine to keep the functionality, as it is the correct behaviour for case 3, but maybe it would make sense to log a warning like: "You are building the admin dashboard for standalone usage, but have not disabled the dashboard from being served along with your server. If you intend to host the dashboard separatly you should disable the admin in your medusa config".

adrien2p:
> > * if admin is not `disabled` and the `build` command is run with the `--admin-only` flag, then the admin output dir will be `.medusa/admin` with the purpose of deploying the admin separately
> 
> This case kind of doesn't make sense, you would never want to run the build command with --admin-only if you haven't disabled it as well, because in that case you end up with two admins.
> 
> I think its fine to keep the functionality, as it is the correct behaviour for case 3, but maybe it would make sense to log a warning like: "You are building the admin dashboard for standalone usage, but have not disabled the dashboard from being served along with your server. If you intend to host the dashboard separatly you should disable the admin in your medusa config".

just added a warning as suggested 👍

packages/medusa/src/utils/admin-consts.ts
```diff
@@ -0,0 +1,3 @@
+export const ADMIN_SOURCE_DIR = "src/admin"
+export const ADMIN_RELATIVE_OUTPUT_DIR = "./public/admin"
```
olivermrbl, line 2: q: isn't the public folder typically associated with something else than build artefacts? maybe we can do `./dist/admin`?

adrien2p: I dont have a strong opinion on this, the thinking was that it is served through express middleware and in general the files you serve through a middleware are often located in a public directory. But fine to move it to dist if you prefer

olivermrbl: All good, let's keep it as is

yannickschuchmann:
Do I understand it correctly that when I set the following:
```js
// medusa-config.js
admin: {
      disable: true, // process.env.DISABLE_MEDUSA_ADMIN === "true", <--- this line
      backendUrl: process.env.MEDUSA_BACKEND_URL,
      path: process.env.MEDUSA_ADMIN_PATH,
},
```

and then run `NODE_ENV=production medusa build` (without the `--admin-only` option), that it should skip the build of the admin interface and only build the server?

Because what I encounter is that it actually builds the admin still and outputs it into `.medusa/server/public/admin/`.

Is this a bug? And is there a way/workaround to really skip the admin build instead? We are building it on a server with less than 2gb RAM and running out of memory..

thx :)

PR #9586, opened 2024-10-15
feat(dashboard) taxes + discount breakdown

fPolic:
**What**
- add tax breakdown to order details
- display discount total and codes

---

CLOSES CC-553


srindom:
Looks good! Would suggest we make a small additional change in this PR if it's not too much work:

1. Include details about the item total, and change order in which totals should be displayed.
2. Show item, shipping and discount totals with or without taxes depending on order.region.automatic_taxes.

Specifically, if automatic_taxes: false then we should show something like:

Item Subtotal - 8.00 DKK
Shipping Subtotal - 8.00 DKK
Discount Subtotal - 2.00 DKK
Tax Total - 3.50 DKK


If automatic_taxes: true we should show something like:

Item Total - 10.00 DKK
Shipping Total - 10.00 DKK
Discount Total - 2.50 DKK
Tax Total (included) - 3.50 DKK

The reason for this is that we always want to show the details in admin as they are likely shown to the customer. automatic_taxes helps us here as it signals whether the store at any moment in time showed prices to the customer without taxes.

E.g., if automatic_taxes: false, then there was a period of time where the customer saw item prices without taxes - up until the point when the tax context was known. This likely also means that taxes is a separate line in the totals breakdown shown to the customer as opposed to being included. This is approach is for example common in the US.

It might be that we will have to revise this decision in the future, but for now this should be a pretty good solution.

srindom (approved):
LGTM

PR #9602, opened 2024-10-16
fix(admin-vite-plugin): Replace slashes for Windows

kasperkristensen:


olivermrbl (approved):

fPolic (approved):

packages/admin/admin-vite-plugin/src/utils.ts
```diff
@@ -17,7 +17,7 @@ import {
 } from "./babel"
 
 export function normalizePath(file: string) {
-  return path.normalize(file.replace(/\\/g, "/"))
+  return path.normalize(file).replace(/\\/g, "/")
```
adrien2p, line 20: suggestion: sorry to jump after the battle 😂 wdyt of the following
```ts
return path.normalize(file).split(path.sep).join('/')
```

PR #9604, opened 2024-10-16
docs: added documentation for build command

shahednasser:
- Added documentation for `medusa build` command
- Update deployment documentation for admin to use `--admin-only` option

### DIFF
--- a/www/apps/resources/app/medusa-cli/commands/build/page.mdx
+++ b/www/apps/resources/app/medusa-cli/commands/build/page.mdx
@@ -11,7 +11,9 @@ export const metadata = {
 
 # {metadata.title}
 
-Create a standalone build of the Medusa application. This is useful to create a build that:
+Create a standalone build of the Medusa application.
+
+This creates a build that:
 
 - Doesn't rely on the source TypeScript files.
 - Has a smaller Docker image size.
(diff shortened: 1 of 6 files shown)


www/apps/resources/app/medusa-cli/commands/build/page.mdx
```diff
@@ -0,0 +1,83 @@
...
+  title: `build Command - Medusa CLI Reference`,
+}
+
+# {metadata.title}
+
+Create a standalone build of the Medusa application. This is useful to create a build that:
+
+- Doesn't rely on the source TypeScript files.
+- Has a smaller Docker image size.
+- Can be copied to a production server reliably.
```
olivermrbl: thought: Don't think these add much right now, would leave them out of the docs

olivermrbl: What I mean is, this is not a special way to build Medusa, it is the only way, so think the guide should explain it as such

shahednasser: Makes sense, I made a slight adjustment to the text. I still think it would make sense to explain what you'll be getting out of the build

www/apps/resources/app/medusa-cli/commands/build/page.mdx
```diff
@@ -0,0 +1,85 @@
...
+}
+
+# {metadata.title}
+
+Create a standalone build of the Medusa application.
+
+This creates a build that:
+
+- Doesn't rely on the source TypeScript files.
+- Has a smaller Docker image size.
```
olivermrbl: nit: Can we remove the Docker comment, think it's a bit out of place 😅

olivermrbl (approved):

PR #9613, opened 2024-10-16
feat: allow html content for notifications

christiananese:


sradevski (commented):
Provider-wise it looks good, but we need to modify the notification module to also support it.

christiananese:
> Provider-wise it looks good, but we need to modify the notification module to also support it.

@sradevski For email notifications I wouldn't store the data to the DB. So this should actually already do the thing, we could move the subject to the data object, but in that case it will get a mess with typings as subject is required  for if no template is used

sradevski (approved):
LGTM, I would maybe just wait for someone else to chime in on the naming (can ask in PRs channel)

packages/modules/providers/notification-sendgrid/src/services/sendgrid.ts
```diff
@@ -58,14 +58,28 @@ export class SendgridNotificationService extends AbstractNotificationProviderSer
 
     const from = notification.from?.trim() || this.config_.from
 
+    let mailContent
+
+    if ("providerContext" in notification) {
```
adrien2p:
```suggestion
    if ("providerContext" in notification && !!notification.providerContext) {
```

just to be sure if an empty object is provided maybe we want to handle it wdyt? because the `in` operator only ensure the key is present in the object, but it can still be undefined for example

packages/modules/providers/notification-sendgrid/src/services/sendgrid.ts
```diff
@@ -58,14 +58,28 @@ export class SendgridNotificationService extends AbstractNotificationProviderSer
...
+      mailContent = {
+        subject: notification.providerContext?.subject,
+        html: notification.providerContext?.html,
+      }
+    } else {
+      // we can't mix html and templates for sendgrid
+      mailContent = {
+        templateId: notification.template,
```
riqwan, line 78: q: don't you still need the subject?

riqwan (approved):
lgtm!

packages/modules/providers/notification-sendgrid/src/services/sendgrid.ts
```diff
@@ -58,14 +58,28 @@ export class SendgridNotificationService extends AbstractNotificationProviderSer
 
     const from = notification.from?.trim() || this.config_.from
 
+    let mailContent
```
adrien2p:
```suggestion
    let mailContent: {
      subject?: string
      html?: string
    } | {
      templateId: string
    }
```

adrien2p (approved):

packages/core/types/src/notification/mutations.ts
```diff
@@ -21,6 +21,10 @@ export interface CreateNotificationDTO {
    * The data that gets passed over to the provider for rendering the notification.
    */
   data?: Record<string, unknown> | null
+  /**
+   * The metadata that gets passed over to the provider.
+   */
+  providerContext?: Record<string, unknown> | null
```
srindom: should be snake_cased

```suggestion
  provider_context?: Record<string, unknown> | null
```

srindom (commented):
One thought here: something feels off about passing html though a field called `providerContext`. 

To me the it would be natural for the notification abstraction to somehow know about html. This would require a larger change that what is suggested here. 

Could a workaround for now be to pass html through `data` and then make the sendgrid plugin "html-aware"?

packages/modules/providers/notification-sendgrid/src/services/sendgrid.ts
christiananese: No, the templates in sendgrid already have one

packages/modules/providers/notification-sendgrid/src/services/sendgrid.ts
christiananese: Good catch.

christiananese:
> One thought here: something feels off about passing html though a field called `providerContext`.
> 
> To me the it would be natural for the notification abstraction to somehow know about html. This would require a larger change that what is suggested here.
> 
> Could a workaround for now be to pass html through `data` and then make the sendgrid plugin "html-aware"?

I'm happy to rename it, but it was what I found most fitting, as it'S basically passing the context from the notification to the underlying provider.

We can't use data for it unless we do a much larger rewrite, data gets always stored to the db and in most cases we don't want that for an email. I first wanted to call it  metadata, but that might confuse people.

srindom (approved):
LGTM

olivermrbl (approved):

PR #9615, opened 2024-10-16
fix(utils): Mikro orm joined selection issue when select-in strategy

adrien2p:
**What**
When in `select-in` strategy selection with mikro orm, the behaviour is different than using the `joined` strategy in the sense that with the joined strategy if a relation is populated then it will be returned entirely. But in `select-in `strategy if a relation is populated but no fields are selected from it then it wont be returned at all.


carlos-r-l-rodrigues (approved):

olivermrbl (approved):
LGTM

PR #9616, opened 2024-10-16
chore(medusa): default endpoints size limit

carlos-r-l-rodrigues:
What:
 * Increase `bodyParser.sizeLimit` to `2mb` for all `/batch` endpoints


adrien2p (approved):

thetutlage:
Will it be possible to configure it within the app too?

carlos-r-l-rodrigues:
> Will it be possible to configure it within the app too?

I think we should. We need to review this route loader and ideally that could be overwritten.
Middlewares, default fields, and some other things we should be able to update.

PR #9623, opened 2024-10-16
feat(dashboard): Submit forms on Cmd + Enter

kasperkristensen:
**What**
- Changes all forms to only submit on Cmd/Ctrl + Enter instead of just Enter.
- Cleans up the position of submit/cancel buttons in many FocusModals that still had them in the header.
- Fixes responsiveness of multiple forms
- Removes the SplitView component, and replaces its usages with StackedDrawer/Modal to align the UX across the project.

Resolves CC-103, CC-535


olivermrbl (approved):
This feels good, nice work @kasperkristensen 💪

PR #9636, opened 2024-10-17
chore: Update modules providers configuration with 'identifier' and 'PROVIDER'

adrien2p:
**What**
- standardize provider `identifier` static property. Now all providers must have a static property `identifier` which will be used to register the provider into the module container.
- The `id` in the providers configuration can still be provided in the defineConfig if some provider can have multiple instances (e.g stripe per region)
- All provider id have been fixed in all modules to be properly generated (e.g `pp_stripe_the-given-id` or `pp_stripe` instead of `pp_stripe_` or `pp_stripe_undefined`

**Breaking changes**
- The provide id in `region_provider_id` might be broken after the provider are reloaded since their id have been fixed. You might need to update those manually or regenerate those link
- The auth module providers have been also aligned to the convention of having identifier and eventually an id and their generated id will following the same convention


carlos-r-l-rodrigues (approved):

adrien2p:
cc @sradevski @christiananese you might need to take a look at this one, we can discuss if you need any help for the changes

sradevski (approved):
LGTM. This shouldn't affect us I believe, we don't have any custom providers implemented for now.

adrien2p:
> LGTM. This shouldn't affect us I believe, we don't have any custom providers implemented for now.

It affect the core providers as well, the auth providers have been aligned as well, maybe that can impact you?

olivermrbl (commented):
wonder if we should snake_case the IDs for consistency? It will be a breaking change regardless, so thought we might as well

olivermrbl:
I don't think this breaks the auth domain, because the `id`s specified in `medusa-config.js` in the platform project happen to be the same as the provider IDs.

adrien2p:
> I don't think this breaks the auth domain, because the `id`s specified in `medusa-config.js` in the platform project happen to be the same as the provider IDs.

-- In order to normalize and be consistent, now the auth also concatenate the identifier and id which was not the case before, that can be breaking too --

nvm the above, we didn't, but I thought we did 😂

olivermrbl (approved):
LGTM

PR #9643, opened 2024-10-17
feat(medusa): Display admin url on start

adrien2p:
**What**
Display the admin url on server start in the console
Screenshot 2024-10-17 at 16 07 10

### DIFF
--- a/packages/medusa/src/commands/start.ts
+++ b/packages/medusa/src/commands/start.ts
@@ -17,6 +17,7 @@ import { logger } from "@medusajs/framework/logger"
 
 import loaders from "../loaders"
 import { MedusaModule } from "@medusajs/framework/modules-sdk"
+import { MedusaContainer } from "@medusajs/framework/types"
 
 const EVERY_SIXTH_HOUR = "0 */6 * * *"
 const CRON_SCHEDULE = EVERY_SIXTH_HOUR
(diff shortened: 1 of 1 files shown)


packages/medusa/src/commands/start.ts
```diff
@@ -109,6 +114,15 @@ async function start(args: {
       const server = GracefulShutdownServer.create(
         http_.listen(port).on("listening", () => {
           logger.success(serverActivity, `Server is ready on port: ${port}`)
+
+          const {
+            admin: { path: adminPath },
```
shahednasser: thought: shouldn't we check if the admin is disabled in the config?

shahednasser (approved):
Love it!

packages/medusa/src/commands/start.ts
kasperkristensen: Yes I think so to, so we don't print a link that goes to no where.

Also think we can skip printing this in production, as the actual link won't be http://localhost:9000/app in that case, and people aren't going to need a link to the admin there anyways.

kasperkristensen (approved):
LGTM! Just small nit on checking if admin is disabled, and not printing in production

packages/medusa/src/commands/start.ts
adrien2p: yes thats true 👍

adrien2p: I have performed the changes, let me know if that's ok for you

kasperkristensen: Looks good @adrien2p!

PR #9646, opened 2024-10-17
feat(medusa): ordem items endpoint

carlos-r-l-rodrigues:
** What **
 * `/admin/orders/:id/line-items` - This endpoint return all line items that existed in an Order. The field `version` represents the latest version of the order it is present, and `history` contain the `version.to` and `version.from` to identify when it was added and potentially removed.


fPolic (approved):
LGTM!

PR #9648, opened 2024-10-17
Feat/notification-twilio

420coupe:
Attempting to extend notification service to add twilio as an sms/mms provider.  However i am having issues trying to test locally. Carlos has provided the package.json resolutions as an option but it does not seem to be pulling all local changes to my cloned medusa repo.


420coupe:
So here's a working **notification-twilio** integration. 

**Pre-reqs:**
- Have a twilio account
- Purchase a sms/mms capable phone number
- create messaging service (optional)

There is also a process to register your Brand, get A2P10DC verified, and finally submit for campaign approval for your use case in production.

### `TwilioServiceConfig` has 3 required properties 1 of which has 2 options 
- `accountSid: string`  (required) 
- `authToken: string`  (required)
- `from?: string` (either `from` phone# (+13055555555) or `messagingServiceSid` required)
- `messagingServiceSid?: string` (either `messageingServiceSid` or `from` required)

You can provide both messagingServiceSid and from, if phone number is in messaging_service_sid's sender pool. and you want to use a specific phone number, otherwise allow messaging service to use senderpool. 

## .env example
```.env
TWILIO_ACCOUNT_SID=
TWILIO_AUTH_TOKEN=
TWILIO_FROM=
TWILIO_MESSAGING_SERVICE_SID=
```
## medusa-config.js
```js
//imports...
// projectConfig settings
modules: {
[Modules.NOTIFICATION]: {
      resolve: "@medusajs/medusa/notification",
      options: {
        providers: [
          {
            resolve: "@medusajs/medusa/notification-twilio",
            id: "twilio",
            options: {
              channels: ["sms"],
              account_sid: process.env.TWILIO_ACCOUNT_SID,
              auth_token: process.env.TWILIO_AUTH_TOKEN,
              messaging_service_sid: process.env.TWILIO_MESSAGING_SERVICE_SID,
              from: process.env.TWILIO_FROM,
            },
          },
        ],
      },
    },
   }) 
   ```

### send an SMS message using Twilio example subscriber for product created below

## subscriber/product-created.ts
```ts
import type { SubscriberArgs, SubscriberConfig } from "@medusajs/framework";
import { Modules } from "@medusajs/framework/utils";
import { INotificationModuleService } from "@medusajs/framework/types";

export default async function productCreateHandler({
  event: { data },
  container,
}: SubscriberArgs<{ id: string }>) {
  const notificationModuleService: INotificationModuleService =
    container.resolve(Modules.NOTIFICATION);

  // SMS/MMS Notification Content
  const smsData = {
    mediaUrls: [], // array of mediaUrls to send MMS
  };

  const smsContent = {
    text: `Test text: product with ID ${data.id} has been created`,
  };

  await notificationModuleService.createNotifications({
    to: "3055555555",
    channel: "sms",
    data: smsData,
    content: smsContent,
    template: "sms-test",
  });

export const config: SubscriberConfig = {
  event: "product.created",
};
```
Next create a product and viola you should have received a sms notification.
`Template` is not used on twilio's api side only internally for medusa notification table.

420coupe:
Should i create the event webhook with this same PR ?

420coupe:
Bump to get this integrated.

420coupe:
@sradevski @fPolic @carlos-r-l-rodrigues just retagging to get this reviewed and integrated to add twilio as a notification service.

sradevski:
Hi @420coupe I've brought this up internally and since community modules can explode without us being able to maintain them properly, we think it's better if you move the module to a separate repository and publish it as a package instead. That also makes ownership more obvious and you will not be blocked by reviews from the core team.

Let me know if we can assist in how to do that instead. Thanks!

420coupe:
> Hi @420coupe I've brought this up internally and since community modules can explode without us being able to maintain them properly, we think it's better if you move the module to a separate repository and publish it as a package instead. That also makes ownership more obvious and you will not be blocked by reviews from the core team.
> 
> Let me know if we can assist in how to do that instead. Thanks!

I have no interest in maintaining my own packages for medusa. IMHO This should be part of core code same as twilio sendgrid. Where will sendgrid package be living moving forward?

sradevski:
@420coupe sendgrid will remain in core since we commit to building and maintaining it. We are also using sendgrid internally, so it's easier for us to catch issues if any happen. At this point we cannot commit to the same support for Twilio so we'd rather not have it as part of the core codebase.

420coupe:
> @420coupe sendgrid will remain in core since we commit to building and maintaining it. We are also using sendgrid internally, so it's easier for us to catch issues if any happen. At this point we cannot commit to the same support for Twilio so we'd rather not have it as part of the core codebase.

they're both twilio products, only supporting one leg doesn't really make any sense when there's not much lift to maintain. 

leaving this open in case someone else wants to tackle and maintain the other repo.

PR #9649, opened 2024-10-17
fix(dashboard, js-sdk): undefined RMA activity items

fPolic:
**What**
- fix for showing removed items in rma flows

---

DEPENDS ON https://github.com/medusajs/medusa/pull/9646
FIXES CC-597


packages/admin/dashboard/src/hooks/api/orders.tsx
```diff
@@ -113,6 +118,28 @@ export const useOrderChanges = (
   return { ...data, ...rest }
 }
 
+export const useOrderLineItems = (
+  id: string,
+  query?: HttpTypes.AdminOrderLineItemsFilters,
+  options?: Omit<
+    UseQueryOptions<
+      HttpTypes.AdminOrderLineItemsResponse,
+      Error,
```
kasperkristensen:
```suggestion
      FetchError,
```

packages/core/types/src/http/order/admin/queries.ts
```diff
@@ -14,6 +14,13 @@ export interface AdminOrderFilters extends FindParams, BaseOrderFilters {
   updated_at?: OperatorMap<string>
 }
 
+export interface AdminOrderItemsFilters extends FindParams {
```
kasperkristensen, line 17: nit: using `*Filters` is inconsistent with how other types in the HTTP layer is named. I can see we have a bit of a mix of *Params and *Filters, so not sure what the right one is here. Maybe we can just leave this as is, but something that we need to clean up and align on.

packages/admin/dashboard/src/routes/orders/order-detail/components/order-activity-section/order-timeline.tsx
```diff
@@ -840,21 +868,24 @@ const OrderEditBody = ({
   )
 }
 
+/**
+ * Returns count of added and removed item quantity
+ */
 function countItemsChange(
```
kasperkristensen, line 874: nit: This section seem somewhat unsafe with all the "!" and type casting. If these are always present and always numbers then it would seem like the types are wrong?

kasperkristensen (approved):
Left a couple of nits, but not something related to the functionality so feel free to leave for now

packages/core/types/src/http/order/admin/queries.ts
fPolic: yup agree, but since there is more *Filter types in the order module I would prefer to refactor this in a separate PR

packages/admin/dashboard/src/routes/orders/order-detail/components/order-activity-section/order-timeline.tsx
fPolic: These fields will be present for RMA changes, but in theory details of the order change action could have different shapes depending on the concrete operation. But yes, we could add more concrete types for RMA changes and actions we know.

PR #9651, opened 2024-10-17
fix(dashboard): Hide usage insights

kasperkristensen:
**What**
- Hides everything related to usage insights from the profile page.

**Why**
- We haven't set up usage insights tracking in V2, so hiding the UI for disabling/enabling it. We can re-add it again later if we decide we want to continue tracking how the dashboard is used.


packages/admin/dashboard/src/routes/profile/profile-detail/components/profile-general-section/profile-general-section.tsx
```diff
@@ -62,14 +62,15 @@ export const ProfileGeneralSection = ({ user }: ProfileGeneralSectionProps) => {
             ?.display_name || "-"}
         </Text>
       </div>
-      <div className="grid grid-cols-2 items-center px-6 py-4">
+      {/* TODO: Do we want to implement usage insights in V2? */}
```
olivermrbl, line 65: Yes, probably something to add almost right after the release

olivermrbl (approved):

packages/admin/dashboard/src/routes/profile/profile-detail/components/profile-general-section/profile-general-section.tsx
kasperkristensen: Alright I'll create a ticket for it in linear 👍

PR #9654, opened 2024-10-18
Chore:workflow sdk cleanup

adrien2p:
**What**
First stab at migrating the workflows-sdk, some refactoring of the method implementation are left for a separate work. The idea of this PR is mainly reorganising the logic and make it easier to change and evolve while adding some implementations improvements


carlos-r-l-rodrigues (approved):

olivermrbl (approved):
Nice, if there are no changes to the public API, let's get this in 💪

adrien2p:
> Nice, if there are no changes to the public API, let's get this in 💪 

I need to include the latest changes made to the workflow sdk on develop

adrien2p:
@olivermrbl I ve managed the backward compatibility API to not have any breaking changes. I still have one thing to figure out for the typings in one place and then we can think about merging. I put it on pause for now

PR #9655, opened 2024-10-18
feature: use application logger to log HTTP requests

thetutlage:
Fixes: FRMW-2749

This PR takes the HTTP logs:

**From this**
CleanShot 2024-10-18 at 12 41 52@2x

**To this**
CleanShot 2024-10-18 at 13 25 44@2x

The logs will be displayed when the `LOG_LEVEL` environment variable is to set `http` or greater.

```dotenv
LOG_LEVE=info # no HTTP logs
LOG_LEVE=http # displays HTTP logs
LOG_LEVE=debug # displays HTTP logs
```


packages/core/framework/src/http/express-loader.ts
```diff
@@ -55,11 +61,33 @@ export async function expressLoader({ app }: { app: Express }): Promise<{
...
+   * and also exclude files served by vite during development
+   */
+  function shouldSkipHttpLog(req: MedusaRequest, res: MedusaResponse) {
+    if (isTest) {
+      return true
+    }
+    if (NOISY_ENDPOINTS_CHUNKS.some((chunk) => req.url.includes(chunk))) {
+      return true
+    }
+    return false
```
adrien2p:
```suggestion
    const isNoisy = NOISY_ENDPOINTS_CHUNKS.some((chunk) => req.url.includes(chunk))
    return isTest || isNoisy
```

if we use the regexp above it can become

```suggestion
    const isNoisy = !!req.url.match(NOISY_ENDPOINTS_CHUNKS)
    return isTest || isNoisy
```

packages/core/framework/src/http/express-loader.ts
```diff
@@ -5,7 +5,11 @@ import session from "express-session"
 import Redis from "ioredis"
 import morgan from "morgan"
 import path from "path"
+import { logger } from "../logger"
 import { configManager } from "../config"
+import { MedusaRequest, MedusaResponse } from "./types"
+
+const NOISY_ENDPOINTS_CHUNKS = ["@fs", "@id", "@vite", "@react", "node_modules"]
```
adrien2p, line 12: suggestion: maybe we can use a regexp directly? `new RegExp(/@fs|@id|@vite|@react|node_modules/, 'gm')`

adrien2p (approved):
LGTM

packages/core/framework/src/http/express-loader.ts
thetutlage: I think more of personal preference thing, or you think it will be faster?

packages/core/framework/src/http/express-loader.ts
thetutlage: Shouldn't we ignore the check if we are `test` environment?

adrien2p: It was returning true in both cases already but otherwise yes we might enable them in dev

packages/core/framework/src/http/express-loader.ts
adrien2p: I think it will be faster on a big number of requests to not have to loop through the while array for every requests wdyt?

adrien2p: some + includes is much better than regexp, lets not change that 👍

packages/core/framework/src/http/express-loader.ts
adrien2p: And in case I missunderstoor, here it is a short-circuit evaluation, the right hand side of the operand wont get evaluated if the left hand side is returning true

PR #9658, opened 2024-10-18
Fix/production logging

thetutlage:
Keep the old format (ie combined) when logging in production.

**Development**
CleanShot 2024-10-18 at 16 51 18@2x

**Production**
CleanShot 2024-10-18 at 16 52 25@2x

### DIFF
--- a/packages/cli/medusa-cli/src/reporter/index.ts
+++ b/packages/cli/medusa-cli/src/reporter/index.ts
@@ -5,7 +5,7 @@ import { ulid } from "ulid"
 import winston from "winston"
 import { panicHandler } from "./panic-handler"
 
-const LOG_LEVEL = process.env.LOG_LEVEL || "info"
+const LOG_LEVEL = process.env.LOG_LEVEL || "http"
 const LOG_FILE = process.env.LOG_FILE || ""
 const NODE_ENV = process.env.NODE_ENV || "development"
 const IS_DEV = NODE_ENV.startsWith("dev")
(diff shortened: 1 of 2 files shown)


olivermrbl (approved):

packages/cli/medusa-cli/src/reporter/index.ts
```diff
@@ -5,7 +5,7 @@ import { ulid } from "ulid"
 import winston from "winston"
 import { panicHandler } from "./panic-handler"
 
-const LOG_LEVEL = process.env.LOG_LEVEL || "info"
+const LOG_LEVEL = process.env.LOG_LEVEL || "http"
```
carlos-r-l-rodrigues, line 8: Is this `http` value intended?
if `env.LOG_LEVEL` is not provided, shouldn't it be `info` for dev and `error` for production?

adrien2p: The idea is to keep the current logs intact, so the http logs have to be present and therefore default to http which includes info, warn, error

carlos-r-l-rodrigues: Is this a common log level though? usually they are like `fatal, error, warn, info, debug, trace`. and a couple of variants of these

adrien2p: Yeah it is from npm ans Winston chooses it as a ref.

carlos-r-l-rodrigues: It is the first time I see this [`http`](https://docs.npmjs.com/cli/v8/using-npm/logging) 🤣 
It doesn't seem to be very standard elsewhere, but this is configurable anyway... so, all good.

PR #9668, opened 2024-10-18
fix(dashboard,js-sdk): remove method and hook for deleting claim

shahednasser:
Remove the JS SDK method and dashboard hook to delete a claim since the API route they're pointing at doesn't exist.


fPolic (approved):
LGTM

PR #9671, opened 2024-10-18
docs-util: fixes to package renames + other options

shahednasser:
- fix tsconfig files used by typedoc pointing to the wrong package names after the packages rename
- fix error when generating references + add more flexibility to reflection kinds

### DIFF
--- a/www/utils/packages/typedoc-generate-references/src/constants/references-details.ts
+++ b/www/utils/packages/typedoc-generate-references/src/constants/references-details.ts
@@ -22,6 +22,16 @@ export const customModulesOptions: Record<string, Partial<TypeDocOptions>> = {
       "packages/core/types/src/stock-location/service.ts"
     ),
   },
+  "inventory-next-models": {
+    entryPoints: getEntryPoints(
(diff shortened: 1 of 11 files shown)


www/utils/packages/typedoc-generate-references/src/constants/references-details.ts
```diff
@@ -22,6 +22,16 @@ export const customModulesOptions: Record<string, Partial<TypeDocOptions>> = {
       "packages/core/types/src/stock-location/service.ts"
     ),
   },
+  "inventory-next-models": {
```
olivermrbl, line 25: Q: should these have "next"?

olivermrbl (approved):

www/utils/packages/typedoc-generate-references/src/constants/references-details.ts
shahednasser: I left them as `next` because otherwise there will be huge renames in the generated files in the docs, which would result in infinity changes again 😅 we can look into renaming them at a later point when we have fewer changes or with some manual work.

PR #9687, opened 2024-10-21
fix: allow additional_data to be undefined or null

thetutlage:
Fixes: FRMW-2751

This PR allows `null` and `undefined` values to be passed when using a custom validation schema for the `additional_data`. 

## Question

However, there is another thing that we might have to either fix in docs or in the implementation. That is the usage of `additional_data` within workflows.

When additional_data is not provided in an API call, it is set to `undefined` and the same value is passed to the workflow hooks. However, I have noticed that our documentation assumes that the additional_data is always an object. For example: https://docs.medusajs.com/v2/learn/advanced-development/api-routes/additional-data#use-additional-data-in-a-hook

The `!additional_data.brand` check will throw an exception in cases where no `additional_data` was passed during the request.

**Question: Do we want to set it to an empty object when its not defined in an API call? Or do we want to update the docs to handle the case.**

### DIFF
--- a/packages/core/framework/src/http/__tests__/validate-body.spec.ts
+++ b/packages/core/framework/src/http/__tests__/validate-body.spec.ts
@@ -18,28 +18,102 @@ describe("validateAndTransformBody", () => {
   it("should pass additionalDataValidator to validator factory", async () => {
     let mockRequest = {
       query: {},
+      body: {
+        additional_data: {},
+      },
+    } as MedusaRequest
+
+    const mockResponse = {} as MedusaResponse
+    const nextFunction = jest.fn()
+
+    mockRequest.additionalDataValidator = zod
+      .object({
+        brand_id: zod.number(),
+      })
+      .nullish()
+
+    const validatorFactory = (
+      schema?: Zod.ZodOptional<Zod.ZodNullable<Zod.ZodObject<any, any>>>
+    ) => {
+      return schema
+        ? createLinkBody().extend({
+            additional_data: schema,
+          })
+        : createLinkBody()
+    }
+
+    let middleware = validateAndTransformBody(validatorFactory)
+
+    await middleware(mockRequest, mockResponse, nextFunction)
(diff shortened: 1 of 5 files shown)


sradevski (approved):
LGTM, I think we just need to make sure it doesn't break anything since we are so close to the official release :)

adrien2p (approved):

thetutlage:
> LGTM, I think we just need to make sure it doesn't break anything since we are so close to the official release :)

Yeah, unfortunately the nature of the change is breaking, since we are going from making something required to optional. However, since we have established that the current behavior is invalid, we have push the fix :)

olivermrbl (approved):
LGTM

thetutlage:
@olivermrbl What do you think of the question in the PR description?

olivermrbl:
> Question: Do we want to set it to an empty object when its not defined in an API call? Or do we want to update the docs to handle the case.

If it's not provided in the payload, I would expect it to be `undefined`, similar to how other properties are treated.

What do you think?

thetutlage:
> If it's not provided in the payload, I would expect it to be undefined, similar to how other properties are treated.

Okay, in that case we have to fix the workflow hooks documentation to account of `additional_data` to be `undefined`.

olivermrbl:
cc @shahednasser

shahednasser:
Thanks for letting me know! I've been adding the check for undefined in most places but probably missed it in that documentation page, will check other places as well 👍🏻

thetutlage:
@olivermrbl @adrien2p @shahednasser 

Between, I see that the `additional_data` in hooks gets converted to a defined property (as per TypeScript types), even though it is marked as optional in the workflow payload.

So let me look into it and see if we have to fix that as-well

thetutlage:
@olivermrbl @shahednasser @adrien2p 

Regarding the previous message, it was more of an issue with us not enabling `strictNullChecks` in `tsconfig.json` file and not an implementation problem.

PR #9689, opened 2024-10-21
docs: fix details on setting / updating customer ID in cart

shahednasser:
Should only be merged after release


olivermrbl (approved):
We can merge this now

PR #9691, opened 2024-10-21
feat(dashboard,types): Translate all OrderBy filters

kasperkristensen:
Resolves CC-616

**Note**
- I have added some missing fields to a couple of types, to resolve TS errors in the dashboard.
- For the ApiKey types I have left the created_at, updated_at, etc. as type Date, as changing them to string, caused @medusajs/medusa to fail building. We need to clean up the types at a later time.


olivermrbl (commented):
LGTM

olivermrbl (approved):

PR #9692, opened 2024-10-21
fix(dashboard): temp hide order column on reservations

fPolic:
**What**
- hide the order column on the reservations list until a link is introduced + impl. of `listOrderItems`


kasperkristensen (approved):

olivermrbl (approved):
LGTM, but we should prob. keep the Linear issue open right?

PR #9693, opened 2024-10-21
fix: API validation management issues

adrien2p:
**What**
Currently, the API validation layer is broken in both responsibilities and validation itself.
This pr introduce the following fixes and patterns:
- Always create a `*Fields` schema that only takes care of defining the schema validation without `effect`
- Use the previous point into the API schema validator including `$and` and `$or` capabilities plus the recursive effects
- remove `normalizeArray` which does not have to exists since array are already treated as they should
- Add recursive transformation to take into account `$and` and `$or` as well or any other similar operators
- New util `applyAndAndOrOperators` to wrap the management of those operators and to be merged to an existing schema

Tasks
- [x] store domain
- [ ] admin domain


packages/medusa/src/api/utils/common-validators/common.ts
```diff
@@ -25,6 +25,21 @@ export const BigNumberInput = z.union([
...
...y<ZodObject<any>>>>}>]: undefined extends baseObjectOutputType<{$and: ZodOptional<ZodLazy<ZodArray<ZodObject<any>>>>, $or: ZodOptional<ZodLazy<ZodArray<ZodObject<any>>>>}>[k] ? never : k}[keyof {$and: ZodOptional<ZodLazy<ZodArray<ZodObject<any>>>>, $or: ZodOptional<ZodLazy<ZodArray<ZodObject<any>>>>}]>]: objectUtil.addQuestionMarks<baseObjectOutputType<{$and: ZodOptional<ZodLazy<ZodArray<ZodObject<any>>>>, $or: ZodOptional<ZodLazy<ZodArray<ZodObject<any>>>>}>, {[k in keyof baseObjectOutputType<{$and: ZodOptional<ZodLazy<ZodArray<ZodObject<any>>>>, $or: ZodOptional<ZodLazy<ZodArray<ZodObject<any>>>>}>]: undefined extends baseObjectOutputType<{$and: ZodOptional<ZodLazy<ZodArray<ZodObject<any>>>>, $or: ZodOptional<ZodLazy<ZodArray<ZodObject<any>>>>}>[k] ? never : k}[keyof {$and: ZodOptional<ZodLazy<ZodArray<ZodObject<any>>>>, $or: ZodOptional<ZodLazy<ZodArray<ZodObject<any>>>>}]>[k_1]}, {[k_2 in keyof baseObjectInputType<{$and: ZodOptional<ZodLazy<ZodArray<ZodObject<any>>>>, $or: ZodOptional<ZodLazy<ZodArray<ZodObject<any>>>>}>]: baseObjectInputType<{$and: ZodOptional<ZodLazy<ZodArray<ZodObject<any>>>>, $or: ZodOptional<ZodLazy<ZodArray<ZodObject<any>>>>}>[k_2]}>["_def"]["catchall"]>}
```
olivermrbl: q: this intentional? 😅

packages/medusa/src/api/utils/validators.ts
```diff
@@ -46,6 +46,17 @@ export const createLinkBody = () => {
   })
 }
 
+export function omitFindParams(schema: z.ZodObject<any>) {
```
olivermrbl: q: is this used?

olivermrbl (approved):
LGTM

packages/medusa/src/api/utils/common-validators/common.ts
adrien2p: my bad, it slipt in by mistake 😂

PR #9697, opened 2024-10-21
fix(dashboard,js-sdk): remove methods / hooks to non existing exchange routes

shahednasser:
Remove the following exchange methods (and hooks from dashboard) which point to routes that don't exist:

- `delete`
- `addItems`
- `updateItem`
- `removeItem`


fPolic (approved):
LGTM!

packages/core/js-sdk/src/admin/exchange.ts
```diff
@@ -73,72 +73,6 @@ export class Exchange {
...
-        body,
-        query,
-      }
-    )
-  }
-
-  async updateItem(
-    id: string,
-    actionId: string,
-    body: HttpTypes.AdminUpdateExchangeItem,
```
fPolic, line 111: nit: think that this types can be removed as well

shahednasser: Good point! will remove them

PR #9705, opened 2024-10-22
Feat/validate query enhancement

adrien2p:


packages/core/framework/src/http/utils/validate-query.ts
```diff
@@ -68,14 +68,19 @@ export function validateAndTransformQuery<TEntity extends BaseEntity>(
     next: NextFunction
   ) {
     try {
+      const restricted = req.restrictedFields.list()
```
olivermrbl: thought: Not sure if we do this already, but we should probably add a strong recommendation in our docs that you should use `validateAndTransformQuery` for custom endpoints in the store domain

olivermrbl: And to ensure best practices are followed, it should probably be the recommendation for any custom endpoint. Same with `validateAndTransformBody`

adrien2p: yes in theory we are already recommending that especially if we talk about the query config

adrien2p: and then if people want something very custom they can just reuse the tools that we are using under the hood in those middleware. But then it is up to the user to manage everything if they don't use our recommended way

adrien2p: Apparently, only the *body is mentioned in the documentation, should we ask shahed to have a look?

Screenshot 2024-10-22 at 11 41 13

olivermrbl: cc @shahednasser

adrien2p: @olivermrbl do we also want to document ? or we do that later?
- `DEFAULT_STORE_RESTRICTED_FIELDS` from `framework/utils`
- `RestrictedFields` from `framework/http`
- medusa config `prjectConfig.http.restrictedFields`

packages/medusa/src/api/store/customers/query-config.ts
```diff
@@ -38,6 +38,10 @@ export const defaultStoreCustomerAddressFields = [
 
 export const retrieveAddressTransformQueryConfig = {
   defaults: defaultStoreCustomerAddressFields,
+  allowed: [
```
olivermrbl: todo: Can you add `validateAndTransformQuery` to the middlewares for the `GET /store/customers/me` endpoint? It is currently missing, so we can't query anything 😄

adrien2p: it is present already

Screenshot 2024-10-22 at 14 44 14

olivermrbl: Oh my bad, but looking at the wrong middelware config

olivermrbl: todo: Can you add `allowed` to `retrieveTransformQueryConfig` in this file? 🙏

olivermrbl: And remove `orders` here. This is for customer addresses

adrien2p: done ✅

olivermrbl (approved):
LGTM!

PR #9710, opened 2024-10-22
fix(dasbhoard): Revert keybound form for public pages

kasperkristensen:
**What**
- We recently changed all forms to require Cmd/Ctrl + Enter to submit, this makes sense for our forms within modals, but not so much for our public pages such as the login page.
- This PR removes the need to hold Cmd or Ctrl when trying to submit the forms on the login page, reset password page, and register page.
- Also fixes some incorrect styling on the reset password and register pages.


olivermrbl (approved):
Nice, makes sesne

PR #9715, opened 2024-10-22
fix(dashboard): truncate price cell titles

fPolic:
Before:
Screenshot 2024-10-22 at 17 14 56

After
Screenshot 2024-10-22 at 17 26 15


packages/admin/dashboard/src/components/data-grid/helpers/create-data-grid-price-columns.tsx
```diff
@@ -62,9 +66,9 @@ export const createDataGridPriceColumns = <
         type: "number",
         header: () => (
           <div className="flex w-full items-center justify-between gap-3">
```
kasperkristensen, line 68: The icon is still cut off, so we should either change the wrapping div to a grid with `grid-cols-[1fr_15px] gap-3`, or add a `shrink-0` to the icon.

olivermrbl:
@kasperkristensen, could you give this another round of review?

kasperkristensen (approved):

PR #9721, opened 2024-10-22
chore: Remove v2 branches from starters

olivermrbl:


packages/cli/medusa-cli/src/commands/new.ts
```diff
@@ -187,7 +187,7 @@ const clone = async (hostInfo, rootPath, v2 = false, inputBranch) => {
       : []
 
   if (v2) {
-    branch = [`-b`, inputBranch || "feat/v2"]
+    branch = [`-b`, inputBranch || "master"]
```
shahednasser, line 190: We don't need this option anyway so better to remove this.

shahednasser (approved):
LGTM just one suggestion

packages/cli/medusa-cli/src/commands/new.ts
olivermrbl: The `inputBranch`?

shahednasser: The `v2` condition, we don't need a `--v2` option anymore

olivermrbl: I noticed this is a more comprehensive change, so I'll save this cleanup for later.

PR #9736, opened 2024-10-23
fix(core-flows, dashboard): adjust stock levels when doing partial fulfilments

fPolic:
**What**
- core-flows: stock levels were adjusted for the entire item quantity instead of the quantity that is being fulfilled
- admin: count the reserved quantity of the item we are fulfilling as available quantity in admin
- admin: refresh reservations after creation order fulfillment

---

CLOSES TRI-418


packages/core/core-flows/src/order/workflows/create-fulfillment.ts
```diff
@@ -202,20 +202,20 @@ function prepareInventoryUpdate({
 
     const inputQuantity = inputItemsMap[item.id]?.quantity ?? item.quantity
 
-    const quantity = reservation.quantity - inputQuantity
+    const remainingReservationQuantity = reservation.quantity - inputQuantity
```
olivermrbl, line 212: q: should we check if input quantity exceeds the reserved quantity?

olivermrbl (approved):
LGTM, just had one q before we merge

PR #9747, opened 2024-10-24
Update docs commerce modules so that tax links to the correct page

jackcaldwell:
### What

Update the href on the tax card on https://docs.medusajs.com/v2 to correctly link to https://docs.medusajs.com/v2/resources/commerce-modules/tax

### Why 

It is currently incorrectly linking to https://docs.medusajs.com/v2/resources/commerce-modules/stock-location

### How

Update the href value.

### Testing

To test open the docs page and verify that the link navigates to the correct page.


shahednasser (approved):
Nice catch thanks!

PR #9748, opened 2024-10-24
fix(medusa): Admin validator  filtering and pagination

adrien2p:
**What**
Fix usage of $and and $or operator and pagination inclusion


olivermrbl (approved):
LGTM, nice getting cleaned up 👍

PR #9753, opened 2024-10-24
docs: show latest version in docs + update automatically

shahednasser:
- Show latest Medusa version in the docs navigation bar
- When there's a new release, we show a badge on that version. The badge is gone after the user hovers it
- Clicking on the version takes you to the release notes
- Added an action to automatically update the version when a new release is out

Demo:

https://github.com/user-attachments/assets/701e131b-a2b7-477b-973e-623295891e2e


olivermrbl (approved):
Nice!

PR #9781, opened 2024-10-24
fix(orchestration): Ids wrongly processed when using operators map

adrien2p:
FIXES FRMW-2754

**What**
Currently, it is not possible to filter a primary key using any operator map operator because of the check and manipulation we are doing. I ve added a check to verify if we are manipulating an object and if that object contains any key part of the supported operators, in that case we drop the manipulation and let it pass as it is.


integration-tests/modules/__tests__/modules/remote-query.spec.ts
```diff
@@ -265,7 +268,27 @@ medusaIntegrationTestRunner({
...
+        expect(data).toEqual([
+          expect.objectContaining({
+            id: product.id,
+            title: product.title,
+          }),
+        ])
+      })
+
       it(`should perform cross module query and apply filters correctly to the correct modules [1]`, async () => {
+        console.log("original product", product.id)
```
olivermrbl: rm
```suggestion
```

olivermrbl (approved):
LGTM

packages/core/orchestration/src/joiner/remote-joiner.ts
```diff
@@ -440,9 +442,18 @@ export class RemoteJoiner {
...
+      isObject(ids) && Object.keys(ids).some((key) => !!FilterOperatorMap[key])
 
-    if (uniqueIds) {
+    let uniqueIds = isIdsUsingOperatorMap
+      ? ids
+      : Array.isArray(ids)
+      ? ids
+      : ids
+      ? [ids]
+      : undefined
```
thetutlage: Do we have an additional conditional here? I am not sure, but looks like we can write the following. Please double check :)

```ts
let uniqueIds = isIdsUsingOperatorMap
      ? ids
      : Array.isArray(ids)
      ? ids
      : [ids]
```

adrien2p: we want to wrap into an array if it is not null or undefined, that is why there is the extra check before wrapping in array or setting it to undefined. But I can clean it up differently

olivermrbl:
@adrien2p, the latest change didn't seem to work as intended. Should we revert? Don't think it did much for readability anyway.

adrien2p:
> @adrien2p, the latest change didn't seem to work as intended. Should we revert? Don't think it did much for readability anyway.

I ve just improved it, let me know what you think

olivermrbl (approved):

PR #9792, opened 2024-10-25
fix(order): add fulfillment labels query key on create shipment page

Kadphol:
What
- Empty tracking number on create shipment page when tracking number is already existed.

Why
- Create shipment component already used fulfillment label items but `useOrder` on `order-create-shipment` page didn't include `*fulfillments.labels` key

How
- Add `*fulfillments.labels` key to `useOrder` on `order-create-shipment`


riqwan (approved):

riqwan:
Thanks for the contribution @Kadphol, apologies for the delay! :

PR #9796, opened 2024-10-25
fix(dashboard): tax override reference

fPolic:
**What**
- tax override references were saved in plural and the tax module query was looking for rules with singular

---

FIXES CC-639
FIXES CC-646


olivermrbl (commented):
Oh interesting, nice catch. LGTM

olivermrbl (approved):

fPolic:
should we mention on the issue that users need to recreate their overrides (or change reference in the DB)

olivermrbl:
> should we mention on the issue that users need to recreate their overrides (or change reference in the DB)

I'll add it to the release notes and we can reference those in the issue.

olivermrbl:
@fPolic, did you also validate the tax calculation on items with a tax override?

olivermrbl:
AFAICT, we actually only account for Product, Shipping and Product Type overrides in the backend

fPolic:
>@fPolic, did you also validate the tax calculation on items with a tax override?

 yup, tested with a product override and works as expected
 
 > AFAICT, we actually only account for Product, Shipping and Product Type overrides in the backend
 
 yes, good point, should we comment out the rest until we add support?

olivermrbl:
> yes, good point, should we comment out the rest until we add support?

Yes, let's do that for now. We might remove these entirely and only support Products, Product Types, and Shipping, but should be good to comment it out now.

PR #9799, opened 2024-10-25
chore(dashboard): Setup test and script for validating i18n

kasperkristensen:
**What**
- Adds a test that validates that en.json and $schema.json matches, this will help us catch missing translations, and prevent us from forgetting to update the schema when we delete/add new keys.
- Adds a script for validating translations against the translation schema. This can be used to validate that community PRs for i18n contain all the required keys. To use the script you can run `yarn i18n:validate <file name>` e.g. `yarn i18n:validate da.json` which will look for a da.json file in the translation folder, and validate it against the schema. We handle this with a script as we don't want to do so through a test. Doing it with a test would mean that if we update the schema, we would also have to update all translations files that we don't maintain ourselves. The purpose of the script is just to allow us to easily review community PRs and also as a tool for people opening PR's to check their translations agains the schema, as it will print missing/additional keys.
- Also adds a script to generate a schema from the en.json file. After adding/deleting keys to en.json you should run `yarn i18n:generate`.


packages/admin/dashboard/package.json
```diff
@@ -7,6 +7,9 @@
     "build": "tsup && node ./scripts/generate-types.js",
     "build:preview": "vite build",
     "preview": "vite preview",
+    "test": "vitest --run",
+    "i18n:validate": "node ./scripts/i18n/validate-translation.js",
+    "i18n:schema": "node ./scripts/i18n/generate-schema.js",
```
sradevski, line 12: I feel like it would be very easy to forget to run this. Should we add a test in our CI that after running this command there are no changes (i.e we haven't forgotten to run it after changing en.json)?

sradevski (approved):
LGTM, just one comment!

packages/admin/dashboard/package.json
kasperkristensen: That's what the test I added ensures. It checks if $schema.json === en.json. So if we make an update to en.json or the schema without updating the other the test will fail and block your PR.

PR #9818, opened 2024-10-27
chore: update common.ts

eltociear:
addreses -> address

### DIFF
--- a/packages/core/types/src/customer/common.ts
+++ b/packages/core/types/src/customer/common.ts
@@ -103,7 +103,7 @@ export interface CustomerAddressDTO {
 export interface FilterableCustomerAddressProps
   extends BaseFilterable<FilterableCustomerAddressProps> {
   /**
-   * Searches for address by properties such as name and street using this search term.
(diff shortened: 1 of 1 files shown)


packages/core/types/src/customer/common.ts
```diff
@@ -103,7 +103,7 @@ export interface CustomerAddressDTO {
 export interface FilterableCustomerAddressProps
   extends BaseFilterable<FilterableCustomerAddressProps> {
   /**
-   * Searches for addreses by properties such as name and street using this search term.
+   * Searches for address by properties such as name and street using this search term.
```
shahednasser:
```suggestion
   * Searches for addresses by properties such as name and street using this search term.
```

shahednasser (changes requested):
Nice catch! But I think it should still be plural.

packages/core/types/src/customer/common.ts
eltociear: Thanks!

shahednasser (approved):
LGTM thanks!

PR #9832, opened 2024-10-28
docs: add common redirects to v1 docs

shahednasser:
Add common redirects to v1 docs that don't affect v2 docs

### DIFF
--- a/www/apps/book/next.config.mjs
+++ b/www/apps/book/next.config.mjs
@@ -159,7 +159,7 @@ const nextConfig = {
       },
       {
         source: "/recipes/:path*",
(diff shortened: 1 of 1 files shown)


www/apps/book/next.config.mjs
```diff
@@ -157,6 +157,21 @@ const nextConfig = {
         destination: "/:path*",
         permanent: true,
       },
+      {
```
olivermrbl, line 160: q: Any reason we shouldn't redirect to the v2 docs here?

shahednasser: Good point! Fixed

olivermrbl (approved):

Issue #6889, opened 2024-03-31
Wrong customer/ last registered user mapped to guest user cart after adding shipping address


sumanchy88:
### Describe the bug
For guest user, while adding product to cart and after filling the shipping address details. On click of Continue to delivery button, last registered customer mapped to guest user cart.

### System information

Medusa version (including plugins):
Node.js version: v16.15.1
Database: Postgres
Operating system: MacOS
Browser (if relevant): Chrome

### Steps to reproduce the behavior

1. Create a sample store fornt App running 
2. Registered a logged in user by click on Sign-In.
3. Logout from the account.
4. Select any product to cart and then click on Go to Cart
5. Add the shipping address details and email address
6. Click on Continue to delivery, observe the response of shipping address with email mapped to last registered account
7. Check the enteries in database for cart id -> customer_id mapped to last registered logged in user.

### Expected behavior
customer id should mapped to the guest user account email address.

### Screenshots
Step 1: Add the Shipping address details and different email address
![image](https://github.com/medusajs/medusa/assets/164988194/9163fe6b-801e-42fc-a89b-56972aaa07d4)

Step 2: After cart update, added response mapped to different user account
![image](https://github.com/medusajs/medusa/assets/164988194/09269fc5-3f0b-4b48-b27d-a227741ef6e1)

If applicable, add screenshots to help explain your problem

### Code snippets

If applicable, add code samples to help explain your problem

### Additional context

Add any other context about the problem here

MgStoreLK:
@kasperkristensen Please escalate this issue if possible. Thanks
Edit: I had to restart my storefront (starter template) ec2 instance to work around this issue

sumanchy88:
@kasperkristensen : Have you got chance to lokk into the issue. This is bit critical for ecommerce application with guest flow.

alsherif-khalaf:
same issue

alsherif-khalaf:
it is a critical issue in storefront

u11d-bartlomiej-galezowski:
Any update? @kasperkristensen @sradevski @olivermrbl @shahednasser 
I've also notice the issue.

devcshort:
https://github.com/medusajs/medusa/pull/9848 Fixes this issue

devcshort:
> This issue is stale because it has been open 30 days with no activity. Remove stale label or comment or this will be closed in 3 days.

Not stale

devcshort:
[...]

PR #9848, opened 2024-10-28
fix: Don't store JWT token in an SSR environment

devcshort:
**What** - This change makes it so that JWT doesn't get stored on the Medusa client when in an SSR environment.

**Why** - Currently when the Medusa client is used in an SSR environment, the last logged in user's JWT gets stored on the Medusa client. This causes customer information to be leaked when it shouldn't be.

**How** - To keep things DRY, I added a typeof check to the jwt-token-manager that checks if the window is undefined. If it's undefined, I'm simply just returning as I don't believe any further action is needed.

This fixes the issue referenced here https://github.com/medusajs/medusa/issues/6889


u11d-bartlomiej-galezowski:
@sradevski @olivermrbl @shahednasser can we ask for your review? This fix is really urgent for us.

thetutlage:
Looks fine to me. I will let @olivermrbl and @sradevski look into it once and especially the failing CI tests

devcshort:
@olivermrbl @sradevski any updates on this? This is a fairly urgent request for us and is a prerequisite before we can go to production. Thanks!

olivermrbl (approved):
LGTM, thanks!

u11d-bartlomiej-galezowski:
Can we expect the fix in v1.20.11?

devcshort:
> This PR is stale because it has been open 30 days with no activity. Remove stale label or comment or this will be closed in 5 days.

Not stale

devcshort:
@olivermrbl @kasperkristensen what is your release cycle like? Curious when we can expect this change for the v1 Medusajs client to be released. Thanks!

SalahAdDin:
> @olivermrbl @kasperkristensen, what is your release cycle like? I'm curious when we can expect this change for the v1 Medusajs client to be released. Thanks!

Being a security issue, they may release it to version 1 too.

olivermrbl:
@devcshort, there is no release cycle for v1, since PRs are very rarely merged. We tend to do it shortly after this happens, so I expect to cut a new version later today.

devcshort:
@olivermrbl sounds great, thank you for the update!

Issue #9849, opened 2024-10-28
Unable to change email address of logged in user during checkout


devcshort:
### Describe the bug

When a user is logged in and tries to go through the checkout flow, they will be forced in to using their account specific email regardless of what they type in to the email address input.

### System information

Medusa version (including plugins): 1.20.10
Node.js version: 20.12.2
Database: Postgres 16.3
Operating system: macOS Sonoma 14.5

### Steps to reproduce the behavior

1. Log in to an account.
2. Add an item to your cart.
3. Go to the checkout page.
4. You should see your account email in the input. Change it to something different.
5. Click `Continue to delivery`
6. Check the address information for your email. It should have reverted to your account email instead of what was typed in.

### Expected behavior

When a user types in an email address, that email should be used regardless of what is tied to their account.

devcshort:
Fixed by https://github.com/medusajs/medusa/pull/9850

PR #9850, opened 2024-10-28
fix: Use the typed in email address when available instead of forcing…

devcshort:
**What** - This change fixes checkout flow so that a typed in email address will be used.

**Why** - Currently when a customer is logged in, the Medusa backend forces them to use their account email address for ALL orders, regardless of what they type in to the email input.

**How** - After the cart has been updated with the logged in customer's information I check to see if an email was provided. If one was provided, I use that one for the cart instead.

This fixes the issue referenced here https://github.com/medusajs/medusa/issues/9849


devcshort:
Not stale

PR #9892, opened 2024-11-01
Feat: add german language support

pixelwiese:
**What**
Added German language json file. 

**Why**
To be able to choose german language.

**How**
The File is translated with Google Translate, some human checks and corrections. I'm sure it's not perfect, but it is a good starting point. First the en.json was duplicated (as de.json) to be able to see the changes made and if the structure is still the same.

**Testing**
As mentioned in How, i checked if the structure is still the same.


kasperkristensen (commented):
Hi @pixelwiese,

There are a couple of keys that are incorrect. You can see a summary here: https://github.com/medusajs/medusa/actions/runs/11775697868/job/32796644875?pr=9892

You can also run `yarn i18n:validate de.json` in the /dashboard folder to check your translation.

Otherwise the PR looks good, so once the required changes have been made we can merge it.

kasperkristensen (approved):

PR #9902, opened 2024-11-03
Update Order service's update method

lukebui:
Since `no_notification` is an optional boolean, should check if it's empty or not.


adrien2p (approved):
LGTM

PR #9917, opened 2024-11-04
fix(utils): Mikro orm repository update many to many should detach all items by default

adrien2p:
FIXES FRMW-2758

**What**
In the current state, when calling a module service update method and providing a partial array for a many to many collection,. mikor orm automatically handle detaching the items that are not present in the array. But in the case an empty array is provided, nothing happen when we would expect all items to be detached automatically. This PR aims at fixing this behaviour (which is expected by mikro orm).

**Note**
As per mikro orm documentation, when relying on auto detach from a collection (such as using remove or remove all), one query per removal will be created, which is not performant. The expectation would be that the end user manage his collection manually and be aware of what is happening by explicitly managing those cases.  for example, providing explicit methods to detach X items from a collection instead of relying on updating deeply from top to bottom.
But in the end, we still provide this shortcut and the user can choose, for simple cases where not a lot of items are involve, to use or not this simili cascade behaviour.


thetutlage:
Sounds good to me! Maybe, we can highlight this behaviour change in our changelog, incase someone was passing empty array during API calls and now they will see that this behaviour results in detach within the pivot table.

thetutlage (approved):

olivermrbl (approved):

PR #9929, opened 2024-11-05
fix(core-flows,medusa): Include region_id in shipping option retrieval

### LINKED ISSUE #9906: Tax inclusive pricing for shipping options not respecting region's settings
### Describe the bug

When I enable tax inclusive pricing for a region and set a shipping option's price for that region, the shipping option’s `is_tax_inclusive` value will be `false`, which causes it to be calculated incorrectly in the cart.

Currently, the workaround is to set the currency as tax inclusive as well, which changes `is_tax_inclusive` to `true`. However, I’m not sure this is the correct setup.

### System information

Medusa version (including plugins): 2.0.1
Node.js version: 20.17.0
Database: PostgreSQL 16.4
Operating system: Windows 11
Browser (if relevant): Edge

1. Create a region with tax inclusive pricing enabled.
2. Create a shipping option for the manual provider.
3. Update the shipping option's price for the created region.
4. Call `medusaClient.store.fulfillment.listCartOptions` via medusa js sdk.
5. You'll see `"is_tax_inclusive": false` in the response.

### Expected behavior

I would expect for the shipping option pricing to respect a region's settings.

olivermrbl:
### What
Include `region_id` when retrieving shipping options for a cart

### Why
Otherwise, region-specific prices will never show in the Store API

Closes CMRC-655 https://github.com/medusajs/medusa/issues/9906


fPolic (approved):

integration-tests/modules/__tests__/shipping-options/store/shipping-options.spec.ts
```diff
@@ -1,9 +1,9 @@
+import { medusaIntegrationTestRunner } from "@medusajs/test-utils"
```
adrien2p, line 1: discussion: I feel like we have http integration in both `integration-tests/http` and `integration-tests/modules`. Originally, modules was more focused on testing the workflows because of the lack of API and location to do that. We might need to rethink that structure at some point to something more simpler and less confusing maybe?
something like
```text
- integration-tests
  - __tests__
    - cart
      - api
        -  store
        - admin
      - (other dir if needed for other kind of integration tests)
```

integration-tests/modules/__tests__/shipping-options/store/shipping-options.spec.ts
```diff
@@ -165,35 +190,77 @@ medusaIntegrationTestRunner({
...
-              currency_code: "usd",
-              email: "test@admin.com",
-              items: [
-                {
-                  variant_id: product.variants[0].id,
-                  quantity: 1,
-                },
-              ],
-            },
+      describe("GET /admin/shipping-options?cart_id=", () => {
```
adrien2p:
```suggestion
      describe("GET /store/shipping-options?cart_id=", () => {
```

integration-tests/modules/__tests__/shipping-options/store/shipping-options.spec.ts
```diff
@@ -165,35 +190,77 @@ medusaIntegrationTestRunner({
...
+            expect.objectContaining({
+              id: shippingOption.id,
+              name: "Test shipping option",
+              amount: 1100,
+              price_type: "flat",
+            })
+          )
+        })
+
+        it("should get shipping options with region prices  for a cart successfully", async () => {
```
adrien2p:
```suggestion
        it("should get shipping options with region prices for a cart successfully", async () => {
```

adrien2p: Q: Both tests are doing the same thing, Maybe we could just merge them and do something like
- create cart with a region
- retrieve shipping options
- expectations
- update cart region
- retrieve shipping options
- expectation

I find it confusing to have two different tests that does the same thing except that if you read the description of the first one we don't know that we are looking for region prices as well. 

Let me know if I overlooked at something 👍

adrien2p (approved):
LGTM, a few questions but nothing major :)

integration-tests/modules/__tests__/shipping-options/store/shipping-options.spec.ts
olivermrbl: Yeah, I think you are right, but that's probably for a separate PR targeting a clean-up of integration tests.

integration-tests/modules/__tests__/shipping-options/store/shipping-options.spec.ts
olivermrbl: Sure, we can merge the two

PR #9935, opened 2024-11-05
fix(types): Prev limit wrong values

adrien2p:
**What**
The values are wrongly set which can lead to infinite depth iteration

**Note**
This will change on how we manage those cases in the future but currently it fixes an issue


olivermrbl (approved):

carlos-r-l-rodrigues (approved):
do we need a changeset?

olivermrbl:
@carlos-r-l-rodrigues, nice catch! We need changesets on all PRs going forward :) 

We were discussing a GH action to check if they were missing with @thetutlage

PR #9939, opened 2024-11-05
fix(telemetry): Pointing and resolving packages wrongly

adrien2p:
FIXES FRMW-2764

**What**
- Fix wrong CLI package
- Fix package.json path resolution


packages/medusa-telemetry/src/telemeter.js
```diff
@@ -82,7 +82,10 @@ class Telemeter {
 
   getMedusaVersion() {
     try {
-      const packageJson = require.resolve(`@medusajs/medusa/package.json`)
+      const packageJson = join(
+        require.resolve(`@medusajs/medusa`).split(`${sep}dist`).shift(),
```
thetutlage: Can we instead expose the package.json file from the package exports?

adrien2p: yes we can, let me do that

olivermrbl (approved):

PR #9941, opened 2024-11-05
fix: commands to run from within the `.medusa/server` directory

thetutlage:
Fixes: DX-1045

Tried it myself on Railway and have the project working with these changes


www/apps/resources/app/deployment/medusa-application/railway/page.mdx
```diff
@@ -47,7 +47,7 @@ So, add the following configuration in `medusa-config.ts`:
 module.exports = defineConfig({
   projectConfig: {
     // ...
-    workerMode: process.env.MEDUSA_WORKER_MODE,
+    workerMode: process.env.MEDUSA_WORKER_MODE as "shared" | "worker" | "server",
```
adrien2p, line 50: Q: would it make sense to change centralize this type as it is being used in many places

www/apps/resources/app/deployment/medusa-application/railway/page.mdx
```diff
@@ -219,7 +219,7 @@ ADMIN_CORS= # ADMIN URL
 AUTH_CORS= # STOREFRONT AND ADMIN URLS, SEPARATED BY COMMAS
 # change to false if you're hosting the admin with the application
 DISABLE_MEDUSA_ADMIN=true
-MEDUSA_WORKER_MODE=worker
+MEDUSA_WORKER_MODE=server
```
adrien2p, line 222: Q: I am wondering if the default shouldn't be shared? wdyt? Otherwise new users might wonder why the events are not being consume at all with their first deployment

www/apps/resources/app/deployment/medusa-application/railway/page.mdx
```diff
@@ -245,7 +245,7 @@ To set the `start` command of your Medusa application in server mode:
 4. For the “Custom Start Command” field, enter the following and click the check mark button:
 
 ```bash npm2yarn
-npm run predeploy && npm run start
+cd .medusa/server && npm run predeploy && npm run start
```
adrien2p, line 248: Q: Do we need to add `npm i` ? I might just have missed it ahah. Same question for line 353

adrien2p (approved):

kasperkristensen (approved):
LGTM, just one question: 

Did you validate that everything is ran in production mode when deploying to Railway? Since Shahed was able to deploy with the previous version of the guide I suspect that Railway does not set NODE_ENV to production for you, and therefore she ended up with the development version of admin being ran. If Railway does not set NODE_ENV to production per default then we need a note about needing to do that.

thetutlage:
> Did you validate that everything is ran in production mode when deploying to Railway? Since Shahed was able to deploy with the previous version of the guide I suspect that Railway does not set NODE_ENV to production for you, and therefore she ended up with the development version of admin being ran. If Railway does not set NODE_ENV to production per default then we need a note about needing to do that.

Yes. I personally verified everything and here is the deployed version. https://deployment-on-railway-production.up.railway.app/app

And yes, we have to set the `NODE_ENV=production`

www/apps/resources/app/deployment/medusa-application/railway/page.mdx
thetutlage: Yeah we can do that!

www/apps/resources/app/deployment/medusa-application/railway/page.mdx
thetutlage: I have zero opinions on it. But in the same guide, we deploy another service in `worker` mode.

www/apps/resources/app/deployment/medusa-application/railway/page.mdx
thetutlage: It is done by Railway automatically by detecting the node app.

www/apps/resources/app/deployment/medusa-application/railway/page.mdx
adrien2p: then it is good 👍

adrien2p (approved):

olivermrbl (approved):

PR #9942, opened 2024-11-05
fix: Support relative path in SDK

sradevski:
This should allow for the SDK to support calls made to the domain where it's running (the origin)

### DIFF
--- a/.changeset/cool-buckets-remember.md
+++ b/.changeset/cool-buckets-remember.md
@@ -1,0 +1,5 @@
+---
+"@medusajs/js-sdk": patch
+---
+
+fix: Add support for relative base URL to js-sdk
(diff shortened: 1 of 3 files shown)


kasperkristensen (approved):
LGTM, should we update the default value here https://github.com/medusajs/medusa/blob/674c86eaa066aa2a9f296f464052d21acaf209f9/packages/admin/dashboard/src/lib/client/client.ts#L3 to "/" in this PR as well so we can remove the requirement to add a backendUrl?

```
export const backendUrl = __BACKEND_URL__ !== undefined ? __BACKEND_URL__ : "/" // Ensure that we use __BACKEND_URL__ even if it is the nullish value "".
```

If you prefer we can do it in another PR

sradevski:
@kasperkristensen I prefer a separate PR for it as I wanted to check all places where we are potentially setting/using this value (there were at least a couple)

PR #9944, opened 2024-11-06
fix: log cause property from the error

thetutlage:
Fixes: FRMW-2762

This PR fixes the error reporting to include the original error message and the error cause when logging it using the JSON formatter of Winston.

> Side note: There is still a lot of room for improvements in our logging. But that require a overhaul cleanup in multiple places and will require dedicated loggers for commands and APP and special treatment for pretty printing errors.

## Error output in JSON format

**Before**
CleanShot 2024-11-06 at 13 12 32@2x

**After**
CleanShot 2024-11-06 at 13 11 28@2x

### DIFF
--- a/packages/cli/medusa-cli/src/reporter/index.ts
+++ b/packages/cli/medusa-cli/src/reporter/index.ts
@@ -179,7 +179,7 @@ export class Reporter {
     let message: string
     let errorAsObject: Error | undefined
 
-    if (isObject(messageOrError)) {
+    if (messageOrError && typeof messageOrError === "object") {
       errorAsObject = messageOrError
       message = messageOrError.message
     } else if (error) {
@@ -215,7 +215,7 @@ export class Reporter {
      * Otherwise, we always loose the message property from the
      * actual error object
      */
-    if (errorAsObject?.message !== message) {
+    if (errorAsObject?.message && errorAsObject?.message !== message) {
       this.loggerInstance_.log({ level: "error", message })
     }
     this.loggerInstance_.log(toLog)
--- a/packages/cli/medusa-cli/src/reporter/index.ts
+++ b/packages/cli/medusa-cli/src/reporter/index.ts
@@ -1,9 +1,10 @@
-import { track } from "@medusajs/telemetry"
 import ora from "ora"
-import { inspect } from "util"
-import stackTrace from "stack-trace"
 import { ulid } from "ulid"
 import winston from "winston"
+import { inspect } from "util"
+import stackTrace from "stack-trace"
+import { isObject } from "@medusajs/utils"
+import { track } from "@medusajs/telemetry"
 import { panicHandler } from "./panic-handler"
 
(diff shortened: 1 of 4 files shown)


packages/cli/medusa-cli/src/reporter/index.ts
```diff
@@ -174,27 +175,57 @@ export class Reporter {
    * Level 0
    */
   error(messageOrError: string | Error, error?: Error) {
-    let message = messageOrError as string
+    let message: string
+    let errorAsObject: Error | undefined
 
-    if (typeof messageOrError === "object") {
+    if (messageOrError && typeof messageOrError === "object") {
```
adrien2p, line 181: suggestion: `isObject(messageOrError)`

packages/cli/medusa-cli/src/reporter/index.ts
```diff
@@ -174,27 +175,57 @@ export class Reporter {
...
 
+    /**
+     * If "errorAsObject" has a message property, then we will
+     * print the standalone message as one log item and then
+     * the actual error object as another log item.
+     *
+     * Otherwise, we always loose the message property from the
+     * actual error object
+     */
+    if (errorAsObject?.message && errorAsObject?.message !== message) {
```
adrien2p, line 217:
```suggestion
    if (errorAsObject?.message !== message) {
```

packages/cli/medusa-cli/src/reporter/index.ts
```diff
@@ -174,27 +175,57 @@ export class Reporter {
...
-    // Give stack traces and details in dev
-    if (error && IS_DEV) {
-      console.log(error)
+    /**
+     * In dev we print the error using `console.error`. I am not sure if
+     * we should be doing this, because now we are giving more info
+     * in Dev and less info in production
+     */
+    if (errorAsObject && IS_DEV) {
+      console.error(errorAsObject)
```
adrien2p, line 227: discussion: I am not sure about that either, also why are we suddenly using the console API there? I think it should be similar no matter the env right?

adrien2p (approved):

packages/cli/medusa-cli/src/reporter/index.ts
thetutlage: Yeah. But I noticed that Winston doesn't display the error stack when not using the `JSON` formatter. 

This is how it looks when I remove the `console.error` statement

CleanShot 2024-11-06 at 13 59 31@2x

adrien2p: weird 😂 also, winston added support for cause lately 🤔

adrien2p (approved):

packages/cli/medusa-cli/src/reporter/index.ts
thetutlage: They use `logform` for formatting and as per this source, the `CLI` formatter does not print anything other than the error message. https://github.com/winstonjs/logform/blob/master/cli.js

We use `CLI` formatter during development

PR #9947, opened 2024-11-06
fix: Default to a relative path for the admin backend URL

sradevski:


packages/admin/dashboard/vite.config.mts
```diff
@@ -8,7 +8,7 @@ export default defineConfig(({ mode }) => {
   const env = loadEnv(mode, process.cwd())
 
   const BASE = env.VITE_MEDUSA_BASE || "/"
-  const BACKEND_URL = env.VITE_MEDUSA_BACKEND_URL || "http://localhost:9000"
```
kasperkristensen, line 11: We should undo this, as this is the vite config that is used when you run the dashboard from within our monorepo, and in that case it's not on the same server.

packages/admin/dashboard/vite.config.mts
```diff
@@ -8,7 +8,7 @@ export default defineConfig(({ mode }) => {
...
+  const BACKEND_URL = env.VITE_MEDUSA_BACKEND_URL || "/"
```
kasperkristensen: We should undo this, as this is the vite config that is used when you run the dashboard from within our monorepo, and in that case it's not on the same server.

carlos-r-l-rodrigues (approved):

PR #9955, opened 2024-11-06
fix(link-modules): generate graphql type of read only links

carlos-r-l-rodrigues:
What:
- Add read-only links to the generated graphql schema

FIXES: FRMW-2768


olivermrbl (approved):
LGTM, but maybe @adrien2p wants to glance over this too

adrien2p (approved):

PR #9959, opened 2024-11-06
chore: Update flow

olivermrbl:


carlos-r-l-rodrigues (approved):

packages/medusa/src/api/auth/utils/validate-token.ts
```diff
@@ -21,7 +21,7 @@ export const validateToken = () => {
     const req_ = req as AuthenticatedMedusaRequest
 
     if (!token) {
-      return next()
+      return res.status(401).json({ message: "Invalid token" })
```
sradevski: I am just not sure if this stops the execution flow or not, cause IIRC you need to do `next(new Error(''))` to return with an error from a middleware.

Worst case we try to do res.status or res.send lower down the call stack and that fails

sradevski (commented):
LGTM!

packages/medusa/src/api/auth/utils/validate-token.ts
olivermrbl: @sradevski, I believe both works, as validated by the tests, but I can change to use the `new Error` to be consistent with other middleware 👍

adrien2p: I would also go with next(new MedusaError) 👌

adrien2p (approved):

integration-tests/http/__tests__/auth/admin/auth.spec.ts
```diff
@@ -264,6 +264,28 @@ medusaIntegrationTestRunner({
...
+        jest.useFakeTimers()
+
+        // Register user
+        await api.post("/auth/user/emailpass/register", {
+          email: "test@medusa-commerce.com",
+          password: "secret_password",
+        })
+
+        // Advance time by 15 minutes
+        jest.advanceTimersByTime(15 * 60 * 1000)
```
adrien2p, line 278: Q: do we have useRealTimers call in the before each? Just be sure that we reset everytime

adrien2p (approved):

PR #9977, opened 2024-11-07
fix(core-flows): keep same customer on cart update

carlos-r-l-rodrigues:
What:
  * When the email is updated:
    * If the Customer associated with it has an account, only the Cart's email is updated
    * If the Customer is a Guest and the email is different, a new guest customer is created and associated with the Cart

FIXES: CMRC-664

### DIFF
--- a/.changeset/bright-bees-decide.md
+++ b/.changeset/bright-bees-decide.md
@@ -1,0 +1,5 @@
+---
+"@medusajs/core-flows": patch
+---
+
+Keep customer when updating cart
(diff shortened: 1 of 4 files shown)


fPolic (approved):

adrien2p (approved):

olivermrbl:
As discussed, we should change this to create new guest customers, instead of updating the existing one

carlos-r-l-rodrigues:
@olivermrbl, I've updated the flow as discussed.

olivermrbl (approved):

PR #9980, opened 2024-11-07
fix(core-flows): capture before order created

carlos-r-l-rodrigues:
What:
 When `autocapture` is enabled, the webhook is processed before the order was created.
 The payment processing workflows were merged into a single one
 
FIXES: SUP-118, SUP-9

https://github.com/medusajs/medusa/issues/9998


packages/core/core-flows/src/payment/workflows/process-payment.ts
```diff
@@ -15,6 +16,37 @@ export const processPaymentWorkflow = createWorkflow(
...
+          paymentSessionResult.data[0].payment_collection_id,
+      },
+    }).config({
+      name: "cart-payment-query",
+    })
+
+    when({ cartPaymentCollection }, ({ cartPaymentCollection }) => {
+      return !!cartPaymentCollection.data.length
+    }).then(() => {
+      completeCartWorkflow.runAsStep({
```
olivermrbl: The reason this logic was separated into two workflows was that the payment processing should not be affected by any subsequent actions we might want to perform as a side effect, here cart completion.

What happens if the cart completion fails now that is is run as a nested workflow? It will roll back the parent workflow too right? We might need to take care of this.

carlos-r-l-rodrigues: Good point, I was trying to figure out the comment in the code yesterday hehe. 
We can config this step to continue in case of permanent failure.

olivermrbl: Nice, that makes sense!

olivermrbl (approved):

carlos-r-l-rodrigues:
Do not merge it yet, as I will add one more change to this PR to have all cases covered even if there is no webhook configured.

fPolic (approved):
LGTM!

PR #9991, opened 2024-11-08
fix(dashboard): Reduce data loaded by product details page

kasperkristensen:
**What**
- Removes variants from the data requested by the product details page.
- Fixes an issue where the layout of the product page would break when there were **many** product options and option values.

Resolves CMRC-666


carlos-r-l-rodrigues (approved):
🔥

fPolic (approved):

PR #10005, opened 2024-11-10
chore: Add actor type check + more tests

olivermrbl:


packages/modules/providers/auth-emailpass/src/services/emailpass.ts
```diff
@@ -43,15 +43,15 @@ export class EmailPassAuthService extends AbstractAuthModuleProvider {
   }
 
   async update(
-    data: { email: string; password: string },
+    data: { password: string; entity_id: string },
```
olivermrbl, line 46: We currently only allow password updates, this could possibly be extended in the future.

carlos-r-l-rodrigues (approved):
LGTM

PR #10011, opened 2024-11-11
breaking: pluralize type helper to account for uncountable nouns and special rules

thetutlage:
Fixes: FRMW-2771

This PR updates the `Pluralize` type helper to have two conditional checks when converting a singular value to plural at types level.

- First, it checks for uncountable nouns in a dictionary.
- Second, it looks for special rules for certain words like `tooth`, `man` and so on.

This change will make us more compatible with the runtime behavior of pluralizing values. However, things won't be in 100% sync yet.


olivermrbl (approved):
Nice!

chuxi:
bad pr.

PR #10013, opened 2024-11-11
fix: pass color to github icon

christiananese:


sradevski (approved):

kasperkristensen:
@christiananese the icons package is autogenerated so you shouldn't make changes to the files as it will just be overwritten the next time I run the script to update icons. And all "brand" icons have fixed colors by design, you can ask Ludvig to add a GITHUB-EX icon to the Medusa UI Figma file, which will result in a icon that allows changing the color being generated.

PR #10021, opened 2024-11-11
docs: fix cart completion in subscriptions recipe

shahednasser:
Fix subscriptions recipe to retrieve the order's details after completion with `completeCartWorkflow` (since the workflow returns only the ID now).

### DIFF
--- a/www/apps/resources/app/recipes/subscriptions/examples/standard/page.mdx
+++ b/www/apps/resources/app/recipes/subscriptions/examples/standard/page.mdx
@@ -591,9 +591,10 @@ The compensation function receives the subscription as a parameter. It cancels t
 Create the file `src/workflows/create-subscription/index.ts` with the following content:
 
 export const createSubscriptionWorkflowHighlights = [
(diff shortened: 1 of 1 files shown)


olivermrbl (approved):

www/apps/resources/app/recipes/subscriptions/examples/standard/page.mdx
```diff
@@ -621,10 +623,22 @@ type WorkflowInput = {
...
   (input: WorkflowInput) => {
-    const order = completeCartWorkflow.runAsStep({
+    const { id } = completeCartWorkflow.runAsStep({
       input: {
-        id: input.cart_id,
+        id: input.cart_id
+      }
+    })
+
+    const order = useRemoteQueryStep({
```
olivermrbl, line 632: thought: Should we use `useQueryStep` so we slowly migrate to use the `query.graph` everywhere?

shahednasser: Oh I wasn't aware of it, we definitely should. Will do it in a separate PR 👍

PR #10024, opened 2024-11-11
feat(ui,dashboard): Add DataTable block

kasperkristensen:
**What**
- Adds opinionated DataTable block to `@medusajs/ui` 
- Adds new DataTable to `@medusajs/dashboard` that uses the above mentioned block as the primitive.

The PR also replaces the table on /customer-groups and the variants table on /products/:id with the new DataTable, to provide an example of it's usage. The previous DataTable component has been renamed to `_DataTable` and has been deprecated.

**Note**
This PR has a lot of LOC. 5,346 of these changes are the fr.json file, which wasn't formatted correctly before. When adding the new translations needed for this PR the file was formatted which caused each line to change to have the proper indentation.

Resolves CMRC-333


dwene:
This would be really nice as I'm building custom admin features! I love it!

Alexnortung:
Hi, I have tried out the datatable, I am not sure if it is intenden. But it seems like when the data for the datatable is empty and it is loading, the `<DataTable.Search>` input gets hidden.
This is a bit annoying if you are typing something and the input disappears.

A workaround is to use the previous data, such that the data is not empty, but is this the intended way to use it?

Alexnortung:
Hello @kasperkristensen,
I would like to be able to make a row a link, I can see that the other datatable has a `navigateTo` prop, which the new data table does not have. Could this be added to the new datatable? Or is there another way to make this work?

kasperkristensen:
> Hello @kasperkristensen, I would like to be able to make a row a link, I can see that the other datatable has a `navigateTo` prop, which the new data table does not have. Could this be added to the new datatable? Or is there another way to make this work?

You can pass an `onRowClick` callback to the hook, you can see how we use it in the dashboard project for navigation: https://github.com/medusajs/medusa/blob/ed003e974d50b38b90ea1565af37ec6dc26d692b/packages/admin/dashboard/src/components/data-table/data-table.tsx#L201

kasperkristensen:
> Hi, I have tried out the datatable, I am not sure if it is intenden. But it seems like when the data for the datatable is empty and it is loading, the `<DataTable.Search>` input gets hidden. This is a bit annoying if you are typing something and the input disappears.
> 
> A workaround is to use the previous data, such that the data is not empty, but is this the intended way to use it?

Hi @Alexnortung, it is intended according the design spec for the component, but I can see why it might be a bit annoying. I'll ping our designer if we should change it so the search bar is never hidden.

But generally I would always avoid having no data while loading, as it will cause layout shifts when the table goes from `data -> no data -> data`. The way we handle this internally is that we keep the old data until the new data is ready. If you are using `@tanstack/react-query` that can be done by using `keepPreviousData` for the `placeholderData`:

```
import { keepPreviousData } from "@tanstack/react-query"

const Component = () => {
  const data = useQuery({
    queryKey: [...],
    queryFn: () => {...},
    placeholderData: keepPreviousData
  })
}
```

Alexnortung:
Thanks for the feedback @kasperkristensen, it was helpful!
I have tried using the datatable a bit more and I have another note:
The toolbar seems to have some kind of z index where it will be on top of tooltips and modals. In the following screenshot you can notice that the first line in the tooltip is missing, but the tooltip looks right on the other rows.
billede

The following shows how it looks with a modal, where the table is behind it:
billede

kasperkristensen:
> Thanks for the feedback @kasperkristensen, it was helpful! I have tried using the datatable a bit more and I have another note: The toolbar seems to have some kind of z index where it will be on top of tooltips and modals. In the following screenshot you can notice that the first line in the tooltip is missing, but the tooltip looks right on the other rows. billede
> 
> The following shows how it looks with a modal, where the table is behind it: billede

Yes the header has a z-index of 1, but this should not cause any issues as the wrapping table has `isolation: isolate` which creates a new stacking context. Here is how it looks on my end:

image

Could you maybe share small reproduction of how you are using the component, and which browser you are using. Also it looks like your CSS is incorrect, if you haven't already make sure to also add a resolution for `@medusajs/ui-preset` with the snapshot version above, and see if that helps.

olivermrbl (approved):

PR #10031, opened 2024-11-11
fix(orchestration): fix set step failure

carlos-r-l-rodrigues:
What:
 - copy data before saving checkpoint
 - removed unused data format function
 - properly handle registerStepFailure to not throw
 - emit onFinish event even when execution failed


packages/core/orchestration/src/transaction/distributed-transaction.ts
```diff
@@ -215,9 +215,11 @@ class DistributedTransaction extends EventEmitter {
       this.modelId,
       this.transactionId
     )
-    await DistributedTransaction.keyValueStore.save(key, data, ttl, options)
 
-    return data
+    const rawData = JSON.parse(JSON.stringify(data))
+    await DistributedTransaction.keyValueStore.save(key, rawData, ttl, options)
```
sradevski, line 220: Was the keyVal store mutating the data?

packages/modules/workflow-engine-inmemory/integration-tests/__fixtures__/workflow_transaction_timeout.ts
```diff
@@ -3,11 +3,14 @@ import {
...
   StepResponse,
 } from "@medusajs/framework/workflows-sdk"
+import { setTimeout } from "timers/promises"
 
 const step_1 = createStep(
   "step_1",
-  jest.fn((input) => {
+  jest.fn(async (input) => {
     input.test = "test"
+    await setTimeout(200)
```
sradevski, line 12: I imagine it's not possible to use the jest fake timers here?

packages/modules/workflow-engine-inmemory/src/services/workflow-orchestrator.ts
```diff
@@ -260,7 +266,7 @@ export class WorkflowOrchestratorService {
     },
     @MedusaContext() sharedContext: Context = {}
   ) {
-    const {
+    let {
```
sradevski: suggestion: Maybe we keep most of the const except for throwOnError in order to avoid accidental mutation?

sradevski (approved):
LGTM

packages/modules/workflow-engine-inmemory/integration-tests/__fixtures__/workflow_transaction_timeout.ts
carlos-r-l-rodrigues: yeah, it is not straightforward :/

packages/core/orchestration/src/transaction/distributed-transaction.ts
carlos-r-l-rodrigues: for some reason it was accessing a #private property apparently inside mikro-orm when inserting it in the db.
here this is the correct behavior though, we first serialize it so the workflow engine only works with serializable data. After a lot of cleanup on the workflows, this was removed.

PR #10043, opened 2024-11-12
feat: convert MikroORM entities to DML entities

thetutlage:
Fixes FRMW-2790, FRMW-2777, FRMW-2798, FRMW-2800

## DML Changelog

- Fix: Mark foreign key as `null` when the relationship is marked a nullable.
- Fix: Do not set `isForeignKey: true` on the generate MikroORM property for the `belongsTo` relationship. However, the index in the database is still created. This is done to match the behaviour of relationships defined via MikroORM.
- Fix: `manyToMany` relationship to infer `owner` side of the relationship without relying on the order in which models are imported. Earlier, the first imported model was marked as the `owner`. **Could result in a breaking change especially in regards to `joinKey` and `inverseJoinKey` columns. But the same can be fixed by explicitly defining these properties.**
```ts
const isOwner =
    !!joinColumn ||
    !!inverseJoinColumn ||
    !!relationship.options.pivotTable ||
    /**
     * We can't infer it from the current entity so lets
     * look at the otherside configuration as well to make a choice
     */
    (!otherSideRelationOptions.pivotTable &&
      !otherSideRelationOptions.joinColumn &&
      !otherSideRelationOptions.inverseJoinColumn &&
      !MANY_TO_MANY_TRACKED_RELATIONS[
        `${relatedModelName}.${otherSideRelationshipProperty}`
      ])

// This means that a relation is infered as owner if there is one of the following options set on the relation
// - joinColumn
// - inverseJoinColumn
// - pivotTable
// if none are present, then we check if the otherside has none of those options as well and if it has not be loaded yet, 
// in that case the current entity (first to be loaded) can be owner. If the other side has any of those options then it will be
// inferred as owner when it will get loaded later
```
- Fix: Set `inversedBy: true` on the `owner` side of the `manyToMany` relationship. 
- fix: `manyToMany` pivot table name generation have been fixed. Meaning that it can be breaking changes in the scenario where no `pivotTable` were provided. In order to fix the breaking change if needed, the user can provide the `pivotTable` configuration with the existing name of the pivot table. (previously was using the models name in alpha order and pluralized last name, now it is using the table name from the model in alpha order with last one pluralized)
```ts
// If this create a breaking change for you then you can specify the `pivotTable` property configuration on the many to many with the current table name

// Something along those lines
const user = model.define('User', () => {
  addresses: model.manyToMany(() => Address, { pivotTable: 'address_users' })
})
```
- fix: `joinerConfigBuilder` now merges the auto-generated primary keys with the one provided in the custom configuration if any
- fix: `joinerConfigBuilder` now merges the auto-generated linkable with the one infered from the custom linkable keys provided if any
- improvement: Always auto generate index for `deleted_at` where is null
- fix: category rank ordering was broken ([see](https://github.com/medusajs/medusa/pull/10043#discussion_r1856460161))


packages/core/utils/src/dal/mikro-orm/mikro-orm-repository.ts
```diff
@@ -467,18 +498,22 @@ export function mikroOrmBaseRepositoryFactory<T extends object = object>(
...
 
-      return await manager.findAndCount(
-        entity as EntityName<T>,
-        findOptions_.where as MikroFilterQuery<T>,
-        findOptions_.options as MikroOptions<T>
-      )
+      return (await manager.findAndCount(
+        this.entity,
+        findOptions_.where,
+        findOptions_.options as any // MikroOptions<T>
```
thetutlage, line 504: Have to fix this

packages/core/utils/src/dml/entity.ts
```diff
@@ -234,4 +252,16 @@ export class DmlEntity<
...
   }
+
+  /**
+   * @experimental
+   * TODO: Write RFC about this, for now it is unstable and mainly
+   * for test purposes
+   * @param hooks
+   * @returns
+   */
+  hooks(hooks: Hooks<Schema, TConfig>): this {
```
thetutlage: Remove

packages/core/utils/src/dml/helpers/create-mikro-orm-entity.ts
```diff
@@ -84,22 +92,58 @@ export function createMikrORMEntity() {
...
+        const hookMethodName = "beforeCreate_" + camelToSnakeCase(modelName)
+        const hookWrapper = function (this: MikroORMEntity) {
+          return hook(this as any)
+        }
+
+        MikroORMEntity.prototype[hookMethodName] = hookWrapper
+        BeforeCreate()(MikroORMEntity.prototype, hookMethodName)
+      }
+    }
+
```
thetutlage: Remove

packages/core/utils/src/dal/mikro-orm/mikro-orm-repository.ts
adrien2p: we need the type utilities we talked about to fix that. for example populate can't get inferred anymore and therefore become `readonly never`

packages/core/utils/src/dml/__tests__/entity-builder.spec.ts
```diff
@@ -2969,11 +2972,11 @@ describe("Entity builder", () => {
             'CREATE UNIQUE INDEX IF NOT EXISTS "IDX_unique-name" ON "user" (organization, account, group_id) WHERE deleted_at IS NULL',
           name: "IDX_unique-name",
         },
-        {
-          expression:
-            'CREATE INDEX IF NOT EXISTS "IDX_user_group_id" ON "user" (group_id) WHERE deleted_at IS NULL',
-          name: "IDX_user_group_id",
-        },
+        // {
```
adrien2p: todo: in the DML either we dont bother and read it as foreign key, or we just create a default index for those props

packages/modules/product/integration-tests/__tests__/product-module-service/product-collections.spec.ts
```diff
@@ -31,27 +32,31 @@ moduleIntegrationTestRunner<IProductModuleService>({
       beforeEach(async () => {
         const testManager = await MikroOrmWrapper.forkManager()
 
-        productOne = testManager.create(Product, {
+        productOne = testManager.create(toMikroORMEntity(Product), {
```
adrien2p, line 35: discussion: here the return type is somewhat incorrect, we would just need to pass the Product.name as the first argument and the template would be `InferEntityType<typeof Product>` but we can discuss that for a separate PR. It goes back to the back and forth type discussion

packages/modules/product/src/migrations/Migration20241125090957.ts
```diff
@@ -0,0 +1,173 @@
+import { Migration } from "@mikro-orm/migrations"
+
+export class Migration20241125090957 extends Migration {
+  async up(): Promise<void> {
+    this.addSql(
+      'alter table if exists "product" drop constraint if exists "product_status_check";'
```
adrien2p, line 6: info: excpeded and get recreated later down

packages/modules/product/src/migrations/Migration20241125090957.ts
```diff
@@ -0,0 +1,173 @@
...
+    )
+    this.addSql(
+      'CREATE INDEX IF NOT EXISTS "IDX_product_category_path" ON "product_category" (mpath) WHERE deleted_at IS NULL;'
+    )
+    this.addSql(
+      'CREATE UNIQUE INDEX IF NOT EXISTS "IDX_category_handle_unique" ON "product_category" (handle) WHERE deleted_at IS NULL;'
+    )
+
+    this.addSql(
+      'CREATE UNIQUE INDEX IF NOT EXISTS "IDX_collection_handle_unique" ON "product_collection" (handle) WHERE deleted_at IS NULL;'
```
adrien2p, line 20: info: there was not part of the snapshot, now they are

packages/modules/product/src/migrations/Migration20241125090957.ts
```diff
@@ -0,0 +1,173 @@
...
+    )
+
+    this.addSql('drop index if exists "IDX_product_image_deleted_at";')
```
adrien2p, line 23: info: Was not following the naming convention, now they are

packages/modules/product/src/migrations/Migration20241125090957.ts
```diff
@@ -0,0 +1,173 @@
...
+      'CREATE INDEX IF NOT EXISTS "IDX_product_image_url" ON "image" (url) WHERE deleted_at IS NULL;'
+    )
+
+    this.addSql(
+      'CREATE UNIQUE INDEX IF NOT EXISTS "IDX_tag_value_unique" ON "product_tag" (value) WHERE deleted_at IS NULL;'
+    )
+
+    this.addSql(
+      'CREATE UNIQUE INDEX IF NOT EXISTS "IDX_type_value_unique" ON "product_type" (value) WHERE deleted_at IS NULL;'
+    )
```
adrien2p, line 37: info: was not part of the snapshot, now they are

packages/modules/product/src/migrations/Migration20241125090957.ts
```diff
@@ -0,0 +1,173 @@
...
+
+    this.addSql(
+      'alter table if exists "product" alter column "status" type text using ("status"::text);'
+    )
+    this.addSql(
+      "alter table if exists \"product\" add constraint \"product_status_check\" check (\"status\" in ('draft', 'proposed', 'published', 'rejected'));"
+    )
+    this.addSql(
+      'alter table if exists "product" alter column "status" set default \'draft\';'
```
adrien2p, line 46: info: re apply check and add a default as draft, was not present before but make sense to have

packages/modules/product/src/migrations/Migration20241125090957.ts
```diff
@@ -0,0 +1,173 @@
...
+    )
+    this.addSql(
+      'CREATE INDEX IF NOT EXISTS "IDX_product_type_id" ON "product" (type_id) WHERE deleted_at IS NULL;'
+    )
+    this.addSql(
+      'CREATE INDEX IF NOT EXISTS "IDX_product_collection_id" ON "product" (collection_id) WHERE deleted_at IS NULL;'
+    )
```
adrien2p, line 53: info: some foregin keys were missing their indices

packages/modules/product/src/migrations/Migration20241125090957.ts
```diff
@@ -0,0 +1,173 @@
...
+    this.addSql(
```
adrien2p, line 54: info: not part of the snapshot, now it is

packages/modules/product/src/migrations/Migration20241125090957.ts
```diff
@@ -0,0 +1,173 @@
...
+      'CREATE UNIQUE INDEX IF NOT EXISTS "IDX_product_handle_unique" ON "product" (handle) WHERE deleted_at IS NULL;'
+    )
+
+    this.addSql(
+      'alter table if exists "product_option" alter column "product_id" type text using ("product_id"::text);'
+    )
+    this.addSql(
+      'alter table if exists "product_option" alter column "product_id" set not null;'
+    )
```
adrien2p, line 63: info: was nullable, not it is not (foreign key)

packages/modules/product/src/migrations/Migration20241125090957.ts
```diff
@@ -0,0 +1,173 @@
...
+    this.addSql(
+      'CREATE UNIQUE INDEX IF NOT EXISTS "IDX_option_product_id_title_unique" ON "product_option" (product_id, title) WHERE deleted_at IS NULL;'
+    )
+
+    this.addSql(
+      'CREATE INDEX IF NOT EXISTS "IDX_product_option_value_option_id" ON "product_option_value" (option_id) WHERE deleted_at IS NULL;'
+    )
+    this.addSql(
+      'CREATE UNIQUE INDEX IF NOT EXISTS "IDX_option_value_option_id_unique" ON "product_option_value" (option_id, value) WHERE deleted_at IS NULL;'
+    )
```
adrien2p, line 76: info: was not part of the snapshot, now they are

packages/modules/product/src/migrations/Migration20241125090957.ts
```diff
@@ -0,0 +1,173 @@
...
+    this.addSql(
+      'alter table if exists "product_variant" alter column "length" type integer using ("length"::integer);'
+    )
+    this.addSql(
+      'alter table if exists "product_variant" alter column "height" type integer using ("height"::integer);'
+    )
+    this.addSql(
+      'alter table if exists "product_variant" alter column "width" type integer using ("width"::integer);'
+    )
```
adrien2p, line 89: info: they use to be `numeric` now they are `integer` should we think about it as it change the max value you can use?

packages/modules/product/src/migrations/Migration20241125090957.ts
```diff
@@ -0,0 +1,173 @@
...
+    this.addSql(
+      'CREATE UNIQUE INDEX IF NOT EXISTS "IDX_product_variant_barcode_unique" ON "product_variant" (barcode) WHERE deleted_at IS NULL;'
+    )
+    this.addSql(
+      'CREATE UNIQUE INDEX IF NOT EXISTS "IDX_product_variant_ean_unique" ON "product_variant" (ean) WHERE deleted_at IS NULL;'
+    )
+    this.addSql(
+      'CREATE UNIQUE INDEX IF NOT EXISTS "IDX_product_variant_upc_unique" ON "product_variant" (upc) WHERE deleted_at IS NULL;'
+    )
```
adrien2p, line 104: info: was not part of the snapshot, now they are

packages/modules/product/integration-tests/__fixtures__/product/data/create-product.ts
```diff
@@ -52,7 +52,7 @@ export const buildProductAndRelationsData = ({
   options,
   variants,
   collection_id,
-}: Partial<ProductTypes.CreateProductDTO>) => {
+}: Partial<ProductTypes.CreateProductDTO> & { tags: { value: string }[] }) => {
```
thetutlage, line 55: Have to double check how it was working earlier

packages/modules/product/src/repositories/product-category.ts
```diff
@@ -20,12 +20,15 @@ export class ProductCategoryRepository extends DALUtils.MikroOrmBaseTreeReposito
     familyOptions: ProductCategoryTransformOptions = {}
   ) {
     const findOptions_ = { ...findOptions }
-    findOptions_.options ??= {
-      orderBy: { rank: "ASC" },
+    findOptions_.options.orderBy = {
```
adrien2p, line 24: note: in the tests, the fields where provided and hence the options always populated, meaning that the order was never applied. By some random luck from god, the order remain the same for a long time, but the truth is that without any ordering we cant guarantee the order of the retrieved data. Now we always apply both rank and id, because rank only would also randomize all categories with the same rand and hence not enforce any ordering among each rank value. Now all categories are always ordered and the order remain constant.

further more, the rank was never selected, meaning that when we include ancerstors or descendant, the sortCategoriesByRank cant work since it is based on the property rank which was never deleted. I am really unsure how it worked so far except luck :)

cc @olivermrbl maybe you have an idea?

packages/core/utils/src/dml/helpers/create-mikro-orm-entity.ts
```diff
@@ -94,12 +100,31 @@ export function createMikrORMEntity() {
...
+  createEntity.clear = function () {
+    MANY_TO_MANY_TRACKED_RELATIONS = {}
+    ENTITIES = {}
+  }
+  return createEntity
 }
 
+/**
+ * Helper function to convert DML entities to MikroORM entity. Use
+ * "toMikroORMEntity" if you are ensure the input is a DML entity
```
olivermrbl, line 123:
```suggestion
 * "toMikroORMEntity" if you are unsure whether the input is a DML entity
```

olivermrbl:
> Fix: manyToMany relationship to infer owner side of the relationship without relying on the order in which models are imported

Maybe this will be clear from the code, but can you describe how we are inferring this now?

packages/core/utils/src/modules-sdk/module.ts
```diff
@@ -49,15 +49,22 @@ export function Module<
       DmlEntity.isDmlEntity(model)
     )
 
+    // TODO: Custom joiner config should take precedence over the DML auto generated linkable
+    // Thats in the case of manually providing models in custom joiner config.
+    // TODO: Add support for non linkable modifier DML object to be skipped from the linkable generation
```
olivermrbl, line 54: question: Judging from the PR description, it sounds like we are doing this now? Can we remove these todos?

packages/modules/product/src/migrations/Migration20241125090957.ts
```diff
@@ -0,0 +1,173 @@
+import { Migration } from "@mikro-orm/migrations"
+
+export class Migration20241125090957 extends Migration {
```
olivermrbl, line 3: 👀  lots of clean up, it seems!

adrien2p:
> > Fix: manyToMany relationship to infer owner side of the relationship without relying on the order in which models are imported
> 
> Maybe this will be clear from the code, but can you describe how we are inferring this now?

sure, do you want it in the description or in a comment?

nvm: I ve updated the description

packages/modules/product/src/migrations/Migration20241125090957.ts
olivermrbl: We talked about using big number instead, wouldn't that be more appropriate?

olivermrbl (commented):
Really nice work guys, love how simple the data model definitions are with the DML 💪

packages/core/utils/src/modules-sdk/module.ts
adrien2p: we are not doing the last todo, you currently can't specify an entity as not linkable, for example in the case of entity that only serves a purpose within the module. Same thing will happen with the order module, where the graphql is generated from the DML but the interface of the objects that are returned to the user does not necessarily correspond to the internal entities, therefore we have to override there too

packages/modules/product/src/migrations/Migration20241125090957.ts
adrien2p: yes indeed 😂

packages/modules/product/src/migrations/Migration20241125090957.ts
adrien2p: oh sorry, I thought we ended up saying we don't do anything :/

thetutlage: Yes, we decided to not do anything because `BigNumber` will create a breaking change at the HTTP layer. Correct me if I am wrong 🙂

olivermrbl: Yeah, that's right. Don't think we should introduce this as part of this refactor. Let's keep it as is.

olivermrbl (approved):
Overall this LGTM, nice work

Can we add more descriptive guides for how you can avoid the breaking changes by setting the configurations explicitly? Would like to include these in the upgrade guide.

packages/core/utils/src/modules-sdk/module.ts
olivermrbl: Ah my bad, I misunderstood the comment. But let's remove the first todo then :)

adrien2p:
> Overall this LGTM, nice work
> 
> Can we add more descriptive guides for how you can avoid the breaking changes by setting the configurations explicitly? Would like to include these in the upgrade guide.

I ve added a mention and an example

adrien2p:
@olivermrbl lets wait for the pipeline but I believe we should be good now

Issue #10062, opened 2024-11-12
Medusa Development Resources: Issue in Select Product in Storefront


Lzok:
### What Medusa version and documentation are you using?

v2

### Preliminary Checks

- [X] This issue is not a duplicate. Before opening a new issue, please search existing issues: https://github.com/medusajs/medusa/issues

### Issue Summary

I think there is an issue at least in the Store->Product section in the [JS SDK Docs](https://docs.medusajs.com/resources/references/js-sdk/admin/product)

As you can see in the image below, I selected `Store->Product` on the left Menu, but it pointed me to the `/admin/product` page, and it is talking about the `sdk.admin.product` method there, where I think it should talk about `sdk.store.product`

![image](https://github.com/user-attachments/assets/9da5d1f7-f3f7-44f0-80df-42aa29267a5e)

### How can this issue be resolved?

1. Load the correct Store Product page instead of the Admin Product one.

### Are you interested in working on this issue?

- [x] I would like to fix this issue

shahednasser:
Nice catch! It seems like the sidebar is pointing to the admin pages rather than storefront pages. Will open a PR to fix it

PR #10063, opened 2024-11-12
docs: fix slugs pointing to admin rather than store routes

shahednasser:
Fix slugs of JS SDK store properties having `/admin` rather than `/store`

Closes DX-1063


Lzok:
Thank you! 🚀 

PS: This PR solves https://github.com/medusajs/medusa/issues/10062

PR #10071, opened 2024-11-13
test: add tests for self relationships

thetutlage:
Initially I thought that we will have to add special checks to allow relationships referencing itself. However, it turned out that not to be the case. Instead, it had more to do with how self relationships are defined in Mikro ORM.

In case of `belongsTo` relationship we have to define the other side as well either as a `hasMany` or `hasOne`. For example:

**❌ The following code will fail, because no children are defined for the parent**
```ts
  id: model.number(),
  username: model.text(),
  parent: model.belongsTo(() => user)
})
```

**✅ Addition of children relationship will make things work**
```ts
  id: model.number(),
  username: model.text(),
  parent: model.belongsTo(() => user, { mappedBy: "children" }),
  children: model.hasMany(() => user, { mappedBy: "parent" }),
})
```

We can see the similar setup here with our `ProductCategory` MikroORM entity. https://github.com/medusajs/medusa/blob/develop/packages/modules/product/src/models/product-category.ts#L87-L94

@adrien2p Correct me if I am wrong. But I have added the tests for now so that we know the behavior of self relationships


packages/core/utils/src/dml/__tests__/entity-builder.spec.ts
```diff
@@ -4702,6 +4702,226 @@ describe("Entity builder", () => {
...
+
+      expectTypeOf(new User()).toMatchTypeOf<{
+        id: number
+        username: string
+        deleted_at: Date | null
+        parent: {
+          id: number
+          username: string
+          deleted_at: Date | null
+        }
```
adrien2p, line 4724: suggestion: We should probably add the children type check here

packages/core/utils/src/dml/__tests__/entity-builder.spec.ts
```diff
@@ -4702,6 +4702,226 @@ describe("Entity builder", () => {
         },
       })
     })
+
+    test("define relationship with self via hasMany", () => {
```
adrien2p:
```suggestion
    test("define entity with relationship to itself via hasMany", () => {
```

packages/core/utils/src/dml/__tests__/entity-builder.spec.ts
```diff
@@ -4702,6 +4702,226 @@ describe("Entity builder", () => {
...
+          name: "deleted_at",
+          fieldName: "deleted_at",
+          nullable: true,
+          getter: false,
+          setter: false,
+        },
+      })
+    })
+
+    test("define relationship with self via hasOne", () => {
```
adrien2p, line 4820: suggestion: if accepted we can rename this test as well

packages/core/utils/src/dml/__tests__/entity-builder.spec.ts
```diff
@@ -4702,6 +4702,226 @@ describe("Entity builder", () => {
...
+
+      expectTypeOf(new User()).toMatchTypeOf<{
+        id: number
+        username: string
+        deleted_at: Date | null
+        parent: {
+          id: number
+          username: string
+          deleted_at: Date | null
+        }
```
adrien2p, line 4838: suggestion: same as previous one

adrien2p (approved):
LGTM, few suggestion

PR #10078, opened 2024-11-13
fix(dashboard): order edit - display item quantity change correctly

fPolic:
**What**
- use a diff form change action details to display edit history


carlos-r-l-rodrigues (approved):

carlos-r-l-rodrigues:
On thing to keep in mind is that if there are multiple actions to update the same item (not possible using the UI I believe), we should only consider the last update as the final one.

PR #10079, opened 2024-11-13
fix(dashboard): Add Breadcrumb components

kasperkristensen:
**What**
- Adds Breadcrumb component to all routes that needs breadcrumbs.
- The Breadcrumb components use a combination of loader data and useQuery to ensure that the displayed value is kept up to date if the underlying data is changed via a mutation.
- Also fixes a couple of places where the breadcrumb was not setup correctly.

Resolves CMRC-688


fPolic (approved):
LGTM 💪

PR #10100, opened 2024-11-14
fix(core-flows): Fix date usage accross workflows

adrien2p:
FIXES CMRC-691

**What**
`Date` is something that get executed, since workflows are meant to compose the definition of what will be executed, the date where always having the same value as they was executed once during composition.
Instead wrap those into transformer that will be executed when needed and fix the Date issues


adrien2p:
cc @frane @carlos-r-l-rodrigues

fPolic (approved):
LGTM!

carlos-r-l-rodrigues (approved):

PR #10106, opened 2024-11-14
feat(medusa, types, utils, core-flows, order) request & accept order transfer

fPolic:
**What**
- add request order transfer workflow
- add admin endpoint for transferring an order to a customer
- accept order transfer storefront endpoint
- accept transfer workflow
- changes in the order module to introduce new change and action types

---

**Note**
- we return 400 instead 409 currently if there is already an active order edit, I will revisit this in a followup
- endpoint for requesting order transfer from the storefront will be added in a separate PR

---

RESOLVES CMRC-701
RESOLVES CMRC-703
RESOLVES CMRC-704
RESOLVES CMRC-705


carlos-r-l-rodrigues (approved):

packages/modules/order/src/utils/actions/transfer-customer.ts
```diff
@@ -0,0 +1,19 @@
+import { ChangeActionType, MedusaError } from "@medusajs/framework/utils"
+import { OrderChangeProcessing } from "../calculate-order-change"
+import { setActionReference } from "../set-action-reference"
+
+OrderChangeProcessing.registerActionType(ChangeActionType.TRANSFER_CUSTOMER, {
```
olivermrbl, line 5: @carlos-r-l-rodrigues, really starting to see the power of this actions API! 💪

packages/medusa/src/api/store/orders/[id]/transfer/accept/route.ts
```diff
@@ -0,0 +1,30 @@
...
+) => {
+  await acceptOrderTransferWorkflow(req.scope).run({
+    input: {
+      orderId: req.params.id,
+      token: req.validatedBody.token,
+    },
+  })
+
+  const workflow = getOrderDetailWorkflow(req.scope)
+  const { result } = await workflow.run({
```
Comment (olivermrbl) [inline packages/medusa/src/api/store/orders/[id]/transfer/accept/route.ts]: nit:
```suggestion
  const { result } = await getOrderDetailWorkflow(req.scope).run({
```

adrien2p (approved):
Very nice 😊

packages/core/types/src/workflow/order/request-transfer.ts
```diff
@@ -0,0 +1,8 @@
+export interface RequestOrderTransferWorkflowInput {
```
olivermrbl, line 1: question: Don't we use a snake_case convention?

packages/modules/order/src/utils/actions/transfer-customer.ts
adrien2p: It really feels easy yo add different type of action while keeping control over it 😇

packages/core/core-flows/src/order/workflows/transfer/request-order-transfer.ts
```diff
@@ -0,0 +1,136 @@
...
+        `Cannot transfer order: ${order.id} to a guest customer account: ${customer.email}`
+      )
+    }
+
+    if (order.customer_id === customer.id) {
+      throw new MedusaError(
+        MedusaError.Types.INVALID_DATA,
+        `Order: ${order.id} already belongs to customer: ${customer.id}`
+      )
+    }
```
olivermrbl, line 50: suggestion/nit(non-blocking): Return idempotently?

olivermrbl (approved):
Overall LGTM

packages/modules/order/src/utils/actions/transfer-customer.ts
olivermrbl: Yes agree!

packages/core/core-flows/src/order/workflows/transfer/request-order-transfer.ts
fPolic: This flow should result in an order change being created so if we don't create one, because initiating transfer in this case doesn't make sense, I would say throwing here is expected.

PR #10111, opened 2024-11-15
docs: fix fulfillment module container name

mimukit:
Here is a small fix of the fulfillment container import name to `Modules.FULFILLMENT` instead of `Modules.API_KEY`.


shahednasser (approved):
Nice catch thanks!

PR #10132, opened 2024-11-17
feat: node_modules plugins resolve

stephane-segning:
## What
Fixed the plugin load when installed 

## Why
Plugins won't load normally if the plugin is in the node_modules

## How
Added a resolve for node_modules
```typescript
resolvedPath = path.resolve(`${process.cwd()}/node_modules/${pluginName}`)
const doesExistsInNodeModules = existsSync(resolvedPath)
if (doesExistsInNodeModules) {
  const packageJSON = JSON.parse(
    fs.readFileSync(`${resolvedPath}/package.json`, `utf-8`)
  )

   const computedResolvedPath = path.join(resolvedPath, "dist")

  return {
    resolve: computedResolvedPath,
    id: createPluginId(packageJSON.name),
    name: packageJSON.name,
    options: {},
    version: packageJSON.version,
  }
}
```

## Testing
No tests written for this feature

### DIFF
--- a/packages/medusa/src/loaders/helpers/resolve-plugins.ts
+++ b/packages/medusa/src/loaders/helpers/resolve-plugins.ts
@@ -93,8 +93,6 @@ function resolvePlugin(pluginName: string): {
         includeExtensionDirectoryPath: true,
       })
     }
-
-    throw new Error(`Unable to find the plugin "${pluginName}".`)
   }
 
   try {
--- a/packages/medusa/src/loaders/helpers/resolve-plugins.ts
+++ b/packages/medusa/src/loaders/helpers/resolve-plugins.ts
@@ -93,6 +93,24 @@ function resolvePlugin(pluginName: string): {
         includeExtensionDirectoryPath: true,
       })
     }
+
+    resolvedPath = path.resolve(`${process.cwd()}/node_modules/${pluginName}`)
+    const doesExistsInNodeModules = existsSync(resolvedPath)
+    if (doesExistsInNodeModules) {
+      const packageJSON = JSON.parse(
+        fs.readFileSync(`${resolvedPath}/package.json`, `utf-8`)
+      )
+
+      const computedResolvedPath = path.join(resolvedPath, "dist")
+
+      return {
+        resolve: computedResolvedPath,
+        id: createPluginId(packageJSON.name),
+        name: packageJSON.name,
+        options: {},
+        version: packageJSON.version,
+      }
+    }
   }
 
   try {


thetutlage:
@stephane-segning 

Thanks for taking out time to create this PR. However, we won't be able to move forward with it for now because of the following reasons.

- Hardcoding the path to `node_modules` could fail in many scenarios, since the Node.js resolution for packages is a little more complex. https://medium.com/outbrain-engineering/node-js-module-resolution-af46715784ef#:~:text=Node%20Modules&text=js%20will%20try%20to%20resolve%20the%20import%20from%20the%20node_modules%20directory.&text=In%20general%2C%20Node.,available%20to%20all%20the%20Node.
- Also, plugins are something we have to rethink in v2 and we will not rush them for now.

stephane-segning:
Hello @thetutlage. Thank you for your prompt reply. But do you have a workaround mechanism for now? 

I've been thinking about a way to use absolute path using process.cwd() in the medusa-config.ts file, so as to resolve the path. But it still doesn't work, because of the resolution mechanism.

adrien2p:
Hi @stephane-segning 
In your medusa config as a workaround you can provide the following 

```ts
resolve: require.resolve("plugin_name")

```
The above should work could you give this a shot please

stephane-segning:
Hello @adrien2p ! Sorry for the long wait. I took my time to test multiple variations of this same workaround and none worked.

hrithikvishwakarma001:
@stephane-segning Did you discover any solutions for using plugins? and @adrien2p when are you planning to add plugins in medusa 2.0 any rough ETA?

nebiyuelias1:
Yeah having similar issues here!

stephane-segning:
Hi guys! @nebiyuelias1, @hrithikvishwakarma001 I cheated by using system links. It's working fine (for now). Documented here https://github.com/vymalo/medusajs/tree/main/packages/vymalo-printful#hack. 

The idea is to use `ln` to create a link to said npm folder:
```shell
mkdir -p plugins/printful-hack
ln -s node_modules/@vymalo/medusa-printful/dist plugins/printful-hack/src
# Write a fake package.json file in the plugins/printful-hack directory
echo "{\"name\":\"printful-hack\",\"version\":\"1.0.0\"}" > plugins/printful-hack/package.json
```

and update the medusajs config file
```typescript
plugins: [
  ...
  {
    resolve: `./plugins/printful-hack`, // TODO: Change this to the correct path
    options: {},
  },
```

It's working fine, even though this looks sketchy. I'll be waiting to see how the plugins evolves in this v2 of medusajs and update accordingly

PR #10134, opened 2024-11-18
refactor: rename workflow to singular

thetutlage:
Deprecated the `createOrdersWorkflow` and exported a new workflow with singular name.

The deprecated method will appear as follows.

CleanShot 2024-11-18 at 14 33 33@2x

Changeset is in - https://github.com/medusajs/medusa/pull/10136


carlos-r-l-rodrigues (approved):
Should we update this route to use it? `packages/medusa/src/api/admin/draft-orders/route.ts`

olivermrbl (approved):

PR #10135, opened 2024-11-18
fix(payment): Idempotent cancellation and proper creationg fail handling

adrien2p:
RESOLVES SUP-188

**What**
Two changes are happening here
- In the stripe payment provider, idempotent cancellation action, if not id is provided then return the existing data unchanged
- Payment module should not try to cancel a session that have failed to be created in the first place


carlos-r-l-rodrigues (approved):

olivermrbl (approved):
LGTM

PR #10138, opened 2024-11-18
chore(): Update module provider retrieval error message and type

adrien2p:
Partially RESOLVES FRMW-2802

**What**
Improve error message and change the error type when retrieving a provider from a local container fail


packages/modules/auth/src/services/auth-provider.ts
```diff
@@ -26,10 +25,11 @@ export default class AuthProviderService {
     try {
       return this.dependencies[`${AuthProviderRegistrationPrefix}${providerId}`]
     } catch (err) {
-      throw new MedusaError(
-        MedusaError.Types.NOT_FOUND,
-        `Could not find a auth provider with id: ${providerId}`
-      )
+      const errMessage = `
+      Unable to retreieve the auth provider with id: ${providerId}
```
carlos-r-l-rodrigues:
```suggestion
      Unable to retrieve the auth provider with id: ${providerId}
```

packages/modules/fulfillment/src/services/fulfillment-provider.ts
```diff
@@ -48,10 +48,11 @@ export default class FulfillmentProviderService extends ModulesSdkUtils.MedusaIn
     try {
       return this.__container__[`fp_${providerId}`]
     } catch (err) {
-      throw new MedusaError(
-        MedusaError.Types.NOT_FOUND,
-        `Could not find a fulfillment provider with id: ${providerId}`
-      )
+      const errMessage = `
+      Unable to retreieve the fulfillment provider with id: ${providerId}
```
carlos-r-l-rodrigues:
```suggestion
      Unable to retrieve the fulfillment provider with id: ${providerId}
```

packages/modules/locking/src/services/locking-provider.ts
```diff
@@ -31,10 +31,11 @@ export default class LockingProviderService {
         `${LockingProviderRegistrationPrefix}${providerId}`
       ]
     } catch (err) {
-      throw new MedusaError(
-        MedusaError.Types.NOT_FOUND,
-        `Could not find a locking provider with id: ${providerId}`
-      )
+      const errMessage = `
+      Unable to retreieve the locking provider with id: ${providerId}
```
carlos-r-l-rodrigues:
```suggestion
      Unable to retrieve the locking provider with id: ${providerId}
```

packages/modules/notification/src/services/notification-provider.ts
```diff
@@ -45,10 +45,11 @@ export default class NotificationProviderService extends ModulesSdkUtils.MedusaI
         `${NotificationProviderRegistrationPrefix}${providerId}`
       ]
     } catch (err) {
-      throw new MedusaError(
-        MedusaError.Types.NOT_FOUND,
-        `Could not find a notification provider with id: ${providerId}`
-      )
+      const errMessage = `
+      Unable to retreieve the notification provider with id: ${providerId}
```
carlos-r-l-rodrigues:
```suggestion
      Unable to retrieve the notification provider with id: ${providerId}
```

packages/modules/payment/src/services/payment-provider.ts
```diff
@@ -32,10 +32,11 @@ export default class PaymentProviderService extends ModulesSdkUtils.MedusaIntern
     try {
       return this.__container__[providerId] as IPaymentProvider
     } catch (e) {
-      throw new MedusaError(
-        MedusaError.Types.NOT_FOUND,
-        `Could not find a payment provider with id: ${providerId}`
-      )
+      const errMessage = `
+      Unable to retreieve the payment provider with id: ${providerId}
```
carlos-r-l-rodrigues:
```suggestion
      Unable to retrieve the payment provider with id: ${providerId}
```

carlos-r-l-rodrigues (approved):

packages/modules/auth/integration-tests/__tests__/auth-module-service/index.spec.ts
```diff
@@ -84,7 +84,8 @@ moduleIntegrationTestRunner({
 
         expect(err).toEqual({
           success: false,
-          error: "Could not find a auth provider with id: facebook",
+          error:
+            "\n      Unable to retreieve the auth provider with id: facebook\n      Please make sure that the provider is registered in the container and it is configured correctly in your project configuration file.\n      ",
```
olivermrbl:
```suggestion
            "\n      Unable to retrieve the auth provider with id: facebook\n      Please make sure that the provider is registered in the container and it is configured correctly in your project configuration file.\n      ",
```

packages/modules/auth/src/services/auth-provider.ts
```diff
@@ -26,10 +25,11 @@ export default class AuthProviderService {
...
+      Please make sure that the provider is registered in the container and it is configured correctly in your project configuration file.
+      `
+      throw new Error(errMessage)
```
olivermrbl, line 39: question: Just wanna make sure the change from `MedusaError` to `Error` intentional?

packages/modules/fulfillment/src/services/fulfillment-provider.ts
```diff
@@ -48,10 +48,11 @@ export default class FulfillmentProviderService extends ModulesSdkUtils.MedusaIn
...
+      Please make sure that the provider is registered in the container and it is configured correctly in your project configuration file.
+      `
+      throw new Error(errMessage)
```
olivermrbl, line 62: Same here.

packages/modules/auth/src/services/auth-provider.ts
olivermrbl: Nvm, I see its added throughout the PR, so I guess it is 😄

olivermrbl (approved):
Should we log the error message?

packages/modules/auth/src/services/auth-provider.ts
adrien2p: yes it is :), I dont think it has to be a medusa error here but if you think otherwise I would be happy to change it

adrien2p:
> Should we log the error message?

Is there a reason why we would want to log it and throw it? the throw should appear in the log no?

packages/modules/auth/src/services/auth-provider.ts
olivermrbl: Fine to keep it like this!

olivermrbl:
> Is there a reason why we would want to log it and throw it? the throw should appear in the log no?

@adrien2p, maybe I am missing something, but we are, in fact, not surfacing the actual error message, right? Only the description of what might have gone wrong (which I still think we should do).

```
} catch (err) {
  logger.debug(err) // <---- wouldn't this make it easier for developers to dig out the issue?
  const errMessage = "..."
  throw Error(errMessage)
}
```

For example, this would show if you are trying to consume the event bus service without having specified it in `dependencies`.

adrien2p:
> > Is there a reason why we would want to log it and throw it? the throw should appear in the log no?
> 
> 
> 
> @adrien2p, maybe I am missing something, but we are, in fact, not surfacing the actual error message, right? Only the description of what might have gone wrong (which I still think we should do).
> 
> 
> 
> ```
> 
> } catch (err) {
> 
>   logger.debug(err) // <---- wouldn't this make it easier for developers to dig out the issue?
> 
>   const errMessage = "..."
> 
>   throw Error(errMessage)
> 
> }
> 
> ```
> 
> 
> 
> For example, this would show if you are trying to consume the event bus service without having specified it in `dependencies`.

I could add it to the cause in fact? Cause you will have to stack trace and two errors if we both logs and throw. What i can do is log it for now and we can revisit our log/cause strategy in the appropriate project 🫡

PR #10144, opened 2024-11-18
fix(product): Update full descendant tree mpath when updating parent category id

adrien2p:
FIXES FRMW-2774

**What**
When updating the parent category id, all descendant mpath should be re computed


packages/modules/product/integration-tests/__tests__/product-category.ts
```diff
@@ -1129,6 +1129,109 @@ moduleIntegrationTestRunner<Service>({
             ])
           )
         })
+
+        it(`should update the mpath of the fulle descenden tree successfully when moving the grand parent in the hierarchy`, async () => {
```
carlos-r-l-rodrigues:
```suggestion
        it(`should update the mpath of the full descendent tree successfully when moving the grand parent in the hierarchy`, async () => {
```

carlos-r-l-rodrigues (approved):

PR #10150, opened 2024-11-19
feat: add optional fields

thetutlage:
Optional fields are precursor to the `default` value. Since it doesn't make sense to assign a default value, unless the field is marked as optional.

Fixes: FRMW-2786


thetutlage:
CC @shahednasser for documentation.

You can now make fields optional using DML using the `optional` modifier.

```ts
model.define('Product', {
  title: model.string(),
  handle: model.string().optional()
})
```

## How is it different from `nullable`?
The `nullable` fields are marked as `nullable` within the database and they should be defined as `null` when no value is provided. Whereas, `optional` fields can be `undefined/missing` altogether.

```ts
// With nullable
const product = {
  handle: null
}

// With optional
const product = {
  handle: undefined
}
```

carlos-r-l-rodrigues (approved):

adrien2p (approved):

PR #10151, opened 2024-11-19
feat(core-flows,medusa): Add API to update cart's customer

riqwan:
what:

- adds an endpoint that updates a cart's customer

RESOLVES CMRC-718


olivermrbl (approved):
LGTM

PR #10158, opened 2024-11-19
feat(types, js-sdk): add updateCustomer to js-sdk

riqwan:
what:

- adds new `POST /store/cart/:id/customer` endpoint to js sdk


packages/core/js-sdk/src/store/index.ts
```diff
@@ -941,6 +941,38 @@ export class Store {
...
+     * @param headers - Headers to pass in the request.
+     * @returns The updated cart.
+     *
+     * @example
+     * sdk.store.cart.updateCustomer("cart_123", {})
+     * .then(({ cart }) => {
+     *   console.log(cart)
+     * })
+     */
+    updateCustomer: async (
```
olivermrbl: nit/suggestion(non-blocking): Wdyt of `transferCustomer`

packages/core/js-sdk/src/store/index.ts
```diff
@@ -941,6 +941,38 @@ export class Store {
...
+      id: string,
+      body: HttpTypes.StoreUpdateCartCustomer,
```
olivermrbl: thought: Can we remove the `body` param and always pass `{}` in the `fetch` call?

That way you could consume it like so:
```ts
sdk.store.cart.transferCustomer("cart_123")
```

olivermrbl (approved):
LGTM, two suggestions :)

packages/core/js-sdk/src/store/index.ts
riqwan: wouldn't it be the other way around? `transferCart('cart_id')`?

olivermrbl: Yes, my bad. Alternatively `setCustomer`. It's super nitpicky so I'll let you decide, but I just feel like `updateCustomer` intuitively tells me that I can update the customer information (not only the ID) on the cart 😄

packages/core/js-sdk/src/store/index.ts
riqwan: done!

PR #10170, opened 2024-11-19
fix: Remove undefined from events in update

olivermrbl:
**What**

Let's imagine we perform a selector update on a resource that doesn't exist:
```
await updateCountries({ region_id: "reg_123" }, { : metadata: { ... } })
```
(`reg_123` doesn't exist). 

The underlying Medusa Service `update` call will return an empty result set. Because we assign the first entry in the result set to `response,` the response will be `[ undefined ]`.

Then, `response` is passed to the event factory builder that builds events using `response.map(r => r.id)` which is the piece that throws and leads to the following error: 
```
message: "Cannot read properties of undefined (reading 'id')",
stack: "TypeError: Cannot read properties of undefined (reading 'id')\n at /[redacted]/node_modules/@medusajs/utils/src/modules-sdk/event-builder-factory.ts:69:30\n"
```

@adrien2p @carlos-r-l-rodrigues let me know if this change is better suited elsewhere.


packages/modules/region/integration-tests/__tests__/region-module.spec.ts
```diff
@@ -403,6 +403,16 @@ moduleIntegrationTestRunner<IRegionModuleService>({
         expect(resp.countries).toHaveLength(2)
       })
 
+      it.only("should delete a region without countries", async () => {
```
adrien2p: Todo: rm .only

adrien2p (approved):
Nice catch 🫡

packages/modules/region/integration-tests/__tests__/region-module.spec.ts
```diff
@@ -403,6 +403,16 @@ moduleIntegrationTestRunner<IRegionModuleService>({
...
       })
 
+      it("should delete a region without countries", async () => {
+        const createdRegion = await service.createRegions({
+          name: "North America",
+          currency_code: "USD",
+          countries: [],
+        })
+
+        await service.softDeleteRegions([createdRegion.id])
```
adrien2p, line 413: Q: should we add an expectation ? You can retrieve with deleted and check that it is properly set

olivermrbl: Yeah, just thought of it, let me do that 😄

carlos-r-l-rodrigues (approved):

PR #10187, opened 2024-11-20
feat(product,dashboard): Allow re-ordering images

### LINKED ISSUE #10128: [Bug]: Admin product media view image in Firefox is zero-width
### Package.json file

```JSON
{
  "name": "medusa-starter-default",
  "version": "0.0.1",
  "license": "MIT",
  "keywords": [
    "sqlite",
    "postgres",
    "typescript",
    "ecommerce",
    "headless",
    "medusa"
  ],
  "scripts": {
    "build": "medusa build",
    "start": "medusa start",
    "dev": "medusa develop",
  },
  "dependencies": {
    "@medusajs/admin-sdk": "latest",
    "@medusajs/cli": "latest",
    "@medusajs/framework": "latest",
    "@medusajs/medusa": "latest",
    "@mikro-orm/core": "5.9.7",
    "@mikro-orm/knex": "5.9.7",
    "@mikro-orm/migrations": "5.9.7",
    "@mikro-orm/postgresql": "5.9.7",
    "awilix": "^8.0.1",
    "pg": "^8.13.0"
  },
  "devDependencies": {
    "@medusajs/test-utils": "latest",
    "@mikro-orm/cli": "5.9.7",
    "@swc/core": "1.5.7",
    "@swc/jest": "^0.2.36",
    "@types/jest": "^29.5.13",
    "@types/node": "^20.0.0",
    "@types/react": "^18.3.2",
    "jest": "^29.7.0",
    "prop-types": "^15.8.1",
    "react": "^18.2.0",
    "react-dom": "^18.2.0",
    "ts-node": "^10.9.2",
    "typescript": "^5.6.2",
    "vite": "^5.2.11"
  },
  "engines": {
    "node": ">=20"
  },
  "packageManager": "yarn@1.22.22+sha512.a6b2f7906b721bba3d67d4aff083df04dad64c399707841b7acf00f6b133b7ac24255f2652fa22ae3534329dc6180534e98d17432037ff6fd140556e2bb3137e"
}
```

### Node.js version

v22.11.0

### Database and its version

PostgreSQL 17.1

Fedora Linux 41

### Browser name

Firefox

### What happended?

When I go to Products, choose a product, and click one of the media items, in Firefox it ends up completely collapsed.

### Expected behavior

I expect it to work like Chrome/Chromium, and show the image.

### Actual behavior

In Firefox, the image doesn't show.  I'm not sure exactly why, but I suspect it's because the image is under a div with `width: fit-content`, but the image has `width: 100%`, making the parent and child sizes each depend on one-another.  If I change the div width to be 100%, it works, and if I change the img tag's width to be a pixel size, it also works.

This is a side-by-side comparison of Firefox and Chromium on the same view:

2024-11-16T22-50-35

### Link to reproduction repo

N/A (this works in a vanilla `npx create-medusa-app@latest` project)

kasperkristensen:
**What**
- Migrates Product <> ProductImage to a OneToMany relationship. The migration script ensures that the current data in the `product-images` pivot table is migrated.
- Adds `ProductImage.rank` field, the field is treated as readonly, and does not allow the user to set it directly. It is instead inferred by the index of an image in the product payload, e.g. `images: [{ url: "./funny-cat.jpg" }, { url: "./funny-dog.jpg" }]` is saved as `[{ url: "./funny-cat.jpg", rank: 0 }, { url: "./funny-dog.jpg", rank: 1 }]`.
- Add DnD for the product media form and product create form in admin.
- Fixes an issue where images in the product media gallery had a width of 0px in Firefox.

Resolves CMRC-723, CMRC-724, CMRC-725

CLOSES #10128


packages/modules/product/src/models/product-image.ts
```diff
@@ -58,17 +58,34 @@ class ProductImage {
...
   deleted_at?: Date
 
-  @ManyToMany(() => Product, (product) => product.images)
-  products = new Collection<Product>(this)
+  @Property({ columnType: "integer", default: 0 })
+  rank: number
+
+  @ManyToOne(() => Product, {
+    columnType: "text",
+    nullable: true,
```
olivermrbl: thought: Should this be not null?

olivermrbl (approved):
Overall LGTM, would be nice with a pair of extra eyes, if possible.

olivermrbl:
@adrien2p, any chance you could glance over the BE in this PR?

adrien2p:
> @adrien2p, any chance you could glance over the BE in this PR?

Yes for sure, i can have a look tomorrow morning 🙏

packages/modules/product/integration-tests/__tests__/product-module-service/products.spec.ts
```diff
@@ -1236,6 +1237,152 @@ moduleIntegrationTestRunner<IProductModuleService>({
...
+              })
+            )
+          )
+
+          // Explicitly verify sequential order
+          retrievedProduct.images.forEach((img, idx) => {
+            if (idx > 0) {
+              expect(img.rank).toBeGreaterThan(retrievedProduct.images[idx - 1].rank)
+            }
+          })
```
adrien2p, line 1336:
```suggestion
          retrievedProduct.images.forEach((img, idx) => {
              expect(img.rank).toBeGreaterThan(retrievedProduct.images[idx - 1].rank)
          })
```

packages/modules/product/src/models/product-image.ts
adrien2p: yes we can remove the nullable, plus with the onDelete cascade it can't happen that the image exists after a product deletion.

adrien2p (approved):

packages/modules/product/src/models/product-image.ts
kasperkristensen: If I remove nullable, creating a product wiht images fails with: `ValidationError: Value for ProductImage.product_id is required`

olivermrbl: You might need to assign the `product_id` directly to the image upon creating the product, similar to what you added in the update path.

olivermrbl: You might need to assign the `product_id` directly to the image upon creating the product, similar to what you added in the update path. We have taken this approach for product options and variants too.

So instead of calling `upsertWithReplace` with the `images` relation, I believe we need to do something along these lines:
```ts
if (product.images?.length) {
    const { entities: images } =
      await this.productImageService_.upsertWithReplace(
          product.images?.map((image, index) => ({
            ...image,
            product_id: upsertedProduct.id,
            rank: index
          })),
        { },
        sharedContext
      )
    upsertedProduct.images = images
  }
```

Please correct me if I am wrong @adrien2p.

kasperkristensen: Yes we can do that, but then I will need to refetch the products before we return from both functions, because otherwise the images won't be present (breaking change and breaks a bunch of pre-existing tests). I can just do that, but just want to check if thats alright since we are adding both more separate DB writes and an additional DB read.

olivermrbl: Not sure I understand, where would you need to refetch products? We are assigning the created images to the returned products.

kasperkristensen: I can do that, but then I will also need to overwrite the ProductService, since it has a bunch of tests for creating/updating/restoring products with images and they start failing. I am assuming it's the same reason for why options and variants have the field as nullable.

kasperkristensen: @adrien2p could you maybe point me to where in the code we create the MedusaInternalServices so I can get an understanding of what I am overwriting, it's a bit hard to gage from the types of the .create, update, upsert, and upsertAndReplace functions.

.changeset/flat-mugs-try.md
```diff
@@ -0,0 +1,5 @@
+---
+"@medusajs/dashboard": patch
```
srindom, line 2: question: shouldn't there be more packages here. 

@medusaja/product and @medusajs/types?

srindom (commented):
Looks good to me - will let someone else take a look as well

olivermrbl (approved):
LGTM!

integration-tests/http/__tests__/product/store/product.spec.ts
```diff
@@ -20,6 +20,7 @@ import { createAuthenticatedCustomer } from "../../../../modules/helpers/create-
 jest.setTimeout(30000)
 
 medusaIntegrationTestRunner({
+  debug: true,
```
olivermrbl: nit: 
```suggestion
```

packages/core/types/src/common/common.ts
```diff
@@ -52,7 +52,7 @@ export interface SoftDeletableEntity extends BaseEntity {
 /**
  * Temporary type fixing to allow any level of orders until we get to properly clean all the types
  */
-type FindConfigOrder = { [Key: string]: "ASC" | "DESC" | FindConfigOrder }
+export type FindConfigOrder = { [Key: string]: "ASC" | "DESC" | FindConfigOrder }
```
adrien2p:
```ts
export type FindConfigOrder = { [Key: string]: "ASC" | "DESC" | string & {} | FindConfigOrder }
```

PR #10202, opened 2024-11-21
feat(medusa,core-flows,types,js-sdk): decline / cancel order transfer

fPolic:
**What**
- **core-flows**: 
  - cancel transfer
  - decline transfer
- **API**:
  - admin cancel transfer
  - store decline transfer 
  - store cancel transfer
- **js-sdk**:
  - add methods for store endpoints 

---

CLOSES CMRC-726


integration-tests/http/__tests__/order/admin/transfer-flow.spec.ts
```diff
@@ -141,6 +141,54 @@ medusaIntegrationTestRunner({
...
+          `/admin/orders/${order.id}/transfer/cancel`,
+          {},
+          adminHeaders
+        )
+
+        orderPreviewResult = (
+          await api.get(`/admin/orders/${order.id}/preview`, adminHeaders)
+        ).data.order
+
+        expect(orderPreviewResult.order_change).not.toBeDefined()
```
riqwan, line 183: Should we check this as well?

```suggestion
        expect(orderPreviewResult.order_change).not.toBeDefined()
        expect(orderPreviewResult.customer_id).not.toBeDefined()
```

packages/medusa/src/api/store/orders/middlewares.ts
```diff
@@ -57,4 +71,15 @@ export const storeOrderRoutesMiddlewares: MiddlewareRoute[] = [
       ),
     ],
   },
+  {
+    method: ["POST"],
+    matcher: "/store/orders/:id/transfer/decline",
+    middlewares: [
```
olivermrbl, line 77: thought(non-blocking): Should this be authenticated too?

integration-tests/http/__tests__/order/admin/transfer-flow.spec.ts
```diff
@@ -344,6 +392,267 @@ medusaIntegrationTestRunner({
...
+          `/admin/orders/${order.id}/transfer/cancel`,
+          {},
+          adminHeaders
+        )
+
+        orderChanges = await orderModule.listOrderChanges({
+          order_id: order.id,
+        })
+
+        expect(orderChanges.length).toEqual(0)
```
riqwan: instead of querying the module:

```suggestion
        const orderChangesResult = (
          await api.get(`/admin/orders/${order.id}/changes`, adminHeaders)
        ).data.order_changes

        expect(orderChangesResult.length).toEqual(0)
```

packages/medusa/src/api/admin/orders/[id]/transfer/cancel/route.ts
```diff
@@ -0,0 +1,37 @@
...
+    },
+  })
+
+  const queryObject = remoteQueryObjectFromString({
+    entryPoint: "order",
+    variables: { id: orderId },
+    fields: req.remoteQueryConfig.fields,
+  })
+
+  const [order] = await remoteQuery(queryObject)
```
Comment (olivermrbl) [inline packages/medusa/src/api/admin/orders/[id]/transfer/cancel/route.ts]: todo: Can we use query graph instead? 🙏

packages/core/core-flows/src/order/workflows/transfer/decline-order-transfer.ts
```diff
@@ -0,0 +1,87 @@
...
+      entry_point: "order_change",
+      fields: ["id", "status", "version", "actions.*"],
+      variables: {
+        filters: {
+          order_id: input.order_id,
+          status: [OrderChangeStatus.PENDING, OrderChangeStatus.REQUESTED],
+        },
+      },
+      list: false,
+    }).config({ name: "order-change-query" })
```
olivermrbl: todo: Can we use `useQueryGraphStep` instead? 🙏

olivermrbl: Would like to slowly phase out remot equery

packages/core/core-flows/src/order/workflows/transfer/cancel-order-transfer.ts
```diff
@@ -0,0 +1,93 @@
...
+      entry_point: "order_change",
+      fields: ["id", "status", "version", "actions.*"],
+      variables: {
+        filters: {
+          order_id: input.order_id,
+          status: [OrderChangeStatus.PENDING, OrderChangeStatus.REQUESTED],
+        },
+      },
+      list: false,
+    }).config({ name: "order-change-query" })
```
olivermrbl: See other comment.

integration-tests/http/__tests__/order/admin/transfer-flow.spec.ts
```diff
@@ -344,6 +392,267 @@ medusaIntegrationTestRunner({
...
+                }),
+              }),
+            ]),
+          })
+        )
+
+        const error = await api
+          .post(
+            `/store/orders/${order.id}/transfer/decline`,
+            { token: "fake-token" },
```
riqwan, line 637: q: shouldn't this be an actual token? or maybe another test with the actual token and non customer headers

packages/core/core-flows/src/order/workflows/transfer/cancel-order-transfer.ts
```diff
@@ -0,0 +1,93 @@
...
+import { deleteOrderChangesStep } from "../../steps"
+import {
+  throwIfIsCancelled,
+  throwIfOrderChangeIsNotActive,
+} from "../../utils/order-validation"
+
+/**
+ * This step validates that a requested order transfer can be canceled.
+ */
+export const cancelTransferOrderRequestValidationStep = createStep(
```
olivermrbl, line 27: thought(non-blocking): This validation step makes the workflow less reusable because we rely on certain actor IDs. Is there a way to make this more generic? Maybe pass a validation flag from the HTTP layer?

packages/core/core-flows/src/order/workflows/transfer/cancel-order-transfer.ts
```diff
@@ -0,0 +1,93 @@
...
+    input,
+  }: {
+    order: OrderDTO
+    orderChange: OrderChangeDTO
+    input: OrderWorkflow.CancelTransferOrderRequestWorkflowInput
+  }) {
+    throwIfIsCancelled(order, "Order")
+    throwIfOrderChangeIsNotActive({ orderChange })
+
+    if (input.logged_in_user_id.startsWith("user_")) {
```
riqwan: sugg: The `req.auth_context` provides a `actor_type` that you can pass in if you want to use it for specific types of users.

packages/core/core-flows/src/order/workflows/transfer/cancel-order-transfer.ts
```diff
@@ -0,0 +1,93 @@
...
+ * This workflow cancels a requested order transfer.
+ *
+ * Customer that requested the transfer or a store admin are allowed to cancel the order transfer request.
+ */
+export const cancelOrderTransferRequestWorkflow = createWorkflow(
+  cancelTransferOrderRequestWorkflowId,
+  function (
+    input: WorkflowData<OrderWorkflow.CancelTransferOrderRequestWorkflowInput>
+  ): WorkflowData<void> {
+    const order: OrderDTO = useRemoteQueryStep({
```
riqwan: sugg: We use `useQueryGraphStep` now as I think this is deprecated.

riqwan: eg: https://github.com/medusajs/medusa/blob/develop/packages/core/core-flows/src/cart/workflows/transfer-cart-customer.ts#L17-L30

packages/core/core-flows/src/order/workflows/transfer/cancel-order-transfer.ts
```diff
@@ -0,0 +1,93 @@
...
+    const orderChange: OrderChangeDTO = useRemoteQueryStep({
+      entry_point: "order_change",
+      fields: ["id", "status", "version", "actions.*"],
+      variables: {
+        filters: {
+          order_id: input.order_id,
+          status: [OrderChangeStatus.PENDING, OrderChangeStatus.REQUESTED],
+        },
+      },
+      list: false,
```
riqwan: q: shouldn't this also throw?

```suggestion
      list: false,
      throw_if_key_not_found: true,
```

packages/core/core-flows/src/order/workflows/transfer/decline-order-transfer.ts
```diff
@@ -0,0 +1,87 @@
...
+    input: OrderWorkflow.DeclineTransferOrderRequestWorkflowInput
+  }) {
+    throwIfIsCancelled(order, "Order")
+    throwIfOrderChangeIsNotActive({ orderChange })
+
+    const token = orderChange.actions?.find(
+      (a) => a.action === ChangeActionType.TRANSFER_CUSTOMER
+    )?.details!.token
+
+    if (!input.token.length || token !== input.token) {
```
riqwan: to account for undefined cases:

```suggestion
    if (!input.token?.length || token !== input.token) {
```

packages/core/core-flows/src/order/workflows/transfer/decline-order-transfer.ts
```diff
@@ -0,0 +1,87 @@
...
+    const orderChange: OrderChangeDTO = useRemoteQueryStep({
+      entry_point: "order_change",
+      fields: ["id", "status", "version", "actions.*"],
+      variables: {
+        filters: {
+          order_id: input.order_id,
+          status: [OrderChangeStatus.PENDING, OrderChangeStatus.REQUESTED],
+        },
+      },
+      list: false,
```
riqwan: same here:

```suggestion
      list: false,
      throw_if_key_not_found: true,
```

packages/medusa/src/api/admin/orders/[id]/transfer/cancel/route.ts
Comment (riqwan) [inline packages/medusa/src/api/admin/orders/[id]/transfer/cancel/route.ts]: sugg: We have the new graph query that you can use. Eg: https://github.com/medusajs/medusa/blob/develop/packages/medusa/src/api/admin/orders/%5Bid%5D/line-items/route.ts#L19-L28

packages/medusa/src/api/store/orders/middlewares.ts
fPolic: This endpoint is used for a guest customer to decline a transfer. (Transfer token is required as a param in body.)

packages/medusa/src/api/store/orders/middlewares.ts
```diff
@@ -57,4 +71,15 @@ export const storeOrderRoutesMiddlewares: MiddlewareRoute[] = [
...
+      validateAndTransformBody(StoreDeclineOrderTransferRequest),
```
riqwan, line 78: q: shouldn't this be auth'd as well?

riqwan (approved):
added some comments/suggestions, overall lgtm!

integration-tests/http/__tests__/order/admin/transfer-flow.spec.ts
fPolic: Could you elaborate. Store headers only pass a publishable key and not auth headers.
Transfer can be accepted/declined by a non registered customer if they have a right token (that was sent to them via email). So we have a test above for successful decline if the customer has token  and this test that check if decline fails without a matching token.

packages/medusa/src/api/store/orders/middlewares.ts
fPolic: A guest customer should be able to accept / decline transfer of their order.
(a guest customer who is the original owner of the order needs to provide a token for this operation which is sent to them via e.g. email)

packages/core/core-flows/src/order/workflows/transfer/cancel-order-transfer.ts
fPolic: Would make sense but we check that in the validation step to be consistent with the pattern that is used across order change workflows.

integration-tests/http/__tests__/order/admin/transfer-flow.spec.ts
riqwan: I thought this endpoint had customer authentication. Your comment on the auth middleware solves this! Nothing to do here.

packages/medusa/src/api/store/orders/middlewares.ts
riqwan: I see, ok cool thanks!

integration-tests/http/__tests__/order/admin/transfer-flow.spec.ts
fPolic: but customer_id of the original customer should be defined here, right?

riqwan: ah yeah, in that case expect it to be the old customer or not to be the new customer 🤞🏻

packages/core/core-flows/src/order/workflows/transfer/cancel-order-transfer.ts
fPolic: good point, changed!

PR #10211, opened 2024-11-21
fix: Update js-sdk with token

olivermrbl:
Right now, the auth. request to update password always return 401 because the token was not passed

Closes CMRC-735


fPolic (approved):
LGTM

PR #10223, opened 2024-11-22
fix(stripe): Wrong container type

adrien2p:
RESOLVES SUP-238

**What**

wrong container type withing payment providers


olivermrbl (commented):
thought(non-blocking): Should we rename containers to `cradle` in the provider interfaces, abstract classes, and classes? Perhaps, this would make the access pattern even more obvious.

srindom (approved):
LGTM

adrien2p:
> thought(non-blocking): Should we rename containers to `cradle` in the provider interfaces, abstract classes, and classes? Perhaps, this would make the access pattern even more obvious.

yes that might be better indeed, let me change that

adrien2p:
@olivermrbl are you fine if I just change the name in the constructor but not the class prop as it would create a breaking change?

olivermrbl:
Yes, sure, makes sense to me.

olivermrbl (approved):

PR #10231, opened 2024-11-22
fix: correct ID used for refetching after updating customer

oDestroyeRo:
What
   correct ID used for refetching after updating customer


integration-tests/modules/__tests__/customer/store/update-customer.spec.ts
```diff
@@ -0,0 +1,60 @@
+import { medusaIntegrationTestRunner } from "@medusajs/test-utils"
```
riqwan: Can you move this test to `integration-tests/http/__tests__/customer/store/customer.spec.ts` under a new describe i.e. what you have from line 15?

riqwan (approved):
Nice, thanks for the contribution @oDestroyeRo!

integration-tests/modules/__tests__/customer/store/update-customer.spec.ts
oDestroyeRo: I’ve moved the test to integration-tests/http/__tests__/customer/store/customer.spec.ts

riqwan: legend! 💪🏻

PR #10292, opened 2024-11-26
chore: Upgrade to MikroORM v6 and Migrate Order module to DML

carlos-r-l-rodrigues:
What:
* Order module - DML Migration
* Mikro-ORM 6 migration

Breaking changes:
* Medusa projects have to update their mikro-orm dependencies from `5.9.7` to `6.4.3`


packages/modules/inventory/src/migrations/Migration20240719123015.ts
```diff
@@ -43,13 +43,13 @@ export class Migration20240719123015 extends Migration {
         ALTER TABLE "reservation_item" DROP COLUMN IF EXISTS "raw_quantity";
 
         ALTER TABLE "inventory_level" ALTER COLUMN "stocked_quantity" TYPE integer;
-        ALTER TABLE "inventory_level" DROP COLUMN IF NOT EXISTS "raw_stocked_quantity";
+        ALTER TABLE "inventory_level" DROP COLUMN IF EXISTS "raw_stocked_quantity";
```
olivermrbl, line 46: 😄

packages/modules/link-modules/src/migration/index.ts
```diff
@@ -233,6 +233,28 @@ export class MigrationsExecutionPlanner implements ILinkMigrationsPlanner {
     }))
   }
 
+  private pickTableRelatedCommands(tableName: string, sqlCommand: string) {
```
olivermrbl, line 236: question: Just out of curiosity, why was this needed?

olivermrbl (approved):
epic stuff 🚢

packages/modules/link-modules/src/migration/index.ts
carlos-r-l-rodrigues: This version 6 was generating some extra commands not related to the Link entities. Like trying to drop a TYPE that was defined on the Order Module. So I added this as a safe guard to avoid touching anything that isn't the related table.

olivermrbl: I see, thanks!

carlos-r-l-rodrigues:
@olivermrbl, it's ready to be merged.

PR #9885, opened 2024-10-31
Add french trad in i18n translations

aurelien-f:
Hello, i send you a new pull request. 

What - Add french trad in i18n translations
Why - Offer a new language and be able to offer my clients a French language
How - I used AI and my knowledge of French to do the translation 
Testing - I have not tested the changes.

PR #10301, opened 2024-11-26
French translation V2

MrCasquette:
This PR improves the French translations already submitted.
- Corrected missing translations.
- Ensured the JSON structure matches the English version.
- Verified consistency across all keys.

The file is very large and it's hard to make sure everything is right.

The previous Similar PR is [Add french trad in i18n translations #9885](https://github.com/medusajs/medusa/pull/9885)

Let me know if additional changes are needed! Happy to help :)


kasperkristensen (approved):
LGTM thanks @triplecasquette

MrCasquette:
> LGTM thanks @triplecasquette

I probably made an error with yarn. I'm more used to pnpm or even npm.

I saw that there were a lot of changes in the lockfile maybe the issue comes from here

kasperkristensen:
> > LGTM thanks @triplecasquette
> 
> I probably made an error with yarn. I'm more used to pnpm or even npm.
> 
> I saw that there were a lot of changes in the lockfile maybe the issue comes from here

We use a specific version of yarn so if you don't have that version it will re-write the lock file. But I have reset it to develop so should be good now 👍

MrCasquette:
> > > LGTM thanks @triplecasquette
> > 
> > 
> > I probably made an error with yarn. I'm more used to pnpm or even npm.
> > I saw that there were a lot of changes in the lockfile maybe the issue comes from here
> 
> We use a specific version of yarn so if you don't have that version it will re-write the lock file. But I have reset it to develop so should be good now 👍

Nice. :)

No relation with PR but maybe we should be able to use another package manager.
I find yarn quite long and verbose, whereas pnpm is fast and easy to use.

PR #10309, opened 2024-11-27
fix: Updates email id

SGFGOV:
the entity_id is the same as the email address. However, the body, in turn data received via the api doesn't have an entity id field, it has only an email field. 
Thus, a small PR to fix this. Otherwise one has to add messy middleware to ensure that the entity id is sent as a part of the body.

### DIFF
--- a/packages/modules/providers/auth-emailpass/src/services/emailpass.ts
+++ b/packages/modules/providers/auth-emailpass/src/services/emailpass.ts
@@ -43,11 +43,11 @@ export class EmailPassAuthService extends AbstractAuthModuleProvider {
   }
 
   async update(
-    data: { password: string; entity_id: string },
+    data: { password: string; email: string },
     authIdentityService: AuthIdentityProviderService
   ) {
-    const { password, entity_id } = data ?? {}
-
+    const { password, email } = data ?? {}
+    const entity_id = email
     if (!entity_id) {
       return {
         success: false,


carlos-r-l-rodrigues:
Hi @SGFGOV,
Thanks for the PR.
Do you mind fixing both http tests failing as well?

packages/modules/providers/auth-emailpass/src/services/emailpass.ts
```diff
@@ -43,11 +43,11 @@ export class EmailPassAuthService extends AbstractAuthModuleProvider {
...
 
   async update(
-    data: { password: string; entity_id: string },
+    data: { password: string; email: string },
     authIdentityService: AuthIdentityProviderService
   ) {
-    const { password, entity_id } = data ?? {}
-
+    const { password, email } = data ?? {}
+    const entity_id = email
```
carlos-r-l-rodrigues, line 50:
```suggestion
    const { password, email: entity_id } = data ?? {}
```

olivermrbl:
@SGFGOV, can I get you to share your use case? The email/entity_id should not be sent in the API body; it's pulled from the decrypted token. You are currently only allowed to update the password through the email-password provider. 

Also, we made some changes to this area a few releases back. Please make sure to update to the latest.

packages/modules/providers/auth-emailpass/src/services/emailpass.ts
```diff
@@ -43,11 +43,11 @@ export class EmailPassAuthService extends AbstractAuthModuleProvider {
   }
 
   async update(
-    data: { password: string; entity_id: string },
+    data: { password: string; email: string },
```
sradevski, line 46: In the docs we mention that we pass `email` and `password`, but in the `auth/<actor_type>/emailpass/update` route we pass the `entity_id` as well that is the `actor_id`. 

Can we please add a test to make sure we are not breaking that endpoint with this?

sradevski: @olivermrbl I am not sure how this worked, but the `actor_id` is eg. user_123456, and we passed this as an entity ID, but in the authModule `entity_id` is expected to be the provider's ID eg. `email`.

I think we need to first filter on auth_identity_id, and only after on entity_id which is eg. the email.

olivermrbl: @sradevski, we currently set the actor ID as the entity ID in the middleware, so there is a slight inconsistency here that we can fix. But this means the entity IDfrom the token is actually passed.

olivermrbl: > I think we need to first filter on auth_identity_id, and only after on entity_id which is eg. the email.

Also, not sure I understand this part. Where do you propose we add this?

sradevski: @olivermrbl 

There are two approaches I am thinking of.

**The first one is to do this in the module:**
We pass the authIdentityId to updateProvider, and then to the method factory (it would be optional here)
```

  async updateProvider(
    provider: string,
    authIdentityId: string,
    data: Record<string, unknown>
  ): Promise<AuthenticationResponse> {
    try {
      return await this.authProviderService_.update(
        provider,
        data,
        this.getAuthIdentityProviderService(provider, authIdentityId)
      )
    } catch (error) {
      return { success: false, error: error.message }
    }
  }

```

Then, in the method factory we essentially apply the auth_identity_id to all methods there (list, retrieve), scoping the requests to this specific auth identity ID.

**The second approach is to deletgate these checks to the caller:**
In the HTTP layer we call listAuthIdentities with the auth identity ID from the context, and we check if the provider being updated is part of the `providerIdentities.entity_id` for that authIdentityId. If it is, call updateProvider as now, if not reject.

olivermrbl: > In the HTTP layer we call listAuthIdentities with the auth identity ID from the context, and we check if the provider being updated is part of the providerIdentities.entity_id for that authIdentityId. If it is, call updateProvider as now, if not reject.

I believe this is more or less what we do right now, when a token is passed to the endpoint. Bear in mind we haven't yet added support for updating your provider without a token. We can add that later down the line when we want to support more use cases than password resets.

In the token validation middleware, we pull the `entity_id` from the token and list auth identities with `(provider, entity_id)`. If the auth identity doesn't exist, we reject the request. 

That said, we should still correct the current inconsistency, such that we don't rely on the actor_id for the entity_id.

SGFGOV: @olivermrbl thanks for the clarification, Will cancel the pr

SGFGOV:
already fixed in an later version

PR #10315, opened 2024-11-27
feat(pricing,utils,types): add operator field to price rule

riqwan:
what:

- adds an operator field to price rule with specific operator fields

RESOLVES CMRC-746


olivermrbl (approved):
LGTM

PR #10318, opened 2024-11-27
fix(utils): read dir recursive

carlos-r-l-rodrigues:
What:
 - Using `Object.defineProperty) to avoid issues on Node versions that set this property as readonly.


riqwan (approved):
will give it a test run 🤞🏻

riqwan (approved):
works on node 21, 22, 23.1 & 23.2! 🎉

PR #10335, opened 2024-11-28
refactor: migrate pricing entities to DML models

thetutlage:
Fixes: FRMW-2810

## Breaking changes
There is only one breaking change

- The `min_quantity` and `max_quantity` properties are now consistently typed as numbers. Earlier, it was [typed as numbers](https://github.com/medusajs/medusa/blob/develop/integration-tests/http/__tests__/price-list/admin/price-list.spec.ts#L68-L69) in some API responses and as [string in others](https://github.com/medusajs/medusa/blob/develop/integration-tests/http/__tests__/price-list/admin/price-list.spec.ts#L186-L187). I did not go to the bottom of this inconsistency, but the tests reveals them.


packages/modules/pricing/integration-tests/__tests__/services/pricing-module/price-list.spec.ts
```diff
@@ -246,19 +246,22 @@ moduleIntegrationTestRunner<IPricingModuleService>({
...
-              ends_at: new Date("10/20/2030").toISOString(),
-            })
+
+          expect(priceList).toHaveProperty("starts_at")
+          expect(priceList).toHaveProperty("ends_at")
+          expect(priceList.starts_at?.toString()).toEqual(
+            new Date("10/10/2010").toISOString()
+          )
+          expect(priceList.ends_at?.toString()).toEqual(
+            new Date("10/20/2030").toISOString()
```
thetutlage, line 264: With DML, the `starts_at` field is a Date object during creation so we have expect against the `toString` values.

packages/modules/pricing/integration-tests/__tests__/services/pricing-module/price-list.spec.ts
```diff
@@ -378,11 +381,14 @@ moduleIntegrationTestRunner<IPricingModuleService>({
...
-              ends_at: new Date("10/20/2030").toISOString(),
-            })
+
+          expect(priceList).toHaveProperty("starts_at")
+          expect(priceList).toHaveProperty("ends_at")
+          expect(priceList.starts_at?.toString()).toEqual(
+            new Date("10/10/2010").toString()
+          )
+          expect(priceList.ends_at?.toString()).toEqual(
+            new Date("10/20/2030").toString()
```
thetutlage, line 391: For some weird reason, the field is a `string` after update. Maybe worth looking into?

packages/modules/pricing/integration-tests/__tests__/services/pricing-module/price-list.spec.ts
```diff
@@ -562,7 +568,6 @@ moduleIntegrationTestRunner<IPricingModuleService>({
                   currency_code: "EUR",
                 }),
                 expect.objectContaining({
-                  rules_count: 0,
```
thetutlage, line 565: It was locally initialised in MikroORM entity and we do not have local initialised values in DML

packages/modules/pricing/src/models/price.ts
```diff
@@ -1,148 +1,46 @@
...
+      on: ["price_list_id"],
+      where: "deleted_at IS NULL",
+    },
+    {
+      on: ["currency_code"],
+      where: "deleted_at IS NULL",
+    },
+  ])
+
+Price.name = "Price" as any
```
thetutlage: The model name is converted to `price` via DML which under the uses the `toCamelCase` util and our `toCamelCase` util ain't correct 🤣

packages/modules/pricing/src/services/pricing-module.ts
```diff
@@ -1402,10 +1416,12 @@ export default class PricingModuleService
 
     const priceListsUpsert = priceLists
       .map((priceList) => {
+        const priceListRules =
+          priceList.price_list_rules as unknown as Collection<
+            InferEntityType<typeof PriceListRule>
+          >
```
thetutlage, line 1422: @adrien2p Here is the place where we have to re-convert the type to MikroORM collection. The same thing we talked about during our initial discussions

packages/modules/pricing/integration-tests/__tests__/services/pricing-module/price-list.spec.ts
adrien2p: wouldnt it be that when it comes from the database it is always a string? 🤔 I think that's why originally we typed it as Date | string, because you can have either of those types depending if you are setting it or it comes from the DB. We can validate that if you want but I believe it is it

packages/modules/pricing/src/models/price.ts
adrien2p: The name of the DML and the name of the mikro orm entity can be different in casing, for the mikro orm entity we apply specific conversion, and the idea is that any adapter can get from the base casing and case it has they need for the underlying tec they are using. In the repositories we can just use this.entity.name or this.entity

packages/modules/pricing/src/services/pricing-module.ts
adrien2p: we shouldnt have to do that in the service module layer, we can manipulate it as a classic array and loop through it as it implements the iterator symbol as well. the toArray will apply toJSON on it but I don't think we care much here for that

packages/modules/pricing/integration-tests/__tests__/services/pricing-module/price-list.spec.ts
adrien2p: as per our investigation, the place the serialization/deserialization occurs for the entity is here (node_modules/@mikro-orm/core/types/DateType.js) except that the update is relying on the upsertWithReplace which goes out of the classic path of managed entities and rather use native methods instead, which explain now why the behaviour is different between create and update

packages/modules/pricing/src/models/price.ts
thetutlage: Fixed in a separate PR and merged

packages/modules/pricing/integration-tests/__tests__/services/pricing-module/calculate-price.spec.ts
```diff
@@ -40,7 +40,7 @@ const createPriceLists = async (
   rules: object = defaultRules,
   prices = defaultPriceListPrices
 ) => {
-  return await service.createPriceLists([
+  return service.createPriceLists([
```
olivermrbl: nit: This intentional?

packages/modules/pricing/src/migrations/Migration20241129094116.ts
```diff
@@ -1,6 +1,6 @@
 import { Migration } from "@mikro-orm/migrations"
 
-export class Migration20241127223829 extends Migration {
+export class Migration20241129094116 extends Migration {
```
olivermrbl: thought: We probably shouldn't rename migrations, as it would otherwise be re-applied no?

olivermrbl (approved):
Overall LGTM, nice work!

packages/modules/pricing/integration-tests/__tests__/services/pricing-module/calculate-price.spec.ts
thetutlage: Nope. By mistake, but it shouldn't change the flow or the behavior of the code. However, happy to revert it back

PR #10337, opened 2024-11-28
fix: do not initialize FKs to null

thetutlage:
Fixes: FRMW-2812

This PR changes how the foreign keys are hydrated on a model instance when you fetch data from the database.

- The foreign key is set when you explicitly `select` it.
- The foreign key is set when you explicitly select its relationship. Even if the relationship doesn't exist and set to `null`.


adrien2p (approved):
LGTM 🎉

olivermrbl (approved):

PR #10350, opened 2024-11-28
feat(pricing, types): add price rule operators to price calculations

riqwan:
what:

- adds price rule operators when doing price calculations
- rules now accepts a key where the value can be an array of objects `({ operator: string, value: number })`
  - validation for available types of operator and value to be a number
```
await service.createPriceSets({
  prices: [
    {
      amount: 50,
      currency_code: "usd",
      rules: {
        region_id: "de",
        cart_total: [
          { operator: "gte", value: 400 },
          { operator: "lte", value: 500 },
        ]
      },
    },
  ]
})
```
- price calculations will now account for the operators - lte, gte, lt, gt when the price context is a number

RESOLVES CMRC-747

### DIFF
--- a/packages/core/types/src/pricing/common/price-set.ts
+++ b/packages/core/types/src/pricing/common/price-set.ts
@@ -7,6 +7,7 @@ import {
   MoneyAmountDTO,
   UpdateMoneyAmountDTO,
 } from "./money-amount"
+import { PricingRuleOperatorValues } from "./price-rule"
 
 export interface PricingRepositoryService {
   calculatePrices(
@@ -206,6 +207,11 @@ export interface CalculatedPriceSet {
   }
 }
 
+export interface RuleWithOperator {
+  operator: PricingRuleOperatorValues
+  value: number
+}
+
 /**
  * @interface
  *
@@ -214,7 +220,8 @@ export interface CalculatedPriceSet {
  * Each key of the object is a the attribute, and its value
  * is the values of the rule.
  */
-export interface CreatePriceSetPriceRules extends Record<string, string> {}
+export interface CreatePriceSetPriceRules
+  extends Record<string, string | RuleWithOperator[]> {}
 
 /**
  * @interface
--- a/packages/modules/pricing/integration-tests/__tests__/services/pricing-module/calculate-price.spec.ts
+++ b/packages/modules/pricing/integration-tests/__tests__/services/pricing-module/calculate-price.spec.ts
@@ -10,6 +10,7 @@ import {
   PriceListType,
 } from "@medusajs/framework/utils"
 import { moduleIntegrationTestRunner } from "@medusajs/test-utils"
+import { withOperator } from "../../../__fixtures__/price-rule"
 import { seedPriceData } from "../../../__fixtures__/seed-price-data"
 
 jest.setTimeout(30000)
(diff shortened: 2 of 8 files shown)


olivermrbl:
@riqwan, I like the idea of the shape of `rules`, however one thought: Is this consistent with how our other rules APIs work? I believe we typically pass an array, but I might be mistaken Ideally, we unify the approach across all modules that have the concept of a rule. 

If so, I guess what you would have to do instead would be something like:
```
rules:[
  {
    key: "region_id": 
    value: "de",
  },
  {
    key: "cart_total",
    value: min,
    operator: "gte"
  },
  {
    key: "cart_total",
    value: max,
    operator: "lte"
  },
}],
```

packages/modules/pricing/integration-tests/__tests__/services/pricing-module/calculate-price.spec.ts
```diff
@@ -1857,6 +1858,180 @@ moduleIntegrationTestRunner<IPricingModuleService>({
           })
         })
       })
+
+      describe("calculatePrices", () => {
```
olivermrbl, line 1862: nice

packages/modules/pricing/src/services/pricing-module.ts
```diff
@@ -596,20 +598,48 @@ export default class PricingModuleService
...
+                  `operator should be one of ${ruleOperators.join(", ")}`
+                )
+              }
+
+              if (typeof customRule.value !== "number") {
+                throw new MedusaError(
+                  MedusaError.Types.INVALID_DATA,
+                  `value should be a number`
+                )
+              }
```
olivermrbl, line 620: comment: Wonder if this should live here, it significantly constrains the underlying rules engine of the pricing module. Would it make more sense for this to be at a higher level?

packages/core/types/src/pricing/common/price-set.ts
```diff
@@ -206,6 +207,11 @@ export interface CalculatedPriceSet {
   }
 }
 
+export interface RuleWithOperator {
+  operator: PricingRuleOperatorValues
+  value: number
```
olivermrbl, line 212: Somewhat related to my other comment, I wonder if we should let this be more open-ended at the module level and enforce the type higher in the stack.

It makes sense right now, because of the operators we support, but if we extend that to have `in` and `nin`, then we'd need to be more flexible with the type of value. 

Alternatively, we just make this always be a string and ensure to cast it to a number whenever that is needed

olivermrbl (approved):
OVerall LGTM, few comments

packages/core/types/src/pricing/common/price-set.ts
riqwan: I'll answer both comments:

There's a catch here - our pricing calculation is on a SQL level, the query we build does transformation that converts a string to a numeric on a db level. If that value that we convert is not an actual number, the entire price calculation would break everytime there is incompatible data in any of the results during calculation. 

We could add some checks on a sql level, but it doesn't look pretty. The module would have to safely store the data for the operators it supports. Thats why I took a strict-first approach. Good news is, to support additional operators, we just have to add input validations on the service and add a conditional in the query building.

I should've added this on the description. 🤞🏻 😄 wdyt?

olivermrbl: Yeah, it does make sense, and I think this is a solid first iteration. My questions were mostly out of curiosity about how we will extend this in the future.

When we introduce new operation, I guess we would also need to include some sort of mapping between what type a value can have for a specific operator. For example, you won't use a string value for `gte`, etc.

riqwan: yes, exactly. Shall we merge this as is then?

olivermrbl: Yes let's do it!

PR #10360, opened 2024-11-29
fix: do not rely on model loading order to find an implicit owner

thetutlage:
Fixes: FRMW-2814

This is something that was supposed to be fixed in the product DML migration PR. However, it slipped through the cracks and hence must be fixed ASAP

### DIFF
--- a/packages/core/utils/src/dml/helpers/entity-builder/define-relationship.ts
+++ b/packages/core/utils/src/dml/helpers/entity-builder/define-relationship.ts
@@ -486,16 +486,12 @@ export function defineManyToManyRelationship(
   }
 
   /**
-   * If the other side has been tracked and any of the properties
-   * are provided on the other side, then we do not expect the
-   * current side to be the owner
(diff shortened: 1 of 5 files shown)


packages/core/utils/src/dml/helpers/entity-builder/define-relationship.ts
```diff
@@ -492,22 +475,36 @@ export function defineManyToManyRelationship(
...
+   */
   const otherSideRelationOptions = otherSideRelationship.parse("").options
+  if (
+    isOwner === undefined &&
+    (otherSideRelationOptions.pivotTable ||
+      otherSideRelationOptions.joinColumn ||
+      otherSideRelationOptions.inverseJoinColumn)
+  ) {
+    isOwner = false
+  }
```
adrien2p: Suggestion: We could probably check if the configuration is on both side and in that case compose an error message to explain that it should only be configured on one side (the owner side)

adrien2p (approved):

PR #10368, opened 2024-11-29
feat(core-flows,types,medusa): ability to update/create custom shipping prices

riqwan:
what:

- adds rules to existing shipping option prices http validations
- update and create shipping option workflows are updated to be able to handle custom prices

The types and workflow steps in shipping options could use a nice cleanup. I'll schedule this after unblocking the FE tickets. 

RESOLVES CMRC-750
RESOLVES CMRC-749


olivermrbl (approved):
LGTM, nice work

PR #10369, opened 2024-11-29
fix(ui,icons,ui-preset): Fix CodeBlock and Command components

kasperkristensen:
**What**
- Updates ui-preset and icons packages with latest changes from Figma.
- Also noticed that all the Code related components were broken because the color and typography styles in Figma was renamed from `Code Block/*` to `Code/*` at some point. I have updated the components to use the proper styles, and tried to implement the CodeBlock as close to the Figma design as possible without introducing any breaking changes. The component will need to be revisited at a point when we are okay with introducing a breaking change, since a lot has been changed in the component as part of the Docs re-design. But this should make it more than usable for the time being.

**Prev**
Skærmbillede 2024-11-29 kl  16 47 43

**Now**

https://github.com/user-attachments/assets/ac5a2678-46c6-4bfe-aaa0-34a6d2d940b2


fPolic (approved):

olivermrbl (approved):
LGTM

PR #10373, opened 2024-11-29
feat(core-flows,medusa,order,types): update orders

fPolic:
**What**
- add order update endpoint
- add workflows and steps for updating orders
- add `registerChanges` method to Order module + workflow step

---

CLOSES CMRC-633


packages/core/core-flows/src/order/workflows/update-shipping-address.ts
```diff
@@ -0,0 +1,118 @@
...
+    const orderChangeInput = transform({ input }, ({ input }) => {
+      return {
+        change_type: "shipping_address_change" as const,
+        order_id: input.order_id,
+        description: input.description,
+        internal_note: input.internal_note,
+      }
+    })
+
+    const change = createOrderChangeStep(orderChangeInput)
```
fPolic: self-comment: should we move the next 4 calls to an order service method like `registerOrderChange` that would create a dummy change record in case we wanna note that some generic change occurred on an order. 
Method would:
1. create already confirmed order change
2. create a generic action for the change that is already "applied"

packages/medusa/src/api/admin/orders/[id]/route.ts
```diff
@@ -21,3 +29,29 @@ export const GET = async (
...
+  res: MedusaResponse<HttpTypes.AdminOrderResponse>
+) => {
+  const remoteQuery = req.scope.resolve(ContainerRegistrationKeys.REMOTE_QUERY)
+
+  const variables = { id: req.params.id }
+
+  await updateOrderWorkflow(req.scope).run({
+    // @ts-ignore - TODO: null/undefined issue
+    input: {
+      ...req.body,
```
Comment (olivermrbl) [inline packages/medusa/src/api/admin/orders/[id]/route.ts]:
```suggestion
      ...req.validatedBody,
```

packages/core/core-flows/src/order/workflows/update-shipping-address.ts
carlos-r-l-rodrigues: That is a good point.  think it would be good to have this generic method on the order class.
That could be used for general "comments" on the order as well.

integration-tests/http/__tests__/order/admin/order.spec.ts
```diff
@@ -20,6 +20,105 @@ medusaIntegrationTestRunner({
...
+        )
+
+        const orderChangesResult = (
+          await api.get(`/admin/orders/${order.id}/changes`, adminHeaders)
+        ).data.order_changes
+
+        expect(orderChangesResult.length).toEqual(1)
+        expect(orderChangesResult[0]).toEqual(
+          expect.objectContaining({
+            version: 2, // TODO: should we bump the version here?
```
carlos-r-l-rodrigues: no, I believe there is no need to bump the version here

packages/modules/order/src/utils/actions/change-shipping-address.ts
```diff
@@ -0,0 +1,20 @@
+import { ChangeActionType } from "@medusajs/framework/utils"
+
+import { OrderChangeProcessing } from "../calculate-order-change"
+import { setActionReference } from "../set-action-reference"
+
+OrderChangeProcessing.registerActionType(
+  ChangeActionType.CHANGE_SHIPPING_ADDRESS,
```
olivermrbl: thought: I would be curious to hear @carlos-r-l-rodrigues's thoughts on creating a more generic action, that can be used for updates that does not have any validation and where the operation is basically just updating the property on the order.

For example:
```ts
OrderChangeProcessing.registerActionType(
  ChangeActionType.ORDER_UPDATE,
  {
    operation({ action, currentOrder, options }) {
      /**
       * NOOP: used as a reference for the change
       */

      setActionReference(currentOrder, action, options)
    },
    validate({ action }) {
      /* noop */
    },
  }
)
```

The order change action generated for this type of update would then be:

**Shipping address**
```ts
const actionInput = transform(
   { order, input, change },
   ({ order, input, change }) => [
     {
       order_change_id: change.id,
       order_id: input.order_id,
       action: ChangeActionType.ORDER_UPDATE,
       version: change.version,
       reference: "shipping_address"
       applied: true,
       reference_id: order.shipping_address?.id,
       details: input.shipping_address as Record<string, unknown>,
     },
   ]
 )
```

**Email**
```ts
const actionInput = transform(
   { order, input, change },
   ({ order, input, change }) => [
     {
       order_change_id: change.id,
       order_id: input.order_id,
       action: ChangeActionType.ORDER_UPDATE,
       version: change.version,
       reference: "email"
       applied: true,
       reference_id: order.email,
       details: { email: input.email },
     },
   ]
 )
```

**Billing address**
```ts
const actionInput = transform(
   { order, input, change },
   ({ order, input, change }) => [
     {
       order_change_id: change.id,
       order_id: input.order_id,
       action: ChangeActionType.ORDER_UPDATE,
       version: change.version,
       reference: "billing_address"
       applied: true,
       reference_id: order.billing_address?.id,
       details: input.billing_address as Record<string, unknown>,
     },
   ]
 )
```

Perhaps, a question to ask: Do we gain anything from having the specific actions? If not, we made as well save ourselves from having multiple actions for these updates.

carlos-r-l-rodrigues (approved):

packages/modules/order/src/utils/actions/change-shipping-address.ts
carlos-r-l-rodrigues: Currently, that action accepts anything, even if there is no "action" associated with it on the Order module.
So, it is really up to us to decide. To me, an ORDER_UPDATE seems to fit well, and anything to distinguish them on the UI, can go on the details

packages/medusa/src/api/admin/orders/[id]/route.ts
```diff
@@ -21,3 +29,28 @@ export const GET = async (
...
+  const variables = { id: req.params.id }
+
+  await updateOrderWorkflow(req.scope).run({
+    input: {
+      ...req.validatedBody,
+      id: req.params.id,
+    },
+  })
+
+  const queryObject = remoteQueryObjectFromString({
```
Comment (olivermrbl) [inline packages/medusa/src/api/admin/orders/[id]/route.ts]: todo: Use query graph

olivermrbl (approved):
This is a neat approach. LGTM – would be good to have @carlos-r-l-rodrigues take a look at it as well

carlos-r-l-rodrigues (approved):
Nice... LGTM

olivermrbl:
@fPolic, should we merge this?

PR #10382, opened 2024-12-01
feat(core-flows, types): update shipping methods upon cart ops

riqwan:
what:

- when cart items change, the shipping methods are changed to reflect the new pricing

RESOLVES CMRC-752


packages/core/core-flows/src/cart/workflows/refresh-cart-items.ts
```diff
@@ -100,7 +101,9 @@ export const refreshCartItemsWorkflow = createWorkflow(
       list: false,
     }).config({ name: "refetch–cart" })
 
-    refreshCartShippingMethodsStep({ cart: refetchedCart })
+    refreshCartShippingMethodsWorkflow.runAsStep({
+      input: { cart_id: cart.id },
```
olivermrbl, line 105: thought: Would it make sense for the workflow to accept `cartOrCartId`? We are already retrieving the cart with shipping methods in the previous step, so it would save us from a DB call

packages/core/core-flows/src/cart/steps/update-shipping-methods.ts
```diff
@@ -0,0 +1,49 @@
...
+import {
+  Modules,
+  getSelectsAndRelationsFromObjectArray,
+} from "@medusajs/framework/utils"
+import { StepResponse, createStep } from "@medusajs/framework/workflows-sdk"
+
+export interface UpdateLineItemsStepInput {
+  id: string
+  items: UpdateLineItemWithSelectorDTO[]
+}
```
olivermrbl: rm
```suggestion
```

packages/core/core-flows/src/cart/steps/refresh-cart-shipping-methods.ts
```diff
@@ -1,73 +0,0 @@
-import {
```
olivermrbl, line 1: suggestion: Maybe just keep this step in case someone is using it (doubt it, but might as well avoid the theorical breaking change)

integration-tests/modules/__tests__/cart/store/cart.workflows.spec.ts
```diff
@@ -2128,18 +2027,15 @@ medusaIntegrationTestRunner({
...
-
-          const { errors } = await listShippingOptionsForCartWorkflow(
+          const { errors, result } = await listShippingOptionsForCartWorkflow(
             appContainer
           ).run({
             input: { cart_id: cart.id },
             throwOnError: false,
           })
 
+          console.log("result -- ", result)
```
olivermrbl:
```suggestion
```

packages/core/core-flows/src/cart/workflows/list-shipping-options-for-cart.ts
```diff
@@ -126,56 +172,49 @@ export const listShippingOptionsForCartWorkflow = createWorkflow(
...
+            MedusaError.Types.INVALID_DATA,
+            `Shipping options with IDs ${ids} do not have a price`
+          )
+        }
 
-        throw new MedusaError(
-          MedusaError.Types.INVALID_DATA,
-          `Shipping options with IDs ${ids} do not have a price`
+        return options.filter((option) =>
+          isDefined(option?.calculated_price?.calculated_amount)
```
olivermrbl: thought: I believe this filter is redundant, as we would otherwise have thrown on undefined prices ealier

packages/core/core-flows/src/cart/workflows/list-shipping-options-for-cart.ts
```diff
@@ -126,56 +172,49 @@ export const listShippingOptionsForCartWorkflow = createWorkflow(
...
-      })
+          return {
+            ...shippingOption,
+            amount: calculated_price?.calculated_amount,
+            is_tax_inclusive:
+              !!calculated_price?.is_calculated_price_tax_inclusive,
+          }
+        })
+
+        const shouldValidatePrices = input.options?.shouldValidatePrices ?? true
```
olivermrbl: thought: What's the idea with this option? When do we not want to validate prices

packages/core/types/src/cart/mutations.ts
```diff
@@ -767,6 +767,11 @@ export interface UpdateShippingMethodDTO {
    */
   amount?: BigNumberInput
 
+  /**
+   * The amount of the shipping method.
```
olivermrbl:
```suggestion
   * The tax inclusivity setting of the shipping method.
```

packages/core/core-flows/src/cart/workflows/add-to-cart.ts
```diff
@@ -111,43 +105,13 @@ export const addToCartWorkflow = createWorkflow(
...
+    emitEventStep({
+      eventName: CartWorkflowEvents.UPDATED,
+      data: { id: input.cart.id },
     })
-
-    refreshPaymentCollectionForCartWorkflow.runAsStep({
-      input: {
-        cart_id: input.cart.id,
-      },
-    })
```
olivermrbl, line 149: question: What happened to these?

riqwan: All of these were already present in the refresh items workflow. Didn't make sense to run them twice

olivermrbl: Oh waow, I hadn't seen. This is great improvement then

packages/core/core-flows/src/cart/workflows/list-shipping-options-for-cart.ts
riqwan: When we create prices for shipping options,we validate them. 

When we're updating price during a cart op, we remove the ones with invalid prices

Happy to choose another way for this, didn't like it either.

packages/core/core-flows/src/cart/workflows/list-shipping-options-for-cart.ts
riqwan: depends on the validation comment here - https://github.com/medusajs/medusa/pull/10382/files#r1867735179

packages/core/core-flows/src/cart/workflows/refresh-cart-shipping-methods.ts
```diff
@@ -0,0 +1,115 @@
...
+                !isPresent(
+                  shippingOptions.find(
+                    (shippingOption) =>
+                      shippingOption.id === shippingMethod.shipping_option_id &&
+                      isDefined(
+                        shippingOption.calculated_price?.calculated_amount
+                      )
+                  )
+                )
+            )
```
olivermrbl: nit: This is a bit hard to read IMO – can we do something about that?

packages/core/core-flows/src/cart/workflows/list-shipping-options-for-cart.ts
```diff
@@ -71,7 +88,7 @@ export const listShippingOptionsForCartWorkflow = createWorkflow(
     )
 
     const customerGroupIds = when({ cart }, ({ cart }) => {
-      return !!cart.id
+      return !!cart.customer_group_id
```
olivermrbl: q: does the cart have a `customer_group_id`? 🤔

packages/core/core-flows/src/cart/workflows/list-shipping-options-for-cart.ts
olivermrbl: I'd almost rather have two steps – one for validating and one without it. I think it's not obvious what's going on here. Wdyt?

integration-tests/http/__tests__/shipping-option/store/shipping-option.spec.ts
```diff
@@ -271,9 +258,11 @@ medusaIntegrationTestRunner({
             expect.objectContaining({
               id: shippingOption.id,
               name: "Test shipping option",
-              amount: 500,
```
olivermrbl, line 274: nit: Think amount is still there right?

packages/core/core-flows/src/cart/workflows/list-shipping-options-for-cart.ts
```diff
@@ -39,6 +30,7 @@ export const listShippingOptionsForCartWorkflow = createWorkflow(
         "id",
         "sales_channel_id",
         "currency_code",
+        "customer.groups.id",
```
olivermrbl: question: If I am not mistaken, we have changed how customer group ID is passed in the pricing context – just want to make sure this is intentional.

We used to pass is as a top-level property:
```
{ ..., customer_group_id: "cusgrp_1234" }
```

Now, it seems we pass it via the cart:
```
{ cart: { customer: { groups: [ { id: "cusgrp_1234" } ] } } }
```

Is that correct, or have I missed something?

olivermrbl (approved):

integration-tests/http/__tests__/shipping-option/store/shipping-option.spec.ts
riqwan: Yes it is, will add that back.

packages/core/core-flows/src/cart/workflows/list-shipping-options-for-cart.ts
riqwan: This was leak from the draft work I had for the changes related to the customer groups i was talking about. Will revert this. 

All of our customer group rules are set as the attribute `customer_group_id`, which doesn't exist in the cart context. The second version you listed should be the actual rule path - `customer.groups.id`. This needs a data migration on price lists and prices, along with some changes in the pricing module, so thought will open a separate PR.

PR #10389, opened 2024-12-02
feat(user): Migrate user module to DML

adrien2p:
RESOLVES FRMW-2817
**What**
Migrate user to DML

### DIFF
--- a/packages/modules/user/integration-tests/__tests__/invite.spec.ts
+++ b/packages/modules/user/integration-tests/__tests__/invite.spec.ts
@@ -8,22 +8,16 @@ import jwt, { JwtPayload } from "jsonwebtoken"
 
 jest.setTimeout(30000)
 
-const expireDate = new Date().setMilliseconds(
-  new Date().getMilliseconds() + 60 * 60 * 24
-)
-
 const defaultInviteData = [
   {
     id: "1",
     email: "user_1@test.com",
     token: "test",
-    expires_at: expireDate,
   },
   {
     id: "2",
     email: "user_2@test.com",
     token: "test",
-    expires_at: expireDate,
   },
 ]
 
--- a/packages/modules/user/src/migrations/Migration20241202103352.ts
+++ b/packages/modules/user/src/migrations/Migration20241202103352.ts
@@ -1,0 +1,21 @@
+import { Migration } from '@mikro-orm/migrations';
+
+export class Migration20241202103352 extends Migration {
+
+  async up(): Promise<void> {
+    this.addSql('drop index if exists "IDX_invite_email";');
+    this.addSql('CREATE UNIQUE INDEX IF NOT EXISTS "IDX_invite_email_unique" ON "invite" (email) WHERE deleted_at IS NULL;');
+
+    this.addSql('drop index if exists "IDX_user_email";');
+    this.addSql('CREATE UNIQUE INDEX IF NOT EXISTS "IDX_user_email_unique" ON "user" (email) WHERE deleted_at IS NULL;');
+  }
+
+  async down(): Promise<void> {
+    this.addSql('drop index if exists "IDX_invite_email_unique";');
+    this.addSql('CREATE UNIQUE INDEX IF NOT EXISTS "IDX_invite_email" ON "invite" (email) WHERE deleted_at IS NULL;');
+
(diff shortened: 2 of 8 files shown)


packages/modules/user/integration-tests/__tests__/invite.spec.ts
```diff
@@ -8,22 +8,16 @@ import jwt, { JwtPayload } from "jsonwebtoken"
...
     id: "1",
     email: "user_1@test.com",
     token: "test",
-    expires_at: expireDate,
   },
   {
     id: "2",
     email: "user_2@test.com",
     token: "test",
-    expires_at: expireDate,
```
adrien2p, line 26: NOTE: The given value was overritten by the module so it was not doing anything

packages/modules/user/src/services/user-module.ts
```diff
@@ -151,9 +157,7 @@ export default class UserModuleService
     const updates = invites.map((invite) => {
       return {
         id: invite.id,
-        expires_at: new Date().setMilliseconds(
-          new Date().getMilliseconds() + this.config.expiresIn * 1000
-        ),
+        expires_at: new Date(Date.now() + this.config.expiresIn * 1000),
```
adrien2p, line 160: NOTE: this was giving a big int instead of a date, now it does the same thing but provide a date instead

packages/modules/user/src/services/user-module.ts
```diff
@@ -311,26 +315,16 @@ export default class UserModuleService
...
-        expires_at: new Date().setMilliseconds(
-          new Date().getMilliseconds() + this.config.expiresIn * 1000
-        ),
-        token: this.generateToken({ id: invite.id, email: invite.email }),
-      }
-    })
-
-    return await this.inviteService_.update(updates, sharedContext)
+    return await this.inviteService_.create(toCreate, sharedContext)
   }
```
adrien2p, line 328: NOTE: This is a simplification, instead of doing a create and then an update, we only perform a create but the result remain unchanged

packages/modules/user/src/migrations/Migration20241202103352.ts
```diff
@@ -0,0 +1,21 @@
+import { Migration } from '@mikro-orm/migrations';
+
+export class Migration20241202103352 extends Migration {
+
+  async up(): Promise<void> {
+    this.addSql('drop index if exists "IDX_invite_email";');
+    this.addSql('CREATE UNIQUE INDEX IF NOT EXISTS "IDX_invite_email_unique" ON "invite" (email) WHERE deleted_at IS NULL;');
+
+    this.addSql('drop index if exists "IDX_user_email";');
+    this.addSql('CREATE UNIQUE INDEX IF NOT EXISTS "IDX_user_email_unique" ON "user" (email) WHERE deleted_at IS NULL;');
```
adrien2p, line 10: NOTE: this apply the right name based on convention

packages/modules/user/src/models/user.ts
```diff
@@ -1,102 +1,18 @@
...
-    columnType: "timestamptz",
-    defaultRaw: "now()",
+import { model } from "@medusajs/framework/utils"
+
+export const User = model
+  .define("user", {
+    id: model.id({ prefix: "user" }).primaryKey(),
+    first_name: model.text().searchable().nullable(),
+    last_name: model.text().searchable().nullable(),
+    email: model.text().searchable().searchable(),
```
olivermrbl:
```suggestion
    email: model.text().searchable(),
```

olivermrbl (approved):
LGTM

packages/modules/user/src/services/user-module.ts
```diff
@@ -311,26 +315,16 @@ export default class UserModuleService
     }
 
     const toCreate = data.map((invite) => {
+      const id = generateEntityId((invite as { id?: string }).id, "invite")
```
olivermrbl, line 318: thought: Don't think we need this anymore, should be generated automatically no?

adrien2p: I manually provide the id here to be able to generate the token without having to first create and then update the create tuples. We have the same approach in the fulfillment where we rely on the id before it gets created, that way it become known from there and allow us to cut down the back and forth

PR #10391, opened 2024-12-02
feature: add support for check constraints in DML

thetutlage:
Fixes: FRMW-2818

The `.checks` method on the DML could be used to define the PostgreSQL constraints. In the following example, we configure a constraint that validates the `amount` column on the `ShippingMethod` model to have a value of `>=0`.

```ts
const ShippingMethod = model.define('ShippingMethod', {
  amount: model.bigNumber(),
})
.checks([
  (columns) => `${columns.amount} >= 0`
])
```

The `.checks` method accepts an array of callbacks or objects. The object signature must match have the properties accepted by the MikroORM's `@Check` decorator. https://mikro-orm.io/docs/5.9/defining-entities#check-constraints

```ts
const ShippingMethod = model.define('ShippingMethod', {
  amount: model.bigNumber(),
})
.checks([
  {
    name: 'shipping_method_amount_check',
    expression: (columns) => `${columns.amount} >= 0`
  }
])
```

Checks will result in the following migration statement.

```ts
this.addSql(`alter table if exists "cart_shipping_method" add constraint shipping_method_amount_check check(amount >= 0);`)
```

### DIFF
--- a/.changeset/hip-suns-sort.md
+++ b/.changeset/hip-suns-sort.md
@@ -1,0 +1,6 @@
+---
+"@medusajs/types": patch
+"@medusajs/utils": patch
+---
+
+feature: add support for check constraints in DML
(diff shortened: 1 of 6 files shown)


olivermrbl (approved):
LGTM

PR #10394, opened 2024-12-02
German language corrections

pixelwiese:
I corrected some terms that were translated wrong.


kasperkristensen (approved):
LGTM thanks @pixelwiese

PR #10412, opened 2024-12-03
fix(dashboard): Add default value to inventory item combobox

kasperkristensen:
**What**
- Resolves CMRC-761
- Fixes TS errors in the same file.


packages/admin/dashboard/src/hooks/use-combobox-data.tsx
```diff
@@ -50,6 +50,8 @@ export const useComboboxData = <
     enabled: !!defaultValue,
   })
 
+  console.log("initial data", initialData, queryInitialDataBy, defaultValue)
```
olivermrbl:
```suggestion
```

olivermrbl (approved):

PR #10441, opened 2024-12-05
feat: add support for defining hasOne with FK

thetutlage:
Fixes: FRMW-2825

In this PR we add support for defining `hasOne` relationship in DML with support for foreign keys. This is added because of the way relationships are defined within our Medusa modules and many of them needs a `hasOne with Cascades` and `FK` on the same side.

A good example of this is `Cart` and `Address`. The relationship is defined as `Cart hasOne Address` and the `Cart` model has the `address_id`. Also, we want a new address to be created automatically via `cascades: ['persist']` setting when defined on a Cart.


adrien2p (approved):
LGTM, lets move with this and see how it goes during the rest of the migration

thetutlage:
@olivermrbl and @shahednasser. We do not want to document this right now, as we might see some breaking changes coming to `hasOne` with foreign keys as we migrate modules.

PR #10445, opened 2024-12-05
fix(core/js-sdk) Add the credentials field in the fetch() only if supported

sgirones:
solves https://github.com/medusajs/nextjs-starter-medusa/issues/421

The root cause of the issue is that `credentials` is not supported by CloudFlare workers.


sgirones:
closing in favour of https://github.com/medusajs/medusa/pull/10446

PR #10459, opened 2024-12-05
fix(workflows-sdk): name for when/then step

carlos-r-l-rodrigues:
What:

`when/then` composition creates a step with a random name (ulid) under the hood when the `then` function is not returning a step. This is causing multiple Medusa instances fail to continue the execution of `async` flows because the step name will differ on each instance.

This PR removes the creation of a step if the `when/then` doesn't return anything, and it introduces the `name` parameter when creating a `when/then` that will return a value different than a step.

```typescript
createWorkflow("wf-when", function (input) {
  step1()
  const subWorkflowRes = when("sub-flow", { input }, ({ input }) => {
    return input.callSubFlow
  }).then(() => {
    const res = subWorkflow.runAsStep({
      input: "hi from outside",
    })

    // The return is an object
    return {
      result: res,
    }
  })

  return new WorkflowResponse(step3(subWorkflowRes.result))
})
```

Some other fixes were applied to execute the same workflow in parallel.

FIXES: FRMW-2824


olivermrbl (commented):
LGTM, were these the only steps we needed to change to not have the warning for our core workflows?

carlos-r-l-rodrigues:
> LGTM, were these the only steps we needed to change to not have the warning for our core workflows?

Yes, only these 2.

olivermrbl (commented):
nice!

olivermrbl (approved):

PR #10471, opened 2024-12-06
refactor: migration stock location module

thetutlage:
Fixes: FRMW-2831


packages/modules/stock-location/src/migrations/Migration20241206092214.ts
```diff
@@ -0,0 +1,13 @@
+import { Migration } from '@mikro-orm/migrations';
+
+export class Migration20241206092214 extends Migration {
+
+  async up(): Promise<void> {
+    this.addSql('alter table if exists "stock_location" add constraint "stock_location_address_id_unique" unique ("address_id");');
```
carlos-r-l-rodrigues: Do we want to make this unique?
In this case, shouldn't it be WHERE on deleted_at is not null?

thetutlage: I think this uniqueness is added by the DML `FK` thing. But yeah, we have to revisit it

olivermrbl: Is this something we should address now? In any case, let's create a ticket for it if we don't have it

thetutlage: Fixed. It was because of the way relationship was setup.

packages/modules/stock-location/src/models/stock-location.ts
```diff
@@ -1,80 +1,19 @@
...
-  Property,
-} from "@mikro-orm/core"
-
-import { StockLocationAddress } from "./stock-location-address"
-
-const StockLocationDeletedAtIndex = createPsqlIndexStatementHelper({
-  tableName: "stock_location",
-  columns: "deleted_at",
-  where: "deleted_at IS NOT NULL",
+import { model } from "@medusajs/framework/utils"
```
olivermrbl, line 1: question: There are some commented out things in this file – what should we do with those?

thetutlage: Ohh sorry. Removed now :)

carlos-r-l-rodrigues (approved):
LGTM

olivermrbl (approved):

PR #10476, opened 2024-12-06
fix(utils): avoid optional properties on graphql generated file

carlos-r-l-rodrigues:
FIXES: SUP-367


thetutlage (approved):

adrien2p (approved):
LGTM, could we update the fixtures in the tests with the new expected output? 🙃

PR #10477, opened 2024-12-06
chore(workflow-engine): Migrate to DML

adrien2p:
RESOLVES FRMW-2832
RESOLVES FRMW-2833

**What**
Migrate workflow engines to DML. Alos includes and update to the linkable generation which now takes into account id and primary keys to generate the linkable instead of only primary keys

### DIFF
--- a/packages/modules/workflow-engine-inmemory/src/utils/workflow-orchestrator-storage.ts
+++ b/packages/modules/workflow-engine-inmemory/src/utils/workflow-orchestrator-storage.ts
@@ -48,6 +48,8 @@ export class InMemoryDistributedTransactionStorage
   }
 
   private async saveToDb(data: TransactionCheckpoint) {
+    const foo = await this.workflowExecutionService_.list()
+    console.log(foo)
     await this.workflowExecutionService_.upsert([
       {
         workflow_id: data.flow.modelId,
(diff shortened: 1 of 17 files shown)


packages/modules/workflow-engine-inmemory/src/utils/workflow-orchestrator-storage.ts
```diff
@@ -48,6 +48,8 @@ export class InMemoryDistributedTransactionStorage
   }
 
   private async saveToDb(data: TransactionCheckpoint) {
+    const foo = await this.workflowExecutionService_.list()
+    console.log(foo)
```
olivermrbl: todo: rm
```suggestion
```

olivermrbl (approved):
Nice, would be nice to apply the provider pattern to the workflow engine (and event bus + cache) at some point, so we can get rid of the duplicated code

adrien2p:
> Nice, would be nice to apply the provider pattern to the workflow engine (and event bus + cache) at some point, so we can get rid of the duplicated code

Yes I do agree, but that is another project and breaking changes, so I didn't do it here

carlos-r-l-rodrigues (approved):

olivermrbl:
> but that is another project and breaking changes, so I didn't do it here

💯

PR #10481, opened 2024-12-06
fix(dashboard,ui): ConditionBlock styling

kasperkristensen:
**What**
- Resolves CMRC-58
- Also fixes some other issues, TS errors, Eslint warnings etc. in the Promotion domain. There are still several TS errors as the return types from `@medusajs/types` don't seem to match how they are used here, but I have left that as is, as I am not super familiar with the Promotion module.


packages/design-system/ui/src/components/code-block/code-block.tsx
```diff
@@ -1,13 +1,13 @@
 "use client"
-import { Highlight, Prism, themes } from "prism-react-renderer"
-import * as React from "react"
-;(typeof global !== "undefined" ? global : window).Prism = Prism
+import { Highlight, Prism, themes } from "prism-react-renderer";
```
olivermrbl: todo: some linting gone wrong here?

olivermrbl (approved):
LGTM

PR #10495, opened 2024-12-08
feat(core-flows, dashboard, fulfillment, fulfillment-manual, utils, types): create shipping options with calculated prices

fPolic:
**What**
- support creating SO with calculated price
- support updating SO for both types of pricing
- update `validateShippingOptionPricesStep` to handle both SO price_types
- add the `validateShippingOptionsForPriceCalculation` method to `FulfillementModule`
- add `canCalculate` and `calculatePrice` to fulfillment provider service service / interface / manual provider
- disable SO pricing edit on Admin if SO price type is calculated

---

CLOSES CMRC-776


packages/core/core-flows/src/fulfillment/workflows/create-shipping-options.ts
```diff
@@ -33,7 +33,13 @@ export const createShippingOptionsWorkflow = createWorkflow(
...
     const data = transform(input, (data) => {
       const shippingOptionsIndexToPrices = data.map((option, index) => {
-        const prices = option.prices
+        /**
+         * Flat rate ShippingOptions always needs to provide a price array.
+         *
+         * For calculated pricing we create an "empty" price set
+         * so we can have simpler update flow for both cases and allow updating price_type.
+         */
+        const prices = (option as any).prices ?? []
```
fPolic, line 42: comment: we always create associated price set for a shipping option, both calculated and flat rate, to make the implementation of create and update flows simpler and more unified (e.g. we don't have to duplicate the logic of price set creation in the update flow if the option price type changes to "flat rate")

packages/core/core-flows/src/fulfillment/workflows/update-shipping-options.ts
```diff
@@ -32,11 +33,18 @@ export const updateShippingOptionsWorkflow = createWorkflow(
 
     const data = transform(input, (data) => {
       const shippingOptionsIndexToPrices = data.map((option, index) => {
-        const prices = option.prices
-        delete option.prices
+        const prices = (option as any).prices
+        delete (option as any).prices
```
olivermrbl: q: can we type this better?

packages/core/types/src/workflow/fulfillment/create-shipping-options.ts
```diff
@@ -1,33 +1,37 @@
-import { ShippingOptionDTO, ShippingOptionPriceType } from "../../fulfillment"
+import { ShippingOptionDTO } from "../../fulfillment"
 import { RuleOperatorType } from "../../common"
 
-export interface CreateShippingOptionsWorkflowInput {
+export type CreateShippingOptionsWorkflowInput = {
```
olivermrbl: nit(non-blocking): this type is starting to get a bit convoluted – perhaps we can refactor to separate types

packages/core/core-flows/src/fulfillment/steps/validate-shipping-option-prices.ts
```diff
@@ -1,26 +1,85 @@
...
+      option.price_type === ShippingOptionPriceType.FLAT
+        ? ((option.prices ?? []) as { region_id: string; amount: number }[])
+        : []
+    )
+
+    const calculatedOptions = options
+      .filter(
+        (option) => option.price_type === ShippingOptionPriceType.CALCULATED
+      )
+      .flatMap((option) => option ?? [])
```
olivermrbl: nit(non-blocking): lots of looping over the same options here – perhaps we could simply:
```
const flatRatePrices = []
const calculatedOptions = []

options.forEach((option) => {
  option.price_type = option.price_type ?? optionsMap.get(option.id)?.price_type
  option.provider_id = option.provider_id ?? optionsMap.get(option.id)?.provider_id

  if (option.price_type === ShippingOptionPriceType.FLAT) {
    flatRatePrices.push(...(option.prices ?? []))
  }

  if (option.price_type === ShippingOptionPriceType.CALCULATED) {
    calculatedOptions.push(option)
  }
})
```

olivermrbl (approved):
Overall LGTM

olivermrbl (approved):

PR #10509, opened 2024-12-09
chore(types): add price types for shipping option endpoints

riqwan:
what:

- adds missing price types for shipping options


olivermrbl (approved):
LGTM

PR #10525, opened 2024-12-10
Feat/tax dml

thetutlage:
Fixes: FRMW-2842

There are zero breaking changes and the users will have to run the migrations


packages/modules/tax/src/migrations/Migration20241212052837.ts
```diff
@@ -0,0 +1,27 @@
+import { Migration } from '@mikro-orm/migrations';
+
+export class Migration20241212052837 extends Migration {
+
+  async up(): Promise<void> {
+    this.addSql('alter table if exists "tax_provider" add column if not exists "created_at" timestamptz not null default now(), add column if not exists "updated_at" timestamptz not null default now(), add column if not exists "deleted_at" timestamptz null;');
```
thetutlage, line 6: These two columns will be created now since they were not defined on the MikroORM model

olivermrbl (approved):
LGTM 💪

PR #10532, opened 2024-12-10
feat(medusa, core-flows, fulfillment): calculate SO price endpoint

fPolic:
**What**
- endpoint + flow for fetching calculated price for a shipping option

---

CLOSES CMRC-777


packages/medusa/src/api/admin/shipping-options/[id]/calculate/route.ts
```diff
@@ -0,0 +1,26 @@
+import { HttpTypes } from "@medusajs/framework/types"
```
Comment (olivermrbl) [inline packages/medusa/src/api/admin/shipping-options/[id]/calculate/route.ts]: q: why is this an admin endpoint?

Comment (fPolic) [reply packages/medusa/src/api/admin/shipping-options/[id]/calculate/route.ts]: should definitely be moved to store 😅

packages/core/types/src/fulfillment/mutations/shipping-option.ts
```diff
@@ -118,3 +118,32 @@ export interface UpdateShippingOptionDTO {
...
+
+  /**
+   * The option data from the provider.
+   */
+  data: Record<string, unknown>
+
+  /**
+   * The calculation context needed for the associated fulfillment provider to calculate the price of a shipping option.
+   */
+  context: {
```
fPolic, line 154: q: is it enough to pass the cart id and the rest of what is needed can be fetched by the provider or should we pass specific customer info, address info etc. (if, for example, a calculated option is used in an RMA flow)

packages/core/types/src/fulfillment/provider.ts
```diff
@@ -53,7 +53,6 @@ export interface IFulfillmentProvider {
    */
   calculatePrice(
     optionData: Record<string, unknown>,
```
olivermrbl: thought: think we still need `data` – I believe the current naming is cause for confusion.

`optionData` – data that was populated when creating the **shipping option**, e.g. `requires_drop_point`. 
`data` – data that is populated when you create the **shipping method**, e.g. `drop_point_id: 1`

For requests that are made prior to adding the shipping method to a cart, the `data` might be empty. 

For example:
```
POST /store/shipping-options/:id/calculate
```

However, when adding the shipping method to cart and the price is calculated, you might need the data for the calculation.

For example:
```
POST /store/carts/:id/shipping-methods
{ option_id: "so_1234", data: { drop_point_id: 1 } }
```

In your implementation of `calculatePrice`, the calculated price might depend on the drop point ID (maybe a farfetched example):
```ts
// pseudo code
calculatePrice() {
  if (drop_point_id) {
    const acceptPackageSizes = await this.fetchDropPointPackageSizes(drop_point_id)

    const price = // one of those package sizes
    return price
  }
}
```

packages/core/utils/src/fulfillment/provider.ts
```diff
@@ -222,7 +222,6 @@ export class AbstractFulfillmentProviderService
    * }
    */
   async calculatePrice(
-    optionData: Record<string, unknown>,
```
olivermrbl: See other comment.

packages/modules/providers/fulfillment-manual/src/services/manual-fulfillment.ts
```diff
@@ -34,7 +34,6 @@ export class ManualFulfillmentService extends AbstractFulfillmentProviderService
   }
 
   async calculatePrice(
-    optionData: Record<string, unknown>,
```
olivermrbl: See other comment.

packages/modules/providers/fulfillment-manual/src/services/manual-fulfillment.ts
```diff
@@ -34,7 +34,6 @@ export class ManualFulfillmentService extends AbstractFulfillmentProviderService
   }
 
   async calculatePrice(
```
shahednasser: suggestion: shouldn't we use `CalculateShippingOptionPriceDTO` for the `data` and `context` for clearer typing?

packages/core/types/src/fulfillment/mutations/shipping-option.ts
olivermrbl: Think we will need more, such as items to calculate prices on product/inventory item dimensions. To be consistent with other providers, we don't want to delegate the responsibility of fetching data from core modules to the provider (yet). This should be a preliminary step and passed to the provider.

Just looping in @srindom – he might have an idea about what we could start with

packages/core/types/src/fulfillment/provider.ts
fPolic: ahh ok, thanks for clarifying - will revert the changes

packages/modules/providers/fulfillment-manual/src/services/manual-fulfillment.ts
fPolic: Good point, there is actually a TODO on top of the file, would prefer to revisit this in a separate PR to fix the entire file.

packages/medusa/src/api/store/shipping-options/[id]/calculate/route.ts
```diff
@@ -0,0 +1,23 @@
+import { HttpTypes } from "@medusajs/framework/types"
+import { MedusaRequest, MedusaResponse } from "@medusajs/framework/http"
+import { calculateShippingOptionsPricesWorkflow } from "@medusajs/core-flows"
+
+import { StoreCalculateShippingOptionPriceType } from "../../validators"
+
+export const POST = async (
+  req: MedusaRequest<StoreCalculateShippingOptionPriceType>,
```
Comment (shahednasser) [inline packages/medusa/src/api/store/shipping-options/[id]/calculate/route.ts]: Can we use an HTTP type here instead? It'll be used by the JS SDK at a later point too.

packages/core/types/src/fulfillment/mutations/shipping-option.ts
olivermrbl: @fPolic to unblock the PR, here's what I think we will definitely need:
- Items, e.g. package dimensions and weight
- Shipping address, e.g. distance from origin to destination

We can add these now and expand in the future in case we need more.

packages/medusa/src/api/store/shipping-options/[id]/calculate/route.ts
```diff
@@ -1,19 +1,34 @@
...
+    entity: "shipping_option",
+    fields: req.remoteQueryConfig.fields,
+    filters: { id: req.params.id },
+  })
+
+  const shippingOption = data[0]
+
+  shippingOption.calculated_price = result[0]?.calculated_amount
+  shippingOption.is_calculated_price_tax_inclusive =
+    result[0]?.is_calculated_price_tax_inclusive
```
Comment (olivermrbl) [inline packages/medusa/src/api/store/shipping-options/[id]/calculate/route.ts]: super nit(non-blocking): We could probably make this change at the fulfillment (provider) interface level to avoid this reshaping – but we can punt that for later

olivermrbl (approved):

packages/medusa/src/api/store/shipping-options/[id]/calculate/route.ts
Comment (fPolic) [reply packages/medusa/src/api/store/shipping-options/[id]/calculate/route.ts]: you mean rename `calculated_amount` to `calculated_price` and use spread?

olivermrbl (approved):

PR #10551, opened 2024-12-11
feat: add support for float properties

thetutlage:
Fixes: FRMW-2844

This PR introduces the support for float values using DML. Under the hood, the value is stored as `numeric` with PostgreSQL and converted to a number using the `Number` serializer.

The float values can be used to represent values with decimal precision. For example:

```ts
const tax = model.define("tax", {
  id: model.number(),
  rate: model.float()
})
```


carlos-r-l-rodrigues (approved):

packages/core/utils/src/dml/helpers/entity-builder/define-property.ts
```diff
@@ -277,6 +279,31 @@ export function defineProperty(
     return
   }
 
+  /**
+   * Handling serial property separately to set the column type
+   */
+  if (field.dataType.name === "float") {
+    Property({
+      columnType: "numeric",
```
olivermrbl: q: shouldn't this be `real` or `double precision` to mirror Postgres?

thetutlage: Double checked it and it seems that you can loose some precision with `real` and `double`. For example: If I run the following query where the `tax_rate_real` is a `REAL` column, then the updated value will be `0.102111116`.

```
UPDATE "public"."foo" SET "tax_rate_real" = 0.102111114;
```

Ofcourse, in our case we don't need these many decimal places, but will be nice to cover them for other use-cases with a more accurate data-type. WDYY?

thetutlage: CleanShot 2024-12-11 at 19 53 38@2x

olivermrbl: Hmm, I thought the argument for not using the `bigNumber` column here and introducing `float` was that we didn't care about high precision for tax rates. If we want to solve for the scenario you describe, would it not be more appropriate to just use `bigNumber`?

thetutlage: I am okay with using `REAL`. Have no strong arguments using `numeric`.

The issue with `bigNumber` was its representation within the JavaScript runtime. Not having the value next to the field and having it inside the `_raw` property as an object.

thetutlage: Done 👍

olivermrbl (approved):

packages/core/utils/src/dml/helpers/entity-builder/define-property.ts
olivermrbl: This looks good to me!

Just to be clear, I think your argument for using `numeric` makes sense–it offers higher precision for the rate of tax rates. That said, we introduced the `bigNumber` property specifically for these use cases and I'd like us to continue using this whenever relevant to ensure consistency and leverage our existing utilities. That’s why I pushed back on adding a new type with somewhat similar characteristics. It’s not that high precision isn’t valid. I hope this rationale makes sense.

Let's go with `real` now and we can revisit later if it leads to issues.

PR #10563, opened 2024-12-11
feat(core-flows): calculate SO price on cart ops

fPolic:
**What**
- calculate the shipping option price when creating a shipping method
- calculate the shipping option price when refreshing cart
- add testing for calculated SO flow
- fix validation on calculated SO creation
- add manual fulfillment provider for testing
- add `from_location` to calculation context

---

RESOLVES CMRC-778
RESOLVES CMRC-602
RESOLVES SUP-136


olivermrbl (approved):
LGTM

PR #10569, opened 2024-12-12
chore(inventory): convert to dml

carlos-r-l-rodrigues:
Fixes: FRMW-2848


packages/modules/inventory/src/models/inventory-item.ts
```diff
@@ -1,156 +1,46 @@
...
+    thumbnail: model.text().nullable(),
+    metadata: model.json().nullable(),
+    location_levels: model.hasMany(() => InventoryLevel, {
+      mappedBy: "inventory_item",
+    }),
+    reservation_items: model.hasMany(() => ReservationItem, {
+      mappedBy: "inventory_item",
+    }),
+    reserved_quantity: model.number().nullable(), // .readonly?
+    stocked_quantity: model.number().nullable(), // .readonly?
```
carlos-r-l-rodrigues: create `.readonly` modifier or similar?

carlos-r-l-rodrigues: @olivermrbl ideally we need a modifier to flag the computed fields here. let me know if we should add it.

olivermrbl: Yeah, I think that's a good idea – can we do `.computed()` maybe?

olivermrbl (approved):

thetutlage (approved):

PR #10493, opened 2024-12-08
fix(carts): Fixes cart modifications not accounting for certain price lists

sergiocampama:
**What**

* Fixes #10490
* Expands any available customer_id into its customer_group_ids for cart
  updates that add line items.

**Why**

* Cart updates from the storefront were overriding any valid price lists
  that were correctly being shown in the storefront's product pages.

**How**

* Adds a new workflow step that expands an optional customer_id into the
  customer_group_ids it belongs to.
* Uses this step in the addToCartWorkflow and
  updateLineItemInCartWorkflow workflows.
* Following a similar approach as in `set-pricing-context.ts` used for the
  `store/products` routes, but that middleware was too specific for carts.

**Testing**
* Using medusa-dev to test on a local backend.
* Adds integration tests for the addToCart and updateLineItemInCart
  workflows.

PR #10579, opened 2024-12-12
feat(medusa,pricing): Cart pricing context with customer group

riqwan:
what:

- cart workflows are updated to now accept cart context as pricing context
- pricing module is updated to handle nested rules
- `customer_group_id` rule is renamed `customer.groups.id` in line with cart shape

Thanks to initial contribution from @sergiocampama - https://github.com/medusajs/medusa/pull/10493

RESOLVES CMRC-745


docloulou:
Well spotted ;)

docloulou:
@riqwan same situation here [Order add line item WF](https://github.com/medusajs/medusa/blob/03803283d68b72c341c3fccf2827eaf04ebc063a/packages/core/core-flows/src/order/workflows/add-line-items.ts#L105C1-L118C6)

riqwan:
@docloulou thanks, there are 10 different workflows in orders that needs to be updated, but that will come in a different PR. 🤞🏻

packages/modules/pricing/src/migrations/Migration20241212190401.ts
```diff
@@ -0,0 +1,11 @@
+import { Migration } from "@mikro-orm/migrations"
+
+export class Migration20241212190401 extends Migration {
+  async up(): Promise<void> {
+    this.addSql(
+      `UPDATE price_list_rule SET attribute = 'customer.groups.id' WHERE attribute = 'customer_group_id';`
```
olivermrbl, line 6: q: Why did we end up renaming the attribute? Feel like this is an awkward key to have

packages/core/core-flows/src/cart/workflows/update-line-item-in-cart.ts
```diff
@@ -4,34 +4,43 @@ import {
...
 import { confirmVariantInventoryWorkflow } from "./confirm-variant-inventory"
 import { refreshCartItemsWorkflow } from "./refresh-cart-items"
 
+const cartFields = cartFieldsForPricingContext.concat(["items.*"])
+
 export const updateLineItemInCartWorkflowId = "update-line-item-in-cart"
 /**
  * This workflow updates a cart's line item.
  */
 export const updateLineItemInCartWorkflow = createWorkflow(
```
olivermrbl, line 25: note to self: Input type change is a breaking change.

olivermrbl (commented):
Overall LGTM, just a comment on the attribute change

packages/modules/pricing/src/migrations/Migration20241212190401.ts
riqwan: The attributes on all our rules based engines are based on what the cart shape looks like - eg: `items.product.id`, `items.subtotal`, etc. This change brings it into alignment. Prevents us from having to mutate the pricing context wherever needed and relying on a singular shape. 

Important to note though that this won't be seen in the presentation layer, for the users they'd just see "Customer Groups"

olivermrbl (approved):
LGTM, the filter utilities are a bit daunting to review, so primarily looked at the tests, and they seem good :)

packages/core/utils/src/common/filter-object-by-keys.ts
```diff
@@ -0,0 +1,86 @@
+import { isDefined } from "./is-defined"
+
+export function filterObjectByKeys(obj, paths) {
```
adrien2p, line 3: suggestion: A description and example here would be nice to have

packages/core/utils/src/common/flatten-object-to-key-value-pairs.ts
```diff
@@ -0,0 +1,116 @@
+type NestedObject = {
+  [key: string]: any
+}
+
+export function flattenObjectToKeyValuePairs(obj: NestedObject): NestedObject {
```
adrien2p, line 5: same suggestion as the previous one

packages/core/utils/src/common/flatten-object-to-key-value-pairs.ts
```diff
@@ -0,0 +1,116 @@
...
+        if (item && typeof item === "object") {
+          Object.keys(item as object).forEach((k) => keys.add(k))
+        }
+      })
+
+      // Process each key
+      keys.forEach((key) => {
+        const values = arrayObjects
+          .map((item) => {
+            if (item && typeof item === "object") {
```
adrien2p, line 90: suggestion: for those type of check would the util `isObject` be more convenient?

packages/core/utils/src/common/flatten-object-to-key-value-pairs.ts
```diff
@@ -0,0 +1,116 @@
...
+              return (item as Record<string, unknown>)[key]
+            }
+            return undefined
+          })
+          .filter((v) => v !== undefined)
+
+        if (values.length > 0) {
+          const newPath = `${pathStr}.${key}`
+          if (values.every((v) => typeof v === "object" && !Array.isArray(v))) {
```
adrien2p, line 99: suggestion: should we use isObject?

adrien2p (approved):

packages/core/utils/src/common/flatten-object-to-key-value-pairs.ts
riqwan: @adrien2p sorry, just saw this. Will open a new PR with these changes to get this PR into the release.

sergiocampama:
thanks @riqwan! do you know when a new version might get released with this fix?

riqwan:
@sergiocampama today!

PR #10587, opened 2024-12-12
feat(dashboard): ability to locate new admin route under existing route

eugenepro2:
This PR add ability to locate new admin route under existing route in sidebar.

For example, new route Brands
image
image

https://github.com/user-attachments/assets/b46b1813-e92e-4b67-84a1-84660023ac7c


packages/admin/dashboard/src/extensions/dashboard-extension-manager/dashboard-extension-manager.tsx
```diff
@@ -117,6 +117,7 @@ export class DashboardExtensionManager {
         to: item.path,
         icon: item.icon ? <item.icon /> : undefined,
         items: [],
+        nested: item.nested,
```
kasperkristensen, line 140: Here it would be good to log a warning if the item has a `nested` field 
but also has children since they will be ignored.

If you create a nested route for `/brand` like this:

```
// /src/admin/routes/brands/page.tsx
import { defineRouteConfig } from "@medusajs/admin-sdk";

const BrandsPage = () => {
  return <div>Brands</div>;
};

    label: "Brands",
    nested: "/products",
})

export default BrandsPage;

// src/admin/routes/brands/create/route.tsx
import { defineRouteConfig } from "@medusajs/admin-sdk";

const CreateBrandPage = () => {
  return <div>Create Brand Page</div>;
};

    label: "Create Brand", // This would normally generate a nested menu item under "Brands".
})

export default CreateBrandPage;
```

The nested MenuItem is created, but never used.

So when looping over the items, it would be great to explicitly warn the user why this item is not being used. We need to keep track of items that have a `nested` field, if we later come across an item which is a child of the item with a `nested` we should skip it, and warn that we are doing so. A little further up in the function you can see we do something similar for other cases.

packages/admin/admin-vite-plugin/src/routes/generate-menu-items.ts
```diff
@@ -17,10 +17,17 @@ import {
 } from "../utils"
 import { getRoute } from "./helpers"
 
+type RouteConfig = {
```
kasperkristensen, line 21: todo: This file needs to be linted

packages/admin/admin-vite-plugin/src/routes/generate-menu-items.ts
```diff
@@ -107,70 +113,58 @@ function generateImport(file: string, index: number): string {
...
-        if (!hasLabel) {
-          return
-        }
+        const hasLabel = hasProperty("label")
+        if (!hasLabel) return
 
-        const hasIcon = properties.some(
-          (prop) =>
-            isObjectProperty(prop) && isIdentifier(prop.key, { name: "icon" })
+        const nested = properties.find(
```
kasperkristensen, line 164: todo: We should validate that the value is one of the allowed ones.

packages/admin/admin-sdk/src/config/types.ts
```diff
@@ -24,6 +24,17 @@ export interface RouteConfig {
...
+  /**
+   * The nested route to display under existing route in the sidebar.
+   */
+  nested?:
+    | "/orders"
+    | "/products"
+    | "/inventory"
+    | "/customers"
+    | "/promotions"
+    | "/price-list"
```
kasperkristensen: It would be good to move this to the admin-shared package both as a const: 

```
export const NESTED_ROUTE_POSITIONS = ["/orders", ...] as const
```
And as a type:
```
export type NestedRoutePosition = (typeof NESTED_ROUTE_POSITIONS)[number]
```

We should then use the const to validate that the user has passed a valid `nested` value in `generate-menu-items.ts`. And use the type in this file, and in the dashboard project.

kasperkristensen (changes requested):
Thanks for the PR @eugenepro2! I have added a few comments, but really great addition which we would love to merge once these are addressed.

packages/admin/dashboard/src/extensions/dashboard-extension-manager/dashboard-extension-manager.tsx
eugenepro2: Good point. But I find out some challange here. The new parameter `nested` passing only on the first route. 
That is means that we cannot check parameter `nested` on the nested page `create`. I hope I described it clear)

Possible solution: locate a new route under existing route by path:
```
// /src/admin/routes/products/brands/page.tsx - located under products route
import { defineRouteConfig } from "@medusajs/admin-sdk";

const BrandsPage = () => {
  return <div>Brands</div>;
};

    label: "Brands",
})

export default BrandsPage;

// /src/admin/routes/products/brands/create/page.tsx - will be ignored
import { defineRouteConfig } from "@medusajs/admin-sdk";

const CreateBrandPage = () => {
  return <div>Create Brand Page</div>;
};

    label: "Create Brand",
})

export default CreateBrandPage;
```

In this case `nested` param will be deleted. 

What do you think about that? Or maybe you have a better solution for that?

eugenepro2: Good point. But I find out some challange here. The new parameter `nested` passing only on the first route. 
That is means that we cannot check parameter `nested` on the nested page `create`. I hope I described it clear)

Possible solution: locate a new route under existing route by path:
```
// /src/admin/routes/products/brands/page.tsx - located under products route
import { defineRouteConfig } from "@medusajs/admin-sdk";

const BrandsPage = () => {
  return <div>Brands</div>;
};

    label: "Brands",
})

export default BrandsPage;

// /src/admin/routes/products/brands/create/page.tsx - will be ignored in sidebar
import { defineRouteConfig } from "@medusajs/admin-sdk";

const CreateBrandPage = () => {
  return <div>Create Brand Page</div>;
};

    label: "Create Brand",
})

export default CreateBrandPage;
```

In this case `nested` param will be deleted and it would be easy to find parent path.

What do you think about that? Or maybe you have a better solution for that?

eugenepro2: I found a more easy solution. We have `menuItems` and we can find nested route there. :)

eugenepro2: Added warning

packages/admin/admin-vite-plugin/src/routes/generate-menu-items.ts
eugenepro2: Added validation

packages/admin/admin-sdk/src/config/types.ts
eugenepro2: Really good point 🚀. Added type.

packages/admin/admin-shared/src/extensions/routes/constants.ts
```diff
@@ -0,0 +1,8 @@
+export const NESTED_ROUTE_POSITIONS = [
+  "/orders",
+  "/products",
+  "/inventory",
+  "/customers",
+  "/promotions",
+  "/price-list",
```
kasperkristensen:
```suggestion
  "/price-lists",
```

kasperkristensen (changes requested):
Thanks for making the changes @eugenepro2, only one little thing to update and then this is good to merge 👍

kasperkristensen (commented):
@eugenepro2 final thing the unit tests are failing because they don't expect the field `nested` for the route items, so if you could update those then we are good to go

eugenepro2:
> @eugenepro2 final thing the unit tests are failing because they don't expect the field `nested` for the route items, so if you could update those then we are good to go

Yep, I'll do that.

kasperkristensen (approved):
LGTM thanks @eugenepro2 🚀

kasperkristensen:
Sorry @eugenepro2, but one more thing, could I get you to remove the changes to the docs, and put those in a separate PR. Otherwise, the changes will go live a soon as we merge this, before the feature is released.

eugenepro2:
> Sorry @eugenepro2, but one more thing, could I get you to remove the changes to the docs, and put those in a separate PR. Otherwise, the changes will go live a soon as we merge this, before the feature is released.

@kasperkristensen Of course. Please check this PR. I'll create a new PR with updated docs for this feature.

kasperkristensen:
> > Sorry @eugenepro2, but one more thing, could I get you to remove the changes to the docs, and put those in a separate PR. Otherwise, the changes will go live a soon as we merge this, before the feature is released.
> 
> @kasperkristensen Of course. Please check this PR. I'll create a new PR with updated docs for this feature.

@eugenepro2 There is still one change in a docs file, which blocks the PR. It's a new line character, if you can remove that then I will make sure this gets merged 👍

eugenepro2:
@kasperkristensen 

> > > Sorry @eugenepro2, but one more thing, could I get you to remove the changes to the docs, and put those in a separate PR. Otherwise, the changes will go live a soon as we merge this, before the feature is released.
> > 
> > 
> > @kasperkristensen Of course. Please check this PR. I'll create a new PR with updated docs for this feature.
> 
> @eugenepro2 There is still one change in a docs file, which blocks the PR. It's a new line character, if you can remove that then I will make sure this gets merged 👍

Oh, Sorry for empty line 😅. Fixed

PR #10599, opened 2024-12-13
breaking: pluralization rule for the info word

thetutlage:
FIXES: FRMW-2847

This PR fixes the pluralization rules for the `info` word at both runtime and the types level.

@olivermrbl I later realized that `VendorLegalInfo` is not a core module, so we should be okay with having a breaking change.


olivermrbl (approved):
LGTM, nice

PR #10617, opened 2024-12-16
chore(fulfillment, utils): Migrate module to DML

adrien2p:
Fixes: FRMW-2838

**What**
- Allow to provide `foreignKeyName` option for hasOne and belongsTo relationships
  - `model.hasOne(() => OtherEntity, { foreignKey: true, foreignKeyName: 'other_entity_something_id' })`
  - The above will also output a generated type that takes into consideration the custom fk name 🔽 
- Update types to account for defined custom foreign key name
- Fix joiner config linkable generation to account for custom linkable keys that provide a public API for their model but are not part of the list of the models included in the MedusaService
  - This was supposed to be handled correctly but the implementation was not considering that custom linkable keys could reference models not part of the one provided to medusa service
- Migrate fulfillment module to DML
- Fix has one with fk behaviour and hooks (the relation should be assigned but not the fk)
- Fix has one belongsTo hooks (the relation should be assigned but not the fk)
- Fix hasOneWithFk and belongsTo non persisted fk to be selectable
- Allow to define `belongsTo` without other side definition for `ManyToOne` with no counter part defined
  - Meaning that if a user defined `belongsTo` on one side andnot mapped by and no counter part on the other entity it will be considered as a `ManyToOne`
- `orphanRemoval` on `OneToOne` have been removed, this means that when assigning a new object relation to an entity, the previous one gets deconected but not deleted automatically. This prevent removing data un volountarely

**NOTE**
As per our convention here are some information to keep in mind

**HasOne <> BelongsTo**
Define `OneToOne`, The foreign key is owned by the belongs to and the relation needs to be provided to cascade if wanted

**HasMany <> BelongsTo**
Define `OneToMane` <> `ManyToOne`, the foreign key is owned by the many to one and for those relation no cascade will be performed, the foreign key must be provided. For the `HasMany` the cascade is available

**HasOne (with FK)**
Will act similarly to belongs to with **HasOne <> BelongsTo**


packages/modules/fulfillment/src/services/fulfillment-module-service.ts
```diff
@@ -1378,6 +1414,13 @@ export default class FulfillmentModuleService
               updatedRuleIds.push(rule.id)
             }
 
+            // @ts-ignore
+            delete rule.created_at
+            // @ts-ignore
+            delete rule.updated_at
+            // @ts-ignore
+            delete rule.deleted_at
```
olivermrbl, line 1421: question: What's this for?

packages/modules/fulfillment/src/services/fulfillment-module-service.ts
```diff
@@ -422,8 +442,10 @@ export default class FulfillmentModuleService
       | FulfillmentTypes.CreateShippingOptionDTO[]
       | FulfillmentTypes.CreateShippingOptionDTO,
     @MedusaContext() sharedContext: Context = {}
-  ): Promise<ShippingOption[]> {
-    const data_ = Array.isArray(data) ? data : [data]
+  ): Promise<InferEntityType<typeof ShippingOption>[]> {
+    const data_ = Array.isArray(data)
+      ? data.map((d) => structuredClone(d))
```
olivermrbl: question: Don't we typically use our own deep copy util?

packages/modules/fulfillment/src/models/shipping-option.ts
```diff
@@ -1,182 +1,42 @@
...
+    }),
+    shipping_profile: model
+      .belongsTo(() => ShippingProfile, {
+        mappedBy: "shipping_options",
+      })
+      .nullable(),
+    provider: model.belongsTo(() => FulfillmentProvider).nullable(),
+    type: model.hasOne(() => ShippingOptionType, {
+      foreignKey: true,
+      foreignKeyName: "shipping_option_type_id",
```
olivermrbl, line 30: thought(non-blocking): Think it's fine to keep this option, but maybe we should actually rename the foreign key to `type_id` .. think this inconsistency is a bit off

packages/modules/fulfillment/src/models/service-zone.ts
```diff
@@ -1,126 +1,60 @@
...
+      unique: true,
+      where: "deleted_at IS NULL",
+    },
+  ])
+  .cascades({
+    delete: ["geo_zones", "shipping_options"],
+  }) as unknown as DmlEntity<
+  DMLEntitySchemaBuilder<ServiceZoneSchema>,
+  "ServiceZone"
+>
```
olivermrbl, line 60: question: This looks a bit funky, what's it for?

packages/modules/fulfillment/integration-tests/__tests__/fulfillment-module-service/shipping-option.spec.ts
```diff
@@ -1059,7 +1072,7 @@ moduleIntegrationTestRunner<IFulfillmentModuleService>({
             )
 
             const types = await service.listShippingOptionTypes()
-            expect(types).toHaveLength(2)
+            expect(types).toHaveLength(4)
```
olivermrbl, line 1075: question: Why did this change? Don't see any changes to seed data or similar 😄

olivermrbl (commented):
Nice work guys, overall LGTM 💪 

Just had a few questions, but nothing blocking

packages/modules/fulfillment/src/services/fulfillment-module-service.ts
adrien2p: This is some this that needs more investigation. For now mikro orm doesn't handle those type properly for some reasons with our current config and we get something like this Data.value error which is very weird and only happen during commit but otherwise all the assign and validation passes first 🤔

packages/modules/fulfillment/src/models/shipping-option.ts
adrien2p: We discussed that to not create breaking changes to the current api (types, links, graphql, etc). But up to you just let us know

packages/modules/fulfillment/src/models/service-zone.ts
adrien2p: When there is too many inferences typescript reach the maximum issue and you have to provide a type yourself in the middle to reduce/break the complexity of the overall inference so that typescript can continue from there. Otherwise you have to ignore and completely loose typings which defeats the purpose

packages/modules/fulfillment/integration-tests/__tests__/fulfillment-module-service/shipping-option.spec.ts
adrien2p: Dml was applying orphan removal on one to one which is pretty aggressive, since dml is completely locked up and do not provide such configuration, it seems dangerous to physically remove resources that are not removed by the user but as a side effect under the hood. So if you attach a new resource, the previous one is detached and the new one created and attached. We have similar behavior in other places as well. But with the dml we enforce the convention everywhere and align everything 👌

olivermrbl: Sounds good – is this something we should mention in the release notes?

packages/modules/fulfillment/src/models/service-zone.ts
olivermrbl: Cool, thx for explaining.

packages/modules/fulfillment/src/models/shipping-option.ts
olivermrbl: Yeah I see, fine to keep it as is

packages/modules/fulfillment/src/services/fulfillment-module-service.ts
olivermrbl: Alright, but we haven't had this issue before correct?

adrien2p: yes it is new and I dont know why it is happening,I will try to investigate one more time but otherwise I would suggest we move forward for now wdyt?

packages/modules/fulfillment/integration-tests/__tests__/fulfillment-module-service/shipping-option.spec.ts
adrien2p: I ve added everything in the pr description, let me know if you feel something is missing

packages/modules/fulfillment/src/services/fulfillment-module-service.ts
olivermrbl: Yeah, agree – not worth spending much time on right now.

packages/modules/fulfillment/integration-tests/__tests__/fulfillment-module-service/shipping-option.spec.ts
olivermrbl: Yeah think this looks good. In the worst case, developers will end up with unassociated resources, which I think is better than mistakenly deleted them.

olivermrbl (approved):

adrien2p:
@olivermrbl can I put the automerge flag here?

olivermrbl:
Yes 💯

PR #10622, opened 2024-12-16
feat(core-flows,dashboard,js-sdk,medusa,types): support Fulfillment Options

fPolic:
**What**
- add a list point for fetching fulfillment options for a provider
- add FO support on SO create & update on dashboard
- pass `cart` and `stockLocation` to `validateFufillmentData` context

---

CLOSES CMRC-789
CLOSES CMRC-790


packages/medusa/src/api/admin/fulfillment-providers/[id]/options/route.ts
```diff
@@ -0,0 +1,27 @@
...
+} from "@medusajs/framework/http"
+
+export const GET = async (
+  req: AuthenticatedMedusaRequest,
+  res: MedusaResponse<HttpTypes.AdminFulfillmentProviderOptionsListResponse>
+) => {
+  const fulfillmentProviderService = req.scope.resolve(Modules.FULFILLMENT)
+
+  const fulfillmentOptions =
+    await fulfillmentProviderService.retrieveFulfillmentOptions(req.params.id)
```
Comment (olivermrbl) [inline packages/medusa/src/api/admin/fulfillment-providers/[id]/options/route.ts:18]: though(non-blocking): I believe we will need to think about the pagination of options at some point, I recall this being an issue in V1. For example, Webshipper would have a default limit of 30 options, and some of our users had more than 30 services set up. This is probably an edge case so we can deal with it separately.

packages/admin/dashboard/src/routes/locations/location-service-zone-shipping-option-edit/components/edit-region-form/edit-shipping-option-form.tsx
```diff
@@ -96,12 +116,20 @@ export const EditShippingOptionForm = ({
...
+    )
+
     await mutateAsync(
       {
         name: values.name,
         price_type: values.price_type,
         shipping_profile_id: values.shipping_profile_id,
         provider_id: values.provider_id,
+        data: {
+          ...(shippingOption.data ?? {}),
```
olivermrbl: thought: Wouldn't this mean we are merging the data of two fulfillment options? That seems incorrect to me. 

Separately, perhaps we should consider disallowing editing the fulfillment option of a shipping option once created. This would avoid concerns like this. We had this in V1. Instead, you would need to create a new option with the other fulfillment option configured.

packages/admin/dashboard/src/routes/locations/location-service-zone-shipping-option-create/components/create-shipping-options-form/create-shipping-options-form.tsx
```diff
@@ -58,6 +60,16 @@ export function CreateShippingOptionsForm({
     resolver: zodResolver(CreateShippingOptionSchema),
   })
 
+  const selectedProviderId = useWatch({
+    control: form.control,
+    name: "provider_id",
+  })
+
+  const { fulfillment_options: fulfillmentProviderOptions } =
```
olivermrbl, line 68: todo: I believe we need to account for the return flow–right now I am getting the same options returns for regular and return options

packages/medusa/src/api/admin/fulfillment-providers/[id]/options/route.ts
Comment (fPolic) [reply packages/medusa/src/api/admin/fulfillment-providers/[id]/options/route.ts:18]: will open a ticket to track this

packages/admin/dashboard/src/routes/locations/location-service-zone-shipping-option-edit/components/edit-region-form/edit-shipping-option-form.tsx
fPolic: Yes, data would be merged. Will disable edit then.

packages/admin/dashboard/src/routes/locations/location-service-zone-shipping-option-create/components/create-shipping-options-form/create-shipping-options-form.tsx
fPolic: good catch, fixed 👍

packages/core/core-flows/src/cart/workflows/add-shipping-method-to-cart.ts
```diff
@@ -80,14 +80,15 @@ export const addShippingMethodToCartWorkflow = createWorkflow(
             provider_id: shippingOption?.provider_id,
             option_data: shippingOption?.data ?? {},
             method_data: inputOption.data ?? {},
+            from_location: shippingOption?.stock_location ?? {},
```
olivermrbl: nit: I believe this should be part of the `context`, alongside the cart data to be consistent with how this is otherwise used

olivermrbl (approved):
Overall LGTM, one final comment.

@kasperkristensen, if you could find time to glance over the UI, that would be great :)

packages/core/core-flows/src/cart/workflows/add-shipping-method-to-cart.ts
fPolic: so you would pass a Record<option_id, stockLocaition> in context since we can have different locations for each option?

olivermrbl: I think I would move the context to be a part of the array, instead of being a "global" context for all options:
```
export interface ValidateShippingMethodsDataInput {
  options_to_validate: {
    id: string
    provider_id: string
    option_data: Record<string, unknown>
    method_data: Record<string, unknown>
    context: Record<string, unknown>
  }[]
}

const validatedData = await promiseAll(
  options_to_validate.map(async (option) => {
    const validated = await fulfillmentModule.validateFulfillmentData(
      option.provider_id,
      option.option_data,
      option.method_data,
      option.context
    )

    return {
      [option.id]: validated,
    }
  })
)
```

What do you think?

olivermrbl: IIRC, we pass the `from_location` and `...cart` in the context to other methods of the fulfillment methods right?

fPolic: ahh the other way around, yea think this is cleaner, pushed a change 👍

packages/admin/dashboard/src/routes/locations/location-service-zone-shipping-option-create/components/create-shipping-options-form/create-shipping-option-details-form.tsx
```diff
@@ -52,6 +58,10 @@ export const CreateShippingOptionDetailsForm = ({
       })),
   })
 
+  useEffect(() => {
+    form.setValue("fulfillment_option_id", undefined)
+  }, [selectedProviderId])
+
```
kasperkristensen: todo: this is missing a dependency, and we should set the value to "". To avoid using a useEffect, we could also update the provider_id select, to set `fulfillment_option_id` to `""` in the onChange.

kasperkristensen (approved):
UI LGTM, just one small thing to fix

PR #10626, opened 2024-12-16
fix(cart): tax rate data type

carlos-r-l-rodrigues:
FIXES: FRMW-2850

### DIFF
--- a/packages/modules/cart/src/migrations/Migration20241205095237.ts
+++ b/packages/modules/cart/src/migrations/Migration20241205095237.ts
@@ -51,9 +51,6 @@ export class Migration20241205095237 extends Migration {
     )
 
     this.addSql(
-      'alter table if exists "cart_line_item_tax_line" alter column "rate" type integer using ("rate"::integer);'
-    )
-    this.addSql(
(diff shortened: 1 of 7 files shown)


packages/modules/cart/src/migrations/Migration20241205095237.ts
```diff
@@ -50,9 +50,6 @@ export class Migration20241205095237 extends Migration {
       'CREATE INDEX IF NOT EXISTS "IDX_cart_line_item_adjustment_item_id" ON "cart_line_item_adjustment" (item_id) WHERE deleted_at IS NULL;'
     )
 
-    this.addSql(
-      'alter table if exists "cart_line_item_tax_line" alter column "rate" type integer using ("rate"::integer);'
-    )
     this.addSql(
```
carlos-r-l-rodrigues, line 53: We need to remove this from previous migration to not convert existing entries into integers

olivermrbl: Nice catch

olivermrbl (approved):

PR #10630, opened 2024-12-17
feat(inventory,dashboard,types,core-flows,js-sdk,medusa): Improve inventory UX

kasperkristensen:
**@medusajs/dashboard**
- Add new bulk editors for managing Inventory: One one the product details page, and another on the inventory item list page.
- Improves performance of DataGrid when performing actions on a large number of cells, e.g. selecting a column of 100 rows and copy/paste, delete, etc.
- Allow bidirectional actions in the DataGrid component when `multiColumnSelection` is true. This allows the users to drag-to-fill, copy/paste, etc, across all columns at once. The feature needs to be enabled with the above flag, as it only makes sense for DataGrids where we can guarantee that all columns are of the same type. Will also enable this for the various price related DataGrids in a follow up PR.
- Adds global "Pending Navigation" UI. We use loaders throughout the admin, to preload the data needed for a page before we render it. Before we would not provide any feedback to the user that this was happening which would make the UI seem unresponsive, if a loader took a noticeable amount of time to resolve. We now dim the content and show a progress bar at the top of the screen, which indicates to the user that a navigation is in progress. This was added as fetching the data needed for the Inventory data grid could take a while to resolve. As we plan to move to more SSR in the future, this pattern will be an even bigger UX improvement.
- Replaces `framer-motion` dependency with `motion`, as the former was deprecated after `motion` was separated from the Framer organization.

**@medusajs/core-flows**
- Adds new batch workflow for creating, updating and deleting location levels.
- Deprecates old create/delete location level workflow

**@medusajs/medusa**
- Adds new `POST /admin/inventory-items/location-levels/batch` endpoint using the above mentioned workflow.
- Updates `POST /admin/inventory-items/:id/location-levels/batch` endpoint to use the above mentioned workflow.
  - The includes two breaking change: To delete a location level you no longer pass the `ID` of the stock location of said location level. Instead you need to pass the ID of the location level. The body of the endpoint has been changes from `{ creates: [], deletes: []}` to `{ create: [], update: [], delete: [] }` to bring it inline with other batch endpoints. As an example an accepted payload went from `{ deletes: ["sloc_123"] }` to `{ delete: ["ilvl_123"] }`.

**@medusajs/js-sdk**
- Adds two new methods `batchInventoryItemsLocationLevels` and `batchInventoryItemLocationLevels`. Notice that one is the plural form of InventoryItems. The two methods correspond to `POST /admin/inventory-items/location-levels/batch` and `POST /admin/inventory-items/:id/location-levels/batch`.
- Deprecates `batchUpdateLevels` in favour of `batchInventoryItemLocationLevels`

**@medusajs/types**
- Adds new Inventory Item types, and some cleanup of existing ones.

**@medusajs/ui**
- Updates the styling of the Switch, RadioGroup, and Checkbox components to the new style guidelines for disabled inputs.

Resolves CMRC-784, CMRC-786, CMRC-787


packages/admin/dashboard/src/routes/products/product-stock/types.ts
```diff
@@ -0,0 +1,9 @@
+import { HttpTypes } from "@medusajs/types"
+
+// TODO: Create a type in the types package
```
olivermrbl: todo?

olivermrbl (approved):
🐐🐐

packages/admin/dashboard/src/routes/products/product-stock/types.ts
kasperkristensen: Fixed! 👍

olivermrbl:
Shall we merge this?

PR #10636, opened 2024-12-17
feat(order, types): Add Credit Line to order module

riqwan:
what:

- adds credit line model
- adds credit line change order
- adds credit line summary total


packages/modules/order/src/services/__tests__/util/actions/credit-line-add.spec.ts
```diff
@@ -0,0 +1,134 @@
...
+      "current_order_total": 30,
+      "pending_difference": 30,
+      "difference_sum": 0,
+      "paid_total": 0,
+      "refunded_total": 0,
+      "credit_line_total": 0
+    }
+
+    Upon adding a credit line, the order total and the pending difference will increase making it possible for the merchant
+    to request the customer for a payment for an arbitrary reason, or prepare the order balance sheet to then allow
```
olivermrbl, line 55: question: I wouldn't have expected credit lines to be used for additional payments, only refunds–can you expand on this?

riqwan: Good catch, this was incorrectly written - I had changed this in my current branch. Will port over the changes in a bit.

`pending_difference` is what the customer needs to pay to settle the difference. If a credit line has been issued, the `credit_line_total` needs to be subtracted from `pending_difference` to accurately depict what should happen, which is not happening in this branch.

olivermrbl: That sounds right yeah

olivermrbl: Do you want to update this branch or do it in a follow up?

riqwan: Can do either. Making good progress rn, so focusing on not hindering that flow. 😄

packages/modules/order/src/models/credit-line.ts
```diff
@@ -0,0 +1,107 @@
...
+    columnType: "text",
+    nullable: true,
+  })
+  reference: string | null = null
+
+  @Property({
+    columnType: "text",
+    nullable: true,
+  })
+  reference_id: string | null = null
```
olivermrbl, line 66: question: How will these references be used in practise?

olivermrbl (approved):

packages/modules/order/src/models/credit-line.ts
riqwan: Its the resource that will be the enabler of this operation, in this case payment-collection and its id.

We can scope it down to payments or refunds, but that would require creating multiple credit lines, which I can't see a benefit for.

PR #10650, opened 2024-12-18
fix(dashboard): order details status

fPolic:
**What**
- disallow cancelling already canceled order
- display payment status alongside fulfillment status in the "general header"

---

FIXES CMRC-758


olivermrbl (approved):
nice

PR #10660, opened 2024-12-18
fix(core-flows): data passed to to fulfillment provider context

fPolic:
**What**
- consolidate data sent to calculate SO price context
- pass `from_location` to `validateFulfillmentData`
  - don't set stock location on a SO in `listShippingOptionsForCartWorkflow` and set it in `listShippingOptionsForCartWithPricingWorkflow`


packages/core/core-flows/src/cart/workflows/list-shipping-options-for-cart-with-pricing.ts
```diff
@@ -61,11 +61,20 @@ export const listShippingOptionsForCartWithPricingWorkflow = createWorkflow(
         "sales_channel_id",
         "currency_code",
         "region_id",
-        "shipping_address.city",
-        "shipping_address.country_code",
-        "shipping_address.province",
-        "shipping_address.postal_code",
+        "shipping_address.*",
+        // ensure that the same subset of fileds needed to calculate shipping options is retrieved here and in calculateShippingOptionsPricesWorkflow when fetching cart
```
olivermrbl: thought(non-blocking): We could create a shared const for these and spread them in the existing fields arrays

olivermrbl (approved):

PR #10673, opened 2024-12-19
fix(core-flows): refresh payment collections upon shipping changes

riqwan:
what:

- refresh payment collections upon shipping changes

RESOLVES CMRC-819


packages/core/core-flows/src/cart/workflows/add-shipping-method-to-cart.ts
```diff
@@ -149,16 +148,8 @@ export const addShippingMethodToCartWorkflow = createWorkflow(
...
-      input: {
-        cart_id: input.cart_id,
-      },
-    })
-
-    updateCartPromotionsWorkflow.runAsStep({
-      input: {
-        cart_id: input.cart_id,
-      },
+    refreshCartItemsWorkflow.runAsStep({
```
olivermrbl, line 151: thought: At some point, we should probably change this to be called `refreshCartWorkflow`–it does a whole bunch of stuff unrelated to items 😄

olivermrbl (approved):

PR #10674, opened 2024-12-19
fix(): Workflow cancellation + gracefully handle non serializable state

adrien2p:
FIXES FRMW-2852

**What**
A workflow distributed transaction expect any response and error to be serializable. When it is not the case, the distributed transaction might fail during the save checkpoint that occurs for async steps. This can lead to unexpected behaviour.

With this pr, we introduce a way to handle non serialazable object in a more sustainable manner, this means the following:

- If a workflow throw any non serialazable error (e.g AWS error that contains full IncomingMessage object that related to network communication, think of req/res) then we identify that this object is not serialzable and we clean up the object to make it serializable without loosing the main information, add a new error to the workflow to informed of this issue and can be handled by the user.
- If a response is not serializable (which should not happen at this point because it is handled before by the value resolver), in that case, we wont be able to reuse that response to continue the workflow which means that the workflow is in a non runnable state. In that case we throw a specific error stating that a non serializable context is being provided

**second what**
This pr refactor the `runAsStep` to add better support for workflow cancelation, especially async ones


carlos-r-l-rodrigues (approved):

packages/core/orchestration/src/transaction/distributed-transaction.ts
```diff
@@ -320,6 +316,76 @@ class DistributedTransaction extends EventEmitter {
...
+            }
+            nonSerializableErrors.push({
+              ...error,
+              error: e,
+            })
+          }
+        }
+
+        if (nonSerializableErrors.length) {
+          this.errors.push(...nonSerializableErrors)
```
sgirones, line 374: shouldn't the existing non-serialisable errors be removed?

adrien2p: We keep it but we cherry pick useful info so that you still have access to it if needed

olivermrbl (approved):

PR #10677, opened 2024-12-19
feat(core-flows,fulfillment, fulfillment-manual, types): make fulfillment typings more specific

fPolic:
**What**
- attempt to add more specific type definitions around fulfillment provider APIs

---

CLOSES CMRC-814
CLOSES CMRC-816
CLOSES CMRC-817


olivermrbl (approved):
Overall LGTM

nit: I think we shouldn't be mixing types from different domains, e.g. the fulfillment service interface should not use the `OrderDTO`, but instead a standalone type that happens to have properties shared with the other type.

fPolic:
> nit: I think we shouldn't be mixing types from different domains

@olivermrbl could you check the last commit

olivermrbl (approved):

PR #10697, opened 2024-12-23
fix: pluralization of words ending in y, where y follows a vowel

thetutlage:
Fixes: FRMW-2851


shahednasser (approved):
LGTM, I believe you need to add a changeset right?

thetutlage:
> LGTM, I believe you need to add a changeset right?

Ohh yes. Thanks for the reminder 👍

PR #10700, opened 2024-12-23
fix(fulfillment): export schema types

shahednasser:
Export `ServiceZoneSchema` and `GeoZoneSchema` so that we can infer types correctly in references


thetutlage (approved):
LGTM!

PR #10737, opened 2024-12-26
feat(medusa): Add health endpoint for all running modes of medusa

sradevski:
When running the service in eg. K8s, we need a way to check the readiness of the service, regardless if it runs in server, shared, or worker mode. This allows for that, as previously we only registered the health endpoint in shared or server mode.

### DIFF
--- a/packages/core/framework/src/http/express-loader.ts
+++ b/packages/core/framework/src/http/express-loader.ts
@@ -95,10 +95,6 @@ export async function expressLoader({ app }: { app: Express }): Promise<{
   // Currently we don't allow configuration of static files, but this can be revisited as needed.
   app.use("/static", express.static(path.join(baseDir, "static")))
 
-  app.get("/health", (req, res) => {
-    res.status(200).send("OK")
-  })
-
   const shutdown = async () => {
     redisClient?.disconnect()
   }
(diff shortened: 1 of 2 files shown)


thetutlage (approved):
Yeah. Looks good to me, given we already start the express server and the `/health` endpoint just returns a 200 OK.

thetutlage:
@sradevski A quick question.

Do we use the `/health` endpoint to check if we should send traffic to this specific server? If yes, then it might cause some issues. For example:

- The `start` command will register the `/health` endpoint and respond with 200.
- The loaders are not yet ready (because booting the app might take a couple of seconds)
- The infra starts sending traffic to this new instance and it won't respond until the entire boot process is finished.

WDYT?

sradevski:
@thetutlage yes I agree, as I wrote in the comment, just sending a 200 isn't ideal due to the scenario you described. 

However, I do think at this point everything should be loaded though, right? If we were to eg. start the server first, and then eg. run the loaders it would be a problem.

thetutlage:
> However, I do think at this point everything should be loaded though, right? If we were to eg. start the server first, and then eg. run the loaders it would be a problem.

Yeah, I think you are right. We load everything and then we call `app.listen`, so we are all sorted then 👍

PR #10771, opened 2024-12-30
fix(pricing): pricing context calculations only takes into account existing rule attributes

riqwan:
what:

- scopes pricing calculations on the context by registered rule attributes


riqwan:
@olivermrbl this snapshot worked for those who had the error

olivermrbl (approved):
LGTM

PR #10791, opened 2025-01-02
feat: generate modules mappings at runtime

thetutlage:
Fixes: FRMW-2693

This PR adds support for generating the container mappings types for modules at runtime by inspecting the registered/enabled modules. As a result of it, the `container.resolve` method will suggest auto-complete for modules that are enabled via the `medusa-config.ts` file (including the custom modules)


packages/medusa/src/commands/start.ts
```diff
@@ -120,21 +121,29 @@ async function start(args: {
...
+        if (gqlSchema) {
+          const outputDirGeneratedTypes = path.join(directory, ".medusa/types")
+          await gqlSchemaToTypes({
+            outputDir: outputDirGeneratedTypes,
+            filename: "remote-query-entry-points",
+            interfaceName: "RemoteQueryEntryPoints",
+            schema: gqlSchema,
+            joinerConfigs: MedusaModule.getAllJoinerConfigs(),
+          })
+          logger.info("Generated modules types")
```
olivermrbl: since: Since we write these on every server restart, these logs clutter things up a bit. Do you think we should remove them or use debug mode?

olivermrbl (approved):
This works well 💪

packages/medusa/src/commands/start.ts
thetutlage: Sure. That makes sense. I will switch from logger.info to logger.debug

PR #10807, opened 2025-01-03
feat: pin dependencies to specific version

LukasGerm:
Hello,

I noticed that if you want to pin your medusa to a specific version and do a fresh install (without a lock file) you will get incorrect versions of the dependencies. 

e.g when I did 

```
pnpm why @medusajs/cart
Legend: production dependency, optional only, dev only

medusa@0.0.1 /Users/lukasgermerott/dev/tarabao/apps/medusa

dependencies:
@medusajs/medusa 2.0.4
└── @medusajs/cart 2.1.3
```

this was causing some weird runtime issues and mismatches.

This PR pins all the packages to the specific versions so that if in the future you want to install this version specific version that also all the peer dependencies will get the correct version.


LukasGerm:
Not stale. I think this is still valid.

adrien2p:
Thank you for you contribution 💪 but we merged another pr for that and missed this one sorry 😅

PR #10829, opened 2025-01-06
feat(auth-google,auth-github): Allow passing a custom callbackUrl to …

sradevski:
This is necessary so we can support different login pages based on the actor. This is also an intended approach to implementing oauth, where the client can control the callbackUrl (but the allowed redirects are limited within the Oauth app configuration)

For example, you might have a `user/github/callback` page and a `customer/github/callback` page, and where you redirect to would depend on which actor is doing the authentication.

The PR also adds the `state` parameter as defined in the Oauth protocol.

Note: If you have overridden the auth module definition in your medusa config, you will need to pass `Modules.CACHE` as a dependency. This can be considered a breaking change due to that fact.

We can also make it non-breaking by not storing the state and always returning the default callbackUrl as the state value. This reduces the security of the implementation, but can be an acceptable fallback.

### DIFF
--- a/packages/modules/providers/auth-github/integration-tests/__tests__/services.spec.ts
+++ b/packages/modules/providers/auth-github/integration-tests/__tests__/services.spec.ts
@@ -118,6 +118,17 @@ describe("Github auth provider", () => {
     })
   })
 
+  it("returns a custom redirect_uri on authenticate", async () => {
+    const res = await githubService.authenticate({
+      body: { callback_url: "https://someotherurl.com" },
+    })
+    expect(res).toEqual({
+      success: true,
+      location:
+        "https://github.com/login/oauth/authorize?redirect_uri=https%3A%2F%2Fsomeotherurl.com&client_id=test&response_type=code",
+    })
+  })
+
   it("validate callback should return an error on empty code", async () => {
     const res = await githubService.validateCallback(
       {
--- a/packages/modules/providers/auth-github/integration-tests/__tests__/services.spec.ts
+++ b/packages/modules/providers/auth-github/integration-tests/__tests__/services.spec.ts
@@ -20,6 +20,20 @@ const sampleIdPayload = {
 }
 
 const baseUrl = "https://someurl.com"
+const callbackUrl = "https%3A%2F%2Fsomeurl.com%2Fauth%2Fgithub%2Fcallback"
+
+let state = {}
+const defaultSpies = {
+  retrieve: jest.fn(),
+  create: jest.fn(),
+  update: jest.fn(),
+  setState: jest.fn().mockImplementation((key, value) => {
+    state[key] = value
+  }),
+  getState: jest.fn().mockImplementation((key) => {
+    return Promise.resolve(state[key])
+  }),
+}
 
 // This is just a network-layer mocking, it doesn't start an actual server
 const server = setupServer(
@@ -29,7 +43,7 @@ const server = setupServer(
       const url = request.url
       if (
         url ===
-        "https://github.com/login/oauth/access_token?client_id=test&client_secret=test&code=invalid-code&redirect_uri=https%3A%2F%2Fsomeurl.com%2Fauth%2Fgithub%2Fcallback"
+        `https://github.com/login/oauth/access_token?client_id=test&client_secret=test&code=invalid-code&redirect_uri=${callbackUrl}`
       ) {
         return new HttpResponse(null, {
           status: 401,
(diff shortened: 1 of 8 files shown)


packages/modules/providers/auth-google/src/services/google.ts
```diff
@@ -85,11 +88,9 @@ export class GoogleAuthService extends AbstractAuthModuleProvider {
       return { success: false, error: "No code provided" }
     }
 
-    const params = `client_id=${this.config_.clientId}&client_secret=${
-      this.config_.clientSecret
-    }&code=${code}&redirect_uri=${encodeURIComponent(
-      this.config_.callbackUrl
-    )}&grant_type=authorization_code`
+    // We can add redirect_uri=${encodeURIComponent(this.config_.callbackUrl)} here as well for enhanced security, although effect is minimal due to having to preconfigure allowed redirect URLs in Google.
```
sradevski: Agh, it seems Google requires it, even though it is optional in the oauth spec. I'll look into implemeting all of this properly now :)

sradevski:
I have already done some e2e tests and things are working as expected.

olivermrbl (approved):
LGTM

packages/modules/providers/auth-github/integration-tests/__tests__/services.spec.ts
```diff
@@ -20,6 +20,20 @@ const sampleIdPayload = {
 }
 
 const baseUrl = "https://someurl.com"
+const callbackUrl = "https%3A%2F%2Fsomeurl.com%2Fauth%2Fgithub%2Fcallback"
```
adrien2p: suggestion: would it be easier to do `new URLSearchParams('https://someurl.com/auth/github/callback').toString()` instead of adding the code which makes the URL hard to read, wdyt? If needed, you can also pass all query params if needed later on so that the transformation is done automatically

packages/modules/providers/auth-github/src/services/github.ts
```diff
@@ -84,9 +91,12 @@ export class GithubAuthService extends AbstractAuthModuleProvider {
       return { success: false, error: "No code provided" }
     }
 
-    const params = `client_id=${this.config_.clientId}&client_secret=${
-      this.config_.clientSecret
-    }&code=${code}&redirect_uri=${encodeURIComponent(this.config_.callbackUrl)}`
+    const state = await authIdentityService.getState(req.query?.state as string)
```
adrien2p: Q: instead of having optional chaining everywhere, would it make sense to start the function with
```ts
const query: Record<string, string> = req.query ?? {}
const body: Record<string, string> = req.body ?? {}
```

packages/modules/providers/auth-github/src/services/github.ts
```diff
@@ -192,16 +202,17 @@ export class GithubAuthService extends AbstractAuthModuleProvider {
...
-
+    const stateKeyParam = `state=${stateKey}`
     const authUrl = new URL(
       `https://github.com/login/oauth/authorize?${[
         redirectUrlParam,
         clientIdParam,
         responseTypeParam,
+        stateKeyParam,
       ].join("&")}`
     )
```
adrien2p: suggestion: What do you think of the following?

```suggestion
    const stateKeyParam = `state=${stateKey}`
    const authUrl = new URL('https://github.com/login/oauth/authorize')
    authUrl.searchParams.append('redirect_uri',  encodeURIComponent(callbackUrl)) 
    authUrl.searchParams.append('client_id',  clientId). 
    authUrl.searchParams.append('responseTypeParam',  'code')
    authUrl.searchParams.append('state',  stateKey)
```

if you agree and that works for you then we can cleanup those variables above as well

packages/modules/providers/auth-google/src/services/google.ts
```diff
@@ -175,18 +182,20 @@ export class GoogleAuthService extends AbstractAuthModuleProvider {
...
     const scopeParam = "scope=email+profile+openid"
+    const stateKeyParam = `state=${stateKey}`
 
     const authUrl = new URL(
       `https://accounts.google.com/o/oauth2/v2/auth?${[
         redirectUrlParam,
         clientIdParam,
         responseTypeParam,
         scopeParam,
+        stateKeyParam,
```
adrien2p: suggestion: all the same as for githuib :)

adrien2p (approved):
few suggestions otherwise LGTM

packages/modules/providers/auth-github/integration-tests/__tests__/services.spec.ts
sradevski: Yeah good call, I did `encodeURIComponent("https://someurl.com/auth/github/callback")` instead.

packages/modules/providers/auth-github/src/services/github.ts
sradevski: Yup sounds good 👍

PR #10831, opened 2025-01-06
chore: rename remote-query-entry-points.d.ts to query-entry-points.d.ts

thetutlage:
Fixes: FRMW-2804


packages/medusa/src/commands/start.ts
```diff
@@ -127,17 +127,23 @@ async function start(args: {
       })
 
       if (generateTypes) {
+        const typesDirectory = path.join(directory, ".medusa/types")
+
+        /**
+         * Cleanup existing types directory before creating new artifacts
+         */
+        await new FileSystem(typesDirectory).cleanup({ recursive: true })
```
thetutlage, line 135: As a hygiene, we should always remove existing files and then create new artifacts.

carlos-r-l-rodrigues (approved):

packages/medusa/src/commands/start.ts
adrien2p: I think thats a good practice, but should we instead let each generator to clean up their own file to be sure we don't have other generated content that gets removed by mistake and not generated at the same time or from the same place?

adrien2p (approved):

packages/medusa/src/commands/start.ts
thetutlage: Yeah. That will be nice too. Right now, I was thinking of this flow

- The `.medusa/types` directory is dedicated to writing generated output.
- We cleanup this directory before we generate any new output.
- Then dedicated actions will generate their output with clean slate.

But I agree, if one of those actions does not run under a certain condition, then the whole cleanup strategy won't work. For now, that's not the case, but I am happy to tweak the logic if you think we should address that right now 👍

olivermrbl (approved):

PR #10835, opened 2025-01-06
chore(medusa): Add handler path to the http tracing to be able to group by

adrien2p:
RESOLVES FRMW-2835

In this PR, we trace HTTP requests using the route pattern and not the request URL. This allows us to aggregate all HTTP requests under a single route.

In terms of implementation, we have to self find the route for a given request, since there is no API in express to do the same and we begin tracing even before the request is handed over to Express.

Since, the route matching happens via RegExp matches, we ensure that this does not add much performance overhead. The matching takes between 0.8ms-1.5ms for various different routes

### DIFF
--- a/.changeset/itchy-walls-call.md
+++ b/.changeset/itchy-walls-call.md
@@ -1,0 +1,5 @@
+---
+"@medusajs/medusa": patch
+---
+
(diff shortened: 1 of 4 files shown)


thetutlage (approved):

adrien2p:
should we get that one merged guys?

PR #10847, opened 2025-01-07
Update page.mdx

DivineStudio:
Updated the if-else example condition for when input.is_active is false.


shahednasser (approved):
Nice catch thanks!

PR #10858, opened 2025-01-07
fix(dashboard): cancel order notifications

fPolic:
**What**
- add proper notifications when canceling an order

---

FIXES CMRC-824

### DIFF
--- a/packages/admin/dashboard/src/i18n/translations/$schema.json
+++ b/packages/admin/dashboard/src/i18n/translations/$schema.json
@@ -258,7 +258,11 @@
               "type": "string"
             }
           },
-          "required": ["header_one", "header_other", "description"],
+          "required": [
+            "header_one",
+            "header_other",
+            "description"
+          ],
           "additionalProperties": false
         }
       },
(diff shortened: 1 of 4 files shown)


packages/admin/dashboard/src/i18n/translations/$schema.json
```diff
@@ -258,7 +258,11 @@
               "type": "string"
             }
           },
-          "required": ["header_one", "header_other", "description"],
+          "required": [
```
olivermrbl, line 261: @kasperkristensen, this file keeps getting updated because of linting 😄 

Not a big deal, but anything we can do here? It clutters the PRs a bit

packages/admin/dashboard/src/i18n/translations/en.json
```diff
@@ -912,6 +912,7 @@
     "exchange": "Exchange",
     "return": "Return",
     "cancelWarning": "You are about to cancel the order {{id}}. This action cannot be undone.",
+    "orderCancelled": "Order cancelled successfully",
```
olivermrbl: todo: should use the american version of cancel
```suggestion
    "orderCanceled": "Order canceled successfully",
```

olivermrbl (approved):
LGTM, one nit

PR #10869, opened 2025-01-07
feat(admin-bundler,admin-vite-plugin,medusa): Add support for loading Admin Extensions from plugins

kasperkristensen:
Should not be merged before https://github.com/medusajs/medusa/pull/10895

**What**
- Introduces a new `plugin` command to `admin-bundler`, currently not used anywhere but will be called from `medusa build:plugin`
- Discovers plugins with extensions and add passes the to `admin-vite-plugin`.
- Updates `admin-vite-plugin` so its able to read built admin extensions.

Resolves CMRC-830, CMRC-839


olivermrbl (approved):
LGTM, looping in @medusajs/framework in case they want to have a look at it.

packages/admin/admin-bundler/src/lib/plugin.ts
```diff
@@ -0,0 +1,57 @@
+import { readFileSync } from "fs"
+import { glob } from "glob"
+import path from "path"
+import { UserConfig } from "vite"
+
+export async function plugin() {
+  const vite = await import("vite")
+  const entries = await glob("src/admin/**/*.{ts,tsx,js,jsx}")
```
adrien2p, line 8: NOTE: since node 22 we have the native glob API `https://nodejs.org/api/fs.html#fsglobpattern-options-callback` maybe we can wait since we also support node 20 😅 but I wanted to mention it for reference

adrien2p (approved):

PR #10874, opened 2025-01-08
feat: remove dead code and refactor the logic of resolving plugins

thetutlage:
Fixes: FRMW-2857

In this PR, I have cleaned up the existing logic of resolving plugins to support plugins imports from relative paths and npm packages. 

Also, we expect plugins to have a `package.json` file and mimic the exact directory structure of a Medusa App.


packages/core/types/src/common/config-module.ts
```diff
@@ -922,10 +922,50 @@ export type ConfigModule = {
...
+  }
+>
+
 export type PluginDetails = {
   resolve: string
   name: string
   id: string
   options: Record<string, unknown>
   version: string
+  modules?: InputConfig["modules"]
```
thetutlage, line 970: Introduced the `modules` property on the plugin detail . Later we will merge these modules with the main config top-level modules

packages/medusa/src/loaders/helpers/resolve-plugins.ts
```diff
@@ -1,148 +1,81 @@
...
-  return "src"
-}
-
-/**
- * Load plugin details from a path. Return undefined if does not contains a package.json
- * @param pluginName
- * @param path
- * @param includeExtensionDirectoryPath should include src | dist for the resolved details
- */
-function loadPluginDetails({
```
thetutlage, line 26: Merged the logic with `resolvePlugin` method

packages/medusa/src/loaders/helpers/resolve-plugins.ts
```diff
@@ -1,148 +1,81 @@
...
  */
-function resolvePlugin(pluginName: string): {
-  resolve: string
-  id: string
-  name: string
-  options: Record<string, unknown>
-  version: string
-} {
-  if (!isAbsolute(pluginName)) {
-    let resolvedPath = path.resolve(`./plugins/${pluginName}`)
```
thetutlage, line 75: When using `require.resolve`, we shouldn't worry about absolute or relative paths separately.

packages/medusa/src/loaders/helpers/resolve-plugins.ts
```diff
@@ -1,148 +1,81 @@
...
-      id: createPluginId(packageJSON.name),
-      name: packageJSON.name,
-      options: {},
-      version: packageJSON.version,
-    }
-  } catch (err) {
-    throw new Error(
-      `Unable to find plugin "${pluginName}". Perhaps you need to install its package?`
-    )
+async function resolvePlugin(
```
olivermrbl, line 51: suggestion(non-blocking): This is super nitpicky, but should we keep the guided error message in case the plugin cannot be found? I realise it will now fail with "Cannot find module", which is sensible, so I am also OK keeping it as is. The other one might make it even more obvious where to look.

olivermrbl (approved):
LGTM

packages/medusa/src/loaders/helpers/resolve-plugins.ts
thetutlage: I agree. Let's capture the error and rethrow it with reworded message

packages/core/utils/src/common/read-dir-recursive.ts
```diff
@@ -2,21 +2,51 @@ import { Dirent } from "fs"
...
+  }
+}
 
-    for (const entry of entries) {
-      const fullPath = join(dir, entry.name)
-      Object.defineProperty(entry, "path", {
-        value: dir,
-      })
-      allEntries.push(entry)
+export async function readDirRecursive(
```
adrien2p, line 24: NOTE: Is there any difference with the native `readdir` function with the recursive options set to true? If not then maybe we can eventually switch to the netive impl?

adrien2p (approved):
LGTM 🎉

packages/core/utils/src/common/read-dir-recursive.ts
thetutlage: Yup. Let's do that in a separate PR

carlos-r-l-rodrigues: There are differences from Node version to Node version. That's why I created the helper function. The loader was breaking in some versions

Issue #10852, opened 2025-01-07
[Bug]: Cannot type a custom refund amount, it always shows as NaN.


Attachai77:
### Package.json file

```JSON
{
	"name": "App",
	"version": "1.0.1",
	"description": "A starter for Medusa projects.",
	"author": "Medusa (https://medusajs.com)",
	"license": "MIT",
	"keywords": [
		"sqlite",
		"postgres",
		"typescript",
		"ecommerce",
		"headless",
		"medusa"
	],
	"scripts": {
		"build": "medusa build",
		"seed": "medusa exec ./src/scripts/seed.ts",
		"migrations": "npx medusa db:migrate",
		"sync": "npx medusa db:sync-links",
		"start": "medusa start",
		"dev": "medusa develop",
		"build-start": "medusa build && medusa start",
		"test": "jest",
		"test:unit": "jest",
		"test:watch": "jest --watch",
		"test:coverage": "jest --coverage",
		"check:changed": "node ./scripts/check-changed.js",
		"format:changed": "node ./scripts/check-changed.js format",
		"lint:changed": "node ./scripts/check-changed.js lint",
		"postinstall": "patch-package"
	},
	"dependencies": {
		"@medusajs/admin-sdk": "2.1.3",
		"@medusajs/api-key": "2.1.3",
		"@medusajs/auth": "2.1.3",
		"@medusajs/auth-emailpass": "2.1.3",
		"@medusajs/auth-google": "2.1.3",
		"@medusajs/cache-inmemory": "2.1.3",
		"@medusajs/cache-redis": "2.1.3",
		"@medusajs/cart": "2.1.3",
		"@medusajs/cli": "2.1.3",
		"@medusajs/currency": "2.1.3",
		"@medusajs/customer": "2.1.3",
		"@medusajs/dashboard": "2.1.3",
		"@medusajs/event-bus-local": "2.1.3",
		"@medusajs/event-bus-redis": "2.1.3",
		"@medusajs/file": "2.1.3",
		"@medusajs/file-local": "2.1.3",
		"@medusajs/file-s3": "2.1.3",
		"@medusajs/framework": "2.1.3",
		"@medusajs/fulfillment": "2.1.3",
		"@medusajs/fulfillment-manual": "2.1.3",
		"@medusajs/inventory": "2.1.3",
		"@medusajs/medusa": "2.1.3",
		"@medusajs/notification": "2.1.3",
		"@medusajs/notification-local": "2.1.3",
[...]

PR #10888, opened 2025-01-09
fix(dashboard, core-flows): improvements to order page on canceled orders

riqwan:
what:

- Remove pending payment for canceled orders
- Hide unfulfilled items for canceled orders
- Disable non refundable payments from being refunded
- Populate refund created_by
- Disable order edit when canceled
- Fix bug https://github.com/medusajs/medusa/issues/10852

RESOLVES CMRC-842


olivermrbl (approved):
Tested with #10923 included and it works 👍

PR #10895, opened 2025-01-09
feat: Merge plugin modules

thetutlage:
Fixes: FRMW-2858

This PR merge the modules exported by the plugins with the modules defined within the user config. As a result, all modules get loaded without changing the internals of the loader.

However, you cannot disable the module of a plugin by re-adding it to the `modules` array. That is something we should handle separately. 

We've added the breaking change label because of the following fix:
We did broke the ability to completely disable modules in the past pr's, in this pr we re add the ability to disable a module and that this modules does not get loaded at all. ([here](https://github.com/medusajs/medusa/pull/10895/commits/6dd164f783d632c927e0a48c50090dcf7387ccdd))


olivermrbl (approved):

packages/core/utils/src/common/__tests__/transform-modules.spec.ts
```diff
@@ -0,0 +1,52 @@
...
+      {
+        resolve: MODULE_PACKAGE_NAMES[Modules.CACHE],
+      },
+      {
+        resolve: MODULE_PACKAGE_NAMES[Modules.CACHE],
+        disable: true,
+      },
+    ])
+
+    expect(modules).toEqual({})
```
adrien2p: Q: if a module is disabled, I thought we expected `{ cache: false }`, the idea is that since some modules configuration can have `defaultPackage`

packages/core/utils/src/common/define-config.ts
```diff
@@ -108,14 +109,86 @@ export function defineConfig(config: InputConfig = {}): ConfigModule {
...
+      }
+    }
+
+    acc[serviceName] = moduleConfig
+
+    return acc
+  }, {})
+
+  // Remove any modules set to false
+  Object.keys(remappedModules).forEach((key) => {
```
adrien2p: NOTE: depends on my previous comment

adrien2p (approved):
Just a little comment that we can handle here or separately if you prefer

packages/core/utils/src/common/__tests__/transform-modules.spec.ts
thetutlage: Fixed in a commit by you :)

PR #10900, opened 2025-01-09
fix(modules): Fix miss leading provider resolution error

adrien2p:
FIXES SUP-560

**What**
Currently, no matter the error when looking for a provider to exists in the container we are throwing a normalized error stating that the provider does not exists in the container. The issue is that the first initialization of the provider occurs the first time we resolve it, and the error can be that the provider failed to be instanciated for any reason. 

In this pr, we are explicitly checking for the error to be an awilix resolution error to throw the classic error and otherwise we provide the issue why the provider failed to be resolved.


packages/modules/auth/src/services/auth-provider.ts
```diff
@@ -31,11 +31,16 @@ export default class AuthProviderService {
     try {
       return this.dependencies[`${AuthProviderRegistrationPrefix}${providerId}`]
     } catch (err) {
-      const errMessage = `
-      Unable to retrieve the auth provider with id: ${providerId}
-      Please make sure that the provider is registered in the container and it is configured correctly in your project configuration file.
-      `
+      if (err.name === "AwilixResolutionError") {
```
sradevski, line 34: NIT: Is it possible or better to do `err instanceof AwilixResolutionError`? This also works, that check is just safer typing-wise if the error object is available

sradevski (approved):
LGTM, just a NIT

packages/modules/auth/src/services/auth-provider.ts
adrien2p: I do agree, I am just always careful with instanceof because depending where the instance vs the object is coming from it will return false 😂 but if you prefer I can change it.

sradevski: Ah yes true, I remember once this being an issue. I think this is fine, thanks JS 😄

olivermrbl (approved):

PR #10902, opened 2025-01-09
feat(core-flows, dashboard, link-modules,medusa, types, utils): fulfillment shipping changes

fPolic:
**What**
- product <> shipping profile link
- create and update product workflows/endpoints accepts shipping profile
- pass shipping option id when creating fulfillment to allow overriding customer selected SO
- validate shipping profile delete
- dashboard
  - set shipping profile on product create
  - manage shipping profile for a product
  - **update the create fulfillment form**
- other
  - fix create product form infinite rerenders
 
---

CLOSES CMRC-831 CMRC-834 CMRC-836 CMRC-837 CMRC-838 CMRC-857 TRI-761


packages/medusa/src/api/admin/products/validators.ts
```diff
@@ -232,6 +232,7 @@ export const CreateProduct = z
     options: z.array(CreateProductOption).optional(),
     variants: z.array(CreateProductVariant).optional(),
     sales_channels: z.array(z.object({ id: z.string() })).optional(),
+    shipping_profile_id: z.string().optional(),
```
fPolic: q: should we require a `shipping_option_id` to be passed to the create product endpoint, also should it be required for the create product workflows?

packages/admin/dashboard/src/routes/products/product-shipping-profile/components/product-organization-form/product-shipping-profile-form.tsx
```diff
@@ -0,0 +1,127 @@
...
+export const ProductShippingProfileForm = ({
+  product,
+}: ProductShippingProfileFormProps) => {
+  const { t } = useTranslation()
+  const { handleSuccess } = useRouteModal()
+
+  // TODO
+  // const { getFormConfigs, getFormFields } = useDashboardExtension()
+  // const configs = getFormConfigs("product", "shipping-profile")
+  // const fields = getFormFields("product", "shipping-profile")
```
olivermrbl: thought: Think we can remove this

packages/core/core-flows/src/order/workflows/create-fulfillment.ts
```diff
@@ -350,8 +350,11 @@ export const createOrderFulfillmentWorkflow = createWorkflow(
       return { data: data.shipping_methods?.[0]?.data }
     })
 
-    const shippingOptionId = transform(order, (data) => {
-      return data.shipping_methods?.[0]?.shipping_option_id
+    const shippingOptionId = transform({ order, input }, (data) => {
+      return (
+        data.input.shipping_option_id ??
+        data.order.shipping_methods?.[0]?.shipping_option_id
```
olivermrbl, line 363: question: This could lead to unexpected results if the order has multiple shipping methods, so maybe we should force it in the input, i.e. validate if it is there?

packages/modules/link-modules/src/definitions/product-shipping-profile.ts
```diff
@@ -0,0 +1,72 @@
+import { ModuleJoinerConfig } from "@medusajs/framework/types"
+import { LINKS, Modules } from "@medusajs/framework/utils"
+
+export const ProductShippingProfile: ModuleJoinerConfig = {
```
olivermrbl, line 4: thought(non-blocking): Can we start using the `defineLink` API instead of these?

olivermrbl: cc @carlos-r-l-rodrigues–can't remember if we are ready for this in the core?

packages/medusa/src/migration-scripts/migrate-product-shipping-profile.ts
```diff
@@ -0,0 +1,55 @@
+import { ExecArgs } from "@medusajs/framework/types"
+import { ContainerRegistrationKeys, Modules } from "@medusajs/framework/utils"
+
+export default async function assignProductsToShippingProfile({
```
olivermrbl: thought: We should probably create a workflow for this to ensure data consistency in the face of errors

packages/medusa/src/api/admin/products/validators.ts
olivermrbl: Yeah, I think we should require them for both. If you don't pass a shipping profile id, the product is consequently not associated with a shipping profile, and that could lead to issues in other workflows, e.g. fulfillments.

olivermrbl: We can consider adding a `is_default` to shipping profiles later and integrate that, but I think it's OK to require this for now.

packages/core/core-flows/src/product/workflows/update-products.ts
```diff
@@ -131,6 +141,25 @@ function findProductsWithSalesChannels({
   return !input.update?.sales_channels ? [] : productIds
 }
 
+function findProductsWithShippingProfile({
```
olivermrbl: If we require shipping profile on product creation all will have it, so we can save ourselves from this loop

packages/core/core-flows/src/order/workflows/create-fulfillment.ts
fPolic: Should we require a shipping option if the item doesn't require shipping?
Perhaps we allow undefined but throw if multiple shipping methods available?

packages/medusa/src/migration-scripts/migrate-product-shipping-profile.ts
fPolic: Moved this to a workflow but kept it in the same file, wdyt? Not sure it is reusable business logic that should live in `core-flows`?

olivermrbl:
@adrien2p, could I get you to glance over this PR? 🙏

packages/admin/dashboard/src/routes/orders/order-create-fulfillment/components/order-create-fulfillment-form/order-create-fulfillment-form.tsx
```diff
@@ -56,30 +59,89 @@ export function OrderCreateFulfillmentForm({
...
+    name: "location_id",
+    control: form.control,
+  })
+
   const { stock_locations = [] } = useStockLocations()
 
+  const { shipping_options = [], isLoading: isShippingOptionsLoading } =
+    useShippingOptions({
+      stock_location_id: selectedLocationId,
+      // is_return: false, // TODO: 500 when enabled
```
adrien2p, line 84: Q: is this todo still happening? is yes, have you been able to identify the reason?

packages/admin/dashboard/src/routes/orders/order-create-fulfillment/components/order-create-fulfillment-form/order-create-fulfillment-form.tsx
```diff
@@ -56,30 +59,89 @@ export function OrderCreateFulfillmentForm({
...
+      },
+      {} as Record<string, string | null>
+    )
+
+    const payload: HttpTypes.AdminCreateOrderFulfillment = {
+      no_notification: !data.send_notification,
+      items: Object.entries(data.quantity)
+        .filter(
+          ([id, value]) =>
+            !!value && itemShippingProfileMap[id] === selectedShippingProfileId
```
adrien2p, line 132: Q: what happen to the one's where the selected shipping profile defer here or the value is 0? I suppose when quantity is 0 we don't really care 😅 but for the other one are they handle some how differently? Cause I thought seeing that we check already if there is discrepancy between the select sp and the actual one.

adrien2p: Also, on a separate note, I havnt reached the backend yet but I suppose we also perform those checks right? Cause the API can be called independently from the client or another API

packages/admin/dashboard/src/routes/orders/order-create-fulfillment/components/order-create-fulfillment-form/order-create-fulfillment-form.tsx
```diff
@@ -89,15 +151,32 @@ export function OrderCreateFulfillmentForm({
...
+        if (shippingOption) {
+          const locationId =
+            shippingOption.service_zone.fulfillment_set.location.id
+
+          form.setValue("location_id", locationId)
+          form.setValue(
+            "shipping_option_id",
+            initialShippingOptionId || undefined
+          )
+        } // else -> TODO: what if original shipping option is deleted?
```
adrien2p, line 169: Q: I believe it should not be possible to hard delete shipping options that are already in use. So maybe we should get the soft deleted one as well and in that case that cant happen? I am not sure though but I wanted to throw it out loud to get your thoughts.

Now a question come to my mind, if a user selected a shipping option before it gets deleted. then I believe the user should still benefit from it, but that might be a business choice probably, right?

packages/admin/dashboard/src/routes/orders/order-create-fulfillment/components/order-create-fulfillment-form/order-create-fulfillment-form.tsx
```diff
@@ -124,14 +203,21 @@ export function OrderCreateFulfillmentForm({
       })
     }
 
-    const quantityMap = itemsToFulfill.reduce((acc, item) => {
-      acc[item.id] = getFulfillableQuantity(item as OrderLineItemDTO)
-      return acc
-    }, {} as Record<string, number>)
+    const quantityMap = itemsToFulfill.reduce(
```
adrien2p, line 199: Q: We already created a similar map previously (l:62) for the form, can't we create it once and then use it for both?

packages/core/core-flows/src/order/workflows/create-fulfillment.ts
adrien2p: if we allow undefined and throw if multiple sm exists, doesn't that looks like requiring a sm anyway for the items?

packages/core/core-flows/src/product/workflows/create-products.ts
```diff
@@ -182,6 +188,30 @@ export const createProductsWorkflow = createWorkflow(
...
+            return {
+              [Modules.PRODUCT]: {
+                product_id: createdProduct.id,
+              },
+              [Modules.FULFILLMENT]: {
+                shipping_profile_id: data.input.products[i].shipping_profile_id,
+              },
+            }
+          })
+          .filter((i) => Boolean(i))
```
adrien2p:
```suggestion
        return data.createdProducts
          .filter((product, i) => data.input.products[i].shipping_profile_id)
          .map((createdProduct, i) => {
            return {
              [Modules.PRODUCT]: {
                product_id: createdProduct.id,
              },
              [Modules.FULFILLMENT]: {
                shipping_profile_id: data.input.products[i].shipping_profile_id,
              },
            }
          })
```

packages/core/types/src/http/product/admin/payloads.ts
```diff
@@ -210,10 +210,14 @@ export interface AdminCreateProduct {
    * The ID of the product's collection.
    */
   collection_id?: string
+  /**
+   * The ID of the product's shipping profile.
+   */
+  shipping_profile_id: string
```
adrien2p, line 216: Q: I ve seen that it is handled as optional in the create product workflow, is this intentional to be required from the API?

packages/core/types/src/workflows/products/mutations.ts
```diff
@@ -33,6 +33,10 @@ export type CreateProductWorkflowInputDTO = Omit<
    * The sales channels that the product is available in.
    */
   sales_channels?: { id: string }[]
+  /**
+   * The product's shipping profile.
+   */
+  shipping_profile_id: string
```
adrien2p, line 39: Q: same question as the previous one

packages/modules/link-modules/src/definitions/product-shipping-profile.ts
adrien2p: I think we probably are and if not maybe a few changes would be needed, but let see separately maybe?

adrien2p (dismissed):
Great work, I was just wondering something, I ve seen a lot of validation on the client side, which I can understand to display the right information to the user, but I don't think I ve seen the same level of validation on the server, is it intentional or did I just missed it 😂 ? Do we expect a user calling the API to get the same level of validation or do we just expect the user to validate whatever they want for there use case if they decide to be more or less flexible?

packages/admin/dashboard/src/routes/orders/order-create-fulfillment/components/order-create-fulfillment-form/order-create-fulfillment-form.tsx
fPolic: We want to create fulfillemnt only for products whose shipping profile matches the selected profile in the dropdown. Items that don't match the shipping profile are disabled in the UI so when a user creates fulfillment they are aware that the fulfillment is created only for those products. 

We do not validate currently if the fulfillment shipping option satisfies the shipping profile of the items but perhaps we should add this in the core flow as well? cc @olivermrbl

packages/admin/dashboard/src/routes/orders/order-create-fulfillment/components/order-create-fulfillment-form/order-create-fulfillment-form.tsx
fPolic: didn't investigate yet, will look into it or open a ticket

packages/core/types/src/http/product/admin/payloads.ts
fPolic: good catch, should be [mandatory](https://github.com/medusajs/medusa/pull/10902#discussion_r1922395161) in API and the workflow

packages/admin/dashboard/src/routes/orders/order-create-fulfillment/components/order-create-fulfillment-form/order-create-fulfillment-form.tsx
olivermrbl: That seems reasonable to me. I believe we recently added similar validation in the cart-completion flow, perhaps we can reuse some of that.

packages/admin/dashboard/src/routes/orders/order-create-fulfillment/components/order-create-fulfillment-form/order-create-fulfillment-form.tsx
fPolic: Yea, it depends on how store operators manage it. Ideally, it can be hidden from the store but used to fulfill remaining orders with that option. I don't see an issue with deleting a shipping option after orders with that option have been fulfilled.

packages/admin/dashboard/src/routes/orders/order-create-fulfillment/components/order-create-fulfillment-form/order-create-fulfillment-form.tsx
fPolic: added validation here: https://github.com/medusajs/medusa/pull/10902/commits/bc5fb9942622a01c276ed51e80ca0667fffe59a6

olivermrbl (approved):

PR #10904, opened 2025-01-10
chore: move build utilities to Compiler class

thetutlage:
Fixes: FRMW-2866


packages/core/framework/src/build-tools/compiler.ts
```diff
@@ -0,0 +1,366 @@
...
+    const emitResult = program.emit()
+    const diagnostics = ts
+      .getPreEmitDiagnostics(program)
+      .concat(emitResult.diagnostics)
+
+    /**
+     * Log errors (if any)
+     */
+    if (diagnostics.length) {
+      console.error(
```
adrien2p, line 181: Q: should we use the logger here?

adrien2p (approved):
Nice cleanup :D

packages/core/framework/src/build-tools/compiler.ts
thetutlage: Yes 👍

carlos-r-l-rodrigues (approved):

olivermrbl (approved):

PR #10924, opened 2025-01-13
docs: update breaking changes for google auth storefront

ranjithkumar8352:
Medusa v2.2.0 has a breaking change where the `state` must also be passed to the auth callback.

This is not mentioned in the required actions for breaking changes in the release notes. I spent more than an hour trying to understand why the upgrade was breaking Google auth login even though I followed the breaking changes release notes and the docs.

Please update it in the release notes as well if possible - https://github.com/medusajs/medusa/releases/tag/v2.2.0

### DIFF
--- a/www/apps/resources/app/storefront-development/customers/third-party-login/page.mdx
+++ b/www/apps/resources/app/storefront-development/customers/third-party-login/page.mdx
@@ -254,7 +254,7 @@ export default function GoogleCallback() {
   }, [])
 
   const state = useMemo(() => {
-    const queryParams = new URLSearchParams(window.location.state)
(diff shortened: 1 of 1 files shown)


shahednasser (approved):
Thanks for your contribution! I just checked with the team and looks like the issue isn't with `state` in particular, but you have to pass all query parameters from the original callback (so that also includes `error` too, for example). And this was the case from before 2.2.0, but we didn't document it correctly.

I'll accept your PR but will open another one to mention this in other areas in the docs, where necessary. Thanks!

ranjithkumar8352:
Thanks for the clarification @shahednasser

PR #10926, opened 2025-01-13
Feat/plugin develop

thetutlage:
Fixes: FRMW-2865

In this PR we add support for developing a plugin in watch mode. During the file change, we re-compile the source code (incrementally), publishes the package, and updates the installations of the plugin. 

We are using `yalc` under the hood and it must be installed as a dev dependency in the plugin project and the main Medusa app.


packages/cli/medusa-cli/src/create-cli.ts
```diff
@@ -254,6 +254,18 @@ function buildLocalCommands(cli, isLocalProject) {
         })
       ),
     })
+    .command({
+      command: "plugin:develop",
+      desc: "Start plugin development process in watch mode. Changes will be re-published to the local packages directory",
```
adrien2p:
```suggestion
      desc: "Start plugin development process in watch mode. Changes will be re-published to the local packages registry",
```

wdyt?

adrien2p (approved):

PR #10934, opened 2025-01-13
docs: remove redundant body parameter in fetch customer

ranjithkumar8352:
Remove redundant body parameter while fetching customer in third-party login docs


shahednasser (approved):
Nice catch thanks!

PR #10950, opened 2025-01-14
feat(promotion,dashboard,types,utils,medusa): Add statuses to promotions

riqwan:
what:

- adds a status column to promotion table
- introduce active promotion query
- scope revert, register and compute actions to active promotions
- admin to create and update promotion with statuses

RESOLVES CMRC-845
RESOLVES CMRC-846
RESOLVES CMRC-847
RESOLVES CMRC-848
RESOLVES CMRC-849
RESOLVES CMRC-850


packages/admin/dashboard/src/lib/promotions.ts
```diff
@@ -1,31 +1,49 @@
...
   ACTIVE = "ACTIVE",
-  DISABLED = "DISABLED",
+  INACTIVE = "INACTIVE",
+  DRAFT = "DRAFT",
+}
+
+export type StatusColors = "grey" | "orange" | "green" | "red" | "grey"
+export type StatusMap = Record<string, [StatusColors, string]>
+export const promotionStatusMap: StatusMap = {
+  [PromotionStatus.ACTIVE]: ["orange", i18n.t("statuses.active")],
```
olivermrbl: todo: Should be green

packages/core/core-flows/src/promotion/workflows/update-promotions-status.ts
```diff
@@ -0,0 +1,55 @@
...
+  updatePromotionsStatusWorkflowId,
+  (input: UpdatePromotionsStatusWorkflowInput) => {
+    updatePromotionsValidationStep(input)
+
+    const updatedPromotions = updatePromotionsStep(input.promotionsData)
+
+    const promotionStatusUpdated = createHook("promotionStatusUpdated", {
+      promotions: updatedPromotions,
+      additional_data: input.additional_data,
+    })
```
olivermrbl, line 49: q: Why was this added? The `updatePromotionsWorkflow` already has a hook.

riqwan: It was a comment from Seb on the RFC where he wanted a workflow that you can hook into when status changes

olivermrbl: Ah alright, I hadn't seen. Would have thought this could be dealt with by the existing hook, but I am fine keeping it as is.

riqwan: Same, this workflow doesn't add much overhead, so just created one.

packages/admin/dashboard/src/lib/promotions.ts
riqwan: @olivermrbl addressed the cache and color issues 🤞🏻

olivermrbl (approved):

PR #10960, opened 2025-01-14
feat(core, medusa, cli): Enable migration scripts

adrien2p:
RESOLVES SUP-602
RESOLVES SUP-603

**What**
Enable the ability to create migration data scripts and execute them as part of the `db:migrate`, also allow to skip running the migration scripts using `--skip-scripts`.

**How**
We are looking for any file (except barrel files) inside the `migration-scripts` directory in `medusa`, the user `project` and the `plugins`.
We expect the files to export a default function as follow `export default function migration1({ container }: { container: MedusaContainer }) => Promise<void>`

The migration flow is the following
- Ensure the `script_migrations` table exists otherwise create it
- Look for pending migration scripts (All scripts that have not been ran previously)
- take a lock
- loop
  - insert script migration that will be executed
  - execute script
  - update finished_at
- release lock


packages/core/framework/src/migrations/migrator.ts
```diff
@@ -0,0 +1,157 @@
...
+
+  async insertMigration(records: Record<string, any>[]): Promise<void> {
+    try {
+      const values = records.map((record) => Object.values(record))
+      const columns = Object.keys(records[0])
+
+      await this.pgConnection.raw(
+        `INSERT INTO ${this.migration_table_name} (${columns.join(
+          ", "
+        )}) VALUES ${values.map((value) => `(${value.join(", ")})`).join(", ")}`
```
carlos-r-l-rodrigues: should we add params to the `raw`, so it doesn't break the SQL in case of the file name constains some `'` on it?

carlos-r-l-rodrigues (approved):
LGTM

packages/core/framework/src/migrations/migrator.ts
adrien2p: oh true I completely forgot 😂 I ll do it right now

packages/medusa/src/commands/db/run-scripts.ts
```diff
@@ -0,0 +1,112 @@
...
+
+    const medusaAppResources = await new MedusaAppLoader().load()
+    onApplicationPrepareShutdown =
+      medusaAppResources.onApplicationPrepareShutdown
+    onApplicationShutdown = medusaAppResources.onApplicationShutdown
+    await medusaAppResources.onApplicationStart()
+
+    const scriptsSourcePaths = [
+      join(dirname(require.resolve("@medusajs/medusa")), "migration-scripts"),
+      ...plugins.map((plugin) => join(plugin.resolve, "migration-scripts")),
```
olivermrbl, line 68: comment: I don't love "two-word folders", so we should find a better name, however, I don't have any suggestions myself right now 😄

packages/core/framework/src/migrations/migrator.ts
adrien2p: @carlos-r-l-rodrigues let me know what you think [about it](https://github.com/medusajs/medusa/pull/10960/commits/7158b91c2b68f0fc6631df2586dab0cc159a13f5)

adrien2p:
@shahednasser let me know if I should add more info or anything for the doc if we do the doc :)

olivermrbl (approved):
Overall LGTM

fPolic (approved):
🔥

olivermrbl:
> @shahednasser let me know if I should add more info or anything for the doc if we do the doc :)

Initially, we should focus on documenting how this is used in the core to run cross-module migration scripts, rather than what the user can do with this feature. Both are relevant, but the former is more important, since the scripts from our core will run without the user knowing.

packages/medusa/src/commands/db/run-scripts.ts
olivermrbl: Anyway, let's merge this now to unblock @fPolic's PR. We can revisit this later.

olivermrbl:
@adrien2p, feel free to add `automerge` when tests are passing.

packages/medusa/src/commands/db/run-scripts.ts
adrien2p: yeah it is not perfect I agree that it feels strange :)

adrien2p:
@olivermrbl I suppose the CLI pipeline is failing but its ok right?

PR #10974, opened 2025-01-15
feat(core-flow): add promotion created event

Attachai77:
- Add emitting `promotion.created` once promotion created


Attachai77:
I think I can consume that event by using hook

PR #10975, opened 2025-01-15
feat: add support for loading admin extensions from the source

thetutlage:
Fixes: FRMW-2874

Loading admin extensions from the source helps in achieving HMR with Vite for the extensions.


kasperkristensen (approved):
LGTM!

packages/core/framework/src/build-tools/compiler.ts
```diff
@@ -440,6 +463,7 @@ export class Compiler {
    * a file has changed.
    */
   async developPluginBackend(onFileChange?: () => void) {
+    await this.#createPluginOptionsFile()
```
kasperkristensen, line 466: @thetutlage just making sure: Can we guarantee that this file will never end up in the final build of the plugin which is published to NPM, or would it be safer to clean this up when the dev process is killed? If it ends up in the final NPM package, admin extensions would fail to load for the plugin.

thetutlage: I see what you mean. Ideally, I expect the users to run `npm run build` at the time of publishing the plugin to npm. And that command will perform a cleanup + build and this file will never part of the published file.

PR #10982, opened 2025-01-15
chore: Use local CLI version in CI

olivermrbl:
**What**

Use `npx` to resolve the locally installed version of `@medusajs/cli` instead of the global one

**Why**

The version mismatch between the globally installed CLI package and the local packages in the project was causing an issue related to loading `medusa-config.ts`. The exact reason for this is still unclear, but this should at least unblock merging PRs. 

We can tackle a larger clean-up of the workflow separately.

---

Sidenote: Using `npx` brings another improvement that is the changes made to `@medusajs/cli` in a PR will now be tested in the CI instead of the global package.

### DIFF
--- a/.github/workflows/test-cli-with-database.yml
+++ b/.github/workflows/test-cli-with-database.yml
@@ -51,7 +51,7 @@ jobs:
 
       - name: Create Medusa project
         run: |
-          medusa new cli-test --skip-db --branch ci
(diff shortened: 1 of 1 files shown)


olivermrbl:
I tested the changes in my fork, and it ran successfully: https://github.com/olivermrbl/medusa/actions/runs/12792860841/job/35664306994

shahednasser (approved):

PR #10984, opened 2025-01-15
feat(core-flows): cart complete shipping validate

fPolic:
**What**
- validate that there is a shipping method if any of the line items have requires_shipping=true
- validate that products shipping profile is supported by a shipping method on the cart
- update tests

---

CLOSES CMRC-683


olivermrbl (approved):
LGTM, nice

packages/core/core-flows/src/cart/steps/validate-shipping.ts
```diff
@@ -0,0 +1,70 @@
...
+
+    const missingShippingPorfiles = requiredShippingPorfiles.filter(
+      (profile) => !availableShippingPorfiles.includes(profile)
+    )
+
+    if (missingShippingPorfiles.length > 0) {
+      throw new MedusaError(
+        MedusaError.Types.INVALID_DATA,
+        `The cart contains items that require shipping profiles that are not available: ${missingShippingPorfiles.join(
+          ", "
```
olivermrbl:
```suggestion
        "The cart items require shipping profiles that are not satisfied by the current shipping methods"
```

olivermrbl (approved):
LGTM

integration-tests/http/__tests__/order/admin/order.spec.ts
```diff
@@ -932,6 +932,7 @@ medusaIntegrationTestRunner({
           .post(
             `/admin/orders/${order.id}/fulfillments`,
             {
+              shipping_option_id: seeder.shippingOption.id, // shipping option with the "regular" shipping profile
```
olivermrbl, line 935: question: This still optional at the HTTP level right? Why did we need to add this here?

fPolic: Yes, it's optional. In this test, the cart has 2 shipping methods for supporting 2 different shipping profiles that products in the cart have. So when creating fulfilment here we want to specify which method is used to heave deterministic behaviour in the test (otherwise `shipping_methods[0]` would be used which is non determinsitic).

---

This test checks if shipping method for fulfilment supports shipping profile of the items that are being fulfilled but the cart previously had only one shipping method with one shipping profile and the item in cart had another profile.
This PR introduced a validation that a cart cannot be completed if there isn't a shipping method for every shipping profile in the cart so now for the order seeder to work and complete the cart we needed to modify the test so 2 methods are on the cart and needed to specify which one is used for fulfillment in this case.

PR #10987, opened 2025-01-16
feat: Remove fields from payment models that were leftovers from v1

sradevski:
BREAKING: 

If someone was using `region_id` from the `payment collection` entity, or `cart_id`, `order_id`, `customer_id` from the `payment` entity, they will no longer exist. Similarly, any code that was passing `region_Id` to `createPaymentCollections` would be recommended to remove the field.


sradevski:
@shahednasser I wasn't sure which parts of the docs need to be removed as well, I can do it if you point the places or you can just push here the necessary changes (or just open a new PR)

shahednasser:
@sradevski no worries, I'll open another PR to make the changes (We don't want the changes to land before this is released) 👍🏻

carlos-r-l-rodrigues (approved):

PR #10994, opened 2025-01-16
feat: Add support for listing saved payment methods in module and Stripe

sradevski:


packages/core/utils/src/payment/abstract-payment-provider.ts
```diff
@@ -623,6 +625,56 @@ export abstract class AbstractPaymentProvider<TConfig = Record<string, unknown>>
...
+   *
+   *       return {payment_methods: response}
+   *     } catch (e) {
+   *       return {
+   *         error: e,
+   *         code: "unknown",
+   *         detail: e
+   *       }
+   *     }
+   *   }
```
adrien2p: suggestion: I think this need some update based on the latest API impl

adrien2p (approved):
I think it looks good for a first iteration 💪

packages/core/utils/src/payment/abstract-payment-provider.ts
sradevski: Yes nice catch!

packages/modules/providers/payment-stripe/src/core/stripe-base.ts
```diff
@@ -294,6 +296,27 @@ abstract class StripeBase extends AbstractPaymentProvider<StripeOptions> {
     }
   }
 
+  async listPaymentMethods(
+    context: PaymentProviderContext
+  ): Promise<PaymentMethodResponse[]> {
+    const customerId = context.customer?.metadata?.stripe_id
```
riqwan, line 304: q: at what point are these displayed? If its during cart, would the stripe_id even be available at that point?

sradevski: If it is a registered customer (prerequisite to store methods), and they have done at least 1 payment with configuration to store the payment method.

The code that stores stripe_id on the customer doesn't work now, it's part of the tickets I've created and I'll be handling that very soon.

riqwan: Ah i see, these are customer saved payment methods 💡

olivermrbl (approved):

packages/core/types/src/payment/provider.ts
```diff
@@ -135,6 +135,11 @@ export type PaymentProviderAuthorizeResponse = {
   data: PaymentProviderSessionResponse["data"]
 }
 
+export type PaymentMethodResponse = {
+  id: string
+  data: Record<string, unknown>
+}
```
olivermrbl, line 141: thought(non-blocking): The return type could also just be:
```ts
export type PaymentMethodResponse = Record<string, unknown>
```
and the data will hold the ID.

sradevski: @olivermrbl I was thinking that maybe the explicit `id` is needed/useful for linking? or am I wrong?

PR #11015, opened 2025-01-17
feat: Revamp payment provider interface

sradevski:
BREAKING:
1. The payment provider interface completely changed. This allowed for a more unified interface, and it makes it more easily extensible to support different payment providers
2. The `context` field from the `POST /payment-collections/:id/payment-sessions` endpoint was removed as it posed a security risk
3. The context accepted by the payment module changed to only accept known fields and limit the surface area of what can be passed through it.
4. Payment provider methods are expected to throw instead of returning an error.

Affected users are only the ones that have implemented a custom payment provider, or they have used the `createPaymentSession` method directly. If you are using Stripe and you haven't manipulated the payment flow of Medusa the breaking changes should not affect you.

Feature: We have added support for account holder information. In Stripe, for example, we will create a customer for every signed up user on your platform so you can track payments more easily within Stripe. These methods are optional for payment providers.


christiananese (approved):
LGTM, would be good to add a ticket for documenting keys we are defining here like `pp_stripe_stripe_customer_id`

sradevski:
Yes definitely, @shahednasser can you have a look and decide what deserves to be documented in the docs? There is the part of `pp_stripe_stripe_customer_id` that people MIGHT want to know about, although most will unlikely need to deal with it directly

There is also the new interface of a payment provider

packages/modules/providers/payment-stripe/src/core/stripe-base.ts
```diff
@@ -272,36 +250,114 @@ abstract class StripeBase extends AbstractPaymentProvider<StripeOptions> {
...
-    const stripeId = context.customer?.metadata?.stripe_id
+    try {
+      const id = data.id as string
+      const sessionData = (await this.stripe_.paymentIntents.update(id, {
+        amount: amountNumeric,
+      })) as unknown as PaymentProviderSessionResponse["data"]
 
-    if (stripeId !== data.customer) {
-      return await this.initiatePayment(input)
-    } else {
```
olivermrbl, line 285: question: Don't we need to deal with this scenario still?
1. I initialize my cart as a guest customer
2. Go through the checkout flow and create payment sessions
3. I sign in -> should update the customer on the payment intent

olivermrbl: Actually, I am not sure our workflows are wired up to deal with this right now.

sradevski: They are not, and this method is not even called.

I was gonna ask you why https://github.com/medusajs/medusa/pull/9348/files was closed actually, as I imagine that's where these flows would be implemented

packages/core/core-flows/src/cart/workflows/complete-cart.ts
```diff
@@ -118,7 +118,6 @@ export const completeCartWorkflow = createWorkflow(
         // We choose the first payment session, as there will only be one active payment session
         // This might change in the future.
         id: paymentSessions[0].id,
-        context: { cart_id: cart.id },
```
sradevski, line 121: This was never passed anywhere.

packages/core/core-flows/src/order/workflows/mark-payment-collection-as-paid.ts
```diff
@@ -120,7 +118,6 @@ export const markPaymentCollectionAsPaid = createWorkflow(
 
     const payment = authorizePaymentSessionStep({
       id: paymentSession.id,
-      context: { order_id: input.order_id },
```
sradevski, line 123: Similarly this wasn't passed or used anywhere

packages/modules/payment/src/migrations/Migration20250123122334.ts
```diff
@@ -0,0 +1,27 @@
+import { Migration } from '@mikro-orm/migrations';
+
+export class Migration20250123122334 extends Migration {
+
+  override async up(): Promise<void> {
+    this.addSql(`create table if not exists "account_holder" ("id" text not null, "provider_id" text not null, "external_id" text not null, "email" text null, "data" jsonb not null default '{}', "metadata" jsonb null, "created_at" timestamptz not null default now(), "updated_at" timestamptz not null default now(), "deleted_at" timestamptz null, constraint "account_holder_pkey" primary key ("id"));`);
+    this.addSql(`CREATE INDEX IF NOT EXISTS "IDX_account_holder_deleted_at" ON "account_holder" (deleted_at) WHERE deleted_at IS NULL;`);
+    this.addSql(`CREATE UNIQUE INDEX IF NOT EXISTS "IDX_account_holder_provider_id_external_id_unique" ON "account_holder" (provider_id, external_id) WHERE deleted_at IS NULL;`);
+
+    this.addSql(`alter table if exists "payment_collection_payment_providers" drop constraint if exists "payment_collection_payment_providers_payment_coll_aa276_foreign";`);
```
sradevski, line 10: Not sure where these came from, any ideas if they are valid or not?

packages/core/core-flows/src/payment-collection/workflows/create-payment-session.ts
```diff
@@ -68,16 +80,89 @@ export const createPaymentSessionsWorkflow = createWorkflow(
...
+      const paymentCustomer = transform({ customer }, (data) => {
+        return {
+          ...data.customer,
+          billing_address:
+            data.customer.addresses?.find((a) => a.is_default_billing) ??
+            data.customer.addresses?.[0],
+        }
+      })
+
+      const accountHolderInput = {
```
olivermrbl, line 119: question: Can we construct data like this? I thought we always needed to use a `transform` to ensure the value of properties are resolved correctly at runtime and not "composition time".

packages/core/types/src/payment/provider.ts
```diff
@@ -26,33 +37,26 @@ export type PaymentActions =
...
-   * The associated customer's email.
-   */
-  email?: string
-
-  /**
-   * The ID of payment session the provider payment is associated with.
+   * The account holder information, if available for the payment provider.
    */
-  session_id?: string
+  accountHolder?: PaymentAccountHolderDTO
```
olivermrbl: nit: Should probably be snake_cased to align with the general convention

packages/modules/payment/src/migrations/Migration20250123122334.ts
olivermrbl: They are valid but for some reason our migration generator wants to delete them and create them anew, as seen on line 11-14

olivermrbl: We can probably keep the migration as is to correct the casing convention (or whatever it is that triggers the recreation), so we don't keep seeing it moving forward.

packages/modules/providers/payment-stripe/src/core/stripe-base.ts
```diff
@@ -95,221 +113,271 @@ abstract class StripeBase extends AbstractPaymentProvider<StripeOptions> {
...
+    const accountHolderId = accountHolder?.data?.id as string | undefined
+    if (!accountHolderId) {
+      throw this.buildError(
+        "No account holder in context",
+        new Error("No account holder provided while deleting account holder")
+      )
+    }
+
+    try {
+      await this.stripe_.customers.del(accountHolderId)
```
olivermrbl, line 362: question: Wouldn't this need to be the `external_id` of the account holder?

packages/modules/providers/payment-stripe/src/core/stripe-base.ts
```diff
@@ -321,21 +389,23 @@ abstract class StripeBase extends AbstractPaymentProvider<StripeOptions> {
...
+    if (!accountHolderId) {
+      throw this.buildError(
         "Account holder not set while saving a payment method",
         new Error("Missing account holder")
       )
     }
 
     const resp = await this.stripe_.setupIntents.create({
-      customer: customer.metadata.stripe_id as string,
+      customer: accountHolderId,
```
olivermrbl, line 408: question: Same as before, wouldn't this need to be the `external_id`?

olivermrbl (commented):
Overall this LGTM–I'd like us to wait merging until 2.4.0 is out, so we can spend some time properly documenting the breaking changes. Will have another deeper look Monday.

I left a few comments, but nothing major.

packages/core/core-flows/src/payment-collection/workflows/create-payment-session.ts
sradevski: If you are not doing spread, ternary operators, etc. you can yes. In many places we construct the object inline where we pass them to the function, here we just assign it to a variable.

packages/core/types/src/payment/provider.ts
sradevski: Yes good point 👍

packages/modules/providers/payment-stripe/src/core/stripe-base.ts
sradevski: Yes, we just get it from `data.id` since `data` is whatever the payment provider returns. We could also get it from `externalId` but that is not required in the context, that is just stored in our table for convenience

packages/modules/providers/payment-stripe/src/core/stripe-base.ts
sradevski: Answered above

packages/modules/providers/payment-stripe/src/core/stripe-base.ts
```diff
@@ -95,221 +113,271 @@ abstract class StripeBase extends AbstractPaymentProvider<StripeOptions> {
...
+    data,
+    context,
+  }: InitiatePaymentInput): Promise<InitiatePaymentOutput> {
+    const additionalParameters = this.normalizePaymentIntentParameters(data)
 
     const intentRequest: Stripe.PaymentIntentCreateParams = {
       amount: getSmallestUnit(amount, currency_code),
       currency: currency_code,
-      metadata: { session_id: session_id! },
+      metadata: { medusa_session_id: data?.medusa_session_id as string },
```
olivermrbl: comment: Think this is also a breaking change, as we rely on the `metadata.session_id` in the webhook. That said, I don't think this is major, so fine to keep it in, now that we are already breaking other parts.

packages/modules/providers/payment-stripe/src/core/stripe-base.ts
olivermrbl: Oh yeah, I see. Makes sense

olivermrbl (approved):

packages/modules/providers/payment-stripe/src/core/stripe-base.ts
sradevski: @olivermrbl ah right, nice catch! I thought it would be nice if this is namespaced, but maybe we can leave it as-is, it's very minor and it's better if we don't add another breaking change item to the list, it's already a lot to handle.

PR #11024, opened 2025-01-17
fix(core-flows): cart validate sales channel

fPolic:
**What**
- validate that sales channel exist on cart create and update
- fix a bug where default SC would be set on a cart on update if no SC was passed

---

CLOSES CMRC-810


olivermrbl (approved):
LGTM

PR #11039, opened 2025-01-18
feat: Add translations for Macedonian

sradevski:
I figured I should at least translate Medusa in my own language 😄 

I did some AI translation first, and did a quick round of obvious improvements. Can be made better but it's pretty good as-is.

### DIFF
--- a/packages/admin/dashboard/src/i18n/languages.ts
+++ b/packages/admin/dashboard/src/i18n/languages.ts
@@ -1,4 +1,17 @@
-import { de, enUS, es, fr, it, ja, pl, ptBR, th, tr, uk } from "date-fns/locale"
+import {
+  de,
+  enUS,
+  es,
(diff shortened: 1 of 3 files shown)


sradevski:
@kasperkristensen I noticed something weird, there are these 2 fields that are missing, but when I added them it says "unknown key", what am I missing here? They are also missing in eg. en.json

olivermrbl (approved):
Haha, I like it 😄

kasperkristensen:
> @kasperkristensen I noticed something weird, there are these 2 fields that are missing, but when I added them it says "unknown key", what am I missing here? They are also missing in eg. en.json

Seems like the keys were deleted from the en.json file, but the schema command wasn't ran.

You can just run `yarn i18n:schema` to fix the schema and then the test will go through 👍

PR #11047, opened 2025-01-20
fix(event): Subscriber ID miss usage

adrien2p:
PARTIALLY RESOLVES FRMW-2876

**What**
Fix wrong usage of the `subscriberId` in the event bus. It happens that the subscriber id coming from the context was not used at all. This issue lead to duplicated event subscriber with the same subscriber id, it also prevent unsubscribing from event since rand id will be assigned.

**NOTE**
This PR does not handle overide strategy for subscribers with the same id. this still needs to be discussed


adrien2p:
@sradevski could you tell me what you think about this pr, it seems that the subscriber id was not used correctly or not at all in some cases 😂

packages/core/utils/src/event-bus/index.ts
```diff
@@ -85,12 +85,14 @@ export abstract class AbstractEventBusModuleService
      * otherwise we generate a random using a ulid
      */
 
-    const randId = ulid()
     const event = eventName.toString()
+    const subscriberId = context?.subscriberId ?? `${event}-${ulid()}`
+
+    ;(subscriber as any).subscriberId = subscriberId
```
sradevski, line 91: In which cases would the context be empty? Mutating the `subscriber` is maybe not the best, but I guess at least we have tests to ensure this keeps working in the future

sradevski (approved):
LGTM, just a comment to hear your thoughts

packages/core/utils/src/event-bus/index.ts
adrien2p: it can be empty if you subscribe outside of the subscribers directory, maybe though a service or something where you don't necessarily manage a unique subscriber id. In that case you are not require to provide a context and you can create a subscriber which will be assigned a uniq id. The thing is that to not break the API and allow for easy unsubscribing, I thought assigning the id to the constructor of the object would be enough for now. That prevent you from having to get all the subscribers and try to identify yours by comparing the functions. 

And as you mentioned, I ve added a set of tests to ensure the behaviour and that any future breaking changes can be identified 😂

olivermrbl:
Can you add a description of the issue to the PR, i.e. how was it used wrong?

adrien2p:
> Can you add a description of the issue to the PR, i.e. how was it used wrong?

@olivermrbl it was not used at all

adrien2p:
@olivermrbl should we merge this bad boy?

PR #11048, opened 2025-01-20
fix: Unique constraint should account for soft deleted records

adrien2p:
FIXES FRMW-2878

**What**
Currently, the `one-to-one` unique constraints does not account for deleted record. This prevents from inserting a new record wth the same fk if another one is deleted.

**Caveat**
`hasOne` with FK option is meant to be a special case, for example a many to one - one to many without defining the other side of the relation. In that case we don't handle this behaviour and keep it as it is


adrien2p:
@sradevski @olivermrbl wdyt?

sradevski (approved):
I think this looks good. I believe this shouldn't have any breaking change consequences, since it is a less strict check right?

adrien2p:
> I think this looks good. I believe this shouldn't have any breaking change consequences, since it is a less strict check right?

Yes that's correct, it is less strict than it was before

olivermrbl (approved):

PR #11058, opened 2025-01-20
feat(dashboard): Refactor Users table to use the new DataTable block

kasperkristensen:
**What**
- Updates the Users table to use the new DataTable
- Fixes an issue where initial parsing of URL filters weren't formatted correctly.


packages/admin/dashboard/src/components/data-table/hooks/general/use-data-table-date-filters.tsx
```diff
@@ -0,0 +1,101 @@
...
+        label: t("filters.date.today"),
+        value: {
+          $gte: today.toISOString(),
+        },
+      },
+      {
+        label: t("filters.date.lastSevenDays"),
+        value: {
+          $gte: new Date(
+            today.getTime() - 7 * 24 * 60 * 60 * 1000
```
riqwan: sugg: not necessarily for change, but we do have date-fns in the deps already to do something like - `subDays(date, 7)`

riqwan (approved):
lgtm!

packages/admin/dashboard/src/components/data-table/hooks/general/use-data-table-date-filters.tsx
kasperkristensen: True, good point!

PR #11063, opened 2025-01-20
feat: Add support for creating payment methods to payment module

sradevski:
CLOSES CLO-407

### DIFF
--- a/packages/core/utils/src/payment/abstract-payment-provider.ts
+++ b/packages/core/utils/src/payment/abstract-payment-provider.ts
@@ -1,14 +1,10 @@
 import {
   CreatePaymentProviderSession,
   IPaymentProvider,
-  PaymentMethodResponse,
-  PaymentProviderContext,
   PaymentProviderError,
   PaymentProviderSessionResponse,
   PaymentSessionStatus,
   ProviderWebhookPayload,
-  SavePaymentMethod,
-  SavePaymentMethodResponse,
   UpdatePaymentProviderSession,
   WebhookActionResult,
 } from "@medusajs/types"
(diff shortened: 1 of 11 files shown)


packages/modules/payment/src/services/payment-module.ts
```diff
@@ -932,7 +939,45 @@ export default class PaymentModuleService
...
+    sharedContext?: Context
+  ): Promise<PaymentMethodDTO[]>
+  @InjectManager()
+  async createPaymentMethods(
+    data: CreatePaymentMethodDTO | CreatePaymentMethodDTO[],
+    @MedusaContext() sharedContext?: Context
+  ): Promise<PaymentMethodDTO | PaymentMethodDTO[]> {
+    const input = Array.isArray(data) ? data : [data]
+
+    const result = await Promise.all(
```
adrien2p: suggestion: I believe we should use promiseAll instead. also, if one of the payment method fail, what should we do? since the other one might have succeed?

adrien2p: with promiseAll, you can get back all the one that have failed and probably handle that error in some ways? wdyt?

packages/core/utils/src/payment/abstract-payment-provider.ts
```diff
@@ -678,6 +680,52 @@ export abstract class AbstractPaymentProvider<TConfig = Record<string, unknown>>
...
+   *       {customer: externalCustomerId, ...data}
+   *     )
+   *
+   *     return response
+   *   }
+   *
+   *   // ...
+   * }
+   */
+  abstract savePaymentMethod(
```
adrien2p: Q: should we make this optional? I believe not all provider can implement that right?

adrien2p (commented):
I ve left a few comments, I am just wondering, does all providers support payments methods? If yes then LGTM otherwise should we handle that a bit differently such that those methods are not required to be implemented and the provider could have a method to tell if the payment methods is supported or not and handle that appropriately? wdyt?

packages/modules/payment/src/services/payment-module.ts
sradevski: I implemented the bulk support simply to match the interface with other methods, but I doubt it will often be used like that. In the case it is used like that, I think if one fails the method should throw, since you intentionally decided to run then in bulk. But this can be changed later on if that's not true.

adrien2p: yes I do agree, but here if it fails some might still have succeeded but you have not information about what happen, which have failed etc? don't you think using promiseAll (our own util) would be better and then you have all the one that have failed and you can rethrow a proper message saying that X and Y have failed because of... ?

adrien2p: we can still do that later it is not blocking but I am just wondering

packages/core/utils/src/payment/abstract-payment-provider.ts
sradevski: Yeah sounds good, I made `listPaymentMethods` optional as well as not all providers would support it.

packages/modules/payment/src/services/payment-module.ts
sradevski: Ah I am not challenging using `promiseAll`, I will definitely use that. Just saying that what is expected to happen might depend on the use-case :)

sradevski: Pushed a new commit with the changes+

packages/modules/payment/src/services/payment-module.ts
```diff
@@ -965,17 +973,20 @@ export default class PaymentModuleService
...
-      data: item.data,
-      provider_id: input[i].provider_id,
-    }))
+    const normalizedResponse = result
+      .map((item, i) => ({
+        id: item?.id,
+        data: item?.data,
+        provider_id: input[i].provider_id,
+      }))
+      .filter((item) => !!item.id && !!item.data) as PaymentMethodDTO[]
```
adrien2p: Q: In whish case we end up with no item id and item data?

packages/modules/payment/src/services/payment-provider.ts
```diff
@@ -157,17 +157,17 @@ Please make sure that the provider is registered in the container and it is conf
...
   }
 
   async savePaymentMethod(
     providerId: string,
     input: SavePaymentMethod
-  ): Promise<SavePaymentMethodResponse> {
+  ): Promise<SavePaymentMethodResponse | undefined> {
     const provider = this.retrieveProvider(providerId)
-    const res = await provider.savePaymentMethod(input)
+    const res = await provider.savePaymentMethod?.(input)
```
adrien2p: suggestion: we should probably update the cast line 176?

packages/modules/payment/src/services/payment-module.ts
sradevski: This is a bit messy I guess... the return of `this.paymentProviderService_.savePaymentMethod(item.provider_id, item)` can either be an object or undefined. 

Now in the normalization, I leave `id` and `data` as undefined if the response for that entry was undefined, because I then use the index of the array to get the provider ID, and only after I filter the results that had undefined response from the provider.

Let me see how I can refactor it.

adrien2p: My thinking is that we are kind of hiding that a provider does not support payment methods no? Would it be possible that we add a method in the provider service to check if payment methods support is available and otherwise throw an error (eventually instead of returning undefined we can just throw that it is not supported) to say that X and Y are trying to create payment methods on provider Z while it is not supported?

sradevski: Pushed a change, LMK what you think and if you have a suggestion how to handle it better. The only annoying thing is that you might pass eg. 5 entries on input but get 2 back, but I don't know how better to handle it TBH

sradevski: I am worried the interface will become a bit bloated with that sort of check. Throwing though might make sense here, and then the caller can catch that and handle it as they deem fit.

adrien2p: Because here we would be in the scenario where you try to create a payment methods but it is not supported, so instead I would throw from the provider (wdyt) instead of returning undefined to indicate that the user is trying to create a payment method on a provider that does not support it

adrien2p: yes exactly, and since we aggregate the errors, all supported providers will create the payment methods and you will get the error for the one that does not support it or something went wrong

sradevski: Yeah let's do that

adrien2p: > interface will become a bit bloated with that sort of check.

It wouldn't be on the provider, but on the module provider service, but as per my second suggestion it would not be necessary

sradevski: Pushed it, have a look

adrien2p: perfect 👍

adrien2p (approved):

packages/modules/payment/src/services/payment-module.ts
adrien2p: should we add a few tests?

PR #11072, opened 2025-01-21
fix(framework): migration scripts regexp

adrien2p:
FIXES FRMW-2879

**What**
The exclusion regexp was broken and instead use the ignore options of the glob sync function to properly ignore definition files


christiananese (approved):
Very nice!

sradevski (approved):

PR #11074, opened 2025-01-21
fix(dashboard): Persist variant_rank

kasperkristensen:
**What**
- Sends variant_rank to the backend
- Sorts variant table by variant_rank if no other order is selected.

We have a project planned for implementing variant_rank correctly, and adding UI for managing the rank after product creation. This PR handles none of that, as it should be handled as part of said project. This PR simply makes the field useable until then.

Resolves CMRC-859


thetutlage (approved):

riqwan (approved):

pepijn-vanvlaanderen:
@kasperkristensen is there an ETA of when variant_rank will be managable through (drag-n-drop?) the Admin UI?

PR #11075, opened 2025-01-21
fix(ui): Fix position of Checkbox indicator

kasperkristensen:
**What**
- Fixes an issue that would cause the Checkbox indicator to get rendered outside of view in the checkbox was wrapped in a relative div.


fPolic (approved):

kasperkristensen:
@olivermrbl can you review this on behalf of UI when you get the chance?

olivermrbl:
Yes for sure. I will add @fPolic to the UI team, so you guys don't depend on my review going forward.

olivermrbl (approved):

PR #11082, opened 2025-01-22
fix: create parent dir before creating the options file

thetutlage:
Fixes: FRMW-2881

### DIFF
--- a/packages/core/framework/src/build-tools/compiler.ts
+++ b/packages/core/framework/src/build-tools/compiler.ts
@@ -1,7 +1,7 @@
 import path from "path"
 import { getConfigFile } from "@medusajs/utils"
 import type { AdminOptions, ConfigModule, Logger } from "@medusajs/types"
-import { rm, access, constants, copyFile, writeFile } from "fs/promises"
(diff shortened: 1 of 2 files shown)


olivermrbl (approved):

shahednasser (approved):

packages/core/framework/src/build-tools/compiler.ts
```diff
@@ -163,6 +163,7 @@ export class Compiler {
    * for MedusaJS loaders during development
    */
   async #createPluginOptionsFile() {
+    await mkdir(path.dirname(this.#pluginOptionsPath), { recursive: true })
     await writeFile(
```
adrien2p, line 167: Q: If we use our FileSystem util, it would do it automatically right?

thetutlage: Good point @adrien2p

adrien2p (approved):
LGTM, just a nit comment

packages/core/framework/src/build-tools/compiler.ts
olivermrbl: Something we want to do, or should we merge as is?

PR #11084, opened 2025-01-22
fix: Flaky behavior of Yalc

thetutlage:
Fixes: FRMW-2882

Yalc is meant to be used as a binary. However, we are importing it as a package and re-calling the same `publishPackage` function on every change. This makes Yalc behave flaky and therefore this PR calls `yalc` as a command.

We do not have to make the same changes in other commands, since they are not long-lived like the `plugin:develop` command.

### DIFF
--- a/packages/core/framework/src/build-tools/compiler.ts
+++ b/packages/core/framework/src/build-tools/compiler.ts
@@ -1,7 +1,7 @@
 import path from "path"
 import { getConfigFile } from "@medusajs/utils"
 import type { AdminOptions, ConfigModule, Logger } from "@medusajs/types"
-import { rm, access, constants, copyFile, writeFile } from "fs/promises"
+import { rm, access, constants, copyFile, writeFile, mkdir } from "fs/promises"
 import type tsStatic from "typescript"
 
 /**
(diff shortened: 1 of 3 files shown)


shahednasser (approved):

packages/core/framework/src/build-tools/compiler.ts
```diff
@@ -140,6 +140,16 @@ export class Compiler {
...
   }
 
+  /**
+   * Ensures a directory exists
+   */
+  async #ensureDir(path: string, clean?: boolean) {
+    if (clean) {
+      await this.#clean(path)
+    }
+    await mkdir(path, { recursive: true })
```
adrien2p, line 150: Q: Would it make sense to reuse our FileSystem util? Like we create a singleton in the constructor and reuse it when necessary?

packages/medusa/src/commands/plugin/develop.ts
```diff
@@ -7,13 +8,43 @@ export default async function developPlugin({
...
+    execFile(
+      yalcBin,
+      ["publish", "--push", "--no-scripts"],
+      {
+        cwd: directory,
+      },
+      (error, stdout, stderr) => {
+        isBusy = false
+        if (error) {
+          console.log(error)
```
adrien2p, line 43: Q: Could we use the logger instead?

adrien2p (approved):
LGTM, nit comments

packages/core/framework/src/build-tools/compiler.ts
thetutlage: One instance might not work, since all these operations don't share the same parent path.

packages/medusa/src/commands/plugin/develop.ts
thetutlage: This is the error from the child process, which might already be formatted in a certain way. So I think logging it as it is makes more sense. But yes, I can use `console.error` instead.

PR #11086, opened 2025-01-22
fix(admin-bundler): Fix Vite imports to prevent deprecation warnings

kasperkristensen:


sradevski (approved):

packages/admin/admin-bundler/src/lib/plugin.ts
```diff
@@ -12,6 +11,7 @@ interface PluginOptions {
 
 export async function plugin(options: PluginOptions) {
   const vite = await import("vite")
+  const react = (await import("@vitejs/plugin-react")).default
```
adrien2p, line 14: Q: how come dynamic import fix it? 🤔

kasperkristensen: Thats how they want people to import Vite going forward if you need to use CJS. This way it will still work when they no longer have a CJS export in their package, whenever they do decide to not include it anymore

```
Dynamically import Vite: If you need to keep using CJS, you can dynamically import Vite using import('vite') instead. This requires your code to be written in an async context, but should still be manageable as Vite's API is mostly asynchronous.
```
_https://vite.dev/guide/troubleshooting.html#vite-cjs-node-api-deprecated_

The React plugin imports Vite, so we also need to import it dynamically the ensure Vite is only ever imported that way.

PR #11087, opened 2025-01-22
chore(ui): add tsdocs for DataTable components

shahednasser:
Add TSDocs for `DataTable` components to generate better docs for their props

### DIFF
--- a/packages/design-system/ui/src/blocks/data-table/components/data-table-table.tsx
+++ b/packages/design-system/ui/src/blocks/data-table/components/data-table-table.tsx
@@ -23,6 +23,10 @@ interface DataTableTableProps {
   emptyState?: DataTableEmptyStateProps
 }
 
+/**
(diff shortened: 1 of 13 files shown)


packages/design-system/ui/src/blocks/data-table/components/data-table-table.tsx
```diff
@@ -23,6 +23,10 @@ interface DataTableTableProps {
   emptyState?: DataTableEmptyStateProps
 }
 
+/**
+ * This component renders the table in a data table, supporting advanced features.
+ * It's useful to create tables similar to those in the Medusa Admin dashboard.
```
kasperkristensen: I would move the "It's useful..." part to the Root component instead

kasperkristensen (approved):
Added a suggestion but otherwise it looks good to me!

PR #11094, opened 2025-01-22
Update admin_products_export.yaml

Dev-Abdullah-H:
fixed typo


shahednasser (changes requested):
Thank you for your contribution! But can you make the change instead in `www/utils/generated/oas-output/operations/admin/post_admin_products_export.ts`? The file you changed is generated from this file.

PR #11106, opened 2025-01-23
feat(ui,dashboard): Migrate SC tables to DataTable

kasperkristensen:


fPolic (approved):
Looks good!

One thing I noticed is that now when mouse is dragged over a row the entire table seems to rerender.

kasperkristensen:
> Looks good!
> 
> One thing I noticed is that now when mouse is dragged over a row the entire table seems to rerender.

Good catch! It was the state that I use to track which row is being hovered that was causing the entire component to re-render each time it changes. Switched it out with a useRef instead.

PR #11107, opened 2025-01-23
fix(dashboard): Format i18n schema with Prettier

kasperkristensen:
**What**
- Formats the schema that is generated by `yarn i18n:schema`. 
- Ensures we don't end up with huge diffs if one team member forgets to format the schema in one PR, and the schema is then updated and formatted later on.


sradevski (approved):

olivermrbl (approved):
yeees, like it

PR #11108, opened 2025-01-23
fix(medusa,cli): Resources not correctly loaded when running the migration scripts

adrien2p:
FIXES FRMW-2888

**What**
Since the migration scripts runner run as part of the migrate script, we were cleaning up the resources but we didn't take into consideration that links are loading automagically by just importing the links. The issue is that once a link have been imported, the next import will rely on the cached module and therefore when cleaning the resources, re importing those links is not enough. Instead, we are now treating the run-script as a separate command that we run during running the migration so that resources are isolated and loaded according to the need of each script themselves


olivermrbl (approved):

adrien2p:
@olivermrbl do not merge it yet please, i need to throw an été on something about it 👌

adrien2p:
@olivermrbl it should be good now, we re tested it with @christiananese to be absolutely sure. I ve just found out that I had fulfillment tests failing though, this sounds weird so I ll see if it happens again during the next run I have fired

PR #11110, opened 2025-01-23
fix: switch from tsc watch to chokidar

thetutlage:
It was very surprising to find that `tsc` watcher cannot remove compiled files when a source file is deleted. Related issue https://github.com/microsoft/TypeScript/issues/16057

Therefore, we have to get rid of it at the last moment and use a custom watcher via `chokidar` and write the compiled output ourselves.

### DIFF
--- a/packages/core/framework/src/build-tools/compiler.ts
+++ b/packages/core/framework/src/build-tools/compiler.ts
@@ -137,6 +137,17 @@ export class Compiler {
   }
 
   /**
+   * Returns a boolean indicating if a file extension belongs
+   * to a JavaScript or TypeScript file
+   */
+  #isScriptFile(filePath: string) {
+    if (filePath.endsWith(".ts") && !filePath.endsWith(".d.ts")) {
+      return true
+    }
+    return filePath.endsWith(".js")
+  }
+
+  /**
    * Loads the medusa config file and prints the error to
    * the console (in case of any errors). Otherwise, the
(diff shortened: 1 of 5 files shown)


packages/core/framework/src/build-tools/compiler.ts
```diff
@@ -472,54 +451,68 @@ export class Compiler {
    * The "onFileChange" argument can be used to get notified when
    * a file has changed.
    */
-  async developPluginBackend(onFileChange?: () => void) {
-    await this.#ensureDir(this.#pluginsDistFolder, true)
-    await this.#createPluginOptionsFile()
-    const ts = await this.#loadTSCompiler()
+  async developPluginBackend(
+    transformer: (filePath: string) => Promise<string>,
+    onFileChange?: () => void
```
carlos-r-l-rodrigues: should this now pass the "file" and the "action" as arguments to `onFileChange`?

shahednasser (approved):
LGTM

carlos-r-l-rodrigues (approved):

packages/core/framework/src/build-tools/compiler.ts
thetutlage: Yeah, we can do that. But this `onFileChange` action for now doesn't need any info. But it won't hurt to provide this info

adrien2p (approved):

PR #11113, opened 2025-01-23
fix(fulfillment): migration so rules

carlos-r-l-rodrigues:


olivermrbl (approved):
third times a charm 😄

PR #11117, opened 2025-01-23
docs: documet support for type aliases

shahednasser:
Document that we don't support type aliases by default + how to configure them.

Closes DX-1280


shahednasser:
@adrien2p maybe you have some input on this?

PR #11151, opened 2025-01-25
feat(medusa): add sku to variant query params

riqwan:
what:

- adds sku to variant query params


carlos-r-l-rodrigues (approved):

kasperkristensen (commented):
Should we also update the HTTP types? 😄

riqwan:
good point, added @kasperkristensen! 💪🏻

PR #11161, opened 2025-01-27
feat: Add tasks orchestrator

thetutlage:
Fixes: FRMW-2885


carlos-r-l-rodrigues (approved):
LGTM

we only need a public `renewLock(entity)` method.

adrien2p (approved):

adrien2p:
> LGTM
> 
> we only need a public `renewLock(entity)` method.

The acquire lock already expand the lock timeout for a given owner id if it already exists (based on the LUA scripts at least 😅 ). Eventually we could make the acquire method public and it would do that but you think it should be handled at a higher level?

PR #11162, opened 2025-01-27
docs: fix tanstack query version

shahednasser:


EzraBerendsen:
Seems like you'd want to point to the exact version, so remove the `^` from the version.

PR #11190, opened 2025-01-28
chore(orchestration): validate PK when throwIfKeyNotFound

carlos-r-l-rodrigues:
CLOSES: FRMW-2895

Breaking change:
- If `query.graph` option `throw_if_key_not_found` is true, the primary key of the entry point is expected to be used as a filter.


packages/core/orchestration/src/joiner/remote-joiner.ts
```diff
@@ -1203,8 +1207,21 @@ export class RemoteJoiner {
       queryObj,
     })
 
+    if (options?.throwIfKeyNotFound) {
+      if (!primaryKeyArg || !primaryKeyArg.value) {
```
riqwan: sugg: should we split this into 2 different error messages to make it more specific? i.e. `primary keys weren't found` and then `value for primary keys weren't found`

riqwan (approved):
legend! 🔥

packages/core/orchestration/src/joiner/remote-joiner.ts
carlos-r-l-rodrigues: yeah. we can do that

PR #11200, opened 2025-01-28
chore(medusa): clear workflow execution

carlos-r-l-rodrigues:
CLOSES: SUP-704


carlos-r-l-rodrigues:
Should we handle this internally in the module and not expose the new method to be accessed from the app?

packages/modules/workflow-engine-inmemory/src/services/workflows-module.ts
```diff
@@ -155,4 +172,12 @@ export class WorkflowsModuleService<
   ) {
     return this.workflowOrchestratorService_.unsubscribe(args as any, context)
   }
+
+  private async clearExpiredExecutions() {
+    return this.manager_.execute(`
+      DELETE FROM workflow_execution
+      WHERE retention_time IS NOT NULL AND
+      updated_at <= (CURRENT_TIMESTAMP - INTERVAL '1 second' * retention_time);
```
riqwan, line 180: q: is the updated_at here correct?

riqwan (approved):
one question, otherwise lgtm!

packages/modules/workflow-engine-inmemory/src/services/workflows-module.ts
carlos-r-l-rodrigues: yeah, this is correct.
it would be the same as 
`(updated_at + INTERVAL '1 second' * retention_time) <= CURRENT_TIMESTAMP`

riqwan: So this would be if a workflow execution has a 5-day retention period and was last updated 6 days ago, it would be deleted. Right?

carlos-r-l-rodrigues: exactly

Issue #11033, opened 2025-01-17
it keeps asking for the database credentials even after specifying --db-url


YoussefRabeiii:
### What Medusa version and documentation are you using?

v2

### Preliminary Checks

- [x] This issue is not a duplicate. Before opening a new issue, please search existing issues: https://github.com/medusajs/medusa/issues

### Issue Summary

when running `npx create-medusa-app@latest --db-url "url"` it keeps asking for the database credentials and keeps failing

### How can this issue be resolved?

1.
2.
3.
...

### Are you interested in working on this issue?

- [ ] I would like to fix this issue

Fab1en:
I'm also stuck by this bug.

felixarjuna:
Me too ... I think i found the bug though

felixarjuna:
i think i found the bug.

felixarjuna:
I opened a pull request on https://github.com/medusajs/medusa/pull/11214. Hope they could review it fast, so we can continue further with our projects. 🤞

PR #11037, opened 2025-01-17
fix(create-medusa-app): --db-url keep asking db credentials

Gederooney:
issue #11033: it keeps asking for the database credentials even after specifying --db-url #11033

fix: dbName is always define. Checking for dbName first in getDbClientAndCredentials() was always skiping getForDbUrl case

PR #11214, opened 2025-01-29
fix: use --db-url argument when passed

felixarjuna:
the latest version of medusa keep asking for database username and password even though the --db-url arguments is specified. People keep looping on this error and therefore cannot create new project when the database is already hosted somewhere else. In my case it is supabase. 

See opened issue: 
[issues/11033](https://github.com/medusajs/medusa/issues/11033)


shahednasser:
Thank you for opening this PR @felixarjuna but we already have a fix at #11037, it just hasn't been merged yet. Will close this

felixarjuna:
okey, i did not see the pr at https://github.com/medusajs/medusa/pull/11037

PR #11226, opened 2025-01-30
Fix/add additionl data to product categories hook

bouazzaayyoub:
Add an `additional_data` property to the `categoriesCreated` hook, enabling users to extend the Product Category model.


carlos-r-l-rodrigues (approved):
LGTM

bouazzaayyoub:
hello,
any updates about this PR.
Thanks

bouazzaayyoub:
hello,
any updates about this PR??
Thanks

riqwan (approved):
lgtm!

riqwan:
@bouazzaayyoub you're missing whitelisting the additional_data param in the http validators - `packages/medusa/src/api/admin/product-categories/validators.ts`. You can use the `WithAdditionalData` helper as used in other places like - `packages/medusa/src/api/admin/products/validators.ts`

If you can write a simple test for this, would be great!

bouazzaayyoub:
> @bouazzaayyoub you're missing whitelisting the additional_data param in the http validators - `packages/medusa/src/api/admin/product-categories/validators.ts`. You can use the `WithAdditionalData` helper as used in other places like - `packages/medusa/src/api/admin/products/validators.ts`
> 
> If you can write a simple test for this, would be great!

thank you for your reply

packages/medusa/src/api/admin/product-categories/validators.ts
```diff
@@ -44,7 +45,7 @@ export const AdminProductCategoriesParams = createFindParams({
   .merge(AdminProductCategoriesParamsFields)
   .merge(applyAndAndOrOperators(AdminProductCategoriesParamsFields))
 
-export const AdminCreateProductCategory = z
+export const CreateProductCategory = z
```
riqwan, line 48: todo: Retain the Admin prefix here and in other places that was changed in this file.

packages/medusa/src/api/admin/product-categories/validators.ts
```diff
@@ -74,6 +79,10 @@ export const AdminUpdateProductCategory = z
   })
   .strict()
 
+export const AdminUpdateProductCategory = WithAdditionalData(
```
riqwan, line 82: todo: categoriesUpdated does not pass `additional_data` into the hook. You can either:

1. revert this change
2. update `packages/core/core-flows/src/product-category/workflows/update-product-categories.ts` similar to what you did for create product categories

riqwan (commented):
Thanks for the change @bouazzaayyoub, i've added a couple of comments on the new change.

packages/medusa/src/api/admin/product-categories/validators.ts
bouazzaayyoub: many thanks

packages/medusa/src/api/admin/product-categories/validators.ts
bouazzaayyoub: I already create another AdminCreateProductCategory with WithAdditionalData, so it import should be fine.

yarn.lock
```diff
@@ -34330,4 +34330,4 @@ __metadata:
   resolution: "zod@npm:3.22.4"
   checksum: 7578ab283dac0eee66a0ad0fc4a7f28c43e6745aadb3a529f59a4b851aa10872b3890398b3160f257f4b6817b4ce643debdda4fb21a2c040adda7862cab0a587
   languageName: node
-  linkType: hard
+  linkType: hard
```
riqwan, line 34333:
```suggestion
```

riqwan, line 34333:
```suggestion
```

riqwan: todo: undo this change here and it should be good to merge. The tests will fail with this

riqwan: @bouazzaayyoub that changed the lock file entirely. Please run this in your console:

`git checkout develop yarn.lock`

We essentially don't want any yarn.lock changes in this PR as we don't change any deps

bouazzaayyoub: @riqwan many thanks

riqwan:
Legend, thanks for the contribution @bouazzaayyoub!

Issue #11191, opened 2025-01-28
[Bug]: RTL Layout Not Applied for Persian/Arabic Languages in Admin Panel Medusa v2.4


behrouz-shafaati:
### Package.json file

```JSON
N/A
```

### Node.js version

v22.12.0

### Database and its version

PostgresSql v17.2

### Operating system name and version

win 11

### Browser name

Google chrom

### What happended?

Set-up a project using npx create-medusa-app@latest with the default settings ( Medusa v2.4 ) .In Medusa admin panel when selecting Persian (Farsi) or Arabic language in the Medusa Admin Panel, the layout does not switch to Right-to-Left (RTL) as expected. The text is displayed in a left-to-right (LTR) format, causing a misalignment in the user interface, which makes it difficult to use for RTL language users.

Steps to Reproduce:

    Go to the Medusa Admin Panel.
    Select Persian (Farsi) or Arabic from the language dropdown.
    Notice that the page content does not switch to RTL layout.
    The layout remains in the default LTR format, making it non-optimal for RTL language users.

### Expected behavior

The Admin Panel should properly switch to RTL layout when selecting Persian or Arabic, with all content aligned appropriately from right to left.

### Actual behavior

The Admin Panel layout stays in LTR format, even after switching to an RTL language like Persian or Arabic.

### Link to reproduction repo

N/A

sradevski:
Hi there! This is currently not a very high priority since we are a pretty small team, but if you or others are willing to contribute and make the admin support RTL we are more than happy to have it merged. 

I'll close this ticket as issues are used for bug reports, but feel free to start a Github Discussion around this and see if here is anyone interested in contributing.

Thanks!

behrouz-shafaati:
Hi there,

Thank you for your response! I'm actually very interested in resolving this issue and contributing to add proper RTL support for the admin panel. However, I've been struggling to set up the Medusa core locally to make the necessary changes.

I followed the documentation provided [here](https://docs.medusajs.com/v1/development/fundamentals/local-development), but I keep encountering the following error when running the medusa-dev command:

Unable to read and parse the workspace definition from medusa package.json.

Could you please guide me on how to resolve this issue? Once I can successfully set up the core locally, I’ll be happy to work on fixing the RTL support for the admin panel.

Thanks in advance for your help

MEClouds:
Hmm , I can solve this issue 🙂

PR #11252, opened 2025-01-31
feat(dashboard): support RTL in dashboard

MEClouds:
**Description:**  
- Set html direction dynamically based on locale
- Fixed some styles to support RTL properly.  
- Added `dir` attributes for components not affected by global `dir` changes.  

This improves RTL support for better usability in languages like Arabic. Let me know if any changes are needed! 🚀
#11191 

image


MEClouds:
@sradevski 
Should I close this pull request 🤔?

sradevski:
@MEClouds the team will have a look when they have time, you can leave it open :)

willbouch (commented):
Thanks for your contribution and sorry for the very late review. I have one suggestion and wanted your opinion on it:

Instead of using document.getAttribute everywhere (which looks hacky to me considering we are using React), why not create a React hook called `use-direction` which would return the information the components need to handle RTL vs LTR. This way, we could directly handle it inside the components rather than needing to pass a prop all the time. Wdyt?

@medusajs/core feel free to jump in and give opinions on this too

MEClouds:
@willbouch You are right, but when I did that, I didn't look for each file , I just used search and replace 
If I used hook, I need to do import and write additional codes for each file ... so 
that was a fast way for me in that time 😅,
But I can fix that easily now with few tabs , just a few minutes

willbouch:
> @willbouch You are right, but when I did that, I didn't look for each file , I just used search and replace If I used hook, I need to do import and write additional codes for each file ... so that was a fast way for me in that time 😅, But I can fix that easily now with few tabs , just a few minutes

You can tag me when it's ready :)

packages/admin/dashboard/src/components/data-grid/components/data-grid-select-cell.tsx
```diff
@@ -28,7 +28,7 @@
 //       render={({ field: { onChange, ref, ...field } }) => {
 //         return (
 //           <DataGridCellContainer {...container}>
-//             <Select {...field} onValueChange={onChange}>
+//             <Select dir={document.documentElement.getAttribute("dir") as "rtl" | "ltr" | undefined}{...field} onValueChange={onChange}>
```
willbouch: The comment on top is more than 1 year old. @olivermrbl what do you think about simply deleting this file?

packages/admin/dashboard/src/routes/promotions/promotion-create/components/create-promotion-form/create-promotion-form.tsx
```diff
@@ -354,8 +355,8 @@ export const CreatePromotionForm = () => {
   return (
     <RouteFocusModal.Form form={form}>
       <KeyboundForm className="flex h-full flex-col" onSubmit={handleSubmit}>
-        <ProgressTabs
-          value={tab}
+         <ProgressTabs
+        dir={direction}value={tab}
```
willbouch: might be a bit of prettier to run here at line 359

willbouch (commented):
@olivermrbl @fPolic what do you guys think about this one? Looks good on my side

willbouch:
Hi! I am testing and here is what I found:
- Promotion creation seems to be broken. I get error in the console about useDirection
- Cannot access the draft order page (which I am not sure is related to your PR or not)

Did you also go over the app entirely? the rest seemed to work. I just tested in English since I want to check that LTR still works fine

willbouch (approved):
LGTM. I asked someone other members of the team to double check locally :) Thanks for your patience 🙏

shahednasser (changes requested):
Thank you for your contribution, this will be a great addition!

While testing your PR I found some issues. Only one of them is major and should be addressed, but the rest are of less priority.

## Must Fix

In the search modal, the clear button covers the search query

Image

## Nice to Fix

- Drawers still open on the right, they should open on the left
- Toast messages still show on the right, they should show on the left

## Nit

- The sidebar on RTL doesn't have the border separating it from the main content similar to LTR
- The theme icon in the profile menu isn't aligned to the left, making it next to the text

Image

- The padding in badges looks weird, should be on the left side

Image

MEClouds:
> Thank you for your contribution, this will be a great addition!
> 
> While testing your PR I found some issues. Only one of them is major and should be addressed, but the rest are of less priority.
> 
> ## Must Fix
> In the search modal, the clear button covers the search query
> 
> Image
> ## Nice to Fix
> * Drawers still open on the right, they should open on the left
> * Toast messages still show on the right, they should show on the left
> 
> ## Nit
> * The sidebar on RTL doesn't have the border separating it from the main content similar to LTR
> * The theme icon in the profile menu isn't aligned to the left, making it next to the text
> 
> Image
> * The padding in badges looks weird, should be on the left side
> 
> Image


**I've made the following fixes and improvements:**

*  Search issue: The clear button no longer overlaps the search query.
*  RTL sidebar: The border is now visible, matching the LTR layout.
*  Theme toggle: I've implemented a temporary fix until a permanent solution is available from the @medusajs/uilibrary.



The issues with drawers and notifications, as well as the badge styling, are still present. These are best fixed within the @medusajs/ui library itself.

This is what I can do for now.
 **Please check it and review these changes.**

shahednasser (approved):
LGTM Thanks!

fPolic (approved):

PR #11263, opened 2025-02-02
feat(orchestration): hydrate resultset

carlos-r-l-rodrigues:
What:
* Add support for aggregating data into existing resultset.
* `query.graph` new option `initialData` containing the resultset to be hydrated
* It fetches data where the requested fields are not present and merge with the existing resultset


thetutlage (approved):
LGTM!

PR #11281, opened 2025-02-03
chore: Move token from params to body

olivermrbl:
**What**
- Accept token for updating provider identity as a Bearer token instead of params
- Remove deprecated `actorType` from reset-password event

**Why**
- The authorization header is a more secure alternative as it avoids security risks such as leaking the token in logs

This is a breaking change; I decided to not make it backward compatible, as this is security-related.


packages/medusa/src/api/auth/utils/validate-token.ts
```diff
@@ -15,12 +15,13 @@ import { decode, JwtPayload, verify } from "jsonwebtoken"
 // Middleware to validate that a token is valid
 export const validateToken = () => {
   return async (
-    req: MedusaRequest,
+    req: MedusaRequest<{ token: string }>,
```
shahednasser: @olivermrbl Passing by and thought I'd point this out 😄 You'd need to update the `updateProvider` in the JS SDK as well, so it makes more sense to move this to an HTTP type.

olivermrbl: Yes, good point thanks 😄

integration-tests/http/__tests__/auth/admin/auth.spec.ts
```diff
@@ -198,13 +198,11 @@ medusaIntegrationTestRunner({
...
 
-        const response = await api.post(
-          `/auth/user/emailpass/update?token=${result}`,
-          {
-            email: "test@medusa-commerce.com",
-            password: "new_password",
-          }
-        )
+        const response = await api.post(`/auth/user/emailpass/update`, {
+          token: result,
```
sradevski: Should we pass the token in a header instead?

olivermrbl: That's also an option yeah, we can pass it as an authorization header

sradevski: I think if there are no issues with that it would be more consistent with the rest of our authentication setup, but I leave it up to you to decide

sradevski (approved):
LGTM 👍

christiananese (approved):
LGTM

PR #11300, opened 2025-02-04
fix(ui,types): Add Metadata form for collection

kasperkristensen:
**What**
- Adds missing Metadata form for collections
- Fixes type of metadata in update payload
- Fixes an issue where deleting the last row of a metadata form would prevent adding new ones.

Resolves SUP-784


fPolic (approved):
LGTM

PR #11305, opened 2025-02-04
feat(dashboard,admin-vite-plugin): Add support for outlet routes, loader, and handle

kasperkristensen:
**What**
- Add support for defining outlet routes using `@`, e.g. `/src/admin/routes/brands/@create/page.tsx`
- Add support for exporting a `loader` from a route file.
- Add support for exporting a `handle` from a route file.

Example usage of a loader and handle:

```tsx
// src/admin/routes/articles/[id]/page.tsx
import { Button, Container, Heading } from "@medusajs/ui";
import {
  Link,
  LoaderFunctionArgs,
  Outlet,
  UIMatch,
  useLoaderData,
} from "react-router-dom";

export async function loader({ params }: LoaderFunctionArgs) {
  const { id } = params;

  return {
    id,
  };
}

export const handle = {
  breadcrumb: (match: UIMatch<{ id: string }>) => {
    const { id } = match.params;
    return `#${id}`;
  },
};

const ProfilePage = () => {
  const { id } = useLoaderData() as Awaited<ReturnType<typeof loader>>;

  return (
    <div>
      <Container className="flex justify-between items-center">
        <Heading>Article {id}</Heading>
        <Button size="small" variant="secondary" asChild>
          <Link to="edit">Edit</Link>
        </Button>
      </Container>
      {/* This will be used for the next example of an Outlet route */}
      <Outlet />
    </div>
  );
};

export default ProfilePage;
```

In the above example we are passing data to the route from a loader, and defining a breadcrumb using the handle.

Example of a outlet route:

```tsx
// src/admin/routes/articles/[id]/@edit/page.tsx
import { Button, Container, Heading } from "@medusajs/ui";

const ProfileEditPage = () => {
  return (
    <div>
      {/* Form goes here */}
    </div>
  );
};

export default ProfileEditPage;
```
This outlet route will be rendered in the <Outlet /> in the above example when the URL is /articles/1/edit

Resolves CMRC-913, CMRC-914, CMRC-915


olivermrbl (approved):
Overall LGTM!

Any specific considerations around adding the `handle` (or `breadcrumb`) as a separate config instead of including it in the existing `define*` utilities?

My simple (and potentially uninformed) argument would be that one config > multiple configs.

kasperkristensen:
> Overall LGTM!
> 
> Any specific considerations around adding the `handle` (or `breadcrumb`) as a separate config instead of including it in the existing `define*` utilities?
> 
> My simple (and potentially uninformed) argument would be that one config > multiple configs.

`handle` is a `react-router-dom` way of providing arbitrary data to your route, we are just using it for adding breadcrumbs (based on their example https://reactrouter.com/6.29.0/hooks/use-matches#breadcrumbs). But you can pass whatever data you want to it and it will be available through `useMatches` on that route, so we should allow people to create a `handle` so they can use it for whatever use cases they might have.

I am considering adding a field called `breadcrumb` to `defineRouteConfig`, but it will require some Vite magic to rewrite the file so we move the field from the `config` to the `handle` (and if there is no handle for a file then we need to write one), which is a bit more involved so didn't want it to block getting this functionality in now for the plugin I am working on.

olivermrbl:
@kasperkristensen, fine to merge this now to unblock you and we can discuss changing the API later down the line.

PR #11312, opened 2025-02-05
fix(test): update flaky fulfillment test

fPolic:
**What**
- ensure that the same 2 items from the cart are tested on each run and the same SO is used for fulfillement


fPolic:
🤔 locally, the test and the suite are consistently passing

olivermrbl (approved):

PR #11344, opened 2025-02-06
fix(utils): add static identifier to AbstractNotificationProviderService

shahednasser:
Creating a notification provider without `identifier` property throws an error. This PR adds the property to `AbstractNotificationProviderService` with TSDocs.

> NOTE: Unlike other modules with providers, the Notification Module doesn't use this identifier when registering the provider. For example, the File Module registers a provider under `fs_{identifier}_{id}`, where `{id}` is the provider's `id` property in the `medusa-config.ts`. The Notification Module only uses the `id` in `medusa-config.ts`, so this probably should be fixed separately.


olivermrbl (approved):
LGTM, have you created a ticket for the related task

PR #11352, opened 2025-02-06
fix(design-system): singleton prompt

fPolic:
**What**
- add a flag to disable rendering multiple prompts on a page

**Why**
- pressing "r" when a prompt is already open would stack additional prompts

---

CLOSES SUP-802


packages/design-system/ui/src/hooks/use-prompt/use-prompt.tsx
```diff
@@ -3,11 +3,20 @@
...
+import { useRef } from "react"
 
 type UsePromptProps = Omit<RenderPromptProps, "onConfirm" | "onCancel" | "open">
 
-const usePrompt = () => {
+const usePrompt = ({ singleton = false }: { singleton?: boolean } = {}) => {
+  const flag = useRef(false)
+
   const prompt = async (props: UsePromptProps): Promise<boolean> => {
+    if (singleton && flag.current) {
```
kasperkristensen: LGTM, but should we perhaps just change it so this is the default behaviour, without needing to do `singleton: true`? I can't really come up with any cases where you would ever want to render multiple prompts since they stack on top of each other, making everyone except the top one visible.

fPolic: Yup, makes more sense for this to be the default. Will push a change.

kasperkristensen (approved):

packages/design-system/ui/src/hooks/use-prompt/use-prompt.tsx
```diff
@@ -1,22 +1,33 @@
...
 const usePrompt = () => {
+  const isPromptActive = useRef(false)
+  const currentPromptPromise = useRef<Promise<boolean> | null>(null)
+
   const prompt = async (props: UsePromptProps): Promise<boolean> => {
-    return new Promise((resolve) => {
-      let open = true
+    if (isPromptActive.current) {
+      // If the prompt is already active, return the current promise
+      return currentPromptPromise.current!
```
kasperkristensen: Instead of returning `false` we return the active promise if `prompt` is called again while the prompt is already active.

PR #11356, opened 2025-02-07
fix: Minor fixes and cleanup to the payments setup

sradevski:
This PR adds a couple new statuses to the payment collection and payment webhook results. The payment collection will now be marked as "completed" once the captured amount is the full amount of the payment collection.

There are several things left to improve the payment setup, so non-happy-path cases are handled correctly.
1. Currently the payment session and payment models serve a very similar purpose. Part of the information is found on one, and the other part on the other model, without any clear reason for doing so. We can simplify the payment module and the data models simply by merging the two.
2. We need to handle failures more gracefully, such as setting the payment session status to failed when such a webhook comes in.
3. We should convert the payment collection status and the different amounts to calculated fields from the payment session, captures, and refunds, as they can easily be a source of inconsistencies.


packages/core/core-flows/src/payment/steps/capture-payment.ts
```diff
@@ -36,6 +36,26 @@ export const capturePaymentStep = createStep(
...
+  async (payment, { container }) => {
+    if (!payment) {
+      return
+    }
+
+    const paymentModule = container.resolve<IPaymentModuleService>(
+      Modules.PAYMENT
+    )
+
+    // Every time we capture a payment, we create a new capture. We want to delete the newset capture if we capture a payment again.
```
carlos-r-l-rodrigues: In which circumstances we can capture a payment more than once?

packages/core/core-flows/src/payment/steps/refund-payment.ts
```diff
@@ -37,5 +37,25 @@ export const refundPaymentStep = createStep(
...
+  async (payment, { container }) => {
+    if (!payment) {
+      return
+    }
+
+    const paymentModule = container.resolve<IPaymentModuleService>(
+      Modules.PAYMENT
+    )
+
+    // Every time we refund a payment, we create a new refund. We want to delete the newset refund if we refund a payment again.
```
carlos-r-l-rodrigues: same question as the above

packages/core/core-flows/src/payment/steps/capture-payment.ts
```diff
@@ -36,6 +36,26 @@ export const capturePaymentStep = createStep(
...
+    const latestCapture = (payment.captures ?? []).sort((a, b) => {
+      return new Date(b.created_at).getTime() - new Date(a.created_at).getTime()
+    })[0]
+
+    if (!latestCapture) {
+      return
+    }
+
+    await paymentModule.deleteCaptures([latestCapture.id])
```
olivermrbl: thought/question: If the payment has been captured at the third-party provider, and the compensation flow kicks in, we are deleting the Medusa captures but not the third-party capture (likely because we can't). 

Could this not leave the application in a state of inconsistency, for example, a captured payment in Stripe without an associated record in Medusa. I would have the same concern for refunds. 

I believe this is why we originally did not add compensation functions to these steps. However, I realise not deleting captures would also lead to an inconsitency, however one that is perhaps easier to manage.

packages/core/core-flows/src/payment/steps/capture-payment.ts
sradevski: Based on the implementation in the payment module and the models, a payment can have multiple partial captures, so this just implements it with that logic in mind. For example, you might have 100 EUR authorized, but you only want to eg. capture 50 before shipping, and the other 50 after shipping.

carlos-r-l-rodrigues: Yeah, that part I get.
But as we are deleting the capture, I'm not sure how that could be possible other than refunding it.

packages/core/core-flows/src/payment/steps/capture-payment.ts
sradevski: I guess it's about what sort of "consistency" you want, but I didn't think about it like that, I think you are right. I'll add comments for the next naive person to not add it again 😄

packages/core/core-flows/src/payment/steps/capture-payment.ts
sradevski: Ah you refer more to Oli's comment then, yes I realized it is probably a mistake to delete the capture from the DB, I'll revert that.

olivermrbl (approved):

PR #11381, opened 2025-02-10
chore(types,framework): pick ssl types from tls lib

### LINKED ISSUE #11341: Postgres SSL connection options aren't valid types
Making a new issue to get this addressed to unblock my deployment.

https://github.com/medusajs/medusa/issues/10243

Seems the fix suggested in this issue never happened, passing through the SSL options to postgres-node isn't covered by the current type spec and therefore self-signed certs cannot be used.

Image
Image

https://node-postgres.com/features/ssl#self-signed-cert

EDIT with more information: I'm using ECS to deploy my Medusa instances attempting to connect them to RDS.

 _Originally posted by @nicolasdabreo in [#10243](https://github.com/medusajs/medusa/issues/10243#issuecomment-2637255159)_

riqwan:
what:

- picks ssl types from tls lib

RESOLVE https://github.com/medusajs/medusa/issues/11341

### DIFF
--- a/packages/core/framework/src/config/types.ts
+++ b/packages/core/framework/src/config/types.ts
@@ -4,6 +4,7 @@ import {
 } from "@medusajs/types"
 
 import type { RedisOptions } from "ioredis"
+import { ConnectionOptions } from "tls"
 // @ts-expect-error
 import type { InlineConfig } from "vite"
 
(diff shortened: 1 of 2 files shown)


packages/core/framework/src/config/types.ts
```diff
@@ -4,6 +4,7 @@ import {
 } from "@medusajs/types"
 
 import type { RedisOptions } from "ioredis"
+import { ConnectionOptions } from "tls"
```
carlos-r-l-rodrigues: should be 'node:tls' ?

riqwan: Screenshot 2025-02-10 at 12 46 31

Just checked, it is from node:tls

riqwan: changed it to request directly from `node:tls`

carlos-r-l-rodrigues (approved):

PR #11403, opened 2025-02-11
fix(core-flows, types): reservation of shared inventory item

fPolic:
**What**
- if a cart contains variants that share inventory items, reservation of the item would fail also causing complete cart to fail
- include `completed_at` when compensating cart update
- account for multiple reservations of the same item when creating the locking key

---

CLOSES SUP-587

### DIFF
--- a/packages/core/core-flows/src/cart/steps/reserve-inventory.ts
+++ b/packages/core/core-flows/src/cart/steps/reserve-inventory.ts
@@ -99,7 +99,9 @@ export const reserveInventoryStep = createStep(
     const locking = container.resolve(Modules.LOCKING)
 
     const inventoryItemIds = data.inventoryItemIds
-    await locking.execute(inventoryItemIds, async () => {
+    const lockingKeys = Array.from(new Set(inventoryItemIds))
+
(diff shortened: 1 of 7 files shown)


packages/core/core-flows/src/cart/steps/reserve-inventory.ts
```diff
@@ -80,7 +79,9 @@ export const reserveInventoryStep = createStep(
       }
     })
 
-    const reservations = await locking.execute(inventoryItemIds, async () => {
+    const lockingKeys = Array.from(new Set(inventoryItemIds))
+
+    const reservations = await locking.execute(lockingKeys, async () => {
```
fPolic, line 84: comment: if we create multiple reservations of the same inventory item in this step (one for each item in a cart that has items which share the inventory item) we need to deduplicate keys otherwise we will have deadlock. We need to update this in few more places if you agree with the approach.

carlos-r-l-rodrigues: Nice catch.

carlos-r-l-rodrigues (approved):

PR #11414, opened 2025-02-12
fix(admin-vite-plugin,admin-bundler,ui,icons,dashboard,framework,types): Update Vite dependencies

kasperkristensen:
**What**
- Bumps the versions of Vite across the entire stack, to prevent an issue similar to what is described here: https://github.com/vitejs/vite/discussions/18271

Not entirely sure what was happening as I couldn't reproduce the issue, but Adrien faced the issue yesterday when working with local versions of our packages. It does appear as if the range we had before could lead to a version of Vite to be installed with said bug.


adrien2p (approved):

olivermrbl (approved):

packages/admin/admin-bundler/package.json
```diff
@@ -38,7 +38,7 @@
     "glob": "^10.3.10",
     "postcss": "^8.4.32",
     "tailwindcss": "^3.3.6",
-    "vite": "^5.2.11"
```
sradevski, line 41: a Q: as I was looking into the node modules generated in production I noticed the node modules size is pretty big - can Vite be a devDependency instead, or is it used for something other than building? The same applies to the rollup and vite plugins above.

Similarly, should `express` be a peerDependency? I see it defined in both devDependencies and dependencies now.

kasperkristensen: The `admin-bundler` is a development tool, and it uses Vite for the commands that it expose so it can't be a devDependency.

What we could do is make it a `peerDependency` and then also make `vite` a peerDependency of `@medusajs/medusa`, and require that people install Vite in their project. But still we might end up with Vite in the production build, because the build and development commands themselves are build directly into the source of `@medusajs/medusa`. To prevent this we would really need to rethink how the build step and dev server of Medusa work, and move it into a separate package that can be installed along with `@medusajs/medusa` as a devDependency.

For `express` it used to be a peerDependency, but Virk recently changed it to a dependency instead. But as long as the versions are the same across packages and we don't end up with multiple install, I don't believe it should effect the bundle size?

sradevski: Yeah the `express` one is not an issue, I just saw it in both places and I found it strange, it's less about the bundle size :D 

If the admin-bundler is only for building the app, but not used at runtime, wouldn't be ok if it is in the `devDependencies` of `@medusajs/medusa`? Since the build commands will never be called in production, I think it should be fine (but we can test it)

PR #11415, opened 2025-02-12
fix(medusa, types): allow passing require_shipping override flag

fPolic:
**What**
- allow passing the `require_shipping` flag when adding items to a cart

**Why**
- useful for variants that don't have managed inventory to override the default `require_shipping: true` flag

### DIFF
--- a/integration-tests/http/__tests__/cart/store/cart.spec.ts
+++ b/integration-tests/http/__tests__/cart/store/cart.spec.ts
@@ -1123,6 +1123,108 @@ medusaIntegrationTestRunner({
             )
           })
 
+          it("should successfully complete cart without shipping for products without inventory management", async () => {
+            const product = (
+              await api.post(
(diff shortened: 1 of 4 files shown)


carlos-r-l-rodrigues (approved):

olivermrbl (changes requested):
We can't let customers decide if their products require shipping or not. Let's chat internally and come up with a solution to the issue

fPolic:
Indeed, good point. Will close this.

Ashish0988:
We have a digital product for downloading resource online which requires payment but no shipping. Please consider this request, as this will save a ton of time.

olivermrbl:
@Ashish0988, in the next release, you'll be able to do one of two things:
- Set `require_shipping` false on the products inventory items
- Remove shipping profile from the product (it will be made optional in the next release)

PR #11416, opened 2025-02-12
fix: Remove unnecessary swc production dependency

sradevski:
We move `@swc/core` to a peer dependency so production builds don't include it in their node_modules, which shaves off 55MB of dependencies. The package is only required when developing plugins.

### DIFF
--- a/packages/medusa/package.json
+++ b/packages/medusa/package.json
@@ -52,7 +52,9 @@
     "@opentelemetry/sdk-node": "^0.53.0",
     "@opentelemetry/sdk-trace-node": "^1.26.0",
     "@swc/core": "^1.7.28",
+    "@swc/helpers": "^0.5.0",
     "@swc/jest": "^0.2.36",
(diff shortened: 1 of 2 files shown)


packages/medusa/package.json
```diff
@@ -107,9 +108,6 @@
     "@medusajs/user": "^2.5.0",
     "@medusajs/workflow-engine-inmemory": "^2.5.0",
     "@medusajs/workflow-engine-redis": "^2.5.0",
-    "@swc/core": "1.5.7",
-    "@swc/helpers": "^0.5.11",
```
adrien2p, line 111: Q: should we move the helper to dev deps?

adrien2p (approved):

packages/medusa/package.json
sradevski: Ah I see it is a peer dependency, moving it 👍  It is already available but from another package

PR #11431, opened 2025-02-12
feat(medusa): Rollout index engine behind feature flag

adrien2p:
**What**
- Add index engine feature flag
- apply it to the `store/products` end point as well as `admin/products`
- Query builder various fixes
- search capabilities on full data of every entities. The `q` search will be applied to all involved joined table for selection/where clauses


olivermrbl (approved):

olivermrbl:
Looks like the added `status` field has caused some of the tests to fail

PR #11441, opened 2025-02-13
fix: listVariantsList types

thetutlage:
Fixes: FRMW-2732, FRMW-2911

### DIFF
--- a/packages/core/types/src/product/common.ts
+++ b/packages/core/types/src/product/common.ts
@@ -899,22 +899,25 @@ export interface FilterableProductVariantProps
    * Search through the title and different code attributes on the variant
    */
   q?: string
+
   /**
    * The IDs to filter product variants by.
    */
-  id?: string | string[]
+  id?: string | string[] | OperatorMap<string | string[]>
+
   /**
    * The SKUs to filter product variants by.
    */
-  sku?: string | string[]
+  sku?: string | string[] | OperatorMap<string | string[]>
   /**
    * Filter the product variants by their associated products' IDs.
    */
-  product_id?: string | string[]
+  product_id?: string | string[] | OperatorMap<string | string[]>
+
   /**
    * Filter product variants by their associated options.
    */
-  options?: Record<string, string>
+  options?: Record<string, any>
 }
 
 /**
--- a/packages/core/types/src/product/common.ts
+++ b/packages/core/types/src/product/common.ts
@@ -726,8 +726,13 @@ export interface FilterableProductProps
    * Filters on a product's variant properties.
    */
   variants?: {
-    options?: { value: string; option_id: string }
+    options?: {
+      value?: string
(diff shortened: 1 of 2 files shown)


adrien2p (approved):

packages/core/types/src/product/common.ts
```diff
@@ -899,22 +899,25 @@ export interface FilterableProductVariantProps
...
    * Filter the product variants by their associated products' IDs.
    */
-  product_id?: string | string[]
+  product_id?: string | string[] | OperatorMap<string | string[]>
+
   /**
    * Filter product variants by their associated options.
    */
-  options?: Record<string, string>
+  options?: Record<string, any>
```
olivermrbl: question: Couldn't this be typed with the actual shape of options?

thetutlage: It could be. But, I can see that we have been lenient everywhere else as well. 

Thinking out loud. Was the idea that if someone provides a custom implementation for a given module, they can continue to use the same interface and have their filters still work?

riqwan: PS: We also need to change for FilterableProductProps as the same issue would be there as well

thetutlage: Made it more specific after finding that `variants.options` in `listProducts` is specific too

olivermrbl: My thinking in this case specifically was that we know the shape of options (which is fixed), so we might as well type it accordingly. Is there a reason we shouldn't do that?

olivermrbl: Ah you beat me to it. Think this is nice :)

thetutlage: @riqwan Fixed https://github.com/medusajs/medusa/pull/11441/files#diff-d39b7d61536c3c804e9498463649a01fbcf75b11747e25f6df396537e577a42cR925-R929

packages/core/types/src/product/common.ts
```diff
@@ -726,8 +726,13 @@ export interface FilterableProductProps
    * Filters on a product's variant properties.
    */
   variants?: {
-    options?: { value: string; option_id: string }
+    options?: {
+      value?: string
+      option_id?: string
+      option?: Record<string, any>
```
fogine, line 732: if its possible to provide `option` with option name as index and option value as value of the hash map, the list product api endpoint validation definition does not reflect that:
https://github.com/medusajs/medusa/blob/develop/packages/medusa/src/api/store/products/validators.ts#L32

thetutlage: @fogine These changes are for the `listProductVariants` and `listProducts` methods and I think the validators are built more around what we want to accept in HTTP payloads.

Correct me if I am missed something from your original message

PR #11464, opened 2025-02-14
fix(medusa): Re throw error in instrumentation after reporting it

adrien2p:
FIXES FRMW-2914

**What**
Currently, when the instrumentation is enabled, some instrument catches the errors in order to set the span status and info, unfortunately, these errors are not re throw leading to swallow them and return nothing in the end leading to potential breaks in flows


thetutlage (approved):
LGTM!

olivermrbl (approved):

PR #11466, opened 2025-02-14
fix(dashboard): Properly delete metadata keys, and fix number parsing

kasperkristensen:
**What**
- Fixes an issue where we would parse a string like `"52 tests"` into `52` in the Metadata form
- If a row is deleted we now send off key with an empty string as its value to signify that it should be deleted. This is needed to introduce the API we had for updating metadata in V1. Adrien is implementing the BE support for this.

RESOLVES SUP-895


carlos-r-l-rodrigues (approved):

packages/admin/dashboard/src/components/forms/metadata-form/metadata-form.tsx
```diff
@@ -393,17 +404,17 @@ function parseValues(
...
     // We try to cast the value to a boolean or number if possible
     if (value === "true") {
       update[key] = true
     } else if (value === "false") {
       update[key] = false
     } else {
-      const parsedNumber = parseFloat(value)
-      if (!isNaN(parsedNumber)) {
-        update[key] = parsedNumber
+      const isNumeric = /^-?\d*\.?\d+$/.test(value)
```
carlos-r-l-rodrigues, line 415:
```suggestion
      const isNumeric = !isNaN(+value)
```

carlos-r-l-rodrigues: the only side effect are hex numbers `!isNaN(+"0xff")`

PR #11467, opened 2025-02-14
feat(js-sdk): implement custom storage config to support react native

ranjithkumar8352:
JS SDK currently doesn't support React native seamlessly. Workarounds are needed to store the auth token using `jwtTokenStorageMethod: 'memory'` and manually set the token or clear the token using the `setToken` and `clearToken` methods.

Example in [Medusa React Native starter](https://github.com/bloomsynth/medusa-mobile-react-native)

This PR adds the support for passing custom storage to JS SDK so the user can choose his preferred storage and pass a simple storage interface to the auth config options.

image


ranjithkumar8352:
Have added unit tests as well. Can someone review this? @sradevski

sradevski (approved):
Thanks for the contribution @ranjithkumar8352 , I think this looks very good!

Let me just circle back with the team if they have any concerns, otherwise we can have it merged soon.

ranjithkumar8352:
Thank you @sradevski 🤗

carlos-r-l-rodrigues (approved):

PR #11469, opened 2025-02-14
feat(): Backport metadata management

adrien2p:
FIXES FRMW-2915

**What**
Backport metadata management. all the metadata get preserved unless a specific empty string is provided for the key which in turn would remove that key from the metadata


.changeset/kind-chairs-cough.md
```diff
@@ -0,0 +1,6 @@
+---
+"@medusajs/utils": patch
+---
+
+feat(utils): Backport metadata management
```
carlos-r-l-rodrigues, line 5: add `modules-sdk`

carlos-r-l-rodrigues (approved):

.changeset/kind-chairs-cough.md
adrien2p: I did and then reverted man, I ve only changed the utils package :D

PR #11472, opened 2025-02-14
fix(medusa): Allow filtering by handle and title as arrays

kasperkristensen:
**What**
- Allows filtering products in the Store API with an array value for both `handle` and `title`.

RESOLVES SUP-893


fPolic (approved):
LGTM

olivermrbl (approved):

PR #11489, opened 2025-02-17
feat(framework): Improve production structured logging with more valuable information as well

adrien2p:
FIXES FRMW-2916

**What**
Improve production logging to improve log search for investigation purposes.

Screenshot 2025-02-17 at 13 30 07

**NOTE**
A separate ticket has been open for the `trust proxy` discussion


packages/core/framework/src/http/express-loader.ts
```diff
@@ -72,21 +72,53 @@ export async function expressLoader({ app }: { app: Express }): Promise<{
...
+
+        // Standard HTTP request properties
+        http_verion: tokens["http-version"](req, res),
+        method: tokens.method(req, res),
+        url: tokens.url(req, res),
+
+        // Response details
+        status: Number(tokens.status(req, res)),
+        content_length: tokens.res(req, res, "content-length") || 0,
+        response_time: `${tokens["response-time"](req, res)} ms`,
```
sgirones: let make this a numeric value so it's easier to process by monitoring tools

packages/core/framework/src/http/express-loader.ts
```diff
@@ -72,21 +72,53 @@ export async function expressLoader({ app }: { app: Express }): Promise<{
...
+        request_id: req.requestId || "-",
+
+        // Standard HTTP request properties
+        http_verion: tokens["http-version"](req, res),
+        method: tokens.method(req, res),
+        url: tokens.url(req, res),
+
+        // Response details
+        status: Number(tokens.status(req, res)),
+        content_length: tokens.res(req, res, "content-length") || 0,
```
sgirones: could we have `response_size` and request_size`?

packages/core/framework/src/http/express-loader.ts
```diff
@@ -72,21 +72,53 @@ export async function expressLoader({ app }: { app: Express }): Promise<{
...
+    const jsonFormat = (tokens, req, res) => {
+      const result = {
+        level: "http",
+        // Request ID can be correlated with other logs (like error reports)
+        request_id: req.requestId || "-",
+
+        // Standard HTTP request properties
+        http_verion: tokens["http-version"](req, res),
+        method: tokens.method(req, res),
+        url: tokens.url(req, res),
```
sgirones: will this contain the query params as well?

packages/core/framework/src/http/express-loader.ts
```diff
@@ -72,21 +72,53 @@ export async function expressLoader({ app }: { app: Express }): Promise<{
...
-      : "combined",
-    {
-      skip: shouldSkipHttpLog,
-      stream: {
-        write: (message: string) => logger.http(message.trim()),
-      },
+  if (!IS_DEV) {
+    const jsonFormat = (tokens, req, res) => {
+      const result = {
+        level: "http",
```
sgirones, line 86: should we call this `component` or something along the lines instead? `level` may be interpreted as `log_level`.

And could we add the log level as well? `info` in this case

packages/core/framework/src/http/express-loader.ts
```diff
@@ -72,21 +72,53 @@ export async function expressLoader({ app }: { app: Express }): Promise<{
...
+        status: Number(tokens.status(req, res)),
+        content_length: tokens.res(req, res, "content-length") || 0,
+        response_time: `${tokens["response-time"](req, res)} ms`,
+
+        // Useful headers that might help in debugging or tracing
+        referrer: tokens.referrer(req, res) || "-",
+        user_agent: tokens["user-agent"](req, res),
+
+        timestamp: new Date().toISOString(),
+      }
```
sgirones, line 109: I would like to see the `client_ip` too

packages/core/framework/src/http/express-loader.ts
sgirones: is this the time of the entire request or just the time it took to process and send the response? should we use `duration` instead to make it obvious?

packages/core/framework/src/http/express-loader.ts
adrien2p: should be good @sgirones 💪 let me know what you think

packages/core/framework/src/http/express-loader.ts
adrien2p: it is the time of the entire request, so it starts when it reaches the middleware and end when we send the response, yes let me apply the changes

packages/core/framework/src/http/express-loader.ts
adrien2p: sure sounds better indeed :)

packages/core/framework/src/http/express-loader.ts
adrien2p: the log level was currently set to HTTP, which when we check was a valid level. Do you want to change it to info is that correct?

adrien2p:
@sgirones here you should have the fully updated log

Screenshot 2025-02-17 at 13 30 07

packages/core/framework/src/http/express-loader.ts
adrien2p: Screenshot 2025-02-17 at 13 31 35

This log level was comming from winston which was following npm approach, let me know if you want it to be different and in that case I can switch to something else

packages/core/framework/src/http/express-loader.ts
```diff
@@ -72,21 +72,57 @@ export async function expressLoader({ app }: { app: Express }): Promise<{
...
+      const result = {
+        level: "http",
+        // client ip
+        client_ip: req.ip || "-",
+
+        // Request ID can be correlated with other logs (like error reports)
+        request_id: req.requestId || "-",
+
+        // Standard HTTP request properties
+        http_verion: tokens["http-version"](req, res),
```
carlos-r-l-rodrigues:
```suggestion
        http_version: tokens["http-version"](req, res),
```

packages/core/framework/src/http/express-loader.ts
sgirones: ah, didn't know `http` was an actual log level. I guess we can keep it then

packages/core/framework/src/http/express-loader.ts
```diff
@@ -60,33 +60,71 @@ export async function expressLoader({ app }: { app: Express }): Promise<{
     })
   }
 
-  app.set("trust proxy", 1)
+  app.set("trust proxy", true)
```
sradevski: Should we maybe do this in a different PR? It has some security implications and it probably needs a bit more thought, WDYT?

sradevski (commented):
Nice, this should makes logs a lot nicer to deal with. Just one comment around the trust proxy

packages/core/framework/src/http/express-loader.ts
adrien2p: We ve discussed it with @sgirones, I will revert it there but the ideal would be to be able to configure it, I ll open a ticket and mentioned oli about it

sradevski: I think even if we do it it might be better if it is a separate PR as it is an important change, security-wise. I leave it up to you guys to decide though :)

adrien2p: here is the ticket FRMW-2917

adrien2p: I ve reverted it, salvador confirmed that it is not too much of an issue in the short term, but the logs will contains the wrong IP address. I ve open the ticket such that we have a proper discussion around it. Then doing the changes should be quick

sgirones (approved):

olivermrbl (approved):
LGTM, I guess this should be deemed a "breaking" change?

adrien2p:
@olivermrbl I ve slapped a flag for breaking changes 👍

PR #11493, opened 2025-02-17
feat(core-flows, types): add has missing inventory flag when listing shipping options

fPolic:
**What**
- add `insufficient_inventory` flag when listing shipping options for a cart
- add `enabled_in_store` flag when creating/editing pickup options


packages/core/core-flows/src/cart/workflows/list-shipping-options-for-cart.ts
```diff
@@ -154,15 +162,42 @@ export const listShippingOptionsForCartWorkflow = createWorkflow(
...
+                return !level ? true : level.available_quantity === 0
+              })
+            }
+          )
+
           return {
             ...shippingOption,
             amount: price?.calculated_amount,
             is_tax_inclusive: !!price?.is_calculated_price_tax_inclusive,
+            has_missing_inventory:
```
olivermrbl: nit: Can we change this to `insufficient_inventory: true`? It's nitpicky, but you cannot have missing inventory 😄

olivermrbl (approved):
LGTM, one comment, think we can merge after it has been addressed.

packages/core/core-flows/src/cart/workflows/list-shipping-options-for-cart.ts
fPolic: 😅 done.

olivermrbl (commented):
Can we add a test too?

PR #11494, opened 2025-02-17
feat: Add support for idempotency key in payments

sradevski:
This PR makes changes to the payment module so an idempotency key is passed to the payment provider, and in cases where the payment provider supports it, it is used to guarantee idempotency of requests (eg. Stripe is one that does support idempotency keys).

CLOSES CLO-408


christiananese (approved):
Finally! 😄

olivermrbl (approved):

PR #11511, opened 2025-02-18
fix(framework): add missing storefrontUrl from configuration type

shahednasser:
The `storefrontUrl` is defined in the `packages/core/types/src/common/config-module.ts` file, but for many purposes, including the documentation, we use the types defined in `packages/core/framework/src/config/types.ts`. This PR adds the `storefrontUrl` to the framework type

### DIFF
--- a/packages/core/framework/src/config/types.ts
+++ b/packages/core/framework/src/config/types.ts
@@ -88,6 +88,24 @@ export type AdminOptions = {
    */
   backendUrl?: string
   /**
+   * The URL of your Medusa storefront application. This URL is used as a prefix to some
+   * links in the admin that require performing actions in the storefront. For example,
+   * this URL is used as a prefix to shareable payment links for orders with
+   * outstanding amounts.
+   *
+   * @example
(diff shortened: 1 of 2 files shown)


packages/core/framework/src/config/types.ts
```diff
@@ -87,6 +87,24 @@ export type AdminOptions = {
...
+   * links in the admin that require performing actions in the storefront. For example,
+   * this URL is used as a prefix to shareable payment links for orders with
+   * outstanding amounts.
+   *
+   * @example
+   * ```js title="medusa-config.js"
+   * module.exports = defineConfig({
+   *   admin: {
+   *     storefrontUrl: process.env.MEDUSA_STOREFRONT_URL ||
+   *       "http://localhost:9000"
```
kasperkristensen: nit: to keep it inline with our starter

```suggestion
   *       "http://localhost:8000"
```

kasperkristensen (approved):
LGTM, just a small suggestion

PR #11517, opened 2025-02-18
change devDependencies in packages

cortopy:
I've built my first application with Medusa, and the docker image is surprisingly large. I've gone through the dependencies declared in the core packages and I have found two that are not needed as production dependencies. This is specially the case of `@medusajs/admin-bundler` which brings in a lot of dev packages.

This PR simply moves those dependencies as `devDependencies`

### DIFF
--- a/packages/core/framework/package.json
+++ b/packages/core/framework/package.json
@@ -69,6 +69,7 @@
     "pg": "^8.13.0",
     "rimraf": "^3.0.2",
     "supertest": "^4.0.2",
+    "tsconfig-paths": "^4.2.0",
     "typescript": "^5.6.2",
     "vite": "^5.4.14"
   },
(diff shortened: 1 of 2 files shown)


kasperkristensen:
Hi @cortopy, 

It's not true that `@medusajs/admin-bundler` is not needed as a dependency. `devDependencies` are dependencies that are only used for building the package, and not used for any of the executable code in the package. This is not the case for `@medusajs/admin-bundler` as it is used in both the `build` command and in our `adminLoader`. If you move it to devDependencies then it won't be installed in the Medusa app that install `@medusajs/medusa`, and the `build` command and loader will no longer work. The same applies for `tsconfig-paths`, its used in executable code, and therefore needs to be a dependency.

We could definitely clean up our package structure, to keep build tools out of `@medusajs/medusa` and move it into a separate package that you would then install along with `@medusajs/medusa` but as a `devDependency` and it is something we are discussing but it requires quite a restructuring compared to how Medusa works today, so not a simple fix.

PR #11518, opened 2025-02-18
fix(order): calculate taxes on order edit flows

carlos-r-l-rodrigues:
non breaking changes:
 - `difference_sum` removed from order summary. This was a very specific field, based on the actions' calculation that could cause more confusion than help.


olivermrbl:
I'd like to avoid the breaking change. Would that be possible? The current behavior is consistent with set operations across all modules.

carlos-r-l-rodrigues:
> I'd like to avoid the breaking change. Would that be possible? The current behavior is consistent with set operations across all modules.

Currently it is similar to Cart.
However Orders are continously edited, and you can add items that belong to an Order continuation (Order edits). To me this behavior of removing taxes or adjustments that aren't present is not correct.
If we don't introduce this change here, all the flows setting taxes and adjustments will have to fetch all the existing items and set them all together again. Which seems a bit of waste of resources.

The alternative would be keeping the set methods, but do not use them on the workflows. And use the upsert instead.

carlos-r-l-rodrigues:
@riqwan can I have a second pair of eyes here?

olivermrbl:
> The alternative would be keeping the set methods, but do not use them on the workflows. And use the upsert instead.

I would prefer this approach. I think the inconsistency between cart and order operations will inevitably lead to issues at some point.

riqwan:
Can you share some notes on what the goal is on the PR? A little lost here 😅 

I'll have a deeper look tomorrow. But from a quick look, if i understand this correctly, what we're fixing here is that we retain or copy over the tax lines and adjustments from an existing item that was ported over from a cart during change actions.

I can't speak confidently about taxes, but for promotion adjustments, if an item is added, removed or updated upon an order edit, the promotion can potentially no longer apply or a different promotion can apply. Can this change be accurately applied on an order module level?

carlos-r-l-rodrigues:
> Can you share some notes on what the goal is on the PR? A little lost here

Oh sorry, I asked you to have a look more on the rma-flow tests that you have added the full flow to test it. This PR fixes tax calculations when new items are added or items are returned.

Regarding promotions/adjustments, they are included on the calculations, however the application of new adjustments during order edits are not covered in this PR. I don't believe this will come anytime soon.
In either case, the important part is that now the calculations of items and shipping methods, including taxes and adjustments are taken into consideration.

olivermrbl (approved):
LGTM, will let @riqwan have a look as well.

integration-tests/http/__tests__/order-edits/order-edits.spec.ts
```diff
@@ -368,7 +368,7 @@ medusaIntegrationTestRunner({
           )
         ).data.order_preview
 
-        expect(result.summary.current_order_total).toEqual(84)
+        expect(result.summary.current_order_total).toEqual(86.4)
```
riqwan, line 371: Item 1 - 12$
Item 2 - 25$
Shipping - 10$

Total before - `(25 * 2) + (12 * 2) + 10 = 84`

If both item 1 and item 2 have a 10% tax, shouldn't it be:

`(25 * 2 + 5) + (12 * 2 + 2.4) + 10 = 91.4`

integration-tests/http/__tests__/order/admin/rma-flows.spec.ts
```diff
@@ -621,8 +597,7 @@ medusaIntegrationTestRunner({
         expect(pendingPaymentCollection).toEqual(
           expect.objectContaining({
             status: "not_paid",
-            // TODO: The payment should also include taxes
-            amount: 115.9,
+            amount: 15.9,
```
riqwan, line 600: q: shouldn't this be 227.9?

integration-tests/http/__tests__/order-edits/order-edits.spec.ts
olivermrbl: Item 1 doesn't have taxes, it is created manually on the order using the order module. Since we don't refresh all tax lines on every request, but only the ones generated by added/updated items, the item remains without taxes.

riqwan: Ah wait, I see that the new item that was added will not have a tax as its being built from scratch. So it will be this

`(25 * 2) + (12 * 2 + 2.4) + 10 = 86.4`

packages/modules/order/src/services/order-module-service.ts
```diff
@@ -3228,11 +3328,12 @@ export default class OrderModuleService
...
+      const initialDiff = MathBN.sub(
+        summary.totals.transaction_total,
+        initialTrxTotal
+      )
       summary.totals.pending_difference = new BigNumber(
-        MathBN.sub(
-          summary.totals.current_order_total,
-          summary.totals.transaction_total
-        )
+        MathBN.sub(summary.totals.pending_difference, initialDiff)
```
riqwan, line 3336: The previous diff looks correct to me here - if `pending_difference` is how much a customer has to pay, shouldn't it be the current order total subtracted by the current transaction total?

riqwan (approved):
An open question on how pending_difference is calculated and its resulting payment collection value, but otherwise looks good to me! 💪🏻

integration-tests/http/__tests__/order/admin/rma-flows.spec.ts
carlos-r-l-rodrigues: Why 227.9? One item is being returned and another one being added. They have the same price, so it stays the same.
But good question, not sure if I need another property on the summary to display these expected values, that are not yet complete. Like returns.

packages/modules/order/src/services/order-module-service.ts
carlos-r-l-rodrigues: Return requests create a negative pending value, if we just get the current total minus the transactions we overwrite them.

integration-tests/http/__tests__/order/admin/rma-flows.spec.ts
riqwan: It came from the formula i knew from previously. I'll resolve this and focus on the other comment, which points to the same issue.

packages/modules/order/src/services/order-module-service.ts
riqwan: Ah I see, that makes sense. The confusion here then is more about why current_order_total includes the positive cost of the returned items then. I don't know if there's anything to fix here, just some education is probably needed as comments inline 🤞🏻

riqwan: we can discuss this later once you're back, merging this in for now 🤞🏻

PR #11525, opened 2025-02-19
fix(utils): Handle 0 correctly in build query

olivermrbl:
**What**

Remove truthy/falsy check for `take` and `skip`

**Why**

To ensure `{ take: 0, skip: 0 }` is not converted to `{ take: undefined, skip: undefined }`


carlos-r-l-rodrigues (approved):
is the strict check `!==` intended, or should it be `config.take != null`?

adrien2p (approved):
+1 on != null

PR #11533, opened 2025-02-19
feat(core-flows, types): calculated shipping in RMA flows

fPolic:
**What**
- support calculated shipping options in the return / exchange / claim flow

**How**
- introduced 2 util workflows (reused across all 3 RMA flows)
  - `fetchShippingOptionForOrderWorkflow`
    - retrieves a shipping option with pricing
    - used when adding a shipping method to an RMA flow 
  - `maybeRefreshShippingMethodsWorkflow` 
    - used to update a shipping method and corresponding action if the shipping option is calcualted
- added `refresh(Return|Claim|Exchange)ShippingWorkflow`
  - retrieves relevant data for the flow and calls `maybeRefreshShippingMethodsWorkflow`
  - called in all places where items in the RMA flow are updated

---

https://github.com/user-attachments/assets/06d1bef5-6e0f-4570-b565-8cab7fb399f1

---

CLOSES CMRC-806


packages/core/types/src/fulfillment/mutations/shipping-option.ts
```diff
@@ -149,9 +149,29 @@ export interface CalculateShippingOptionPriceDTO {
...
     /**
-     * The location that the items will be shipped from.
+     * The location that the items will be shipped from (or to if it is a return).
      */
     from_location?: StockLocationDTO
+
+    /**
+     * The return details - if present, the return shipping needs to be calculated.
+     */
+    orderReturn?: {
```
olivermrbl: todo: Let's clean up this entire type while we're at it. The mixing of the casing is triggering 😄 

I think this needs to be camel_cased. Please deprecate (and add a todo) the ones that are changed, so we avoid breaking the type.

olivermrbl (commented):
Could I get you to describe how we have achieved this in the PR description?

packages/core/types/src/fulfillment/mutations/shipping-option.ts
fPolic: I refactored the type. The type is only extended with new properties so there shouldn't be a breaking change.

packages/core/core-flows/src/order/workflows/return/refresh-shipping.ts
```diff
@@ -0,0 +1,96 @@
...
+ * @summary
+ *
+ * Refresh return shipping.
+ */
+export const refreshReturnShippingWorkflow = createWorkflow(
+  refreshReturnShippingWorkflowId,
+  function (
+    input: WorkflowData<RequestItemReturnValidationStepInput>
+  ): WorkflowResponse<void> {
+    const orderChange: OrderChangeDTO = useRemoteQueryStep({
```
olivermrbl: nit: Use query graph instead so we can move toward deprecating this step

packages/core/core-flows/src/order/workflows/return/create-return-shipping-method.ts
```diff
@@ -142,6 +145,7 @@ export const createReturnShippingMethodWorkflow = createWorkflow(
         "claim_id",
         "exchange_id",
         "canceled_at",
+        "items.*",
```
olivermrbl: nit: Can we be more specific about the fields we choose? No need to fetch everything if we can avoid it.

olivermrbl: Are these items even used?

packages/core/core-flows/src/order/workflows/exchange/refresh-shipping.ts
```diff
@@ -0,0 +1,150 @@
...
+ * @summary
+ *
+ * Refresh exchange shipping.
+ */
+export const refreshExchangeShippingWorkflow = createWorkflow(
+  refreshExchangeShippingWorkflowId,
+  function (
+    input: WorkflowData<RefreshExchangeShippingWorkflowInput>
+  ): WorkflowResponse<void> {
+    const orderChange: OrderChangeDTO = useRemoteQueryStep({
```
olivermrbl: nit: Use query graph

packages/core/core-flows/src/order/workflows/exchange/refresh-shipping.ts
```diff
@@ -0,0 +1,150 @@
...
+            order_id: orderChange.order_id,
+            action_id: inboundShippingAction.id,
+            context: {
+              return_id: inboundShippingAction.return_id,
+              return_items: items,
+            },
+          })
+        } else {
+          result.push(null)
+        }
```
olivermrbl: question: Do we need to push `null` here?

packages/core/core-flows/src/order/workflows/exchange/refresh-shipping.ts
```diff
@@ -0,0 +1,150 @@
...
+
+    // Refresh outbound shipping method
+    when({ refreshArgs }, ({ refreshArgs }) => refreshArgs[1] !== null).then(
+      () =>
+        maybeRefreshShippingMethodsWorkflow
+          .runAsStep({
+            input: refreshArgs[1],
+          })
+          .config({ name: "refresh-outbound-shipping-method" })
+    )
```
olivermrbl: thought: What happens if there are more than one outbound or inbound shipping method on the exchange?

packages/core/core-flows/src/order/workflows/exchange/refresh-shipping.ts
olivermrbl: I guess I understand how it works with the subsequent when–then, but I think we need to consider the scenario where multiple methods are applied.

packages/core/core-flows/src/order/workflows/claim/refresh-shipping.ts
```diff
@@ -0,0 +1,150 @@
...
+ * @summary
+ *
+ * Refresh claim shipping.
+ */
+export const refreshClaimShippingWorkflow = createWorkflow(
+  refreshClaimShippingWorkflowId,
+  function (
+    input: WorkflowData<RefreshClaimShippingWorkflowInput>
+  ): WorkflowResponse<void> {
+    const orderChange: OrderChangeDTO = useRemoteQueryStep({
```
olivermrbl: nit: Use query graph

packages/core/core-flows/src/order/workflows/claim/refresh-shipping.ts
```diff
@@ -0,0 +1,150 @@
...
+
+    // Refresh outbound shipping method
+    when({ refreshArgs }, ({ refreshArgs }) => refreshArgs[1] !== null).then(
+      () =>
+        maybeRefreshShippingMethodsWorkflow
+          .runAsStep({
+            input: refreshArgs[1],
+          })
+          .config({ name: "refresh-outbound-shipping-method" })
+    )
```
olivermrbl: Same question as for exchanges applies here.

packages/core/core-flows/src/order/utils/maybe-refresh-shipping-methods.ts
```diff
@@ -0,0 +1,170 @@
...
+  context: CalculatedRMAShippingContext
+}
+
+export const maybeRefreshShippingMethodsWorkflowId =
+  "maybe-refresh-shipping-methods"
+/**
+ * This workflows refreshes shipping method an order (used in RMA flows).
+ * The shipping method and the action is updated if the shipping option is calculated.
+ */
+export const maybeRefreshShippingMethodsWorkflow = createWorkflow(
```
olivermrbl, line 74: nit: Use query graph in this workflow

packages/core/core-flows/src/order/utils/maybe-refresh-shipping-methods.ts
```diff
@@ -0,0 +1,170 @@
...
+            {
+              id: shippingOption.id as string,
+              optionData: shippingOption.data,
+              context: {
+                ...order,
+                ...input.context,
+                from_location:
+                  shippingOption.service_zone.fulfillment_set.location,
+              },
+              // data: {}, // TODO: add data
```
olivermrbl, line 133: question: What should we do here?

packages/core/core-flows/src/order/utils/maybe-refresh-shipping-methods.ts
```diff
@@ -0,0 +1,170 @@
...
+            },
+          ]
+        }
+      )
+
+      parallelize(
+        updateOrderChangeActionsStep([updateData[0]]),
+        updateOrderShippingMethodsStep([updateData[1]!])
+      )
+    })
```
olivermrbl, line 169: question: Not sure I understand this logic, could I get you to explain?

packages/core/core-flows/src/order/workflows/return/create-return-shipping-method.ts
fPolic: good catch, initially it was used but then forgot to remove

packages/core/core-flows/src/order/utils/maybe-refresh-shipping-methods.ts
fPolic: Atm we are not handling fulfillment data validation for shipping methods across the RMA flows, we just copy shipping option data. Should we revisit this separately?

packages/core/core-flows/src/order/workflows/exchange/refresh-shipping.ts
fPolic: Good point, `maybeRefreshShippingMethodsWorkflow` would need to be updated to handle an array of methods. We cannot add more than one currently from the Admin, but we should support multiple methods in theory. Let me know if you think we should tackle this in this one or in a follow-up.

packages/core/core-flows/src/order/utils/maybe-refresh-shipping-methods.ts
fPolic: So if the shipping option is calculated, we update the amount on the method and the associated order change. Not sure if that answers your question?

packages/core/core-flows/src/order/workflows/exchange/refresh-shipping.ts
olivermrbl: Fine to keep for a follow up iteration.

packages/core/core-flows/src/order/utils/maybe-refresh-shipping-methods.ts
olivermrbl: Let's revisit this separately. Can you create a ticket for this?

olivermrbl (approved):

packages/core/core-flows/src/order/utils/maybe-refresh-shipping-methods.ts
fPolic: yup, created CMRC-987

PR #11534, opened 2025-02-19
chore(cli): Prevent swallowing error in non production env

adrien2p:
**What**
It happened many times that people can't run their project locally because an error is thrown during the config file loading. It occurs that this error can often happen because of some missing deps such as `swc/core`. But since the error is being swallowed, the import API will look for the js file which often does not exists (as it has not been built) and lead to hard debugging situation.

In order to prevent that, we could warn about any error that might occur when running the CLI as long as we are not in production. this error can be ignored if ts-node is not of any use in particular scenarios

### DIFF
--- a/packages/cli/medusa-cli/cli.js
+++ b/packages/cli/medusa-cli/cli.js
@@ -3,6 +3,14 @@
 try {
   require("ts-node").register({})
   require("tsconfig-paths").register({})
-} catch {}
+} catch (e) {
+  const isProduction = process.env.NODE_ENV === "production"
+  if (!isProduction) {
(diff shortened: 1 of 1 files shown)


sgirones (approved):

carlos-r-l-rodrigues (approved):

adrien2p:
@olivermrbl @thetutlage if I could get your opinion on this?

olivermrbl (approved):

olivermrbl:
Think this is nice, very helpful!

thetutlage:
Looks good to me

PR #11538, opened 2025-02-19
chore(core-flows): reserve inventory from available location

carlos-r-l-rodrigues:
What:
 - If multiple stock locations are available on the sales channel, the inventory is checked only on location where available inventory >= required quantity * quantity.


packages/core/core-flows/src/cart/utils/prepare-confirm-inventory-input.ts
```diff
@@ -67,6 +80,22 @@ export const prepareConfirmInventoryInput = (data: {
...
+        const availabilty = MathBN.sub(
+          location_levels.raw_stocked_quantity ??
+            location_levels.stocked_quantity ??
+            0,
+          location_levels.raw_reserved_quantity ??
+            location_levels.reserved_quantity ??
+            0
+        )
+
+        mapLocationAvailability.set(
```
fPolic: q: if the variant has an inventory kit i.e. multiple inventory items, wouldn't this set availability, for that location, for the last item only?

integration-tests/modules/__tests__/cart/store/cart.workflows.spec.ts
```diff
@@ -844,6 +845,168 @@ medusaIntegrationTestRunner({
...
+          ])
+
+          // complete 2 carts
+          for (let i = 1; i <= 2; i++) {
+            const cart = await cartModuleService.createCarts({
+              currency_code: "usd",
+              sales_channel_id: salesChannel.id,
+            })
+
+            console.log(`Cart ${i}`, cart.id)
```
fPolic: nit: log

packages/core/core-flows/src/cart/utils/prepare-confirm-inventory-input.ts
carlos-r-l-rodrigues: this is a good point. We need to map by item_id

fPolic (approved):
LGTM

PR #11544, opened 2025-02-20
fix(): handle empty q filters - allow to query deleted records from graph API - staled_at fixes

adrien2p:
**What**
It is currently not possible to provide the `withDeleted` flag to graph queries making it not possible to query for soft deleted record and hence the index engine also can't retrieve the data to be sync on deleted/detached events and making the index only relying on the application start sync.

This pr update the graph API to allow the consumer to provide the `withDeleted` flag and the remote query will now take that flag to be send to the underlying queried module allowing the querier to retrieve soft deleted items.

This pr also include a fix in how the index query builder is checking for `q` filter

This pr also include a fix on the staled_at management

this pr also fix a wrong import from `dist`


adrien2p:
[WIP] I still have one thing to figure out

olivermrbl (approved):

Issue #11559, opened 2025-02-21
[Bug]: Error when using `createOrderEditShippingMethodWorkflow` - "OrderClaim id not found"


JohnGemstone:
### What happended?

I've been trying to update the shipping methods for existing draft orders and came across the following error after fixing #11558

When using:
`createOrderEditShippingMethodWorkflow`

I get the following error:
`OrderClaim id not found: order_01XXXX`

```
Error: OrderClaim id not found: order_01JMMRNEJPCS2RFQCBWB3H1B28\n' +
    '    at RemoteJoiner.checkIfKeysExist (medusa/node_modules/@medusajs/orchestration/src/joiner/remote-joiner.ts:556:13)\n' +
    '    at RemoteJoiner.fetchData (medusa/node_modules/@medusajs/orchestration/src/joiner/remote-joiner.ts:493:10)\n' +
    '    at processTicksAndRejections (node:internal/process/task_queues:105:5)\n' +
    '    at async RemoteJoiner.query (medusa/node_modules/@medusajs/orchestration/src/joiner/remote-joiner.ts:1230:22)\n' +
    '    at async RemoteQuery.query (medusa/node_modules/@medusajs/modules-sdk/src/remote-query/remote-query.ts:289:12)\n' +
    '    at async Query.query (medusa/node_modules/@medusajs/modules-sdk/src/remote-query/query.ts:126:12)\n' +
    '    at async Object.<anonymous> (medusa/node_modules/@medusajs/core-flows/src/common/steps/use-remote-query.ts:135:22)\n' +
    '    at async Object.invoke (medusa/node_modules/@medusajs/workflows-sdk/src/utils/composer/helpers/create-step-handler.ts:72:52)\n' +
    '    at async DistributedTransaction.handler (medusa/node_modules/@medusajs/orchestration/src/workflow/workflow-manager.ts:214:16)\n' +
    '    at async stepHandler (medusa/node_modules/@medusajs/orchestration/src/transaction/transaction-orchestrator.ts:827:20)\n' +
    '⮑ sat medusa/apps/backend/src/workflows/order/workflows/create-order-edit-shipping-method.ts: [create-order-edit-shipping-method2 -> order-query (invoke)]
```

This also affects the api endpoint for order edits shipping methods e.g:
`/admin/order-edits/order_01XXXX/shipping-method`

The error is due to the wrong entity being queried for the workflow, it is currently looking at "order_claim" rather than "orders"

I have fixed and will create a pull request,

Thanks!

Issue #11558, opened 2025-02-21
[Bug]: error when using `createOrderEditShippingMethodWorkflow` - prepare-shipping-method


JohnGemstone:
### What happended?

I've been trying to update the shipping methods for existing draft orders and came across the following error.

When using:
`createOrderEditShippingMethodWorkflow`

I get the following error:
`Cannot read properties of undefined (reading 'order_id')`

```
{
  message: "Cannot read properties of undefined (reading 'order_id')",
  name: 'TypeError',
  stack: "TypeError: Cannot read properties of undefined (reading 'order_id')\n" +
    '    at Function.<anonymous> (medusa-project/node_modules/@medusajs/core-flows/src/order/utils/prepare-shipping-method.ts:21:36)\n' +
    '    at Proxy.returnFn (medusa-project/node_modules/@medusajs/workflows-sdk/src/utils/composer/transform.ts:192:30)\n' +
    '    at processTicksAndRejections (node:internal/process/task_queues:105:5)\n' +
    '    at async resolveProperty (medusa-project/node_modules/@medusajs/workflows-sdk/src/utils/composer/helpers/resolve-value.ts:17:11)\n' +
[...]
    '    at async unwrapInput (medusa-project/node_modules/@medusajs/workflows-sdk/src/utils/composer/helpers/resolve-value.ts:50:14)\n' +
    '    at async unwrapInput (medusa-project/node_modules/@medusajs/workflows-sdk/src/utils/composer/helpers/resolve-value.ts:65:26)\n' +
    '    at async resolveValue (medusa-project/node_modules/@medusajs/workflows-sdk/src/utils/composer/helpers/resolve-value.ts:79:7)\n' +
    '⮑ sat medusa-project/node_modules/@medusajs/core-flows/dist/order/workflows/order-edit/create-order-edit-shipping-method.js: [create-order-edit-shipping-method -> create-order-shipping-methods (invoke)]'
}
```

This also affects the api endpoint for order edits shipping methods e.g:
`/admin/order-edits/order_01XXXX/shipping-method`

As you can see the error occurs inside the transformer `prepare-shipping-method`.

I have fixed and will create a pull request,

Thanks!

JohnGemstone:
Actually there's a pull request already open for this fix
https://github.com/medusajs/medusa/pull/11504

olivermrbl:
Thanks for highlighting. I'll close this, as this will be part of the next release.

PR #11560, opened 2025-02-21
fix(core-flows): Allow create shipping methods through order edits

JohnGemstone:
Fixes for #11559 and #11558

Noticed there's a recent pull request https://github.com/medusajs/medusa/pull/11504 which fixes what's passed to `prepare-shipping-method`.

As for updating the query for `CreateOrderEditShippingMethodWorkflowInput` entry point from `order_claim` to `orders`, it seems like this has already been fixed?? 

For my own education, how can github say I'm two commits ahead from the main `develop` branch but show what I'm merging into is different from the codebase...
See:
https://github.com/medusajs/medusa/pull/11560/commits/9fb265c169834bb74f220745b6a0a08dfea79be1

I can't even seem to find when this line was changed:
[/packages/core/core-flows/src/order/workflows/order-edit/create-order-edit-shipping-method.ts#L118](https://github.com/medusajs/medusa/blob/develop/packages/core/core-flows/src/order/workflows/order-edit/create-order-edit-shipping-method.ts#L118)

What has happened?

### DIFF
--- a/packages/core/core-flows/src/order/utils/prepare-shipping-method.ts
+++ b/packages/core/core-flows/src/order/utils/prepare-shipping-method.ts
@@ -18,7 +18,7 @@ export function prepareShippingMethod(relatedEntityField?: string) {
       data: option.data ?? {},
       name: option.name,
       version: orderChange.version,
-      order_id: data.relatedEntity.order_id,
+      order_id: data.input.order_id,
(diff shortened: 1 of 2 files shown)


carlos-r-l-rodrigues:
cc: @fPolic

PR #11585, opened 2025-02-24
chore(product): revamp upsertWithReplace and Remove its usage from product creation

adrien2p:
**What**
- Move create product to use native create by structuring the data appropriately, it means no more `upsertWithReplace` being very poorly performant and got 20x better performances on staging
- Improvements in `upsertWithReplace` to still get performance boost for places that still relies on it. Mostly bulking the operations when possible


carlos-r-l-rodrigues (approved):

packages/modules/product/src/services/product-module-service.ts
```diff
@@ -1568,75 +1569,82 @@ export default class ProductModuleService
...
+        ;(product as any).tags = (product as any).tags.map(
+          (tag: { id: string }) => {
+            const existingTag = existingTagsMap.get(tag.id)
+            if (existingTag) {
+              return existingTag
+            }
+
+            throw new MedusaError(
+              MedusaError.Types.INVALID_DATA,
+              `Tag with id ${tag.id} not found. Please create the tag before associating it with the product.`
```
olivermrbl, line 1633: question: Was this not possible previous?

adrien2p: Maybe in an uncrontrolled way, the api only allows tag_ids and the normalisation takes the tag ids and assign them on tags in the current implementation

adrien2p: If you want we could add the support for it properly in a controlled manner.

Note: it would create a change in the front facing api though

olivermrbl: If it is not too much work, I'd prefer that to avoid a theoretical breaking change (even though it might not affect many)

adrien2p: They would have had a typescript error if they were passing tags instead of tag ids, so unless they ignore the error they wouldn't have been able to pass the build

olivermrbl: As per internal discussion, let's keep it as is, as this is more a correction of an inconstency.

PR #11589, opened 2025-02-24
Update configuration.mdx

grandpaSam:
I changed the description of the backend field of admin dashboard config. 

Originally It said that it is only used if serve is false. This is not accurate as you can put a value there with server true and the admin dashboard will direct requests to the set url. 

In my opinion this should also remain as you need to change the backend value if your dev environment is on a remote server and I would much prefer if this worked while the server option is true


shahednasser (approved):
LGTM thanks!

PR #11592, opened 2025-02-25
feat: add routes loader

thetutlage:
Fixes: FRMW-2919

This PR adds a new routes loader with a single responsibility of scanning the filesystem and collecting routes. Sorting of routes, merging middleware and registering them with express are going to separate implementations.

The new `RoutesLoader` class allows overriding routes as-well (not recommended though) and this is how routes are de-duplicated.

- When two routes for the exact route pattern/matcher are discovered, the routes loader will only keep the last one.

- Routes files can also override handlers for specific HTTP methods. For example, the original route file exported handlers for `GET` and `POST`, but the overriding one only defines `GET`. In that case, we will continue using the original implementation for the `POST` handler.

- If an overriding route file exports additional configuration like `export const AUTHENTICATION=false`, then this will only impact the handlers exported from this file and not the original handlers.

Routes sorting has been already been implemented in a separate PR and you can visualize it using this URL. https://routes-visualizer.fly.dev/


adrien2p (approved):
Perfect 💪

PR #11601, opened 2025-02-25
chore(): Improve internal repository delete algo

adrien2p:
**What**
Improve internal native delete performance by removing the need to prefecth data before deleting them and rather delete them and return the id from the id


olivermrbl (approved):
LGTM, nice one

PR #11602, opened 2025-02-25
chore(core-utils): avoid overfetching to refresh cart

carlos-r-l-rodrigues:
What:
 * Not all Cart operations need a full refresh updating items. This PR introduces a flag to force the refresh for special ocasions, like updating the Cart's region, or transfering the Cart to another customer. For all other flows it will update only promotions, taxes and payment collection if needed.


adrien2p (approved):

packages/core/core-flows/src/cart/workflows/refresh-cart-items.ts
```diff
@@ -101,62 +101,59 @@ export const refreshCartItemsWorkflow = createWorkflow(
...
         },
       }).config({ name: "fetch-variants" })
-    })
 
-    validateVariantPricesStep({ variants })
-
-    const validate = createHook("validate", {
-      input,
-      cart,
-    })
```
olivermrbl, line 111: thought: Was this intentionally removed?

olivermrbl (commented):
What about add to cart, update line item in cart, and remove item from cart? I would imagine these can trigger promotion changes.

packages/core/core-flows/src/cart/workflows/refresh-cart-items.ts
carlos-r-l-rodrigues: Yeah, this was intentional. That hook is present in all workflows where this one is called as a step.
Here we are not mutating the cart per se, I believe the hook here is not valid.

PR #11629, opened 2025-02-26
fix(dashboard): filter outbound shipping options in RMA

fPolic:
**What**
- filter out return shipping options for Claims/Exchanges
- fix the incorrect display of "missing inventory" warning

---

CLOSES CMRC-946
CLOSES CMRC-947

### DIFF
--- a/packages/admin/dashboard/src/routes/orders/order-create-exchange/components/exchange-create-form/exchange-outbound-section.tsx
+++ b/packages/admin/dashboard/src/routes/orders/order-create-exchange/components/exchange-create-form/exchange-outbound-section.tsx
@@ -30,6 +30,7 @@ import { ItemPlaceholder } from "../../../order-create-claim/components/claim-cr
 import { AddExchangeOutboundItemsTable } from "../add-exchange-outbound-items-table"
 import { ExchangeOutboundItem } from "./exchange-outbound-item"
 import { CreateExchangeSchemaType } from "./schema"
+import { log } from "console"
 
 type ExchangeOutboundSectionProps = {
   order: AdminOrder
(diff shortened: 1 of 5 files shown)


packages/admin/dashboard/src/routes/orders/order-create-exchange/components/exchange-create-form/exchange-outbound-section.tsx
```diff
@@ -62,6 +63,13 @@ export const ExchangeOutboundSection = ({
     fields: "*prices,+service_zone.fulfillment_set.location.id",
   })
 
+  const outboundShippingOptions = shipping_options.filter(
+    (shippingOption) =>
+      !!shippingOption.rules.find(
+        (r) => r.attribute === "is_return" && r.value === "false"
```
olivermrbl, line 69: question: Would it make sense to allow a `rules` filter in the API Route, so we can filter on the backend?

fPolic: You mean like `is_return` flag that we have on the store endpoint? or how would the query look like?

olivermrbl: Hmm, forget what I said, I don't think we can easily create a `rules` filter for the endpoint.

Basically, what I had in mind was something like:
```
GET /admin/shipping-options?rules[attribute]=is_return&rules[value]=false
```

fPolic: Yeah, also issue is if we want to filter by `enabled_in_store` as well in the same query. I can open a ticket to revisit filtering options on the admin endpoint.

olivermrbl (approved):

PR #11630, opened 2025-02-26
fix(medusa): deleting location level

fPolic:
**What**
- when deleting an inventory location level, a wrong RQ argument would result in a wrong location level being returned


packages/medusa/src/api/admin/inventory-items/[id]/location-levels/[location_id]/route.ts
```diff
@@ -27,8 +27,10 @@ export const DELETE = async (
       remoteQueryObjectFromString({
         entryPoint: "inventory_level",
         variables: {
-          inventory_item_id: id,
-          location_id,
+          filters: {
```
Comment (olivermrbl) [inline packages/medusa/src/api/admin/inventory-items/[id]/location-levels/[location_id]/route.ts]: suggestion: Should we change this to use `query` now that we are at it?

olivermrbl (approved):
LGTM

PR #11632, opened 2025-02-26
chore: prevent workflow steps to call modules when not necessary

adrien2p:
**What**
Some steps were calling the modules even when nothing was needed which for some operations would create transaction for nothing leading to extra execution time that add up very quickly on cloud network


carlos-r-l-rodrigues (approved):

olivermrbl (approved):
LGTM 💪

PR #11638, opened 2025-02-27
feat: add middleware-file-loader

thetutlage:
Fixes: FRMW-2920

The feature is called `middleware-file-loader`, because it does more than just collecting middlewares. It also collects the bodyParser config for routes


packages/core/framework/src/http/middleware-file-loader.ts
```diff
@@ -0,0 +1,153 @@
...
+              route,
+              null,
+              2
+            )}`
+          )
+        }
+
+        const matcher = String(route.matcher)
+        if ("bodyParser" in route && route.bodyParser !== undefined) {
+          result.bodyParserConfigRoutes[matcher] = {
```
adrien2p: suggestion: We should probably handle it per matcher and method as there is scenarios where the config is dependent on the route verb

packages/core/framework/src/http/middleware-file-loader.ts
```diff
@@ -0,0 +1,153 @@
...
+import type {
+  MiddlewaresConfig,
+  ScannedMiddlewareDescriptor,
+  BodyParserConfigRoutes,
+} from "./types"
+
+/**
+ * File name that is used to indicate that the file is a middleware file
+ */
+const MIDDLEWARE_NAME = "middlewares"
```
adrien2p:
```suggestion
const MIDDLEWARE_FILE_NAME = "middlewares"
```

packages/core/framework/src/http/types.ts
```diff
@@ -111,6 +111,20 @@ export type FileSystemRouteDescriptor = ScannedRouteDescriptor & {
   relativePath: string
 }
 
+export type ScannedMiddlewareDescriptor = {
+  matcher: string
+  method?: MiddlewareVerb | MiddlewareVerb[]
+  handler: MiddlewareFunction
+}
+
+export type BodyParserConfigRoutes = {
```
adrien2p: suggestion: This type might need some adjustments based on previous comments

adrien2p (approved):
Excellent, just a few comments 💪

packages/core/framework/src/http/middleware-file-loader.ts
thetutlage: Converted to an array instead

packages/core/framework/src/http/middleware-file-loader.ts
thetutlage: Done

packages/core/framework/src/http/types.ts
thetutlage: Converted to an array instead

Issue #11620, opened 2025-02-26
Global place for handling errors


namlt3820:
### What Medusa version and documentation are you using?

v2

### Preliminary Checks

- [x] This issue is not a duplicate. Before opening a new issue, please search existing issues: https://github.com/medusajs/medusa/issues

### Issue Summary

At first, I thought I could use middleware as a central place to catch all errors for validation, formatting, and sending them to third parties.
```
export default defineMiddlewares({
  errorHandler: (err, req, res, next) => {
    console.log({ err });
  },
});
```
However, it turns out that service-level errors (such as those from the cart, sale channel, product, etc.) do not appear here. Is there a dedicated place to handle these types of errors? Thank you.

### Are you interested in working on this issue?

- [ ] I would like to fix this issue

thetutlage:
Fixed in https://github.com/medusajs/medusa/pull/11646

Issue #11595, opened 2025-02-25
[Bug]: Custom Middleware/Validation calls report body as empty when using predefined api routes


Miguel-Absa:
### Package.json file

```JSON
{
  "name": "stock_up",
  "version": "0.0.1",
  "description": "A starter for Medusa projects.",
  "author": "Medusa (https://medusajs.com)",
  "license": "MIT",
  "keywords": [
    "sqlite",
    "postgres",
    "typescript",
    "ecommerce",
    "headless",
    "medusa"
  ],
  "scripts": {
    "build": "medusa build",
    "seed": "medusa exec ./src/scripts/seed.ts",
    "start": "medusa start",
    "dev": "medusa develop -p 9001",
    "test:integration:http": "TEST_TYPE=integration:http NODE_OPTIONS=--experimental-vm-modules jest --silent=false --runInBand --forceExit",
    "test:integration:modules": "TEST_TYPE=integration:modules NODE_OPTIONS=--experimental-vm-modules jest --silent --runInBand --forceExit",
    "test:unit": "TEST_TYPE=unit NODE_OPTIONS=--experimental-vm-modules jest --silent --runInBand --forceExit"
  },
  "dependencies": {
    "@medusajs/admin-sdk": "2.5.0",
    "@medusajs/cli": "2.5.0",
    "@medusajs/framework": "2.5.0",
    "@medusajs/medusa": "2.5.0",
    "@mikro-orm/core": "6.4.3",
    "@mikro-orm/knex": "6.4.3",
    "@mikro-orm/migrations": "6.4.3",
    "@mikro-orm/postgresql": "6.4.3",
    "@opentelemetry/exporter-metrics-otlp-grpc": "^0.57.2", 
    "@opentelemetry/exporter-trace-otlp-grpc": "^0.57.2",  
    "@opentelemetry/instrumentation-pg": "^0.51.0",
    "@opentelemetry/resources": "^1.30.1",
    "@opentelemetry/sdk-node": "^0.57.2",
    "@opentelemetry/sdk-trace-node": "^1.30.1",
    "awilix": "^8.0.1",
    "pg": "^8.13.0"
  },
  "devDependencies": {
    "@medusajs/test-utils": "2.5.0",
    "@mikro-orm/cli": "6.4.3",
    "@swc/core": "1.5.7",
    "@swc/jest": "^0.2.36",
    "@types/jest": "^29.5.13",
    "@types/node": "^20.0.0",
    "@types/react": "^18.3.2",
[...]

Issue #11355, opened 2025-02-07
[Bug]: possible zod validation bug


yeshuman:
### Package.json file

```JSON
{
  "name": "medusa-starter-default",
  "version": "0.0.1",
  "description": "A starter for Medusa projects.",
  "author": "Medusa (https://medusajs.com)",
  "license": "MIT",
  "keywords": [
    "sqlite",
    "postgres",
    "typescript",
    "ecommerce",
    "headless",
    "medusa"
  ],
  "scripts": {
    "build": "medusa build",
    "seed": "medusa exec ./src/scripts/seed.ts",
    "start": "medusa start",
    "dev": "medusa develop",
    "inspect": "node --inspect ./node_modules/@medusajs/medusa/dist/index.js",
    "test:integration:http": "TEST_TYPE=integration:http NODE_OPTIONS=--experimental-vm-modules jest --silent=false --runInBand --forceExit",
    "test:integration:modules": "TEST_TYPE=integration:modules NODE_OPTIONS=--experimental-vm-modules jest --silent --runInBand --forceExit",
    "test:unit": "TEST_TYPE=unit NODE_OPTIONS=--experimental-vm-modules jest --silent --runInBand --forceExit"
  },
  "dependencies": {
    "@medusajs/admin-sdk": "2.4.0",
    "@medusajs/cli": "2.4.0",
    "@medusajs/framework": "2.4.0",
    "@medusajs/medusa": "2.4.0",
    "@mikro-orm/core": "6.4.3",
    "@mikro-orm/knex": "6.4.3",
    "@mikro-orm/migrations": "6.4.3",
    "@mikro-orm/postgresql": "6.4.3",
    "awilix": "^8.0.1",
    "pg": "^8.13.0"
  },
  "devDependencies": {
    "@faker-js/faker": "^9.4.0",
    "@medusajs/test-utils": "2.4.0",
    "@mikro-orm/cli": "6.4.3",
    "@swc/core": "1.5.7",
    "@swc/jest": "^0.2.36",
    "@types/jest": "^29.5.13",
    "@types/node": "^20.0.0",
    "@types/react": "^18.3.2",
    "@types/react-dom": "^18.2.25",
    "jest": "^29.7.0",
    "prop-types": "^15.8.1",
    "react": "^18.2.0",
    "react-dom": "^18.2.0",
    "ts-node": "^10.9.2",
    "typescript": "^5.6.2",
    "vite": "^5.2.11",
[...]

PR #11646, opened 2025-02-27
feat: Replace existing router with the new implementation

thetutlage:
Fixes: FRMW-2925

This PR revamps the internal implementation of our HTTP router to be more predicatable and fixes some long pending issues like https://github.com/medusajs/medusa/issues/11595, https://github.com/medusajs/medusa/issues/11355, and https://github.com/medusajs/medusa/issues/11620

The root of the issues with the old implementation were mainly around the following decisions.

- Plugins, Medusa core and the application used isolated router instances of Express. This means global middlewares and error handlers are isolated too. If you define an error handler in your app, it won't be able to handle errors from the core routes.

- Middlewares defined inside the `src/api/middlewares.ts` file were always defined before the filesystem routes and this lead to unexpected behavior in many scanerios. For example, the middleware handler registered for route `admin/payments/payment-providers` will also run when a request comes in for `/admin/payments/:id`, even though there is a static route in place for the `/admin/payments/payment-providers` URI.

- There was no clarity on the ordering in which routes and middleware were executed. With the new implementation, this has been fixed (more on ordering is discussed below).

- Overriding an existing route defines multiple express handlers under the hood. It worked previously, because the route handlers never call `next`, hence the previous implementation was never executed. With the newer implementation, the route overrides from a Medusa application will be registered over the routes from core or the plugins.

To resolve these issues, we’ve rebuilt the loader from scratch to ensure predictable behavior in how routes and middleware are registered within the application. Additionally, the old API Routes loader unintentionally allowed overriding core API routes and middleware. While this behavior remains for now, it will be removed in a future release. Please prepare your application based on the updated behavior outlined below.

**Global middleware**

A global middleware is defined using the `defineMiddlewares` function without specifying an HTTP method. 
```
defineMiddlewares([
  {
    matcher: '/admin',
    middleware: [
      (req, res, next) => console.log("Hi from global middleware")
    ],
  }
])
```

The order of registering global middleware is as follows:
1. Global middleware from the core
2. Global middleware from plugins (in the order plugins are registered)
3. Global middleware from the app

**Route middleware**

A route middleware is also defined using the `defineMiddlewares` method. However, they must always specify the HTTP method(s) they want to target.
```
defineMiddlewares([
  {
    matcher: '/admin/*',
    methods: ["GET", "POST"],
    middleware: [
      (req, res, next) => console.log("Hi from route middleware")
    ],
  }
])
```

The order of registering route middleware is as follows:
1. Route middleware from the core
2. Route middleware from plugins (in the order plugins are registered)
3. Route middleware from the app

**API Routes**

An API Route is always attached to a method, decided by the exported const from the route file. API Routes are registered by scanning the filesystem.

Two types of API Routes are supported:
1. Static segments: `/admin/products/route.ts`
2. Dynamic segments: `/admin/products/[id]/route.ts`

**Sorting routes and middleware**

Before registering routes + middleware with Express, they are sorted as described below:

1. Global middleware
2. Route middleware
3. API Routes

On top of this, each segment from the API Route matcher is sorted by its specificity:

1. Wildcard segments, e.g. `/admin/products/*`
2. Regex segments, e.g. `/admin/(product-types|product-collections)`
3. Static segments, e.g. `/admin/products`
4. Dynamic segments, e.g. `/admin/products/:id`

**Important note about middleware**
Middleware **never overrides any existing middleware**; they are added to the stack of existing middleware. For example, if you define a custom validation middleware on an existing route, then both the original and the custom validation middleware will run.


packages/core/framework/src/http/utils/unless-path.ts
```diff
@@ -14,6 +14,7 @@ import {
 export const unlessPath =
   (onPath: RegExp, middleware: MiddlewareFunction) =>
   (req: MedusaRequest, res: MedusaResponse, next: MedusaNextFunction) => {
+    console.log({ onPath, path: req.path })
```
carlos-r-l-rodrigues: rm

packages/medusa/src/instrumentation/index.ts
```diff
@@ -66,72 +66,72 @@ export function instrumentHttpLayer() {
    * Instrumenting the route handler to report traces to
    * OpenTelemetry
    */
-  ApiRoutesLoader.traceRoute = (handler) => {
-    return async (req, res) => {
-      if (shouldExcludeResource(req.originalUrl)) {
-        return await handler(req, res)
-      }
+  // ApiRoutesLoader.traceRoute = (handler) => {
```
carlos-r-l-rodrigues: will this be removed?

packages/medusa/src/api/admin/workflows-executions/middlewares.ts
```diff
@@ -44,18 +44,18 @@ export const adminWorkflowsExecutionsMiddlewares: MiddlewareRoute[] = [
   },
   {
     method: ["POST"],
-    matcher: "/admin/workflows-executions/:id/run",
+    matcher: "/admin/workflows-executions/:workflow_id/run",
```
carlos-r-l-rodrigues, line 47: nice catch

packages/core/framework/src/http/utils/unless-path.ts
thetutlage: Yo!

packages/medusa/src/instrumentation/index.ts
thetutlage: Nope. Have to add it back to the new implementation. That's the next thing I will do now

integration-tests/http/package.json
```diff
@@ -6,7 +6,7 @@
   "private": true,
   "scripts": {
     "test:integration": "NODE_OPTIONS=--experimental-vm-modules jest --no-cache --maxWorkers=50% --bail --detectOpenHandles --forceExit --logHeapUsage",
-    "test:integration:chunk": "NODE_OPTIONS=--experimental-vm-modules jest --silent --no-cache --bail --maxWorkers=50% --forceExit --testPathPattern=$(echo $CHUNKS | jq -r \".[${CHUNK}] | .[]\")",
+    "test:integration:chunk": "NODE_OPTIONS=--experimental-vm-modules jest --silent=false --no-cache --bail --maxWorkers=50% --forceExit -- __tests__/shipping-option/admin/shipping-option.spec.ts",
```
carlos-r-l-rodrigues: revert

carlos-r-l-rodrigues (approved):

packages/core/framework/src/http/__tests__/middleware-file-loader.spec.ts
```diff
@@ -14,12 +14,20 @@ describe("Middleware file loader", () => {
...
+            "DELETE",
+            "OPTIONS",
+            "HEAD",
+          ],
         },
         {
           "config": false,
           "matcher": "/webhooks/*",
-          "method": "POST",
+          "methods": "POST",
```
adrien2p: suggestion: I think it would be better to normalize the expected type according to the naming (plural) to always be an array of string. We could deprecate (but still support for the time being) string

packages/core/framework/src/http/__tests__/routes-loader.spec.ts
```diff
@@ -4,177 +4,194 @@ import { RoutesLoader } from "../routes-loader"
...
-          "optedOutOfAuth": false,
-          "relativePath": "/customers/route.ts",
-          "route": "/customers",
-          "shouldAppendAdminCors": false,
-          "shouldAppendAuthCors": false,
-          "shouldAppendStoreCors": false,
-        },
-      ]
+      {
+        "list": [
```
adrien2p: suggestion: should we get back to a flat array instead of an object?

packages/core/framework/src/http/__tests__/routes-loader.spec.ts
```diff
@@ -184,189 +201,207 @@ describe("Routes loader", () => {
...
-          "optedOutOfAuth": false,
-          "relativePath": "/store/[customer_id]/orders/[order_id]/route.ts",
-          "route": "/store/:customer_id/orders/:order_id",
-          "shouldAppendAdminCors": false,
-          "shouldAppendAuthCors": false,
-          "shouldAppendStoreCors": true,
-        },
-      ]
+      {
+        "list": [
```
adrien2p: see previous suggestion

packages/core/framework/src/http/middlewares/bodyparser.ts
```diff
@@ -0,0 +1,96 @@
...
+        )
+        const parserMiddleware = parsers[parser]
+
+        if (!matchingRoute) {
+          return parserMiddleware()(req, res, next)
+        }
+
+        if (matchingRoute.config === false) {
+          logger.debug(
+            `skipping ${parser} bodyparser middleware ${req.method}${req.path}`
```
adrien2p: nit: add space between method and path

packages/core/framework/src/http/middlewares/bodyparser.ts
```diff
@@ -0,0 +1,96 @@
...
+          )
+          return next()
+        }
+
+        logger.debug(
+          `using custom ${parser} bodyparser config ${req.method}${req.path}`
```
adrien2p: nit: same as above

packages/core/framework/src/http/router.ts
```diff
@@ -288,773 +45,344 @@ export class ApiRoutesLoader {
...
+      )
+    )
 
-      const extension = extname(route_)
-      if (extension) {
-        route_ = route_.replace(extension, "")
-      }
+    const routes = routesLoader.getRoutes()
+    return {
+      routes: routes.list,
```
adrien2p: suggestion: to be aligned upon the acceptation of previous suggestion

packages/core/framework/src/http/router.ts
```diff
@@ -288,773 +45,344 @@ export class ApiRoutesLoader {
...
-          const handlers = Object.keys(import_).filter((key) => {
-            /**
-             * Filter out any export that is not a function
-             */
-            return typeof import_[key] === "function"
+  #registerExpressHandler(
+    route: MiddlewareDescriptor | RouteDescriptor | RouteDescriptor
+  ) {
+    if ("isRoute" in route) {
+      logger.debug(`registering route ${route.methods}:${route.matcher}`)
```
adrien2p: nit: should we unified the logs to either use a space of column between method and path/matcher

packages/core/framework/src/http/router.ts
```diff
@@ -288,773 +45,344 @@ export class ApiRoutesLoader {
...
+    methods.forEach((method) => {
+      logger.debug(`registering route middleware ${method}:${route.matcher}`)
+      const handler = ApiLoader.traceMiddleware
+        ? (ApiLoader.traceMiddleware(wrapHandler(route.handler), {
+            route: route.matcher,
+            method: method,
+          }) as RequestHandler)
+        : wrapHandler(route.handler)
+
+      this.#app[method.toLowerCase()](route.matcher, handler)
```
adrien2p, line 141: nit: align method <-> methods naming

packages/core/framework/src/http/router.ts
```diff
@@ -288,773 +45,344 @@ export class ApiRoutesLoader {
...
-   * Applies middleware that checks if a valid publishable key is set on store request
+   * Assigns global cors middleware for a given prefix
    */
-  applyStorePublishableKeyMiddleware(route: string | RegExp) {
-    let middleware = ensurePublishableApiKeyMiddleware as unknown as
-      | RequestHandler
-      | MiddlewareFunction
+  #applyCorsMiddleware(
+    routesFinder: RoutesFinder<RouteDescriptor>,
+    route: string,
```
adrien2p: nit: lets rename all global middleware identifier to `namespace` instead of `route`?

packages/core/framework/src/http/routes-loader.ts
```diff
@@ -252,14 +217,20 @@ export class RoutesLoader {
...
-      (ScannedRouteDescriptor | FileSystemRouteDescriptor)[]
-    >((result, routePattern) => {
-      const methodsRoutes = this.#routes[routePattern]
-      Object.keys(methodsRoutes).forEach((method) => {
-        result.push(methodsRoutes[method])
-      })
-      return result
-    }, [])
+    return Object.keys(this.#routes).reduce<{
+      list: RouteDescriptor[]
```
adrien2p: suggestion: align to not use list

packages/core/framework/src/http/routes-matcher.ts
```diff
@@ -0,0 +1,36 @@
+import pathToRegExp from "path-to-regexp"
```
adrien2p: suggestion: if not use we can delete that?

packages/core/framework/src/http/routes-sorter.ts
```diff
@@ -244,18 +251,34 @@ export class RoutesSorter {
...
+      keyof RoutesBranch<T>,
+      keyof RoutesBranch<T>,
+      keyof RoutesBranch<T>,
+      keyof RoutesBranch<T>
+    ]
+  ) {
+    if (orderBy) {
+      this.#orderBy = orderBy
+    }
     this.#routesToProcess.map((route) => this.#processRoute(route))
```
adrien2p: suggestion: here all subsequent sort would be using the last set orderBy. As per discussion, we might pass it down to the method and either use that or the default orderBy

packages/medusa/src/api/admin/promotions/middlewares.ts
```diff
@@ -137,6 +134,10 @@ export const adminPromotionRoutesMiddlewares: MiddlewareRoute[] = [
     method: ["GET"],
     matcher: "/admin/promotions/rule-attribute-options/:rule_type",
     middlewares: [
+      (_, __, next) => {
+        console.log("specific one")
+        next()
+      },
```
adrien2p: todo: rm test :D

adrien2p (approved):
LGTM 🎉 few comment that we need to resolve but excellent work 💪

PR #11650, opened 2025-02-27
fix: Exclude the health endpoint from being logged

sradevski:
The health endpoint is almost exclusively used by the infrastructure and there is very little value in logging it. 

Let me know if you think we should handle this differently, or if we should make it configurable.


riqwan (approved):
lgtm!

carlos-r-l-rodrigues (commented):
LGTM

we will likely have a few more in the future.
but we can transform it to an array when we get there.

carlos-r-l-rodrigues (approved):

PR #11660, opened 2025-02-28
feat: added hook for completeCartWorkflow

Avia-Code:
What
This commit introduces the `cartCompleted` hook to the `completeCartWorkflow`.

Why
Currently, an order can be created in 2 ways:

- via `createOrdersWorkflow`
- via `completeCartWorkflow`

The issue with completeCartWorkflow is that it does not produce any hook with created order ID, so a user will need to come up with some workarounds

How
Added the `cartCompleted` hook and updated related logic within the workflow.

Testing

1. Create new medusa project via create-medusa-app
2. Connect this custom branch of medusa to it
3. Create `src/workflows/hooks/cart-completed.ts` file with the following content:

```
import { completeCartWorkflow } from "@medusajs/medusa/core-flows"

completeCartWorkflow.hooks.cartCompleted(
  (async ({ order_id, additional_data }, { container }) => {
    console.log({order_id})
  })
)
```
4. Create new admin account and set up region, location, and sales_channel, create a product, go to the storefront, add the product to the cart, and complete the cart.
5. The log should be displayed in terminal


Avia-Code:
related to https://github.com/medusajs/medusa/discussions/9765

Avia-Code:
Hey @sradevski , could you please have a review ?

olivermrbl:
@Avia-Code, can you share what you need this hook for? We typically recommend a subscriber for reacting asynchronously to cart completions.

Avia-Code:
@olivermrbl we are building a multi-vendor marketplace using approach described here https://medium.com/@igorkhomenko/building-a-multivendor-marketplace-with-medusa-js-2-0-a-dev-guide-f55aec971126

We use it because the action we're performing is integral to the main flow of the core commerce feature.
And as described above, an order can be created in two ways:

via `createOrdersWorkflow`
via `completeCartWorkflow`
The issue with `completeCartWorkflow` is that it does not produce any hook with the created order ID.

That's why we specifically need a `hook`, not a `subscriber`.

panalgin:
We also need this,

We need the order id generated from this workflow and process some internal data based on workflows success or failed result. If we go with order.placed event, it would be too late, and it separates the logic to an another async operation.

A hook at the end of this completeCartWorkflow and possibly before order.placed event firing would be so helpful.

Main idea is to inbox this operation and retry the completion of cart later when some previously failed (third party) criteria is met.

Issue #11645, opened 2025-02-27
[Bug]: raw_pending_difference disappear after order edit - Can't refund


docloulou:
### Package.json file

```JSON
fresh starter
```

### Node.js version

v20

### Database and its version

15

### Operating system name and version

linux

### Browser name

chrome

### What happended?

raw_pending_difference in order.summary disappear after order edit that lead to impossible refund.

raw_pending_difference is use in the validate step for refund : https://github.com/medusajs/medusa/blob/c250de79192bda1b1cd217bcca37f458c693de1d/packages/core/core-flows/src/payment/workflows/refund-payment.ts#L64-L71

i think that happen after 2.5.1

### Expected behavior

-

### Actual behavior

-

### Link to reproduction repo

https://jam.dev/c/bd9a5ab8-5545-453c-aa68-796b59173c94

srinivasulu-narayanam:
I am also facing same issue  after upgrading to v2.5.0. After items return,I am not able to do refund.

carlos-r-l-rodrigues:
`2.6.0` fixes this

PR #11712, opened 2025-03-04
fix(order): summary raw totals

carlos-r-l-rodrigues:
What:
  - Generate `raw` properties from BigNumbers of Order summary totals

### DIFF
--- a/.changeset/thick-readers-add.md
+++ b/.changeset/thick-readers-add.md
@@ -1,0 +1,5 @@
+---
+"@medusajs/order": patch
+---
+
+fix(order): summary raw totals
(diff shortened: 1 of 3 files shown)


riqwan (approved):

adrien2p (approved):

adrien2p:
Should we run a migration just for safety?

carlos-r-l-rodrigues:
> Should we run a migration just for safety?

I don't believe this is needed. the workflow will use the numeric field and after any order change the `raw` will be generated.

docloulou:
this fix #11645 👍

PR #11713, opened 2025-03-04
fix(core-flows): add validation to check if pending_difference is present before proceeding with refund

riqwan:
what:

- adds a validation to check if pending_difference is present before proceeding with refund


adrien2p (approved):
Always good to have 👌

riqwan:
Closing as Carlos made a change on his pr

PR #11731, opened 2025-03-05
fix(dashboard): display minute instead of month

fPolic:
**What**
- display minutes instead of months
- remove manual offsetting since the server returns ISO format and JS date converts it to local time

---

CLOSES SUP-1069


packages/admin/dashboard/src/components/table/table-cells/common/date-cell/date-cell.tsx
```diff
@@ -13,10 +13,9 @@ export const DateCell = ({ date }: DateCellProps) => {
   }
 
   const value = new Date(date)
-  value.setMinutes(value.getMinutes() - value.getTimezoneOffset())
 
   const hour12 = Intl.DateTimeFormat().resolvedOptions().hour12
-  const timestampFormat = hour12 ? "dd MMM yyyy hh:MM a" : "dd MMM yyyy HH:MM"
+  const timestampFormat = hour12 ? "dd MMM yyyy hh:mm a" : "dd MMM yyyy HH:mm"
```
kasperkristensen: suggestion: Use the getFullDate function from `useDate` instead so we ensure that we use the same format as we do for all other dates

thetutlage (approved):

PR #11700, opened 2025-03-04
fix(icons): rename ArrowRight component to fix typo and update references

jaruesink:

PR #11734, opened 2025-03-05
chore(ui-preset,icons,dashboard)

kasperkristensen:
In favor of #11700


olivermrbl:
@kasperkristensen, lots of changes here, but I assume these are from Figma, so we should be good to merge?

olivermrbl (approved):

PR #11740, opened 2025-03-05
fix(workflow-engine-redis): Split the queues and respect worker mode for job executions

adrien2p:
**What**
Currently, the workflow engine redis does not make any distinction between worker modes, when starting as server, the engine listen to the queue which contains everything and try to execute the corresponding workflow which does not exists since job workflows are not loaded in server mode. Now, a dedicated queue is created for jobs and the worker is only started if the instance is not in server mode. In order to clean up the old queue, if the old queue trigger a scheduled job then it gets removed from the queue since it will get re added to the new queue by the new worker instances


packages/modules/workflow-engine-redis/src/utils/workflow-orchestrator-storage.ts
```diff
@@ -76,6 +94,11 @@ export class RedisDistributedTransactionStorage
       JobType.TRANSACTION_TIMEOUT,
     ]
 
+    const workerOptions = {
+      connection:
+        this.redisWorkerConnection /*, runRetryDelay: 100000 for tests */,
+    }
+
     this.worker = new Worker(
```
sradevski, line 102: Isn't it better for this worker/queue only run on a server that is in `shared` or `worker` mode? Like that, the server only deals with HTTP, and the worker deals with async workflows, jobs, and subscribers

packages/modules/workflow-engine-redis/src/loaders/redis.ts
```diff
@@ -59,12 +61,14 @@ export default async ({
   }
 
   container.register({
+    isWorkerMode: asValue(moduleDeclaration.worker_mode !== "server"),
```
sradevski, line 64: NIT: This is a bit confusing as the actual check we do is `isWorkerOrSharedMode`. I think we have it elsewhere like this as well so it's fine to leave it, but for a minute here I was gonna comment that we also need to check the "shared" mode 😄

sradevski (commented):
Just one question on the standard worker queue and what we think is better for it to do, otherwise LGTM

packages/modules/workflow-engine-redis/src/loaders/redis.ts
adrien2p: yes I agree, I ve just kept it that way for consistency and we can do a pass on all of them later in one shot 👍

packages/modules/workflow-engine-redis/src/utils/workflow-orchestrator-storage.ts
adrien2p: I think it is fine for workflow to be executed on the server instance as well no? I don't think you should be forced to have a second instance of medusa to run async workflows wdyt? but it is ok for jobs and subscribers. but I ll follow what the majority prefers.

For example, in the user infra that reported this issue, he would still have two instances that runs async workflows

sradevski: I think it really depends on the load - but how I see it is, if you want your server to process async workflows and other events then set it to "shared", otherwise the server should only handle http requests (so eg. listing products is more performant)

adrien2p: that would be a breaking change in the users infrastructure/api whereas it is not for jobs and subscribers since they are not loaded in server instances. I do agree with you overall as it would prevent increasing the pressure on the event loop, spare some memory and CPU resources more than increasing performances but I understand and agree with your point 👍 

@olivermrbl since this suggestion would have breaking implication on users deployment, infrastructure and API wdyt?

carlos-r-l-rodrigues (approved):

adrien2p:
@olivermrbl should we merge the fix and then subsequently do any other changes you might see fit but are not related to the issue being fixed?

olivermrbl:
Yeah, let's do it to move forward 👍

PR #11756, opened 2025-03-06
feat(workflows-sdk): Allow when then in parallelize

adrien2p:
**What**
Update typings to allow using when then inside parallelize


adrien2p:
@riqwan this should help you fix the batch workflow issue more easily

riqwan (approved):
Nice, tested on my PR - works great!

PR #11767, opened 2025-03-07
feat: add check for uniqueness when creating links with isList=false

thetutlage:
Fixes: FRMW-2907

**Error message**
CleanShot 2025-03-10 at 09 54 00@2x


packages/core/modules-sdk/src/link.ts
```diff
@@ -403,15 +479,34 @@ export class Link {
...
+      serviceLinks
+        .get(service.__definition.key)
+        ?.linksToCreate.push(fields as any)
     }
 
-    const promises: Promise<unknown[]>[] = []
+    for (const [serviceName, data] of serviceLinks) {
+      if (data.linksToValidateForUniqueness.filters.length) {
+        const service = this.modulesMap.get(serviceName)!
+        const existingLinks = await service.count({
```
sradevski: A quick Q: Wouldn't it make more sense to create a unique index on the pivot table? This check is not Atomic, so if  call the same create twice (it can happen if a user click a button twice very quickly, this is the issue we witnessed) the check might not work correctly.

I leave it up to you guys to decide how you approach it in the end, it's just that DB indices are atomic so you have some better guarantess

thetutlage: > Wouldn't it make more sense to create a unique index on the pivot table?

Yeah, we considered that. It has a bit of friction when you switch between `isList: true` and `isList: false`.

For example: All existing apps with `isList: false` having multiple entries in the pivot table will fail to start until they remove all the entries except one and then only the link module migration will apply.

Similarly switching from `isList: true` to `false` will require first fixing the database and then running the app.

If we are okay with this friction, then indeed a database level constraint is the best way to move forward.

sradevski: I'd say that should be expected. Even like this, switching from `isList: true` to `isList: false` would have an issue since the queried data results will be inconsistent (sometimes you'll get an array, and sometimes an object), and it's better for it to fail early during migration than at runtime

carlos-r-l-rodrigues: suggestion: Instead of implementing the new method `count`, maybe we should just use the method `list` with a `take: 1`, as the number of rows here is not important.

carlos-r-l-rodrigues (approved):

packages/core/modules-sdk/src/link.ts
thetutlage: Makes sense!

thetutlage: Will handle it in a separate PR as it will require creating a migration that warns instead of throwing the error and then removing itself from the completed migrations list.

thetutlage: Updated usage to `list`. Will remove the custom count method after testing the snapshot

olivermrbl (approved):

sradevski:
@thetutlage I rebased so I can test the other changes that landed on main around the workflows in one go, I hope you don't mind

PR #11771, opened 2025-03-07
fix(workflow-engines): race condition when retry interval is used

adrien2p:
what:
 * When `retryInterval` is used on async steps, it is executing the same step more than once leading to parallel executions of the same workflow. This PR fixes behavior and even if the step is started more than once, only one execution will be acknowledge to continue the workflow execution.

CLOSES: FRMW-2931


carlos-r-l-rodrigues (approved):

adrien2p:
@sradevski we endled all race conditions during concurrent execution. Only one thing we would like to check before getting it merged but after that it should be good 👍 let us know what you think

packages/modules/workflow-engine-inmemory/src/utils/workflow-orchestrator-storage.ts
```diff
@@ -148,7 +156,103 @@ export class InMemoryDistributedTransactionStorage
     }
 
     if (hasFinished) {
-      this.storage.delete(key)
+      const tenMinutes = 10 * 60 * 1000
```
sradevski: As we talked with Carlos on a separate Slack thread, unless the user has marked a workflow idempotent, we should clear it from storage as soon as it finishes.

This can of course be done as part of a separate PR, as I assume the storage removal should also be applied elsewhere.

packages/modules/workflow-engine-redis/src/utils/workflow-orchestrator-storage.ts
```diff
@@ -457,4 +465,97 @@ export class RedisDistributedTransactionStorage
       repeatableJobs.map((job) => this.jobQueue?.removeRepeatableByKey(job.key))
     )
   }
+
+  async #preventRaceConditionExecutionIfNecessary({
```
sradevski, line 465: I wonder when the day will come where we unify the in memory and redis logic and only use providers for the underlying storage 😄

sradevski (commented):
LGTM, I can test it with a snapshot when you guys are done.

packages/modules/workflow-engine-inmemory/src/utils/workflow-orchestrator-storage.ts
adrien2p: I agree, but here it is more in the scenario where you have concurrent execution due to retryal on awaiting for example where you need to have the state while the concurrent one get to that point. But we can reduce it a bit, it is for the inmemory to prevent any subsequent execution and prevent another type of race condition :)

packages/modules/workflow-engine-redis/src/utils/workflow-orchestrator-storage.ts
adrien2p: we need to do it def :D same for the event bus

packages/modules/workflow-engine-inmemory/src/utils/workflow-orchestrator-storage.ts
sradevski: Would it cause issues in the case where I do this though:
1. Execute workflow A with `transactionID: "id"`
2. After a minute, while workflow A is still running I cancel workflow A
3. Immediately after, I execute workflow A with `transactionID: "id"`
?

Basically I want to ensure the same workflow with a specific transaction ID is not running concurrently, but as soon as it is completed I should be able to execute it again with the same transaction ID

adrien2p: keep in mind that the in memory is not meant for production usage but more for testing locally (eventually) apart from that point, yes you can do that, but as soon as you will cancel it you will have the rollback kicking in and then you will be able to run it normally again using the same transaction id, but it ensure consistency between invocation and rollback. does that answer the question? :) 

just to be absolutely sure we will double check on our side 👍

sradevski: We use the in-memory for the tests so the tests won't work if what I described doesn't work the same way as the Redis engine.

The rollback is fine, that's not an issue at all. Would the `await myWorkflow(container).cancel(transaction)` resolve only once the rollback is done? or it won't wait for the rollback?

adrien2p: It works the same way as the redis does. 
for your question about the cancel, we will introduce a cancel method as part of the engine which is missing at the moment, but otherwise it works like for the run, if it is an async workflow (with async compensation) then you will have an acknowledgment otherwise it will have finished the rollback when the promise resolve.
and this cancel will be part of this pr 👍

adrien2p: @sradevski we will generate a snapshot for you to test, but we should be good now. As mentioned, we delete the memory storage immediately after finishing an execution to completion (success/reverted). Any other execution would be considered as superfluous and be discarded right away. In regards to the cancellation is everything ok in the understanding of its execution?

PR #11773, opened 2025-03-07
Update page.mdx

saranshisatgit:
This fixes some type issues that it will throw , so the  /// <reference types="vite/types/importMeta.d.ts" /> will fix this issue which is Property 'env' does not exist on type 'ImportMeta'.ts


shahednasser (changes requested):
Thank you for your contribution @saranshisatgit , but we already have documentation around this [here](https://docs.medusajs.com/learn/fundamentals/admin/environment-variables#type-error-on-importmetaenv). Maybe you can instead add a note after the code block that if you face type errors you should refer to that documentation? Something like this:

```
<Note>

If you get a type error for `import.meta.env`, refer to the [Environment Variables documentation](!docs!/learn/fundamentals/admin/environment-variables#type-error-on-importmetaenv).

</Note>
```

PR #11790, opened 2025-03-10
feat(core-flows): revisit fulfillment flow reservations and support inventory kit

fPolic:
**What**
- handle inventory kit reservations properly when creating fulfillment
- restore reservations when a fulfillment is canceled
- prevent canceling already canceled fulfillment

---

CLOSES CMRC-981


packages/core/core-flows/src/order/workflows/cancel-order-fulfillment.ts
```diff
@@ -114,44 +125,120 @@ function prepareCancelOrderFulfillmentData({
...
   return {
     order_id: order.id,
     reference: Modules.FULFILLMENT,
     reference_id: fulfillment.id,
-    items: fulfillment.items!.map((i) => {
+    items: lineItemIds.map((lineItemId) => {
+      // find order item
+      const orderItem = order.items!.find((i) => i.id === lineItemId)
+      // find inventory items
+      const iitems = (orderItem as any)!.variant?.inventory_items
```
olivermrbl: nit: I've seen `(orderItem as any).variant` a few times in this PR, would it make sense to update the type to reference it?

packages/core/core-flows/src/order/workflows/cancel-order-fulfillment.ts
```diff
@@ -199,6 +286,9 @@ export const cancelOrderFulfillmentWorkflow = createWorkflow(
           "id",
           "status",
           "items.*",
```
olivermrbl: question: Can we remote `items.*` in favor of specific properties?

integration-tests/http/__tests__/order/admin/order.spec.ts
```diff
@@ -1008,6 +1010,386 @@ medusaIntegrationTestRunner({
...
+          { provider_id: "pp_system_default" },
+          storeHeaders
+        )
+
+        order = (
+          await api.post(`/store/carts/${cart.id}/complete`, {}, storeHeaders)
+        ).data.order
+      })
+
+      it("should correctly manage reservations when canceling a fulfillment (with inventory kit)", async () => {
```
olivermrbl, line 1244: suggestion: This is great, but we could probably cover a few additional cases, e.g. the ones we expect to throw

olivermrbl (commented):
Nice work @fPolic. No major comments, I'll let you decide if we merge now or after addressing the nits/suggestions.

olivermrbl (approved):

packages/core/core-flows/src/order/workflows/create-fulfillment.ts
```diff
@@ -157,31 +174,59 @@ function prepareFulfillmentData({
...
+          (ii) => ii.inventory.id === r.inventory_item_id
+        )
+
+        return {
+          line_item_id: i.id,
+          inventory_item_id: r.inventory_item_id,
+          quantity: MathBN.mult(
+            iItem?.required_quantity ?? 1,
+            i.quantity
+          ).toNumber(), // TODO: type
```
olivermrbl: question: is it correctly understood, that we don't necessarily fulfill all reserved items here? If not, I would have expected us to just use the reservation quantity here.

olivermrbl (approved):
Overall LGTM, nice work. Only a minor question from my side

packages/core/core-flows/src/order/workflows/create-fulfillment.ts
fPolic: yes in case a line item would be partially fulfilled, we need to reduce a part of the reservation item's quantity here

PR #11798, opened 2025-03-10
fix: preserve payment sessions during certain Stripe errors for webhook reconciliation

Saryazdi-Saman:
### Issue
When creating a payment intent with specific parameters (off_session: true, confirm: true, capture_method: "automatic"), the payment module throws errors on certain Stripe responses. This triggers a cleanup function that deletes the payment session.

This creates a significant problem in the payment flow: if the issue resolves and a webhook event is later received from Stripe, the payment session no longer exists to be matched and updated.

Additionally, the payment module currently doesn't update the payment_session status when calling authorizePaymentSession, which creates inconsistency in the payment state.

### Solution
This PR implements several improvements to make payment processing more robust:

1. **Better error handling**: Errors from Stripe are now handled appropriately:
   - If the error is retry-able (network errors, rate limits), a retry is triggered
   - If the issue can't be solved instantly, the error details are stored in payment_session.data
   - This preserves the payment_session for webhook reconciliation or future resolution

2. **Retry mechanism**: Added automatic retry with exponential backoff for errors that might be fixed after retry (connection issues, rate limits)

3. **Status consistency**: Modified authorizePaymentSession method to update payment_session.status regardless of whether the authorization is successful

## Benefits
This change improves payment reliability by:
- Preventing data loss during transient errors
- Maintaining a consistent record for webhook reconciliation 
- Enabling better debugging of payment issues
- Ensuring payment status is always reflected accurately

Note: I haven't included tests yet as I wanted to get feedback on the approach first. I'm happy to add tests once the solution direction is confirmed.

### DIFF
--- a/packages/modules/providers/payment-stripe/src/core/stripe-base.ts
+++ b/packages/modules/providers/payment-stripe/src/core/stripe-base.ts
@@ -42,6 +42,7 @@ import {
 import {
   ErrorCodes,
   ErrorIntentStatus,
+  HandledErrorType,
   PaymentIntentOptions,
   StripeOptions,
 } from "../types"
@@ -115,6 +116,87 @@ abstract class StripeBase extends AbstractPaymentProvider<StripeOptions> {
     return res
   }
 
+  handleStripeError(error: any): HandledErrorType | null {
+    if (error instanceof Stripe.errors.StripeError) {
+      switch (error.type) {
+        case 'StripeCardError':
+          const stripeError = error.raw as Stripe.errors.StripeCardError
+          if (stripeError.payment_intent) {
+            return {
+              retry: false,
+              data: stripeError.payment_intent
+            }
+          } else {
+            throw this.buildError(
+              "An error occurred in InitiatePayment during the creation of the stripe payment intent",
+              error
+            )
+          }
+
+        case 'StripeConnectionError':
+        case 'StripeRateLimitError':
+          return {
+            retry: true,
+            data: null,
+          }
+        case 'StripeAPIError': {
+          return null
+        }
+        default:
+          throw this.buildError(
+            "An error occurred in InitiatePayment during the creation of the stripe payment intent",
+            error
+          )
+      }
+    } else {
+      throw this.buildError(
+        "An error occurred in InitiatePayment during the creation of the stripe payment intent",
+        error
+      )
+    }
+  }
+
+  async executeWithRetry<T>(
+    apiCall: () => Promise<T>,
+    maxRetries: number = 3,
+    baseDelay: number = 1000,
+    currentAttempt: number = 1
+  ): Promise<T | null> {
+    try {
+      return await apiCall();
+    } catch (error) {
+      const handledError = this.handleStripeError(error)
+      if (handledError === null) {
+        return null
+      }
+
+      if (
+        handledError.retry &&
+        currentAttempt <= maxRetries
+      ) {
+        const delay = baseDelay * Math.pow(2, currentAttempt - 1) * (0.5 + Math.random() * 0.5);
+        await new Promise(resolve => setTimeout(resolve, delay));
+        return this.executeWithRetry(apiCall, maxRetries, baseDelay, currentAttempt + 1);
+      }
+
+      if (
+        handledError.retry && 
+        currentAttempt > maxRetries
+      ) {
+        return null
+      }
+
+      if (handledError.data) {
+        return handledError.data as T
+      }
+      throw this.buildError(
+        "An error occurred in InitiatePayment during the creation of the stripe payment intent",
+        error
+      )
+    }
+  }
+
+
   async getPaymentStatus({
     data,
   }: GetPaymentStatusInput): Promise<GetPaymentStatusOutput> {
@@ -173,22 +255,21 @@ abstract class StripeBase extends AbstractPaymentProvider<StripeOptions> {
       | string
       | undefined
 
-    let sessionData
-    try {
-      sessionData = (await this.stripe_.paymentIntents.create(intentRequest, {
-        idempotencyKey: context?.idempotency_key,
-      })) as unknown as Record<string, unknown>
-    } catch (e) {
-      throw this.buildError(
-        "An error occurred in InitiatePayment during the creation of the stripe payment intent",
-        e
+      const sessionData = await this.executeWithRetry<Stripe.PaymentIntent>(
+        () => this.stripe_.paymentIntents.create(intentRequest, {
+          idempotencyKey: context?.idempotency_key,
+        })
       )
-    }
-
-    return {
-      id: sessionData.id,
-      data: sessionData,
-    }
+      if (sessionData === null) {
+        return {
+          id: data?.session_id as string,
+        }
+      }
+  
+      return {
+        id: sessionData.id,
+        data: sessionData as unknown as Record<string, unknown>,
+      }
   }
 
   async authorizePayment(
--- a/packages/modules/providers/payment-stripe/src/core/stripe-base.ts
+++ b/packages/modules/providers/payment-stripe/src/core/stripe-base.ts
@@ -263,14 +263,9 @@ abstract class StripeBase extends AbstractPaymentProvider<StripeOptions> {
         idempotencyKey: context?.idempotency_key,
       })
     )
-    if (sessionData === null) {
-      return {
-        id: data?.session_id as string,
-      }
-    }
 
     return {
-      id: sessionData.id,
+      id: sessionData?.id ?? data?.session_id as string,
       data: sessionData as unknown as Record<string, unknown>,
     }
   }
--- a/packages/modules/providers/payment-stripe/src/core/stripe-base.ts
+++ b/packages/modules/providers/payment-stripe/src/core/stripe-base.ts
@@ -1,5 +1,5 @@
 import Stripe from "stripe"
-
+import { setTimeout } from "timers/promises"
 import {
   AuthorizePaymentInput,
   AuthorizePaymentOutput,
@@ -42,7 +42,6 @@ import {
 import {
   ErrorCodes,
   ErrorIntentStatus,
-  HandledErrorType,
   PaymentIntentOptions,
   StripeOptions,
 } from "../types"
@@ -50,6 +49,10 @@ import {
   getAmountFromSmallestUnit,
   getSmallestUnit,
 } from "../utils/get-smallest-unit"
+
+type HandledErrorType = 
+| { retry: true; } 
+| { retry: false; data: Stripe.PaymentIntent };
 
 abstract class StripeBase extends AbstractPaymentProvider<StripeOptions> {
   protected readonly options_: StripeOptions
@@ -137,7 +140,6 @@ abstract class StripeBase extends AbstractPaymentProvider<StripeOptions> {
         case 'StripeRateLimitError':
           return {
             retry: true,
-            data: null,
           }
         case 'StripeAPIError': {
           return null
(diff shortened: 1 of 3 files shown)


packages/modules/providers/payment-stripe/src/types/index.ts
```diff
@@ -21,6 +23,10 @@ export interface StripeOptions {
   paymentDescription?: string
 }
 
+export type HandledErrorType =
```
sradevski: Since this is just an internal data type, it should live inside the Stripe payment provider (i.e. you can put it at the top of `stripe-base`.

packages/modules/providers/payment-stripe/src/core/stripe-base.ts
```diff
@@ -115,6 +116,87 @@ abstract class StripeBase extends AbstractPaymentProvider<StripeOptions> {
     return res
   }
 
+  handleStripeError(error: any): HandledErrorType | null {
+    if (error instanceof Stripe.errors.StripeError) {
```
sradevski: I think we can remove the `if` check as the `switch` will serve a similar purpose (it already checks for stripe-explicit errors, and the default is to throw a generic error)

packages/modules/providers/payment-stripe/src/core/stripe-base.ts
```diff
@@ -115,6 +116,87 @@ abstract class StripeBase extends AbstractPaymentProvider<StripeOptions> {
...
+      if (handledError === null) {
+        return null
+      }
+
+      if (
+        handledError.retry &&
+        currentAttempt <= maxRetries
+      ) {
+        const delay = baseDelay * Math.pow(2, currentAttempt - 1) * (0.5 + Math.random() * 0.5);
+        await new Promise(resolve => setTimeout(resolve, delay));
```
sradevski: You can do `import { setTimeout } from "timers/promises"`

and then just do `await setTimeout(delay)` here.

packages/modules/providers/payment-stripe/src/core/stripe-base.ts
```diff
@@ -115,6 +116,87 @@ abstract class StripeBase extends AbstractPaymentProvider<StripeOptions> {
...
+              "An error occurred in InitiatePayment during the creation of the stripe payment intent",
+              error
+            )
+          }
+
+        case 'StripeConnectionError':
+        case 'StripeRateLimitError':
+          return {
+            retry: true,
+            data: null,
```
sradevski: We can just return `{ retry: true }` thereby data being `undefined`.

packages/modules/providers/payment-stripe/src/core/stripe-base.ts
```diff
@@ -115,6 +116,87 @@ abstract class StripeBase extends AbstractPaymentProvider<StripeOptions> {
...
+          }
+        case 'StripeAPIError': {
+          return null
```
sradevski: Shouldn't this just throw? We can let it fallback to the default case

packages/modules/providers/payment-stripe/src/core/stripe-base.ts
```diff
@@ -173,22 +255,21 @@ abstract class StripeBase extends AbstractPaymentProvider<StripeOptions> {
...
+          idempotencyKey: context?.idempotency_key,
+        })
       )
-    }
-
-    return {
-      id: sessionData.id,
-      data: sessionData,
-    }
+      if (sessionData === null) {
```
sradevski: You can remove this if check and only return the data below
```
{
        id: sessionData.id,
        data: sessionData as unknown as Record<string, unknown>
}
```

Since if sessionData is not defined we should throw as the payment intent wasn't created successfully

packages/modules/providers/payment-stripe/src/core/stripe-base.ts
```diff
@@ -115,6 +116,87 @@ abstract class StripeBase extends AbstractPaymentProvider<StripeOptions> {
...
+  async executeWithRetry<T>(
+    apiCall: () => Promise<T>,
+    maxRetries: number = 3,
+    baseDelay: number = 1000,
+    currentAttempt: number = 1
+  ): Promise<T | null> {
+    try {
+      return await apiCall();
+    } catch (error) {
+      const handledError = this.handleStripeError(error)
```
sradevski, line 182: With the suggested changes, this should always be defined, with either `{ retry: false, data: somedata}` or `{retry: true}`, so we can remove the if check below

packages/modules/providers/payment-stripe/src/core/stripe-base.ts
```diff
@@ -115,6 +116,87 @@ abstract class StripeBase extends AbstractPaymentProvider<StripeOptions> {
...
+        const delay = baseDelay * Math.pow(2, currentAttempt - 1) * (0.5 + Math.random() * 0.5);
+        await new Promise(resolve => setTimeout(resolve, delay));
+        return this.executeWithRetry(apiCall, maxRetries, baseDelay, currentAttempt + 1);
+      }
+
+      if (
+        handledError.retry && 
+        currentAttempt > maxRetries
+      ) {
+        return null
```
sradevski: We should throw here

packages/modules/providers/payment-stripe/src/core/stripe-base.ts
```diff
@@ -115,6 +116,87 @@ abstract class StripeBase extends AbstractPaymentProvider<StripeOptions> {
...
+      }
+
+      if (handledError.data) {
```
sradevski: This check should go above
```
 if (
        handledError.retry &&
        currentAttempt <= maxRetries
      ) {
```

If the handleError returned some data, then return that and complete. Otherwise we retry

sradevski (commented):
Thanks a lot for the contribution! I left some comments on how we can improve it, and once they're addressed we can get it merged

packages/modules/providers/payment-stripe/src/core/stripe-base.ts
Saryazdi-Saman: I should have left a descriptive comment on this line to explain the reasoning.

According to [Stripe API errors documentation](https://docs.stripe.com/error-handling?lang=node#api-errors), StripeAPIError should be treated as indeterminate - we shouldn't assume it succeeded or failed, and should rely on webhooks for the outcome.

Throwing an error here would trigger the cleanup function that deletes the payment session. By returning null instead, we preserve the payment session, which is critical for when webhooks arrive later with the actual outcome (which could be successful).

This approach follows Stripe's best practice of relying on webhooks for the final state in indeterminate cases. The user can still manually throw an error if payment_session.data is null if they prefer different behavior.

What are your thoughts?

packages/modules/providers/payment-stripe/src/core/stripe-base.ts
Saryazdi-Saman: Great! Done!

packages/modules/providers/payment-stripe/src/core/stripe-base.ts
Saryazdi-Saman: I modified the type definition to simplify it as you suggested. this line is now { retry: true}

packages/modules/providers/payment-stripe/src/core/stripe-base.ts
Saryazdi-Saman: I have removed this if check, and simplified it in a one liner.
Before i remove the null possibilities, I need to know your thoughts on returning null as a valid value for cases where result is uncertain and requires webhook updates.

packages/modules/providers/payment-stripe/src/core/stripe-base.ts
Saryazdi-Saman: Applied.

packages/modules/providers/payment-stripe/src/core/stripe-base.ts
Saryazdi-Saman: Perfect, this check is moved to the top to return immediately if there is data.

packages/modules/providers/payment-stripe/src/core/stripe-base.ts
Saryazdi-Saman: I removed the if check, just switching over types and falling back on default

packages/modules/providers/payment-stripe/src/core/stripe-base.ts
Saryazdi-Saman: The purpose was to avoid throwing if the result is uncertain so the payment_session wont be deleted in case a webhook event arrives that addresses that payment session.
I'll wait for your feedback before applying any changes

packages/modules/providers/payment-stripe/src/core/stripe-base.ts
```diff
@@ -115,6 +119,87 @@ abstract class StripeBase extends AbstractPaymentProvider<StripeOptions> {
...
+        // We should retry the operation and ultimately rely on webhooks
+        // for the final status if retries are exhausted
+        return {
+          retry: true,
+        }
+      case 'StripeAPIError': {
+        // API errors should be treated as indeterminate per Stripe documentation
+        // We return null to preserve the payment session and rely on webhooks
+        // for the actual outcome rather than assuming failure
+        return null
```
sradevski: I would still maybe return
```
{
  retry: false,
  data: {
    state: "indeterminate_error"
  }
}
```

or something like that so at least we have a trace of what happened in the DB

packages/modules/providers/payment-stripe/src/core/stripe-base.ts
```diff
@@ -115,6 +119,87 @@ abstract class StripeBase extends AbstractPaymentProvider<StripeOptions> {
...
+    apiCall: () => Promise<T>,
+    maxRetries: number = 3,
+    baseDelay: number = 1000,
+    currentAttempt: number = 1
+  ): Promise<T | null> {
+    try {
+      return await apiCall();
+    } catch (error) {
+      const handledError = this.handleStripeError(error)
+      if (handledError === null) {
```
sradevski: If we do the suggestion above, we can remove this check.

sradevski (commented):
Final comment, otherwise LGTM

packages/modules/providers/payment-stripe/src/core/stripe-base.ts
sradevski: Got it, thanks for explaining, I think it makes sense. I just added a small suggestion on how we can improve it.

packages/modules/providers/payment-stripe/src/core/stripe-base.ts
sradevski: I think at this point we can be certain the payment intent wasn't created, or it is highly unlikely, so it's safer to throw.

packages/modules/providers/payment-stripe/src/core/stripe-base.ts
```diff
@@ -115,6 +119,87 @@ abstract class StripeBase extends AbstractPaymentProvider<StripeOptions> {
...
+        // Logic for retrying
+        const delay = baseDelay * Math.pow(2, currentAttempt - 1) * (0.5 + Math.random() * 0.5);
+        await setTimeout(delay);
+        return this.executeWithRetry(apiCall, maxRetries, baseDelay, currentAttempt + 1);
+      }
+
+      // Retries are exhausted or result is null
+      // handleStripeError() didn't throw an error
+      // return null to avoid triggering the cleanup function 
+      return null
```
sradevski: I think if we get up to this point, it means the payment intent wasn't created (i.e. the error wasn't StripeAPIError, nor was it something where we know the payment intent was created). I think we should throw in this case.

packages/modules/providers/payment-stripe/src/core/stripe-base.ts
Saryazdi-Saman: You're right, returning a state is better than returning null. I've implemented this change to preserve a trace of what happened

packages/modules/providers/payment-stripe/src/core/stripe-base.ts
Saryazdi-Saman: I agree. I think I was overcorrecting. At this point it is safe to assume payment intent wasn't created.
Changes are applied.
Let me know if there are any other concerns to address.

sradevski (approved):

PR #11800, opened 2025-03-11
fix(workflow-engine-*): scheduled jobs interval

adrien2p:
**What**
Currently only cron pattern are supported by scheduled jobs, this can lead to issue. for example you set the pattern to execute every hours at minute 0 and second 0 (as it is expected to execute at exactly this constraint) but due to the moment it gets executed we our out of the second 0 then the job wont get executed until the next scheduled cron table execution.

With this pr we introduce the `interval` configuration which allows you the specify a delay between execution in ms (e.g every minute -> 60 * 1000 ms) which ensure that once a job is executed another one is scheduled for a minute later.

**Usage**
```ts
// jobs/job-1.ts
const thirtySeconds = 30 * 1000

export const config = {
  name: "job-1",
  schedule: {
    interval: thirtySeconds
  },
}
```


packages/modules/workflow-engine-redis/src/utils/workflow-orchestrator-storage.ts
```diff
@@ -438,19 +466,12 @@ export class RedisDistributedTransactionStorage
...
+        repeat: repeatOptions,
         removeOnComplete: {
           age: 86400,
           count: 1000,
         },
-        removeOnFail: {
-          age: 604800,
-          count: 5000,
-        },
+        removeOnFail: false,
```
sradevski: I think you typically don't want to store failures indefinitely, you need them long-enough to debug them. Instead of removing it completely, we can increase the age if you want, but I think we should still keep it.

packages/modules/workflow-engine-inmemory/src/utils/workflow-orchestrator-storage.ts
```diff
@@ -428,7 +443,13 @@ export class InMemoryDistributedTransactionStorage
       return
     }
 
-    const nextExecution = job.expression.next().getTime() - Date.now()
+    let nextExecution = 0
```
sradevski: Let's extract this into a helper function, as it can be used above as well

packages/modules/workflow-engine-inmemory/src/utils/workflow-orchestrator-storage.ts
```diff
@@ -383,16 +383,31 @@ export class InMemoryDistributedTransactionStorage
...
 
     const timer = setTimeout(async () => {
       this.jobHandler(jobId)
     }, nextExecution)
 
     this.scheduled.set(jobId, {
       timer,
-      expression,
+      expression:
+        expression ??
```
sradevski: Maybe we can move the `interval:..` part to the expression definition above? So above we'd have

```
let expression
if ("cron" in schedulerOptions) {
 expression = ...
} else if ("interval" in schedulerOptions {
expression = ...
} else throw

const nextExecution = getNextExecution(expression)
```

sradevski (commented):
Just some small comments. LGTM otherwise

packages/modules/workflow-engine-inmemory/src/utils/workflow-orchestrator-storage.ts
adrien2p: I ve done all the changes [here](https://github.com/medusajs/medusa/pull/11800/commits/8ed96fea0915f69eae3a386923cd625ac04fbfb8) let me know wyt 👍

sradevski (approved):
Nice!

PR #11802, opened 2025-03-11
refactor: use module name as the snapshot name

thetutlage:
Fixes: FRMW-2930

This PR updates the MikroORM config to use the module name as the snapshot name when generating migration files. Otherwise, MikroORM defaults to the database name and every time you update the database name, it will create a new snapshot. 

Also, we migrate existing snapshot files to be same the new file name to avoid breaking changes.


adrien2p (approved):
LGTM! as per our discussion we can add a snapshot renaming strategy upon calling the generate command to ensure to always be in a good state

PR #11805, opened 2025-03-11
feat(medusa,core-flows,types,js-sdk): Draft Order workflows and API endpoints

kasperkristensen:


packages/medusa/src/api/admin/draft-orders/[id]/edit/promotions/route.ts
```diff
@@ -0,0 +1,48 @@
...
+  })
+}
+
+export const DELETE = async (
+  req: AuthenticatedMedusaRequest<AdminRemoveDraftOrderPromotionsType>,
+  res: MedusaResponse<HttpTypes.AdminDraftOrderPreviewResponse>
+) => {
+  const { id } = req.params
+
+  console.log("req.validatedBody", JSON.stringify(req.validatedBody, null, 2))
```
Comment (olivermrbl) [inline packages/medusa/src/api/admin/draft-orders/[id]/edit/promotions/route.ts]:
```suggestion
```

packages/core/core-flows/src/draft-order/workflows/update-draft-order.ts
```diff
@@ -0,0 +1,258 @@
...
+    return new StepResponse(updatedOrder, order)
+  },
+  async function (prevData, { container }) {
+    if (!prevData) {
+      return
+    }
+
+    const service = container.resolve<IOrderModuleService>(Modules.ORDER)
+
+    await service.updateOrders([prevData as UpdateOrderDTO])
```
olivermrbl, line 86: todo: This doesn't seem right, `prevData` is the new data, so the compensation action would trigger an update with the new data.

packages/core/core-flows/src/draft-order/workflows/update-draft-order-shipping-method.ts
```diff
@@ -0,0 +1,154 @@
...
+
+      if (!promotionLink) {
+        return []
+      }
+
+      if (Array.isArray(promotionLink)) {
+        return promotionLink.map((promo) => promo.promotion.code)
+      }
+
+      return [promotionLink.promotion.code]
```
olivermrbl, line 111: question: Why do we need to query the link here? Nothing wrong with it, just seems like an anti pattern

packages/core/core-flows/src/draft-order/steps/create-draft-order-shipping-method-adjustments.ts
```diff
@@ -0,0 +1,54 @@
...
+    if (!shippingMethodAdjustmentsToCreate?.length) {
+      return new StepResponse(void 0, [])
+    }
+
+    const service = container.resolve<IOrderModuleService>(Modules.ORDER)
+
+    const shippingMethodAdjustments =
+      await service.createOrderShippingMethodAdjustments(
+        shippingMethodAdjustmentsToCreate
+      )
```
olivermrbl, line 32: question: Should we also filter out `undefined` amounts here, like we do for line items/

olivermrbl (approved):

olivermrbl:
FYI, I will merge this now, so we can proceed with the draft order plugin. This doesn't touch any other domains, so it should be a safe action.

PR #11808, opened 2025-03-11
Polish translations improvements

radeknapora:


sradevski (approved):

radeknapora:
Polish has more plural forms than English, requiring explicit pluralization support in the translation system. The standard English approach (one and other) was not sufficient because Polish includes additional forms (few, many).

Without this approach, many UI elements defaulted to English due to missing keys for proper Polish grammar. To correctly handle pluralization, I added the necessary keys for dynamic values, such as the number of keys (numberOfKeys).

Now, translations are correctly inflected based on quantity (1 klucz, 2 klucze, 5 kluczy, etc.), improving readability and ensuring proper grammatical correctness in Polish.

PR #11850, opened 2025-03-14
feat: Add support for generating a production-ready config out of the…

sradevski:
Our default config generation includes in-memory modules, which are not production ready.

This PR adds a production-ready config that is used as the default when running in a cloud environment

### DIFF
--- a/packages/core/utils/src/common/__tests__/define-config.spec.ts
+++ b/packages/core/utils/src/common/__tests__/define-config.spec.ts
@@ -958,4 +958,188 @@ describe("defineConfig", function () {
       }
     `)
   })
+
+  it("should include cloud-based modules when in cloud execution context", function () {
+    const originalEnv = { ...process.env }
+
+    process.env.EXECUTION_CONTEXT = "medusa-cloud"
+    process.env.REDIS_URL = "redis://localhost:6379"
+    process.env.S3_FILE_URL = "https://s3.amazonaws.com/medusa-cloud-test"
+    process.env.S3_PREFIX = "test"
+    process.env.S3_REGION = "us-east-1"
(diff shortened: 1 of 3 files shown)


olivermrbl (approved):
Nice, this looks good to me. Will just let @adrien2p or @carlos-r-l-rodrigues have a look as well

packages/core/utils/src/common/__tests__/define-config.spec.ts
```diff
@@ -958,4 +958,188 @@ describe("defineConfig", function () {
...
+    process.env.S3_FILE_URL = "https://s3.amazonaws.com/medusa-cloud-test"
+    process.env.S3_PREFIX = "test"
+    process.env.S3_REGION = "us-east-1"
+    process.env.S3_BUCKET = "medusa-cloud-test"
+    process.env.S3_ENDPOINT = "https://s3.amazonaws.com"
+    const res = defineConfig({})
+
+    process.env = { ...originalEnv }
+
+    expect(res).toMatchInlineSnapshot(`
```
carlos-r-l-rodrigues, line 976: suggestion: maybe we should use a regular equal here. Snapshots are super strict and any property order that changes, will break it.

carlos-r-l-rodrigues (approved):

packages/core/utils/src/common/__tests__/define-config.spec.ts
sradevski: I just followed the pattern of the previous tests. I'm more than happy to change it everywhere if you think it's better.

carlos-r-l-rodrigues: It is alright to keep it as is.
This is more personal preference than anything else.

PR #11876, opened 2025-03-17
feat(medusa,types): add enabled plugins route

riqwan:
what:

- adds a route to list all plugins within a project


integration-tests/http/package.json
```diff
@@ -34,6 +34,7 @@
     "@medusajs/workflow-engine-inmemory": "workspace:*",
     "form-data": "^4.0.0",
     "jsonwebtoken": "^9.0.2",
+    "medusa-plugin-navigation": "^0.0.4",
```
olivermrbl: question: What plugin is this?

riqwan: just a random public v2 plugin i found in the web for testing.

riqwan: I've removed it.

olivermrbl (approved):

PR #11885, opened 2025-03-18
feat(dashboard,types): add credit lines + loyalty changes

riqwan:
what:

- adds conditional refund form
- adds credit lines tab in order summary

Screenshot 2025-03-18 at 09 10 12

Screenshot 2025-03-18 at 09 10 42

Screenshot 2025-03-18 at 09 11 08


packages/admin/dashboard/src/routes/orders/order-balance-settlement/components/order-balance-settlement-form/order-balance-settlement-form.tsx
```diff
@@ -0,0 +1,420 @@
...
+
+                              if (fieldValue && !isNaN(fieldValue)) {
+                                if (
+                                  fieldValue < 0 ||
+                                  fieldValue > pendingDifference
+                                ) {
+                                  form.setError(`refund.amount`, {
+                                    type: "manual",
+                                    message: t(
+                                      "orders.payment.createRefundWrongQuantity",
```
fPolic: todo: validation message seems off

Screenshot 2025-03-18 at 17 14 40

fPolic (commented):
Overall, Admin UX and code looks good.

Not sure if I missed something in the setup or this is still todo but when I submit a credit line refund request I get a `Store credit account not found` error. However, credit/refund lines are still created:
Screenshot 2025-03-18 at 17 38 04

packages/admin/dashboard/src/routes/orders/order-balance-settlement/components/order-balance-settlement-form/order-balance-settlement-form.tsx
riqwan: good catch on both, will fix these!

packages/admin/dashboard/src/routes/orders/order-balance-settlement/components/order-balance-settlement-form/order-balance-settlement-form.tsx
```diff
@@ -0,0 +1,398 @@
...
+  const { mutateAsync: createRefund, isPending: isRefundPending } =
+    useRefundPayment(order.id, activePayment?.id!)
+
+  const settlementType = form.watch("settlement_type")
+
+  const handleSubmit = form.handleSubmit(async (data) => {
+    if (data.settlement_type === "credit_line") {
+      await createCreditLine(
+        {
+          amount: parseFloat(data.credit_line!.amount! as string) * -1,
```
olivermrbl, line 86: thought: I might have missed a critical detail in a previous PR–I would have expected credit lines to always have a positive value. Negative values would have a significantly different meaning.

For example:

1. Customer places order
```
Sale: $100 (debit)
Credit line: $0
Net total: $100
```
2. Customer requests return (and refund)
```
Sale: $100 (debit)
Credit line: $100
Net total: $0 (debit - credit)
```

Let me know if I am missing something? I haven't spotted this earlier, so it's likely that I am missing some context for why it's created as a negative value here.

riqwan: We add negative credit lines to refund to store credit. Otherwise we need a different way to indicate that a credit line "takes from the balance" or "adds to the balance".

When its a refund, we take from the balance through the negative credit line. When its a "payment", we add to the balance through a positive credit line. This is exactly how we do for payments/refunds as well. 

Here is an example of an order refunded partially through payment method and the other through store credit

Screenshot 2025-03-31 at 23 21 54

olivermrbl: > When its a "payment", we add to the balance through a positive credit line

Is it correctly understood that this used when paying with store credit? Or what does "payment" refer to here?

riqwan: Yes exactly! When we apply store credit to cart, we create a positive credit line.

olivermrbl (approved):
With the new increased scope that includes anonymous gift cards, some of this UI might need to change. However, I will merge this for now, since 1) it is only exposed to users with the plugin installed, and 2) those users have already been using this UI for a while. With that in mind, merging this as is should not cause any issues.

This is to avoid having this PR hang any longer.

PR #11886, opened 2025-03-18
fix: hooks auto-generated types to have precise input options

thetutlage:
Fixes: FRMW-2933

**Types before**

CleanShot 2025-03-18 at 14 53 40@2x

**Types after**
CleanShot 2025-03-18 at 14 57 08@2x


carlos-r-l-rodrigues (approved):
🔥

PR #11891, opened 2025-03-18
feat: Set pricing context hook

thetutlage:


carlos-r-l-rodrigues (dismissed):

packages/core/core-flows/src/cart/workflows/update-line-item-in-cart.ts
```diff
@@ -73,6 +78,11 @@ export const updateLineItemInCartWorkflow = createWorkflow(
       return [item.variant_id].filter(Boolean)
     })
 
+    const setPricingContext = createHook("setPricingContext", {
+      cart,
```
olivermrbl: question: Should we also hard-code the currency code and region ID here?

olivermrbl (commented):
We also need to add these to workflows calculating shipping option prices.

Separately, I believe we missed a workflow `refresh-cart-items.ts`.

Finally, should we add a few tests to ensure it behaves as expected?

thetutlage:
> Finally, should we add a few tests to ensure it behaves as expected?

Testing with hooks is a little tricky, since the hooks are applied globally on all the executions of a workflow. So, if I define a hook that mutates the `pricingContext`, then that mutation will leak into all the tests as well.

@carlos-r-l-rodrigues Do you have any tips on how we can approach testing in this case?

integration-tests/modules/__tests__/cart/store/cart.workflows.spec.ts
```diff
@@ -627,6 +638,337 @@ medusaIntegrationTestRunner({
...
+            })
+
+            setPricingContextHook = undefined
+            pricingModule.listPriceSets = originalFn
+
+            expect(pricingContext).toEqual(
+              expect.objectContaining({
+                unit_price: 200,
+                region_id: region.id,
+                currency_code: "usd",
```
carlos-r-l-rodrigues, line 959: should we also add `customer_id` on the check?

packages/core/core-flows/src/cart/workflows/add-to-cart.ts
```diff
@@ -93,6 +94,24 @@ export const addToCartWorkflow = createWorkflow(
...
+      items: input.items,
+      additional_data: input.additional_data,
+    })
+
+    const setPricingContextResult = setPricingContext.getResult() as any
+    const pricingContext = transform(
+      { cart, setPricingContextResult },
+      (data) => {
+        return {
+          ...data.cart,
```
carlos-r-l-rodrigues, line 116: I know this was like that before, but maybe we shouldn't pass the entire cart to the pricing module.
We can just use the same fields we are passing on other workflows. 
Apparently there is a util `filterObjectByKeys(data.cart, cartFieldsForPricingContext)` used on other places.

packages/core/core-flows/src/cart/workflows/list-shipping-options-for-cart.ts
```diff
@@ -122,7 +137,12 @@ export const listShippingOptionsForCartWorkflow = createWorkflow(
             },
           },
 
-          calculated_price: { context: cart },
+          calculated_price: {
+            context: {
+              ...cart,
```
carlos-r-l-rodrigues, line 149: same comment as above

carlos-r-l-rodrigues (approved):

packages/core/core-flows/src/cart/workflows/add-to-cart.ts
```diff
@@ -93,6 +94,24 @@ export const addToCartWorkflow = createWorkflow(
...
     })
 
+    const setPricingContext = createHook("setPricingContext", {
+      cart,
+      variantIds,
+      items: input.items,
+      additional_data: input.additional_data,
+    })
+
+    const setPricingContextResult = setPricingContext.getResult() as any
```
olivermrbl: question: What would it take to get rid of `as any`?

olivermrbl: This is not super important right now, but I was just curious.

thetutlage: We will have to settle on the properties the hook must/can return and then define a zod schema to validate those properties.

packages/core/core-flows/src/cart/workflows/add-to-cart.ts
thetutlage: So essentially all these workflows should send the same set of properties? Which are

- currency_code
- region_id
- customer_id

Do you think there could be an unintentional breaking change after this?

carlos-r-l-rodrigues: There is a bit more that is passed apparently, including shipping address. basically everything that is declared on that variable `cartFieldsForPricingContext`.

Yes, this is unintentional breaking change, but I don't expect that to be widely used. 
@oli can decide here what is the best to be done. I believe the entire cart can be a huge parameter to be passing all the time, but if we want to avoid even this minimal breaking change, ideally all the other places also receive the full cart.

packages/core/core-flows/src/cart/workflows/add-to-cart.ts
olivermrbl: Ah I see, should we set it to a record then? So we can get rid of the `any`

packages/core/core-flows/src/cart/workflows/add-to-cart.ts
olivermrbl: @thetutlage @carlos-r-l-rodrigues, should we perhaps leave a TODO that we can revisit when we can be more strict about what we pass in the pricing context? I'd like to avoid having to decide on this right now AND avoid the breaking change.

packages/core/core-flows/src/cart/workflows/add-to-cart.ts
thetutlage: Sure

packages/core/core-flows/src/cart/workflows/add-to-cart.ts
thetutlage: Yeah, I tried cherry picking the fields and tests started to fail. Ofcourse, I would have missed some, but I can imagine it being an exercise to find which precise data (per workflow) has to be passed down

olivermrbl: Yeah my thinking too

integration-tests/modules/__tests__/cart/store/cart.workflows.spec.ts
thetutlage: Just realized, that check is here. https://github.com/medusajs/medusa/pull/11891/files#diff-3084247c8dbb6a88ec078934d653e7c0e243bd1f59a2d7f982e94c451f2bcffaR947

This is because I do not have the customer_id handy. Its pulled under the hood from the email

packages/core/core-flows/src/cart/workflows/update-line-item-in-cart.ts
```diff
@@ -73,6 +79,30 @@ export const updateLineItemInCartWorkflow = createWorkflow(
...
+      }
+    )
+
+    const setPricingContextResult = setPricingContext.getResult()
+    const pricingContext = transform(
+      { cart, setPricingContextResult },
+      (data) => {
+        return {
+          ...data.cart,
+          ...(data.setPricingContextResult ? data.setPricingContextResult : {}),
```
olivermrbl, line 101: suggestion: I wonder if we should enforce currency, region, and customer here? Now that we are opening up for extensibility, users could mistakenly overwrite those. We probably don't want that to happen.

**Same thought for all cart workflows.**

olivermrbl (commented):
Nice work @thetutlage!

Left a thought, but otherwise LGTM.

packages/core/core-flows/src/cart/workflows/update-line-item-in-cart.ts
thetutlage: Done

packages/core/core-flows/src/cart/workflows/update-line-item-in-cart.ts
thetutlage: Yup, fixed all those with tests. https://github.com/medusajs/medusa/pull/11891/commits/279d2fa5e86bc37aea796f8118b5f4a8639132ad

olivermrbl (approved):

PR #11898, opened 2025-03-18
feat(index): add filterable fields to link definition

carlos-r-l-rodrigues:
What:
 - `defineLink` accepts `filterable` fields. That will be used to add the entities and their fields to the Index module.
 
```ts
defineLink(
  ProductModule.linkable.product,
  {
      linkable: Merchant.linkable.merchant,
      filterable: ["id", "active", "name"]
  }
)
```

- Fix circular relationship references in schemas (e.g Product -> Variant -> Product)
- Fix meta information on hierarchical structure config building
- Add new test on index engine config building
- Build query now use the id instead of data->>'id'
- Build query now rely on `ANY(ARRAY[...]::JSONB[])` instead of `IN` in order to improve performances of IN operator by a magnitude of 15-20x for queries like (SKU in ['red-1', 'red-2'])
- Event Bus is injected as dependency on all custom modules


docloulou:
> What:
> 
> * `defineLink` accepts `filterable` fields. That will be used to add the entities and their fields to the Index module.
> 
> ```ts
> defineLink(
>   ProductModule.linkable.product,
>   {
>       linkable: Merchant.linkable.merchant,
>       filterable: ["id", "active", "name"]
>   }
> )
> ```
hello,
this will allow to query.graph filter relational nested properties ?
Thanks

carlos-r-l-rodrigues:
> hello,
> this will allow to query.graph filter relational nested properties ?

Hi @docloulou,
yes, but using `query.index` using the Index module.

docloulou:
> ```ts
> merchant
> ```

it is applicable for core modules or only custom modules ?


is it correct ? : 
```
  const {
    data: products,
  } = await query.index({
    entity: "product",
    fields: req.queryConfig.fields,
    filters: { 
      merchant: {
        active: true,
      },
    },
  });
```

Thanks,

carlos-r-l-rodrigues:
> it is applicable for core modules or only custom modules ?

For both, you set the filterable fields when defining the link and these will be indexed.

> 
> is it correct ? :
> 
> ```
>   const {
>     data: products,
>   } = await query.index({
>     entity: "product",
>     fields: req.queryConfig.fields,
>     filters: { 
>       merchant: {
>         active: true,
>       },
>     },
>   });
> ```

yes, when this PR is done, this should work.
it will be in test for a while before officially releasing it through.

adrien2p:
@olivermrbl do you have some bandwith for this one please?

integration-tests/modules/__tests__/index/search.spec.ts
```diff
@@ -95,6 +91,14 @@ medusaIntegrationTestRunner({
...
         })
 
+        /*
+        const manager = (indexEngine as any).container_.baseRepository.manager_
+        const datas = await manager.execute(`
+          SELECT DISTINCT name FROM index_data
+        `)
+        console.log(JSON.stringify(datas, null, 2))
+        */
+
```
thetutlage: Do we want to keep this in?

carlos-r-l-rodrigues: I removed it

olivermrbl:
This works well–nice work!

Can we do something about the generated types? Right now, it seems the filter defaults to `undefined`. 

For example, say I link product with brand with the following link:
```ts
export default defineLink(
  {
    linkable: ProductModule.linkable.product,
    isList: true,
  },
  {
    linkable: BrandModule.linkable.brand,
    filterable: ["id", "name"],
  }
);
```

The type for the `brand` filter is:
image

carlos-r-l-rodrigues:
> Can we do something about the generated types?

Yes, they were missing. Good catch.
The latest snapshot should have this fixed.

olivermrbl:
Hmm, doesn't seem to be working on my side. Can you reproduce the types locally?

packages/core/types/src/index-data/query-config/query-input-config.ts
```diff
@@ -1,7 +1,56 @@
-import { RemoteQueryInput } from "../../modules-sdk/remote-query-object-from-string"
 import { IndexServiceEntryPoints } from "../index-service-entry-points"
 import { ObjectToIndexFields } from "./query-input-config-fields"
 import { IndexFilters } from "./query-input-config-filters"
+import { IndexOrderBy } from "./query-input-config-order-by"
+
+export type IndexQueryInput<TEntry extends string> = {
+  // service: string This property is still supported under the hood but part of the type due to types missmatch towards fields
```
olivermrbl:
```suggestion
```
nit: guess we might as well remove this?

packages/core/types/src/index-data/query-config/query-input-config.ts
```diff
@@ -1,7 +1,56 @@
...
+    order?: IndexOrderBy<TEntry>
+  }
+  /**
+   * Filters to apply on the retrieved items.
+   */
+  filters?: IndexFilters<TEntry>
+  /**
+   * Apply a query context on the retrieved data. For example, to retrieve product prices for a certain context.
+   */
+  context?: any
```
olivermrbl: thought(non-blocking): could this be typed as `QueryContext`?

packages/modules/index/src/utils/query-builder.ts
```diff
@@ -281,14 +342,25 @@ export class QueryBuilder {
             return
           }
 
-          const castType = this.getPostgresCastType(attr, field).cast
+          // const castType = this.getPostgresCastType(attr, field).cast
```
olivermrbl: rm?
```suggestion
```

packages/medusa/src/api/store/products/route.ts
```diff
@@ -80,7 +80,7 @@ async function getProductsWithIndexEngine(
   await wrapProductsWithTaxPrices(req, products)
   res.json({
     products,
-    count: metadata!.count,
+    count: metadata!.estimate_count,
```
olivermrbl, line 83: thought: Would it make sense to return the estimated count more explicitly here when the index engine is used?
```
{ 
  products: [ ... ], 
  estimate_count: 1000,
  count: 1000, // ensures compatibility with existing implementations
  offset: 0, 
  limit: 50 
}
```

This would make the behavior more obvious to the end-user. I fear using `count` for the estimate could be confusing in the presence of incorrect results.

For example, the case we had with the estimated count being 56 and the actual number of rows 1.

olivermrbl (approved):

packages/core/types/src/index-data/query-config/query-input-config.ts
carlos-r-l-rodrigues: I will create a new type for this, as QueryContext is on Utils.

packages/medusa/src/api/store/products/route.ts
carlos-r-l-rodrigues: Yeah, that's better.
I added this property as optional to the paginated response.

PR #11914, opened 2025-03-20
feat: intra-module relation

Vineet1877:


thetutlage:
Hello @Vineet1877 

Can you please share the purpose of this PR and was there any discussion in which we agree to setup this relationship in the codebase?

PR #11919, opened 2025-03-20
fix(order): caculate received total for pending difference

fPolic:
**What**
- use the return received total to compute pending difference with return requested total

**Why**
- this would show incorrect outstanding amount when receiving a return and also, the outstanding amount after the return is received would be incorrect


packages/modules/order/src/utils/calculate-order-change.ts
```diff
@@ -246,16 +246,24 @@ export class OrderChangeProcessing {
...
-      if (MathBN.gt(item_.return_requested_total, 0)) {
-        orderSummary.pending_difference = MathBN.sub(
-          orderSummary.pending_difference,
-          item_.return_requested_total
-        )
-      }
+      ;[
+        "return_requested_total",
+        "return_received_total",
+        "return_dismissed_total",
```
fPolic: comment: should dismissed total be calcualted into pending difference here?

olivermrbl: How is the dismissed total currently calculated for returns?

fPolic: It is the total of the items that are received as damaged.

olivermrbl: For returns and exchanges, the customer bears the responsibility of returning the items in a good shape. If that is not the case, and the merchant receives damaged items, the customer should not be refunded. This does not directly answer your question, but hopefully it provides the context needed to ensure correct behavior.

fPolic: The issue is that we use Returns for the Claim flow as well. Could we use ReturnReasons with an additional flag here. So we would:
- add a flag on a return reason indicating if it is refundable or not
- introduce additional quantity filed on line items, e.g. `return_dismisssed_refundable_quantity`

fPolic: Removed the dismissed total from the calculation so we can merge the fix for the received total. Opened CMRC-1000 to revisit this.

olivermrbl (approved):

Issue #11470, opened 2025-02-14
[Bug]: Input Component Text Always Highlighted, Blocking Price Entry (ServiceZone -> Conditional Price Component)


lst97:
### Package.json file

```JSON
{
	"name": "medusa",
	"version": "0.0.1",
	"description": "A starter for Medusa projects.",
	"author": "Medusa (https://medusajs.com)",
	"license": "MIT",
	"keywords": [
		"sqlite",
		"postgres",
		"typescript",
		"ecommerce",
		"headless",
		"medusa"
	],
	"scripts": {
		"build": "medusa build",
		"seed": "medusa exec ./src/scripts/seed.ts",
		"start": "medusa start",
		"dev": "medusa develop",
		"test:integration:http": "TEST_TYPE=integration:http NODE_OPTIONS=--experimental-vm-modules jest --silent=false --runInBand --forceExit",
		"test:integration:modules": "TEST_TYPE=integration:modules NODE_OPTIONS=--experimental-vm-modules jest --silent --runInBand --forceExit",
		"test:unit": "TEST_TYPE=unit NODE_OPTIONS=--experimental-vm-modules jest --silent --runInBand --forceExit"
	},
	"dependencies": {
		"@medusajs/admin-sdk": "latest",
		"@medusajs/cli": "latest",
		"@medusajs/framework": "latest",
		"@medusajs/medusa": "latest",
		"@mikro-orm/core": "latest",
		"@mikro-orm/knex": "latest",
		"@mikro-orm/migrations": "6.4.3",
		"@mikro-orm/postgresql": "6.4.3",
		"awilix": "^8.0.1",
		"pg": "^8.13.3"
	},
	"devDependencies": {
		"@medusajs/test-utils": "latest",
		"@mikro-orm/cli": "6.4.3",
		"@swc/core": "1.10.16",
		"@swc/jest": "^0.2.37",
		"@types/jest": "^29.5.14",
		"@types/node": "^22.13.4",
		"@types/react": "^18.3.18",
		"@types/react-dom": "^18.3.5",
		"jest": "^29.7.0",
		"prop-types": "^15.8.1",
		"react": "^18.3.1",
		"react-dom": "^18.3.1",
		"ts-node": "^10.9.2",
		"typescript": "^5.7.3",
		"vite": "^5.4.14",
		"yalc": "^1.0.0-pre.53"
	},
	"engines": {
		"node": ">=20"
	},
	"pnpm": {
		"onlyBuiltDependencies": [
[...]

PR #11944, opened 2025-03-22
fix(dashboard): disable broken autofocus in SO cond. price form

leobenzol:
What

- Disabled auto-focus on the currency input component in the form to edit conditional prices for shipping options

Why

- Currently there is an autofocus loop, on Firefox only, between the currency input and modal close button elements. This prevents firefox users from typing in a price until the loop is broken(e.g. by clicking on the browser navbar or devtools), see #11470. 
- The auto-focus on the currency input doesn't work regardless on other browsers, the close button for the modal is focused instead when opening the stacked modal. See screenshot. Since it doesn't work on other browsers i think it's best to disable it entirely

Edge:
edge


leobenzol:
not stale

leobenzol:
bot hasn't removed the stale label yet, weird

willbouch (commented):
Can I ask you to quickly add a changeset in this PR? This would be a patch. I was able to reproduce on other browsers and it indeed does not work. Is it necessary passing autofocus to false? What happens if we just remove the prop? I see that autofocus is passed in a lot of places to that component. Why does only that one fail?

leobenzol:
@willbouch thanks for the reply, added the changeset :) 

It could be left out, at the time I set it to false explicitly so it doesn't get set again in the future by mistake until it's properly fixed, I could add a comment too if you'd like.

I tried to debug it and find a solution but I gave up after a while. Afaik it was something about stacked modals(or the close button) stealing focus from eachother in firefox, which prevents typing in the field, but in other browsers the autofocus loop stops after a few cycles iirc, but the field is still not focused

It'd be best to disable it for now, I might give another stab at it in the future

willbouch (approved):
Sounds good, let's do it like this

PR #11958, opened 2025-03-24
fix(core-flows,dashboard): handling authorized payment collection on OE

fPolic:
**What**
- fix the creation of payment collection upon order edit when an authorized but not captured payment collection is associated
- remove payment collections refresh from order edit request and only refresh collections when order edit is confirmed (since order edit refresh can include cancelation of authorized payments we want to do that only on confirmed edit)

**How**
- when there is an existing payment collection that has authorized amount and difference due changed, cancel the authorized payments and create a new payment collection for the difference due
- if there is a not_paid collection, update the amount on that collection

---

CLOSES TRI-823
CLOSES CMRC-932
CLOSES CMRC-937
CLOSES CMRC-971
CLOSES CMRC-940


packages/core/core-flows/src/payment-collection/workflows/cancel-payment-collection.ts
```diff
@@ -0,0 +1,105 @@
...
+    )
+
+    cancelPaymentStep({
+      ids: authorizedPaymentIds,
+    })
+
+    const updatedPaymentCollections = updatePaymentCollectionStep({
+      selector: { id: paymentCollection.id },
+      update: {
+        status: PaymentCollectionStatus.CANCELED, // TODO: should it be PARTIALLY_CAPTURED if there is caputred amount?
```
olivermrbl: Yes, I think that would make sense

olivermrbl (commented):
Should we add some tests?

packages/core/core-flows/src/order/workflows/create-or-update-order-payment-collection.ts
```diff
@@ -123,10 +144,30 @@ export const createOrUpdateOrderPaymentCollectionWorkflow = createWorkflow(
       }) as PaymentCollectionDTO[]
     })
 
+    when(
```
riqwan: sugg: Would move this to the last call before wrapping the workflow as there are no reverts for this.

packages/core/core-flows/src/payment-collection/workflows/cancel-payment-collection.ts
```diff
@@ -0,0 +1,105 @@
...
+      { paymentCollection },
+      ({ paymentCollection }) =>
+        paymentCollection.payments
+          ?.filter((p) => !p.captured_at)
+          .map((p) => p.id) ?? []
+    )
+
+    cancelPaymentStep({
+      ids: authorizedPaymentIds,
+    })
```
riqwan: same here, move it after payment collection have been updated 🤞🏻

packages/core/core-flows/src/order/workflows/order-edit/request-order-edit.ts
```diff
@@ -153,12 +152,6 @@ export const requestOrderEditRequestWorkflow = createWorkflow(
     const updateOrderChangesData = getOrderChangesData({ input, orderChange })
     updateOrderChangesStep(updateOrderChangesData)
 
-    createOrUpdateOrderPaymentCollectionWorkflow.runAsStep({
```
riqwan, line 156: wait, when does this happen now ? :D

fPolic: We run this when OE request is confirmed because, now, this workflow might include payment cancellation and we don't want to do that if OE ends up being declined or canceled.

packages/core/core-flows/src/payment-collection/steps/cancel-payment.ts
```diff
@@ -0,0 +1,56 @@
...
+   * The IDs of the payments to cancel.
+   */
+  ids: string[]
+}
+
+export const cancelPaymentStepId = "cancel-payment"
+/**
+ * This step cancels one or more authorized payments.
+ */
+export const cancelPaymentStep = createStep(
```
olivermrbl, line 23: thought: We typically don't send multiple third-party requests in steps, as that could lead to issues with data inconsistencies between Medusa and the third-party system.

For example, if the first payment cancellation succeeds but the second one fails, the workflow would trigger a rollback aborting the entire cancel operation. However, the first payment is still left canceled in the third-party provider.

Should we rethink this flow?

cc @carlos-r-l-rodrigues

fPolic: Seems that we been doing this already for [cancel order](https://github.com/medusajs/medusa/blob/develop/packages/core/core-flows/src/payment/steps/cancel-payment.ts#L23) so we should revisit that as well. (Also just realised this is almost identical step so will remove this one).

olivermrbl (approved):

PR #11963, opened 2025-03-24
feat: run workflow hooks inside a when/then block

thetutlage:
What:
 * Hooks respecting when/then block to be executed.
 * when/then can return hooks to be exported on the workflow

FIXES: SUP-1683


carlos-r-l-rodrigues (approved):

olivermrbl (approved):
Judging from the tests, this looks good to me!

adrien2p (approved):

PR #11992, opened 2025-03-26
fix(utils,promotion): fix case where mutliple percentage promotions werent applied

riqwan:
what:

- fix case where mutliple percentage promotions werent applied


packages/modules/promotion/integration-tests/__tests__/services/promotion-module/compute-actions.spec.ts
```diff
@@ -748,6 +748,81 @@ moduleIntegrationTestRunner({
             ])
           })
 
+          it("should compute actions when 2 percentage promotions are applied to the same item", async () => {
```
olivermrbl, line 751: nit/question: Do we already have a test for `across`?

olivermrbl (approved):

packages/modules/promotion/integration-tests/__tests__/services/promotion-module/compute-actions.spec.ts
riqwan: Yes there are plenty, I saw the previous test was wrong by a small amount and added some precision into the calculation while I was in the area. Nothing much else changes for across. But good push to add the same test for across as well - added!

PR #11999, opened 2025-03-26
fix(dashboard): Product export fields

olivermrbl:
What
- Remove `fields` from query to export products

Why
- The product export route uses the `fields` in the export request if specified. The `fields` didn't used to be part of the request, but in #11532 we accidentally started using a narrow set of fields as a result of shaving down on the requested data elsewhere in the dashboard.


packages/admin/dashboard/src/routes/products/product-export/product-export.tsx
```diff
@@ -25,8 +25,7 @@ export const ProductExport = () => {
 
 const ProductExportContent = () => {
   const { t } = useTranslation()
-  const { searchParams } = useProductTableQuery({})
```
fPolic, line 28: nit: rm unused import

fPolic (approved):
LGTM

PR #12007, opened 2025-03-27
fix(dashboard): return form pending diff

fPolic:
**What**
- display est. difference in returns as for claims/exchanges
- display order pending difference instead of value computed on FE

---

CLOSES CMRC-949


packages/admin/dashboard/src/i18n/translations/en.json
```diff
@@ -1133,7 +1133,7 @@
       "sendNotificationHint": "Notify customer about return.",
       "returnTotal": "Return total",
       "inboundTotal": "Inbound total",
-      "refundAmount": "Refund amount",
+      "refundAmount": "Estimated difference",
```
olivermrbl: nit: should we rename the key?

olivermrbl (approved):

PR #12010, opened 2025-03-27
feat(core-flows,events): emit missing analytics event

VariableVic:
This PR introduces (changes to) the following events emitted from core flows:

*   `cart.updated` (refactored)
*   `payment-session.created` (added)
*   `payment-session.deleted` (added)

All cart updates now emit a `cart.updated` event. The payload structure has been updated for better type safety and to support multiple actions (added, updated, deleted) per field type (`line_items`, `promo_codes`) in a single event:

```ts
// Definition of possible actions
export enum ChangeAction {
  ADDED = "added",
  UPDATED = "updated",
  DELETED = "deleted",
}

// Structure for a single change
type Change<T> = {
  action: ChangeAction
  value: T
}

// Payload for the cart.updated event
// Specific types depend on context (e.g., StoreCartLineItem, AddressDTO)
type CartWorkflowEventPayload = {
  id: string // Cart ID
  changes: {
    // Fields supporting multiple changes (Array of Change objects)
    line_items?: Change<Partial</* StoreCartLineItem | CartLineItemDTO */ any>>[]
    promo_codes?: Change<string>[]

    // Fields supporting a single change (Change object)
    region_id?: Change<string>
    customer_id?: Change<string>
    sales_channel_id?: Change<string>
    email?: Change<string>
    currency_code?: Change<string>
    metadata?: Change<Record<string, unknown>>
    shipping_address?: Change<Partial</* StoreCartAddress | AddressDTO */ any>>
    billing_address?: Change<Partial</* StoreCartAddress | AddressDTO */ any>>
  }
}
```

**Examples:**

*   **Add shipping address to cart:**
    ```json
    {
      "id": "cart_1234567890",
      "changes": {
        "shipping_address": {
          "action": "added",
          "value": { "first_name": "John", "last_name": "Doe", "address_1": "...", ... }
        }
      }
    }
    ```
*   **Add multiple line items:**
    ```json
    {
      "id": "cart_1234567890",
      "changes": {
        "line_items": [
          {
            "action": "added",
            "value": { "variant_id": "variant_abc", "quantity": 1 }
          },
          {
            "action": "added",
            "value": { "variant_id": "variant_def", "quantity": 2 }
          }
        ]
      }
    }
    ```
*   **Remove a line item:**
    ```json
    {
      "id": "cart_1234567890",
      "changes": {
        "line_items": [
          {
            "action": "deleted",
            "value": { "id": "item_xyz" } // Include identifier for the item
          }
        ]
      }
    }
    ```
*   **Update region and add a promo code:**
    ```json
    {
      "id": "cart_1234567890",
      "changes": {
        "region_id": {
           "action": "updated",
           "value": "reg_new_region"
        },
        "promo_codes": [
           {
             "action": "added",
             "value": "SUMMERDEAL"
           }
        ]
      }
    }
    ```

**Why?**

These events are essential for setting up server-side ecommerce analytics in tools like GA and Segment. Some of these actions already emitted a `cart.updated` event, but the previous payload was too generic and didn't support tracking multiple changes efficiently. This new structure provides more detailed, type-safe information.

**How?**

By adding the `emitEventStep` in relevant core flows. In some cases, the events are emitted conditionally based on the changes detected within the workflow.

**Testing?**

The easiest way would be to add test subscribers that listen for the new events and log the payload to verify the structure and data for different cart operations.


.changeset/seven-kids-train.md
VariableVic, line 1: Not sure if this is required, reviewer please check

packages/core/core-flows/src/cart/workflows/add-to-cart.ts
VariableVic, line 1: Emits `cart.item_added` with the added items as the payload

packages/core/core-flows/src/cart/workflows/update-line-item-in-cart.ts
```diff
@@ -146,6 +150,16 @@ export const updateLineItemInCartWorkflow = createWorkflow(
       input: { cart_id: input.cart_id },
     })
 
+    // when the quantity is 0, we emit the item removed event
```
VariableVic, line 153: Emit `cart.item_removed` when a line item quantity reaches 0

packages/core/core-flows/src/line-item/workflows/delete-line-items.ts
```diff
@@ -32,18 +34,31 @@ export const deleteLineItemsWorkflowId = "delete-line-items"
...
+ *
  * @summary
- * 
+ *
  * Delete line items from a cart.
  */
 export const deleteLineItemsWorkflow = createWorkflow(
   deleteLineItemsWorkflowId,
   (input: WorkflowData<DeleteLineItemsWorkflowInput>) => {
+    const items = useQueryGraphStep({
```
VariableVic: If this extra query has a significant negative impact on performance, we can leave it out and handle the query in a subscriber.

packages/core/core-flows/src/cart/workflows/add-to-cart.ts
```diff
@@ -195,10 +195,18 @@ export const addToCartWorkflow = createWorkflow(
...
+      emitEventStep({
+        eventName: CartWorkflowEvents.UPDATED,
+        data: { id: cart.id },
+      }).config({ name: "emit-cart-updated-event" }),
+
+      emitEventStep({
+        eventName: CartWorkflowEvents.ITEM_ADDED,
+        data: { id: cart.id, items: allItems },
+      }).config({ name: "emit-item-added-event" })
+    )
```
srindom: thought: I am not sure that it makes sense to emit multiple events here. 

Maybe a better approach would be to emit more metadata in the cart.updated event. But curious to hear what others think. 

separately, if we go with the item_added event I would rename this to `items_added` since we are passing an array of items and not just a single item.

packages/core/core-flows/src/cart/workflows/update-cart.ts
```diff
@@ -253,9 +253,19 @@ export const updateCartWorkflow = createWorkflow(
...
+    // when the shipping address is added, we emit the shipping address added event
+    when({ input }, ({ input }) => {
+      return isDefined(input.shipping_address)
+    }).then(() => {
+      emitEventStep({
+        eventName: CartWorkflowEvents.SHIPPING_ADDRESS_ADDED,
+        data: { id: input.id },
+      }).config({ name: "emit-shipping-address-added-event" })
+    })
+
```
srindom: again an alternative approach could be to add metadata to the updated event. E.g.:

```
      emitEventStep({
         eventName: CartWorkflowEvents.UPDATED,
         data: { id: input.id, fields: ["shipping_address"] },
       })
```

packages/core/core-flows/src/cart/workflows/add-to-cart.ts
VariableVic: I think emitting the `cart.updated` event makes sense for updates via `updateCartWorkflow` - such as changes to the region, address, etc. - but not for adding or removing line items.

From an analytics perspective, we only need to know _that_ a cart update occurred, not the specific details (e.g., the updated address or region). So, an event like:

```
emitEventStep({
   eventName: CartWorkflowEvents.UPDATED,
   data: { id: input.id, fields: ["shipping_address"] },
})
```

would be sufficient.

However, for line item modifications, we need to capture exactly what was added or removed (including item metadata) to properly populate tracking events. I think emitting dedicated events with a structured, predictable payload for these cases will provide better DX

packages/core/core-flows/src/cart/workflows/update-cart.ts
VariableVic: Think that could be a nice approach for cart updates, see my other response

PR #12012, opened 2025-03-27
chore(utils): fix TSDoc example for payment provider

kunutch:
Docs Link: https://docs.medusajs.com/resources/references/payment/provider


shahednasser (changes requested):
Thank you for your contribution! But this is actually generated from the comments in `packages/core/utils/src/payment/abstract-payment-provider.ts`. can you make the change there instead?

kunutch:
> Thank you for your contribution! But this is actually generated from the comments in `packages/core/utils/src/payment/abstract-payment-provider.ts`. can you make the change there instead?

Thanks for checking my PR. Updated ✅

shahednasser (approved):
LGTM thanks!

PR #12021, opened 2025-03-27
fix(dashboard): RMA allow shipping unset

fPolic:
**What**
- allow unsetting shipping in RMA flows + bug cleanup around shipping flows in claims / exchanges
- additional fixes:
  - product type create form 
  - product type delete
  - delete confirmation messages
  - fix a few inventory strings + material field

---

CLOSES CMRC-945
CLOSES CMRC-953
CLOSES CMRC-967
CLOSES CMRC-963
CLOSES CMRC-954


packages/admin/dashboard/src/routes/orders/order-create-claim/components/claim-create-form/claim-outbound-section.tsx
```diff
@@ -403,10 +407,11 @@ export const ClaimOutboundSection = ({
                   <Form.Item>
                     <Form.Control>
                       <Combobox
+                        allowClear
```
olivermrbl, line 410: question: Does `allowClear` remove the shipping option on the BE too?

For example:
1. You add shipping -> send a request
2. You clear shipping -> do we send a request here?

The reason I am asking is that we now have a check for `selectedOptionId`, which I presume would be undefined in the second case.

fPolic: Yes we clear all the other (outbound in this case) shipping methods that were added for this order change since we support only one inbound/outbound shipping from the admin flow.

So if `selectedOptionId ` is undefined we only send a request to remove existing shipping method that is result of `SHIPPING_ADD` but we do not send a request to add a shipping method.

olivermrbl (approved):
LGTM, niiice

PR #12025, opened 2025-03-28
fix: validate module names to disallow unallowed characters

thetutlage:


carlos-r-l-rodrigues (approved):

packages/core/utils/src/common/validate-module-name.ts
```diff
@@ -0,0 +1,25 @@
...
+ * a very simple example, but cases like these will lead to naming
+ * conflicts, so its better to use the names as it is and restrict
+ * them to be variable safe
+ */
+const RE = /^[a-zA-Z_][0-9a-zA-Z_]*$/
+export function validateModuleName(name: string) {
+  if (!RE.test(name)) {
+    throw new Error(
+      `Invalid module name "${name}". Module names must be alpha numeric and may contain an underscore`
+    )
```
olivermrbl, line 23: suggestion: Should we log the error and skip the module registration to avoid the breaking change?

thetutlage: You mean still register the module?

Wouldn't that eventually fail when you will try to run migrations or generate GQL schema?

olivermrbl: Can we say with certainty that module names that don't follow these rules would have led to errors at some point, e.g. for running migrations? 

If that's the case, I'd be fine to proceed with this. But if not, I think we should log + skip the registration.

thetutlage: I can run some tests locally

thetutlage: Yes, I can confirm that using any of the unallowed values doesn't work. I cannot start the server

olivermrbl: Great, thanks for testing. Let's merge as is

PR #12027, opened 2025-03-28
feat(orchestration): skip on permanent failure

carlos-r-l-rodrigues:
What:
 - Added step config `skipOnPermanentFailure`. Skip all the next steps when the current step fails. If a string is used, the workflow will resume from the given step.
 - Fix `continueOnPermanentFailure` to continue the execution of the flow when a step fails.
 
```ts
createWorkflow("some-workflow", () => {
  errorStep().config({
    skipOnPermanentFailure: true,
  })
  nextStep1() // skipped
  nextStep2() // skipped
})


createWorkflow("some-workflow", () => {
  errorStep().config({
    skipOnPermanentFailure: "resume-from-here",
  });
  nextStep1(); // skipped
  nextStep2(); // skipped
  nextStep3().config({ name: "resume-from-here" }); // executed
  nextStep4(); // executed
});
```


NicolasGorga:
Hey @carlos-r-l-rodrigues was this included in 2.7.0? 

Inside a step i have an unhandled error and even though i've configured it inside of my workflow with option continueOnPermanentFailure to true, it is not continuing.

carlos-r-l-rodrigues:
Oh, we missed this PR @NicolasGorga.
Thanks for the heads up, we'll soon have a release w/ some patches and bug fixes and I will make sure this one is included.

packages/core/orchestration/src/transaction/transaction-orchestrator.ts
```diff
@@ -696,12 +700,28 @@ export class TransactionOrchestrator extends EventEmitter {
...
+          if (step.definition.skipOnPermanentFailure) {
+            const until = isString(step.definition.skipOnPermanentFailure)
+              ? step.definition.skipOnPermanentFailure
+              : undefined
+
+            let stepsToSkip: string[] = [...step.next]
+            while (stepsToSkip.length > 0) {
+              const currentStep = flow.steps[stepsToSkip.pop()!]
+
+              if (until && currentStep.definition.action === until) {
```
riqwan, line 716: q: how would this work with parallelize(step1, step2, step3)?

packages/core/orchestration/src/transaction/transaction-orchestrator.ts
```diff
@@ -696,12 +700,28 @@ export class TransactionOrchestrator extends EventEmitter {
...
-          step.definition.continueOnPermanentFailure &&
+          (step.definition.continueOnPermanentFailure ||
+            step.definition.skipOnPermanentFailure) &&
           !TransactionTimeoutError.isTransactionTimeoutError(timeoutError!)
         ) {
-          for (const childStep of step.next) {
-            const child = flow.steps[childStep]
-            child.changeState(TransactionStepState.SKIPPED_FAILURE)
+          if (step.definition.skipOnPermanentFailure) {
+            const until = isString(step.definition.skipOnPermanentFailure)
```
riqwan, line 708: q: should we validate the existence of this step id?

riqwan (approved):
Asked a couple of questions, but overall lgtm!

packages/core/orchestration/src/transaction/transaction-orchestrator.ts
carlos-r-l-rodrigues: Good push. Just added validation to it.

packages/core/orchestration/src/transaction/transaction-orchestrator.ts
carlos-r-l-rodrigues: At the Transaction Orchestrator level, it will skip all the steps that are children of the step that has failed.
```
[
  a - a.1 - a.2,
  b - b.1 - b.2 - b.3,
  c - c.1,
]
```
if `b` fails, `b.1, b.2, b.3` will be skipped, but the rest will continue.

For our workflow SDK that has a limitation where you can't have multiple next steps for each "branch" parallelized, next steps are only append as child of the last step defined on the `parallelize` 

```
[
  a,
  b,
  c - c.1 - c.2 - c.3,
]
```
If `a` or `b` fail, nothing will be skipped, only if `c` fail it will skip the children steps.

thetutlage (approved):

riqwan (approved):

shahednasser (approved):

PR #12039, opened 2025-03-29
feat: add `hasMany` flag to enforce in app link uniqueness

thetutlage:
Earlier we used the `isList` flag to enforce uniqueness constraint on a link. However, that leads to a breaking change with the remote query when the `isList` is not explicitly provided. For example:

When `isList` is not provided, we want the `isList` to be `false` (as earlier). However, we want to allow creating multiple links from the primary service. This lead us to changing the default from `isList=false` to `isList=true`, but that had a breaking change in the remote query types and naming scheme.

In this PR, we separate both the concepts and have an internal flag called `createMultiple`. This flag has different defaults from the `isList` flag but infers its value from the `isList` flag when defined explicitly.

### DIFF
--- a/integration-tests/modules/__tests__/link-modules/define-link.spec.ts
+++ b/integration-tests/modules/__tests__/link-modules/define-link.spec.ts
@@ -46,7 +46,8 @@ medusaIntegrationTestRunner({
               entity: "Currency",
               primaryKey: "code",
               foreignKey: "currency_code",
-              isList: true,
+              isList: false,
+              createMultiple: true,
               alias: "currency",
               args: {
                 methodSuffix: "Currencies",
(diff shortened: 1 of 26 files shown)


olivermrbl:
@carlos-r-l-rodrigues, it would be great if you could take a look at this one

carlos-r-l-rodrigues (approved):
LGTM

adrien2p:
As per our discussion:
- renaming: createMultiple -> hasMany
- remove isList from link relationships (but kept on the extends relationships/aliases as it server the query purpose)
- fix the constraint for the links creation

cc @carlos-r-l-rodrigues  @olivermrbl

carlos-r-l-rodrigues:
@adrien2p @olivermrbl this one is ready

adrien2p (approved):

PR #12072, opened 2025-04-03
fix(order): update add item unit price

fPolic:
**What**
- spread of existing `lineItem` would contain raw properties which would override newly set unit_price

---

CLOSES CMRC-1005


olivermrbl (approved):
LGTM, can we add a test for this? This can be a unit test in the order module if we want to keep it simple.

PR #12088, opened 2025-04-05
Update algolia page.mdx

huynhdieutuong:
Update some routes from store/products/search/route to store/search/route


shahednasser (changes requested):
Thank you for pointing this out! Actually the error is that the route should be created in `src/api/store/products/search/route.ts`. can you instead change the file path to fix that and revert the other changes?

shahednasser:
Closing this PR as it's been fixed by #12141

PR #12091, opened 2025-04-05
update zod

tomer-yechiel:
zod added support for [standard schema](https://standardschema.dev/) at version 3.24.0

needed for using Tanstack form

thanks


tomer-yechiel:
..

PR #12097, opened 2025-04-07
chore(): Emit events in batch and index process event ids in batch

adrien2p:
**What**
First iteration to prevent events from overwhelming the systems.
- Group emitted event ids when possible instead of creating a message per id which leads to reduced amount of events to process massively in cases of import for example
- Update the index engine to process event data in batches of 100
- Update event handling by the index engine to be able to upsert by batch as well
- Fix index engine build config for intermediate listeners inferrence


packages/modules/link-modules/src/services/link-module-service.ts
```diff
@@ -388,4 +384,21 @@ export default class LinkModuleService implements ILinkModule {
   ): Promise<[object[], Record<string, string[]>]> {
     return await this.linkService_.restore(data, sharedContext)
   }
+
+  protected async emitEvents_(groupedEvents) {
```
thetutlage, line 388: Where is this method used?

thetutlage (approved):

packages/modules/link-modules/src/services/link-module-service.ts
adrien2p: by the emit events decorator which would otherwise warn that this is not implemented

thetutlage: Ahh okay.

olivermrbl (approved):
Epic work guys

PR #12115, opened 2025-04-08
fix(core-flows): draft order reservations

fPolic:
**What**
- On edit confirmation all previous reservations are deleted. When we confirm an edit we create reservations only for added/updated items which would leave previous items without recreated reservations.
- Dispatch order placed event when a draft order is converted to an order.

---

CLOSES CMRC-1015


packages/core/core-flows/src/draft-order/workflows/confirm-draft-order-edit.ts
```diff
@@ -114,19 +114,25 @@ export const confirmDraftOrderEditWorkflow = createWorkflow(
       throw_if_key_not_found: true,
     }).config({ name: "order-items-query" })
 
-    const lineItemIds = transform(
+    const { removedLineItemIds } = transform(
```
olivermrbl, line 117: comment: We should be able to write a test for this no?

fPolic: @oli - added

olivermrbl (approved):

PR #12120, opened 2025-04-08
chore(pricing): improve calculate prices performances

adrien2p:
**What**
Optimize the pricing retrieval query to improve performances (especially on large dataset)

It removed the overall sub query and many joins in favor of single join and very specific sub queries for rules check in order to benefit from indexes and pre filtered data.

Also, the produced query will be very much simpler if there is no context for rules check, in that case we only want prices that have 0 rules since all other prices would fail to match their expectations leading to more performant queries in these cases as well.

The resulting query plan is much cleaner and rely more on filtered data and indexes as well as sub plan that gets executed only when required and pre filtered the data

**TEST**
We ran some test on dataset containing about 1.2M prices and 1.2M price rules
1. The original query takes about 718ms execution time with a planning of 0.5ms
2. The new query takes about 17ms with a slightly higher execution plan time of 3.7ms

Overall this indicates a performance improvement at the query level of 42 times better performances form the original query


packages/modules/pricing/src/repositories/pricing.ts
```diff
@@ -53,28 +52,14 @@ export class PricingRepository
       )
     }
 
-    const isContextPresent = Object.entries(context).length || !!currencyCode
-
-    // Only if the context is present do we need to query the database.
-    // We don't get anything from the db otherwise.
-    if (!isContextPresent) {
+    // Early return if no context
+    if (!Object.keys(context).length && !currencyCode) {
```
riqwan:
```suggestion
    if (!Object.keys(context).length) {
```

packages/modules/pricing/src/repositories/pricing.ts
```diff
@@ -95,198 +80,197 @@ export class PricingRepository
...
+            WHERE pr.price_id = price.id 
+            AND pr.deleted_at IS NULL
+            AND (
+              ${flattenedContext
+                .map(([key, value]) => {
                   if (typeof value === "number") {
-                    buildOperatorQueries(orBuilder, value)
+                    return `
+                    (pr.attribute = ? AND (
+                      (pr.operator = 'eq' AND pr.value::numeric = ?) OR
```
riqwan:
```suggestion
                      (pr.operator = 'eq' AND pr.value = ?) OR
```

packages/modules/pricing/src/repositories/pricing.ts
```diff
@@ -95,198 +80,197 @@ export class PricingRepository
...
+            SELECT COUNT(*) 
+            FROM price_rule pr
+            WHERE pr.price_id = price.id 
+            AND pr.deleted_at IS NULL
+          )
+        )
+        `,
+        flattenedContext.flatMap(([key, value]) => {
+          if (typeof value === "number") {
+            return [key, value, value, value, value, value]
```
riqwan: This will fix a bug we have right now

```suggestion
            return [key, value.toString(), value, value, value, value]
```

packages/modules/pricing/src/repositories/pricing.ts
```diff
@@ -95,198 +80,197 @@ export class PricingRepository
...
+      const priceRuleConditions = knex.raw(
+        `
+        (
+          price.rules_count = 0 OR
+          (
+            /* Count all matching rules and compare to total rule count */
+            SELECT COUNT(*) 
+            FROM price_rule pr
+            WHERE pr.price_id = price.id 
+            AND pr.deleted_at IS NULL
```
riqwan, line 148: nice, this was missed previously!

riqwan (approved):
nice work! 🔥 

Do you have any stats on this?

packages/modules/pricing/src/repositories/pricing.ts
```diff
@@ -95,198 +75,197 @@ export class PricingRepository
...
+                    ))
+                    `
                   } else {
                     const normalizeValue = Array.isArray(value)
                       ? value
                       : [value]
-
-                    orBuilder.whereIn("pr.value", normalizeValue)
+                    const placeholders = normalizeValue.map(() => "?").join(",")
+                    return `(pr.attribute = ? AND pr.value IN (${placeholders}))`
```
peterlgh7, line 167: does `placeholders` come from user input?

adrien2p: the placeholders are just the `?` char representing bindings to pick from the bindings list provided later on that gets sanitized to prevent injection 👍

olivermrbl (approved):

PR #12128, opened 2025-04-09
chore(pricing): Pricing retrieval improvements

adrien2p:
**What**
I have removed the check for the context key where it was fetching all attributes available and then stripping out the one that does not exists.. On big dataset these would remove multiple hundreds of ms of query execution


packages/modules/pricing/src/repositories/pricing.ts
```diff
@@ -53,28 +52,49 @@ export class PricingRepository
...
+    const candidateContext = Object.entries(flattenedKeyValuePairs).filter(
+      ([, value]) => {
         const isValuePresent = !Array.isArray(value) && isPresent(value)
         const isArrayPresent = Array.isArray(value) && value.flat(1).length
 
-        return (
-          allowedRuleAttributes.includes(key) &&
-          (isValuePresent || isArrayPresent)
-        )
+        return isValuePresent || isArrayPresent
```
riqwan, line 61: This just needs the test we spoke about regarding a full cart context, if its no longer needed, killer! 🔥

packages/modules/pricing/src/repositories/pricing.ts
```diff
@@ -53,28 +52,49 @@ export class PricingRepository
...
+    //   `,
+    //       [...contextKeys, ...contextKeys]
+    //     )
+    //     .then((result) => result.rows.map((row) => row.attribute))
+
+    //   // Filter context to only include keys that exist in rules
+    //   flattenedContext = candidateContext.filter(([key]) =>
+    //     existingKeys.includes(key)
+    //   )
+    // }
```
riqwan:
```suggestion
const flattenedContext = candidateContext
```

This is a nice alternative if the test doesn't pass! Remove if it does 🤞🏻 

Irregardless, i think this would still reduce the query size, might be good to have.

riqwan (approved):

packages/modules/pricing/src/repositories/pricing.ts
carlos-r-l-rodrigues: let's just get this removed and it's perfect.

carlos-r-l-rodrigues (approved):

packages/modules/pricing/src/repositories/pricing.ts
adrien2p: I ve fixed the flatten util now it should be good to remove and I did just that 😂

packages/modules/pricing/integration-tests/__tests__/services/pricing-module/calculate-price.spec.ts
```diff
@@ -286,6 +286,171 @@ moduleIntegrationTestRunner<IPricingModuleService>({
...
+                  updated_at: "2025-04-09T14:59:24.250Z",
+                  deleted_at: null,
+                },
+              ],
+            },
+            promotions: [],
+          }
+
+          const calculatedPrice = await service.calculatePrices(
+            { id: ["price-set-EUR", "price-set-PLN"] },
```
riqwan, line 414: You might need to setup a complex rule for this?

riqwan (approved):

olivermrbl (approved):
Nice, let's get this included in the release 💪

PR #12129, opened 2025-04-09
chore(promotion): Improve performances [1]

adrien2p:
**What**
Reduce database queries when possible and use proper data structure and aggregation when possible in order to reduce performance decrease overall


packages/modules/promotion/src/services/promotion-module.ts
```diff
@@ -144,24 +144,23 @@ export default class PromotionModuleService
...
         },
         {
-          status: PromotionStatus.ACTIVE,
           ...filters,
           campaign: {
             ...filters?.campaign,
+            $or: [{ starts_at: null }, { starts_at: { $lte: now } }],
             $and: [
               {
-                $or: [{ starts_at: null }, { starts_at: { $lte: new Date() } }],
```
adrien2p, line 161: comment: this could have created a bug eventually

packages/modules/promotion/src/services/promotion-module.ts
```diff
@@ -144,24 +144,23 @@ export default class PromotionModuleService
...
-          status: PromotionStatus.ACTIVE,
           campaign_id: null,
           ...filters,
         },
         {
-          status: PromotionStatus.ACTIVE,
           ...filters,
           campaign: {
             ...filters?.campaign,
+            $or: [{ starts_at: null }, { starts_at: { $lte: now } }],
```
riqwan: q: shouldn't this be a part of line 163?

packages/modules/promotion/src/services/promotion-module.ts
```diff
@@ -225,16 +222,20 @@ export default class PromotionModuleService
...
+        // Check if it exceeds the limit and cap it if necessary
         if (
           campaignBudget.limit &&
-          MathBN.gt(campaignBudgetData.used, campaignBudget.limit)
+          MathBN.gt(newUsedValue, campaignBudget.limit)
         ) {
-          continue
+          campaignBudgetData.used = campaignBudget.limit
+        } else {
+          campaignBudgetData.used = newUsedValue
```
riqwan, line 239: nice catch!

todo: you can skip the capping here and always assign `newUsedValue`, we allow going over the limit in the same transaction.

packages/modules/promotion/src/utils/compute-actions/buy-get.ts
```diff
@@ -128,101 +130,113 @@ export function getComputedActionsForBuyGet(
...
   )
 
-  const targetQuantity = MathBN.convert(
-    promotion.application_method?.apply_to_quantity ?? 0
-  )
+  if (!eligibleTargetItems.length) {
+    return computedActions
+  }
 
-  /*
```
riqwan, line 143: can we keep all these comments thats been removed? 😄

riqwan (approved):
lgtm!

packages/modules/promotion/src/services/promotion-module.ts
adrien2p: the constraint will be equivalent but for readability I ll do that 👍

adrien2p:
@riqwan I ve applied your feedback if you could have a look please :)

packages/modules/promotion/src/services/promotion-module.ts
adrien2p: when removing this constraint it break one of the test 🤔

riqwan (approved):
nice, lgtm!

packages/modules/promotion/src/services/promotion-module.ts
adrien2p: dont know if you have seen my comment @riqwan 😄

riqwan: ah no, sorry it was resolved. Yeah, I think the continue might be necessary here. Can't remember for what reason though. If a test is failing its a good indication that this was added on purpose 😄 

Essentially, when a promotion is applied, but there is a little bit of budget left, we allow all promotions with the same budget to pass through and go over.

PR #12137, opened 2025-04-10
chore(medusa-test-utils): Prevent waiting for event indefinately

adrien2p:
**What**
Currently the util await for event infinitely, this can lead to chain crashes in the jest tests suites leading to too much noise to investigate proper issues.
We now have a default time out raced against the promise that is configurable to prevent from waiting for an excessive amount of time


sradevski (approved):
Nice!

olivermrbl (approved):

PR #12140, opened 2025-04-10
Add support for dynamoDB for storing sessions and some types cleanup

thetutlage:
Fixes: FRMW-2939

In this PR we have added the support for using DynamoDB for storing sessions. Here's how it will work.

- The `defineConfig` method will set the `dynamoDBOptions` when we are running in Cloud and have defined `SESSION_STORE=dynamodb` environment variable. It will allow us to slowly roll dynamoDB setup to certain customers and not all of them.
- If `SESSION_STORE=dynamodb` is set. It becomes the responsibility of the build pipeline to perform the following tasks.
   - Set `AWS_REGION`, `DYNAMO_DB_ENDPOINT`, `AWS_ACCESS_KEY_ID`, `AWS_SECRET_ACCESS_KEY` environment variables. These are read by the AWS SDK automatically except the `DYNAMO_DB_ENDPOINT` variable.
   - Install `connect-dynamodb@3.0.5` and `@aws-sdk/client-dynamodb@3.218.0` in the app as dependencies

Once these conditions are met, we will be able to use DynamoDB for storing sessions.


packages/core/framework/src/config/types.ts
```diff
@@ -1,978 +1,8 @@
...
-  ExternalModuleDeclaration,
-  InternalModuleDeclaration,
+export {
+  AdminOptions,
+  SessionOptions,
+  ProjectConfigOptions,
+  HttpCompressionOptions,
+  ConfigModule,
+  PluginDetails,
 } from "@medusajs/types"
```
thetutlage, line 8: These were duplicated with slight variations. Ideally we should have one copy. But happy to revert if there was a reason for duplicating them

packages/core/framework/src/http/express-loader.ts
```diff
@@ -37,7 +38,7 @@ export async function expressLoader({ app }: { app: Express }): Promise<{
     name: sessionOptions?.name ?? "connect.sid",
     resave: sessionOptions?.resave ?? true,
     rolling: sessionOptions?.rolling ?? false,
-    saveUninitialized: sessionOptions?.saveUninitialized ?? true,
+    saveUninitialized: sessionOptions?.saveUninitialized ?? false,
```
thetutlage, line 41: This was the root cause of creating too many sessions. Using `true` is not recommended by `express-session` package and they have deprecated their default true value now

packages/core/framework/src/http/express-loader.ts
```diff
@@ -50,7 +51,15 @@ export async function expressLoader({ app }: { app: Express }): Promise<{
...
-  if (configModule?.projectConfig?.redisUrl) {
+  if (configModule?.projectConfig.dynamodbOptions) {
+    const storeFactory = await dynamicImport("connect-dynamodb")
+    const client = await dynamicImport("@aws-sdk/client-dynamodb")
+    const DynamoDBStore = storeFactory({ session })
+    sessionOpts.store = new DynamoDBStore({
+      ...configModule?.projectConfig.dynamodbOptions,
+      client: new client.DynamoDBClient(),
+    })
+  } else if (configModule?.projectConfig?.redisUrl) {
```
thetutlage: Preferred over Redis when the config is defined

packages/core/utils/src/common/to-number.ts
```diff
@@ -0,0 +1,15 @@
+/**
+ * Transforms a value to number or returns the default value
+ * when original value cannot be casted to number
+ */
+export function toNumber(value: unknown): number | undefined
+export function toNumber<T>(value: unknown, defaultValue: T): number | T
+export function toNumber<T>(
```
carlos-r-l-rodrigues: nit: I think this name is a bit confusing if we are not returning always a number...
maybe `tryConvertToNumber`?

packages/core/framework/src/config/config.ts
```diff
@@ -164,7 +164,9 @@ export class ConfigManager {
 
     this.#config = {
       projectConfig: normalizedProjectConfig,
-      admin: projectConfig.admin ?? {},
+      admin: projectConfig.admin ?? {
+        path: "/",
```
olivermrbl: question: Just out of curiousity, why was this added?

packages/core/types/src/common/config-module.ts
```diff
@@ -354,11 +364,39 @@ export type ProjectConfigOptions = {
...
+   * - @aws-sdk/client-dynamodb@^3.218.0
+   * - connect-dynamodb@^3.0.5
+   *
+   * Environment variables to set
+   * - AWS_REGION=<value>
+   * - ENDPOINT=<value>
+   * - AWS_ACCESS_KEY_ID=<value>
+   * - AWS_SECRET_ACCESS_KEY=<value>
+   */
+  dynamodbOptions?: {
```
olivermrbl: thought(non-blocking): Should we nest this within `sessionStore` to make it more obvious what it is intended for?

packages/core/framework/src/config/types.ts
olivermrbl: I guess this is technically a breaking change, so should we instead deprecate the types? We can perhaps assign these types to the other exported ones to remove the duplicated code.

packages/core/types/src/common/config-module.ts
thetutlage: We have similar setup for redis. So should be also nest that within the `sessionStore`? Because there could be case where I want to use separate Redis stores for events and for storing sessions

packages/core/framework/src/config/types.ts
thetutlage: I think renaming these to something else will possibly create two copies of types. One exported from `@medusajs/types` and one exported from `@medusajs/framework` with a different name.

So, if we want to prevent the breaking change, then I can revert this

packages/core/framework/src/config/config.ts
thetutlage: This is because, the `path` was optional in the duplicated types defined by the `framework` https://github.com/medusajs/medusa/blob/develop/packages/core/framework/src/config/types.ts#L59

Now, if you look at the `admin-bundler` it picks its types from the `@medusajs/types` package and therefore the `path` property is required by it. https://github.com/medusajs/medusa/blob/develop/packages/admin/admin-bundler/src/types.ts#L1

So essentially, framework having its own types could have runtime values mismatch when passing config to the other parts of the stack

packages/core/types/src/common/config-module.ts
olivermrbl: > Because there could be case where I want to use separate Redis stores for events and for storing sessions

I am not sure I understand how this is related to the config–can you elaborate?

thetutlage: Currently when we pick Redis as the session store for storing the session data, we rely on top-level `projectConfig.redisUrl` and `projectConfig.redisOptions` config variables. 

So for consistency both Redis and DynamoDB should be at the same level

packages/core/framework/src/config/config.ts
olivermrbl: Maybe I am mistaken, but I believe the default path for admin is `/app`, so this would be a breaking change (that we don't want to do)? I might be wrong though, but might be good to double check the expected behavior here.

thetutlage: Yup, you are right!

packages/core/types/src/common/config-module.ts
olivermrbl: Redis being top-level is more a legacy thing than intended actually. I would be OK with moving the dynamo DB options to `sessionStore` and keep Redis as is to avoid breaking changes.<br><br>The thinking here is that suddenly having `dynamodbOptions` live top-level in the project config might be cause of confusion. What is it for and do I need it?

This is not a strong opinion though, so happy to go with whatever.

packages/core/framework/src/config/config.ts
olivermrbl: I actually don't think we need this default value, as the default admin values are set in `normalizeAdminConfig` in `defineConfig`

packages/core/types/src/common/config-module.ts
thetutlage: Yup happy to move it to the `sessionStore` 👍

packages/core/framework/src/config/config.ts
thetutlage: In that case we might need a bit more strict types and here's why.

- Most of the types are written from perspective of what the end-user types.
- However, we go with the assumption that all configs returned from the `medusa-config.ts` file are using the `defineConfig` method and therefore these normalizations have been applied.

In that case, there should be an `InputConfig` type (that we have but not for all nested properties) and then the `OutputConfig` type that reflects the normalizations being applied.

packages/core/framework/src/config/types.ts
olivermrbl: Do we have to rename them?

The thinking here would be to deprecate them and do:
```
import { TypeFromTypesPackage } from "@medusajs/types"

export type ThisType = TypeFromTypesPackage
```

olivermrbl (approved):

packages/core/framework/src/config/types.ts
thetutlage: Actually the names were the same. Now, if I want to deprecate the one from the framework package then either they have to named differently or the re-export from the types package has to have a different name.

```ts
import { AdminOptions } from '@medusajs/types'
export { AdminOptions }

// 👇 Duplicate name
// @deprecated
export type AdminOptions = {} 
```

sgirones:
this is great. We will also need the following options to be configurable through an env var:
* prefix
* hashKey
* initialized

packages/core/utils/src/common/__tests__/define-config.spec.ts
```diff
@@ -1147,6 +1153,389 @@ describe("defineConfig", function () {
...
+
+    process.env.EXECUTION_CONTEXT = "medusa-cloud"
+    process.env.REDIS_URL = "redis://localhost:6379"
+    process.env.S3_FILE_URL = "https://s3.amazonaws.com/medusa-cloud-test"
+    process.env.S3_PREFIX = "test"
+    process.env.S3_REGION = "us-east-1"
+    process.env.S3_BUCKET = "medusa-cloud-test"
+    process.env.S3_ENDPOINT = "https://s3.amazonaws.com"
+    process.env.SESSION_STORE = "dynamodb"
+    process.env.DYNAMO_DB_ENDPOINT = "http://localhost:8000"
```
sgirones: this is not used if I'm not mistaken

sgirones:
tested in AWS and works great!

sgirones (approved):

packages/core/utils/src/common/__tests__/define-config.spec.ts
thetutlage: The `DYNAMO_DB_ENDPOINT` is not used and rest are used

packages/core/framework/src/http/express-loader.ts
```diff
@@ -50,7 +51,17 @@ export async function expressLoader({ app }: { app: Express }): Promise<{
 
   let redisClient: Redis
 
-  if (configModule?.projectConfig?.redisUrl) {
+  if (configModule?.projectConfig.sessionOptions?.dynamodbOptions) {
```
adrien2p, line 54: Q: Do we need to update the shutdown function or is it not necessary?

adrien2p (approved):

packages/core/framework/src/http/express-loader.ts
thetutlage: When I looked into their source code, I found they do not have any long-lived connections like Redis. However, there is a `destroy` method to close the connection of an HTTP request, which seems like an HTTP request made by an executed DynamoDB command.

thetutlage: CleanShot 2025-04-16 at 14 52 32@2x

Calling `dynamoDbClient.destroy` will call this method under the hood and I think we don't want to do this.

PR #12144, opened 2025-04-10
chore: Cache available price rule attributes

adrien2p:
**What**
We found out that the pricing context from the cart always contains the entire cart, even though it is kind of wrong. The issue is that even though we improve the performances of the query, it will cost a lot to have hundreds of constraint for nothing potentially. For that reason, we cache the attributes in memory with the best possible query we can do to gather them and we renew them when we perform a calculate prices if it has been reset. That way, we ensure we don't have unnecessary checks on attributes that does not have rules.

Since we don't have the type table anymore which was doing that for us and until we have a proper caching layer it would do IMO. But the rules type table was very useful for these attributes findings


carlos-r-l-rodrigues (approved):

packages/modules/pricing/src/repositories/pricing.ts
```diff
@@ -19,12 +19,46 @@ export class PricingRepository
...
+    rows.forEach(({ attribute }) => {
+      this.#availableAttributes.add(attribute)
+    })
+  }
+
+  async #getCachedAvailableAttributes() {
+    if (this.#availableAttributes.size === 0) {
+      await this.#cacheAvailableAttributes()
+    }
+  }
```
olivermrbl: super-nit(non-blocking): This is very nitpicky, but the name here indicates a return value. I would either return the attributes or rename them to `checkForCachedAvailableAttributes`

Also, feel free to discard this, just a thought to improve readability for the next dev coming by this :)

olivermrbl (approved):

packages/modules/pricing/src/repositories/pricing.ts
adrien2p: done, I had the same feeling :)

PR #12150, opened 2025-04-11
chore(product): Missing index in migration

adrien2p:
**What**
Align snapshot and migration by re running the migration:create command. It adds missing indexes


adrien2p:
@thetutlage wdyt could have happen in your opinion? someone manually removed something and we got discrepancies between migrations and snapshot?

olivermrbl (approved):

PR #12157, opened 2025-04-11
fix(): Event group id propagation and event managements

adrien2p:
**What**
- Event group id was not propagated through the entire chain as expected, leading to some events being fired early without being part of any event groups
- Copy events before clearing them to prevent race conditions where the reference gets deleted before being used
- Make copied context properties enumerable to prevent loosing event group id and other properties from chained decorator
- Preserve event group id on workflow continuation
- Fix save checkpoint during next execution


packages/modules/event-bus-local/src/services/event-bus-local.ts
```diff
@@ -112,7 +113,8 @@ export default class LocalEventBusService extends AbstractEventBusModuleService
   }
 
   async releaseGroupedEvents(eventGroupId: string) {
-    const groupedEvents = this.groupedEventsMap_.get(eventGroupId) || []
+    let groupedEvents = this.groupedEventsMap_.get(eventGroupId) || []
+    groupedEvents = JSON.parse(JSON.stringify(groupedEvents))
```
thetutlage, line 117: Do you think this could have some major impact on blocking the event loop when the grouped events will be huge? Like 200K ids or we will never reach that number

adrien2p: in theory we should not reach that, it should be batched by smaller group a bit like we do with a limit of 100 at a time. But indeed people are very capable of sending millions of data, but I believe it would create other issues before reaching that place. What would you suggest instead?

thetutlage: No recommendation to be honest. I was thinking out loud to see if we need cloning of data. But, I think you would have thought about that already 👍

integration-tests/http/package.json
```diff
@@ -5,7 +5,7 @@
   "license": "MIT",
   "private": true,
   "scripts": {
-    "test:integration": "NODE_OPTIONS=--experimental-vm-modules jest --no-cache --maxWorkers=50% --bail --detectOpenHandles --forceExit --logHeapUsage",
+    "test:integration": "NODE_OPTIONS=--experimental-vm-modules jest --no-cache --maxWorkers=50% --bail --detectOpenHandles --forceExit --logHeapUsage -- __tests__/product/admin/product-import.spec.ts",
```
olivermrbl: nit: revert

olivermrbl (approved):
Nice catch :)

PR #12172, opened 2025-04-13
feat(core-flows,js-sdk,medusa,types): draft order delete

fPolic:
What
- support removal of draft orders
- override `deleteOrders` to remove related entities
- fix issue where duplicate addresses were created on order create
- add a migration for missing `order` <> `order_summary` FK

---

CLOSES CMRC-1016


olivermrbl (dismissed):
LGTM, does this delete "cascade" to all related entities, including those local to the order module and anything that is linked to it (if any)?

fPolic:
Delete cascades to items, shipping_methods etc.
I would additionally consider deleting the `OrderChanges` of the deleted order from the workflow? cc @carlos-r-l-rodrigues

carlos-r-l-rodrigues:
To cascade delete all internal relations we need to overwrite the method on the Order service and also cleanup actions and potentially other things related to the order that don't have the db cascade.
Also, the workflow should call the Link.remove (there's a step for that) to remove all the links associated with the deleted draft-orders as well.

olivermrbl:
> To cascade delete all internal relations we need to overwrite the method on the Order service and also cleanup actions and potentially other things related to the order that don't have the db cascade.

@carlos-r-l-rodrigues, aside from what is addressed in this PR, can you think of other related resources, we need to delete manually?

olivermrbl:
@carlos-r-l-rodrigues, any chance you could help with a review of this one?

packages/modules/order/src/services/order-module-service.ts
```diff
@@ -844,6 +851,45 @@ export default class OrderModuleService
...
+
+    await Promise.all([
+      this.orderAddressService_.delete(orderAddressIds, sharedContext),
+      this.orderChangeService_.delete(orderChangeIds, sharedContext),
+    ])
+
+    // Delete order changes & actions
+    await this.orderChangeService_.delete(orderChangeIds, sharedContext)
+
+    await super.deleteOrders(ids, sharedContext)
```
carlos-r-l-rodrigues, line 907: We can add a couple more entities to be deleted here:
Follow the pseudo code of the tables. You can just use the internal services to delete them.

```
select item_id from order_item where order_id in (ids)

delete from "order_line_item" where id in ( [result of previous query] )

delete order_item where order_id in (ids)


select shipping_method_id from order_shipping where order_id in (ids)

delete from order_shipping_method where order_id in ( [result of previous query] )

delete from order_shipping where order_id in (ids)

delete order_summary where order_id in (ids)
delete from order_transaction where order_id in ()
```

These will delete in cascade item and shipping adjustments and tax_lines.

carlos-r-l-rodrigues (commented):
It looks good overall.
As we are overriding the `deleteOrders` method, I believe we can add some more tables that can be cleaned up.

carlos-r-l-rodrigues (approved):
LGTM

packages/modules/order/src/services/order-module-service.ts
fPolic: Some of these relations have cascade already, I added deletion for ones that don't have cascade (`order_line_item `, `order_shipping_method`).

---

Order `summary` has cascade defined in dml/snapshot but `order_summary_order_id_foreign` FK is missing so I added a migration. cc @olivermrbl.

packages/modules/order/src/migrations/Migration20250522181137.ts
```diff
@@ -0,0 +1,21 @@
+import { Migration } from "@mikro-orm/migrations"
+
+export class Migration20250522181137 extends Migration {
+  override async up(): Promise<void> {
+    this.addSql(
+      `DELETE FROM "order_summary" WHERE "order_id" NOT IN (SELECT id FROM "order");`
```
olivermrbl, line 6: question: Is this a safe operation? Seems a bit drastic to delete all orphaned order summaries, but perhaps, they aren't used so it's more a clean-up that anything? @carlos-r-l-rodrigues

packages/modules/order/src/services/order-module-service.ts
```diff
@@ -844,6 +851,70 @@ export default class OrderModuleService
...
+
+    const orderShippingMethodIds = orderShipping.map(
+      (orderShipping) => orderShipping.shipping_method_id
+    )
+
+    await Promise.all([
+      this.orderAddressService_.delete(orderAddressIds, sharedContext),
+      // Delete order changes & actions
+      this.orderChangeService_.delete(orderChangeIds, sharedContext),
+    ])
```
olivermrbl: todo: use `promiseAll` utility

packages/modules/order/src/migrations/Migration20250522181137.ts
carlos-r-l-rodrigues: Yeah, it is unlike there will be any orphaned summary. This is just avoid any potential issue when creating the constraint.

fPolic: Yes, this is a cleanup of summaries of already deleted orders. Since cascade wasn't properly set up, they weren't deleted. We also cannot add the foreign key constraint if we don't remove summaries without their order first.

olivermrbl: Shouldn't we only query orders that are not deleted in the `IN(...)` clause?

fPolic: That would remove summaries of soft deleted order as well which we don't want in case soft deleted order is restored.

olivermrbl: Okay, I think I partly thought this was the intention for a second.

I can't see how summaries could exist with an order ID, that isn't in the DB, since we are not hard-deleting any entities, but I guess it's fine to include the script just in case.

PR #12193, opened 2025-04-15
fix(payment): properly delete refund

fPolic:
**What**
- properly delete the refund record if refunding with the provider fails

---

CLOSES SUP-1398


olivermrbl (approved):
Nice  catch

PR #12194, opened 2025-04-15
fix: apply additional data validator using a global middleware

### LINKED ISSUE #12109: [Bug]: additionalDataValidator does not work
### Package.json file

```JSON
"@medusajs/admin-sdk": "2.6.1",
"@medusajs/cli": "^2.6.1",
"@medusajs/framework": "2.6.1",
"@medusajs/medusa": "2.6.1",
```

### Node.js version

N/A

### Database and its version

N/A

N/A

### Browser name

_No response_

### What happended?

1. Followed the [guide](https://docs.medusajs.com/learn/fundamentals/api-routes/additional-data)
2. Send a request with any additional_data

Because the user defined middleware is executed after medusa defined middleware.

```
Route middlewares in the following order:
Route middleware defined in the Medusa's core.
Route middleware defined in the plugins (in the order the plugins are registered in).
Route middleware you define in the application.
```

### Expected behavior

request failed with validation error

### Actual behavior

request successfully

### Link to reproduction repo

N/A

thetutlage:
Fixes: FRMW-2941

This PR treats the `additionalDataValidator` as a first-class citizen within the request handling layer and does not convert them to a middleware behind the scenes. The treat is similar to how the `bodyParserConfig` is treated.

**Current problem**

Since middleware defined inside the user app runs after the core middleware, the `additionalDataValidator` middleware skips the validation layer and hence its schema has no impact on the request.

When running in debug mode you can see which routes have additional data validator

CleanShot 2025-04-15 at 15 54 05@2x

Closes: #12109


packages/core/framework/src/http/__fixtures__/routers-middleware/middlewares.ts
```diff
@@ -56,3 +69,6 @@ export default defineMiddlewares([
     middlewares: [raw({ type: "application/json" })],
   },
 ])
+
+console.log(middlewares.routes)
```
olivermrbl:
```suggestion
```

packages/core/framework/src/http/middleware-file-loader.ts
```diff
@@ -90,6 +99,24 @@ export class MiddlewareFileLoader {
           })
         }
 
+        if (
+          "additionalDataValidator" in route &&
```
olivermrbl: suggestion(non-blocking): use isPresent util

olivermrbl (commented):
LGTM

olivermrbl (approved):

adrien2p (approved):
LGTM, just the existing comments 💪

packages/core/framework/src/http/middleware-file-loader.ts
thetutlage: Doesn't narrow down the types 😔

CleanShot 2025-04-16 at 15 03 50@2x

packages/core/framework/src/http/__fixtures__/routers-middleware/middlewares.ts
thetutlage: Nice catch 😅

PR #12196, opened 2025-04-15
fix(): Properly handle workflow as step now that events are fixed entirely

adrien2p:
**What**
Now that all events management are fixed in the workflows life cycle, the run as step needs to leverage the workflow engine if present (which should always be the case for async workflows) in order to ensure the continuation and the ability to mark parent step in parent workflow as success or failure


adrien2p:
@sradevski now everything should be fixed on all account with the previous pr's + this one 💪

carlos-r-l-rodrigues (approved):

PR #12197, opened 2025-04-15
fix(medusa): Ensure specified sales channel is used for inventory

olivermrbl:
**What**
Solves inventory calculation for products in the case where:
- A publishable API key has multiple sales channels
- A sales channel is passed as a query param

Until now, if a publishable API key was associated with multiple sales channels, the inventory availability wrapper would fail with: 
```
Inventory availability cannot be calculated in the given context. Either provide a sales channel id or configure a single sales channel in the publishable key
```

This happens because:
1. we don't know which sales channel to pick when calculating the inventory
2. the sales channel from the query param was not used

### DIFF
--- a/integration-tests/http/__tests__/product/store/product.spec.ts
+++ b/integration-tests/http/__tests__/product/store/product.spec.ts
@@ -1478,7 +1478,7 @@ medusaIntegrationTestRunner({
       describe("with inventory items", () => {
         let location1
         let location2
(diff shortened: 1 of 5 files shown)


fPolic (approved):
LGTM

carlos-r-l-rodrigues (approved):

olivermrbl:
Closing in favor of #12234

PR #12198, opened 2025-04-15
fix(dashboard): properly register settings custom routes

fPolic:
**What**
- custom setting routes were registered under `/settings/settings/custom-route` instead of `/settings/custom-route`

---

CLOSES SUP-1384

### DIFF
--- a/.changeset/green-penguins-kiss.md
+++ b/.changeset/green-penguins-kiss.md
@@ -1,0 +1,5 @@
+---
+"@medusajs/dashboard": patch
+---
+
+fix(dashboard): properly register settings custom routes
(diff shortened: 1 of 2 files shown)


olivermrbl (approved):
I presume you tested this, because the change looks a bit funky. Probably just because I don't know how it works under the hood 😄

fPolic:
Yup, tested it with algoliasearch example. 
The issue was that the `settingsRoutes` shape that is passed:
`{path: "/settings", children: [{path: "/custom-route-1", ...}, {path: "/custom-route-2", ...}]}`. 
We were nesting this already inside the `/settings` path so the fix is to spread children instead of the root object.

PR #12224, opened 2025-04-17
feat: Add support for uploading a file directly to the file provider from the client

sradevski:
REFERENCES CMRC-306

This PR adds support for getting a signed upload URL that the client can then use to upload a file to the chosen file provider.

The changes have been tested with Digitalocean and AWS S3, but other S3-compatible providers should work similarly.


packages/modules/file/src/services/file-module-service.ts
```diff
@@ -49,6 +51,27 @@ export default class FileModuleService implements FileTypes.IFileModuleService {
     return Array.isArray(data) ? result : result[0]
   }
 
+  getUploadFileUrls(
```
sradevski, line 54: I just need to add a simple integration test to test the wiring. I'll do that tomorrow

packages/core/types/src/file/service.ts
```diff
@@ -41,6 +41,44 @@ export interface IFileModuleService extends IModuleService {
...
+   * @example
+   * const [uploadInfo] = await fileModuleService.getUploadFileUrls([{
+   *   filename: "product.png",
+   *   mimeType: "image/png",
+   * }])
+   */
+  getUploadFileUrls(
+    data: GetUploadFileUrlDTO[],
+    sharedContext?: Context
+  ): Promise<UploadFileUrlDTO[]>
```
adrien2p: todo: we might need to align the example, they don't seem to match the expected input

packages/modules/file/src/services/file-module-service.ts
```diff
@@ -49,6 +51,27 @@ export default class FileModuleService implements FileTypes.IFileModuleService {
...
+    sharedContext?: Context
+  ): Promise<UploadFileUrlDTO>
+
+  async getUploadFileUrls(
+    data: GetUploadFileUrlDTO[] | GetUploadFileUrlDTO
+  ): Promise<UploadFileUrlDTO[] | UploadFileUrlDTO> {
+    const input = Array.isArray(data) ? data : [data]
+
+    const result = await Promise.all(
+      input.map((file) => this.fileProviderService_.getPresignedUploadUrl(file))
```
adrien2p, line 69: suggestion: should we provide `getPresignedUploadUrls`?

packages/modules/file/src/services/file-provider-service.ts
```diff
@@ -48,4 +48,24 @@ export default class FileProviderService {
...
+        "Provider does not support presigned upload URLs"
+      )
+    }
+
+    if (!fileData.filename) {
+      throw new MedusaError(
+        MedusaError.Types.INVALID_DATA,
+        "File name is required to get a presigned upload URL"
+      )
+    }
```
adrien2p, line 67: Q: wouldnt this validation be handled by the provider directly instead?

packages/modules/providers/file-s3/src/services/s3-file.ts
```diff
@@ -175,4 +178,49 @@ export class S3FileService extends AbstractFileProviderService {
...
+        `No file data provided`
+      )
+    }
+
+    if (!fileData.filename) {
+      throw new MedusaError(
+        MedusaError.Types.INVALID_DATA,
+        `No filename provided`
+      )
+    }
```
adrien2p:
```suggestion
    if (!fileData?.filename) {
      throw new MedusaError(
        MedusaError.Types.INVALID_DATA,
        `No filename provided`
      )
    }
```

packages/modules/providers/file-s3/src/services/s3-file.ts
```diff
@@ -175,4 +178,49 @@ export class S3FileService extends AbstractFileProviderService {
...
+
+    const parsedFilename = path.parse(fileData.filename)
+    const fileKey = `${this.config_.prefix}${parsedFilename.name}${parsedFilename.ext}`
```
adrien2p:
```suggestion
    const fileKey = `${this.config_.prefix}${parsedFilename.base}`
```

adrien2p (approved):

packages/core/types/src/file/service.ts
sradevski: Ah left the content there, the rest should match :)

packages/modules/file/src/services/file-module-service.ts
sradevski: I just followed the same pattern as the `getPresignedDownloadUrl`, but since none of the providers provide this natively it's better to just handle the "concurrency" here

packages/modules/file/src/services/file-provider-service.ts
sradevski: It could be, but since it's a mandatory field for all providers it's good to have a top-level check. WDYT?

packages/modules/providers/file-s3/src/services/s3-file.ts
sradevski: OFC :D

packages/modules/providers/file-s3/src/services/s3-file.ts
sradevski: 👍 Initially I thought it's better to separate them in case we want to apply some suffix to the filename, but we can do it later on if necessary

sradevski:
Tested it in Cloud, working as expected :) Merging now

PR #12227, opened 2025-04-17
fix(core-flows, link-modules): return fulfillment creation

fPolic:
**What**
- fix algo for finding the return shipping method
- fix returns <> fulfillements link
- fix cancel claim flow
- fix cancel exchange flow

---

CLOSES SUP-1425


olivermrbl (approved):
LGTM, should we add a test?

fPolic:
@olivermrbl, added testing for this and fixed issues causing failed tests, should be ready for merge

packages/core/core-flows/src/order/workflows/claim/cancel-claim.ts
```diff
@@ -78,7 +81,7 @@ export const cancelClaimValidateOrderStep = createStep(
     const notCanceled = (o) => !o.canceled_at
 
     throwErrorIf(
-      orderClaim_.fulfillments,
+      orderReturn_.fulfillments,
```
olivermrbl, line 84: question: Don't we still need to ensure the claims fulfillments are canceled too? Or am I missing something?

fPolic: We don't have claim/exchange specific fulfillments i.e. we don't have links between those entities AFAIK. We have Return <> Fulfillment link for ReturnFulfillments otherwise it's an order fulillment. Maybe @carlos-r-l-rodrigues can confirm.

olivermrbl: Yeah, but what I mean is, if you create an exchange or claim, and you then create the fulfillment for those outbound items. Those fulfillments would need to be canceled as well before you can cancel the entire exchange or claim. 

Does that make sense?

fPolic: Yeah sorry misunderstud what part of the flow you are reffering to. Makes sense!

So we would need to ensure that any fulfillment on the order that has outbound items from the claim is canceled, right?

This is not handled at the moment. I would suggest to add this in a follow up to not block this fix if you agree?

olivermrbl: > So we would need to ensure that any fulfillment on the order that has outbound items from the claim is canceled, right?

Yes, exactly.

> This is not handled at the moment. I would suggest to add this in a follow up to not block this fix if you agree?

Sure, we can do that. So I guess what we already had in place wasn't working because the fulfillments weren't tied to the claims and exchanges? Is that correct?

trevster344: Any idea on a timeframe for this fix? Currently running into this issue in production.

PR #12234, opened 2025-04-18
fix(medusa): sales_channel_id middleware manipulation leading to lost  of the sc

adrien2p:
**What**
Sales channel manipulation through the middleware (especially the usage of maybe apply link filter) lead to the removal of the filterable key from which other middlewares might depends. Instead, make other middlewares always rely on the source of truth from validated query and clean up potential miss usage


olivermrbl:
Is this a duplicate of #12197? I am fine going with either

adrien2p:
> Is this a duplicate of #12197? I am fine going with either

Sorry I didnt know 😅, I'll close this one then.

olivermrbl:
All good, we can also go with this PR, just wanted to highlight, so we can decide 😄

adrien2p:
> All good, we can also go with this PR, just wanted to highlight, so we can decide 😄

Can I get you to have a check on this one still, I can grab what you have done and put it here. But the idea was that they are two different middleware and we should not rely on a previous middleware mutation (to make them independent and usable separately). Instead we rely on the validated query as the source of truth and I ve extracted the validation of the sc into a separate function. Also, not removing the filterable fields could lead to its usage in places where it might not be expected. let me know what you think and maybe we can have a mix of our two pr's 😄

adrien2p:
depending on what you think we can either pick up some thing from this one into your or vice versa I can get your integration tests into this one 💪 @olivermrbl

thetutlage (approved):

carlos-r-l-rodrigues (approved):

juanzgc:
Do we potentially want to also update `maybeApplyLinkFilter` to no longer delete `req.filterableField.sales_channel_id`? 
By doing so we would have `filterByValidSalesChannels()` handle the source of truth for which `sales_channel_id` should be used when handling inventory.
Let me know what you think @adrien2p / @olivermrbl

adrien2p:
> Do we potentially want to also update `maybeApplyLinkFilter` to no longer delete `req.filterableField.sales_channel_id`? 
> 
> By doing so we would have `filterByValidSalesChannels()` handle the source of truth for which `sales_channel_id` should be used when handling inventory.
> 
> Let me know what you think @adrien2p / @olivermrbl 

In my opinion, the source of truth remains validatedQuery from which filterableFields are derived (minus pagination and ordering). You could even argue that you could only use validated query to prepare your own filters if you wanted 🙃 and since middlewares can be applied in any order and any numbers I would say that it is better to apply validation and transformation independently from each other as I have tried to do here by sharing the transform and validate sales channel in both middlewares. Let me know wyt.

juanzgc:
Yes, I actually agree, lgtm! Thanks for the quick turnaround on this PR btw! 💪🏽

olivermrbl (approved):
LGTM

PR #12242, opened 2025-04-18
docs: Replace api_key_id with publishable_key_id

Wcdaren:
image

### DIFF
--- a/packages/core/modules-sdk/src/utils/linking-error.ts
+++ b/packages/core/modules-sdk/src/utils/linking-error.ts
@@ -13,7 +13,7 @@ type LinkingErrorMessageInput = {
 
 /**
  *
- * Example: Module to dismiss salesChannel and apiKey by keys sales_channel_id and publishable_key_id was not found. Ensure the link exists, keys are correct, and the link is passed in the correct order to method 'remoteLink.dismiss'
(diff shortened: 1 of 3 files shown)


shahednasser (commented):
Nice catch! Can you also make the change in https://docs.medusajs.com/resources/commerce-modules/sales-channel/links-to-other-modules#manage-with-link ?

Wcdaren:
Regarding the modifications to the document, I have added some, but I'm not sure whether the following section needs to be changed.
https://github.com/medusajs/medusa/blob/develop/packages/core/utils/src/link/links.ts#L79
image

packages/core/modules-sdk/src/utils/linking-error.ts
```diff
@@ -13,7 +13,7 @@ type LinkingErrorMessageInput = {
 
 /**
  *
- * Example: Module to dismiss salesChannel and apiKey by keys sales_channel_id and api_key_id was not found. Ensure the link exists, keys are correct, and the link is passed in the correct order to method 'remoteLink.dismiss'
+ * Example: Module to dismiss salesChannel and apiKey by keys sales_channel_id and publishable_key_id was not found. Ensure the link exists, keys are correct, and the link is passed in the correct order to method 'remoteLink.dismiss'
```
shahednasser: Revert this, as the original message was correct in this context.

shahednasser (changes requested):
Just a last change!

shahednasser (approved):
LGTM thanks!

PR #12251, opened 2025-04-21
fix: allow steps to have undefined or null in response

thetutlage:
Fixes: FRMW-2946


sradevski (approved):
👍

thetutlage:
Took another stab at it here. This approach was become complicated. https://github.com/medusajs/medusa/pull/12255

PR #12255, opened 2025-04-22
fix: steps to return undefined and still chain the config method

thetutlage:
Fixes: FRMW-2946

### DIFF
--- a/packages/core/workflows-sdk/src/utils/composer/type.ts
+++ b/packages/core/workflows-sdk/src/utils/composer/type.ts
@@ -50,6 +50,8 @@ type ConvertHooksToFunctions<THooks extends any[]> = THooks extends [
   ? ConvertHookToObject<A> & ConvertHooksToFunctions<R>
   : {}
 
+export type Void = { " $$type": "void" }
+
 /**
  * A step function to be used in a workflow.
  *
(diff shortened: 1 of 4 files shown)


packages/core/workflows-sdk/src/utils/composer/type.ts
```diff
@@ -62,7 +64,9 @@ export type StepFunction<
 > = (KeysOfUnion<TInput> extends []
   ? // Function that doesn't expect any input
     {
-      (): WorkflowData<TOutput> & StepFunctionReturnConfig<TOutput>
+      (): TOutput & {} extends never
```
sradevski, line 67: Is the `& {}` necessary? I am just not sure what purpose it has :D

thetutlage: The goal here is to check if merging `TOutput` (aka step output) with an object produces `never`, then we should treat that output differently. 

It will produce `never` when the return value is `undefined` or `null`. But not a union of `undefined + something else`.

sradevski: 👍  TIL

carlos-r-l-rodrigues (approved):

PR #12263, opened 2025-04-22
fix(core-flows): Refund and recreate payment session on cart complete failure

fPolic:
**What**
- if cart complete flow fails due don't cancel authorized session
- authorize and capture payment if we receive `payment.success` event in the webhook (and it is autocapture flow) instead of relying on the cart complete workflow do to that
  - in case cart complete fails before payment steps are reached, payment session inside `compensateIPaymentIfNeeded` will have stale status: `pending`


packages/core/core-flows/src/cart/steps/compensate-payment-if-needed.ts
```diff
@@ -51,20 +48,7 @@ export const compensatePaymentIfNeededStep = createStep(
...
-          `Error was thrown trying to cancel payment session - ${paymentSession.id} - ${e}`
-        )
-      }
-    }
-
-    if (
-      paymentSession.status === PaymentSessionStatus.CAPTURED &&
-      paymentSession.payment?.id
-    ) {
+    if (paymentSession.payment?.captured_at) {
```
fPolic, line 64: comment: should we add to docs that once payment is refunded in the autocapture flow if cart complete fails, a new session needs to be initialized

packages/core/core-flows/src/cart/steps/compensate-payment-if-needed.ts
```diff
@@ -51,20 +48,7 @@ export const compensatePaymentIfNeededStep = createStep(
       }
     )
 
-    if (paymentSession.status === PaymentSessionStatus.AUTHORIZED) {
-      try {
-        await paymentModule.cancelPayment(paymentSession.id)
```
olivermrbl, line 55: comment: This seems to have been in a broken state up until now, since `cancelPayment` requires a payment ID, not a session ID. If we revert to keep this, will you update this? 

I think we should be able to do `paymentSession.payment.id`

olivermrbl (approved):
LGTM

PR #12272, opened 2025-04-23
Update 6.3. Using TypeScript Aliases - Alias should run after build, not before

AlexanderLindkjaer:
https://docs.medusajs.com/learn/configurations/ts-aliases

`npm run resolve:aliases` should run after `medusa build` to change out the aliases in the build output. 

So instead of the current guide
`"build": "npm run resolve:aliases && medusa build"`

I should be 
`"build": "medusa build && npm run resolve:aliases"`


shahednasser (approved):
Nice catch thanks!

PR #12278, opened 2025-04-24
test: use jest shards and simplify test-cli-pipeline

thetutlage:
# PR Summary

- **CLI pipeline down from `12mins` to `4mins`**.
- **Medusa Pipeline (tests) down from `16mins` to `8mins`**.

- Updated `setup-node` action to its latest version (ie `v4`).
- Removed usage of `medusa-dev-cli`, since this package is barely maintained. Instead, we are using yarn resolutions to install source code from the current checked out codebase in the CI.
- Remove custom logic around creating chunks. Instead using `jest --shards`.
- Divided packages integration tests into two pipelines. One for `slow` and another for `fast`. This allows these tests to run in parallel.
- Removed `--runInBand` from many integration tests, since that makes test files run serially.
- Removed duplicate builds running before running every test chunk.


carlos-r-l-rodrigues (approved):
Amazing 🔥

olivermrbl (approved):
💪 nice!

PR #12284, opened 2025-04-24
fix(medusa): Fix store return API

riqwan:
what:

- fixes the middleware to point to the right API route
- adds a test to store return API


packages/medusa/src/api/store/return/middlewares.ts
```diff
@@ -1,15 +1,15 @@
...
 } from "@medusajs/framework"
+import { MiddlewareRoute } from "@medusajs/framework/http"
 import * as QueryConfig from "./query-config"
 import { ReturnsParams, StorePostReturnsReqSchema } from "./validators"
 
 export const storeReturnRoutesMiddlewares: MiddlewareRoute[] = [
   {
     method: ["POST"],
-    matcher: "/store/returns",
+    matcher: "/store/return",
```
fPolic: q: should the enpoint be `/store/returns`?

fPolic (approved):
LGTM

packages/medusa/src/api/store/return/middlewares.ts
riqwan: @olivermrbl shall we change the route and call it a breaking change?

olivermrbl: Would it be fair to say it wasn't functional in the previous state, and label it as a fix rather than a breaking change?

riqwan: thats true indeed!

riqwan: pushed! 🥂

olivermrbl (approved):

PR #12285, opened 2025-04-24
fix(dashboard,core-flows): create multiple transactions when multiple refunds occur

riqwan:
what:

When multiple refunds are currently performed on the same payment, we don't create additional transactions due to a validation that was in place. Removing it continues to add the transactions, but still fails when the refunded value is greater than captured value (which is expected) 

This also fixes an error that does not display the refundable amount when there is a paid payment collection.

### DIFF
--- a/packages/core/core-flows/src/order/steps/add-order-transaction.ts
+++ b/packages/core/core-flows/src/order/steps/add-order-transaction.ts
@@ -5,12 +5,16 @@ import { StepResponse, createStep } from "@medusajs/framework/workflows-sdk"
 /**
  * The transaction(s) to add to the order.
  */
-export type AddOrderTransactionStepInput = CreateOrderTransactionDTO | CreateOrderTransactionDTO[]
+export type AddOrderTransactionStepInput =
+  | CreateOrderTransactionDTO
+  | CreateOrderTransactionDTO[]
 
 /**
  * The added order transaction(s).
  */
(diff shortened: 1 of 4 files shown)


packages/core/core-flows/src/order/steps/add-order-transaction.ts
```diff
@@ -5,48 +5,34 @@ import { StepResponse, createStep } from "@medusajs/framework/workflows-sdk"
...
-          reference: trx.reference,
-          reference_id: trx.reference_id,
-        },
-        {
-          select: ["id"],
-        }
-      )
-
-      if (existing.length) {
-        return new StepResponse(null)
```
riqwan, line 42: Hey @carlos-r-l-rodrigues , can you think of any reason why this would skip if there is an existing transaction?

carlos-r-l-rodrigues: The reference_id here should not repeat. Every transaction ideally has it's own reference_id.
A different return_id, a different payment_id, or capture_id, anything.
but ideally not duplicated.

carlos-r-l-rodrigues: This is on purpose to avoid duplicated transactions

olivermrbl: @riqwan, @carlos-r-l-rodrigues did we get to a resolution on this?

PR #12297, opened 2025-04-25
feat(dashboard, js-sdk, medusa, tax, types): custom tax providers

fPolic:
**What**
- add support for custom tax providers in core
- list tax providers endpoint
- admin: choose tax provider on TaxRegion create
- admin: update tax provider for a Region
- add missing update tax region methods
- add update TaxRegion on admin

---
NOTE/TODO
- [x] we need to add a migration script since system provider is not default anymore to migrate existing regions to "system" provider


packages/medusa/src/api/admin/tax-providers/route.ts
```diff
@@ -0,0 +1,35 @@
...
+
+import { HttpTypes } from "@medusajs/framework/types"
+
+export const GET = async (
+  req: AuthenticatedMedusaRequest<HttpTypes.AdminGetTaxProvidersParams>,
+  res: MedusaResponse<HttpTypes.AdminTaxProviderListResponse>
+) => {
+  const remoteQuery = req.scope.resolve(ContainerRegistrationKeys.REMOTE_QUERY)
+
+  const { rows: tax_providers, metadata } = await remoteQuery(
```
olivermrbl: todo: Use `query.graph`, remote query will be deprecated over time

packages/medusa/src/migration-scripts/migrate-tax-region-provider.ts
```diff
@@ -0,0 +1,49 @@
...
+        }))
+    })
+
+    updateTaxRegionsStep(updateData)
+
+    return new WorkflowResponse(void 0)
+  }
+)
+
+export default async function assignProductsToShippingProfile({
```
olivermrbl: todo: assignTaxSystemProviderToTaxRegions
```suggestion
export default async function assignTaxSystemProviderToTaxRegions({
```

packages/medusa/src/migration-scripts/migrate-tax-region-provider.ts
```diff
@@ -0,0 +1,49 @@
...
+  container,
+}: ExecArgs) {
+  const logger = container.resolve(ContainerRegistrationKeys.LOGGER)
+
+  logger.info("Assigning products to shipping profile")
```
olivermrbl:
```suggestion
  logger.info("Assigning tax system provider to tax regions")
```

olivermrbl (approved):
Overall LGTM, nice one!

This provider title in the header doesn't look super nice though, could we format it differently? Also, it's not clear that this is the provider, so maybe it should also say that.

CleanShot 2025-05-04 at 13 13 31@2x

fPolic:
> This provider title in the header doesn't look super nice though, could we format it differently? Also, it's not clear that this is the provider, so maybe it should also say that.

Good point, added a separate provider section.

PR #12322, opened 2025-04-29
feat(tax, cart, core-flows, utils, types): Compound TaxRate (Tax on Tax) Feature

panalgin:
# Compound Tax Calculation Feature

## Overview
This PR introduces support for compound taxation, allowing taxes to be calculated on top of other taxes. This feature is essential for accurately modeling tax systems like Turkey's, where VAT (KDV) is calculated on a price that already includes Special Consumption Tax (ÖTV).

## Key Changes
- Added `is_compound` flag to tax rate models
- Implemented compound tax calculation logic in tax and cart totals calculation
- Updated admin UI to support creating and editing compound tax rates
- Added database migrations for the new fields
- Updated documentation to reflect new capabilities

## Technical Implementation
The implementation properly handles tax-inclusive and tax-exclusive pricing scenarios when compound taxes are present. Compound taxes are:
1. Applied after non-compound taxes
2. Calculated based on the price + all non-compound taxes
3. Properly sequenced in the calculation pipeline

## Use Cases
This feature supports various international tax systems that require compound calculations:

- **Turkey**: VAT calculated on price including SCT
- **Brazil**: ICMS (state VAT) calculated on price including IPI (Industrial Product Tax)
- **Philippines**: VAT calculated on price including excise taxes
- **Russia**: VAT calculated on price including excise duties
- **India**: Additional cess tax on top of GST for luxury goods
- **Mexico**: VAT calculated on price including IEPS (Special Tax)

## Example
For a product costing $100 with 10% SCT and 20% VAT (as compound):
- SCT = $10 (calculated on base price)
- VAT = $22 (calculated on price + SCT = $110)
- Total tax = $32
- Final price = $132

Without the compound flag, VAT would incorrectly be $20 (on base price), leading to a total tax of $30 and final price of $130.

## Migration Notes
This PR includes necessary migrations for database schema updates in both tax and cart modules.

### DIFF
--- a/packages/core/utils/src/totals/tax/index.ts
+++ b/packages/core/utils/src/totals/tax/index.ts
@@ -1,4 +1,4 @@
-import { BigNumberInput, TaxLineDTO } from "@medusajs/types"
+import { BigNumberInput, ItemTaxLineDTO } from "@medusajs/types"
 import { BigNumber } from "../big-number"
 import { MathBN } from "../math"
 
@@ -7,7 +7,7 @@ export function calculateTaxTotal({
   taxableAmount,
   setTotalField,
 }: {
-  taxLines: Pick<TaxLineDTO, "rate">[]
+  taxLines: any[]
   taxableAmount: BigNumberInput
   setTotalField?: string
 }) {
@@ -16,10 +16,30 @@ export function calculateTaxTotal({
   }
 
   let taxTotal = MathBN.convert(0)
+  let baseWithTax = MathBN.convert(taxableAmount)
 
-  for (const taxLine of taxLines) {
+  // First process all non-compound taxes (like SCT)
+  const standardTaxLines = taxLines.filter(taxLine => !taxLine.is_compound);
+  const compoundTaxLines = taxLines.filter(taxLine => taxLine.is_compound);
+
+  // Calculate regular taxes
+  for (const taxLine of standardTaxLines) {
     const rate = MathBN.div(taxLine.rate, 100)
     let taxAmount = MathBN.mult(taxableAmount, rate)
+
+    if (setTotalField) {
+      ;(taxLine as any)[setTotalField] = new BigNumber(taxAmount)
+    }
+
+    taxTotal = MathBN.add(taxTotal, taxAmount)
+    baseWithTax = MathBN.add(baseWithTax, taxAmount) // Add tax to base for compound calculation
+  }
+
+  // Calculate compound taxes (like VAT on SCT)
+  for (const taxLine of compoundTaxLines) {
+    const rate = MathBN.div(taxLine.rate, 100)
+    // Apply to price INCLUDING other taxes
+    let taxAmount = MathBN.mult(baseWithTax, rate)
 
     if (setTotalField) {
       ;(taxLine as any)[setTotalField] = new BigNumber(taxAmount)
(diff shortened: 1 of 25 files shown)


packages/core/utils/src/totals/tax/index.ts
```diff
@@ -7,7 +7,7 @@ export function calculateTaxTotal({
   taxableAmount,
   setTotalField,
 }: {
-  taxLines: Pick<TaxLineDTO, "rate">[]
+  taxLines: any[]
```
adrien2p, line 10: suggestion: we could probably type it here

packages/modules/cart/src/migrations/Migration20250411131500.ts
```diff
@@ -0,0 +1,15 @@
+import { Migration } from "@mikro-orm/migrations"
+
+export class Migration20250326151602 extends Migration {
```
adrien2p, line 3: todo: could you run `migration:create` instead of writing the migration by hand 😃 here it misses the snapshot update

packages/modules/tax/README.md
```diff
@@ -1 +1,95 @@
 # Tax Module
+
```
adrien2p, line 2: todo: this can be removed from that readme cc @olivermrbl

packages/modules/tax/src/migrations/Migration20250411000002.ts
```diff
@@ -0,0 +1,17 @@
+import { Migration } from "@mikro-orm/migrations"
+
+export class Migration20240800000002 extends Migration {
```
adrien2p, line 3: todo: same as previous

packages/modules/tax/src/providers/system.ts
```diff
@@ -12,30 +12,50 @@ export default class SystemTaxService implements ITaxProvider {
     shippingLines: TaxTypes.ShippingTaxCalculationLine[],
     _: TaxTypes.TaxCalculationContext
   ): Promise<(TaxTypes.ItemTaxLineDTO | TaxTypes.ShippingTaxLineDTO)[]> {
```
adrien2p, line 14: suggestion: let me know what you think

```ts
async getTaxLines(
    itemLines: TaxTypes.ItemTaxCalculationLine[],
    shippingLines: TaxTypes.ShippingTaxCalculationLine[],
    _: TaxTypes.TaxCalculationContext
  ): Promise<(TaxTypes.ItemTaxLineDTO | TaxTypes.ShippingTaxLineDTO)[]> {
    const taxLines: (TaxTypes.ItemTaxLineDTO | TaxTypes.ShippingTaxLineDTO)[] = []

    const itemAndShippingLines = [...itemLines, ...shippingLines]

    for (const line of itemAndShippingLines) {
      line.rates.map(rate => {
        const taxLine = {
          rate_id: rate.id,
          rate: rate.rate || 0,
          name: rate.name,
          code: rate.code,
          line_item_id: (line as TaxTypes.ItemTaxCalculationLine).line_item?.id,
          shipping_line_id: (line as TaxTypes.ShippingTaxCalculationLine).shipping_line?.id,
          provider_id: this.getIdentifier(),
          is_compound: rate.is_compound || false,
        }

        if (rate.is_compound) {   
          taxLines.push(taxLine)
        } else {
          taxLines.unshift(taxLine)
        }
      })
    }

    return taxLines
  }
```

srindom:
This is nice work - we will need some tests of the behavior though.

adrien2p:
yeah some integration tests will be required for the pr

packages/modules/tax/README.md
olivermrbl: Agree, but this is great for docs cc @shahednasser 


Edit: I don't mind keeping it in though. Don't think it hurts

adrien2p: In that case we would need to add the other features as well no? Cause it will feel that the rest might not be supported since only on feature is described here?

PR #12324, opened 2025-04-29
feat(fulfillment): support postal_expression filtering for geo zone matching

Saryazdi-Saman:
🚀 Adds postal code-based zone matching for improved fulfillment filtering.

## Summary

This PR introduces support for filtering geo zones based on the `postal_expression` field when listing shipping options for a given address. It enables both exact and prefix-based postal code matching to improve fulfillment granularity.

Previously, geo zone matching only supported country, province, and city fields. There was no way to match based on full or partial postal codes.
This made it difficult to define fine-grained service areas or neighbourhoods using real-world postal code patterns.

With this change:
- Exact postal code matches (`matches` field) and postal prefix matches (`starts_with` field) are now supported.
- Shipping options are matched only if the service zone fully contains the provided address. (e.g., a shippingOption serving Toronto/Canada won't be matched with an address in Montreal/Canada just because they share the same Country)
- Improves flexible, efficient postal code-based fulfillment territory management.

## Changes Introduced

- Added `PostalExpression` and `PostalExpressionFilterableProps` interfaces for structured postal filtering.
- Extended `buildGeoZoneConstraintsFromAddress` to generate dynamic `$or` conditions for postal expressions.
- Injected the `type` field (e.g., `zip`, `city`) into constraints to improve matching precision.
- Added GIN index on `geo_zone.postal_expression` (JSONB) for optimized lookup.
- Added integration tests validating:
  - Exact postal code matching
  - Prefix postal code matching
  - Non-matching behaviour

## Testing

- Added integration tests for exact and prefix postal code scenarios
- Verified service zone matching works correctly for both direct and partial postal codes
- Confirmed that shipping options which don't fully cover an area are excluded from results (This scenario would have failed prior to this PR)
- All tests pass locally and reflect expected behaviour

## Notes

- Assumes PostgreSQL is being used, as it leverages JSONB operators and GIN indexing


Saryazdi-Saman:
Thanks for reviewing! Happy to adjust based on feedback.

adrien2p (approved):
Excellent, that looks good to me but I ll wait for @olivermrbl opinion as well

packages/core/types/src/fulfillment/common/geo-zone.ts
```diff
@@ -90,5 +121,49 @@ export interface FilterableGeoZoneProps
...
+//   /**
+//    * Matches JSONB objects that contain all the specified keys.
+//    */
+//   $hasKeys?: (keyof T)[]
+
+//   /**
+//    * Matches JSONB objects that contain at least one of the specified keys.
+//    */
+//   $hasSomeKeys?: (keyof T)[]
+// }
```
olivermrbl: question: can we remove this, or is there a reason to keep it here?

olivermrbl (approved):
LGTM, appreciate the contribution.

packages/modules/fulfillment/src/services/fulfillment-module-service.ts
```diff
@@ -2419,17 +2419,71 @@ export default class FulfillmentModuleService
...
+    return [
+      { postal_expression: { matches: { $contains: [postalCode] } } },
+      ...Array.from(
+        { length: Math.max(0, Math.min(postalCode.length - 1, 4)) },
+        (_, i) => ({
+          postal_expression: { starts_with: { $contains: [postalCode.slice(0, i + 2)] } }
+        })
+      ),
+    ]
+  }
```
adrien2p:
```suggestion
protected static buildPostalExpressionFilters(postalCode: string) {
  const postalExpressions = [];

  postalExpressions.push({
    postal_expression: {
      matches: { $contains: [postalCode] }
    }
  });

  // Add partial match expressions for first 2-5 characters
  // We start with at least 2 characters and add more if available
  const minPartialLength = 2;
  const maxPartialLength = Math.max(0, Math.min(postalCode.length - 1, 5));
  
  for (let length = minPartialLength; length <= maxPartialLength; length++) {
    const partialCode = postalCode.slice(0, length);
    postalExpressions.push({
      postal_expression: {
        starts_with: { $contains: [partialCode] }
      }
    });
  }

  return postalExpressions;
}
```

Suggestion to make it more human friendly and simplify some computation that can be controlled differently and make it easier to evolve for potential future updates 👍 let me know what you think

packages/core/types/src/fulfillment/common/geo-zone.ts
Saryazdi-Saman: Good catch — this was leftover from a previous iteration. Cleaning it up now 🙌

packages/modules/fulfillment/src/services/fulfillment-module-service.ts
Saryazdi-Saman: Great suggestion — I agree this is more readable and easier to maintain. I’ve pushed the update 🙌

Saryazdi-Saman:
I noticed that the last commit had caused a TypeScript build error (due to a function name mismatch and a missing type on the postalExpressions array).

I’ve fixed both issues in the latest commit — build now passes locally and in CI ✅

Saryazdi-Saman:
Hey @adrien2p @olivermrbl — just looping you in since I pushed a quick fix last night to resolve a build issue. CI is all green now, let me know if it needs another approval. Thanks!

Saryazdi-Saman:
Hey team! Just checking in  — I saw this was approved (thank you!), and I wanted to see if there’s anything else you’d like me to update or if it’s just waiting for the right time to merge.

I’m also about to start working on the related admin UI for service zone management. 
Would you prefer that I add the UI changes to this same PR, or create a separate one building on top of this branch? Happy to do whichever makes more sense for your workflow.

PR #12325, opened 2025-04-30
feat: add needed methods to the file module and providers

thetutlage:
Fixes: FRMW-2956

Once we implement direct file uploads, we will need the APIs to read the uploaded files as a Buffer or a Stream to process its content. This PR introduces those methods


packages/core/types/src/file/provider.ts
```diff
@@ -144,4 +145,14 @@ export interface IFileProvider {
   getPresignedUploadUrl?(
     fileData: ProviderGetPresignedUploadUrlDTO
   ): Promise<ProviderFileResultDTO>
+
+  /**
+   * Get the file contents as a readable stream.
+   */
+  getAsStream(fileData: ProviderGetFileDTO): Promise<Readable>
```
sradevski, line 152: What's the use-case for stream methods for reading? SInce `getPresignedUploadUrl` gives you a link, you can then load the content as a stream or not, the caller is in full control of it, so they seem unnecessary.

Maybe I am missing something though

adrien2p: for programatic ingestion from stream, for example parsing a file incrementally with async iterators over stream and import. It seems weird from a programmatic point of view to get a presigned URL and then load it from there no?

adrien2p (approved):

packages/core/types/src/file/provider.ts
sradevski: It's not that weird, and it's a bit more versatile as the response is serializable. I am fine with adding it btw, just want to understand if we already have a use-case for it.

Also, do we need both as stream and as buffer? Wouldn't as stream suffice? The caller then can turn it in a buffer if they wanted to. Again, just thinking if we can keep the file module a bit leaner in case we don't have a use-case for it yet, but not a blocker for sure

thetutlage: @sradevski Yup missing a lot of context 😅

The signed URL is to upload the file and not to read it. This is how the flow looks like:

- The client (which is the browser) gets the signed URL to upload the file.
- The client triggers the file upload and sends it to AWS S3.
- Now the client has just the filename which was uploaded to S3.
- Now the client sends this filename to `/products/import` to process the file for imports. How would the implementation of `/products/import` will read the file now from its name? This is what these methods are meant to solve.

sradevski: `getPresignedDownloadUrl` is for downloading, so in your flow in the last step you would call
```
const file = fileService.retrieveFile(theId)
await fetch(input.file.url).then((response) =>
  fs.writeFile(filename, stream.Readable.fromWeb(response.body))
)

```
(In this example we just write the s3 file to a local file, but you can do anything with the stream at that point)

sradevski: I just noticed that I pasted the wrong method above, apologies for that :D

thetutlage: > It's not that weird, and it's a bit more versatile as the response is serializable

Which response is serializable?

sradevski: The `getPresignedDownloadUrl` as you get a URL to a file that you can then stream as in the example above

packages/core/types/src/file/provider.ts
```diff
@@ -144,4 +145,14 @@ export interface IFileProvider {
...
+
+  /**
+   * Get the file contents as a Node.js Buffer
+   */
+  getAsBuffer(fileData: ProviderGetFileDTO): Promise<Buffer>
```
olivermrbl, line 157: thought(non-blocking): Do we have a use case for this right now, otherwise I would leave it out?

packages/core/types/src/file/provider.ts
thetutlage: Why should I first generate a URL and then read it as a stream myself using fetch, if the same can be encapsulated using the existing API's?

- `s3`: Use the SDK to fetch the file as a stream. Why create a URL, which includes semi-slow operations like generating a signature.
- `local`: Directly read it from the disk. Why create a URL, then hit an endpoint and then read the file from there?

Infact, this approach seems the opposite of keeping things lean, as we are giving up on the providers ability to perform an operation and instead self implementing those operations in other parts of the codebase.

olivermrbl, line 152: very-nit: Can we call this `getDownloadStream`? 

Two reasons here:
- Aligns the implementation with what we had in v1, so it'll be slightly more intuitive to those coming from the previous major
- Makes the action explicit from the call signature

packages/core/types/src/file/provider.ts
thetutlage: Right now we need the Buffer method for the `products/import` endpoint as it deals with the buffer only. 

We can convert a readable stream to a Buffer in that endpoint, but that will require having an additional util to make sure we cleanup the readable stream properly and handle errors. Whereas, with the `fs` module and `s3` SDK we get it for free.

packages/core/types/src/file/provider.ts
thetutlage: Actually, it's just a stream that you might never use to trigger a download. Like read the stream and process each chunk to perform an import (no download happens)

olivermrbl: But wouldn't this always get (download) the stream from a remote location, such as with S3?

thetutlage: Yup. Makes sense!

PR #12328, opened 2025-04-30
feat: implement direct upload

thetutlage:
Fixes: FRMW-2953, FRMW-2955, FRMW-2954

This PR introduces a new endpoint called `/admin/direct_upload` which takes the details of a file selected by the user in the browser and issues a signed URL to upload this file with either S3 or the local file provider (depending upon the configured provider).

Once the signed URL is generated, the client will upload the file with the signed URL. In this request, we cannot send additional headers like cookies or authorization header, because the remote services like S3 will give a CORS error. 

Finally, I have changed the implementation of the `/admin/products/import` endpoint to accept the uploaded file unique name, read it as a buffer and process the file.


packages/medusa/src/api/admin/direct-upload/route.ts
```diff
@@ -0,0 +1,35 @@
...
+  AuthenticatedMedusaRequest,
+} from "@medusajs/framework/http"
+import type { HttpTypes } from "@medusajs/framework/types"
+import type { AdminDirectUploadRequestPayloadType } from "./validators"
+
+export const POST = async (
+  req: AuthenticatedMedusaRequest<AdminDirectUploadRequestPayloadType>,
+  res: MedusaResponse<HttpTypes.AdminDirectUploadSignatureResponse>
+) => {
+  const fileProvider = req.scope.resolve("file")
```
sradevski: NIT: Use the Modules.FILE variable here

packages/medusa/src/api/admin/direct-upload/route.ts
```diff
@@ -0,0 +1,35 @@
...
+  const { filename: originalname, mimeType, size } = req.body
+
+  const type = new MIMEType(mimeType)
+  const extension = type.subtype
+  const uniqueFilename = `${ulid()}.${extension}`
+
+  const response = await fileProvider.getUploadFileUrls({
+    filename: uniqueFilename,
+    mimeType: mimeType,
+    access: "private",
```
sradevski: Would this depend on what is being uploaded? Eg. for images you'd want them public

packages/medusa/src/api/admin/direct-upload/route.ts
```diff
@@ -0,0 +1,35 @@
...
+export const POST = async (
+  req: AuthenticatedMedusaRequest<AdminDirectUploadRequestPayloadType>,
+  res: MedusaResponse<HttpTypes.AdminDirectUploadSignatureResponse>
+) => {
+  const fileProvider = req.scope.resolve("file")
+  const { filename: originalname, mimeType, size } = req.body
+
+  const type = new MIMEType(mimeType)
+  const extension = type.subtype
+  const uniqueFilename = `${ulid()}.${extension}`
```
sradevski: I think we should allow passing the filename from the UI (and default to ulid) so they can upload files in nested folders (eg. `/media/videos/<ulid>`)

packages/medusa/src/api/admin/products/import/route.ts
```diff
@@ -2,27 +2,21 @@ import {
...
-    )
-  }
+  const fileProvider = req.scope.resolve("file")
+  const file = await fileProvider.getAsBuffer(req.body.filename)
 
   const { result, transaction } = await importProductsWorkflow(req.scope).run({
     input: {
-      filename: input.originalname,
-      fileContent: input.buffer.toString("utf-8"),
+      filename: req.body.originalname,
```
sradevski: `validatedbody`?

packages/medusa/src/api/admin/products/import/route.ts
```diff
@@ -2,27 +2,21 @@ import {
...
 ) => {
-  const input = req.file as Express.Multer.File
-
-  if (!input) {
-    throw new MedusaError(
-      MedusaError.Types.INVALID_DATA,
-      "No file was uploaded for importing"
-    )
-  }
+  const fileProvider = req.scope.resolve("file")
```
sradevski: similar, use Modules.FILE

packages/medusa/src/api/admin/products/import/route.ts
```diff
@@ -2,27 +2,21 @@ import {
...
+  const file = await fileProvider.getAsBuffer(req.body.filename)
```
sradevski: NIT: Maybe it's better to call it `fileId` or `fileKey` so it's clear that it is not a user-friendly name but the identifier in storage

packages/medusa/src/api/admin/products/middlewares.ts
```diff
@@ -102,7 +97,7 @@ export const adminProductRoutesMiddlewares: MiddlewareRoute[] = [
   {
     method: ["POST"],
     matcher: "/admin/products/import",
-    middlewares: [upload.single("file")],
```
sradevski, line 105: Should this PR also change the import products UI? Otherwise I guess it will be broken until we do it

packages/medusa/src/api/admin/direct-upload/route.ts
thetutlage: Yeah, it depends how we want to categorize them. Should we treat all `images` as public and everything else as private?

packages/medusa/src/api/admin/direct-upload/route.ts
thetutlage: Is there any specific reason that frontend would want to do that? Because, we do not display a gallery of files anywhere, so the hierarchy might not make much difference anyways.

packages/medusa/src/api/admin/products/import/route.ts
thetutlage: Yup. Doing it here https://github.com/medusajs/medusa/pull/12328/files#diff-0d02e04e3c70c173d2c2ad5ca4e0133d7d9377243b624b68a058d3f1c7929a35R345-R351

packages/medusa/src/api/admin/products/import/route.ts
thetutlage: Yeah, the `fileKey` makes more sense.

packages/medusa/src/api/admin/products/middlewares.ts
thetutlage: The imports UI works fine. However, I have updated the `js-sdk` to now make 3 calls instead of one, when a new file is uploaded.

packages/medusa/src/api/admin/products/validators.ts
```diff
@@ -341,3 +341,12 @@ export type AdminBatchVariantInventoryItemsType = BatchMethodRequest<
   AdminBatchUpdateVariantInventoryItemType,
   AdminBatchDeleteVariantInventoryItemType
 >
+
+export const AdminImportProducts = z.object({
+  filename: z.string(),
+  originalname: z.string(),
+  extension: z.string(),
+  size: z.number(),
+  mimeType: z.string(),
```
olivermrbl: todo: Should be snake_cased

packages/medusa/src/api/admin/products/import/route.ts
olivermrbl: todo: Use `validatedBody`

packages/medusa/src/api/admin/direct-upload/route.ts
sradevski: Maybe the caller should just pass if they want the upload to be private or public? But yes, all images are typically public

packages/medusa/src/api/admin/direct-upload/route.ts
sradevski: It's not needed for our current functionalities, it was more if someone wants to reuse the endpoint for some custom content, and they don't want everything flat in the bucket. Eg. in Cloud we structure dumps, some image content, etc. in different folders.

It was just a suggestion though, we can always add it later on if needed

srindom:
thought: think the API path is inconsistent with the general conventions. Since the resource that is being created here is a URL we could maybe do something like: `POST /admin/upload-urls` or `POST /admin/uploads/presigned-urls`.

thetutlage:
@srindom Sure. I wasn't sure about the naming convention anyways 😅.

packages/medusa/src/api/admin/direct-upload/route.ts
carlos-r-l-rodrigues: +1 for a param, and default it to private if not present.

packages/core/types/src/http/file/admin/responses.ts
```diff
@@ -16,3 +16,36 @@ export interface AdminFileListResponse {
 }
 
 export type AdminFileDeleteResponse = DeleteResponse<"file">
+
+export interface AdminUploadPreSignedUrlResponse {
```
olivermrbl, line 20: todo: Assuming this is the HTTP response, this needs to be snake_case to follow our convention

packages/medusa/src/api/admin/uploads/presigned-urls/route.ts
```diff
@@ -0,0 +1,34 @@
...
+  const extension = type.subtype
+  const uniqueFilename = `${ulid()}.${extension}`
+
+  const response = await fileProvider.getUploadFileUrls({
+    filename: uniqueFilename,
+    mimeType: req.validatedBody.mime_type,
+    access: req.validatedBody.access ?? "private",
+  })
+
+  res.json({
```
olivermrbl, line 41: See other comment. Needs to be snake_cased

packages/core/js-sdk/src/admin/product.ts
```diff
@@ -58,17 +58,93 @@ export class Product {
...
+   *
+   * @example
+   * sdk.admin.product.directImport({
+   *   file // uploaded File instance
+   * })
+   * .then(({ transaction_id }) => {
+   *   console.log(transaction_id)
+   * })
+   */
+  async directImport(
```
olivermrbl: thought: I am not loving the naming here, but I also don't have any great suggestions for alternatives 😅 

Maybe we can do `createImport`?

olivermrbl (approved):
Overall LGTM, minor comments.

packages/core/js-sdk/src/admin/product.ts
thetutlage: Yeah, same. Struggled with naming here a bit. I think we can go with `createImport`, since there is another method called `confirmImport`, so they both feel a bit in sync

olivermrbl (approved):

adrien2p (approved):

packages/medusa/src/api/admin/products/imports/route.ts
```diff
@@ -0,0 +1,27 @@
+import {
```
olivermrbl, line 1: comment: I imagine this will look different when we have the new product import workflow right? I am specifically referring to the part about getting the file buffer vs. stream.

PR #12333, opened 2025-04-30
fix: batch updates in upsertMany_

peterlgh7:
Batching updates in `mikro-orm-repository.ts` to bring down the duration of the `product/batch` endpoint. I kept changes as minimal as possible, but I have one question: Can we not use mikro-orm's own `_upsertMany` instead of our own version?


carlos-r-l-rodrigues (approved):
Nice 🔥 
LGTM

PR #12338, opened 2025-05-01
fix(core-flows,utils): move fulfillment workflow events

shahednasser:
Move fulfillment workflow events to be with other workflow events.

Could be considered a breaking change for users using the previously `FulfillmentEvents` variable


thetutlage (approved):
LGTM!

PR #12345, opened 2025-05-02
fix(medusa): Support additional data on admin/collections

olivermrbl:


packages/medusa/src/api/admin/collections/[id]/route.ts
```diff
@@ -26,7 +26,7 @@ export const GET = async (
 }
 
 export const POST = async (
-  req: AuthenticatedMedusaRequest<AdminUpdateCollectionType>,
+  req: AuthenticatedMedusaRequest<AdminUpdateCollectionType & AdditionalData>,
```
Comment (adrien2p) [inline packages/medusa/src/api/admin/collections/[id]/route.ts:29]: thought: Would it make sense to have `WithAdditionalData` and be able to do `AuthenticatedMedusaRequest<WithAdditionalData<YourOriginalType>>` That way you don't have to manage the intersection manually and additional data can be managed by us?

adrien2p (approved):

shahednasser (changes requested):
Shouldn't you add the additional data to the route validator?

olivermrbl:
> Shouldn't you add the additional data to the route validator?

Nice catch, that is easy to forget 😄

shahednasser (approved):
LGTM

PR #12346, opened 2025-05-02
fix: Added missing admin dashboard dependencies

pepijn-vanvlaanderen:
When working on a custom admin dashboard, I noticed 2 depencies that are directly used were not installed. This PR resolves that issue.


olivermrbl (approved):
Nice catch :)

PR #12347, opened 2025-05-02
fix(dashboard): support creating filters in new DataTable

### LINKED ISSUE #12068: [Bug]: Filter button in new DataTable does nothing
### Package.json file

```JSON
clean 2.6.1 starter
```

### Node.js version

v22.13.1

### Database and its version

PostgreSQL 17.2

W11

### Browser name

Firefox and Edge

### What happended?

The filter button on the new `DataTable` component does not work, see video. This affects all medusa pages that use it like customer groups, sales channels, api keys etc and also custom UI(for which the new DataTable is recommended on the docs). Tested on both firefox and chromium(edge)

I don't know if this is already in the backlog but i couldnt find any open issue about it so i wanted to document it.

https://github.com/user-attachments/assets/e3fcd1cf-ecf1-4327-9106-575748cdeff4

While writing i also found that the sorting button for the tax regions has undefined fields, see below. Not related to DataTable but kinda on topic.

https://github.com/user-attachments/assets/7c84e3db-f737-4548-9c43-e3a4afb90e46

### Expected behavior

filters work on new DataTables, sorting in Tax Regions page shows correct labels

### Actual behavior

filters do nothing, sorting labels in tax regions page are undefined

### Link to reproduction repo

2.6.1 starter

leobenzol:
**What**

- adds support for creating new filters in the new DataTable component, which is currently disabled/not working
- adds filter parameters to the query for UserListTable, which were simply missing

**How**

I introduced a new state `newFilter` in DataTable that holds the id of the current new filter. If present, it gets added to the `filtering` state of the DataTable with value undefined(which is what DataTableFilterBar expects).
The `handleFilteringChange` function manages the new state which is preemptively reset at every filter change. This handles the case when the value has been selected or the modal has been dismissed. 

**Why**
Currently all (filter) params are directly added and read from the url query params. But since FilterMenu and FilterBar respectively write and read a filter with undefined value to signal a new one, the new filter can't be added to the url query, both for syntactic and semantic reasons: 
1. undefined is not a valid JSON token and would require extra checks and parsing in many places 
2. whatever token might be chosen to represent the new filter state("undefined", "null", "_newFilterDontUse" might still be used by a user for whatever reason and disallowing a valid value just to implement the UI more easily is clearly to be avoided.

Closes #12068


packages/admin/dashboard/src/components/data-table/data-table.tsx
```diff
@@ -175,10 +182,14 @@ export const DataTable = <TData,>({
...
       Object.entries(value).forEach(([key, filter]) => {
         if (
-          prefixedFilterIds.includes(getQueryParamKey(key, prefix)) &&
-          filter
+          prefixedFilterIds.includes(getQueryParamKey(key, prefix))
         ) {
-          prev.set(getQueryParamKey(key, prefix), JSON.stringify(filter))
+          const prefixedKey = getQueryParamKey(key, prefix)
+          if (filter) {
+            prev.set(prefixedKey, JSON.stringify(filter))
```
anthid, line 189: Here, `JSON.stringify` will add quotes to strings (status="draft") resulting in a 400 - Bad request.

leobenzol: I couldn't reproduce that on any existing DataTable in the dashboard, since none of them has a "status" filter.
Could you give me a more complete example

leobenzol:
While troubleshooting I found a problem when adding filters with prefixes(which are used when the table is a nested component, so anywhere but the list pages), which is weird cause I swear remembering I checked that when making the PR. I'll look into it later

packages/admin/dashboard/src/components/data-table/data-table.tsx
anthid: Sorry for not adding more context. I had identified the issue in an admin customization I'm working on. Right now medusa only uses the new DataTable component in a couple of routes, and only with date filters. However you can try (directly typing the url) with `products?status=published` and `products?status="published"` to see how it behaves with and without quotes.

As _DataTable is deprecated in favor of the new DataTable, I understand that the new DataTable will eventually replace the old one in all routes.

leobenzol: I see. Well the sales channel table has a is_disabled="true"/"false" in addition to date filters, which works fine, so other string properties should too. The old data table must do URL parsing differently, not a problem here

anthid: Indeed. I missed a `JSON.parse(status)` in my searchParams definition. Adding it solves the issue.

willbouch:
hi @SteelRazor47 and thanks for the contribution. It looks to me like you are fixing the bug linked and also adding a new functionality correct? If that's the case, do you mind simply fixing the bug for now and then opening another PR to support the feature you are trying to add? Or maybe I am missing something and you can let me know

leobenzol:
@willbouch do you mean the created_at/updated_at fields? there is already a useDataTableDateFilters defined for the user table, which needs them. The filter wasn't working before so it wasn't noticed. Afaik all other tables have them already.

I think it belongs in this PR since it's needed to make filters work in the users table, but I can make it a separate one no problem, let me know.

P. S I'll wait until srindom is done with the data table changes to fix the conflicts, unless they end up fixing it as well.

willbouch:
> @willbouch do you mean the created_at/updated_at fields? there is already a useDataTableDateFilters defined for the user table, which needs them. The filter wasn't working before so it wasn't noticed. Afaik all other tables have them already.
> 
> I think it belongs in this PR since it's needed to make filters work in the users table, but I can make it a separate one no problem, let me know.
> 
> P. S I'll wait until srindom is done with the data table changes to fix the conflicts, unless they end up fixing it as well.

Yes! I mean, my goal here is to get the error fixed asap. So, my question is "does adding the created_at/updated_at fields fix the error that was flagged in the issue linked? If yes by itself, then let's merge this first, otherwise, keep the PR the way it is

leobenzol:
@willbouch I tried resolving the conflict but the changes were significant. Is it possible that *this* filter button will be removed? I see that the new View API has changed the Filter Bar which is now always present and also has a filter and sorting button(which both work perfectly). It doesn't make sense to have the same button twice.  
Let me know if for some reason it's still needed and I'll try and work it out, otherwise you can close this :)

Screenshot 2025-09-09 alle 00 09 26

willbouch:
@srindom I believe you would be better answering that question. My guess would be that yes, but could take a look at @SteelRazor47's last comment?

srindom:
heyo - the plan with the new data table is that the filter button next to the search bar will move down to this new "filter bar".

i bumped into similar issues with respect to the filter behavior when a filter is selected but hasn't yet been assigned a value. it should be fixed in the new version of the data table (as long as the filters from the URL are being parsed correctly).

willbouch:
So I guess we could close this PR then @SteelRazor47 ?

leobenzol:
@willbouch yeah the buttons in the new filter bar work perfectly so I'll close this 👍 . Could I ask you to have someone take a look at #13360 sometime before next release instead? Normally I'd let the team get to it when they have the time without bothering yall, but it's a one line change and currently draft orders are unusable in production for anyone that deploys the admin on a separate domain.

PR #12353, opened 2025-05-03
chore: Minor restructure of cart completion

olivermrbl:
The payment webhook is due for a larger revamp; I've kept the changes in this PR specific to the cart-completion to not touch too many things at once.

This PR primarily does two things:
- Removes idempotency from the cart-completion workflow, see explanation in the comment below
- Moves payment authorization to the end of the workflow, see explanation in the comment below


packages/core/core-flows/src/cart/workflows/complete-cart.ts
```diff
@@ -88,7 +88,7 @@ export const completeCartWorkflow = createWorkflow(
   {
     name: completeCartWorkflowId,
     store: true,
-    idempotent: true,
+    idempotent: false,
```
olivermrbl, line 92: comment: `idempotent: true` will return the result of a previous execution even in case of failures. This makes for a difficult FE implementation as it requires a fresh cart to execute again. 

For example, if the shipping profile constraints are not satisfied, the workflow throws when shipping methods are attempted created. All subsequent runs will fail with the same error, even if the cart is updated to satisfy the constraints.

packages/core/core-flows/src/cart/workflows/complete-cart.ts
```diff
@@ -325,44 +318,47 @@ export const completeCartWorkflow = createWorkflow(
...
+        input,
+      })
 
-          for (const adjustment of shippingAdjustments) {
-            promotionUsage.push({
-              amount: adjustment.amount,
-              code: adjustment.code!,
-            })
-          }
+      // We authorize payment sessions at the very end of the workflow to minimize the risk of
```
olivermrbl, line 348: comment: See inline comment for why this has been moved to the end of the workflow.

packages/core/core-flows/src/cart/workflows/complete-cart.ts
```diff
@@ -111,22 +111,7 @@ export const completeCartWorkflow = createWorkflow(
...
-    // but needs to be before the validation step
-    const paymentSessions = when(
-      "create-order-payment-compensation",
-      { orderId },
-      ({ orderId }) => !orderId
-    ).then(() => {
-      const paymentSessions = validateCartPaymentsStep({ cart })
-      // purpose of this step is to run compensation if cart completion fails
-      // and tries to cancel or refund the payment depending on the status.
-      compensatePaymentIfNeededStep({
```
olivermrbl: comment: @fPolic, @carlos-r-l-rodrigues I believe this is not needed if the payment authorization is the final step. Please correct me if I am wrong.

packages/core/core-flows/src/cart/workflows/complete-cart.ts
olivermrbl: See [FRMW-2958](https://linear.app/medusajs/issue/FRMW-2958/rethink-workflow-idempotency-options) for more.

packages/core/core-flows/src/cart/workflows/complete-cart.ts
```diff
@@ -298,6 +260,37 @@ export const completeCartWorkflow = createWorkflow(
         }
       })
 
+      const promotionUsage = transform(
```
olivermrbl, line 279: comment: @riqwan, would it be OK to move this to a parallelized step? It used to be a separate one, if you look at the diff.

riqwan: looks like adjustments have already been settled by then, should be ok 🤞🏻

packages/core/core-flows/src/cart/workflows/complete-cart.ts
fPolic: It depends on how we want to handle errors with `autocapture` flow.
If the cart complete flow is triggered from the webhook on payment success, and something fails before the authorize step, there would be a payment but order wouldn't be created right?

carlos-r-l-rodrigues: The issue here is what Frane said, when the Payment was authorized/captured before and then this workflow is triggered by the webhook. We have to remember that the "authorize step" in practice is only reading a status from the provider, and not doing any action per se.
At that point, anything that goes wrong in this workflow, we should compensate the payment.

packages/core/core-flows/src/order/steps/add-order-transaction.ts
```diff
@@ -5,27 +5,32 @@ import { StepResponse, createStep } from "@medusajs/framework/workflows-sdk"
...
-  async (
-    data: AddOrderTransactionStepInput,
-    { container }
-  ) => {
+  async (data: AddOrderTransactionStepInput, { container }) => {
     const service = container.resolve(Modules.ORDER)
 
     const trxsData = Array.isArray(data) ? data : [data]
 
+    if (!trxsData.length) {
```
carlos-r-l-rodrigues, line 30: 👍🏻

packages/core/core-flows/src/cart/workflows/complete-cart.ts
olivermrbl: > If the cart complete flow is triggered from the webhook on payment success, and something fails before the authorize step, there would be a payment but order wouldn't be created right?

The idea here would be to ensure that no payment exists before the very end of the cart-completion flow, so I don't think this is a realistic scenario. Let me know if I understood this right?

carlos-r-l-rodrigues (approved):
LGTM

fPolic (approved):
LGTM

PR #12354, opened 2025-05-04
fix(product): Return updated collections

olivermrbl:
Collection updates without a `products` payload were not returned from the service layer due to an incorrect early return

Fixes SUP-1488


adrien2p (approved):
LGTM, do we have any integration tests to add or update?

fPolic (approved):

PR #12357, opened 2025-05-05
fix: export ProductImage to allow for custom links to add alt text

thetutlage:
Fixes: FRMW-2959


packages/modules/product/src/services/product-module-service.ts
```diff
@@ -96,6 +96,9 @@ export default class ProductModuleService
     ProductVariant: {
       dto: ProductTypes.ProductVariantDTO
     }
+    ProductImage: {
```
adrien2p, line 99: sugg: you might need to check if you need to add it to the custom joiner config in the module. Also, you can update the integration tests that validate the output linkable 👍

adrien2p (approved):

PR #12362, opened 2025-05-05
feat(workflows-*): Allow to re run non idempotent but stored workflow with the same transaction id if considered done

adrien2p:
RESOLVES FRMW-2958

**What**
- If a workflow is not considered done, is not idempotent and is stored, allow to execute a new instance with the same transaction id but a different run id
- Ensure concurrent run to start duplicated workflow with the same manual transaction id provided will fail on the second one (the workflow must stored for that to kicks)


adrien2p:
We will add some more tests and add a flag for cancellation in cases it is not idempotent but stored and need to be cancelled. In that case the last run must be returned
cc @olivermrbl

sradevski (commented):
Nice, I'll give it a try later today

olivermrbl:
@adrien2p very nice, will give this a spin on the cart-completion flow 👍

olivermrbl (approved):
This works as expected on my end, nice work!

adrien2p:
@sradevski can you try again, it should be good now 👍

PR #12367, opened 2025-05-05
fix: refactor batch product update

peterlgh7:
An improvement on https://github.com/medusajs/medusa/pull/12333 based on discussion with @adrien2p . This PR refactors updateProducts to not use `upsertWithReplace` and instead rely on the entity manager's unit of work for the batch updates.

The previous optimization on `upsertWithReplace` (actually on upsertMany_) didn't fully solve the performance issues, cause `upsertWithReplace` was being called on a loop for each product variant and option, which negates the effects of `upsertWithReplace`'s batching.

### DIFF
--- a/packages/modules/product/integration-tests/__tests__/product-module-service/products.spec.ts
+++ b/packages/modules/product/integration-tests/__tests__/product-module-service/products.spec.ts
@@ -12,10 +12,10 @@ import {
   ProductStatus,
 } from "@medusajs/framework/utils"
 import {
-  ProductImage,
   Product,
   ProductCategory,
   ProductCollection,
+  ProductImage,
   ProductType,
 } from "@models"
 
@@ -400,9 +400,7 @@ moduleIntegrationTestRunner<IProductModuleService>({
               options: { size: "x", color: "red" }, // update options
             },
             {
-              id: existingVariant2.id,
-              title: "new variant 2",
-              options: { size: "l", color: "green" }, // just preserve old one
+              id: existingVariant2.id, // just preserve old one
             },
             {
               product_id: product.id,
@@ -722,30 +720,6 @@ moduleIntegrationTestRunner<IProductModuleService>({
           expect(error).toEqual(`Product with id: does-not-exist was not found`)
         })
 
-        it("should throw because variant doesn't have all options set", async () => {
-          const error = await service
-            .createProducts([
-              {
-                title: "Product with variants and options",
-                options: [
-                  { title: "opt1", values: ["1", "2"] },
-                  { title: "opt2", values: ["3", "4"] },
-                ],
-                variants: [
-                  {
-                    title: "missing option",
-                    options: { opt1: "1" },
-                  },
-                ],
-              },
-            ])
-            .catch((e) => e)
-
-          expect(error.message).toEqual(
-            `Product "Product with variants and options" has variants with missing options: [missing option]`
-          )
-        })
-
         it("should update, create and delete variants", async () => {
           const updateData = {
             id: productTwo.id,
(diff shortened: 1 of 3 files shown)


packages/modules/product/integration-tests/__tests__/product-module-service/products.spec.ts
```diff
@@ -722,30 +720,6 @@ moduleIntegrationTestRunner<IProductModuleService>({
           expect(error).toEqual(`Product with id: does-not-exist was not found`)
         })
 
-        it("should throw because variant doesn't have all options set", async () => {
```
peterlgh7, line 725: This test wasn't removed, it was just moved under "create"

packages/modules/product/src/services/product-module-service.ts
```diff
@@ -1661,124 +1661,65 @@ export default class ProductModuleService
...
-          } else {
-            await this.productImageService_.delete(
-              { product_id: upsertedProduct.id },
-              sharedContext
-            )
-          }
-        }
-      })
-    )
+    await (sharedContext.transactionManager as EntityManager).flush()
```
adrien2p: todo: we should not flush here, the transaction will be flushed automatically. Also, this service is not supposed to know about mikro orm internals. Eventually, the product repository can if we needed

packages/modules/product/src/services/product-module-service.ts
```diff
@@ -1661,124 +1661,65 @@ export default class ProductModuleService
...
-            {
-              product_id: upsertedProduct.id,
-              id: {
-                $nin: upsertedProduct.variants.map(({ id }) => id),
-              },
-            },
-            sharedContext
-          )
-        }
+      wrap(product!).assign(update)
```
adrien2p: todo: we should not use wrap here, it means that the module service is aware of the form internals. I would instead assign the target props directly wdyt?

packages/modules/product/src/services/product-module-service.ts
```diff
@@ -1661,124 +1661,65 @@ export default class ProductModuleService
...
-            ProductModuleService.assignOptionsToVariants(
-              product.variants.map((v) => ({
-                ...v,
-                product_id: upsertedProduct.id,
-              })) ?? [],
-              allOptions
-            )
+      // Assign the options first, so they'll be available for the variants loop below
+      if (update.options) {
+        wrap(product).assign({ options: update.options })
```
adrien2p: todo: same comment as bellow, sorry I ve started the review from bottom to top 😂

packages/modules/product/src/services/product-module-service.ts
```diff
@@ -1661,124 +1661,65 @@ export default class ProductModuleService
...
-          if (product.variants?.length) {
-            allOptions = await this.productOptionService_.list(
-              { product_id: upsertedProduct.id },
-              { relations: ["values"] },
-              sharedContext
-            )
-          }
-        }
+    for (const normalizedProduct of normalizedProducts) {
+      const update = normalizedProduct as any
```
adrien2p: suggestion: should we use `UpdateProductInput` as the type here?

packages/modules/product/integration-tests/__tests__/product-module-service/products.spec.ts
```diff
@@ -849,6 +823,136 @@ moduleIntegrationTestRunner<IProductModuleService>({
...
+                {
+                  title: "variant 1",
+                  options: { size: "small" },
+                },
+              ],
+            })
+            .catch((e) => e)
+
+          expect(error?.message).toEqual(
+            `Product has 2 but there were 1 provided option values for the variant: variant 1.`
```
adrien2p:
```suggestion
            `Product has 2 option values but there were 1 provided option values for the variant: variant 1.`
```

adrien2p (commented):
Excellent work man 💪 few adjustments in comments

packages/modules/product/src/services/product-module-service.ts
peterlgh7: I understand. I was flushing to maintain the previous behavior, i.e. since upsertWithReplace was firing native queries, they were triggered right away, but I agree it should not make a difference, and indeed I just checked tests continue to pass without it.

packages/modules/product/src/services/product-module-service.ts
peterlgh7: I'd love to, but we're not respecting the type in `normalizeUpdateProductInput`. Among other transformations, we're doing:
```
if (productData.tag_ids) {
  ;(productData as any).tags = productData.tag_ids.map((cid) => ({
    id: cid,
  }))
  delete productData.tag_ids
}
```
which adds a new field. Or even changing field types here:

```
if (productData.options?.length) {
  ;(productData as any).options = productData.options?.map((option) => {
    const dbOption = dbOptions.find(
      (o) => o.title === option.title && o.product_id === productData.id
    )
    return {
      title: option.title,
      values: option.values?.map((value) => {
        const dbValue = dbOption?.values?.find(
          (val) => val.value === value
        )
        return {
          value: value,
          ...(dbValue ? { id: dbValue.id } : {}),
        }
      }),
      ...(dbOption ? { id: dbOption.id } : {}),
    }
  })
}
```

adrien2p:
one suggestion @peterlgh7 , since the pr will be squashed before being merged, you can just push any new commit without squashing or amending, it will be easier to see the changes made from a particular commit, otherwise it become harder to understand what have changed in between 😄 let me know if that's alright 👍

peterlgh7:
@adrien2p sure, I'm too used to amending, but that makes sense!

packages/modules/product/src/services/product-module-service.ts
peterlgh7: My bad for not realizing the service should be ORM-agnostic. I'm looking into alternatives now, it's just that assign makes applying deep updates very simple.

adrien2p: You can use the repository layer or create a new repository to handle it if you think it would make it easier, here you can have access to the orm. Repositories are auto generated by our loaders for every entity available unless one is created in the repository directory of the module 👍

peterlgh7: @adrien2p I agree moving to the repo is the best way forward, but I read your comment just now and I had done some experimentation yesterday. Somehow I managed to make it work by using the default `update` in `MedusaInternalService`. You can see it in the latest commit. I don't like the solution though, cause it uses some tricks (you can see it in the code comments). So while I agree I should just move my original implementation to the repo, I also feel we should add some jsdocs to `MedusaInternalService` explaining how it handles deep updates. In a nutshell, it will handle them just fine if you have previously loaded all relations for all entities to update, but this is something I know only because I read it in mikroORM's `em.assign` documentation, which defeats the purpose of not using the em directly. Currently without any docs, I think the behavior of `update` can be surprising. Alternatively we could ensure update always handles deep updates by loading relations within it, but that's a heavier lift right now I think (and it's probably the problem `insertWithReplace` was trying to solve?).

Please also see my comment regarding metadata. The products/batch endpoint updates metadata all-or-nothing (no-merging), and there's a specific integration test to check that. But I see you implemented some specific metadata merging logic in update recently. Should we align `products/batch`, or is the difference intended?

PS: assigning the props directly didn't work for relations, cause I need to pass entities, not just POJOs, and I need access to the em for creating them.

packages/modules/product/src/services/product-module-service.ts
```diff
@@ -1661,124 +1660,82 @@ export default class ProductModuleService
...
-          // If the options weren't affected, but the user is changing the variants, make sure we have all options available locally
-          if (product.variants?.length) {
-            allOptions = await this.productOptionService_.list(
-              { product_id: upsertedProduct.id },
-              { relations: ["values"] },
-              sharedContext
-            )
-          }
+    await this.productService_.update(
+      normalizedProducts.map((normalizedProduct: any) => {
```
adrien2p: nit: I prefer extracting data processing outside a function call but as you prefer. I think it would make it more readable

packages/modules/product/src/services/product-module-service.ts
```diff
@@ -1661,124 +1660,82 @@ export default class ProductModuleService
...
+        const update = { ...normalizedProduct }
```
adrien2p: Q: Here you are doing a shallowcopy, so I would say that when you assign to nested object it will be also applied to the original normalized product. Is that intended?

packages/modules/product/src/services/product-module-service.ts
```diff
@@ -1661,124 +1660,82 @@ export default class ProductModuleService
...
 
-    const { entities: productData } =
-      await this.productService_.upsertWithReplace(
-        normalizedProducts,
-        {
-          relations: ["tags", "categories"],
-        },
-        sharedContext
-      )
+    const products = await this.productService_.list(
```
adrien2p: discussion: We can keep it that way for now and see the impact, otherwise we might just need (if needed) to decoupled the relation processing and let the unit of work group them accordingly

adrien2p (approved):
Great work 💪 just a few points in comment let me know what you think

packages/modules/product/src/services/product-module-service.ts
peterlgh7: Not sure what you mean. normalizedProduct is the update POJO, not the entity. In any case, the reason I'm doing the shallow copy is because I want to remove the variants from the first update, but still have it on the second update.

packages/modules/product/src/services/product-module-service.ts
peterlgh7: Do you mean moving this mapping function to a helper function?

packages/modules/product/src/services/product-module-service.ts
peterlgh7: What would decoupling look like? Updating these relations separately from the main entity?

packages/modules/product/src/services/product-module-service.ts
adrien2p: I was thinking just assigning the map to a const and then calling the update method but as you prefer. No strong opinions it is more a preference matter 🙃

packages/modules/product/src/services/product-module-service.ts
adrien2p: Yes exactly, we have for each entity it’s counter part internal service and repository. I believe for now it is not something we have to do but just throwing out loud this approach. Knowing that mikro orm philosophy is being explicit instead of implicit but we can discuss that in a huddle separately from this pr as it is not blocking and can be decided separately wdyt?

packages/modules/product/src/services/product-module-service.ts
adrien2p: Yes I would say the metadata update should have the same behavior in batch and single update. I thought we added that somewhere in the internal service from the top of my head

packages/modules/product/src/services/product-module-service.ts
peterlgh7: ah ok, I agree. I did have a feeling this was too "single statementy".

packages/modules/product/src/services/product-module-service.ts
peterlgh7: sure, let's discuss separately.

PR #12396, opened 2025-05-07
feat: create CSV normalizer to normalize a CSV file

thetutlage:
Fixes: FRMW-2962

This PR adds a new CSV normalizer that does the following things.

- Remove normalizations of v1 columns
- Update the columns to be processed as per the doc published here. https://docs.medusajs.com/user-guide/products/import#csv-file-format
- Remove support for processing timestamps and `Variant Product Id`.
- Do not make any DB calls when normalizing the CSV or validating its format. DB calls must be made when processing rows in chunks.
- Make validations contain the CSV row number. See screenshot
- Add tests for the normalizer and move it to the `utils`.
- Validate format for all fields. Earlier some fields like `Product Weight` allowed for non-numeric values as well.


packages/core/utils/src/product/csv-normalizer.ts
```diff
@@ -0,0 +1,498 @@
...
+  csvRow: Record<string, string | boolean | number>,
+  rowColumns: string[],
+  output: Output
+) => void
+
+/**
+ * Collection of static product columns whose values must be copied
+ * as it is without any further processing.
+ */
+const productStaticColumns: {
```
thetutlage, line 18: While going through these columns, you might feel why are we repeating all this code that looks almost the same for all the columns.

The goal here is to not write terse code. Instead, the goal is to be descriptive with the fields we handle within the CSV and localize the logic of how a field is handled/normalized. 

To me, this code is easy to reason about vs going through a loop and writing inline conditions there for certain or using transforms like `snakeCase`, `camelCase` to convert the CSV column to DTO property name. 🙂

packages/core/core-flows/src/product/steps/normalize-products-v1.ts
```diff
@@ -0,0 +1,45 @@
...
+ */
+export type NormalizeProductCsvStepInput = string
+
+export const normalizeCsvStepId = "normalize-product-csv"
+/**
+ * This step parses a CSV file holding products to import, returning the products as
+ * objects that can be imported.
+ *
+ * @example
+ * const data = parseProductCsvStep("products.csv")
```
olivermrbl, line 17:
```suggestion
 * const data = normalizeCsvStep("products.csv")
```

packages/core/utils/src/common/try-convert-to-boolean.ts
```diff
@@ -0,0 +1,23 @@
+/**
+ * Transforms a value to number or returns the default value
+ * when original value cannot be casted to number
```
olivermrbl:
```suggestion
 * Transforms a value to a boolean or returns the default value
 * when original value cannot be casted to a boolean
```

packages/core/utils/src/product/csv-normalizer.ts
```diff
@@ -0,0 +1,520 @@
...
+   */
+  if (priceTokens.length === 1) {
+    return {
+      region: null,
+      iso: priceTokens[0].toLowerCase(),
+    }
+  }
+
+  /**
+   * The price was specified as "Variant Price [EUR]" or "Variant Price Europe [EUR]"
```
olivermrbl: question: Is this description accurate? Wouldn't this be:
```
"Variant Price Europe EUR" or "Variant Price North America USD"
```

Do we need the square brackets?

olivermrbl: We could say the currency should always be at the end of the column

packages/core/utils/src/product/csv-normalizer.ts
```diff
@@ -0,0 +1,520 @@
+import {
```
olivermrbl, line 1: thought: lots of good stuff in here, that could be used for a more generic parser

olivermrbl (commented):
Clean, easy to read

packages/core/utils/src/product/csv-normalizer.ts
```diff
@@ -0,0 +1,520 @@
...
+        rowNumber,
+        `Invalid price format used by "${columnName}". Expect column name to contain the ISO code inside square brackets. For example: "Variant Price [USD]"`
+      )
+    }
+
+    return {
+      iso: iso
+        .replace(new RegExp(`^[`), "")
+        .replace(new RegExp(`]$`), "")
+        .toLowerCase(),
```
carlos-r-l-rodrigues:
```suggestion
      iso: iso.replace(/^\[|\]$/g, "")
        .toLowerCase(),
```

packages/core/utils/src/product/csv-normalizer.ts
thetutlage: I referenced the docs here. Maybe, I misunderstood the formatting syntax.

CleanShot 2025-05-12 at 22 16 12@2x

carlos-r-l-rodrigues (approved):
LGTM

question: does `convertCsvToJson` already apply some ".trim()" on each column? just to avoid some common issues working with this kind of file format?

packages/core/utils/src/product/csv-normalizer.ts
thetutlage: Yeah, @adrien2p mentioned that too. But, I resisted the urge to turn it into a generic parser. But we can always do that the moment we write another one

packages/core/utils/src/product/csv-normalizer.ts
olivermrbl: Ah yes, now I remember. 

To consistently capture the region name, we decided to use the brackets to surround it. 

With that in mind, the current parsing is actually slightly incorrect. Here are a few examples:
```
Variant Price [Europe] EUR
Variant Price [North America] USD
Variant Price [A super duper long region name] EUR
```

Having said that, I am also fine going with the current approach. This small change would just unify how currency is parsed, i.e. always as the last word and without brackets.

packages/core/utils/src/product/csv-normalizer.ts
olivermrbl: Makes sense 👍

packages/core/utils/src/product/csv-normalizer.ts
thetutlage: Alright. So the region name is inside the brackets and the code is at last word without any brackets? If so, I will fix the logic

packages/core/core-flows/src/product/steps/normalize-products-v1.ts
```diff
@@ -0,0 +1,45 @@
...
+ */
+export const normalizeCsvStep = createStep(
+  normalizeCsvStepId,
+  async (fileContent: NormalizeProductCsvStepInput, { container }) => {
+    const csvProducts = convertCsvToJson(fileContent)
```
adrien2p:
```suggestion
    const csvProducts = convertCsvToJson<ConstructorParameters<typeof CSVNormalizer>[0][0]>(fileContent)
```

packages/core/core-flows/src/product/steps/normalize-products-v1.ts
```diff
@@ -0,0 +1,45 @@
...
+    const normalizer = new CSVNormalizer(csvProducts as any)
```
adrien2p:
```suggestion
    const normalizer = new CSVNormalizer(csvProducts)
```

packages/core/core-flows/src/product/steps/normalize-products-v1.ts
```diff
@@ -0,0 +1,45 @@
...
+    const projects = normalizer.proccess()
```
adrien2p:
```suggestion
    const products = normalizer.proccess()
```

adrien2p: let me know what you think

packages/core/core-flows/src/product/steps/normalize-products-v1.ts
```diff
@@ -0,0 +1,45 @@
...
+
+    const create = Object.keys(projects.toCreate).reduce<
+      (typeof projects)["toCreate"][keyof (typeof projects)["toCreate"]][]
+    >((result, toCreateId) => {
```
adrien2p: question: should it be handle instead of toCreateId?

packages/core/utils/src/common/try-convert-to-boolean.ts
```diff
@@ -0,0 +1,23 @@
+/**
+ * Transforms a value to a boolean or returns the default value
+ * when original value cannot be casted to a boolean
+ */
+export function tryConvertToBoolean(value: unknown): boolean | undefined
+export function tryConvertToBoolean<T>(
+  value: unknown,
+  defaultValue: T
+): number | T
```
adrien2p: Q: number is not represented in the implementation function, but also, we always return a boolean or default value or undefined. Do we want to return a number?

packages/core/core-flows/src/product/steps/normalize-products-v1.ts
thetutlage: Holding this off until I add Zod validations to the normalized output. That will be another PR since I do not want to put too many things in the same PR :)

packages/core/utils/src/product/csv-normalizer.ts
```diff
@@ -0,0 +1,518 @@
...
+    MedusaError.Types.INVALID_DATA,
+    `Row ${rowNumber}: ${message}`
+  )
+}
+
+/**
+ * Normalizes a CSV value by removing the leading "\r" from the
+ * value.
+ */
+function normalizeValue<T>(value: T): T {
```
adrien2p, line 34: Q: should we also trim?

packages/core/utils/src/product/csv-normalizer.ts
```diff
@@ -0,0 +1,518 @@
...
+}
+
+/**
+ * Parses different patterns to extract variant price iso
+ * and the region name. The iso is converted to lowercase
+ */
+function parseVariantPriceColumn(columnName: string, rowNumber: number) {
+  const priceTokens = normalizeValue(columnName)
+    .replace("Variant Price ", "")
+    .split(" ")
```
adrien2p: Q: if we follow the logic of ''Variant Price [A super duper long region name] EUR" then splitting by space would not work. in that case would the following work

```ts
const normalizedValue = normalizeValue(columnName)
const potentialRegion = normalizedValue.match(/\[(.*)\]/g)?.[0]
const price = normalizedValue.split(' ').pop()
```

packages/core/utils/src/product/csv-normalizer.ts
```diff
@@ -0,0 +1,518 @@
...
+  return (csvRow, _, __, output) => {
+    const value = normalizeValue(csvRow[inputKey])
+    if (isPresent(value)) {
+      output[outputKey as any] = value
+    }
+  }
+}
+
+/**
+ * Processes the column value as a string
```
adrien2p:
```suggestion
 * Processes the column value as a boolean
```

packages/core/utils/src/product/csv-normalizer.ts
```diff
@@ -0,0 +1,518 @@
...
+    const existingIds = output[outputKey].map((item) => item[arrayItemKey])
+
+    rowColumns
+      .filter((rowKey) => inputMatcher.test(rowKey))
+      .forEach((rowKey) => {
+        const value = normalizeValue(csvRow[rowKey])
+        if (!existingIds.includes(value) && isPresent(value)) {
+          output[outputKey].push({ [arrayItemKey]: value })
+        }
+      })
```
adrien2p, line 142: It is more of a suggestion but as you prefer

```suggestion
    const existingIds = new Set(output[outputKey].map((item) => item[arrayItemKey]))

    rowColumns
      .filter((rowKey) => inputMatcher.test(rowKey))
      .forEach((rowKey) => {
        const value = normalizeValue(csvRow[rowKey])
        if (isPresent(value) && !existingIds.has(value)) {
          output[outputKey].push({ [arrayItemKey]: value })
        }
      })
```

packages/core/utils/src/product/csv-normalizer.ts
```diff
@@ -0,0 +1,518 @@
...
+/**
+ * Collection of static product columns whose values must be copied
+ * as it is without any further processing.
+ */
+const productStaticColumns: {
+  [columnName: string]: ColumnProcessor<{
+    [K in keyof AdminCreateProduct | "id"]?: any
+  }>
+} = {
+  "Product Id": processAsString("Product Id", "id"),
```
adrien2p: discussion: I am wondering something, since the column names have many upper/lower case, would it be reasonable to be case insensitive? I believe that in any case when we retrieve meaningful information from a column name we lowercase it to normalize. In that case we could just always lowercase the column name when we normalize them and process them. wdyt?

packages/core/utils/src/product/csv-normalizer.ts
```diff
@@ -0,0 +1,518 @@
...
+  }
+
+  /**
+   * Initializes a product object or returns an existing one
+   * by its id. The products with ids are treated as updates
+   */
+  #getOrInitializeProductById(id: string) {
+    if (!this.#products.toUpdate[id]) {
+      this.#products.toUpdate[id] = {
+        variants: [],
```
adrien2p: Q: would it make sense to not initialize the variants here and instead below assign variants to `product['variants']`  directly? variants are optional if I am correct and in case you just want to update the product attribute I think providing an empty variant array would result in removing all existing variants but to verify

packages/core/utils/src/product/csv-normalizer.ts
adrien2p: as we discussed, being case insensitive can prevent silly mistakes from the user but also missing data because not parsed (since column names are currently strict). That plus checking any column that cant be processed should allow for easier usage of the import tool and prevent missing data by mistake that the user would only be able to see by going through the imported data which can be massive

thetutlage: Done. Made CSV case insensitive and also we will throw an Error when unknown columns are defined in the CSV. CC @olivermrbl

packages/core/utils/src/product/csv-normalizer.ts
thetutlage: Done

packages/core/utils/src/product/csv-normalizer.ts
thetutlage: Done

olivermrbl:
Just removing automerge until release is out :)

olivermrbl:
Feel free to add again 😄

PR #12397, opened 2025-05-07
fix(core-flows): use product title for line item title

fPolic:
**What**
- use product title as title and variant title as subtitle

**TODO**
- [x] tests
- [ ] changeset


sradevski (approved):
This annoyed me personally :D

olivermrbl (approved):

PR #12406, opened 2025-05-08
fix: Use logger in the test runner and migration commands

sradevski:
Tests can get very noisy due to migrations running over and over again, and there is no way to disable the logs.

The PR changes the logging to always use the logger, so the LOG_LEVEL envvar can be used to control the logs

### DIFF
--- a/packages/core/utils/src/modules-sdk/migration-scripts/migration-down.ts
+++ b/packages/core/utils/src/modules-sdk/migration-scripts/migration-down.ts
@@ -28,8 +28,8 @@ export function buildRevertMigrationScript({ moduleName, pathToMigrations }) {
   > = {}) {
     logger ??= console as unknown as Logger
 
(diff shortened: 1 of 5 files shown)


packages/core/utils/src/modules-sdk/migration-scripts/migration-down.ts
```diff
@@ -28,8 +28,8 @@ export function buildRevertMigrationScript({ moduleName, pathToMigrations }) {
   > = {}) {
     logger ??= console as unknown as Logger
 
-    console.log(new Array(TERMINAL_SIZE).join("-"))
```
sradevski, line 31: Maybe using console was intentional, but it nevertheless adds a lot of noise so it's better to always use the logger

carlos-r-l-rodrigues (approved):

riqwan (approved):

PR #12425, opened 2025-05-09
feat(core-flows): Emit cart trasnferred customer

adrien2p:
**What**
When a cart is transferred emit an event notifying of the action with the cart id and the customer id


shahednasser (approved):
LGTM

olivermrbl (approved):

PR #12439, opened 2025-05-09
Chore/test runner improvements

adrien2p:
**What**
Make sure there is no open handles left and that the shutdown function are properly called. Refactor and improve the medusa test runner. Make sure all modules instances are released and cleaned up

**NOTE:**
On a separate PR we can continue the investigation for the memory growing over time while the tests execute


adrien2p:
@olivermrbl @sradevski shall we merge that one already?

carlos-r-l-rodrigues (approved):
LGTM

it's missing a changeset

thetutlage (approved):

PR #12441, opened 2025-05-09
fix(tax): Make system provider registration backward compatible

olivermrbl:
### What
Register the out-of-the-box system tax provider as `tp_system` in the dependency container.

### Why
To avoid the minor breaking change introduced in #12297 in which the registration key was renamed to `tp_system_system`.

### DIFF
--- a/integration-tests/http/__tests__/order-edits/order-edits.spec.ts
+++ b/integration-tests/http/__tests__/order-edits/order-edits.spec.ts
@@ -58,7 +58,7 @@ medusaIntegrationTestRunner({
         await api.post(
           "/admin/tax-regions",
           {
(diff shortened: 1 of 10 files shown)


fPolic (approved):
LGTM

adrien2p (approved):
LGTM, one question, should we still have a script for users that used this version and created default the tax provider with id? Or was it just for the snapshot and not part of an official release?

olivermrbl:
> LGTM, one question, should we still have a script for users that used this version and created default the tax provider with id? Or was it just for the snapshot and not part of an official release?

I believe what you are describing would only be an issue if you had been using the snapshot, but I am not entirely sure I understand the case you are describing. Can you give an example?

PR #12447, opened 2025-05-12
fix(core-flows, dashboard): reservation recreation on fulfilment cancel + allocation button display

fPolic:
**What**
- fix recreation of reservations on fulfilment cancel
- fix allocate items button display

---

CLOSES CMRC-1018

### DIFF
--- a/packages/core/core-flows/src/order/workflows/cancel-order-fulfillment.ts
+++ b/packages/core/core-flows/src/order/workflows/cancel-order-fulfillment.ts
@@ -203,6 +203,7 @@ function prepareInventoryUpdate({
     inventory_item_id: string
     location_id: string
     quantity: BigNumberInput
+    line_item_id: string
   }[] = []
   const toUpdate: {
     id: string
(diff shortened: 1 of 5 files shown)


packages/core/core-flows/src/order/workflows/cancel-order-fulfillment.ts
```diff
@@ -203,6 +203,7 @@ function prepareInventoryUpdate({
     inventory_item_id: string
     location_id: string
     quantity: BigNumberInput
+    line_item_id: string
```
olivermrbl, line 206: suggestion: Could we add a test for this?

olivermrbl (approved):

packages/core/core-flows/src/order/workflows/cancel-order-fulfillment.ts
fPolic: Added!

olivermrbl (approved):
Niice

PR #12449, opened 2025-05-12
fix(workflows-sdk): Miss match context usage within run as step

adrien2p:
**What**

Currently, runAsStep keep reference of the workflow context that is being run as step, except that the step is composed for the current workflow composition and not the workflow being run as a step. Therefore, the context are currently miss matched leading to wrong configuration being used in case of async workflows.

**BUG**
This fix allow the runAsStep to use the current composition context to configure the step for the sub workflow to be run

**BUG BREAKING**
fix the step config wrongly used to wrap async step handlers. Now steps configured async through .config that returns a new step response will indeed marked itself as success without the need for background execution or calling setStepSuccess (as it was expected originally)

**FEATURE**
This pr also add support for cancelling running transaction, the transaction will be marked as being cancelled, once the current step finished, it will cancel the transaction to start compensating all previous steps including itself


carlos-r-l-rodrigues (approved):

thetutlage (approved):

olivermrbl:
Is this safe to merge?

adrien2p:
I would wait tomorrow 🙏

carlos-r-l-rodrigues:
Safe it is, what we need to confirm is that this will fix a bug Stevche has found.

adrien2p:
@olivermrbl could you review this one as well please, we d like to get it in 🚀 we tested with stevche as well to be sure of the behaviour

sradevski (approved):
Tested the latest snapshot on Platform, all green! I'll test it out with the deployment cancellation in the next couple of days as well

PR #12450, opened 2025-05-12
fix(dashboard, medusa): validate provider for top level tax regions

fPolic:
**What**
- prevent top level tax region creation without a provider on Admin and in HTTP validation layer

### DIFF
--- a/integration-tests/http/__tests__/tax-region/admin/tax-region.spec.ts
+++ b/integration-tests/http/__tests__/tax-region/admin/tax-region.spec.ts
@@ -24,7 +24,7 @@ medusaIntegrationTestRunner({
             "/admin/tax-regions",
             {
               country_code: "us",
-              provider_id: "tp_system", // TODO: update when Oli's PR is merged
+              provider_id: "tp_system",
             },
             adminHeaders
           )
--- a/integration-tests/http/__tests__/tax-region/admin/tax-region.spec.ts
+++ b/integration-tests/http/__tests__/tax-region/admin/tax-region.spec.ts
@@ -108,7 +108,7 @@ medusaIntegrationTestRunner({
           expect(status).toEqual(400)
           expect(data).toEqual({
             message:
-              "Provider is required when creating a non-province tax region.",
+              "Invalid request: Provider is required when creating a non-province tax region.",
             type: "invalid_data",
           })
         })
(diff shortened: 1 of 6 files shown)


integration-tests/http/__tests__/tax-region/admin/tax-region.spec.ts
```diff
@@ -20,12 +20,21 @@ medusaIntegrationTestRunner({
         let taxRegion
 
         beforeEach(async () => {
+          const parentRegion = await api.post(
+            "/admin/tax-regions",
+            {
+              country_code: "us",
+              provider_id: "tp_system_system", // TODO: update when Oli's PR is merged
```
fPolic: todo: depnds on https://github.com/medusajs/medusa/pull/12441

packages/medusa/src/api/admin/tax-regions/validators.ts
```diff
@@ -57,6 +58,16 @@ export const AdminCreateTaxRegion = z.object({
   metadata: z.record(z.unknown()).nullish(),
 })
 
+export const AdminCreateTaxRegion = WithAdditionalData(
+  CreateTaxRegion,
+  (schema) => {
+    return schema.refine((data) => data.parent_id || data.provider_id, {
```
olivermrbl, line 64: question: Does this work as expected? I think I would have expected this to be `!data.parent_id && !data.provider_id`, but perhaps I am just reading this wrong 😄

integration-tests/http/__tests__/tax-region/admin/tax-region.spec.ts
```diff
@@ -20,12 +20,21 @@ medusaIntegrationTestRunner({
         let taxRegion
 
         beforeEach(async () => {
+          const parentRegion = await api.post(
```
olivermrbl, line 23: suggestion: should we also add a test ensuring that we can create a province region without a provider? It seems to be missing

fPolic: this is tested implcitley in `beforeEach` where we create a parent tax region and a province region, but I added an explicit case as well

packages/medusa/src/api/admin/tax-regions/validators.ts
fPolic: Yeah was expecting the same but refinement function should test the expected condition, i.e. return falsy to signal failure. So the condition is a negation of your condition.

olivermrbl: TIL, interesting design choice 😄

olivermrbl (approved):

PR #12473, opened 2025-05-14
feat: define validators and use normalize-products step

thetutlage:
Fixes: FRMW-2965

In this PR we replace/remove the existing step to normalize a CSV file with the newly written CSV normalizer and also we validate the file contents further using a Zod schema.

I have duplicated the schema for now. But it is makes sense to re-use the schema for CSV validating and `/admin/products/batch`, then I can keep one source of truth under utils and re-export it. WDYT?

**Screenshots of some errors after validating the file strictly**

CleanShot 2025-05-15 at 16 36 46@2x

CleanShot 2025-05-15 at 16 36 34@2x


integration-tests/http/__tests__/product/admin/product-import.spec.ts
```diff
@@ -554,52 +576,19 @@ medusaIntegrationTestRunner({
...
-          expect.objectContaining({
-            field: "Test product",
-          })
+        expect(batchJobRes.response.data.message).toEqual(
+          'Invalid column name(s) "Product field"'
         )
       })
 
-      it("supports importing the v1 template", async () => {
+      it.skip("supports importing the v1 template", async () => {
```
thetutlage: Should we remove this test altogether?

integration-tests/http/__tests__/product/admin/product-import.spec.ts
```diff
@@ -454,7 +505,7 @@ medusaIntegrationTestRunner({
         )
       })
 
-      it("should fail on invalid region in prices being present in the CSV", async () => {
+      it.skip("should fail on invalid region in prices being present in the CSV", async () => {
```
thetutlage: Currently we fetch and test the imported file for the region ids.

However, in the revamp, I have moved all DB calls to the processing of the CSV file. Therefore we cannot verify for valid regions until the file is processed.

Should I remove this test and add a new one once the processing of CSV is also revamped?

carlos-r-l-rodrigues (approved):
LGTM

ps: it is missing a changeset

adrien2p (approved):
LGTM 🔥

PR #12478, opened 2025-05-14
fix(product, utils): handle metadata key deletion on product update

fPolic:
**What**
- handle metadata updates properly when updating products
- remove unused `setMetadata`, which has been replaced with `mergeMetadata`

**Why**
- metadata key couldn't be deleted on product entities

reference: https://github.com/medusajs/medusa/pull/11466


packages/modules/product/src/repositories/product.ts
```diff
@@ -69,6 +74,10 @@ export class ProductRepository extends DALUtils.mikroOrmBaseRepositoryFactory(
         }))
       }
 
+      if (isPresent(update.metadata)) {
+        update.metadata = mergeMetadata(product.metadata ?? {}, update.metadata)
+      }
```
adrien2p: Q: could you try `mergeObjectProperties` in the assign bellow? This is what we are using in the base repository

fPolic: This doesn't delete a key if it is set to "" like we do in `mergeMetadata`, right?

olivermrbl (approved):
LGTM, @adrien2p could you re-review this :)

adrien2p (approved):

olivermrbl (approved):

PR #12486, opened 2025-05-14
fix: returned updates order in upsertWithReplace

peterlgh7:


olivermrbl (approved):
LGTM

adrien2p (approved):
Lgtm, i agree that the coupling is not right. I have something in mind to remove it 👌

PR #12489, opened 2025-05-15
chore(core-flows): Remove data order tight coupling in update products

adrien2p:
**What**
Attempt to remove tight coupling with data index ordering in update products workflow


packages/core/core-flows/src/product/workflows/update-products.ts
```diff
@@ -85,6 +86,42 @@ export type UpdateProductWorkflowInput =
...
 
+type PreparedVariantPrices = {
+  product_id: string
+  variant_id: string
+  prices: CreateMoneyAmountDTO[]
+}
+
+// Helper function to create a hash from variant properties
+// @ts-ignore
+function createVariantHash(variant: {
```
adrien2p, line 97: can be removed

adrien2p:
can't really rely on hash like or proper hasing because data can changed because of default values for example which the workflow is not aware off and is the responsibility of the module

PR #12493, opened 2025-05-15
fix(core-flows): cart complete order address creation

fPolic:
**What**
- order addresses would be created with cart ids
- this would also prevent cart completion if a step after `createOrders` failed, since creation of an address with an existing id would be attempted

---

DEPENDS ON https://github.com/medusajs/medusa/pull/12172

---

CLOSES SUP-1628


olivermrbl:
Shouldn't the address be deleted in the compensation flow?

olivermrbl:
> Shouldn't the address be deleted in the compensation flow?

Another thought: what if the address is a saved address? Then, we probably don't want to delete it. There are some scenarios we might need to think through.

fPolic:
Good points. Regarding saved addresses, I might be missing something, but aren't they on the customer address table?
So I would say that the fix for the id here is still valid, but we need to override `OrderModuleService.deleteOrders` and additionally delete the addresses there, wdyt? I can do that as part of DraftOrder delete PR I'm working on now.

packages/core/core-flows/src/cart/workflows/complete-cart.ts
```diff
@@ -215,15 +215,30 @@ export const completeCartWorkflow = createWorkflow(
           .map((adjustment) => adjustment.code)
           .filter(Boolean)
 
+        const shippingAddress = cart.shipping_address
```
carlos-r-l-rodrigues, line 218: We need a transformer to check if cart.something is present, otherwise at this point in the workflow composition, it will always be true.

turistua:
@fPolic  please advice when this can be finalized?

olivermrbl (commented):
Can we add a few tests? 😄

olivermrbl (approved):

PR #12500, opened 2025-05-15
fix: fix weight/length/height/width types in updates

peterlgh7:
There's a mismatch in the types of weight/length/height/width between DTOs and entities. This caused troubles when calling `em.assign`. This fix just maps the type in the repository for now, but I think the solution should be to migrate the DB types to be numbers, which is already the case for ProductVariant.
This also causes e.g. that the types returned by `retrieveProduct` are string instead of the declared numbers.

### DIFF
--- a/integration-tests/http/__tests__/product/admin/product.spec.ts
+++ b/integration-tests/http/__tests__/product/admin/product.spec.ts
@@ -3043,12 +3043,20 @@ medusaIntegrationTestRunner({
       describe("batch methods", () => {
         it("successfully creates, updates, and deletes products", async () => {
           const createPayload = getProductFixture({
+            weight: 100,
+            length: 200,
+            height: 300,
+            width: 400,
             title: "Test batch create",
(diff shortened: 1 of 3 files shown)


olivermrbl:
Just in case :)

olivermrbl (approved):

adrien2p (approved):
LGTM, but I agree with you that we should migrate properly more than trying to mitigate the issue. Do we plan to do that in a separate pr? I imagine that can be considered as a breaking change but IMO still worth doing and explaining

olivermrbl:
> LGTM, but I agree with you that we should migrate properly more than trying to mitigate the issue. Do we plan to do that in a separate pr? I imagine that can be considered as a breaking change but IMO still worth doing and explaining

We agreed to do this next week 👍

lolzivkovic:
@adrien2p @olivermrbl Gents, what do you think about setting step values to allow a numbers with decimals as well?

PR #12501, opened 2025-05-15
fix(index): Apply various fixes to the index engine

carlos-r-l-rodrigues:
What:
 * Fix partition name length to avoid postgres limits of 63 chars
 * Index Enum fields as String
 * Make operators optional instead of required
 * Fix `buildSafeJsonPathQuery` for `$eq` and `$ne` against `null` value 
 * `query.index` hydration now have a default pagination of the index response size in order to force pagination and therefore `select-in` strategy
 * Fix event name normalization in the index config builder for object representation
 
FIXES: SUP-1640


packages/modules/workflow-engine-redis/integration-tests/__tests__/index.spec.ts
```diff
@@ -654,7 +654,6 @@ moduleIntegrationTestRunner<IWorkflowEngineService>({
             "workflow_transaction_timeout",
             {
               input: {},
-              transactionId: "trx",
```
adrien2p: Q: what is the issue with providing the transactionId? 🤔

adrien2p (approved):
LGTM

olivermrbl (approved):

PR #12503, opened 2025-05-15
fix(core-flows): fulfilment cancelation with shared inventory kit item

fPolic:
**What**
- manage reservations correctly when cancelling a fulfillment when the order has a shared inventory item
- improve testing - tests now check reservations and inventory levels when fulfilments is (re)created and canceled in case of
  - regular line item with a single inventory item
  - line item with multiple inventory items
  - multiple items that have the same inventory item


packages/core/core-flows/src/order/workflows/cancel-order-fulfillment.ts
```diff
@@ -228,24 +223,36 @@ function prepareInventoryUpdate({
...
+      (r) =>
+        r.inventory_item_id === iitem.inventory.id &&
+        r.line_item_id === fulfillmentItem.line_item_id
+    )
+
+    if (!reservation) {
+      toCreate.push({
+        inventory_item_id: iitem.inventory.id,
+        location_id: fulfillment.location_id,
+        quantity: fulfillmentItem.quantity, // <- this is the inventory quantity that is being fulfilled so it menas it does include the required quantity
```
adrien2p:
```suggestion
        quantity: fulfillmentItem.quantity, // <- this is the inventory quantity that is being fulfilled so it means it does include the required quantity
```

adrien2p (approved):
Sounds good, if I understand correctly, previously it was managed by inventory items and therefore each one was processed individually right?

PR #12505, opened 2025-05-15
feat: Add an analytics module and local and posthog providers

sradevski:
Question: Should we load the analytics module with a local provider by default?


adrien2p:
About the question in the description, I think this module wont be opt-in by default right? In any case, we can probably load the local provider by default and probably display a warning in the logs if the analytics is enabled but without a provider configured? just to make sure people are aware that they need to configure the provider or use their own wdyt?

packages/core/types/src/analytics/mutations.ts
```diff
@@ -0,0 +1,45 @@
...
+   */
+  group?: {
+    type?: string
+    id?: string
+  }
+  /**
+   * The properties that will be set on either the actor or group.
+   */
+  properties?: Record<string, any>
+}
```
adrien2p: Q: If one or the other are required, should we split this interface in two and maybe create a union type of them?

packages/modules/analytics/src/services/analytics-service.ts
```diff
@@ -0,0 +1,35 @@
...
+export default class AnalyticsService {
+  protected readonly analyticsProviderService_: AnalyticsProviderService
+
+  constructor({ analyticsProviderService }: InjectedDependencies) {
+    this.analyticsProviderService_ = analyticsProviderService
+  }
+
+  __hooks = {
+    onApplicationShutdown: async () => {
+      this.analyticsProviderService_.shutdown()
```
adrien2p: suggestion: we should probably await it, such that we are sure that it finishes once we resolve the promise, wdyt?

packages/modules/analytics/src/services/analytics-service.ts
```diff
@@ -0,0 +1,35 @@
...
+    },
+  }
+
+  getProvider() {
+    return this.analyticsProviderService_
+  }
+
+  async track(data: TrackAnalyticsEventDTO): Promise<void> {
+    this.analyticsProviderService_.track(data)
```
adrien2p: suggestion: since all the methods are async but we don't await or return the promise to be resolve, should we handle the errors in case there is any? Just to not have unhandled promises

sradevski:
@adrien2p my take is, there is little value is using the local provider so it's better to not include anything in the default config, but let's see what others think as well. I am happy to go either way. There are no consumers of the analytics module yet anyway (customers would need to add a subscriber themselves)

packages/modules/analytics/src/services/provider-service.ts
```diff
@@ -0,0 +1,56 @@
...
+    this.analyticsProvider_.track(data)
+  }
+
+  async identify(data: ProviderIdentifyAnalyticsEventDTO): Promise<void> {
+    this.analyticsProvider_.identify(data)
+  }
+
+  async shutdown(): Promise<void> {
+    await this.analyticsProvider_.shutdown?.()
+  }
```
adrien2p, line 55: Q: same question as for the module service promise handlin

adrien2p (approved):
LGTM 👍

packages/core/types/src/analytics/mutations.ts
sradevski: Yes definitely better 👍

packages/modules/analytics/src/services/analytics-service.ts
sradevski: Adding await, and I'll also add a try-catch to make the error nicer

adrien2p: that sounds good to me 👍

packages/modules/providers/analytics-posthog/CHANGELOG.md
```diff
@@ -0,0 +1,94 @@
+# @medusajs/file-local
```
olivermrbl: todo: update package name and remove content

packages/modules/providers/analytics-local/CHANGELOG.md
```diff
@@ -0,0 +1,94 @@
+# @medusajs/file-local
```
olivermrbl: todo: update package + remove content

packages/modules/analytics/package.json
```diff
@@ -0,0 +1,48 @@
+{
+  "name": "@medusajs/analytics",
+  "version": "2.8.1",
```
olivermrbl:
```suggestion
  "version": "2.8.2",
```

packages/core/types/src/analytics/mutations.ts
```diff
@@ -0,0 +1,45 @@
...
+  /**
+   * The event name
+   */
+  event: string
+  /**
+   * The actor of the event, if there is any
+   */
+  actor?: {
+    id: string
+  }
```
olivermrbl: question/nit: Do we need this to be an object? Could we just do `actorId`?

packages/core/types/src/analytics/mutations.ts
```diff
@@ -0,0 +1,45 @@
...
+  /**
+   * The group that the event is for, such as an organization or team.
+   * The "type" defines the name of the group (eg. "organization"), and the "id" is the id of the group.
+   */
+  group?: {
+    type?: string
+    id?: string
+  }
```
olivermrbl, line 17: nit/suggestion: This feels a little bit Posthog-specific; would it make more sense to add a `context` object, that can carry information specific to providers?

packages/core/types/src/analytics/mutations.ts
```diff
@@ -0,0 +1,45 @@
...
+  /**
+   * The properties of the event. The format and content depends on the provider.
+   */
+  properties?: Record<string, any>
+}
+
+export interface IdentifyActorDTO {
+  actor: {
+    id: string
+  }
```
olivermrbl: suggestion: See other comment, `actorId`

packages/core/types/src/analytics/mutations.ts
```diff
@@ -0,0 +1,45 @@
...
+  properties?: Record<string, any>
+}
+
+export interface IdentifyGroupDTO {
```
olivermrbl, line 29: Same comments as for the other type.

packages/core/types/src/analytics/mutations.ts
sradevski: There are multiple providers that support the same construct (eg. https://amplitude.com/docs/apis/analytics/group-identify) so I figured I should make it a first-class citizen. A lot of their APIs are the way they are due to a necessity, and I think it's better to have better/strongly typed support for it, WDYT?

packages/core/types/src/analytics/mutations.ts
sradevski: @olivermrbl I did it for two reasons:
1. Make it more flexible (so eg. adding actor-specific data is easier, such as eg. email)
2. Consistency with the group (they are both objects)

We could of course name it `actorId` and then, if we need more data, add an `actor` object later on without causing a breaking change, but I figured the difference is minimal. 

Which one would you prefer? I am happy to go either way

olivermrbl: I don't have a strong opinion on this, to be honest, but in my experience, the actor/user ID is typically top-level, and the actor's properties or traits are in a separate object. Now that I think of it, this is probably only relevant for the `identify` call since you wouldn't populate actor characteristics on the regular tracking event.

packages/core/types/src/analytics/mutations.ts
olivermrbl: Yeah, I see. Fine to keep this as is.

packages/core/types/src/analytics/mutations.ts
sradevski: @olivermrbl you can populate actor characteristics on the tracking event as well (we just don't currently), which is why I thought having it as an object would make that easier. Just make the final call and let's go with that 😄

sradevski:
@olivermrbl how do we go about the CLI tests? Do we just force merge so the package is released? I don't remember what was the process to add a new package 😄

packages/core/types/src/analytics/mutations.ts
olivermrbl: Let's do `actorId` if it's not too much work :)

olivermrbl:
Yes, that's what we typically do 👍

packages/core/types/src/analytics/mutations.ts
sradevski: Sure, not at all :D

sradevski:
@olivermrbl ready to re-review, if all good please go ahead and merge it!

olivermrbl (approved):
Nice, LGTM

olivermrbl:
I've just published the official packages now, so we should actually be able to get this to turn green before merging :)

PR #12508, opened 2025-05-16
chore(product, modules-sdk): Add missing index for select-in strategy + allow to pass top level strategy to force the behaviour

adrien2p:
**What**
- Add missing index for query that falls in select in strategy, specifically the heaviest one with variant IN filtering
- Allow to force the strategy to be sent to the entry point module when using the graph API. It is useful with big dataset where filtering is enough without pagination and select in can offer better performances

e.g
```ts
       await query.graph({
            entity: 'product',
            fields,
            filters,
            strategy: 'select-in' // <-- this will force the module to use receive this value and apply it accordingly 
        })
```


adrien2p:
virk will change the csv tests to not be strictly checking a snapshot since ordering is not set. Therefore it can break like in this pr when adding a new index

carlos-r-l-rodrigues (approved):

olivermrbl (approved):
LGTM

PR #12522, opened 2025-05-16
fix(dashboard): combobox initial item cache

fPolic:
**What**
- fix caching the initial combobox item under the same key which caused initial item not being shown properly
- fix loading of selected data - prevents issue when query results doesn't contain selected value
- fix infinite query response being cached under the same keys we store list responses - since the shapes don't match it can cause different issues
- prevent duplicate inventory items on variant kit edit

---

CLOSES SUP-1682


packages/admin/dashboard/src/hooks/use-combobox-data.tsx
```diff
@@ -53,8 +55,21 @@ export const useComboboxData = <
...
+  const { data: selectedData } = useQuery({
+    queryKey: [...queryKey, selectedValue].filter(Boolean) as QueryKey,
+    queryFn: async () => {
+      return queryFn({
+        id: selectedValue,
+        limit: 1,
+      } as TParams)
+    },
+    enabled: !!selectedValue && enabled,
+  })
```
olivermrbl, line 68: question: Just out of curiosity, what's the scenario where this happens?

olivermrbl (approved):
It looks good to me, but I don't know much about the combobox to know with certainty. Feel free to merge this if you have tested it across the dashboard.

packages/admin/dashboard/src/hooks/use-combobox-data.tsx
fPolic: for example, if the user selects an item that is not on the initial page of the paginated endpoint and closes the dropdown, that item won't be in the state anymore so input will appear empty

PR #12527, opened 2025-05-19
feat: process import from pre-processed chunks

thetutlage:
Fixes: FRMW-2968

In this PR we have done two major things.

- First, we remove storing CSV contents within the workflow storage and neither store the JSON payloads to be created/updated in workflows. Earlier, they all were workflow inputs, hence were stored in the workflow
- Introduce a naive concept of chunks and process chunks one by one. The next PR making chunking a bit more robust while using streams, adding ability to resume from the failed chunk and so on.

> [!IMPORTANT]  
> The new endpoint `/admin/product/imports` is not in use yet. But it will be after the next (final) PR.

## Old context in workflow storage

CleanShot 2025-05-19 at 17 11 08@2x

## New context in workflow storage

CleanShot 2025-05-19 at 17 15 08@2x


packages/core/core-flows/src/product/steps/normalize-products-v1.ts
```diff
@@ -0,0 +1,70 @@
...
+} from "@medusajs/framework/utils"
+import { StepResponse, createStep } from "@medusajs/framework/workflows-sdk"
+import { convertCsvToJson } from "../utils"
+
+/**
+ * The CSV file content to parse.
+ */
+export type NormalizeProductCsvV1StepInput = string
+
+export const normalizeCsvV1StepId = "normalize-product-csv-v1"
```
thetutlage: This might not be the best naming. So feel free to send some recommendations

packages/core/core-flows/src/product/steps/normalize-products-v1.ts
```diff
@@ -0,0 +1,70 @@
...
+      mimeType: "application/json",
+    })
+
+    return new StepResponse({
+      chunks: [id],
+      summary: {
+        toCreate: create.length,
+        toUpdate: update.length,
+      },
+    })
```
thetutlage: No more storing the CSV contents (as JSON) in workflow. Instead we store the chunk ids.

packages/core/core-flows/src/product/steps/process-import-chunks.ts
```diff
@@ -0,0 +1,30 @@
...
+    for (let chunk of input.chunks) {
+      const contents = await file.getAsBuffer(chunk)
+      console.log(`processing chunk ${chunk}`)
+
+      await batchProductsWorkflow(container).run({
+        input: JSON.parse(contents.toString("utf-8")),
+      })
+    }
+
+    return new StepResponse({ completed: true })
```
thetutlage, line 26: This step will return the progress of chunks in the next PR, so that we can resume from where things failed.

packages/medusa/src/api/admin/products/imports/route.ts
```diff
@@ -14,15 +13,14 @@ export const POST = async (
...
-    input: {
-      filename: req.validatedBody.originalname,
-      fileContent: file.toString("utf-8"),
-    },
-  })
+  const { result, transaction } = await importProductsV1Workflow(req.scope).run(
+    {
+      input: {
+        filename: req.validatedBody.originalname,
+        fileKey: req.validatedBody.file_key,
```
thetutlage: No longer storing the CSV in the workflow

adrien2p (approved):
LGTM, I suppose retention and checks will come in the follow up 👍

packages/core/core-flows/src/product/steps/normalize-products-v1.ts
olivermrbl: question: should we remove V1 from the name?

olivermrbl (approved):

packages/core/core-flows/src/product/steps/normalize-products-v1.ts
thetutlage: Then it will conflict with the existing step we have used by the current workflows. Lemme think of some better naming here.

PR #12529, opened 2025-05-19
chore(test-cli-with-database): Update cli tag from latest to preview

adrien2p:
**What**
Using the preview version instead of the latest should allow us to catch potential issues earlier before releasing


thetutlage (approved):

adrien2p:
@olivermrbl is that ok with you?

olivermrbl (approved):

PR #12538, opened 2025-05-20
fix(product): Deep update data retrieval bottleneck

adrien2p:
RESOLVES SUP-1663
**What**
- Fix not found product not managed in deep update, was silently doing nothing
- Fix deep update algorithm to benefit from separate queries, especially for large catalog without impacting smaller ones
- Only load relations that needs to be loaded to perform update
- Overall refactoring to remove redundant wrapping 

**NOTE**
On large catalog get the products update workflow down from +40s to <1s


packages/modules/product/src/repositories/product.ts
```diff
@@ -13,28 +17,78 @@ export class ProductRepository extends DALUtils.mikroOrmBaseRepositoryFactory(
     super(...arguments)
   }
 
+  /**
+   * Identify the relations to load for the given update.
+   * @param update
+   * @returns
+   */
+  getRelationsToLoad(update: any[]): string[] {
```
SimonGodefroid: `update` is singular but it's an array it's misleading.

packages/modules/product/src/repositories/product.ts
```diff
@@ -13,28 +17,78 @@ export class ProductRepository extends DALUtils.mikroOrmBaseRepositoryFactory(
...
+        relationsToLoad.add("options.values")
+      }
+      if (update.variants) {
+        relationsToLoad.add("options")
+        relationsToLoad.add("options.values")
+        relationsToLoad.add("variants")
+        relationsToLoad.add("variants.options")
+        relationsToLoad.add("variants.options.option")
+      }
+      if (update.tags) relationsToLoad.add("tags")
```
SimonGodefroid: Would it make sense to have an `update.meta` property that stores "options", "variants"... and segregate cases where `update.meta === 'variants'` and for the rest just add `update.meta` to the `relationsToLoad` set?

packages/modules/product/src/repositories/product.ts
SimonGodefroid: Also seems to be `updates` in other methods see [`deepUpdate`](https://github.com/medusajs/medusa/pull/12538/files#diff-9b1728f8225386d0357a222f3b4b9f5c8cccb8e3a27f72d569aaf4e104e19d26R48)

packages/modules/product/src/repositories/product.ts
adrien2p: just saw your comment, the pr is still in progress 😅 
This suggestion would require each product to be fetched independently for the relations to be determined by product to update right?

packages/modules/product/src/repositories/product.ts
adrien2p: ahah it was still in progress, I ve changed the naming completely as I prefer explicit naming personally

carlos-r-l-rodrigues (approved):

olivermrbl (approved):

PR #12539, opened 2025-05-20
fix: mark posthog-node as a peer dependency

thetutlage:
Since the runtime of the `@medusajs/analytics-posthog` relies on `posthog-node` package. It should be either installed as a dependency or a peerDependency that will be satisfied by the user project.

In this PR, I have added it as a peer dependency


sradevski (approved):
Ah I missed it, thanks Virk

adrien2p (approved):

PR #12540, opened 2025-05-20
fix(core-flows): skip location check if inventory is not managed

fPolic:
**What**
- skip location association check if variant is not managed when preparing inventory confirm

### DIFF
--- a/packages/core/core-flows/src/cart/utils/prepare-confirm-inventory-input.ts
+++ b/packages/core/core-flows/src/cart/utils/prepare-confirm-inventory-input.ts
@@ -100,7 +100,6 @@ export const prepareConfirmInventoryInput = (data: {
             0
         )
 
-        // TODO: check if this stock location is avaialable in the input.sales_channel_id
         if (!mapLocationAvailability.has(location_levels.location_id)) {
           mapLocationAvailability.set(location_levels.location_id, new Map())
         }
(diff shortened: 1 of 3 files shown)


adrien2p (approved):

adrien2p:
LGMT, dont forget the chageset

packages/core/core-flows/src/cart/utils/prepare-confirm-inventory-input.ts
```diff
@@ -100,6 +100,7 @@ export const prepareConfirmInventoryInput = (data: {
             0
         )
 
+        // TODO: check if this stock location is avaialable in the input.sales_channel_id
```
adrien2p: Q: This will be handled in a separate pr?

olivermrbl (approved):
LGTM

yanchesky:
Probably not a big deal but as I've done some research on that case too, maybe you will find my findings valuable.

In the `draft-order.spec.tsx` the second item is created with `manage_inventory: false` but it is also linked to the inventory item. That was the reason why the test passed without `prepareConfirmInventoryInput` throwing the error.

As I saw that you removed `inventory_items` from the unit tests of the `prepareConfirmInventoryInput`, maybe it should not be linked either.

https://github.com/medusajs/medusa/blob/1f5f50010ae42f8d5b9924edcd75612b1f336463/integration-tests/modules/__tests__/order/draft-order.spec.ts#L168C10-L175C13

PR #12546, opened 2025-05-20
fix(core-flows): reservation management on order edit and draft order confirm

fPolic:
**What**
- revisit reservations management when confirming draft order and order edit

**How**
- delete all reservations for removed or updated items
- recreate reservations for for new and updated items

### DIFF
--- a/packages/core/core-flows/src/draft-order/workflows/confirm-draft-order-edit.ts
+++ b/packages/core/core-flows/src/draft-order/workflows/confirm-draft-order-edit.ts
@@ -36,10 +36,10 @@ export interface ConfirmDraftOrderEditWorkflowInput {
 /**
  * This workflow confirms a draft order edit. It's used by the
  * [Confirm Draft Order Edit Admin API Route](https://docs.medusajs.com/api/admin#draft-orders_postdraftordersideditconfirm).
- * 
+ *
  * You can use this workflow within your customizations or your own custom workflows, allowing you to wrap custom logic around
  * confirming a draft order edit.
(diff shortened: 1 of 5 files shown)


packages/core/core-flows/src/draft-order/workflows/confirm-draft-order-edit.ts
```diff
@@ -134,31 +134,21 @@ export const confirmDraftOrderEditWorkflow = createWorkflow(
...
     }).config({ name: "order-items-query" })
 
-    const { removedLineItemIds } = transform(
-      { orderItems, previousOrderItems: order.items },
-      (data) => {
-        const previousItemIds = (data.previousOrderItems || []).map(
-          ({ id }) => id
-        )
-        const currentItemIds = data.orderItems.items.map(({ id }) => id)
+    const { variants, items, toRemoveReservationLineItemIds } = transform(
```
olivermrbl, line 137: thought: Can we consolidate these transforms? I believe we have almost the exact same code in many places, and I fear we can easily end up updating one but not the others.

olivermrbl: We can leave this for a separate PR, if we want to include this in the release today :)

fPolic: Sounds good, I can open a separate ticket.

olivermrbl (approved):
LGTM

PR #12547, opened 2025-05-20
chore(docs): Update npm command

### LINKED ISSUE #12523: Incorrect yarn run install in documentation, should be yarn install
### What Medusa version and documentation are you using?

v2

### Preliminary Checks

- [x] This issue is not a duplicate. Before opening a new issue, please search existing issues: https://github.com/medusajs/medusa/issues

### Issue Summary

In the docs, the installation step says:

`cd .medusa/server && yarn run install && yarn run start`

This leads to an error because install is a built-in yarn command and shouldn't be run with yarn run.

The correct command should be:

`cd .medusa/server && yarn install && yarn start`

Please fix this in the official documentation. Thanks!

`cd .medusa/server && yarn install && yarn start`

- [x] I would like to fix this issue

waynegeance:
`npm run install` in the first place is not a valid command, `yarn run install` will also be fixed by this change as we are using  **npm-to-yarn** package for its conversion (tested).

`yarn run start` is a valid command, i know `yarn start` is a **short-hand** but i cant help here as we have't hardcoded it in the docs its being converted through the same package mentioned above, anyways both are valid so should be fine ig.

### DIFF
--- a/www/apps/book/app/learn/deployment/general/page.mdx
+++ b/www/apps/book/app/learn/deployment/general/page.mdx
@@ -305,7 +305,7 @@ The Medusa application's production build, which is created using the `build` co
 If your hosting provider doesn't support setting a current-working directory, set the start command to the following:
 
 ```bash npm2yarn
-cd .medusa/server && npm run install && npm run start
+cd .medusa/server && npm install && npm run start
 ```
 
 ---


shahednasser (approved):
LGTM Thanks!

PR #12557, opened 2025-05-21
chore(index): Few adjustments

adrien2p:
**What**
- Adjust lock duration, it is in seconds and not in ms
- Log on lock release 
- Renew lock separately from the other promises
- Add more logs 
  - Log when lock can't be acquired. It can be expected in case two processes try to sync the same entity and in that case it can be ignored. But at least it gives some information in case it happens for another reason
  - Log when the release lock failed, the lock will remain in the locking provider for 1 minute before being removed. But it wont prevent other entities to be synced


adrien2p:
@olivermrbl it wont do any harm to get that in, but it does not solve the current investigation.

carlos-r-l-rodrigues (approved):

olivermrbl (approved):

Issue #12429, opened 2025-05-09
[Bug]: Medusa is unable to handle 100% discounts


xtul9:
### Package.json file

```JSON
{
  "name": "gmc-shop",
  "version": "0.0.1",
  "description": "A starter for Medusa projects.",
  "author": "Medusa (https://medusajs.com)",
  "license": "MIT",
  "keywords": [
    "sqlite",
    "postgres",
    "typescript",
    "ecommerce",
    "headless",
    "medusa"
  ],
  "scripts": {
    "build": "medusa build && ln -s .medusa/server/public/ public",
    "seed": "medusa exec ./src/scripts/seed.ts",
    "start": "medusa start",
    "dev": "medusa develop",
    "test:integration:http": "TEST_TYPE=integration:http NODE_OPTIONS=--experimental-vm-modules jest --silent=false --runInBand --forceExit",
    "test:integration:modules": "TEST_TYPE=integration:modules NODE_OPTIONS=--experimental-vm-modules jest --silent --runInBand --forceExit",
    "test:unit": "TEST_TYPE=unit NODE_OPTIONS=--experimental-vm-modules jest --silent --runInBand --forceExit"
  },
  "dependencies": {
    "@emotion/react": "^11.13.3",
    "@emotion/styled": "^11.13.0",
    "@mailerlite/mailerlite-nodejs": "^1.3.1",
    "@medusajs/admin-sdk": "2.6.1",
    "@medusajs/cli": "2.6.1",
    "@medusajs/framework": "2.6.1",
    "@medusajs/icons": "^2.1.3",
    "@medusajs/medusa": "2.6.1",
    "@mikro-orm/core": "6.4.3",
    "@mikro-orm/knex": "6.4.3",
    "@mikro-orm/migrations": "6.4.3",
    "@mikro-orm/postgresql": "6.4.3",
    "@mui/material": "^6.1.2",
    "awilix": "^8.0.1",
    "csv-writer": "^1.6.0",
    "mailersend": "^2.5.0",
    "multer": "^1.4.5-lts.1",
    "pdf-lib": "^1.17.1",
    "pdfkit": "^0.15.1",
    "pg": "^8.13.0",
    "react-hook-form": "^7.53.0",
    "react-windowed-select": "^5.2.0",
    "recharts": "^2.13.3",
    "stripe": "^18.0.0"
  },
  "devDependencies": {
    "@medusajs/test-utils": "2.6.1",
    "@mikro-orm/cli": "6.4.3",
[...]

Issue #12226, opened 2025-04-17
[Bug]: Tax inclusive issue with fixed promotion amount off order


pepijn-vanvlaanderen:
### Package.json file

```JSON
{
  "name": "my-medusa-store",
  "version": "0.0.1",
  "description": "A starter for Medusa projects.",
  "author": "Medusa (https://medusajs.com)",
  "license": "MIT",
  "keywords": [
    "sqlite",
    "postgres",
    "typescript",
    "ecommerce",
    "headless",
    "medusa"
  ],
  "scripts": {
    "build": "medusa build",
    "seed": "medusa exec ./src/scripts/seed.ts",
    "start": "medusa start",
    "dev": "medusa develop",
    "test:integration:http": "TEST_TYPE=integration:http NODE_OPTIONS=--experimental-vm-modules jest --silent=false --runInBand --forceExit",
    "test:integration:modules": "TEST_TYPE=integration:modules NODE_OPTIONS=--experimental-vm-modules jest --silent --runInBand --forceExit",
    "test:unit": "TEST_TYPE=unit NODE_OPTIONS=--experimental-vm-modules jest --silent --runInBand --forceExit"
  },
  "dependencies": {
    "@medusajs/admin-sdk": "2.7.0",
    "@medusajs/cli": "2.7.0",
    "@medusajs/framework": "2.7.0",
    "@medusajs/medusa": "2.7.0",
    "@mikro-orm/core": "6.4.3",
    "@mikro-orm/knex": "6.4.3",
    "@mikro-orm/migrations": "6.4.3",
    "@mikro-orm/postgresql": "6.4.3",
    "awilix": "^8.0.1",
    "pg": "^8.13.0"
  },
  "devDependencies": {
    "@faker-js/faker": "^9.7.0",
    "@medusajs/test-utils": "2.7.0",
    "@mikro-orm/cli": "6.4.3",
    "@swc/core": "1.5.7",
    "@swc/jest": "^0.2.36",
    "@types/jest": "^29.5.13",
    "@types/node": "^20.0.0",
    "@types/react": "^18.3.2",
    "@types/react-dom": "^18.2.25",
    "jest": "^29.7.0",
    "prop-types": "^15.8.1",
    "react": "^18.2.0",
    "react-dom": "^18.2.0",
    "ts-node": "^10.9.2",
    "typescript": "^5.6.2",
    "vite": "^5.2.11",
    "yalc": "^1.0.0-pre.53"
  },
  "engines": {
[...]

PR #12571, opened 2025-05-21
fix: rebuilt parts of the discount/adjustment/subtotal logic with refactored isTaxInclusive logic

scherddel:
# Fix for Promotion Calculation Issues with Tax-Inclusive Products

## Issue Description
I encountered several issues with promotion logic (especially for tax-inclusive products) where incorrect discount totals were being generated. These issues have been reported by multiple users:
- https://github.com/medusajs/medusa/issues/12429
- https://github.com/medusajs/medusa/issues/12226

## Investigation
After extensive debugging, I initially looked at @fPolic's PR approach but discovered that the issue is more fundamental - promotions shouldn't be directly related to a tax-inclusiveness field in the promotion. 

The root cause is in the discount calculation logic, where:
- The wrong price elements were being used as inputs
- This error propagated throughout the calculation chain
- Ultimately affected both cart totals and order totals

Edit: Ohhh the issue was nested much deeper. I realized that many calculations for taxInclusiveness were incorrect (I am askin myself how people have this productive with the issues I had to correct) as well as the calculations of subtotals between items and the cart which I had to correct first which then led to different adjustments in the discount logic.

But now it should all work fine and promotions will consider tax_inclusivess as well without an extra field to maintain

## Testing
I've verified this fix works with:
- Regular products
- Tax-inclusive products
- Various promotion types (percentage, fixed amount, cross, each)

This fix should resolve discount calculation issues for everyone using tax-inclusive products. 

**Given how fundamental this issue is to correct pricing, I hope it can be implemented soon.**

@riqwan I see you as the promotion expert, maybe this is something you can also have a look on.


packages/core/utils/src/totals/line-item/index.ts
```diff
@@ -124,7 +125,7 @@ function getLineItemTotals(
     Original Price = Total Price / (1 + Tax Rate)
   */
   const subtotal = isTaxInclusive
-    ? MathBN.div(totalItemPrice, MathBN.add(1, sumTaxRate))
+    ? MathBN.mult(totalItemPrice, MathBN.div(100, MathBN.add(100, sumTax)))
     : totalItemPrice
```
srindom: question: can you please explain the change here. 

The subtotal should always be a value without tax.

scherddel: Edit:  @srindom I think i missread your statement and then replied. Sorry. It was more a explanation on why I thought all subsequent changes where necessary and its absolutely right what you say.

...

The subtotal hast to be differently calculated if you consider the taxInclusive context. When your items unit price is X then the subtotal will change based on the perspective of tax inclusion:

Here X is the items unit price that will be handled through by tax inclusion rules (and we take a tax of 19%):
**Tax Inclusive (Gross Price)**
- **Total**: X
- **Tax**: X × (19/119)
- **Subtotal**: X × (100/119)

**Non-Tax Inclusive (Net Price)**  
- **Subtotal**: X
- **Tax**: X × 0.19
- **Total**: X × 1.19

At least this is my understanding how tax works ;-) And this should also be the logic throughout the whole tax calculation for each subtotal, total and tax (i mean in the cart as well)

scherddel: Just to provide some official context: https://www.ldoceonline.com/Business%2Bbasics-topic/subtotal

```
the total of a set of numbers, especially on a bill, that is added to other numbers, such as a tax, to form a complete total
```

scherddel: But to your question: the change on line 128 was not really needed as the formulars are the same but it helped me to streamline the whole calculation adjustments.

I just dont fully get your concern "The subtotal should always be a value without tax." as both formulars are exactly in line with the subtotal definition (at least in these lines)

scherddel:
This took longer than expected due to several complex issues I discovered during implementation:

**Key Issues Identified:**
1. **Promo code removal bug**: Adding a promo code that reduced an item's subtotal to $0 would unexpectedly remove the promo code entirely. This shouldn't happen, as reducing items to $0 may be the intended purpose of certain promotions.

2. **Flawed promotion calculation basis**: Promotions were incorrectly based on item subtotals, which creates problems when the same promo code is already applied to the cart. This circular logic led to incorrect calculations.

3. **Adjustment recalculation issues**: The existing logic deletes all applied promotions first, then recalculates. However, since subtotals remain unchanged (with adjustments still applied), the system needs to use Unit Price as the basis for adjustment calculations to properly handle recalculations.

**Changes Made:**
- Added tax inclusion logic to adjustments, as return values vary based on tax context
- Added missing subtotal calculations where they were absent
- Fixed taxable amount calculations, which differ based on tax inclusion context
- Corrected the promotion calculation flow to use proper price bases

**Testing:**
I've tested this extensively across multiple scenarios and the calculations now work correctly for both tax-inclusive and tax-exclusive products with various promotion types.

www/apps/resources/app/commerce-modules/cart/tax-lines/page.mdx
```diff
@@ -30,7 +30,11 @@ The following diagram is a simplified showcase of how a subtotal is calculated f
 
 ![A diagram showing an example of calculating the subtotal of a line item using its taxes](https://res.cloudinary.com/dza7lstvk/image/upload/v1711535295/Medusa%20Resources/cart-tax-inclusive_shpr3t.jpg)
```
scherddel, line 36: I dont know how to change the picture but it also shows a wrong definition as it should show the values as described in the text below.

PR #12572, opened 2025-05-22
fix(core-flows, dashboard, types): improve allocation flows on Admin

fPolic:
**What**
- handle single inventory items with `required_quantity > 1` correctly in the allocate UI flow
- display correct availability amount in the create fulullment form for inventory kit items
- fix check in the create fulfillment flow that would allow negative reservation because `required_quantity` wasn't included in the check
- show the most recent reservations first in the reservations table
- display an error in the allocation form if a reservation is not created for some inventory items 
- display inventory kit in the order summary if the product has multiple same inventory items

---

CLOSES SUP-1655


olivermrbl (commented):
Nice, let's add some tests for the fix, so we know it's working as expected and won't break in the future.

packages/admin/dashboard/src/routes/orders/order-allocate-items/components/order-create-fulfillment-form/order-allocate-items-form.tsx
```diff
@@ -81,14 +82,19 @@ export function OrderAllocateItemsForm({ order }: OrderAllocateItemsFormProps) {
         return
       }
 
-      const promises = payload.map(([itemId, inventoryId, quantity]) =>
+      const promises = payload.map(([itemId, inventoryId, quantity]) => {
+        const item = itemsToAllocate.find((i) => i.id === itemId)
+        const variantInventoryItem = item?.variant?.inventory_items?.find(
+          (i) => i.inventory_item_id === inventoryId
+        )
```
olivermrbl: question: I assume these are guaranteed to be present? If not, I think we should add a safe-guard returning early.

packages/admin/dashboard/src/routes/orders/order-allocate-items/components/order-create-fulfillment-form/utls.ts
```diff
@@ -0,0 +1,28 @@
...
+  const variant = item.variant
+
+  if (!variant) {
+    return false
+  }
+
+  return (
+    (!!variant.inventory_items.length && variant.inventory_items.length > 1) ||
+    (variant.inventory_items.length === 1 &&
+      variant.inventory_items[0].required_quantity! > 1)
```
olivermrbl: question: Might be missing an obvious detail, but why is the second condition deemed an inventory kit?

packages/admin/dashboard/src/routes/orders/order-create-fulfillment/components/order-create-fulfillment-form/order-create-fulfillment-item.tsx
```diff
@@ -35,36 +35,91 @@ export function OrderCreateFulfillmentItem({
...
     {
       enabled: !!item.variant,
     }
   )
 
   const { availableQuantity, inStockQuantity } = useMemo(() => {
-    if (!variant || !locationId) {
+    if (
+      !variant?.inventory_items?.length ||
+      !variant?.inventory?.length ||
```
olivermrbl, line 48: question: Is the second one not an alias of the first one?

fPolic: `inventory` is an array of associated `InventoryItem` entities and `inventory_item` contains an array of `ProductVariantInventoyItem` entities which conatin `required_quantity` etc.

The lengths of the arrays should be the same but the check is for redundancy.

packages/admin/dashboard/src/routes/orders/order-allocate-items/components/order-create-fulfillment-form/utls.ts
fPolic: If there is a single inventory item for the variant, but it requires multiple items, isn't this just a special case of an inventory kit? For example, a 3-pack bottle variant that only has one inventory item but needs 3 of them for fulfillment. 
Handling that as a special inventory kit case simplifies UI implementation and ensures quantities take `required_quantity` into consideration.

olivermrbl: I was missing a detail; this makes sense. Thanks 😄

packages/admin/dashboard/src/routes/orders/order-allocate-items/components/order-create-fulfillment-form/utls.ts
```diff
@@ -0,0 +1,28 @@
+import {
```
olivermrbl: todo: typo in file name

packages/admin/dashboard/src/routes/orders/order-create-fulfillment/components/order-create-fulfillment-form/order-create-fulfillment-item.tsx
```diff
@@ -35,36 +35,91 @@ export function OrderCreateFulfillmentItem({
...
+    }
 
     return {
-      availableQuantity:
-        locationInventory.available_quantity + reservedQuantityForItem,
-      inStockQuantity: locationInventory.stocked_quantity,
+      availableQuantity: Math.floor(
+        maxAvailableQuantity + reservedQuantityForItem
+      ),
+      inStockQuantity: Math.floor(maxStockedQuantity),
```
olivermrbl, line 120: q: what's the _flooring_ for here?

olivermrbl (approved):

packages/admin/dashboard/src/routes/orders/order-create-fulfillment/components/order-create-fulfillment-form/order-create-fulfillment-item.tsx
fPolic: If we have an inventory kit product which requires 3x InventoryItemA and 1x InventoryItemB, if there are 10 InventoryItemA in stock, we can fulfilll max of `floor (10 / 3)` -> 3 products

PR #12573, opened 2025-05-22
fix: Expose complete cart hooks

### LINKED ISSUE #12456: [Bug/Question]: not exposed hooks ?
### Package.json file

```JSON
-
```

### Node.js version

-

### Database and its version

-

-

### Browser name

_No response_

### What happended?

```ts
// completeCartWorkflow
// ...
      createHook("orderCreated", {
        order_id: createdOrder.id,
        cart_id: cart.id,
      })
// ...
```

I've seen `createHook` calls like these. What is the purpose of them when they aren't exposed in the end anyway? The `completeCartWorkflow` exposes a `validate` hook. So `const validate = createHook(..` is at the end of the workflow exposed, e.g.
```ts
    return new WorkflowResponse(createdOrder, {
      hooks: [validate], // why not [validate, orderCreated, there are more..] ?
    })
```

So is there something I don't see or are these createHook calls unused leftovers? If so I'd consider them as bugs because apparently they need to be there but are not usable right now?

I am currently looking for a hook I can use after a order with applied promotions is completed from my store / cart.

Love,
Y

### Expected behavior

Created hooks should be accessible.

### Actual behavior

Created hooks look abandoned.

### Link to reproduction repo

-

sradevski:
FIXES #12456

One issue is, the hooks get called even if we don't enter in the `when/then` condition, but without data.

### DIFF
--- a/integration-tests/modules/__tests__/cart/store/cart.completion.ts
+++ b/integration-tests/modules/__tests__/cart/store/cart.completion.ts
@@ -226,6 +226,13 @@ medusaIntegrationTestRunner({
             },
           })
 
+          completeCartWorkflow.hooks.beforePaymentAuthorization((input) => {
+            expect(input.input.id).toBeDefined()
+          })
+          completeCartWorkflow.hooks.orderCreated((input) => {
+            expect(input.order_id).toBeDefined()
+          })
+
           await completeCartWorkflow(appContainer).run({
             input: {
               id: cart.id,
--- a/packages/core/core-flows/src/cart/workflows/complete-cart.ts
+++ b/packages/core/core-flows/src/cart/workflows/complete-cart.ts
@@ -122,15 +122,19 @@ export const completeCartWorkflow = createWorkflow(
       payment_session_id: paymentSessions[0].id,
     })
 
-    const validate = createHook("validate", {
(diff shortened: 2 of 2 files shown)


integration-tests/modules/__tests__/cart/store/cart.completion.ts
```diff
@@ -226,6 +226,13 @@ medusaIntegrationTestRunner({
             },
           })
 
+          completeCartWorkflow.hooks.beforePaymentAuthorization((input) => {
+            expect(input.input.id).toBeDefined()
+          })
+          completeCartWorkflow.hooks.orderCreated((input) => {
+            expect(input.order_id).toBeDefined()
+          })
+
```
thetutlage, line 235: Can we test if these methods are undefined when `orderId` does not exist?

sradevski: @thetutlage yeah I wanted to discuss in Slack first, but they are defined, and they execute even if we don't enter in the when with data missing (i.e. `input.input.id` is defined, but `input.order_id` isn't) .

olivermrbl:
Closing this as we don't intend to expose these hooks for the time being.

PR #12574, opened 2025-05-22
feat: implement stream based processing of the files

thetutlage:
Fixes: FRMW-2960

This PR adds support for processing large CSV files by breaking them into chunks and processing one chunk at a time. This is how it works in nutshell.

- The CSV file is read as a stream and each chunk of the stream is one CSV row.
- We read upto 1000 rows (plus a few more to ensure product variants of a product are not split into multiple chunks).
- Each chunk is then normalized using the `CSVNormalizer` and validated using zod schemas. If there is an error, the entire process will be aborted and the existing chunks will be deleted.
- Each chunk is written to a JSON file, so that we can process them later (after user confirms) without re-processing or validating the CSV file.
- The confirmation process will start consuming one chunk at a time and create/update products using the `batchProducts` workflow.

## Resume or not to resume processing of chunks

Let's imagine during processing of chunks, we find that chunk 3 leads to a database error. However, till this time we have processed the first two chunks already. How do we deal with this situation? Options are:

- We store at which chunk we failed and then during the re-upload we ignore chunks before the failed one. In my conversation with @olivermrbl we discovered that resuming will have to work with certain assumptions if we decide to implement it.
   - What if a user updates the CSV rows which are part of the already processed chunks? These changes will be ignored and they will never notice it.
   - Resuming works if the file name is still the same. What if they made changes and saved the file with "Save as - New name". In that case we will anyways process the entire file.
   - We will have to fetch the old workflow from the workflow engine using some `ilike` search, so that we can see at which chunk the last run failed for the given file.


packages/core/utils/src/product/__tests__/csv-normalizer.spec.ts
```diff
@@ -10,16 +10,16 @@ async function loadFixtureFile(fileName: string) {
 
 describe("CSV processor", () => {
   it("should error when both Product Id and Handle are missing", async () => {
-    const processor = new CSVNormalizer([{}])
-
-    expect(() => processor.proccess()).toThrow(
+    expect(() => CSVNormalizer.preProcess({}, 1)).toThrow(
       "Row 1: Missing product id and handle. One of them are required to process the row"
```
adrien2p:
```suggestion
      "Row 1: Missing product id and handle. One of these columns are required to process the row"
```

adrien2p (approved):
Lovely 🎉

carlos-r-l-rodrigues (approved):
LGTM

PR #12590, opened 2025-05-22
Fix: remove "Province of China" from Taiwan country name

ccwei:
Taiwan is not a province of China so the current naming is incorrect. Update the name to keep Taiwan only.

The U.S. House of Representatives also recently started sending letters to ask Hotels to correct the naming, see the following info.

https://selectcommitteeontheccp.house.gov/sites/evo-subsites/selectcommitteeontheccp.house.gov/files/evo-media-document/Letter%20-%20U.S.%20Hotels%20Labeling%20of%20Taiwan.pdf

https://selectcommitteeontheccp.house.gov/media/press-releases/moolenaar-blackburn-hotel-giants-stop-pushing-communist-chinas-talking-points#:~:text=WASHINGTON%2C%20D.C.%20%E2%80%93%20Chairman%20John%20Moolenaar,their%20websites%20and%20promotional%20materials.


ccwei:
@sradevski @riqwan @kasperkristensen could you help taking a quick look? It's a simple text change.

Thanks!

PR #12612, opened 2025-05-26
fix: expand method ALL for bodyparser config and additional validator

thetutlage:
Fixes: FRMW-2969

### DIFF
--- a/packages/core/framework/src/http/__tests__/middleware-file-loader.spec.ts
+++ b/packages/core/framework/src/http/__tests__/middleware-file-loader.spec.ts
@@ -7,8 +7,24 @@ describe("Middleware file loader", () => {
     const loader = new MiddlewareFileLoader()
     await loader.scanDir(BASE_DIR)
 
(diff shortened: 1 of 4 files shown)


packages/core/framework/src/http/__tests__/middleware-file-loader.spec.ts
```diff
@@ -7,8 +7,24 @@ describe("Middleware file loader", () => {
     const loader = new MiddlewareFileLoader()
     await loader.scanDir(BASE_DIR)
 
+    console.log(loader.getBodyParserConfigRoutes())
```
sradevski: This is probably not needed?

sradevski (approved):
Nice 👍

packages/core/framework/src/http/__tests__/middleware-file-loader.spec.ts
thetutlage: Ahhh! My bad

PR #12614, opened 2025-05-26
feat: introduce bulkDelete method for IFileProvider

thetutlage:
Fixes: FRMW-2974

Currently during the product imports, we create multiple chunks that must be deleted after the import has finished (either successfully or with an error). Deleting files one by one leads to multiple network calls and slows down everything.

The `bulkDelete` method deletes multiple files (with their fileKey) in one go


sgirones (approved):

packages/core/types/src/file/provider.ts
```diff
@@ -128,6 +129,15 @@ export interface IFileProvider {
    */
   delete(fileData: ProviderDeleteFileDTO): Promise<void>
```
sradevski: thought/suggestion: Should we, instead, have the provider `delete` method accept `ProviderDeleteFileDTO | ProviderDeleteFileDTO[]` instead of adding a new method?

sradevski (approved):
Just a suggestion on the provider's interface, otherwise LGTM

packages/core/types/src/file/provider.ts
olivermrbl: +1, I also think this is a bit more in line with how we usually define the public service interfaces

olivermrbl (approved):

packages/core/types/src/file/provider.ts
thetutlage: Sure. Let's do that!

PR #12620, opened 2025-05-26
docs: add missing equal sign for DATABASE_URL in env example

Amar2Kk:
The DATABASE_URL variable in the .env file examples was missing an equal sign, preventing it from parsing correctly. This fix ensures the environment variable is properly loaded and available during application runtime.


shahednasser (approved):
Nice catch thanks!

PR #12642, opened 2025-05-29
fix(analytics): fix type of provider option

shahednasser:
Use `providers` instead of `provider` and change the value to an array


carlos-r-l-rodrigues (approved):

olivermrbl (approved):
LGTM

olivermrbl:
Always get a bit scared when I see this haha 😄 

CleanShot 2025-05-29 at 10 41 52@2x

PR #12643, opened 2025-05-29
feat: wire up direct uploads with local file provider

thetutlage:


packages/admin/dashboard/src/hooks/api/products.tsx
```diff
@@ -400,7 +400,7 @@ export const useImportProducts = (
   >
 ) => {
   return useMutation({
-    mutationFn: (payload) => sdk.admin.product.import(payload),
+    mutationFn: (payload) => sdk.admin.product.createImport(payload),
```
thetutlage, line 403: Instructing frontend to use the new endpoint that uses direct file uploads

packages/core/core-flows/src/product/steps/normalize-products-to-chunks.ts
```diff
@@ -192,36 +194,52 @@ async function createChunks(
...
 
-    const summary = chunks.reduce<{ toCreate: number; toUpdate: number }>(
-      (result, chunk) => {
-        result.toCreate = result.toCreate + chunk.toCreate
-        result.toUpdate = result.toUpdate + chunk.toUpdate
-        return result
-      },
-      { toCreate: 0, toUpdate: 0 }
-    )
+        contents.on("error", reject)
```
thetutlage, line 211: Handling stream errors and rejecting the promise

packages/core/core-flows/src/product/steps/process-import-chunks.ts
```diff
@@ -12,7 +12,10 @@ export const processImportChunksStepId = "process-import-chunks"
  * const data = parseProductCsvStep("products.csv")
  */
 export const processImportChunksStep = createStep(
-  processImportChunksStepId,
+  {
+    name: processImportChunksStepId,
+    async: true,
```
thetutlage, line 17: If I do not convert this step to `async`, then the imports confirm endpoint waits until this step is done (which takes minutes with larger uploads).

@carlos-r-l-rodrigues can you confirm if my assumption is correct?

packages/core/js-sdk/src/admin/product.ts
```diff
@@ -114,10 +114,40 @@ export class Product {
...
+        headers: {
+          ...headers,
+          // Let the browser determine the content type.
+          "content-type": null,
+        },
+        body: form,
+        query,
+      })
+
+      response.filename = localUploadResponse.files[0].id
```
thetutlage, line 149: If we are using local file provider to uploading a file, we have to handle it specially in the frontend.

- In case of s3, we have to send a `PUT` request. Whereas, with Medusa we do not implement `PUT` endpoints.
- In case of s3, we cannot send additional headers like `Cookies` or `Authorization`, otherwise it will result in CORS error. Whereas, with Medusa, we have to send Auth headers.

packages/core/core-flows/src/product/steps/process-import-chunks.ts
carlos-r-l-rodrigues: When composing the workflow, as the step to wait for confirmation is async, this workflow will be automatically flagged as async.
so, here the `async` flag is just optional.

thetutlage: Okay. But I noticed that without this flag, the endpoint waits for the workflow to finish. Maybe what I need is `background` execution

thetutlage: Btw, I have been following this guide https://docs.medusajs.com/learn/fundamentals/workflows/long-running-workflow

carlos-r-l-rodrigues (approved):

packages/core/core-flows/src/product/steps/process-import-chunks.ts
carlos-r-l-rodrigues: > Okay. But I noticed that without this flag, the endpoint waits for the workflow to finish. Maybe what I need is `background` execution

This shouldn't happen. When there is one async step, the worklfow.run will return as soon as it get to this step. When that step is finished, it will continue in another process.

integration-tests/http/__tests__/product/admin/product-imports.spec.ts
```diff
@@ -605,6 +605,22 @@ medusaIntegrationTestRunner({
...
+          name: "test.csv",
+          key: "test.csv",
+          size: 0,
+        })
+
+        const batchJobRes = await api
+          .post("/admin/products/imports", body, meta)
+          .catch((e) => e)
+
+        expect(batchJobRes.response.data.message).toEqual(
```
adrien2p, line 620: suggestion: would it be possible to have a comprehensive error message here?

packages/core/core-flows/src/product/steps/process-import-chunks.ts
adrien2p: programatically speaking, we do not await when we execute an async step. if you want we can take a look together to be sure of what is happening

adrien2p (approved):

PR #12654, opened 2025-06-01
fix(docs): correct module reference in useQueryGraphStep tip

Amar2Kk:
Corrected an inaccurate example in the useQueryGraphStep tip within the review workflow documentation.

Updated the entity reference from "cart's promotions" to "product" to align with the actual code context.


shahednasser (approved):
Nice catch thanks!

PR #12675, opened 2025-06-06
fix(workflow-sdk): Async/nested runAsStep propagation

adrien2p:
FIXES CLO-524

**What**
Add hidden stepDefinition object as part of the step argument and ensure the runAsStep handlers rely on the latest definition when config is being used on the returned step in order to ensure async configuration propagation and nested configuration


packages/core/workflows-sdk/src/utils/composer/create-step.ts
```diff
@@ -173,6 +173,10 @@ export function applyStep<
         ...localConfig,
       }
 
+      if (isDefined(newConfig.nested)) {
+        newConfig.nested ||= newConfig.async
```
sradevski, line 177: Is there any other config here that we'd like to store?

sradevski (approved):
Nice 👍

packages/core/workflows-sdk/src/utils/composer/create-step.ts
adrien2p: not that are calculated, I think we are not missing anymore 😅 the rest are stored when passed and are part of newConfig directly 👍

adrien2p: I have also added some tests to validate the propagation of async and nested

adrien2p:
@sradevski I am generating a new snapshot right now with the latest changes 👍

sradevski:
@adrien2p looks good, and all tests are passing, so all good from my side!

PR #12684, opened 2025-06-09
fix: keep enum values in types generation

peterlgh7:
An idea for solving the issue of enum values being changed when the GraphQL types are generated.

Two-step approach:
1. Add a custom directive to enum values when GraphQL schema is generated. The directive contains the enum value.
2. Read that custom directive when generating the types from GraphQL

The approach could be extended to also keep enum keys, but I wonder how useful that really is, other than for aesthetic purposes, since the generated type and the original one won't be compatible anyway, despite being identical.

Another question is whether we want to treat actual enums differently from union types, since the DML allows to specify enums either way. We could e.g. keep union types as union types when we generate the types, but that's a more breaking change that I didn't want to tackle here.

Then on a separate note (unrelated to this PR), I noticed we have a failing case when generated enum names overlap. E.g. take DML `.enum("new-user", "new_user")`. The generated GraphQL is:
```
enum Role {
  NEW_USER
  NEW_USER
}
```
since all non-alphanumeric characters are converted to _. Here the GraphQL parsing fails because of the duplicate. One easy workaround I can think of is to hex-encode enum values if there are duplicates. Not pretty, but should work fine, and there's no impact on the final generated types anyway.

### DIFF
--- a/packages/core/modules-sdk/src/medusa-app.ts
+++ b/packages/core/modules-sdk/src/medusa-app.ts
@@ -234,6 +234,7 @@ function cleanAndMergeSchema(loadedSchema) {
   const defaultMedusaSchema = `
     scalar DateTime
     scalar JSON
+    directive @enumValue(value: String) on ENUM_VALUE
   `
   const { schema: cleanedSchema, notFound } = GraphQLUtils.cleanGraphQLSchema(
     defaultMedusaSchema + loadedSchema
--- a/packages/core/modules-sdk/src/medusa-app.ts
+++ b/packages/core/modules-sdk/src/medusa-app.ts
@@ -231,8 +231,14 @@ function isMedusaModule(mod) {
 }
 
 function cleanAndMergeSchema(loadedSchema) {
-  const { schema: cleanedSchema, notFound } =
-    GraphQLUtils.cleanGraphQLSchema(loadedSchema)
+  const defaultMedusaSchema = `
+    scalar DateTime
+    scalar JSON
+    directive @enumValue(value: String) on ENUM_VALUE
+  `
+  const { schema: cleanedSchema, notFound } = GraphQLUtils.cleanGraphQLSchema(
+    defaultMedusaSchema + loadedSchema
+  )
   const mergedSchema = GraphQLUtils.mergeTypeDefs(cleanedSchema)
   return {
     schema: GraphQLUtils.makeExecutableSchema({ typeDefs: mergedSchema }),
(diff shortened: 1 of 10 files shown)


adrien2p (approved):
LGTM, I think the union types only would make more sense to prevent types incompatibility and especially that it would still enforce (from a type perspective) the values being passed. The only down side (which is acceptable) is that enum are translated to js and allow to centralize usage allowing changes in one place to be applied everywhere. cc @thetutlage

thetutlage:
Just to make sure I understand the problem we are trying to solve. Let's say we have this enum used inside the product model.

```ts
export enum ProductStatus {
  DRAFT = "draft",
  PROPOSED = "proposed",
  PUBLISHED = "published",
  REJECTED = "rejected",
}

model.define('Product', {
  status: model.enum(ProductStatus)
})
```

What should be its output in the generated types written inside the `.medusa/types/query-entry-points.d.ts`?

adrien2p:
> Just to make sure I understand the problem we are trying to solve. Let's say we have this enum used inside the product model.
> 
> ```ts
> export enum ProductStatus {
>   DRAFT = "draft",
>   PROPOSED = "proposed",
>   PUBLISHED = "published",
>   REJECTED = "rejected",
> }
> 
> model.define('Product', {
>   status: model.enum(ProductStatus)
> })
> ```
> 
> What should be its output in the generated types written inside the `.medusa/types/query-entry-points.d.ts`?

I would something something like 
`type ProductStatus = ‘draft’ | ‘proposed’ | …`

in that case both strings and enum values should satisfy the type right?

peterlgh7:
@thetutlage before this PR, the output would be:
```
export enum ProductStatusEnum {
  DRAFT = "DRAFT",
  PROPOSED = "PROPOSED",
  PUBLISHED = "PUBLISHED",
  REJECTED = "REJECTED",
}
```
After this PR:
```
export enum ProductStatusEnum {
  DRAFT = "draft",
  PROPOSED = "proposed",
  PUBLISHED = "published",
  REJECTED = "rejected",
}
```
The main problem solved here is that the DML values are assignable to the GraphQL model and backwards. 

Whether to switch to using enum types is another question I raise in the description. I'd lean towards always generating union types, which @adrien2p seems to be suggesting as well. I can implement that too. I didn't do it because it was a bigger change and needed alignment.

thetutlage:
@peterlgh7 I think let's go with generating union types, because enums are not really useful in the generated types output.

peterlgh7:
@thetutlage finishing some cloud stuff now, but will add the union types thing today

peterlgh7:
@thetutlage I've pushed the union types change now, see the last commit. Are we ok to merge after that?

PR #12694, opened 2025-06-10
feat(core-flows,types): pass cart in create payment session context

shahednasser:
Some payment providers requires details related to the cart, such as its items and totals. This PR passes the cart information in the create payment session context


packages/core/core-flows/src/payment-collection/workflows/create-payment-session.ts
```diff
@@ -77,7 +77,26 @@ export const createPaymentSessionsWorkflow = createWorkflow(
   ): WorkflowResponse<PaymentSessionDTO> => {
     const paymentCollection = useRemoteQueryStep({
       entry_point: "payment_collection",
-      fields: ["id", "amount", "currency_code", "payment_sessions.*"],
+      fields: [
```
adrien2p, line 80: suggestion: we could add tests in `integration-tests/modules/__tests__/payment/payment-session.workflows.spec.ts` just to ensure the data are correctly pass and we don't break it later, wdyt?

packages/core/core-flows/src/payment-collection/workflows/create-payment-session.ts
```diff
@@ -77,7 +77,26 @@ export const createPaymentSessionsWorkflow = createWorkflow(
...
+        "id", 
+        "amount", 
+        "currency_code", 
+        "payment_sessions.*", 
+        "cart.id",
+        "cart.currency_code",
+        "cart.items.*",
```
adrien2p, line 87: Q: should items totals be present as well?

olivermrbl (approved):

packages/core/core-flows/src/payment-collection/workflows/create-payment-session.ts
```diff
@@ -77,7 +77,28 @@ export const createPaymentSessionsWorkflow = createWorkflow(
   ): WorkflowResponse<PaymentSessionDTO> => {
     const paymentCollection = useRemoteQueryStep({
       entry_point: "payment_collection",
-      fields: ["id", "amount", "currency_code", "payment_sessions.*"],
+      fields: [
+        "id", 
+        "amount", 
+        "currency_code", 
+        "payment_sessions.*", 
+        "cart.id",
```
olivermrbl, line 85: thought: If I am not mistaken, this workflow can also be used to create payment sessions for orders. I wonder if we should reconsider the approach here to account for this?

One idea that comes to mind would be to:
1. Allow passing `cart_id` and `order_id` to the workflow
2. Fetch the data based on which of these was passed
3. Pass `cart_or_order` in the context passed to the provider

Not loving it, but this was the first approach that came to mind, so mainly sharing to start the discussion.

shahednasser: I think that makes sense, but maybe we can have separate `cart` and `order` properties in the context? They're both optional, so we don't guarantee they're passed anyway and that would allow us to account for other use cases in the future.

shahednasser:
@olivermrbl let me know what we should do about this one

NicolasGorga (approved):
LGTM

langovoi:
@shahednasser any reasons why this PR was closed?

PR #12704, opened 2025-06-11
docs: add missing export for CreateBundledProduct component

Amar2Kk:
This commit adds a missing `export` statement to the `CreateBundledProduct` component in the "Create Form Component" example.

Without the `export`, readers copying the example wouldn't be able to use the component elsewhere in their project, which could be confusing. This ensures the example is complete and ready to use as-is.


shahednasser (changes requested):
Nice catch! but can you instead change to add the following line after the component?

```
export default CreateBundledProduct
```

shahednasser (approved):
LGTM Thanks!

PR #12714, opened 2025-06-12
fix(create-medusa-app): ensure the same package manager is used consistently

shahednasser:
Ensure that we select a package manager to choose early on in the installation and use it consistently.


juanzgc (approved):
LGTM! Confirmed defaults to using `yarn` if exists, otherwise uses `npm`.

PR #12715, opened 2025-06-12
fix(utils): medusa internal service returned data should match typings

adrien2p:
FIXES SUP-1824

**What**
The medusa internal service update should always return the data in the expected shape described by the interface. The medusa service should not have to handle the reshapre


thetutlage (approved):

packages/core/utils/src/modules-sdk/medusa-service.ts
```diff
@@ -245,12 +245,7 @@ export function MedusaService<
         ): Promise<T | T[]> {
           const serviceData = Array.isArray(data) ? data : [data]
           const service = this.__container__[serviceRegistrationName]
-          const models = await service.update(serviceData, sharedContext)
-          const response = models.length
-            ? Array.isArray(data)
-              ? models
-              : models[0]
-            : []
+          const response = await service.update(serviceData, sharedContext)
```
olivermrbl, line 248: thought: I guess this is technically a breaking change? All service calls that have used the selector-based update will now return an array right?

adrien2p: yes thats right, but it was actually a bug, so I consider it more as a fix than a breaking change. The types was describing that you would receive an array but in fact you were receiving an object. Also, this logic is handled in the internal service

adrien2p: if you push further down, the result was different if you call the medusa service or the medusa internal service for the same input except that we were describing the same result

olivermrbl: Yeah, that's true. Think it should be OK to treat it as a bug fix.

olivermrbl (approved):

PR #12724, opened 2025-06-12
chore(): re enable integration tests bp

adrien2p:
RESOLVES FRMW-2980

**What**
Remove old experimental flag from integration tests as it interfere with v8 inspector protocol and break the breakpoints usage


olivermrbl (approved):
❤️

PR #12731, opened 2025-06-13
feat: Improve startup time by parallelizing module and link loading

sradevski:
REFERENCES CLO-842

### DIFF
--- a/packages/core/core-flows/src/order/workflows/transfer/decline-order-transfer.ts
+++ b/packages/core/core-flows/src/order/workflows/transfer/decline-order-transfer.ts
@@ -116,7 +116,7 @@ export const declineOrderTransferRequestWorkflow = createWorkflow(
   ): WorkflowData<void> {
     const orderQuery = useQueryGraphStep({
       entity: "order",
-      fields: ["id", "version", "declineed_at"],
+      fields: ["id", "version", "declined_at"],
       filters: { id: input.order_id },
       options: { throwIfKeyNotFound: true },
     }).config({ name: "order-query" })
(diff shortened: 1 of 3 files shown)


packages/core/core-flows/src/order/workflows/transfer/decline-order-transfer.ts
```diff
@@ -116,7 +116,7 @@ export const declineOrderTransferRequestWorkflow = createWorkflow(
   ): WorkflowData<void> {
     const orderQuery = useQueryGraphStep({
       entity: "order",
-      fields: ["id", "version", "declineed_at"],
+      fields: ["id", "version", "declined_at"],
```
adrien2p, line 119: nice catch 😅

packages/core/modules-sdk/src/medusa-module.ts
```diff
@@ -395,104 +395,109 @@ class MedusaModule {
...
-        moduleExports,
-        sharedContainer,
-        moduleDefinition,
-        injectedDependencies,
-      } = moduleOptions
-
-      const hashKey = simpleHash(
-        stringifyCircular({ moduleKey, defaultPath, declaration })
-      )
+    await Promise.all(
```
adrien2p: Q: just wondering if it would make sense to use PromiseAll with aggregated errors such that if multiple modules have issues we can show all of their errors in one go? the idea being that if the user have multiple errors he will have to handle them one by one and restart before seeing the next one. wdyt?

adrien2p (approved):

packages/core/modules-sdk/src/medusa-module.ts
sradevski: I just wanted to preserve the current behavior, but yes we can definitely do that. let me look into it

PR #12735, opened 2025-06-13
fix: add operators to RemoteQueryFilters

peterlgh7:
This `graph.query` call in the platform gives a typing error:

```
filters: {
  $or: [
    { id: input.build_id },
    { external_id: input.external_build_id },
  ],
}
```

since `RemoteQueryFilters` only allows the entity fields as keys. This PR allows `$or`, `$and` and `$not`, which all work fine at runtime.

### DIFF
--- a/packages/core/types/src/modules-sdk/to-remote-query.ts
+++ b/packages/core/types/src/modules-sdk/to-remote-query.ts
@@ -46,9 +46,9 @@ type ExtractFiltersOperators<
 }
 
 type RemoteQueryFilterOperators<T> = {
-  $or?: T[]
-  $and?: T[]
-  $not?: T
+  $or?: (T & RemoteQueryFilterOperators<T>)[]
+  $and?: (T & RemoteQueryFilterOperators<T>)[]
+  $not?: T & RemoteQueryFilterOperators<T>
 }
 
 /**
--- a/packages/core/types/src/modules-sdk/to-remote-query.ts
+++ b/packages/core/types/src/modules-sdk/to-remote-query.ts
@@ -45,6 +45,34 @@ type ExtractFiltersOperators<
     : never
 }
 
+type OperatorsOnlyFilters<
+  TEntry extends string,
+  RemoteQueryEntryPointsLevel = RemoteQueryEntryPoints,
+  Exclusion extends string[] = [],
+  Lim extends number = Depth[3]
+> = Lim extends number
+  ? {
+      $or?: RemoteQueryFilters<
+        TEntry,
+        RemoteQueryEntryPointsLevel,
+        Exclusion,
+        Depth[Lim]
+      >[]
+      $and?: RemoteQueryFilters<
+        TEntry,
+        RemoteQueryEntryPointsLevel,
+        Exclusion,
+        Depth[Lim]
+      >[]
+      $not?: RemoteQueryFilters<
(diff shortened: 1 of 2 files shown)


olivermrbl (approved):
LGTM, pending @adrien2p's review :)

packages/core/types/src/modules-sdk/to-remote-query.ts
```diff
@@ -69,4 +90,5 @@ export type RemoteQueryFilters<
           >
         >
     : Record<string, any>
-  : never
+  : never) &
+  OperatorsOnlyFilters<TEntry, RemoteQueryEntryPointsLevel, Exclusion, Lim>
```
adrien2p: suggestion: Let me know what do you think about the suggestion, the main idea being to simplify operators to be easier to extend if we need and also separate the responsibility of the types and finally have the type composition without re computing the types to many times (cc @thetutlage )

```ts
type RemoteQueryFilterOperators<T> = {
  $or?: T[]
  $and?: T[]
  $not?: T
}

/**
 * Extract all available filters from a remote entry point deeply
 */
export type InternalRemoteQueryFilters<
  TEntry extends string,
  RemoteQueryEntryPointsLevel = RemoteQueryEntryPoints,
  Exclusion extends string[] = [],
  Lim extends number = Depth[3]
> = Lim extends number
  ? TEntry extends keyof RemoteQueryEntryPointsLevel
    ? TypeOnly<RemoteQueryEntryPointsLevel[TEntry]> extends Array<infer V>
      ? Prettify<
          OmitNever<ExtractFiltersOperators<V, Lim, [TEntry, ...Exclusion]>>
        >
      : Prettify<
          OmitNever<
            ExtractFiltersOperators<
              RemoteQueryEntryPointsLevel[TEntry],
              Lim,
              [TEntry, ...Exclusion]
            >
          >
        >
    : Record<string, any>
  : never

export type RemoteQueryFilters<
  TEntry extends string,
  RemoteQueryEntryPointsLevel = RemoteQueryEntryPoints,
  Exclusion extends string[] = [],
  Lim extends number = Depth[3]
> = RemoteQueryFilterOperators<
  InternalRemoteQueryFilters<
    TEntry,
    RemoteQueryEntryPointsLevel,
    Exclusion,
    Lim
  >
> &
  InternalRemoteQueryFilters<
    TEntry,
    RemoteQueryEntryPointsLevel,
    Exclusion,
    Lim
  >
```

thetutlage: Yeah, this feels a bit more cleaner 👍

peterlgh7: looks good. Will try it out today

peterlgh7: @adrien2p suggestion applied. Should have done it like that from the start tbh 😄

peterlgh7: @adrien2p please see my last commit. I made `RemoteQueryFilterOperators` deep to support stuff like
```javascript
{
  $or: {
    $and: [...],
    $and: [...]
}
```
Let me know if that's ok.

adrien2p: that looks good to me man 💪

PR #12753, opened 2025-06-17
fix(core, medusa-test-utils): Fix medusa test runner plugin modules loading

adrien2p:
FIXES SUP-1812
**What**
- Prevent plugin compiler to ignore plugin containing chunk name within its package name
- Fix medusa test runner to actually load plugin modules


thetutlage (approved):
LGTM!

PR #12761, opened 2025-06-18
fix: initiate request container before other express middleware

thetutlage:


packages/core/framework/src/http/middlewares/error-handler.ts
```diff
@@ -20,7 +20,9 @@ export function errorHandler() {
     res: Response,
     _: NextFunction
   ) {
-    const logger = req.scope.resolve(ContainerRegistrationKeys.LOGGER)
+    const logger = req.scope
```
olivermrbl, line 23: question: When is `req.scope` not defined? I would assume this could only happen if something is not starting up correctly.

thetutlage: That's the mystery 😅 It should be defined all the time and we are not able to reproduce it locally.

So thought of using `console.log` to get access to the original error and find if there is something weird making `req.scope` not exist.

adrien2p: yeah it would be nice to have a log when it is not defined to be able to pin point it? maybe something that says that the container is not available so switching to console instead and some info like the URL could be nice?

thetutlage: Yeah, we found it is an issue with the DynamoDB connection which was initialized before we initialize the container. Therefore, the `req.scope` was undefined.

packages/core/framework/src/http/middlewares/error-handler.ts
```diff
@@ -20,14 +20,20 @@ export function errorHandler() {
...
+      ? req.scope.resolve(ContainerRegistrationKeys.LOGGER)
+      : console
 
-    err = formatException(err)
+    if (!req.scope) {
+      logger.error(
+        "req.scope is missing unexpectedly. It should be defined in all the cases"
```
shahednasser, line 29: suggestion: if this is an error that the user might face, I think the error should be more helpful to the user, like what should they do if they see this?

thetutlage: This is not really for the user, but for us to dig further into the error and see why `req.scope` is undefined.

For example, currently it was `undefined` because we were not able to connect to DynamoDB because the user had defined some env vars in their cloud dashboard, that they shouldn't define.

olivermrbl (approved):
LGTM

PR #12775, opened 2025-06-19
fix(workflow-engine-redis): Ensure PK is set without errors

olivermrbl:
### What

Drop the PK before re-setting it

### Why

We mistakenly named this migration differently in the workflow engines; this has caused issues with the migrations. Switching between engines will fail because the primary key is attempted to be set twice.

The issue happens in the following scenario:
1. In memory engine is used
2. Migration is run
3. Primary is key added
4. Redis engine is used
5. Migration is run
6. Same primary key is attempted to be set again
7. Migration fails

The same scenario can happen if you go from Redis to In memory.

With this fix, we ensure the primary key is only ever set once.

Long-term we want to upgrade the workflow engine modules to use the provider architecture, however a short-term solution is needed to resolve the error.

### DIFF
--- a/packages/modules/workflow-engine-inmemory/src/migrations/Migration20250505092459.ts
+++ b/packages/modules/workflow-engine-inmemory/src/migrations/Migration20250505092459.ts
@@ -19,6 +19,24 @@ export class Migration20250505092459 extends Migration {
     this.addSql(
       `CREATE UNIQUE INDEX IF NOT EXISTS "IDX_workflow_execution_workflow_id_transaction_id_run_id_unique" ON "workflow_execution" (workflow_id, transaction_id, run_id) WHERE deleted_at IS NULL;`
     )
+    /* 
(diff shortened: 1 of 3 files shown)


adrien2p (approved):
LGTM

olivermrbl:
@adrien2p, I had to go with the other approach, since solely renaming the file will cause the migration to be re-run for some customers, e.g. if they installed the Redis engine immediately, they won't have this name in their `mikro_orm_migrations` table, and therefore it won't be skipped. And when that happens, the same error is thrown.

adrien2p:
> @adrien2p, I had to go with the other approach, since solely renaming the file will cause the migration to be re-run for some customers, e.g. if they installed the Redis engine immediately, they won't have this name in their `mikro_orm_migrations` table, and therefore it won't be skipped. And when that happens, the same error is thrown.

No problem, I thought so too when I have read your pr earlier and thought the same.

PR #12788, opened 2025-06-20
fix(utils): build query withDeleted remove auto detection

adrien2p:
**What**
Currently, filtering data providing a `deleted_at` value will automatically apply the `withDeleted` flag which in turns remove the default constraint apply to all queries `deleted_at: null`. The problem is that it does not account for the value assign to `deleted_at` leading to inconsistent behaviour depending on the value. e.g filtering with `deleted_at: { $eq: null }` where the expectation is to only filter the non deleted record will end up returning deleted record as well by applying the `withDeleted` filters.

This pr revert this auto detection if favor of the user providing `withDeleted` explicitly, as it is already supported , plus the filters.

Further more, some integration tests demonstrate how to filter deleted records (e.g product) from the api. While the api did not properly support it, this pr adds support to pass with_deleted flags to the query and being handled accordingly to our api support. Validators have been updated and product list end point benefit from it. Also, the list config type was already accepting such value which I have translated to the remote query config.

Also, since the previous pr was adjusting the product types, I ve adjusted them to match the expectation


adrien2p:
WIP, I am still adding support globally, we can discuss it once don

adrien2p:
@olivermrbl would you have a minute sometime today to discuss this bad boy?

olivermrbl:
@adrien2p, let's do it. Ping me whenever

olivermrbl (approved):

PR #12795, opened 2025-06-23
fix(workflow-engine-*): Cleanup expired executions and reduce redis storage usage

adrien2p:
FIXES FRMW-2981
RESOLVES FRMW-2982

**What**
Fix clean up of the expired executions. Currently, all expired executions are removed no matter their state leading to un finished long running workflows being removed prematurely. This pr modify the way we are processing the cleanup and introduce check on the execution state.

This pr also remove the full context from being stored into redis and instead go to the DB when resuming a transaction


olivermrbl (approved):
Very nice stuff; pending @carlos-r-l-rodrigues's review 💪

adrien2p:
@olivermrbl I had to change the DML because currently even though the timestamps have a default `now()` they are applying application level date `new Date()` which means that it can create (I faced it) a diff between the database timestamp and the server timestamp. wdyt?

cc @carlos-r-l-rodrigues

carlos-r-l-rodrigues (approved):

olivermrbl (approved):
Nice catch @adrien2p, I think it makes sense to rely on the DB query for the timestamps. 

Separately, I don't think this will break anything, so it should be good to merge.

adrien2p:
@olivermrbl do you mind me adding the resolution for FRMW-2982 in that pr? it is pretty small

olivermrbl:
> @olivermrbl do you mind me adding the resolution for FRMW-2982 in that pr? it is pretty small

As in, you have the solution to only store part of workflow executions in Redis? 😄

adrien2p:
> > @olivermrbl do you mind me adding the resolution for FRMW-2982 in that pr? it is pretty small
> 
> As in, you have the solution to only store part of workflow executions in Redis? 😄

yes that's correct :)

adrien2p:
@olivermrbl @carlos-r-l-rodrigues should be good to re review that pr

carlos-r-l-rodrigues:
LGTM!

adrien2p:
@olivermrbl shall we merge?

PR #12803, opened 2025-06-23
fix(payment): round currency precision

### LINKED ISSUE #12789: [Bug]: Rounding Issue when capturing payment from stripe webhook
### Package.json file

```JSON
Medusa 2.8.4
"stripe": "^17.6.0",
```

### Node.js version

21

### Database and its version

16

mac

### Browser name

_No response_

### What happended?

We are capturing payments from stripe via webhook and we noticed an interesting bug:

In medusa, the payment_session has this amount:

```
98.9009
{"value": "98.9009", "precision": 20}
```

Now, in Stripe we are handing over the amount in **cents**, so that would be `9890`.

Now whenever we get the stripe webhook that the payment_intent was indeed successful with the amount (`9890`) the medusa backend matches its entry with the amount owed (`98.9009`) and detects a remainder since `98.9009 - 98.90 > 0.` 

Hence, it puts the payment_session in the state "Partially Captured" whereas it should be "Captured".... This in turn creates a lot of downstream issues for subscribers and stateful logics that depend on the payment being captured.

Since I cannot pass 4 digits to Stripe I need to somehow adjust this in medusa. Is anyone else having that issue?

This is rather urgent for us

### Expected behavior

Match the amounts correctly, e.g. round in medusa or allow a margin of error

### Actual behavior

Rounding issue

### Link to reproduction repo

n/A

carlos-r-l-rodrigues:
What:
 - Payment Authorization and Capture are rounded by payment providers. The Payment module must consider the currency precision to manage payment statuses correctly.
 
FIXES: https://github.com/medusajs/medusa/issues/12789


olivermrbl (approved):
LGTM.

I wonder if it would make sense to add this field to the big number class as well at some point? 

E.g. `rounded_amount`

PR #12809, opened 2025-06-24
fix: Allow setting the status of a payment session when updating

sradevski:
Currently there is no way to set the payment session status due to eg. a webhook coming in.

This PR allows for the status to be set, both by the caller, as well as based on the results of an update session call.

It also allows for CRUD on the metadata, which wasn't handled previously

### DIFF
--- a/packages/modules/payment/src/services/payment-module.ts
+++ b/packages/modules/payment/src/services/payment-module.ts
@@ -401,6 +401,7 @@ export default class PaymentModuleService
         currency_code: data.currency_code,
         context: data.context,
         data: data.data,
+        metadata: data.metadata,
       },
       sharedContext
     )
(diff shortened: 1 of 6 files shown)


packages/modules/payment/src/services/payment-module.ts
```diff
@@ -435,6 +436,9 @@ export default class PaymentModuleService
         amount: data.amount,
         currency_code: data.currency_code,
         data: providerData.data,
+        // Allow the caller to explicitly set the status (eg. due to a webhook), fallback to the update response, and finally to the existing status.
+        status: data.status ?? providerData.status ?? session.status,
```
carlos-r-l-rodrigues, line 440: I believe if you use `providerData.status`, it would require a normalization or check if it is a valid `PaymentSessionStatus` enum.

carlos-r-l-rodrigues (approved):
LGTM

packages/modules/payment/src/services/payment-module.ts
sradevski: The provider is supposed to return a valid PaymentSessionStatus (already happening), so there is no need for normalization here. It's the same normalization logic in authorizePaymentSession and createPaymentSession

PR #12812, opened 2025-06-24
fix(dashboard): fix currency input locale formatting

### LINKED ISSUE #12797: [Bug]: Refund amount not parsed properly when value contains comma
### What happended?

Issue on the Medusa v2.8.4 with the core Stripe payment provider.

#### Steps to reproduce
1. Open the Refund modal on the order detail in the Admin UI
2. Enter a value, i.e. `6,95`
3. The amount is changed to 6, also the succes widget show an amount of `€NaN` is refunded.
4. Also in the Stripe refund it is processed as a € 6 refund.

Note: in my admin I am not allowed to enter numeric values with a `.`, so there is also no workaround, except for doing the refund in Stripe and manually update database with refund value.

### Expected behavior

I expected to have refunded the actual amount of € 6,95

### Actual behavior

A refund was done of € 6

fPolic:
**What**
- fix float parsing for currency inputs that have float values (when the value is formatted with comma as decimal separator)
- cleanup and fix refund forms
- update currency input forms in RMA flows

---

FIXES SUP-1914

### DIFF
--- a/packages/admin/dashboard/src/lib/format-currency.ts
+++ b/packages/admin/dashboard/src/lib/format-currency.ts
@@ -1,5 +1,5 @@
 export const formatCurrency = (amount: number, currency: string) => {
-  return new Intl.NumberFormat("en-US", {
+  return new Intl.NumberFormat(undefined, {
     style: "currency",
     currency,
     signDisplay: "auto",
--- a/packages/admin/dashboard/src/routes/orders/order-balance-settlement/components/order-balance-settlement-form/order-balance-settlement-form.tsx
+++ b/packages/admin/dashboard/src/routes/orders/order-balance-settlement/components/order-balance-settlement-form/order-balance-settlement-form.tsx
@@ -5,7 +5,6 @@ import {
   clx,
   CurrencyInput,
   Divider,
-  Input,
   Label,
   RadioGroup,
   Select,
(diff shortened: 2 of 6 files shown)


packages/admin/dashboard/src/lib/format-currency.ts
```diff
@@ -1,5 +1,5 @@
 export const formatCurrency = (amount: number, currency: string) => {
-  return new Intl.NumberFormat("en-US", {
+  return new Intl.NumberFormat(undefined, {
```
olivermrbl, line 2: question: Why was this change required?

olivermrbl:
Just tried the snapshot; I cannot seem to use comma as a separation, e.g. for Danish kroner.

packages/admin/dashboard/src/lib/format-currency.ts
fPolic: With this change, the default runtime locale is used for formatting instead of forcing `en-US`. This way currency formatting should be the same for the currency input and the places we use this helper (previously for example with DE locale formatting would be `1.234,5` but the helper would return `1,234.5`.

fPolic:
> Just tried the snapshot; I cannot seem to use comma as a separation, e.g. for Danish kroner.

What language/locale you have set in the browser? If you change to `de-DE` for example, you should be able to use a comma as a decimal separator.

olivermrbl:
> > Just tried the snapshot; I cannot seem to use comma as a separation, e.g. for Danish kroner.
> 
> What language/locale you have set in the browser? If you change to `de-DE` for example, you should be able to use a comma as a decimal separator.

Shouldn't this be automatically detected and set?

fPolic:
Yes, the browser sets the locale depending on OS settings, browser preferences etc. (I suggested changing to German to simulate comma formatting).

If English is detected locale will be set to `navigator.language = "en-GB"`, in that case all currencies on the Admin will be formatted with `.` as a decimal separator. If German is detected, `navigator.language` will be `de-DE` and all currencies on admin will use a comma for decimal separator.

olivermrbl (approved):

PR #12813, opened 2025-06-24
feat(): Add support for jwt asymetric keys

adrien2p:
**What**
Add support for JWT asymetric keys validation.
This pr add support for users to specify a private and public key alongside other options as per [SignOptions](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/e507b7a12247e581d58e1fa669d3c59835b30649/types/jsonwebtoken/index.d.ts#L27C18-L27C29) and [VerifyOptions](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/e507b7a12247e581d58e1fa669d3c59835b30649/types/jsonwebtoken/index.d.ts#L57)

The symetric validation is still supported without changing anything.

to benefit from the new asymetric validation you can now set the following (as example)
```ts
const jwtOptions = {
  algorithm: "RS256",
  expiresIn: "1h",
  issuer: "medusa",
  keyid: "medusa",
}


defineConfig({
  admin: {
    disable: true,
  },
  projectConfig: {
    http: {
      jwtSecret: {
        key: privateKey,
        passphrase,
      },
      jwtPublicKey: publicKey,
      jwtOptions: jwtOptions,
      jwtVerifyOptions: { /* ... */ }
    },
  }
})
```


riqwan (approved):
lgtm! Would get a second pair of eyes 🤞🏻

packages/core/utils/src/auth/token.ts
```diff
@@ -1,21 +1,27 @@
...
-  if (!jwtConfig.secret || !jwtConfig.expiresIn) {
+  if (
+    !jwtConfig.secret ||
+    (!jwtConfig.expiresIn && !jwtConfig.jwtOptions?.expiresIn)
+  ) {
     throw new MedusaError(
       MedusaError.Types.INVALID_ARGUMENT,
       "JWT secret and expiresIn must be provided when generating a token"
     )
   }
```
carlos-r-l-rodrigues, line 21: Just for sake of awareness, I would at least `warn` the user if a public key is present on the `jwtConfig` but the options doesnt not contain the `algorithm`. I'm affraid people will just set the public key, but the default algorithm will be symetric.

carlos-r-l-rodrigues (approved):
LGTM

packages/core/types/src/common/config-module.ts
```diff
@@ -505,7 +506,50 @@ export type ProjectConfigOptions = {
...
+     *         keyid: "medusa",
+     *       }
+     *     }
+     *     // ...
+     *   },
+     *   // ...
+     * })
+     * ```
+     */
+    jwtOptions?: SignOptions
```
sgirones, line 551: I see you reuse most of the `SignOptions` in the `verify` method, but for example, important options like `audience` and `issuer` don't map 100%. Should we also add `VerifyOptions`?
We could add `jwtVerifyOptions` as well and fallback to `jwtOptions` if not provided.

If we offer full support for `VerifyOptions`, it should open the door to offer authentication through other OIDC providers in the future, and given we are already working on this area, why not adding it now.

packages/core/utils/src/auth/token.ts
adrien2p: I ll had a descriptive warn

packages/modules/user/src/services/user-module.ts
```diff
@@ -155,10 +184,17 @@ export default class UserModuleService
       }
     }
 
+    const expiresAt =
```
sradevski: NIT: Extract this as it's used in a couple of places now

packages/modules/user/src/services/user-module.ts
```diff
@@ -51,23 +54,35 @@ export default class UserModuleService
   protected readonly inviteService_: ModulesSdkTypes.IMedusaInternalService<
     InferEntityType<typeof Invite>
   >
-  protected readonly config: { jwtSecret: string; expiresIn: number }
+  protected readonly config: {
+    jwtSecret: string
+    jwtPublicKey?: string
```
sradevski, line 59: Just for my understanding I assume using asymmetric auth for the invite tokens isn't as useful, but this change was done since the same algorithm is used for all JWT tokens in the stack right?

sradevski (approved):

packages/modules/user/src/services/user-module.ts
adrien2p: yes it was just to align everything. Since the user module options can be passed and be different from the project config I though it just enabled the support here too. Currently the user config just take the config from the project config

adrien2p:
@sgirones @sradevski @carlos-r-l-rodrigues I ve included your feedback 🚀

PR #12814, opened 2025-06-24
feat: Search by billing_address and shipping_address on Order model

juanzgc:
**What**
Added `.searchable()` method to the `shipping_address` and `billing_address` relationships in the `Order` model, enabling search functionality.

**Why**
Previously searching via `shipping_address` and `billing_address` was not supported, limiting the ability to find orders based on shipping or billing address information.

Resolves SUP-1838

### DIFF
--- a/integration-tests/http/__tests__/order/admin/order.spec.ts
+++ b/integration-tests/http/__tests__/order/admin/order.spec.ts
@@ -66,6 +66,35 @@ medusaIntegrationTestRunner({
         expect(response.data.orders).toHaveLength(0)
         expect(response.data.orders).toEqual([])
       })
+
+      it("should search orders by shipping address", async () => {
+        let response = await api.get(`/admin/orders?fields=+shipping_address.address_1,+shipping_address.address_2`, adminHeaders)
+
+        expect(response.data.orders).toHaveLength(1)
+        expect(response.data.orders).toEqual([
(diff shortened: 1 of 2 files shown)


integration-tests/http/__tests__/order/admin/order.spec.ts
```diff
@@ -66,6 +66,35 @@ medusaIntegrationTestRunner({
         expect(response.data.orders).toHaveLength(0)
         expect(response.data.orders).toEqual([])
       })
+
+      it("should search orders by shipping address", async () => {
```
olivermrbl, line 70: thought: Should we also add a test for when there is no address with the search parameter to avoid false positives?

juanzgc: Updated integration tests to include billing address and also non existing addresses.

olivermrbl (approved):

integration-tests/http/__tests__/order/admin/order.spec.ts
```diff
@@ -66,6 +66,69 @@ medusaIntegrationTestRunner({
         expect(response.data.orders).toHaveLength(0)
```
adrien2p, line 66: Q: I feel like you have some issues with the linting rules. I would have expected the format to be a bit different :) could you have a look at it please

PR #12880, opened 2025-07-02
fix: medusa-service generated update functions always returns an array even if single object is passed.

dwene:
Related to the change with Medusa Internal Service here: https://github.com/medusajs/medusa/pull/12715/files#diff-4d287d33cf799750dfc3957d3c22d87a5d369ca15bbbc7db29265fc864cedfadL248

Context: all of my medusa-service generated update functions are always returning an array, forcing me to have to check the response myself.

ex:

```
export const updateThemeStep = createStep(
  'update-theme-step',
  async (input: UpdateCheckoutThemeWorkflowStepInput, { container }) => {
    const themeModuleService: ThemeModuleService = container.resolve(CHECKOUT_THEME_MODULE);

    const { channel_id, ...updateThemeData } = input;

    const originalTheme = await queryCheckoutTheme({ container, themeId: input.id });

    const theme = await themeModuleService.updateCheckoutThemes(updateThemeData);

    if (Array.isArray(theme)) return new StepResponse(theme[0], originalTheme);

    return new StepResponse(theme, originalTheme);
  },
  async (originalTheme, { container }) => {
    if (!originalTheme) return;

    const themeModuleService: ThemeModuleService = container.resolve(CHECKOUT_THEME_MODULE);

    await themeModuleService.updateCheckoutThemes(originalTheme);
  }
);
```


adrien2p (approved):

packages/core/utils/src/modules-sdk/medusa-service.ts
```diff
@@ -243,7 +243,7 @@ export function MedusaService<
           data = [],
           sharedContext: Context = {}
         ): Promise<T | T[]> {
-          const serviceData = Array.isArray(data) ? data : [data]
+          const serviceData = data
```
olivermrbl: @adrien2p, should we add a few tests for this before we merge?

adrien2p: Yea definitely 💯 forgot to mention it. Also, removing the service data const as it does not do anything anymore

olivermrbl: cc @dwene

olivermrbl (approved):

dwene:
@olivermrbl OK all the tests are passing now. Looks like create had the same issue.

dwene:
Looks like this was fixed by a core team member in PR https://github.com/medusajs/medusa/pull/13225 so I'm going to close this.

Thanks for fixing this @carlos-r-l-rodrigues !

PR #12881, opened 2025-07-03
fix(core-flows): Locations levels check in draft order and order edit flows

olivermrbl:


carlos-r-l-rodrigues (approved):
just need a changeset

carlos-r-l-rodrigues:
We could also bring the `raw_` ones, like the cart utils,for the big precision calculations.

olivermrbl:
@carlos-r-l-rodrigues, teah, let's do that. Maybe we should centralise the `fields` actually? To avoid having to update these whenever we find a bug.

carlos-r-l-rodrigues:
@olivermrbl, yeah, I believe having a constant would be great.
Some places it is `items.variant` and other `product.variants`. I guess this is the only thing to pay attention to.

carlos-r-l-rodrigues (approved):

PR #12886, opened 2025-07-04
fix(orchestration): Handle expected lifecycle errors

adrien2p:
**What**
Correctly handle all expected error during lifecycle in relation to conccurency


olivermrbl:
@adrien2p, is this ready to be reviewed/merged? We expect to cut a release tomorrow and would be great to get included, if possible :)

adrien2p:
Yes it is good, I was actually waiting for stev to be sure his scenario was covered which worked for me during my tests. But I think if anything comes up I can open a separate pr wdyt?

olivermrbl:
> Yes it is good, I was actually waiting for stev to be sure his scenario was covered which worked for me during my tests. But I think if anything comes up I can open a separate pr wdyt?

Sounds good!

olivermrbl (approved):

PR #12897, opened 2025-07-08
fix(dashboard): show TIP flag only for fixed promotions

fPolic:


packages/admin/dashboard/src/routes/promotions/promotion-create/components/create-promotion-form/templates.ts
```diff
@@ -10,7 +11,9 @@ export const templates = [
     type: "standard",
     title: "Amount off products",
     description: "Discount specific products or collection of products",
-    hiddenFields: [...commonHiddenFields],
+    hiddenFields: [
+      ...commonHiddenFields.filter((field) => field !== "is_tax_inclusive"),
```
olivermrbl: thought: This approach seems a bit hacky; would it make sense to make a dedicated array with hidden fields for the different promotions? In case we need to update these in the future

olivermrbl (approved):
LGTM

Issue #5048, opened 2023-09-13
Can't setup 9.975% tax rate - Québec Sales Tax (QST)


alexdx-ca:
### Describe the bug

I can't setup Québec Sales Tax (QST). Because it's value is 9.975%
But Medusa is asking me 9.97% or 9.98%.

### System information

Medusa version: 1.3.20
Node.js version: 
Database: PostgreSQL 15.4
Operating system: macOS 12.6.1
Browser (if relevant): Firefox 

### Steps to reproduce the behavior

1. Go to http://localhost:7001/a/settings
2. Click on Taxes.
3. Click on "+ New Tax Rate" button.
4. Enter name, tax rate (9.975) and tax code.
5. Click "Create" and see error.

### Expected behavior

Tax rate field have to accept more than 2 decimal places.

### Screenshot

<img width="757" alt="Screen Shot 2023-09-13 at 6 04 12 PM" src="https://github.com/medusajs/medusa/assets/7256258/e1cc5190-4b7b-4ccd-a8a5-06d11b1b29ef">

pevey:
Just adding info that this is not a limitation on the database.  I can manually add a tax rate with high specificity.  It might be just a matter of the admin app validators.

Have you tried adding the rate via the medusa API?  If it works, that or adding to the database directly could be a workaround until this is fixed.

SalahAdDin:
@alexdx-ca  any update on this?

alexdx-ca:
Just tried latest version of Medusa!
The issue is still there, form accepts only 2 decimal places.
So if possible, can you fix it please for future versions.
Thank you so much for your time and help!

SalahAdDin:
Need to ping someone from the staff then,

SalahAdDin:
@alexdx-ca anything here?

SalahAdDin:
@alexdx-ca anything here?

AmbroziuBaban:
@SalahAdDin . I've created an PR which will allow percentage value with 4 decimals. Now I'm waiting the code review.

SalahAdDin:
> [@SalahAdDin](https://github.com/SalahAdDin) . I've created an PR which will allow percentage value with 4 decimals. Now I'm waiting the code review.

@AmbroziuBaban was it fixed?

SalahAdDin:
> [@SalahAdDin](https://github.com/SalahAdDin) . I've created an PR which will allow percentage value with 4 decimals. Now I'm waiting the code review.

[...]

PR #12902, opened 2025-07-08
fix(dashboard): percentage decimals tax rates support

AmbroziuBaban:
Added support for tax rates with 4 decimals and modified the Tax Rates Forms to support 4 decimals value input.

Fixed this [issue](https://github.com/medusajs/medusa/issues/5048)


packages/admin/dashboard/src/lib/percentage-helpers.ts
```diff
@@ -1,6 +1,7 @@
 const formatter = new Intl.NumberFormat([], {
   style: "percent",
   minimumFractionDigits: 2,
+  maximumFractionDigits: 4,
```
olivermrbl, line 4: question: do we need to constrain this to maximum four?

packages/admin/dashboard/src/routes/tax-regions/tax-region-tax-override-create/components/tax-region-override-create-form/tax-region-tax-override-create.tsx
```diff
@@ -389,7 +389,8 @@ export const TaxRegionCreateTaxOverrideForm = ({
                             <Form.Control>
                               <PercentageInput
                                 {...field}
-                                placeholder="0.00"
+                                placeholder="0.0000"
```
olivermrbl: todo: let's not change the placeholder; four decimals is pretty unusual for a tax rate, so no need to have that as the placeholder

packages/admin/dashboard/src/lib/percentage-helpers.ts
olivermrbl, line 4: question: do we need to constrain this to maximum four?

AmbroziuBaban: No necessary, but we need to set a maximum. We can to set a value between 0 and 100. I chose 4 because that is the number required in the related issues. I also don't know if more than 4 will be needed, considering that 2 are generally used.

.changeset/large-months-yawn.md
```diff
@@ -0,0 +1,5 @@
+---
+"@medusajs/dashboard": patch
+---
+
+Added support for 4 decimals tax rates.
```
willbouch, line 5: could I ask you to change this here and set it to the commit name please?

packages/admin/dashboard/src/lib/percentage-helpers.ts
willbouch: 4 would be good enough for now

willbouch (dismissed):

willbouch:
Thanks for the contribution! I actually copied the PR and created [this one](https://github.com/medusajs/medusa/pull/13407) so I can add a few things. We need this feature asap so will take care of it for now :)

PR #12903, opened 2025-07-08
fix(orchestration): Prevent workf. cancellation to execute while rescheduling

adrien2p:
**What**
Currently, when cancelling async workflows, the step will get rescheduled while the current worker try to continue the execution leading to concurrency failure on compensation. This pr prevent the current worker from executing while an async step gets rescheduled


carlos-r-l-rodrigues (approved):

adrien2p:
@sradevski i ll wait your confirmation on top of my tests before getting that in 👌

sradevski (approved):
Merged, tests are looking good, trying in staging now, but I think we can merge it already

adrien2p:
@sradevski I ve added the support to pass `{ triggerCount }` as part of the wait subscriber execution util. let me know what you think

sradevski:
@adrien2p it didn't quite work in staging, I need to see why. The deployment got stuck in a deploying state

PR #12908, opened 2025-07-09
refactor(framework): optimize middleware file loading logic

OnlyWick:


olivermrbl:
@OnlyWick, please add a PR description explaining the change. I think the existing code works and reads fine.

adrien2p (commented):
thank you for your contribution, though, I am not really sure how it improve the loading logic, could you explain a bit more please? Also, looping through 2 items which is unlikely to expand I think it might be a bit overkill but I d like your thoughts about it 👍

OnlyWick:
OK. Yeah, it might be a bit overkill.

This is my first thought when I saw this code:
1. If you want to support more extensions, such as `.mjs` or `.cjs`, you need to add another `else if`, and write `fs.exists` and `processMiddlewareFile` again
2. There is a priority for searching these extensions, and `array` and `for...of` can clearly see the priority of loading

olivermrbl:
@OnlyWick, we just discussed internally, and while we highly appreciate your willingness to contribute, we will leave this one out, as it doesn't seem to introduce any significant improvements.

I'd love to accept other PRs; feel free to scour our issues board, if you are up for it.

PR #12919, opened 2025-07-10
fix(core-flows): updating tax lines when draft order shipping is removed

fPolic:
**What**
- don't call `updateOrderTaxLinesWorkflow` when a shipping method is removed from a draft order (tax lines will be cascade deleted with the method)


packages/core/core-flows/src/draft-order/workflows/remove-draft-order-shipping-method.ts
```diff
@@ -83,12 +82,6 @@ export const removeDraftOrderShippingMethodWorkflow = createWorkflow(
 
     validateDraftOrderChangeStep({ order, orderChange })
 
-    updateOrderTaxLinesWorkflow.runAsStep({
```
olivermrbl, line 90: thought: should we include a test? 🙏

fPolic: yes, writing one now, this is why I left it as a draft PR

olivermrbl (approved):

PR #12920, opened 2025-07-10
fix(link-modules,core-flows): Carry over cart promotions to order promotions

riqwan:
what:

- Carry over cart promotions to order promotions


packages/core/core-flows/src/draft-order/utils/fields.ts
```diff
@@ -27,10 +27,7 @@ export const draftOrderFieldsForRefreshSteps = [
   "shipping_methods.tax_lines.*",
   "customer.*",
   "customer.groups.*",
-  "promotion_link.*",
-  "promotion_link.promotion",
-  "promotion_link.promotion.id",
-  "promotion_link.promotion.code",
+  "promotions.*",
```
fPolic: nit: think we only need `id` and `code`

fPolic (approved):
LGTM

packages/modules/link-modules/src/definitions/order-promotion.ts
```diff
@@ -44,8 +44,8 @@ export const OrderPromotion: ModuleJoinerConfig = {
       serviceName: Modules.ORDER,
       entity: "Order",
       fieldAlias: {
-        promotion: {
-          path: "promotion_link.promotion",
+        promotions: {
```
olivermrbl, line 47: thought: this is a breaking change; should we keep + deprecate the old alias for now, and remove it later? would like to avoid having to cut a minor for a small change like this

riqwan: like this? https://github.com/medusajs/medusa/pull/12925/files

Issue #12901, opened 2025-07-08
[Bug]: Vercel Deployment of Admin UI


ejsinfuego:
### Package.json file

```JSON
{
  "name": "medusa-backend",
  "version": "0.0.1",
  "description": "A starter for Medusa projects.",
  "author": "Medusa (https://medusajs.com)",
  "license": "MIT",
  "keywords": [
    "sqlite",
    "postgres",
    "typescript",
    "ecommerce",
    "headless",
    "medusa"
  ],
  "scripts": {
    "build": "npx medusa build --admin-only",
    "build:vercel": "npx medusa build --admin-only",
    "predeploy": "medusa db:migrate",
    "seed": "medusa exec ./src/scripts/seed.ts",
    "seed:pickup": "medusa exec ./src/scripts/seed-pickup.ts",
    "setup:pickup": "node ./scripts/setup-pickup.js",
    "start": "medusa start",
    "start:server": "medusa start",
    "start:worker": "medusa start",
    "start:prod": "NODE_ENV=production medusa start",
    "dev": "medusa develop",
    "seed:products": "medusa exec ./src/scripts/seed-product.ts"
  },
  "dependencies": {
    "@medusajs/admin-sdk": "2.8.2",
    "@medusajs/cli": "2.8.2",
    "@medusajs/framework": "2.8.2",
    "@medusajs/medusa": "2.8.2",
    "awilix": "^8.0.1",
    "dotenv": "^16.5.0",
    "pg": "^8.13.0"
  },
  "devDependencies": {
    "@medusajs/test-utils": "2.8.2",
    "@swc/core": "1.5.7",
    "@swc/jest": "^0.2.36",
    "@types/jest": "^29.5.13",
    "@types/node": "^20.0.0",
    "@types/react": "^18.3.2",
    "@types/react-dom": "^18.2.25",
    "jest": "^29.7.0",
    "prop-types": "^15.8.1",
    "react": "^18.2.0",
    "react-dom": "^18.2.0",
    "ts-node": "^10.9.2",
    "typescript": "^5.6.2",
    "vite": "^5.2.11",
    "yalc": "^1.0.0-pre.53"
  },
  "engines": {
    "node": ">=20"
  }
}
```

### Node.js version

22.14.0

### Database and its version

Posgresql 16.2

### Operating system name and version

macOS Sequoia 15.5

### Browser name

Edge

### What happended?

[...]

PR #12922, opened 2025-07-11
docs(dashboard): add guidance for standalone admin dashboard build and deployment

ejsinfuego:
This update enhances the documentation for the Medusa build process by adding a dedicated section on how to build and deploy the admin dashboard independently from the main application. It explains the use of the --admin-only flag with the medusa build command, which outputs the admin dashboard to a separate directory suitable for standalone deployment (e.g., on Vercel).

this is related to this issue #12901


shahednasser:
Thank you for your contribution! However, we only recommend in our documentation deploying the Medusa server with the admin, as you can see in [this guide](https://docs.medusajs.com/learn/deployment). For that reason, we can't accept this PR.

PR #12926, opened 2025-07-11
fix(workflows-sdk): fix step name config used before default

carlos-r-l-rodrigues:
what:
* When same step is used twice, if the first one uses a `.config({name: ""})` the second one is not overwriting it. So the result will be the same as the first step.


olivermrbl:
@adrien2p, could you have a look at this? Think you are better suited to review it, than I am.

adrien2p (approved):

PR #12927, opened 2025-07-11
bump @medusajs/admin-bundler

carlos-r-l-rodrigues:


olivermrbl:
@carlos-r-l-rodrigues, I think this is no longer relevant right?

carlos-r-l-rodrigues:
> @carlos-r-l-rodrigues, I think this is no longer relevant right?

I don't know TBH. There is a PR where this package was changed and we didn't include it to the changeset.
Is it still needed for new releases? or maybe another PR in develop changed this package too and it is all good?

olivermrbl:
Yeah it should be all good. Currently, we use changesets mainly to generate an entry in the changelog of the package and less for versioning. <br><br>So in this case, I think we can close this PR

PR #12928, opened 2025-07-12
chore(link-modules): keep promotion alias

riqwan:
what:

- keeps promotion and order link


olivermrbl:
This looks good.

Another thing that I think we missed in the first PR: we need to add `promotions` as a field/relation in addition to `promotion` when querying orders across the core, otherwise the new field alias won't be returned to the end-user, unless they specify it.

riqwan:
Hey @olivermrbl, I don't quite follow the last comment. *promotion or *promotions are not a default field or relation in any of the core apis. The workflows are already accounted for.

olivermrbl:
> I don't quite follow the last comment. *promotion or *promotions are not a default field or relation in any of the core apis. The workflows are already accounted for.

I believe we are querying for `promotion` from the admin dashboard, at least on the order details page.

olivermrbl (approved):
LGTM

PR #12930, opened 2025-07-12
chore(cli): use constants to override npm levels in winston

OnlyWick:


sradevski:
This doesn't really help with anything, and it allocated an unnecessary variable, closing

PR #12957, opened 2025-07-14
fix: only log server errors

peterlgh7:
We're seeing a lot of "expected" client-side 4xx errors in our logs, which just create a lot of noise. A typical case is stale UI: someone performs an action on an entity that doesn't exist anymore, or isn't anymore in the state the UI is showing, resulting in 404 or other errors. I think we should only log 5xx as errors, as they are the server's responsibility. For 4xx we have the request logging which should be enough.

### DIFF
--- a/packages/core/framework/src/http/middlewares/error-handler.ts
+++ b/packages/core/framework/src/http/middlewares/error-handler.ts
@@ -1,9 +1,9 @@
+import { ErrorRequestHandler, NextFunction, Response } from "express"
 import { fromZodIssue } from "zod-validation-error"
-import { NextFunction, ErrorRequestHandler, Response } from "express"
 
 import { ContainerRegistrationKeys, MedusaError } from "@medusajs/utils"
-import { formatException } from "./exception-formatter"
 import { MedusaRequest } from "../types"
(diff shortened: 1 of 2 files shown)


sradevski (approved):

packages/core/framework/src/http/middlewares/error-handler.ts
```diff
@@ -82,6 +81,10 @@ export function errorHandler() {
         break
     }
 
+    if (statusCode >= 500) {
```
adrien2p, line 84: Q: should we continue to log when we get there but instead we can log something like warning or info? wdyt

peterlgh7: You mean for <500 ? Yeah that was an option I considered. My preference would be info, cause warning still gives a sense of something going wrong, and in my view these errors are business as usual from the server's perspective. I'll log with with info now but let me know if you have a strong preference for warning.

adrien2p: I am fine with info as well 💪

adrien2p (approved):

PR #12958, opened 2025-07-14
fix(core-flows): properly delete variant inventory item

fPolic:
**What**
- properly delete variant's inventory item when a variant is deleted (variant inventory items were queried after links were removed)

---

CLOSES CMRC-1012

### DIFF
--- a/packages/core/core-flows/src/product/workflows/delete-product-variants.ts
+++ b/packages/core/core-flows/src/product/workflows/delete-product-variants.ts
@@ -20,7 +20,7 @@ import { deleteInventoryItemWorkflow } from "../../inventory"
 /**
  * The data to delete one or more product variants.
  */
-export type DeleteProductVariantsWorkflowInput = { 
+export type DeleteProductVariantsWorkflowInput = {
   /**
    * The IDs of the variants to delete.
    */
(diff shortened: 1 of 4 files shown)


packages/core/core-flows/src/product/workflows/delete-product-variants.ts
```diff
@@ -71,6 +67,10 @@ export const deleteProductVariantsWorkflow = createWorkflow(
       },
     })
 
+    removeRemoteLinkStep({
+      [Modules.PRODUCT]: { variant_id: input.ids },
+    }).config({ name: "remove-variant-link-step" })
```
olivermrbl, line 72: question: does this account for inventory items associated with more than one variants?

fPolic: items that are only associated with variants being deleted are removed:

https://github.com/medusajs/medusa/blob/db61289fba02c337a34671dc129bdcd5e75b7e69/packages/core/core-flows/src/product/workflows/delete-product-variants.ts#L89

olivermrbl: Oh yeah, that's right. My bad.

olivermrbl (approved):

PR #12969, opened 2025-07-16
fix: accepted values in import with template

juanzgc:
**What**
Fixed CSV import functionality to properly handle columns that were previously cuasing import errors.

**Why**
Users were encountering "Invalid column name(s)" errors when importing CSV files containing system-generated columns like "Product Created At", "Product Updated At", etc. These columns are automatically added by export templates but should be ignored during import since they're not part of the product creation schema.

Resolves FRMW-2983


adrien2p (approved):

olivermrbl (approved):
LGTM

PR #12981, opened 2025-07-17
fix(pricing): fix pricing query when max_quantity is null

riqwan:
what: 

Prices when max_quantity value is null is accounted for when quantity is passed in context.


packages/modules/pricing/integration-tests/__tests__/services/pricing-module/calculate-price.spec.ts
```diff
@@ -451,6 +461,59 @@ moduleIntegrationTestRunner<IPricingModuleService>({
...
+
+          const calculatedPrice = await service.calculatePrices(
+            { id: ["price-set-EUR", "price-set-PLN"] },
+            { context }
+          )
+
+          console.log(
+            "calculatedPrice - ",
+            JSON.stringify(calculatedPrice, null, 2)
+          )
```
olivermrbl:
```suggestion
```

packages/modules/pricing/integration-tests/__tests__/services/pricing-module/calculate-price.spec.ts
```diff
@@ -451,6 +461,59 @@ moduleIntegrationTestRunner<IPricingModuleService>({
...
+                min_quantity: 20,
+                max_quantity: null,
+              },
+              original_price: {
+                id: "price-PLN-min-quantity-only",
+                price_list_id: null,
+                price_list_type: null,
+                min_quantity: 20,
+                max_quantity: null,
+              },
```
olivermrbl, line 508: thought(non-blocking): Should these not contain an amount by default? Without it, they are of no good use.

olivermrbl (approved):
lgtm

PR #12989, opened 2025-07-18
fix(dashboard): allocation UI for orders with more than 20 reservation items

fPolic:


packages/admin/dashboard/src/lib/orders.ts
```diff
@@ -1,8 +1,23 @@
...
+
+/**
+ * Returns a limit for number of reservations that order can have.
+ */
+export function getReservationsLimitCount(order: AdminOrder) {
+  if (!order?.items?.length) {
+    return 0
+  }
+
+  return order.items.reduce(
```
srindom, line 18: thought (non-blocking): not sure if this can happen - but could it happen that all items are not included here due to a limit in the list call made by remote query? 

not something to fix in this PR but might be worth double checking

srindom (approved):
lgtm!

packages/admin/dashboard/src/lib/orders.ts
fPolic: test with an order with 24 items and everything looks correct, item counts and reservations
Screenshot 2025-07-18 at 14 49 29

Issue #12986, opened 2025-07-18
[Bug]: 2.8.7 entity code prompt missing in query.graph


X-X-X-X-X-X-X-X-X-X-X-X-X:
### Package.json file

```JSON
{
  "name": "my-medusa-store",
  "version": "0.0.1",
  "description": "A starter for Medusa projects.",
  "author": "Medusa (https://medusajs.com)",
  "license": "MIT",
  "keywords": [
    "sqlite",
    "postgres",
    "typescript",
    "ecommerce",
    "headless",
    "medusa"
  ],
  "scripts": {
    "build": "medusa build",
    "seed": "medusa exec ./src/scripts/seed.ts",
    "start": "medusa start",
    "dev": "medusa develop",
    "test:integration:http": "TEST_TYPE=integration:http NODE_OPTIONS=--experimental-vm-modules jest --silent=false --runInBand --forceExit",
    "test:integration:modules": "TEST_TYPE=integration:modules NODE_OPTIONS=--experimental-vm-modules jest --silent --runInBand --forceExit",
    "test:unit": "TEST_TYPE=unit NODE_OPTIONS=--experimental-vm-modules jest --silent --runInBand --forceExit"
  },
  "dependencies": {
    "@medusajs/admin-sdk": "2.8.7",
    "@medusajs/cli": "2.8.7",
    "@medusajs/framework": "2.8.7",
    "@medusajs/medusa": "2.8.7",
    "@mikro-orm/core": "6.4.3",
    "@mikro-orm/knex": "6.4.3",
    "@mikro-orm/migrations": "6.4.3",
    "@mikro-orm/postgresql": "6.4.3",
    "awilix": "^8.0.1",
    "pg": "^8.13.0"
  },
  "devDependencies": {
    "@medusajs/test-utils": "2.8.7",
    "@mikro-orm/cli": "6.4.3",
    "@swc/core": "1.5.7",
    "@swc/jest": "^0.2.36",
    "@types/jest": "^29.5.13",
    "@types/node": "^20.0.0",
    "@types/react": "^18.3.2",
    "@types/react-dom": "^18.2.25",
    "jest": "^29.7.0",
    "prop-types": "^15.8.1",
    "react": "^18.2.0",
    "react-dom": "^18.2.0",
    "ts-node": "^10.9.2",
    "typescript": "^5.6.2",
    "vite": "^5.2.11",
    "yalc": "^1.0.0-pre.53"
  },
  "engines": {
    "node": ">=20"
[...]

PR #13002, opened 2025-07-21
fix(modules-sdk): Entity types

olivermrbl:
Fixes regression introduced in https://github.com/medusajs/medusa/pull/12712

Highlighted in https://github.com/medusajs/medusa/issues/12986

**What**

Re-introduce autocomplete on the `entity` property of the `query.graph` API


carlos-r-l-rodrigues (approved):

adrien2p (approved):
Q: Is it sufficient?

olivermrbl:
@adrien2p, yeah the issue was that with the changes introduced a few weeks back, the type of `entity` is always `TEntry`, and `TEntry` is always just a string. This is fine for custom entities that are not present in our autogenerated types, however for the ones that are, we need to explicitly expect that `entity` might be one of them. Otherwise, there will be no suggestions in the IDE.

Is there anything else you think I should test before I merge?

PR #13007, opened 2025-07-21
fix: fix script migrations order

peterlgh7:
`glob.sync` does not seem to guarantee ascending order. Instead, it seems to depend on the OS. In my Mac e.g., I get sometimes reversed order, which means script migrations got executed in reverse order.


carlos-r-l-rodrigues (approved):
LGTM

PR #13019, opened 2025-07-22
fix(dashboard, product): product attributes update with a relation update

fPolic:
**What**
- a relation update would unset some of the product attributes such as width, height etc.

---

CLOSES CMRC-1074
CLOSES CMRC-1076


packages/modules/product/src/repositories/product.ts
```diff
@@ -57,10 +58,18 @@ export class ProductRepository extends DALUtils.mikroOrmBaseRepositoryFactory(
...
+    }
+    if (isDefined(productToUpdate.length)) {
+      productToUpdate.length = productToUpdate.length?.toString()
+    }
+    if (isDefined(productToUpdate.height)) {
+      productToUpdate.height = productToUpdate.height?.toString()
+    }
+    if (isDefined(productToUpdate.width)) {
+      productToUpdate.width = productToUpdate.width?.toString()
+    }
```
olivermrbl, line 72: thought: we should probably find a better way to do this at some point 😄

can you create a ticket for it?

olivermrbl (approved):

packages/modules/product/src/repositories/product.ts
fPolic: yup, will open a ticket!

PR #13020, opened 2025-07-23
fix(medusa): fetching a product without a category with `categories` filed passed

fPolic:
**What**
- adding `categories` to `fields` param would return 404 for an existing product if the product was not associated with a category

---

CLOSES CMRC-1046


integration-tests/http/__tests__/product/store/product.spec.ts
```diff
@@ -1928,6 +1928,130 @@ medusaIntegrationTestRunner({
...
+        expect(response.status).toEqual(200)
+        expect(response.data.product).toEqual(
+          expect.objectContaining({
+            id: product.id,
+            categories: [expect.objectContaining({ id: category.id })],
+          })
+        )
+      })
+
+      it("should return 404 for a product in a disabled category", async () => {
```
fPolic: discussion: should a product be returned in this case as well? what about case where a product belongs to multiple categories where some of them are active

olivermrbl: The category being disabled shouldn't affect the "visibility" of the product. In this case, the disabled categories should be omitted from `categories` on the retrieved product.

I.e. specifically in this case, the response would be:
```
id: "prod_1234",
categories: [],
...
```

Similarly when the product belongs to multiple categories, the disabled categories will not be included in `categories` on the product.

I am curious what made you think it should return 404 instead?

fPolic: I agree with you this is why I added the comment. Added the test since this is the current behaviour and I wasn't sure if it was intentional but on second tought looks like an additional bug caused by the same underlying issue.

integration-tests/http/__tests__/product/store/product.spec.ts
```diff
@@ -17,7 +17,7 @@ import {
 import { getProductFixture } from "../../../../helpers/fixtures"
 import { createAuthenticatedCustomer } from "../../../../modules/helpers/create-authenticated-customer"
 
-jest.setTimeout(30000)
+jest.setTimeout(300000)
```
adrien2p: todo: revert?

adrien2p (approved):

willbouch (approved):

PR #13037, opened 2025-07-24
fix(core-flows): guest customer updates email to another guest account

fPolic:
CLOSES CMRC-1026

### DIFF
--- a/integration-tests/modules/__tests__/cart/store/carts.spec.ts
+++ b/integration-tests/modules/__tests__/cart/store/carts.spec.ts
@@ -1843,7 +1843,7 @@ medusaIntegrationTestRunner({
             {
               email: guestsMainEmail, // update with the same mail
               metadata: {
-                test: "test updated 2, new customer",
+                test: "test updated 2, same customer",
               },
             },
             storeHeaders
--- a/integration-tests/modules/__tests__/cart/store/carts.spec.ts
+++ b/integration-tests/modules/__tests__/cart/store/carts.spec.ts
@@ -1730,6 +1730,99 @@ medusaIntegrationTestRunner({
           })
         })
 
+        it("should update the cart email from one guest account to another", async () => {
+          const guestsMainEmail = "guest.main@acme.com"
+          const guestMain = await customerModule.createCustomers({
+            email: guestsMainEmail,
+            has_account: false,
+          })
+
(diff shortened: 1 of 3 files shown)


integration-tests/modules/__tests__/cart/store/carts.spec.ts
```diff
@@ -1730,6 +1730,157 @@ medusaIntegrationTestRunner({
           })
         })
 
+        it("should update the cart email from one guest account to another", async () => {
```
olivermrbl, line 1733: thought: It might also make sense to ensure customer IDs are updated correctly. 

If you update a cart with a new _existing_ customer email, the ID of that existing customer should be on the cart.

If you update a cart with a new _non-existing_ customer email, we should create a new guest customer and the ID of that should be on the cart.

What do you think?

fPolic: > If you update a cart with a new existing customer email, the ID of that existing customer should be on the cart.

Isn't [this](https://github.com/medusajs/medusa/pull/13037/files#diff-1bdfb5c8c6dbd70ecf424afc8eb0dbb0aa3fd95f9699b33194407fe844edf77cR1808-R1809) assertion covering that?

> If you update a cart with a new non-existing customer email, we should create a new guest customer and the ID of that should be on the cart.

Also this is checked in the [test](https://github.com/medusajs/medusa/blob/1e4bf919195b41837109bda06f3154f618e3778a/integration-tests/modules/__tests__/cart/store/carts.spec.ts#L1722) above.

olivermrbl: > Isn't this assertion covering that?

Think the link might be wrong; I am taken to the test asserting that updating the same email works as expected

olivermrbl: > Also this is checked in the test above.

Great, didn't check the other tests; my bad.

olivermrbl: > Think the link might be wrong; I am taken to the test asserting that updating the same email works as expected

Sorry, now I see the link to. This looks good.

olivermrbl (approved):

PR #13038, opened 2025-07-24
Fixed premature teardown in medusa-test-runner

### LINKED ISSUE #12956: [Bug]: medusaTestRunner tear down
dbUtils is provided for the end user to implement their own clean up and teardown per each or wherever in their tests but it seems the test runner is calling teardown after each. I believe this is an error and forces the end users to override the dbUtils teardown method just to avoid premature teardowns between their tests. This also breaks fixtures in beforeAll.

https://github.com/medusajs/medusa/blob/e74044af4de213ba6326a8fb2b5d02e5875b8a4a/packages/medusa-test-utils/src/medusa-test-runner.ts#L285

trevster344:
Fixes #12956


adrien2p:
Hi @trevster344 , thank you for your contribution. Though, I am not sure to understand something here. The teardown is expected to run after each in order to cleanup all the data after each test and prevent collision or remaining data in any follow up  test in order to always start from a clean slate. With your changes it would break all the tests that are not expecting other data to exists during their execution and probably the setup that expects to create data that are not supposed to remain present.

what I think would be interesting here instead is to provide a new option to the runner to prevent running the teardown in the after each in case you want to manage it manually. Something like preventTearingDatabaseDown, Does that sounds right?

just to be absolutely sure, could you explain me your setup to have a better idea of your need or what you are trying to achieve ?

let me know if I am misunderstanding something or missed a piece of information.

trevster344:
Hello @adrien2p, thanks for getting back to me quickly. 

I setup a series of simple tests. One describe block, a beforeAll block and a couple tests. I simply wanted to create some test data before these tests ran. The first test would succeed but subsequent tests in the same describe block would fail due to missing fixture data. I could easily create each test and repeat some of the setup code to do the same. I figured however that this was the intention of beforeAll. I noticed dbUtils had some helpful methods for this and assumed this would not be called unless explicitly stated. Would I be mistaken in my understanding? 

If this is expected behavior and my situation isn't, I'd be more than willing to amend the changes for an options approach instead.

Edit: Also wanted to note that I thought of using beforeEach as well but in my case this greatly slowed down the testing due to the amount of data I am setting up repeatedly. I also noticed the teardown is only a partial teardown as it explicitly ignores certain data.

adrien2p:
Thank you for the information, I think it would make sense to add the option to improve the tool flexibility for people that would like to handle their tests differently indeed. 
But the default behaviour should remain the same, teardown in the after each, why?
Because the idea is that the data of test one should not impact test 2. I believe in your case you might have a lot of data in your hook because of the way you want to test it and therefore you might be creating the data of the entire test suite in the before all. In our mind, tests are working in isolation, each tests is independent and should not leak into other tests.

But +1 to add the option cc @olivermrbl

trevster344:
@adrien2p I figured isolation was the goal but was slightly confused when I noticed the teardown excludes certain data from being torn down. I agree with a configuration option being added for this sort of scenario. I can make that contribution if and when you'd like. Thanks for the input.

leobenzol:
I'd also like to see an option to have the teardown in afterAll, with the default being afterEach. When testing an integration with a 3rd party API currently you can't divide the tests neatly if you want to avoid making the same requests over and over.

adrien2p:
I cant see anyhting anymore in the pr 🤔 but otherwise, I believe it is a good addition to have @olivermrbl ?

willbouch:
@trevster344 Hi and thanks for the contribution! Is there any chance you can take a look at the PR? I cannot seem to find any file change either as Adrien mentionned 😵

trevster344:
@willbouch How about now? Sorry about the delay.

willbouch (commented):
Hey! I believe it might be better to switch it around. The way I see it, auto teardown should be default. In this case, why not have `disableAutoTeardown` with a default of false. It achieves the same but it makes more sense in my opinion

trevster344:
@willbouch Sounds good to me! I'll make the modification.

packages/medusa-test-utils/src/medusa-test-runner.ts
```diff
@@ -45,6 +45,8 @@ interface TestRunnerConfig {
   env?: Record<string, any>
   dbName?: string
   medusaConfigFile?: string
+  disableAutoTeardown?: boolean
+  disableAutoTeardown?: boolean
```
willbouch: Thios looks like duplicate

willbouch (commented):
Also, can you change the name of the PR + the commit name in the changeset to meet "fix(package affected): description"? Thanks :)

packages/medusa-test-utils/src/medusa-test-runner.ts
trevster344: Corrected in following commit. Sorry.

packages/medusa-test-utils/src/medusa-test-runner.ts
```diff
@@ -290,7 +293,11 @@ class MedusaTestRunner {
   public async afterEach(): Promise<void> {
     try {
       await waitWorkflowExecutions(this.globalContainer as MedusaContainer)
-      await this.dbUtils.teardown({ schema: this.schema })
+
+      if (!this?.disableAutoTeardown) {
```
willbouch: Is there really a context where `this` is undefined ? Why adding the `?`

trevster344: It's been sometime. I don't recall there needing to be any context. I'll review this again.

willbouch (dismissed):

packages/medusa-test-utils/src/medusa-test-runner.ts
```diff
@@ -290,7 +293,11 @@ class MedusaTestRunner {
   public async afterEach(): Promise<void> {
     try {
       await waitWorkflowExecutions(this.globalContainer as MedusaContainer)
-      await this.dbUtils.teardown({ schema: this.schema })
+
+      if (!this.disableAutoTeardown) {
+        // Perform automatic teardown
+        await this.dbUtils.teardown({ schema: this.schema })
+      }
```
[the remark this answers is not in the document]
willbouch: @trevster344 I believe this is true here. Nothing seems to be passed to the `medusaIntegrationTestRunner` function, so how can I set the property to true?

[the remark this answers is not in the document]
adrien2p: @trevster344 you are indeed missing the options in the function down this file which instanciate this class and propagate the option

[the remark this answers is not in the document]
trevster344: Let me take a look. I think I must've done something with my git environment to overwrite myself. I noticed a duplication in a previous commit too. I'll edit this comment when I go back over it more thoroughly. 

Edit: I went over it more. I don't see anything else standing out. Seems to be only the exported function was missing the necessary implementation. Updated. Appreciate the patience from you all.

willbouch (approved):

willbouch:
@adrien2p, I'll let you merge once you have reviewed :)

adrien2p (approved):

PR #13053, opened 2025-07-25
fix(dashboard): table inclusive date filter

fPolic:
**What**
- make end date inclusive for search results

---

https://github.com/user-attachments/assets/8367ffd1-e882-44a9-a876-a8e4731a6cd5

---

CLOSES CMRC-1072


willbouch (approved):
Tested and it works. 

Nitpick: I guess it is okay to add 24h because those filters only provide a date and no time, but maybe you could use setHours instead and set it to 23:59:59? That would be more befitting of the comment you left. Because it is a bit unclear from that line only that adding 24h offsets to end of day

olivermrbl (approved):
I assume the value when selecting `lte` is always the beginning of the day?

olivermrbl:
> Nitpick: I guess it is okay to add 24h because those filters only provide a date and no time, but maybe you could use setHours instead and set it to 23:59:59? That would be more befitting of the comment you left. Because it is a bit unclear from that line only that adding 24h offsets to end of day

I didn't see this comment, but I agree. If possible, this would make it more obvious

fPolic:
> I assume the value when selecting lte is always the beginning of the day?

Yes it's 00:00 in that timezone

> That would be more befitting of the comment you left

Valid point - changed offsetting to use `setHours`

olivermrbl (approved):

PR #13064, opened 2025-07-28
chore(framework): rm internal container types

carlos-r-l-rodrigues:
What:
 - Remove internal types from Medusa container types

### DIFF
--- a/.changeset/angry-terms-fry.md
+++ b/.changeset/angry-terms-fry.md
@@ -1,0 +1,5 @@
+---
+"@medusajs/framework": patch
+---
+
+Remove deprecated registration keys from container type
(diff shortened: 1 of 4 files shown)


willbouch (approved):
Would like to know why if you can give me a bit of context :)

olivermrbl:
> Would like to know why if you can give me a bit of context :)

+1

carlos-r-l-rodrigues:
> > Would like to know why if you can give me a bit of context :)
> 
> +1

Converted it to Draft, because this PR will require a much bigger work on the types, than just that.
The idea is to display only what is available/registered for each Module's container. Not even sure if this is possible to infer in many cases where things are registered on loaders.

adrien2p (commented):
From my undersatanding of the needs, it is not currently possible. The module dependencies are managed at runtime and we can really convert that easily to typings. On the other hand, if the issue is that we show all the available module, maybe what we can do is not type the container resolution and provide a new parameter to the module option being the list of the available resolution keys? wdyt? If my explanation are not clear, we can discuss over a call and then write it down in a better way 👍

PR #12960, opened 2025-07-14
This fixes the discount_ calculation logic and promotion tax inclusiveness calculation

scherddel:
This PR addresses a bug where the timing to consider the is_tax_inclusive parameter was not right when calculating promotion adjustments. The issue was caused by applying the tax_inclusive calculation too late in the process

Changes:

- Refactored is_tax_inclusive calculation logic in the "promotion to adjustment" process to consitently make adjustments "subtotal" of a promotion amount considering the "is_tax_inclusive" flag as all calculations performed to apply promotions are based on item subtotals 
- Refactored adjustments/index.ts for improved readability
- Removed is_tax_inclusive from cart item line adjustements to reduce redundancy and only keep the flag in the promotion object
- Refactored test cases in promotion and cart to correct misscalculations causes by the mentioned issues

**Known Issue:**
This fix reveals an additional issue that requires follow-up work: In tax-inclusive contexts, promotion adjustments are incorrectly calculated using the line item's subtotal rather than the full price.
For example, with an item priced at 119 (tax-inclusive) and 19% tax, the adjustment calculation (of a 119 adjustment) uses 100 (subtotal) as the basis instead of the full 119 price. This results in incorrect promotion amounts and needs to be addressed in a future PR.

Edit: Also fixes the "Known Issue": 
- Adjustments are now subtotals only which is in line with the implementation of the promotion logic (also only applies subtotals)
- Cleaned up some code to increase readability

Edit2: Due to some concerns raised by oliver I have refactored the solution to use the same logic as promotions and the amount of adjustments now the same but considers different totals based on the is_tax_inclusive flag

PR #13070, opened 2025-07-28
fix(medusa): include tax inclusive adjustment flag in default cart query config

fPolic:
**What**
- default cart query config wasn't including `is_tax_inclusive` flag for `items.adjustments` so depending on the fileds params in some cases the filed wasn't properly passed to the totals utils


carlos-r-l-rodrigues (approved):

olivermrbl:
this seems empty; should we close?

fPolic:
yes, fix was added in the meantime [here](https://github.com/medusajs/medusa/pull/12960/files#diff-af0b76e29bd88a47a376268479d40743e24e5bd39a7726a917a19d24c991642dR81)

PR #13076, opened 2025-07-29
 fix(core-flows): pass backorder flag when recreating reservations after fulfilment cancelation

fPolic:
**What**
- pass backorder flag recreating reservations after fulfillment cancelation

---

CLOSES SUP-1913


integration-tests/http/__tests__/order/admin/order.spec.ts
```diff
@@ -1571,6 +1589,212 @@ medusaIntegrationTestRunner({
...
+        )
+
+        reservations = (
+          await api.get(
+            `/admin/reservations?line_item_id[]=${lineItemId}`,
+            adminHeaders
+          )
+        ).data.reservations
+
+        expect(reservations).toEqual(
```
adrien2p, line 1783: Q: If the fulfilled order is cancelled, the items are still reserved? How can we unreserved them?

adrien2p (approved):
just one question, LGTM 🎉

integration-tests/http/__tests__/order/admin/order.spec.ts
fPolic: Presume you mean when fulfillment is cancelled here....in that case, we want to keep items reserved for that order so the order can be fulfilled again. A user can cancel the order or manually delete reservations to free the inventory.

Issue #12669, opened 2025-06-05
[Bug]: Promotions matching on product tags do not seem to be working


mortenengel:
### Package.json file

```JSON
{
  "name": "medusa-starter-default",
  "version": "0.0.1",
  "description": "A starter for Medusa projects.",
  "author": "Medusa (https://medusajs.com)",
  "license": "MIT",
  "keywords": [
    "sqlite",
    "postgres",
    "typescript",
    "ecommerce",
    "headless",
    "medusa"
  ],
  "scripts": {
    "build": "medusa build",
    "seed": "medusa exec ./src/scripts/seed.ts",
    "start": "medusa start",
    "dev": "medusa develop",
    "migrate": "medusa db:migrate",
    "test:integration:http": "TEST_TYPE=integration:http NODE_OPTIONS=--experimental-vm-modules jest --silent=false --runInBand --forceExit",
    "test:integration:modules": "TEST_TYPE=integration:modules NODE_OPTIONS=--experimental-vm-modules jest --silent --runInBand --forceExit",
    "test:unit": "TEST_TYPE=unit NODE_OPTIONS=--experimental-vm-modules jest --silent --runInBand --forceExit",
    "generate-wsdl-example": "wsdl-tsclient ./src/modules/drl-stock/generated/WSItemStock.wsdl -o ./src/modules/drl-stock/generated"
  },
  "dependencies": {
    "@azure/monitor-opentelemetry-exporter": "^1.0.0-beta.28",
    "@ariakit/react": "^0.4.15",
    "@azure/storage-blob": "^12.26.0",
    "@medusajs/admin-sdk": "2.8.3",
    "@medusajs/cli": "2.8.3",
    "@medusajs/framework": "2.8.3",
    "@medusajs/medusa": "2.8.3",
    "@mikro-orm/core": "6.4.3",
    "@mikro-orm/knex": "6.4.3",
    "@mikro-orm/migrations": "6.4.3",
    "@mikro-orm/postgresql": "6.4.3",
    "@opentelemetry/instrumentation-pg": "^0.51.1",
    "@opentelemetry/resources": "^1.30.1",
    "@opentelemetry/sdk-node": "^0.57.2",
    "@opentelemetry/sdk-trace-node": "^1.30.1",
    "@relewise/client": "^2.4.0",
    "@relewise/integrations": "^1.0.2",
    "@tiptap/extension-color": "^2.11.5",
    "@tiptap/pm": "^2.11.5",
[...]

PR #13078, opened 2025-07-29
Made in operator work as In insted of equal logic

mikkel-lindstrom:
Changed the In operator to actually work like an In instead of being same logic as Equals. This means that in promotions you can add a rule to apply when a product is in a category or multiple different. Before the logic had to match all the products categories to apply, which doesnt really make sense when you have nested category structure on products. The logic also applies to tags where you can make a rule apply based on a tag before it also had to match all tags on a product to apply.

Issue:

https://github.com/medusajs/medusa/issues/12669


olivermrbl (approved):
LGTM, thanks for the contribution.

Issue #13074, opened 2025-07-29
[Bug]: [Price List] : Product price is applied from earliest price list (by created_at) that contains the product, not the cheapest


Attachai77:
### Package.json file

```JSON
{
	"name": "aaa",
	"version": "1.5.1",
	"description": "A backend fo e-commerce platform",
	"author": "AAA",
	"license": "MIT",
	"keywords": [
		"sqlite",
		"postgres",
		"typescript",
		"ecommerce",
		"headless",
		"medusa",
		"utech",
		"it",
		"gadget",
		"backend",
		"typescript"
	],
	"scripts": {
		"build": "medusa build",
		"seed": "medusa exec ./src/scripts/seed.ts",
		"migrations": "npx medusa db:migrate",
		"sync": "npx medusa db:sync-links",
		"start": "medusa start",
		"dev": "medusa develop",
		"build-start": "medusa build && medusa start",
		"test:integration:http": "TEST_TYPE=integration:http NODE_OPTIONS=--experimental-vm-modules jest --silent=false --runInBand --forceExit",
		"test:integration:modules": "TEST_TYPE=integration:modules NODE_OPTIONS=--experimental-vm-modules jest --silent --runInBand --forceExit",
		"test:unit": "TEST_TYPE=unit NODE_OPTIONS=--experimental-vm-modules jest --silent --runInBand --forceExit",
		"test:coverage": "jest --coverage",
		"check:changed": "node ./scripts/check-changed.js",
		"format:changed": "node ./scripts/check-changed.js format",
		"lint:changed": "node ./scripts/check-changed.js lint",
		"prepare": "husky"
	},
	"dependencies": {
		"@medusajs/admin-sdk": "2.8.7",
		"@medusajs/cli": "2.8.7",
		"@medusajs/framework": "2.8.7",
		"@medusajs/medusa": "2.8.7",
		"@mikro-orm/core": "6.4.3",
		"@mikro-orm/knex": "6.4.3",
		"@mikro-orm/migrations": "6.4.3",
		"@mikro-orm/postgresql": "6.4.3",
		"@opentelemetry/api": "^1.9.0",
		"@opentelemetry/exporter-trace-otlp-grpc": "^0.57.2",
		"@opentelemetry/sdk-node": "^0.57.2",
		"@sentry/node": "9.11",
		"@sentry/opentelemetry-node": "^7.114.0",
		"@zimpligital/medusa-plugin-auth-otp": "^1.2.9",
[...]

Issue #13034, opened 2025-07-24
[Bug]: Price calculation fails when customer is associated with multiple customer groups


rubenalpha:
### Package.json file

```JSON
{
  "name": "medusa-b2b-starter",
  "version": "0.0.1",
  "description": "A starter for Medusa B2B projects.",
  "author": "Medusa (https://medusajs.com)",
  "license": "MIT",
  "keywords": [
    "sqlite",
    "postgres",
    "typescript",
    "ecommerce",
    "headless",
    "medusa"
  ],
  "scripts": {
    "build": "medusa build",
    "seed": "medusa exec ./src/scripts/seed.ts",
    "start": "medusa start",
    "dev": "medusa develop",
    "test:integration:http": "TEST_TYPE=integration:http NODE_OPTIONS=--experimental-vm-modules jest --silent=false --runInBand --forceExit",
    "test:integration:modules": "TEST_TYPE=integration:modules NODE_OPTIONS=--experimental-vm-modules jest --silent --runInBand --forceExit",
    "test:unit": "TEST_TYPE=unit NODE_OPTIONS=--experimental-vm-modules jest --silent --runInBand --forceExit"
  },
  "dependencies": {
    "@medusajs/admin-sdk": "2.4.0",
    "@medusajs/cli": "2.4.0",
    "@medusajs/framework": "2.4.0",
    "@medusajs/js-sdk": "2.4.0",
    "@medusajs/medusa": "2.4.0",
    "@medusajs/ui": "4.0.0",
    "@mikro-orm/core": "6.4.3",
    "@mikro-orm/knex": "6.4.3",
    "@mikro-orm/migrations": "6.4.3",
    "@mikro-orm/postgresql": "6.4.3",
    "@sentry/browser": "^9.2.0",
    "@sentry/vite-plugin": "^3.2.1",
    "algoliasearch": "^4",
    "awilix": "^8.0.1",
    "axios": "^1.7.9",
    "cors": "^2.8.5",
    "csv-parser": "^3.1.0",
    "globalpayments-api": "^3.10.3",
    "multer": "^1.4.5-lts.1",
    "pg": "^8.13.0",
    "zod": "3.22.4"
  },
  "devDependencies": {
    "@medusajs/test-utils": "2.4.0",
    "@medusajs/ui-preset": "latest",
    "@mikro-orm/cli": "6.4.3",
    "@swc/core": "1.5.7",
    "@swc/jest": "^0.2.36",
    "@types/cors": "^2",
[...]

PR #13079, opened 2025-07-29
fix(pricing): Calculate prices with multiple rule values

olivermrbl:
### What

This PR fixes two issues described in #13074 and #13034.

**Price list price priority**

Up until now, if multiple price list prices of type override were part of the result of calculating prices, we would always pick the most recently created price, which is not necessarily the cheapest price.

This PR changes this logic to always return the cheapest price of all the price list prices in the result set. We do this by sorting those prices by amount instead of creation date.

**Multiple rule values**

Up until now, if the pricing context contained an array of values for one of its keys, it would break the calculation logic in the pricing repository and lead to an empty result set.

For example, let's imagine the following: 
- Customer has a registered account
- Two customer groups exist: Customer Group 1 and Customer Group 2
- Customer is part of both customer groups
- Two price lists exist: Price List 1 and Price List 2
- Price List 1 has a rule limiting pricing to customers of Customer Group 1
- Price List 2 has a rule limiting pricing to customers of Customer Group 2

In this setup, the registered customer has "access" to prices in both price lists. We want to show the customer the cheapest price across these.

However, the way the pricing context for this customer is transformed to SQL leaves us with the following subquery:
```
(plr.attribute = 'customer.groups.id' AND plr.value @> '["cusgr_1234","cusgr_4321"]')
```

This means we are looking for prices that strictly have _both_ customer groups defined as a rule. 

Because none of our price list have both customer groups assigned to them, no prices will be returned.

Instead, what we actually want to do is query prices for each of the customer groups:
```
(plr.attribute = 'customer.groups.id' AND plr.value @> '["cusgr_4321"]')
OR
(plr.attribute = 'customer.groups.id' AND plr.value @> '["cusgr_1234"]')
```

This PR changes the calculation logic to solve for this scenario.


carlos-r-l-rodrigues (approved):
LGTM

PR #13084, opened 2025-07-30
fix(promotion): check currency when computing actions for promotions

riqwan:
what:

- fixing regression that skipped checking for currency when computing actions for promotions

RESOLVES CMRC-1079

### DIFF
--- a/packages/modules/promotion/integration-tests/__tests__/services/promotion-module/compute-actions.spec.ts
+++ b/packages/modules/promotion/integration-tests/__tests__/services/promotion-module/compute-actions.spec.ts
@@ -2694,6 +2694,7 @@ moduleIntegrationTestRunner({
             })
 
             const result = await service.computeActions(["PROMOTION_TEST"], {
+              currency_code: "usd",
               customer: {
                 customer_group: {
                   id: "VIP",
@@ -2743,6 +2744,7 @@ moduleIntegrationTestRunner({
             })
 
             const result = await service.computeActions(["PROMOTION_TEST"], {
+              currency_code: "usd",
               customer: {
                 customer_group: {
                   id: "VIP",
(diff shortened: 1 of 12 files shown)


packages/modules/promotion/src/services/promotion-module.ts
```diff
@@ -520,13 +520,17 @@ export default class PromotionModuleService
         continue
       }
 
+      const isCurrencyCodeValid =
+        !isDefined(applicationMethod.currency_code) ||
```
olivermrbl, line 524: nitpicky: Wouldn't it be more appropriate to check the type of application method here?

If it is of type `fixed` and currency codes match, it is a valid code. If this is not true, we are in a bad state and should probably throw? Because the currency code should be there in this case.

Let me know what you think.

packages/modules/promotion/integration-tests/__tests__/services/promotion-module/compute-actions.spec.ts
```diff
@@ -2694,6 +2694,7 @@ moduleIntegrationTestRunner({
             })
```
olivermrbl, line 2694: question: It is not obvious why we now have to explicitly pass `currency_code` and `is_discountable`; can you explain?

packages/modules/promotion/src/services/promotion-module.ts
riqwan: I thought about this, but remembered the case of percentage promotions that will need to have a currency_code when adding it to a campaign. In that case its no longer about application method type

olivermrbl: Got it

packages/modules/promotion/integration-tests/__tests__/services/promotion-module/compute-actions.spec.ts
riqwan: Its part of the required type, just added them to remove the silent TS fails. No idea why this was manadated though.

olivermrbl (approved):

PR #13086, opened 2025-07-30
feat: added pacific franc currency

tehaulp:
Trying to add the Pacific Franc (XPF) currency in medusajs since the country is listed but not its currency.


sradevski (approved):

tehaulp:
anyone for this ?

PR #13092, opened 2025-07-30
chore(ui): move toast to top right

fPolic:
CLOSES CMRC-1043


willbouch (approved):
:shipit:

packages/design-system/ui/src/utils/toast.tsx
```diff
@@ -68,7 +74,9 @@ function info(
   /**
    * The props of the toast.
    */
-  props: VariantToastProps = {}
+  props: VariantToastProps = {
+    position: "top-right",
```
adrien2p: nit: should we create a DEFAULT_TOAST_POSITION = 'top-right' at the top of the file and use it? I am just thinking that if we ever decide to change it again it would be easier to change it in one place. Dont have strong opinion though

adrien2p (approved):

willbouch (approved):
LGTM. What do we do about UI changes like this? We just move the buttons and hope the users find them? Do we do any communication? I am genuinely curious 😆

fPolic:
The majority of forms already have buttons in the footer, so this should make the buttons easier to find now that all forms are consistent :)

olivermrbl (approved):

PR #13094, opened 2025-07-30
chore(dashboard): add missing US states

fPolic:
CLOSES CMRC-1025


packages/admin/dashboard/src/lib/data/country-states.ts
```diff
@@ -1227,6 +1227,7 @@ const countryProvinceMap: Record<string, CountryProvinceObject> = {
       "US-CA": "California",
```
olivermrbl, line 1227: thought: now, the list contains 52 states, which doesn't seem right

a quick grok on this tell me this: 
> The list you provided includes 50 states of the United States, but it contains two inaccuracies:
>
> 1. **Puerto Rico ("US-PR")** is not a U.S. state; it is a U.S. territory.
> 2. **Washington, D.C. ("US-DC")** is not a state; it is a federal district.
>
> - Your list includes all 50 states except for two incorrect entries (Puerto Rico and Washington, D.C.).
> - To correct your list, remove "US-PR": "Puerto Rico" and "US-DC": "Washington DC".
> - The remaining 50 entries correctly represent the 50 U.S. states.
>
> If you need further clarification or assistance, let me know!


Could you double check this?

fPolic: Checked the states before opening PR, but on second thought, should the list be based on this: https://en.wikipedia.org/wiki/ISO_3166-2:US, which means we need to add a few missing territories to the list.

olivermrbl: Yeah I think all of these should be included

olivermrbl (approved):

PR #13095, opened 2025-07-30
fix(dashboard): order list fixes

fPolic:
**What**
- fix order list country flag
- fix totals cell edge case

---

CLOSES CMRC-1064

### DIFF
--- a/packages/admin/dashboard/src/routes/orders/order-list/const.ts
+++ b/packages/admin/dashboard/src/routes/orders/order-list/const.ts
@@ -10,7 +10,11 @@ const DEFAULT_PROPERTIES = [
   "currency_code",
 ]
 
-const DEFAULT_RELATIONS = ["*customer", "*sales_channel"]
+const DEFAULT_RELATIONS = [
+  "*customer",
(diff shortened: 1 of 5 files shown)


packages/admin/dashboard/src/routes/orders/order-list/const.ts
```diff
@@ -10,7 +10,11 @@ const DEFAULT_PROPERTIES = [
   "currency_code",
 ]
 
-const DEFAULT_RELATIONS = ["*customer", "*sales_channel"]
+const DEFAULT_RELATIONS = [
+  "*customer",
+  "*sales_channel",
+  "+shipping_address.country_code",
```
willbouch, line 16: Probably a noob question but what is the difference between * and + here?

willbouch (approved):
Tested and looks good

olivermrbl:
@fPolic should we close this in favor of the customizable table that Seb is adding? 

If there is a fix in this PR, we can separate it into its own PR.

fPolic:
yes closing in favour of Seb's PR

PR #13106, opened 2025-07-31
fix(types,utils,promotion): Move from total to original_total to resolve edge case for adjustments calculation

scherddel:
Discovered an edge case in the promotion adjustment logic that occurs when:

1. A promotion is applied that reduces the cart total to zero
2. The same promotion code is applied again subsequently
3. The system incorrectly removes the adjustment instead of handling the duplicate application

This scenario is difficult to catch in standard test cases since it requires the specific sequence of applying identical promotions after reaching a zero total.

Solution:
Moved calculation logic from total to original_total to ensure consistent base values


riqwan:
Thanks @scherddel, I think this makes sense to me. Can you add a test that describes this case?

scherddel:
@riqwan Will do!

scherddel:
@riqwan Added a test. All works fine now.

olivermrbl (approved):

PR #13111, opened 2025-07-31
fix(dashboard): create product selected inventory item display

fPolic:
CLOSES CMRC-1081


olivermrbl (commented):
Could I get you to explain what was wrong, and how this fixes it? It is not super obvious from the changes.

fPolic:
> Could I get you to explain what was wrong, and how this fixes it? It is not super obvious from the changes.

When an item is selected, the `inventory_item` field array is not updated i.e. `inventory_item.inventory_item_id` is still `undefined` so Combobox component didn't fetch the selected item for display.

`useWatch` is more reliable here since it will always provide the selected value from the form state to `Combobox`.

willbouch (approved):

PR #13140, opened 2025-08-04
chore(core-flows): throw error on invalid promo code

willbouch:
**What**: Added a mechanism to throw an error when applying an invalid promo code
**Why**: This improves the developer experience when building storefronts by allowing them to handle invalid discount codes directly from the API response, rather than checking cart promotions after the API request.
**How**: When replacing or adding a promo code, check that the code exists in the database and if not, throw an error
**Testing**: integration tests

CLOSES: CMRC-1020

### DIFF
--- a/.changeset/ninety-kangaroos-grin.md
+++ b/.changeset/ninety-kangaroos-grin.md
@@ -1,0 +1,5 @@
+---
+"@medusajs/core-flows": minor
+---
+
+chore(code-flows): throw error on invalid promo code
--- a/integration-tests/http/__tests__/cart/store/cart.spec.ts
+++ b/integration-tests/http/__tests__/cart/store/cart.spec.ts
@@ -2815,6 +2815,25 @@ medusaIntegrationTestRunner({
           )
         })
 
+        it("should throw an error when adding a promotion that does not exist", async () => {
+          const invalidPromoCode = "SOME_INVALID_PROMO_CODE"
+
+          try {
+            await api.post(
+              `/store/carts/${cart.id}/promotions`,
+              { promo_codes: [invalidPromoCode] },
+              storeHeaders
+            )
+            fail()
+          } catch (e) {
+            expect(e.response.status).toEqual(400)
+            expect(e.response.data.type).toEqual("invalid_data")
+            expect(e.response.data.message).toEqual(
+              `The promotion code ${invalidPromoCode} is invalid`
+            )
+          }
+        })
+
         it("should remove promotion adjustments when promotion is deleted", async () => {
           let cartBeforeRemovingPromotion = (
             await api.get(`/store/carts/${cart.id}`, storeHeaders)
--- a/integration-tests/http/__tests__/cart/store/cart.spec.ts
+++ b/integration-tests/http/__tests__/cart/store/cart.spec.ts
@@ -2818,20 +2818,19 @@ medusaIntegrationTestRunner({
         it("should throw an error when adding a promotion that does not exist", async () => {
           const invalidPromoCode = "SOME_INVALID_PROMO_CODE"
 
-          try {
-            await api.post(
+          const { response } = await api
+            .post(
               `/store/carts/${cart.id}/promotions`,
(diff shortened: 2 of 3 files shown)


integration-tests/http/__tests__/cart/store/cart.spec.ts
```diff
@@ -2815,6 +2815,25 @@ medusaIntegrationTestRunner({
           )
         })
 
+        it("should throw an error when adding a promotion that does not exist", async () => {
```
willbouch, line 2818: I wasn't too sure about where to add this test because I am not calling quite exactly the endpoint that is defined in the `describe`. New .spec file inside a sub promotion directory?
Screenshot 2025-08-04 at 1 21 32 PM

Or a new sub describe thats adds a `/promotions`?

packages/core/core-flows/src/cart/steps/get-promotion-codes-to-apply.ts
```diff
@@ -94,17 +96,38 @@ export const getPromotionCodesToApply = createStep(
...
+    }
+
+    if (
+      action === PromotionActions.ADD ||
+      action === PromotionActions.REPLACE
+    ) {
+      const validPromoCodes: Set<string> = new Set(
+        promo_codes.length
+          ? (
+              await promotionService.listPromotions(
```
willbouch, line 114: There is a way to make this call happen in parallel with the call on line 91 but the code will become a bit harder to read in my opinion. I was not too sure about what you prioritize. Feel free to let me know if it would be better to add the call in //

packages/core/core-flows/src/cart/steps/get-promotion-codes-to-apply.ts
```diff
@@ -94,17 +96,38 @@ export const getPromotionCodesToApply = createStep(
...
+                { code: promo_codes },
+                { select: ["code"] }
+              )
+            ).map((p) => p.code!)
+          : []
+      )
+
+      promo_codes.forEach((code) => {
+        if (!validPromoCodes.has(code)) {
+          throw new MedusaError(
```
willbouch, line 124: I fail immediately. Let me know if it would make more sense to accumulare multiple errors to throw all at once if multiple invalid discount codes were added

.changeset/ninety-kangaroos-grin.md
```diff
@@ -0,0 +1,5 @@
+---
+"@medusajs/core-flows": minor
```
willbouch, line 2: breaking change so I put minor but let me know if it should be patch

integration-tests/http/__tests__/cart/store/cart.spec.ts
```diff
@@ -2815,6 +2815,25 @@ medusaIntegrationTestRunner({
...
+            fail()
+          } catch (e) {
+            expect(e.response.status).toEqual(400)
+            expect(e.response.data.type).toEqual("invalid_data")
+            expect(e.response.data.message).toEqual(
+              `The promotion code ${invalidPromoCode} is invalid`
+            )
+          }
+        })
+
```
carlos-r-l-rodrigues: suggestion: you can grab the error and check it outside the try catch.

`const { response } = await api.post(...).catch((e) => e)`

generally speaking, ideally we avoid using `try/catch` on tests to avoid hiding errors. And we don't have to safeguard it with `fail`

integration-tests/http/__tests__/cart/store/cart.spec.ts
carlos-r-l-rodrigues: It's fine to keep it here.

carlos-r-l-rodrigues (approved):
LGTM
🔥

integration-tests/http/__tests__/cart/store/cart.spec.ts
willbouch: That would explain why I didn't find fail anywhere in the repo haha

olivermrbl (approved):

PR #13145, opened 2025-08-05
chore(util): avoid promotion value precision limit

carlos-r-l-rodrigues:
What:
 - Removes precision limit from promotion calculations
 - introduce `MEDUSA_EPSILON` environment variable to use as comparison when a small number needs to be considered 0.

### DIFF
--- a/.changeset/clever-deers-shout.md
+++ b/.changeset/clever-deers-shout.md
@@ -1,0 +1,5 @@
+---
+"@medusajs/utils": patch
+---
+
+chore(utils): avoid limit precision for promotion value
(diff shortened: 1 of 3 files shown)


willbouch (commented):
LGTM but will let another review for now. Just for my curiosity, can you explain why? What happened for us to do that?

carlos-r-l-rodrigues:
> LGTM but will let another review for now. Just for my curiosity, can you explain why? What happened for us to do that?

[This test](https://github.com/medusajs/medusa/blob/develop/packages/modules/promotion/integration-tests/__tests__/services/promotion-module/compute-actions.spec.ts#L3522) was applying a promotion where the amount was `2e-18`

olivermrbl (approved):

PR #13150, opened 2025-08-05
fix(dashboard): variants disappearing when removing an option on product creation

willbouch:
**What**: Fixed 2 issues for variants on product selection:
1. When creating a second option and setting values, removing those values caused the product variants to vanish from the table below
2. When creating a second option and setting values, deleting the option caused the variants to be unselected and even caused problems in the console upon product creation

**Why**: Quirky behaviour otherwise
**How**: Small tweaks in the code. See comments below for more details
**Testing**: spinned up the dashboard locally and manually tested (see linked tickets for videos of buggy behaviour)

CLOSES: CMRC-1083
CLOSES: CMRC-1084

### DIFF
--- a/packages/admin/dashboard/src/routes/products/product-create/components/product-create-details-form/components/product-create-details-variant-section/product-create-details-variant-section.tsx
+++ b/packages/admin/dashboard/src/routes/products/product-create/components/product-create-details-form/components/product-create-details-variant-section/product-create-details-variant-section.tsx
@@ -332,10 +332,18 @@ export const ProductCreateVariantsSection = ({
                       )}
                       <ul className="flex flex-col gap-y-4">
                         {options.fields.map((option, index) => {
+                          const hasError =
+                            !!form.formState.errors.options?.[index]
                           return (
                             <li
                               key={option.id}
-                              className="bg-ui-bg-component shadow-elevation-card-rest grid grid-cols-[1fr_28px] items-center gap-1.5 rounded-xl p-1.5"
+                              className={clx(
+                                "bg-ui-bg-component shadow-elevation-card-rest grid grid-cols-[1fr_28px] items-center gap-1.5 rounded-xl p-1.5",
+                                {
+                                  "border-ui-border-error shadow-borders-error":
+                                    hasError,
+                                }
+                              )}
                             >
                               <div className="grid grid-cols-[min-content,1fr] items-center gap-1.5">
                                 <div className="flex items-center px-2 py-1.5">
--- a/packages/admin/dashboard/src/routes/products/product-create/components/product-create-details-form/components/product-create-details-variant-section/product-create-details-variant-section.tsx
+++ b/packages/admin/dashboard/src/routes/products/product-create/components/product-create-details-form/components/product-create-details-variant-section/product-create-details-variant-section.tsx
@@ -3,6 +3,7 @@ import {
   Alert,
   Button,
   Checkbox,
+  clx,
   Heading,
   Hint,
   IconButton,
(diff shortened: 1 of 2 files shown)


packages/admin/dashboard/src/routes/products/product-create/components/product-create-details-form/components/product-create-details-variant-section/product-create-details-variant-section.tsx
```diff
@@ -104,7 +104,9 @@ export const ProductCreateVariantsSection = ({
     const newOptions = [...watchedOptions]
     newOptions[index].values = value
 
-    const permutations = getPermutations(newOptions)
+    const permutations = getPermutations(
```
willbouch, line 107: Issue here was that when removing every value for an option, the array of values becomes naturally `[]`. So, when trying to get the permutations between `[Red, Blue]` and `[]`, the result is also naturally `[]`. This was causing the issue. So, to fix it, I simply removed the empty arrays when getting the permutations, so, the result is essentially getting the permutations for only `[Red, Blue]`, which will end up being itself

packages/admin/dashboard/src/routes/products/product-create/components/product-create-details-form/components/product-create-details-variant-section/product-create-details-variant-section.tsx
```diff
@@ -157,46 +159,29 @@ export const ProductCreateVariantsSection = ({
...
     newOptions.splice(index, 1)
+    const validOptionTitles = new Set(newOptions.map((option) => option.title))
 
     const permutations = getPermutations(newOptions)
     const oldVariants = [...watchedVariants]
 
-    const findMatchingPermutation = (options: Record<string, string>) => {
-      return permutations.find((permutation) =>
-        Object.keys(options).every((key) => options[key] === permutation[key])
+    const newVariants = permutations.reduce((variants, permutation) => {
```
willbouch, line 167: This resolves CRC-1084:
The issue here is that the `match` was always undefined because when trying to find the matching permutation, we are essentially comparing apples and oranges. The permutations will not have the deleted option inside, but the `oldVariants` will. So, `Object.keys(options).every((key) => options[key] === permutation[key])` always returns false. So, I simply changed the comparison to only compare the `validOptionTitles`, which are essentially all the options - the deleted one. I switched the reduce to do it on permutations isntead of `oldVariants` to prevent duplicates

olivermrbl:
I came across something very nitpicky, that I thought would make sense to fix as well. If you try to continue without adding values to an option, you can't (for good reason). 

The current flow works, but it would be nice if we could highlight the input field of the option missing the evalues:

https://github.com/user-attachments/assets/7e445a93-3c53-4e95-9275-a6d7fc8c4ffe


Having said that, if this is not a quick addition, I wouldn't spend time on it.

olivermrbl (approved):
This LGTM, I like the comments explaining the change. Helps a lot!

willbouch:
> I came across something very nitpicky, that I thought would make sense to fix as well. If you try to continue without adding values to an option, you can't (for good reason).
> 
> The current flow works, but it would be nice if we could highlight the input field of the option missing the evalues:
> 
>  CleanShot.2025-08-06.at.12.33.54.mp4 
> Having said that, if this is not a quick addition, I wouldn't spend time on it.

Screenshot 2025-08-06 at 8 25 55 AM

Would this work for you? It is a small addition. The behaviour currently is that the message "Please create at least one option." does not disappear when you input a value, so I would keep the same behaviour for the border. In other words, the border will and error message will remain there until the user clicks Continue again. Does this work for you? I pushed the code if you want to test it out

olivermrbl:
@willbouch, think this is a great addition. Nice one.

olivermrbl:
Feel free to merge as you please

PR #13151, opened 2025-08-05
fix(workflow-engine-inmemory): fix cron job schedule

carlos-r-l-rodrigues:
What:
 - Fix next timeout schedule. Sometimes the calculation can return negative numbers, executing the same job twice in a row.

RESOLVES FRMW-2988

### DIFF
--- a/packages/modules/workflow-engine-inmemory/src/utils/workflow-orchestrator-storage.ts
+++ b/packages/modules/workflow-engine-inmemory/src/utils/workflow-orchestrator-storage.ts
@@ -32,13 +32,13 @@ function calculateDelayFromExpression(expression: CronExpression): number {
   const nextTime = expression.next().getTime()
   const now = Date.now()
   const delay = nextTime - now
-  
-  // If the calculated delay is negative or very small, get the next occurrence
+
+  // If the calculated delay is negative or zero, get the next occurrence
   if (delay <= 0) {
     const nextNextTime = expression.next().getTime()
     return Math.max(1, nextNextTime - now)
   }
-  
+
   return delay
 }
 
--- a/packages/modules/workflow-engine-inmemory/src/utils/workflow-orchestrator-storage.ts
+++ b/packages/modules/workflow-engine-inmemory/src/utils/workflow-orchestrator-storage.ts
@@ -28,27 +28,41 @@ import { WorkflowOrchestratorService } from "@services"
 import { type CronExpression, parseExpression } from "cron-parser"
 import { WorkflowExecution } from "../models/workflow-execution"
 
+function calculateDelayFromExpression(expression: CronExpression): number {
+  const nextTime = expression.next().getTime()
+  const now = Date.now()
+  const delay = nextTime - now
+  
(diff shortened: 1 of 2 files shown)


packages/modules/workflow-engine-inmemory/src/utils/workflow-orchestrator-storage.ts
```diff
@@ -28,27 +28,41 @@ import { WorkflowOrchestratorService } from "@services"
 import { type CronExpression, parseExpression } from "cron-parser"
 import { WorkflowExecution } from "../models/workflow-execution"
 
+function calculateDelayFromExpression(expression: CronExpression): number {
+  const nextTime = expression.next().getTime()
+  const now = Date.now()
+  const delay = nextTime - now
+  
+  // If the calculated delay is negative or very small, get the next occurrence
```
willbouch: comment says "or very small" but the if condition does not fit or I am missing something? By very small you mean 0.000001?

willbouch: If it is, then maybe it would be a good idea to have <= EPSILON (the one you introduced in another PR)

carlos-r-l-rodrigues: Here we are only handling integers. Let me update the comment to not make it dubious hehe.

willbouch (commented):
Again, LGTM but will let another one review for now

packages/modules/workflow-engine-inmemory/src/utils/workflow-orchestrator-storage.ts
```diff
@@ -28,27 +28,41 @@ import { WorkflowOrchestratorService } from "@services"
...
 import { WorkflowExecution } from "../models/workflow-execution"
 
+function calculateDelayFromExpression(expression: CronExpression): number {
+  const nextTime = expression.next().getTime()
+  const now = Date.now()
+  const delay = nextTime - now
+
+  // If the calculated delay is negative or zero, get the next occurrence
+  if (delay <= 0) {
+    const nextNextTime = expression.next().getTime()
```
olivermrbl, line 38: q: I presume the `expression.next()` advances time of `expression` and saves this state internally? Such that the next time it is called, it starts from n+1. 

Otherwise, I would have expected this to be something like `expression.next().next()`.

olivermrbl (approved):
LGTM

packages/modules/workflow-engine-inmemory/src/utils/workflow-orchestrator-storage.ts
carlos-r-l-rodrigues: Yes, every time next is called, it advances in time internally. This was sometimes returning the same from the previous call. I just assumed it is a bug from `cron-parser`. we could try to bump it's version instead of this fix if you prefer.

olivermrbl: Got it. Fine to keep this as a safe guard!

Issue #8548, opened 2024-08-11
Deleting an image does not trigger the delete method of AbstractFileProviderService


SummerFans:
## Bug report
During the development of the image service module, I found that the delete method in the AbstractFileProviderService type could not be executed, which resulted in the inability to delete images stored by the third-party image service provider.### Describe the bug

### System information

Medusa version (including plugins): 
- @medusajs/medusa(1.20.6-preview-20240809113122)
- @medusajs/file-local-next(0.0.3-preview-20240809113122)
Node.js version: v20.11.1
Database: postgresql 16
Operating system: macos 14.4.1
Browser (if relevant): Chrome

### Screenshots
<img width="885" alt="image" src="https://github.com/user-attachments/assets/4730a018-5669-49ad-9cd0-0ebbfc140810">

<img width="864" alt="image" src="https://github.com/user-attachments/assets/eaae940b-1824-4a8f-9700-e604c3e98b7b">

### Code snippets
```javascript
// [PATH] @medusajs/utils/dist/file/abstract-file-provider.d.ts
....
// Executed successfully. 
upload(file: FileTypes.ProviderUploadFileDTO): Promise<FileTypes.ProviderFileResultDTO>;

// Unable to execute
delete(file: FileTypes.ProviderDeleteFileDTO): Promise<void>;
....
```

thetutlage:
Hello @SummerFans 

Can you please share more information on the issue? Like how you are trying to trigger this method and what error do you see? What's the expected outcome.

Bonus: It will be great, if you can put together an example and share it for reproduction with us.

420coupe:
> Hello @SummerFans
> 
> Can you please share more information on the issue? Like how you are trying to trigger this method and what error do you see? What's the expected outcome.
> 
> Bonus: It will be great, if you can put together an example and share it for reproduction with us.

I know what he's talking about. So when you go to product -> media -> edit / gallery. If you delete one of the images, it does a soft delete in db of the image. But doesn't actually call the file service to delete the image. The one experience i am talking about specifically, is with an S3 provider. (supabase s3) in my instance, There was no delete request being sent when deleting image from product media. 

Would also sometimes get a weird error if i tried deleting from the image options vs pressing 'd'  on selected images to delete.

SalahAdDin:
@SummerFans did you check if it was solved in the latest version?

SummerFans:
> @SummerFans did you check if it was solved in the latest version?

[...]

PR #13155, opened 2025-08-06
fix(dashboard): Allow images to be deleted from the dashboard

TheHypnoo:
Currently, if we delete an image from the dashboard, the image will be deleted from the product, but it will not be deleted from the service system being used. For example, AWS.

This solves the problem by first deleting the image from the service and then updating the product.

issue: #8548


olivermrbl:
@TheHypnoo appreciate the contribution. I discussed this with the team, and while your solution solves the problem, we want to move the deletion server-side to ensure we can compensate in case one of the operations fails to complete.

E.g. if the request to update the product fails after deleting the images, your product will end up referencing non-existing links. 

We expect to tackle this in the near future, but feel free to take a stab at this if you want to. Happy to help as well.

SalahAdDin:
> @TheHypnoo appreciate the contribution. I discussed this with the team, and while your solution solves the problem, we want to move the deletion server-side to ensure we can compensate in case one of the operations fails to complete.
> 
> E.g. if the request to update the product fails after deleting the images, your product will end up referencing non-existing links.
> 
> We expect to tackle this in the near future, but feel free to take a stab at this if you want to. Happy to help as well.

Near future mean?

olivermrbl:
> Near future mean?

I suggest following our monthly communicate updates where we will provide information about the tasks/projects we pick up. I cannot give you a timeline right now.

SalahAdDin:
> > Near future mean?
> 
> I suggest following our monthly communicate updates where we will provide information about the tasks/projects we pick up. I cannot give you a timeline right now.

Is it ok to remove this issue when it is not fixed yet?

PR #13156, opened 2025-08-06
feat: custom logger

carlos-r-l-rodrigues:
What:
 - Option to replace the default Logger with a custom implementation.

```ts
// medusa-config.ts
import { MyCustomLogger } from "./custom-logger"

export default defineConfig({
   // ...,
   logger: MyCustomLogger
})

```
The class must respect the `Logger` interface:

https://github.com/medusajs/medusa/blob/a1949a3c4d079e04233d80edd09f56248d711f41/packages/core/types/src/logger/index.ts#L1-L18

CLOSES: CORE-1103


packages/core/framework/src/config/config.ts
```diff
@@ -128,17 +128,19 @@ export class ConfigManager {
    * @protected
    */
   protected normalizeProjectConfig(
-    projectConfig: Partial<ConfigModule["projectConfig"]>
+    projectConfig: Partial<ConfigModule>
```
willbouch: nit: this could probably be called `configs` to avoid having `projectConfig.projectConfig`

packages/core/framework/src/config/loader.ts
```diff
@@ -1,16 +1,16 @@
...
 
 const handleConfigError = (error: Error): void => {
-  logger.error(`Error in loading config: ${error.message}`)
+  defaultLogger.error(`Error in loading config: ${error.message}`)
   if (error.stack) {
-    logger.error(error.stack)
+    defaultLogger.error(error.stack)
   }
-  process.exit(1)
+  // process.exit(1)
```
willbouch: probably forgot to uncomment?

packages/core/framework/src/http/router.ts
```diff
@@ -58,18 +58,23 @@ export class ApiLoader {
    */
   readonly #sourceDirs: string[]
 
+  readonly #logger: Logger
```
willbouch, line 61: why the #? Just curious

packages/core/framework/src/http/router.ts
```diff
@@ -234,6 +242,7 @@ export class ApiLoader {
     authType: AuthType | AuthType[],
     options?: { allowUnauthenticated?: boolean; allowUnregistered?: boolean }
   ) {
+    const logger = this.#logger
```
willbouch, line 245: why not juste replace the logger.debug with `this.logger.debug`

packages/core/types/src/logger/index.ts
```diff
@@ -1,15 +1,18 @@
...
-  activity: (message: string, config?: Record<string, any>) => string
-  progress: (activityId: string, message: string) => void
-  error: (messageOrError: string | Error, error?: Error) => void
-  failure: (activityId: string, message: string) => Record<string, any> | null
-  success: (activityId: string, message: string) => Record<string, any> | null
-  debug: (message: string) => void
-  info: (message: string) => void
-  warn: (message: string) => void
-  log: (...args: any[]) => void
+  panic(data: any): void
```
willbouch: I have noticed you guys use a log of `any` in the code, but I was wondering if `unknown` would be more appropriate in this case? Don't have to change, I am mostly curious as to why `any` seems to be the go to everywhere

packages/core/framework/src/http/router.ts
carlos-r-l-rodrigues: Because here there is another function where the "this" have a diff scope.

packages/core/framework/src/http/router.ts
carlos-r-l-rodrigues: To make the property private in JS

packages/core/framework/src/config/loader.ts
carlos-r-l-rodrigues: Yes, thanks

packages/core/framework/src/config/config.ts
carlos-r-l-rodrigues: Good call.

packages/core/types/src/logger/index.ts
carlos-r-l-rodrigues: Yeah, it can be `unknown` here. No particular reason it is any.

packages/medusa/src/commands/db/create.ts
```diff
@@ -156,13 +161,18 @@ export async function dbCreate({
...
+    const created = await dbCreate({ directory, interactive, db, logger })
     process.exit(created ? 0 : 1)
   } catch (error) {
     if (error.name === "ExitPromptError") {
       process.exit()
     }
-    logger.error(error)
+    defaultLogger.error(error)
     process.exit(1)
   }
```
adrien2p: I would suggest the following here:

```ts
  let logger = defaultLogger as Logger

  try {
    const container = await initializeContainer(directory, {
      skipDbConnection: true,
    })
    logger = container.resolve(ContainerRegistrationKeys.LOGGER)

    const created = await dbCreate({ directory, interactive, db, logger })
    process.exit(created ? 0 : 1)
  } catch (error) {
    if (error.name === "ExitPromptError") {
      process.exit()
    }
    logger.error(error)
    process.exit(1)
  }
```

packages/medusa/src/commands/db/generate.ts
```diff
@@ -1,25 +1,26 @@
...
 } from "@medusajs/framework/utils"
 import { join } from "path"
-
 import { initializeContainer } from "../../loaders"
 import { ensureDbExists } from "../utils"
 
 const TERMINAL_SIZE = process.stdout.columns
 
 const main = async function ({ directory, modules }) {
   try {
```
adrien2p, line 19: Same suggestion here:

```ts
  let logger = defaultLogger as Logger

  try {
    const container = await initializeContainer(directory)
    logger = container.resolve(ContainerRegistrationKeys.LOGGER)
    /**
     * Setup
     */

    await ensureDbExists(container)

    const medusaAppLoader = new MedusaAppLoader()
    const configModule = container.resolve(
      ContainerRegistrationKeys.CONFIG_MODULE
    )

    const plugins = await getResolvedPlugins(directory, configModule, true)
    mergePluginModules(configModule, plugins)

    const linksSourcePaths = plugins.map((plugin) =>
      join(plugin.resolve, "links")
    )
    await new LinkLoader(linksSourcePaths, logger).load()

    /**
     * Generating migrations
     */
    logger.info("Generating migrations...")

    await medusaAppLoader.runModulesMigrations({
      moduleNames: modules,
      action: "generate",
    })

    logger.log(new Array(TERMINAL_SIZE).join("-"))
    logger.info("Migrations generated")

    process.exit()
  } catch (error) {
    logger.log(new Array(TERMINAL_SIZE).join("-"))
    if (error.code && error.code === MedusaError.Codes.UNKNOWN_MODULES) {
      logger.error(error.message)
      const modulesList = error.allModules.map(
        (name: string) => `          - ${name}`
      )
      logger.error(`Available modules:\n${modulesList.join("\n")}`)
    } else {
      logger.error(error.message, error)
    }
    process.exit(1)
  }
  ```

File: packages/medusa/src/commands/db/migrate.ts [hunk 8]
```diff
@@ -38,6 +37,8 @@ export async function migrate({
    * Setup
    */
   const container = await initializeContainer(directory)
+  const logger = container.resolve(ContainerRegistrationKeys.LOGGER)
```
adrien2p: same as the previous one here

packages/medusa/src/commands/db/rollback.ts
```diff
@@ -43,20 +44,20 @@ const main = async function ({ directory, modules }) {
...
       action: "revert",
     })
-    console.log(new Array(TERMINAL_SIZE).join("-"))
+    logger.log(new Array(TERMINAL_SIZE).join("-"))
     logger.info("Migrations reverted")
 
     process.exit()
   } catch (error) {
-    console.log(new Array(TERMINAL_SIZE).join("-"))
+    defaultLogger.log(new Array(TERMINAL_SIZE).join("-"))
```
adrien2p: I believe we can go through all these files and apply the same approach as suggested above, I wont comment that anymore ahah:)

packages/medusa/src/commands/db/run-scripts.ts
```diff
@@ -123,7 +127,7 @@ const main = async function ({
     })
     process.exit(migrated ? 0 : 1)
   } catch (error) {
-    logger.error(error)
+    defaultLogger.error(error)
```
adrien2p: suggestion: perhaps we can move the container init in this main function and benefit from the logger all the way down?

adrien2p:
@carlos-r-l-rodrigues , if you want we can discuss my feedback and get this one ready to merge?

packages/core/framework/src/http/__fixtures__/server/index.ts
```diff
@@ -66,11 +65,8 @@ export const createServer = async (rootDir) => {
 
   container.register(ContainerRegistrationKeys.PG_CONNECTION, asValue({}))
   container.register("configModule", asValue(config))
+  container.register(ContainerRegistrationKeys.LOGGER, asValue(defaultLogger))
```
adrien2p, line 68: Q: is there a reason we register the defaultLogger and we dont register config.logger here?

packages/core/framework/src/http/__fixtures__/server/index.ts
```diff
@@ -88,7 +84,7 @@ export const createServer = async (rootDir) => {
   })
 
   await featureFlagsLoader()
-  await moduleLoader({ container, moduleResolutions, logger })
+  await moduleLoader({ container, moduleResolutions, logger: defaultLogger })
```
adrien2p, line 87: Q: isnt it possible to use config.logger here?

packages/core/framework/src/utils/resource-loader.ts
```diff
@@ -22,8 +27,11 @@ export abstract class ResourceLoader {
    */
   #excludes: RegExp[] = [/^_[^/\\]*(\.[^/\\]+)?$/]
 
-  constructor(sourceDir: string | string[]) {
+  protected logger: Logger
+
+  constructor(sourceDir: string | string[], container: MedusaContainer) {
     this.#sourceDir = sourceDir
+    this.logger = container.resolve(ContainerRegistrationKeys.LOGGER)
```
adrien2p, line 34: very nit: if the class provides  (all these type of class) a logger getter which returns container.resolve of the getter, we wouldnt have to care when the logger get registered in the container, with this approach you cant instanciate the class without the logger being registered prior the class instanciation. I ll let you choose :D

packages/core/types/src/logger/index.ts
```diff
@@ -1,15 +1,18 @@
...
-  activity: (message: string, config?: Record<string, any>) => string
-  progress: (activityId: string, message: string) => void
-  error: (messageOrError: string | Error, error?: Error) => void
-  failure: (activityId: string, message: string) => Record<string, any> | null
-  success: (activityId: string, message: string) => Record<string, any> | null
-  debug: (message: string) => void
-  info: (message: string) => void
-  warn: (message: string) => void
-  log: (...args: any[]) => void
+  panic(data: unknown): void
```
adrien2p, line 2: nit/suggestion: as we are here, would it make sense for the method that only accept a string message to also accept an object to provide structured logging ability?

adrien2p (approved):
LGTM, just a few comments 🎉

packages/core/framework/src/http/__fixtures__/server/index.ts
carlos-r-l-rodrigues: this is a fixture

packages/core/framework/src/http/__fixtures__/server/index.ts
carlos-r-l-rodrigues: fixture

packages/core/framework/src/utils/resource-loader.ts
carlos-r-l-rodrigues: We discussed and that will be done in another PR

packages/core/types/src/logger/index.ts
carlos-r-l-rodrigues: We discussed and that will be done in another PR

PR #13161, opened 2025-08-07
fix(dashboard,promotion): Fix regression on applying percentage promotions

riqwan:
what:

- Fixes regression on being able to apply percentage promotions
- Fixes bug where currency code operator was not being set in promotions form

### DIFF
--- a/integration-tests/http/__tests__/cart/store/cart.spec.ts
+++ b/integration-tests/http/__tests__/cart/store/cart.spec.ts
@@ -2725,1013 +2725,1142 @@ medusaIntegrationTestRunner({
           )
         })
 
-        it("should only apply promotion on discountable items", async () => {
-          const notDiscountableProduct = (
-            await api.post(
-              "/admin/products",
-              {
-                title: "Medusa T-Shirt not discountable",
-                handle: "t-shirt-not-discountable",
-                discountable: false,
-                options: [
-                  {
-                    title: "Size",
-                    values: ["S"],
-                  },
-                ],
-                variants: [
-                  {
-                    title: "S",
-                    sku: "s-shirt",
-                    options: {
-                      Size: "S",
+        describe("With promotions", () => {
+          it("should only apply promotion on discountable items", async () => {
+            const notDiscountableProduct = (
+              await api.post(
+                "/admin/products",
+                {
+                  title: "Medusa T-Shirt not discountable",
+                  handle: "t-shirt-not-discountable",
+                  discountable: false,
+                  options: [
+                    {
+                      title: "Size",
+                      values: ["S"],
                     },
-                    manage_inventory: false,
-                    prices: [
-                      {
-                        amount: 1000,
-                        currency_code: "usd",
-                      },
-                    ],
-                  },
-                ],
-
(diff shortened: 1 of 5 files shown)


integration-tests/http/__tests__/cart/store/cart.spec.ts
```diff
@@ -2725,1013 +2725,1142 @@ medusaIntegrationTestRunner({
...
-          ).data.promotion
 
-          const product = (
-            await api.post(
-              `/admin/products`,
-              {
-                title: "Product for free",
-                description: "test",
-                options: [
+            it("should not apply a percentage promotion to a cart if cart currency is not the same as the promotion currency", async () => {
```
riqwan, line 3805: This is a test addition

integration-tests/http/__tests__/cart/store/cart.spec.ts
```diff
@@ -2725,1013 +2725,1142 @@ medusaIntegrationTestRunner({
...
 
-          let updated = await api.post(
-            `/store/carts/${cart.id}`,
-            {
-              promo_codes: [taxInclPromotion.code],
-            },
-            storeHeaders
-          )
+          describe("Percentage promotions", () => {
+            it("should apply a percentage promotion to a cart", async () => {
```
riqwan, line 3742: This is a test addition

integration-tests/http/__tests__/cart/store/cart.spec.ts
```diff
@@ -2725,1013 +2725,1142 @@ medusaIntegrationTestRunner({
...
-                    values: ["S"],
-                  },
-                ],
-                variants: [
-                  {
-                    title: "S",
-                    sku: "s-shirt",
-                    options: {
-                      Size: "S",
+        describe("With promotions", () => {
```
riqwan, line 2728: Reorg'd some tests within this describe, I've added comments on whats actually been added at the bottom of the file for easy review

packages/admin/dashboard/src/routes/promotions/common/edit-rules/components/rules-form-field/rules-form-field.tsx
```diff
@@ -157,11 +157,23 @@ export const RulesFormField = ({
...
+                      if (currentAttributeOption?.operators?.length === 1) {
+                        fieldRuleOverrides.operator =
+                          currentAttributeOption.operators[0].value
+                      }
+
+                      if (fieldRuleOverrides.operator === "eq") {
+                        fieldRuleOverrides.values = ""
+                      } else {
+                        fieldRuleOverrides.values = []
+                      }
```
olivermrbl, line 174: comment: Could I get you to explain this change?

riqwan: If there is only one operator for an attribute, we can just set the chosen value as the one that exists. In this case, `eq` is the only operator for `currency_code`.

The default values for "eq" operator is a string, whereas for in and nin, are array of strings. Because we're manually setting the operator here, we set the values for it as well

olivermrbl (approved):

PR #13175, opened 2025-08-09
feat: add settings module

srindom:
**What**
- Adds a new module to Medusa that can hold various global settings. The initial implementation comes with ViewConfigurations and UserPreferences.
- A ViewConfiguration is used to store a user's configuration of a data table in the admin dashboard. This includes information about columns, column order, filters, and sorting behavior. Views can be saved with a name and can either be user scoped or default scoped (which is for everyone).
- A UserPreference holds a user's configured value for a given key. For example active_order_view could be a key and a value could be a ViewConfiguration id to persist the selected view configuration between sessions.


packages/medusa/package.json
```diff
@@ -108,6 +108,7 @@
     "@medusajs/promotion": "2.8.8",
     "@medusajs/region": "2.8.8",
     "@medusajs/sales-channel": "2.8.8",
+    "@medusajs/settings": "workspace:^",
```
srindom: q: what's the correct way to add this?

packages/modules/settings/src/services/settings-module-service.ts
```diff
@@ -0,0 +1,388 @@
...
+  ) {
+    super(...arguments)
+    this.baseRepository_ = baseRepository
+    this.viewConfigurationService_ = viewConfigurationService
+    this.userPreferenceService_ = userPreferenceService
+  }
+
+  @InjectManager()
+  // @ts-expect-error
+  async retrieveViewConfiguration(
```
carlos-r-l-rodrigues: `retrieveViewConfiguration` will be auto generated

packages/modules/settings/src/services/settings-module-service.ts
```diff
@@ -0,0 +1,388 @@
...
+    const created = await this.viewConfigurationService_.create(
+      input,
+      sharedContext
+    )
+
+    const serialized = await this.baseRepository_.serialize<
+      SettingsTypes.ViewConfigurationDTO[]
+    >(created, { populate: true })
+
+    return Array.isArray(data) ? serialized : serialized[0]
```
carlos-r-l-rodrigues:
```suggestion  
  return await super.createViewConfigurations(input, sharedContext)
```
We can call the auto generated method that will also emit events

packages/modules/settings/src/services/settings-module-service.ts
```diff
@@ -0,0 +1,388 @@
...
+    }
+
+    // If we have an explicit null preference, or no preference, or a deleted view
+    // We should check for defaults in this order:
+
+    // Check if user has any personal views (only if no explicit null preference)
+    if (!activeViewPref || activeViewPref.value?.viewConfigurationId !== null) {
+      const personalViews = await this.listViewConfigurations(
+        { entity, user_id: userId },
+        { order: { created_at: "ASC" } },
```
carlos-r-l-rodrigues: we can limit it to 1

carlos-r-l-rodrigues (approved):
LGTM,
there are just a couple of suggestions.

PR #13177, opened 2025-08-09
feat(api): add view configuration API routes

srindom:
This is a stacked PR for the view configuration feature.

This PR introduces the API for managing view configurations, allowing users to save and manage their personalized views for entities (orders, products, etc.) in the dashboard. The views hold things like columns, filters and sorting. 

  **What**
- Added view configuration endpoints under /admin/views/{entity}/configurations for creating, reading, updating, and deleting saved view configurations
- Added active view management to set and retrieve which view configuration is currently active for a user
- Added system default views that admins can create to provide default configurations for all users
- Added support for saving filters, sorting, and search state within view configurations
- Added comprehensive integration tests covering all functionality including personal views, system defaults, and state persistence
- Updated validators and route handlers to support the new RESTful structure with entity in the URL path
- Added feature flag (MEDUSA_FF_VIEW_CONFIGURATIONS) to control access to the new endpoints

API Endpoints
```
  // List view configurations for an entity
  GET    /admin/views/{entity}/configurations

  // Create a new view configuration
  POST   /admin/views/{entity}/configurations

  // Get a specific view configuration
  GET    /admin/views/{entity}/configurations/:id

  // Update a view configuration
  POST   /admin/views/{entity}/configurations/:id

  // Delete a view configuration
  DELETE /admin/views/{entity}/configurations/:id

  // Get the active view configuration for an entity
  GET    /admin/views/{entity}/configurations/active

  // Set the active view configuration for an entity
  POST   /admin/views/{entity}/configurations/active
```


packages/medusa/src/api/admin/views/[entity]/configurations/[id]/route.ts
```diff
@@ -0,0 +1,105 @@
...
+} from "@medusajs/framework/http"
+import { AdminUpdateViewConfigurationType } from "../validators"
+import { HttpTypes, ISettingsModuleService } from "@medusajs/framework/types"
+import { MedusaError, Modules } from "@medusajs/framework/utils"
+
+export const GET = async (
+  req: AuthenticatedMedusaRequest<HttpTypes.AdminGetViewConfigurationParams>,
+  res: MedusaResponse<HttpTypes.AdminViewConfigurationResponse>
+) => {
+  const settingsService: ISettingsModuleService = req.scope.resolve(
```
Comment (carlos-r-l-rodrigues) [inline packages/medusa/src/api/admin/views/[entity]/configurations/[id]/route.ts]: Could you add `ISettingsModuleService` type here:
https://github.com/medusajs/medusa/blob/develop/packages/core/framework/src/types/container.ts#L38

Then you can remove `: ISettingsModuleService` everywhere

packages/medusa/src/api/admin/views/[entity]/configurations/[id]/route.ts
```diff
@@ -0,0 +1,105 @@
...
+  )
+
+  // Handle set_active if requested
+  if (set_active) {
+    await settingsService.setActiveViewConfiguration(
+      existing.entity,
+      req.auth_context.actor_id,
+      req.params.id
+    )
+  }
```
Comment (carlos-r-l-rodrigues) [inline packages/medusa/src/api/admin/views/[entity]/configurations/[id]/route.ts]: When there is more than one operation, ideally we use a workflow.

packages/medusa/src/api/admin/views/[entity]/configurations/route.ts
```diff
@@ -0,0 +1,70 @@
...
+    input
+  )
+
+  if (set_active) {
+    await settingsService.setActiveViewConfiguration(
+      viewConfiguration.entity,
+      req.auth_context.actor_id,
+      viewConfiguration.id
+    )
+  }
```
Comment (carlos-r-l-rodrigues) [inline packages/medusa/src/api/admin/views/[entity]/configurations/route.ts]: same here. two operations.

carlos-r-l-rodrigues (approved):

PR #13178, opened 2025-08-10
feat(dashboard,currency): added Tajikistani somoni currency

Amirkhon:
Added currency of the Tajikistan - [Tajikistani Somoni](https://en.wikipedia.org/wiki/Tajikistani_somoni).


tehaulp:
Hi friend, you should also add the currency in the 

**packages/admin/dashboard/src/lib/data/currencies.ts**

 file.
Otherwise it will only be present in database but some errors will appear in admin dashboard

willbouch:
@Amirkhon thanks for the contribution. There is one small comment to address and also could you please add a changeset? Thanks!

willbouch (commented):
Can you also add a changeset? In your case, it would.be a patch. Just run `yarn changeset` in the repo

adrien2p (approved):
thanks a lot, just missing the changeset with patch 💪

Amirkhon:
> Can you also add a changeset? In your case, it would.be a patch. Just run `yarn changeset` in the repo

added

.changeset/witty-spies-move.md
```diff
@@ -0,0 +1,8 @@
+---
+"@medusajs/currency": patch
+"@medusajs/dashboard": patch
+"integration-tests-http": patch
+"@medusajs/utils": patch
+---
+
+Added Tajikistani somoni currency support
```
willbouch: can you replace this with the commit name I added for the PR? Also, you can remove line 4 :) No need to add the integration tests here

willbouch (approved):

willbouch:
I'll get this merged after the small tweak

.changeset/witty-spies-move.md
Amirkhon: done

willbouch (approved):

Amirkhon:
@willbouch  can we now merge this?

willbouch:
> @willbouch can we now merge this?

Will do today

willbouch:
The integration tests keep failing for some reason

Issue #13168, opened 2025-08-08
[Bug]: Draft order creation with no variant_id throws an unexpected error


almousa1990:
### Package.json file

```JSON
{
  "name": "@x/backend",
  "version": "0.0.2",
  "description": "A starter for Medusa projects.",
  "author": "Medusa (https://medusajs.com)",
  "license": "MIT",
  "keywords": [
    "sqlite",
    "postgres",
    "typescript",
    "ecommerce",
    "headless",
    "medusa"
  ],
  "scripts": {
    "build": "medusa build && pnpm resolve:aliases",
    "predeploy": "medusa db:migrate",
    "resolve:aliases": "tsc --showConfig -p tsconfig.json > tsconfig.resolved.json && tsc-alias -p tsconfig.resolved.json && rimraf tsconfig.resolved.json",
    "seed": "medusa exec ./src/scripts/seed.ts",
    "seed:color": "medusa exec ./src/scripts/seed-colors.ts",
    "seed:asset": "medusa exec ./src/scripts/seed-assets.ts",
    "start": "medusa start",
    "dev": "medusa develop",
    "test:integration:http": "TEST_TYPE=integration:http NODE_OPTIONS=--experimental-vm-modules jest --silent=false --runInBand --forceExit",
    "test:integration:modules": "TEST_TYPE=integration:modules NODE_OPTIONS=--experimental-vm-modules jest --silent --runInBand --forceExit",
    "test:unit": "TEST_TYPE=unit NODE_OPTIONS=--experimental-vm-modules jest --silent --runInBand --forceExit"
  },
  "dependencies": {
    "@dnd-kit/accessibility": "^3.1.1",
    "@dnd-kit/core": "^6.3.1",
    "@dnd-kit/modifiers": "^9.0.0",
    "@dnd-kit/sortable": "^10.0.0",
    "@measured/puck": "^0.18.2",
    "@medusajs/admin-sdk": "2.8.7",
    "@medusajs/cli": "2.8.7",
    "@medusajs/framework": "2.8.7",
    "@medusajs/medusa": "2.8.7",
    "@medusajs/ui": "^4.0.7",
    "@x/image-utils": "workspace:*",
    "@x/transactional": "workspace:*",
    "@x/utils": "workspace:*",
    "@mikro-orm/core": "6.4.3",
    "@mikro-orm/knex": "6.4.3",
    "@mikro-orm/migrations": "6.4.3",
    "@mikro-orm/postgresql": "6.4.3",
[...]

PR #13179, opened 2025-08-10
fix: handle missing variants and preserve zero unit_price in prepareLineItems

almousa1990:
Fixed two issues in prepareLineItems:
1. unit_price value of 0 was previously treated as falsy and overwritten with the variant's price. Updated the check to only fallback when unitPrice is null or undefined.
2. When no variants array was provided, the code could throw due to non-null assertions. Now safely handles missing or empty variants.

Additional adjustments:
- Replaced `||` with `??` for array defaults to preserve valid empty arrays.
- Kept changes minimal to avoid altering function signature.

solves: #13168


carlos-r-l-rodrigues (approved):
LGTM

olivermrbl (commented):
Overall LGTM, I believe a [similar function is used when adding to cart](https://github.com/medusajs/medusa/blob/f4ae615a3eba9f1dc058c2c374c95dbb5a0fd423/packages/core/core-flows/src/order/workflows/add-line-items.ts#L32). Could you update that as of this PR too?

almousa1990:
@olivermrbl done.

packages/core/core-flows/src/order/workflows/add-line-items.ts
```diff
@@ -31,7 +31,7 @@ import { productVariantsFields } from "../utils/fields"
 
 function prepareLineItems(data) {
   const items = (data.input.items ?? []).map((item) => {
-    const variant = data.variants?.find((v) => v.id === item.variant_id)!
+    const variant = data.variants?.find((v) => v.id === item.variant_id);
```
olivermrbl: todo: we don't use semi colons
```suggestion
    const variant = data.variants?.find((v) => v.id === item.variant_id)
```

packages/core/core-flows/src/order/workflows/create-order.ts
```diff
@@ -33,21 +33,21 @@ import { updateOrderTaxLinesWorkflow } from "./update-tax-lines"
 
 function prepareLineItems(data) {
   const items = (data.input.items ?? []).map((item) => {
-    const variant = data.variants.find((v) => v.id === item.variant_id)!
+    const variant = data.variants?.find((v) => v.id === item.variant_id);
```
olivermrbl:
```suggestion
    const variant = data.variants?.find((v) => v.id === item.variant_id)
```

adrien2p (approved):
Thank you for your contribution 💪 LGTM

PR #13182, opened 2025-08-10
chore(core-flows): only allow published products in addToCartWorkflow

### LINKED ISSUE #13163: [Bug]: Draft product can be added to cart
Medusa v2.9.0 (preview)

### What happended?

1. Unpublished a product
2. Add a variant of the product to a cart
3. Product is added

### Expected behavior

I expect an error when adding a draft product to the cart.

### Actual behavior

Product is added to the cart.

leobenzol:
Closes #13163 

I have a few questions about expected behaviour, since this currently breaks some tests:

- Many tests use the productModule to create products, with default status == "draft", and use the addToCart workflow which now throws. Should I change all breaking tests to specify status == "published" whne creating the product? The alternative would be to check the status in the store API route before the workflow but 1. it would be an extra query and 2. the addToCart workflow is only used in the store currently, and even if it was to be used admin-side, it still doesn't make sense to add a draft product to cart

- After this PR an unpublished product would give the same error as a variant that doesn't exist. While imho this is correct, the thrown error (for both) is "Items  do not have a price" which doesn't make much sense(i believe the workflows goes through with an empty variants list and then errors at the price check point). Should I throw a different error when a variant doesn't exists/isn't published?

---

> [!NOTE]
> Enforces that only variants from published products can be added to carts, adds status fetching, refines errors, and updates tests to use ProductStatus.PUBLISHED.
> 
> - **Core Flows**:
>   - addToCart: Validate variants exist and belong to `product.status = PUBLISHED`; throw clear `INVALID_DATA` when not found/unpublished.
>   - Data fetching: Include `product.status` in `cart` and `order` variant field selections.
> - **Tests/Fixtures**:
>   - Update integration tests to set `status: ProductStatus.PUBLISHED` when creating products and import `ProductStatus` where needed.
>   - Add cases for unpublished products and non-existent variants producing the new error message.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit ca72532e957964d2d8e6bcecbb0905054c677ded. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


leobenzol:
figured I would push the changes I proposed for sake of expediency when someone gets a chance to review it, they are all separate commits so I can revert if need be

willbouch (commented):
In discussion with the team, will get back to you soon

willbouch:
Sorry, forgot about this one. Pinged the team again @leobenzol

carlos-r-l-rodrigues (approved):

willbouch:
@leobenzol for the tests, I think we should just fix in the tests directly

leobenzol:
@willbouch got it, still working on the merge conflict. I thought it was a trivial reordering at first but there is some change in behavior that I'm not fully understanding and tests are broken(and also some new ones popped up that also need to be fixed)

I'll let you know when I'm done :)

leobenzol:
should be good to go. I ended up moving the logic to the new getVariantsAndItemsWithPricesWorkflow which revealed a lot more broken tests.

maybe @adrien2p should also have a look at this since he added the workflow recently? I'm not sure in which case, in addToCartWorkflow, `lineItems` is empty and `variants/fetchedVariants` is used instead. I wouldn't want to miss an edge case.

carlos-r-l-rodrigues (approved):
Great.
Thanks for the PR @leobenzol.

PR #13183, opened 2025-08-11
feat: add column introspection api

srindom:
This is part of a set of stacked PRs to add support for view configurations, which will allow users to customize the columns shown in admin tables.

The functionality in this PR is behind the view_configuration feature flag.

**What**
- Adds an API to introspect the remote query schema and extract columns that users can add to their table views.

**Notes**
- This is a brute forced approach to get the functionality in place and I expect it to evolve to something more elegant over time. Some ideas for things we might want to consider:
  - Compute the entity columns during build time and store as static data the API can serve quickly.
  - Offer developers more control over how their data can be exposed in this API with additional decorators in the DML.


packages/medusa/src/api/admin/views/[entity]/columns/route.ts
```diff
@@ -0,0 +1,352 @@
...
+
+  const entityMapping = ENTITY_MAPPINGS[entity as keyof typeof ENTITY_MAPPINGS]
+  if (!entityMapping) {
+    return res.status(400).json({
+      message: `Unsupported entity: ${entity}`,
+      type: "invalid_data",
+    } as any)
+  }
+
+  try {
```
Comment (carlos-r-l-rodrigues) [inline packages/medusa/src/api/admin/views/[entity]/columns/route.ts:26]: Can we have this schema introspection extracted to a util?

carlos-r-l-rodrigues (approved):

packages/medusa/src/api/admin/views/[entity]/columns/route.ts
Comment (srindom) [reply packages/medusa/src/api/admin/views/[entity]/columns/route.ts:26]: yeah - I definitely think we should; that said I suspect this will evolve a little more in the near future so might make sense to wait with extracting it right now.

Comment (srindom) [reply packages/medusa/src/api/admin/views/[entity]/columns/route.ts:26]: moved it into the helpers for now!

PR #13184, opened 2025-08-11
fix: update morgan logging skip condition to include SKIP_HTTP_LOGS environment

bgalezowski:
For now we are not able to disable morgan HTTP logs on production, since we have to set `NODE_ENV=production`. This change will allow us to use an additional `SKIP_HTTP_LOGS` environment variable to disable HTTP logs on production, which during high traffic are really problematic because they:

- Make logs difficult to parse: HTTP request logs create noise that makes it hard to find potential errors and bugs in the application logs
- Impact performance: Excessive logging operations consume CPU cycles and I/O resources, especially under heavy load
- Consume storage: Large volumes of HTTP logs quickly fill up disk space and increase log storage costs
- Overwhelm monitoring systems: Log aggregation tools can struggle with the high volume of repetitive HTTP entries

This environment variable provides a clean way to maintain detailed HTTP logging in development while keeping production logs focused on application errors and important business events.

### DIFF
--- a/packages/medusa/src/loaders/express.ts
+++ b/packages/medusa/src/loaders/express.ts
@@ -11,8 +11,11 @@ type Options = {
   configModule: ConfigModule
 }
 
-export default async ({ app, configModule }: Options): Promise<{
-  app: Express,
+export default async ({
+  app,
+  configModule,
+}: Options): Promise<{
+  app: Express
   shutdown: () => Promise<void>
 }> => {
   let sameSite: string | boolean = false
@@ -46,7 +49,7 @@ export default async ({ app, configModule }: Options): Promise<{
   if (configModule?.projectConfig?.redis_url) {
     const RedisStore = createStore(session)
     redisClient = new Redis(
-      configModule.projectConfig.redis_url, 
+      configModule.projectConfig.redis_url,
       configModule.projectConfig.redis_options ?? {}
     )
     sessionOpts.store = new RedisStore({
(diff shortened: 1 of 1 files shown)


carlos-r-l-rodrigues:
Hi @bgalezowski,
thanks for PR, but setting the env var `LOG_LEVEL` to `warn` will remove all the noise from the logs.
In addition to that, soon it will be possible to [replace the default Logger](https://github.com/medusajs/medusa/pull/13156) and have more control on what/when to log.

bgalezowski:
Hi @carlos-r-l-rodrigues, thanks for responding! However, this isn’t actually correct. Maybe it works that way in v2 — I’m not sure — but in v1, LOG_LEVEL=warn doesn’t remove the HTTP logs. LOG_LEVEL isn’t related to morgan at all.

carlos-r-l-rodrigues:
Oh @bgalezowski, I'm sorry, I didn't notice that this was for v1.
I'm reopening the PR, and will share it internally.

bgalezowski:
Actually I believe that in Medusa v2 is the same problem. If you want I can also prepare the feature for v2.

adrien2p:
Hey @bgalezowski, thank you very much for your contribution, unfortunately, we only accept critical security vulnerabilities on v1 now. So for now I ll close this pr

adrien2p:
if you want to look at v2 I would be more than happy, we already support log levels based on NPM levels

PR #13185, opened 2025-08-11
fix(types): remove order_id in apis that do not handle it

willbouch:
CLOSES: CORE-1017

### DIFF
--- a/packages/core/types/src/order/common.ts
+++ b/packages/core/types/src/order/common.ts
@@ -2488,11 +2488,6 @@ export interface FilterableOrderLineItemProps
   id?: string | string[]
 
   /**
-   * Filter line items by their associated order's ID.
-   */
-  order_id?: string | string[]
-
-  /**
    * Filter by line items' title.
    */
   title?: string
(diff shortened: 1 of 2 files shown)


packages/core/types/src/order/common.ts
```diff
@@ -2487,11 +2487,6 @@ export interface FilterableOrderLineItemProps
    */
   id?: string | string[]
 
-  /**
-   * Filter line items by their associated order's ID.
-   */
-  order_id?: string | string[]
```
willbouch, line 2493: This is the one causing issue in the ticket

packages/core/types/src/order/common.ts
```diff
@@ -2549,11 +2544,6 @@ export interface FilterableOrderShippingMethodProps
    */
   id?: string | string[]
 
-  /**
-   * Filter the shipping methods by their associated order's ID.
-   */
-  order_id?: string | string[]
```
willbouch, line 2555: This is another one that is failing, which I figured we would remove for the same reasons

packages/core/types/src/order/common.ts
```diff
@@ -2816,32 +2806,6 @@ export interface FilterableOrderTransactionProps
   created_at?: OperatorMap<string>
 }
 
-/**
- * The filters to apply on the retrieved order items.
- */
-export interface FilterableOrderItemProps
```
willbouch, line 2822: unused

olivermrbl (approved):

PR #13189, opened 2025-08-12
feat(core-flows): Refresh adjustments when editing orders

olivermrbl:
**What**

Refresh adjustments when editing orders.

**Why**

This ensures promotions are reapplied to the "new" order after an edit. For example, if an item has been added and the order had an order-level adjustment, we must redistribute adjustments across the new and original items.


packages/core/core-flows/src/order/workflows/order-edit/refresh-order-edit-adjustments.ts
```diff
@@ -0,0 +1,74 @@
...
+      { input, orderPreview },
+      ({ input, orderPreview }) => {
+        return {
+          ...orderPreview,
+          items: orderPreview.items.map((item) => ({
+            ...item,
+            // Buy-Get promotions rely on the product ID, so we need to manually set it before refreshing adjustments
+            // TODO: Ideally, this is returned by the previewOrderChange step; let's look into this before merging
+            product: { id: item.product_id },
+          })),
```
olivermrbl: self-review: Not proud of this; ideally, the product is populated on the preview, but that's not really possible, since it's a link. I.e. the order module cannot expand that relation

olivermrbl: cc @carlos-r-l-rodrigues

packages/core/framework/package.json
```diff
@@ -85,7 +85,7 @@
     "@opentelemetry/api": "^1.9.0",
     "@types/express": "^4.17.17",
     "chokidar": "^3.4.2",
-    "compression": "1.7.4",
+    "compression": "1.8.0",
```
olivermrbl, line 88: self-review: aligns the version with `@medusajs/medusa`

packages/core/core-flows/src/order/workflows/order-edit/begin-order-edit.ts
```diff
@@ -54,33 +56,33 @@ export const beginOrderEditOrderWorkflowId = "begin-order-edit-order"
...
+ *
  * Create an order edit request.
  */
 export const beginOrderEditOrderWorkflow = createWorkflow(
   beginOrderEditOrderWorkflowId,
   function (
     input: WorkflowData<OrderWorkflow.BeginorderEditWorkflowInput>
   ): WorkflowResponse<OrderChangeDTO> {
-    const order: OrderDTO = useRemoteQueryStep({
+    const order = useRemoteQueryStep({
```
willbouch: Nit: maybe we could refactor to use the new query? Same for others. I'll let you decide if you want to go for the trooper move 🫡 🪖

olivermrbl: Yessir, good push!

packages/core/core-flows/src/order/workflows/order-edit/refresh-order-edit-adjustments.ts
adrien2p: in all workflow functions the data are copied, so there is no real need to re shallow copy object over again, it might just create more createObject calls to the underlying v8 engine. Instead here you can do:

```ts
// Before your workflow you can create a new local type
type OrderPreviewWithPromotions = Omit<OrderPreviewDTO, "items"> & {
  items: (OrderPreviewDTO["items"][number] & {
    product: { id?: string | null }
  })[]
  promotions: PromotionDTO[]
}

// Within your workflow
const orderToRefresh = transform(
      { input, orderPreview },
      ({ input, orderPreview }): OrderPreviewWithPromotions => {
        const orderPreview_ = orderPreview as OrderPreviewWithPromotions
        orderPreview_.items.forEach((item) => {
          item.product = { id: item.product_id }
        })
        orderPreview_.currency_code = input.order.currency_code
        orderPreview_.promotions = input.order.promotions
        return orderPreview_
      }
    )
```

Let me know if that works.

packages/core/core-flows/src/order/workflows/order-edit/begin-order-edit.ts
adrien2p: agree! 👍

adrien2p (approved):
Nice one 🎉 Just a suggestion to test and if you could move the usage from remote query step to graph query step it would be fantastic 💪

packages/core/core-flows/src/order/workflows/order-edit/begin-order-edit.ts
olivermrbl: Done!

packages/core/core-flows/src/order/workflows/order-edit/refresh-order-edit-adjustments.ts
```diff
@@ -0,0 +1,79 @@
...
+import {
+  refreshDraftOrderAdjustmentsWorkflow
+} from "../../../draft-order/workflows/refresh-draft-order-adjustments"
+import { previewOrderChangeStep } from "../../steps"
+
+export const refreshOrderEditAdjustmentsWorkflowId =
+  "refresh-order-edit-adjustments"
+
+/**
+ * The details of the draft order to refresh the adjustments for.
```
fPolic:
```suggestion
 * The details of the order to refresh the adjustments for.
```

fPolic (approved):
LGTM 💪 
Should we add adjustment refresh for exchanges as well?

olivermrbl:
> LGTM 💪 Should we add adjustment refresh for exchanges as well?

Yes, I plan to add that in a follow-up PR 👍

PR #13190, opened 2025-08-12
feat(core-flows): add translations to line item

pepijn-vanvlaanderen:
### What
This PR serves to give localization support on some product / product variant / product collection data. 

We have updated the helper function prepareLineItemData to support reading any translations from metadata fields. If these are absent, the usual core fields are used. It was also needed to add metadata fields to queried fields.

### Why
Many times when working with multi lingual shops, the variants / product titles als needs localization. Though we always use CMS with localization support, cart & order item data is only available in one language in Medusa. With this small PR we can allready make cart & order line items translatable, given any metadata translations object provided.

One can then add UI widgets to manage those translations for a variant / product variant / collection.


willbouch:
Hi @pepijn-vanvlaanderen and thank you for your contribution 🚀 . Although this looks like a working solution, after a discussion with the team, we decided that we would not go forward with translations in the metadata, since it was never meant to store this kind of information

pepijn-vanvlaanderen:
Hi @willbouch, agreed this is a 'patch'. But are there then any plans in the near future to support translations of product (variant) data? Line item data is currently fixed to 1 language. When running a multilingual store, you will run into issues even if you have a multilingual CMS in front, which is also a workaround tbh. Besides translating data for a storefront, you also have transactional emails, server-side tracking, analytics & reporting which depend on localized data.

willbouch:
Let me circle back with the team and I will get back to you

TheGP:
So? Why do u store product information in one language already? I suggest completely removing any language data or adding multi language natively

willbouch:
@pepijn-vanvlaanderen @TheGP We added this ticket in our board and will get to it soon. We will expose a hook rather than using the metadata. That should cover what you need for your use case :)

SalahAdDin:
> @pepijn-vanvlaanderen @TheGP We added this ticket in our board and will get to it soon. We will expose a hook rather than using the metadata. That should cover what you need for your use case :)

Which ticket?

mdiab1996:
Any updates?

SalahAdDin:
> @pepijn-vanvlaanderen @TheGP We added this ticket in our board and will get to it soon. We will expose a hook rather than using the metadata. That should cover what you need for your use case :)

which one?

PR #13193, opened 2025-08-13
Feat/datatable core enhancements

srindom:
**What**

This PR adds core DataTable enhancements to support view configurations in the admin dashboard. This is part of a set of stacked PRs to add the feature to Medusa.

- Puts handlers in place to update the visible columns in a table and the order in which they appear.
- Adds a ViewPills component for displaying and switching between saved view configurations
- Integrated view configuration hooks (useViewConfigurations) with the DataTable

Note: Column drag-and-drop reordering and the column visibility UI controls are not included in this PR as they require additional UI library updates - which will come in the next PR.

Example of what this looks like with the feature flag turned on - note the view pills with "default" in the top. This will expand when the data table behavior adds configuration.
CleanShot 2025-08-13 at 2  31 35@2x


packages/admin/dashboard/src/hooks/table/columns/use-column-state.ts
```diff
@@ -0,0 +1,160 @@
...
+}
+
+/**
+ * Gets the initial column order from API columns
+ */
+function getInitialColumnOrder(
+  apiColumns: HttpTypes.AdminViewColumn[]
+): string[] {
+  const sortedColumns = [...apiColumns].sort((a, b) => {
+    const orderA = a.default_order ?? DEFAULT_COLUMN_ORDER
```
fPolic, line 154: nit: use `apiColumns.length` instead of `DEFAULT_COLUMN_ORDER`

fPolic (approved):

PR #13197, opened 2025-08-13
chore(core-flows): import zod

carlos-r-l-rodrigues:
What:
 * Add missing package on `core-flows`. That can cause issues when using `pnpm`

### DIFF
--- a/packages/core/workflows-sdk/package.json
+++ b/packages/core/workflows-sdk/package.json
@@ -55,8 +55,7 @@
     "@mikro-orm/postgresql": "6.4.3",
     "awilix": "^8.0.1",
     "express": "^4.21.0",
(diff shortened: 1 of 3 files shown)


packages/core/workflows-sdk/package.json
```diff
@@ -55,8 +55,7 @@
     "@mikro-orm/postgresql": "6.4.3",
     "awilix": "^8.0.1",
     "express": "^4.21.0",
-    "pg": "^8.13.0",
-    "zod": "3.25.76"
```
carlos-r-l-rodrigues, line 59: this was repeated on devDependencies and peerDependencies.

riqwan (approved):

PR #13199, opened 2025-08-14
fix(utils): fix promotion case of each allocation not applying its total amount

riqwan:
what:

When promotion allocation is each, the scope of the promotion is each line item. We were not applying the full scope of its applicable amount to the line item


integration-tests/http/__tests__/cart/store/cart.spec.ts
```diff
@@ -3739,6 +3739,107 @@ medusaIntegrationTestRunner({
             )
           })
 
+          it("should apply promotions to multiple quantity of the same product", async () => {
```
olivermrbl, line 3742: suggestion: should we also add a test where `max_quantity` is less than `line_item.quantity` just be sure all cases are covered?

olivermrbl (dismissed):

integration-tests/http/__tests__/cart/store/cart.spec.ts
riqwan: good catch. There are already tests failing for this case. Will fix the issue

riqwan: Fixed now @olivermrbl

carlos-r-l-rodrigues (approved):
LGTM

olivermrbl (approved):

PR #13225, opened 2025-08-15
fix(utils): auto generated update method return

carlos-r-l-rodrigues:


olivermrbl:
This change was introduced in https://github.com/medusajs/medusa/pull/12715. Judging from that PR, it was an intentional change, so there must be an misalignment/inconsistency somewhere that we need to update as well to prevent this from happening again.

Here's what I would expect in terms of behavior of the service methods:
```ts
createBrands({ name: "Hermes" }): Brand
createBrands([{ name: "Hermes" }]): Brand[]
createBrands([{ name: "Hermes" }, { name: "Rimowa" }]): Brand[]

updateBrands({ id: "br_1234", name: "Hermes" }): Brand
updateBrands([{ id: "br_1234", name: "Hermes" }]): Brand[]
updateBrands({ selector: { name: "Hermes" }, { coo: "France" }): Brand[] // Always returns an array, regardless of how many rows were updated
```

Could it be because our core services behave slightly differently in their method overrides? Maybe this has gotten us to mix this up. Wdyt?

cc @adrien2p

olivermrbl:
> Could it be because our core services behave slightly differently in their method overrides?

I don't love that this is actually the case, but we should deal with this separately when time allows for it.

adrien2p:
> This change was introduced in #12715. Judging from that PR, it was an intentional change, so there must be an misalignment/inconsistency somewhere that we need to update as well to prevent this from happening again.
> 
> Here's what I would expect in terms of behavior of the service methods:
> 
> ```ts
> createBrands({ name: "Hermes" }): Brand
> createBrands([{ name: "Hermes" }]): Brand[]
> createBrands([{ name: "Hermes" }, { name: "Rimowa" }]): Brand[]
> 
> updateBrands({ id: "br_1234", name: "Hermes" }): Brand
> updateBrands([{ id: "br_1234", name: "Hermes" }]): Brand[]
> updateBrands({ selector: { name: "Hermes" }, { coo: "France" }): Brand[] // Always returns an array, regardless of how many rows were updated
> ```
> 
> Could it be because our core services behave slightly differently in their method overrides? Maybe this has gotten us to mix this up. Wdyt?
> 
> cc @adrien2p

If I recall correctly, the internal services was doing the right job. What was wrong and that I ve removed was that the base medusa service was trying to do what we do in our core service ( i.e if data is an array then return an array otherwise return a single item which is not entirely correct). But the base medusa service should not have to do that since the internal service is supposed to already return the right output.

so what Carlos has done to remove this check that mutate the returned result looks correct to me, and correspond to your expectations no?

adrien2p:
What I ve changed in the mentioned pr was mostly linked to the internal service, and some cleanup of the medusa service + fixing some core services but some other medusa service generated methods seemed to also had this issue and I missed to removed the normalization at the beginning of the generated update, so your expectations seems to be solved by this pr which complements the mentioned pr 👌

let me know if it makes sense @olivermrbl

adrien2p (approved):

olivermrbl:
Yeah, that makes sense, thanks for clarifying. I agree this PR introduces the remaining changes needed to align with our expectations. I didn’t mean to suggest the previous PR was wrong, but I wanted to loop you in to make sure we are "fixing" this once and for all.

On that note, we should probably do a minor bump since it technically changes behavior again, even if it’s a correction.

What do you think @adrien2p?

olivermrbl (approved):

adrien2p:
Yes I agree a minor would be better, perhaps we should also highlight it in the changelog 🙌

.changeset/young-mangos-suffer.md
```diff
@@ -0,0 +1,5 @@
+---
+"@medusajs/utils": patch
```
olivermrbl:
```suggestion
"@medusajs/utils": minor
```

PR #13226, opened 2025-08-15
feat(dashboard, core-flows): associate shipping option to type

willbouch:
**WHAT**: Allowing associating shipping options to shipping option types
**WHY**: Groundwork to enable free shipping promotions based on shipping option types
**HOW**: Accepting a `type_id` at the API level (it lives next to the already existing `type` in the API in order to make sure there is no breaking change)
**TESTING**: Manual on dashboard + integration tests

CLOSES: CORE-1115


packages/core/core-flows/src/fulfillment/steps/validate-shipping-option-prices.ts
```diff
@@ -96,7 +93,7 @@ export const validateShippingOptionPricesStep = createStep(
 
     const validation =
       await fulfillmentModuleService.validateShippingOptionsForPriceCalculation(
-        calculatedOptions as FulfillmentWorkflow.CreateShippingOptionsWorkflowInput[]
+        calculatedOptions as CreateShippingOptionDTO[]
```
willbouch, line 96: It was just the wrong type cast

packages/core/core-flows/src/fulfillment/steps/upsert-shipping-options.ts
```diff
@@ -65,7 +61,7 @@ export const upsertShippingOptionsStep = createStep(
 
     const upsertedShippingOptions: ShippingOptionDTO[] =
       await fulfillmentService.upsertShippingOptions(
-        input as UpsertShippingOptionDTO[]
+        input.map(i => ({...i, type: i.type_id ? i.type_id: i.type})) as UpsertShippingOptionDTO[]
```
willbouch: There is another option, which would be to expose in the API something like `type: ShippingOptionType | string` rather than `type: ShippingOptionType, type_id: string`. I figured it would be a bit confusing at the API level, so I chose to go with `type: ShippingOptionType, type_id: string`, which means I have to do a bit of conversion at the step level so that the module is able to save the data properly

packages/medusa/src/api/admin/shipping-options/validators.ts
```diff
@@ -138,7 +138,8 @@ export const AdminCreateShippingOption = z
     data: z.record(z.unknown()).optional(),
     price_type: z.nativeEnum(ShippingOptionPriceTypeEnum),
     provider_id: z.string(),
-    type: AdminCreateShippingOptionTypeObject,
+    type: AdminCreateShippingOptionTypeObject.optional(),
```
willbouch, line 139: type OR type_id should be defined. They cannot be undefined at the same time. Do you guys usually handle this kind of logic with zod? I believe it is possible

packages/core/core-flows/src/fulfillment/steps/upsert-shipping-options.ts
```diff
@@ -65,7 +61,13 @@ export const upsertShippingOptionsStep = createStep(
 
     const upsertedShippingOptions: ShippingOptionDTO[] =
       await fulfillmentService.upsertShippingOptions(
-        input as UpsertShippingOptionDTO[]
+        input.map(i => {
```
willbouch: There is another option, which would be to expose in the API something like type: ShippingOptionType | string rather than type: ShippingOptionType, type_id: string. I figured it would be a bit confusing at the API level, so I chose to go with type: ShippingOptionType, type_id: string, which means I have to do a bit of conversion at the step level so that the module is able to save the data properly

willbouch:
Should I add a ` - {type}`? Only thing I am afraid is that current users will be the funky default types that were created since the beginning :o 
Screenshot 2025-08-18 at 10 21 44 AM

packages/core/core-flows/src/fulfillment/steps/upsert-shipping-options.ts
adrien2p: suggestion: I am not super fan of one letter naming for multi line function (I do accept it for oneliner thouhg ;) )

packages/core/types/src/fulfillment/mutations/shipping-option.ts
```diff
@@ -87,7 +87,7 @@ export interface UpdateShippingOptionDTO {
    * The shipping option type associated with the shipping option.
    */
   type?:
-    | Omit<CreateShippingOptionTypeDTO, "shipping_option_id">
+    | CreateShippingOptionTypeDTO | string
```
adrien2p, line 90: suggestion: I think there is a discrepency between the two interface, let me know wyt:

in the create interface we accept for a type the DTO | string, but for the update we accept DTO | string | { id }. wdyt of aligning both by either changing the craete to DTO | { id } or changing the update by removing the { id }. I might be missing some context but I just wanted to raise the point 🙈

packages/core/types/src/http/shipping-option/admin/payloads.ts
```diff
@@ -156,7 +156,14 @@ export interface AdminCreateShippingOption {
...
    */
-  type: AdminCreateShippingOptionType
+  type?: AdminCreateShippingOptionType
+  /**
+   * The ID of the type of shipping option.
+   *
+   * Learn more in the [Shipping Option](https://docs.medusajs.com/resources/commerce-modules/fulfillment/shipping-option#shipping-profile-and-types)
+   * documentation.
+   */
+  type_id?: string
```
adrien2p, line 166: suggestion: for the API, should we also do something similar to the DTO mutation? meaning type can be AdminCreateShippingOptionType | { id } or depending on previous comment AdminCreateShippingOptionType | string. In the first case we wouldnt even need any conversion i believe and mikro orm would handle it natively. I just find it a bit confusing that the shape and expectation changes all the time, also, what happen when both are provided? I believe only the id will be taken into account right? (but if we cant do that differently for now without creating too much breaking changes then lets go) 🙈

packages/core/types/src/http/shipping-option/admin/payloads.ts
```diff
@@ -254,6 +261,13 @@ export interface AdminUpdateShippingOption {
    * documentation.
    */
   type?: AdminCreateShippingOptionType
+  /**
+   * The ID of the type of shipping option.
+   *
+   * Learn more in the [Shipping Option](https://docs.medusajs.com/resources/commerce-modules/fulfillment/shipping-option#shipping-profile-and-types)
+   * documentation.
+   */
+  type_id?: string
```
adrien2p, line 270: same question/suggestion as previously

packages/modules/fulfillment/src/services/fulfillment-module-service.ts
```diff
@@ -1410,7 +1410,7 @@ export default class FulfillmentModuleService
         shippingOption.id
       )! // Garuantueed to exist since the validation above have been performed
 
-      if (shippingOption.type && !("id" in shippingOption.type)) {
+      if (shippingOption.type && typeof shippingOption.type === 'object' &&  !("id" in shippingOption.type)) {
```
adrien2p: suggestion: you can use `isObject` when you need to do that and then we can simplify the condition in both places to `if (isObject(shippingOption.type) &&  !("id" in shippingOption.type))`

packages/medusa/src/api/admin/shipping-options/validators.ts
adrien2p: yes I believe something like the following should be fine
```ts
.refine((data) => isDefined(data.type) || isDefined(data.type_id), {
    message: "Invalid type or type_id",
    path: ["type", "type_id"],
  })
  ```

but lets validate with @olivermrbl

File: packages/core/types/src/http/shipping-option/admin/payloads.ts (hunk 6 above)
Comment (olivermrbl) [reply packages/core/types/src/http/shipping-option/admin/payloads.ts:166]: I would go with what @willbouch has done here at HTTP level, and apply it all the way down the stack to avoid the breaking change + follow the convention we have elsewhere.

That is:
```
type: DTO
type_id: string
```

If both are provided, I would throw:
```
Only one of `type` and `type_id` can be provided
```

packages/medusa/src/api/admin/shipping-options/validators.ts
olivermrbl: Yup agree, see message regarding the error above.

olivermrbl (approved):
Overall LGTM; will let you address @adrien2p's comments

adrien2p (approved):

PR #13236, opened 2025-08-18
chore: idempotent cart operations

carlos-r-l-rodrigues:
What:
 - Avoid concurrent mutations on cart operations using flag `idempotent` from workflow engine. 

No retention time is set to cleanup the workflow execution when it's done. unless the workflow is a nested workflow execution, in this case, it will get cleaned up once the top level workflow is considered finished

Breaking change, Workflow Engine:
- The `idempotent` flag by itself does not keep executions in the database after completion. To retain them, `retentionTime` must also be set.

**NOTE**
In a separate pr, we will have to introduce `retentionTime` propagation. It means that the top level workflow retention time will be propagated to the nested workflow chain in order to ensure consistency across the execution chain either forward or backward.


packages/modules/workflow-engine-redis/src/utils/workflow-orchestrator-storage.ts
```diff
@@ -416,7 +416,7 @@ export class RedisDistributedTransactionStorage
     })
 
     // Database operations
-    if (hasFinished && !retentionTime && !idempotent) {
+    if (hasFinished && !retentionTime) {
```
adrien2p, line 417: Q: Does that mean that if it was idempotent without any retention time then it was just saved and not removed? Just want to be sure ahah. Should we apply the same to the inmemory storage as well? Also, does that mean that now it would only be idempotent for as long as the retention time allows it to be otherwise, if it is idempotent without retention time then it is idempotent for as long as the workflow is being executed?

adrien2p (approved):
LGTM

packages/modules/workflow-engine-redis/src/utils/workflow-orchestrator-storage.ts
carlos-r-l-rodrigues: Yes, this is it. you answered it already hehe.
it is applied in both, redis and in-memory.

olivermrbl (commented):
thought: I feel like this change to the _idempotency_ changes the purpose of it quite a bit. It goes from being idempotency in the "traditional" sense, i.e. being able to issue multiple requests and receive the same response to now being a concurrency mechanism.

With this in mind, wouldn't it be more appropriate to introduce a new config for this, so we don't blurry the purpose of `idempotent`?

That way we can also avoid the breaking change.

adrien2p:
> thought: I feel like this change to the _idempotency_ changes the purpose of it quite a bit. It goes from being idempotency in the "traditional" sense, i.e. being able to issue multiple requests and receive the same response to now being a concurrency mechanism.
> 
> With this in mind, wouldn't it be more appropriate to introduce a new config for this, so we don't blurry the purpose of `idempotent`?
> 
> That way we can also avoid the breaking change.

I think we have aligned the purpose a bit more here. Previously, it was idempotent for either as long as the retention times allows it or indefinately if there was no retention time. which is probably why we introduced a retention time of 3 days to the complete cart workflow ouselves right?

Now, it remains idempotent for as long as the retention time, but not indefinately unless you specify an infinite retention time. 

So the behaviour is the same except not for an infinite amount of time by default. does that make sense?

In short, we are applying the same approach as we did for the complete cart, we dont want the user to perform the same action twice during the idempotency lifetime of a workflow.

but I understand that there is an ambiguity between idempotency and concurrency. Ideally, what would you see? if you consider something idempotent, it should remain in the database indefinately no matter the specified retention time (which would defeat the purpose) and cannot be run ever again and instead always return the same result right? Here the subtility is that you get this behaviour for a specified amount of time, but you can specify `Infinity` if you want

olivermrbl:
> it should remain in the database indefinately no matter the specified retention time (which would defeat the purpose) and cannot be run ever again and instead always return the same result right? Here the subtility is that you get this behaviour for a specified amount of time, but you can specify Infinity if you want

What if we had a _reasonable_ default retention time when `idempotent` was set without an explicitly configured retention time?

I do agree that storing it indefinitely does not make sense, but not storing it at all also seems off.

adrien2p:
> > it should remain in the database indefinately no matter the specified retention time (which would defeat the purpose) and cannot be run ever again and instead always return the same result right? Here the subtility is that you get this behaviour for a specified amount of time, but you can specify Infinity if you want
> 
> What if we had a _reasonable_ default retention time when `idempotent` was set without an explicitly configured retention time?
> 
> I do agree that storing it indefinitely does not make sense, but not storing it at all also seems off.

Well, it stays stored for as long as the workflow is executing. If you have an async workflow with human in the loop for example or async without human, then it will still be stored and idempotent until completion (kind of default retention time). Now, that being said, I am not against a default retention time unless specified. But with the retention time propagation described for a next pr, then it will get restored to none if the parent doesn't have one but a nested workflow is idempotent (issue we found with nested workflow without retention but idempotent). The tricky thing is that a workflow can be executed independently or as a sub workflow as well and the behaviour might be slightly different. Here my expectation was that we can either make it idempotent for as long as it is executing but not more unless specified (number or infinite), alternatively we can use the locking module to achieve the same. But I circle back to the propagation mentioned above. If I am not describing it properly we can huddle around it if you prefer 😅


Edit: for me, it stays idempotent for a same transaction id. The concurrency would involved concurrent transaction that have different transaction id no?

adrien2p:
@carlos-r-l-rodrigues @olivermrbl shall we get this one ready to be merged?

packages/modules/workflow-engine-redis/src/utils/workflow-orchestrator-storage.ts
```diff
@@ -220,8 +220,8 @@ export class RedisDistributedTransactionStorage
   private async deleteFromDb(data: TransactionCheckpoint) {
     await this.workflowExecutionService_.delete([
       {
-        workflow_id: data.flow.modelId,
-        transaction_id: data.flow.transactionId,
+        // workflow_id: data.flow.modelId,
+        // transaction_id: data.flow.transactionId,
```
olivermrbl: is this intentional?

adrien2p: Yes we want to delete all the workflows for a runId, the runId is shared from the top level workflow down to all nested ones. In that case, the run id being unique will return the entire workflow chain to be deleted. Does that make sense?

olivermrbl: my question was more about why they were commented out? 😄 

should we not remove it if it's no longer used?

adrien2p: Because we only rely on the runId here which is uniq, and shared across all nested workflows of a same execution

PR #13243, opened 2025-08-19
feat: Expose beforePaymentAuthorization and orderCreated hooks in complete-cart workflow

Bhuvi100:
This PR exposes beforePaymentAuthorization and orderCreated hooks in complete-cart workflow.

These hooks are essential for adding any custom functionality before authorizing the payment or after creating the order (apart from the event).

I'm not sure why these two hooks are present but are not exposed.


Bhuvi100:
@olivermrbl This PR was dismissed earlier without any proper reason mentioned. Please ensure that Medusa provides the flexibility by exposing hooks that are defined, not sure why this is ignored. If there's any reason, kindly mention the reason for it.
This way, Medusa can be made flexible as per the claims that it's the `The world's most flexible commerce platform`.

@dwene Tagged as you had mentioned of a same issue on the PR that ignored these hooks

olivermrbl:
@Bhuvi100, these hooks aren’t meant to be exposed at this stage. They serve a specific purpose for our Loyalty plugin, which already comes with a pre-registered hook when the plugin is used. Since we don’t currently support multiple hook registrations, we’ll keep these hooks “private” until that changes.

dwene:
Oooh interesting, that makes sense. If we try to use the loyalty plugin AND try to register that hook Medusa will throw an error. Thanks for the clarification @olivermrbl

olivermrbl:
@dwene, exactly. Sorry for not clarifying earlier; I must have missed this completely

PR #13244, opened 2025-08-19
chore(medusa): fetch shipping related attributes

willbouch:
**WHAT**: This is a groundwork PR that is easier to review on its own without mixing UI stuff in there. The goal here is simply to let the API return shipping related stuff for promotions. Basically, we need the backend to send us the shipping attributes/values that will be used in the dropdowns to set target_rules for the free shipping promotions (like free shipping only for shipping option type X).

**WHY**: Necessary for free shipping promotion

**HOW**: There was already an existing `application_method.target_type` called `shipping_methods`, so I simply reused that. The idea is basically to update a bit the behaviour so that IF the `target_type` is indeed `shipping_methods`, then the API does not return the default product attributes/values, but it rather returns the shipping attributes/values. For the moment, there is only the shipping method type, but it could be expanded in the future.

**TESTING**: integration tests and manual API testing


packages/medusa/src/api/admin/promotions/rule-value-options/[rule_type]/[rule_attribute_id]/route.ts
```diff
@@ -25,12 +25,7 @@ export const GET = async (
   req: AuthenticatedMedusaRequest<AdminGetPromotionRuleParamsType>,
   res: MedusaResponse<HttpTypes.AdminRuleValueOptionsListResponse>
 ) => {
-  const {
-    rule_type: ruleType,
-    rule_attribute_id: ruleAttributeId,
-    promotion_type: promotionType,
-    application_method_type: applicationMethodType,
-  } = req.params
+  const { rule_type: ruleType, rule_attribute_id: ruleAttributeId } = req.params
```
Comment (willbouch) [inline packages/medusa/src/api/admin/promotions/rule-value-options/[rule_type]/[rule_attribute_id]/route.ts:32]: I genuinely do not know how the `promotion_type` and `application_method_type` could be retrieved from the params. Removing them could be a breaking change assuming perhaps some users send GET requests with JSON bodies? I've never seen that but I wanted to flag it and get your opinion on this.

packages/medusa/src/api/admin/promotions/rule-value-options/[rule_type]/[rule_attribute_id]/route.ts
```diff
@@ -43,12 +38,18 @@ export const GET = async (
...
     ruleType,
     ruleAttributeId,
-    applicationMethodType,
+    promotionType: undefined,
+    applicationMethodType: undefined,
+    applicationMethodTargetType:
+      filterableFields.application_method_target_type as string | undefined,
   })
 
+  if (filterableFields.application_method_target_type) {
```
Comment (willbouch) [inline packages/medusa/src/api/admin/promotions/rule-value-options/[rule_type]/[rule_attribute_id]/route.ts:55]: This is not the final implementation. I wanted to make it work to keep moving forward but I wanted your thoughts on this. Basically, what happens here is that I need the `application_method_target_type` and the only way without changing the endpoint is to pass it in the query params. We already passed things like `q` or `offset` as query params, which meant that passing `filterableFields` in the filters at line 57 made sense. But, adding `application_method_target_type` as a query param also means that it somehow ends up in the `filterableFields`, so I need to get rid of it before passing it at line 57. Am I missing something? I wanted your ideas on this topic

@adrien2p @fPolic @olivermrbl

packages/medusa/src/api/admin/promotions/validators.ts
```diff
@@ -76,6 +78,7 @@ export const AdminGetPromotionsRuleValueParams = createFindParams({
   z.object({
     q: z.string().optional(),
     value: z.union([z.string(), z.array(z.string())]).optional(),
+    application_method_target_type: z.string().optional(),
```
willbouch, line 81: This looks quirky. It is related to a comment I left above

packages/medusa/src/api/admin/promotions/rule-value-options/[rule_type]/[rule_attribute_id]/route.ts
Comment (riqwan) [reply packages/medusa/src/api/admin/promotions/rule-value-options/[rule_type]/[rule_attribute_id]/route.ts:32]: These endpoints were specifically added for the promotions UI, my recommendation would be to not worry about breaking changes for users. We just need to handle those breaking changes on the UI that this would affect. Maybe @olivermrbl can give a final say here. 

Question here - why is this being removed? 

This endpoint would be used in the create promotion UI where the promotion type and application method type can switch.

packages/medusa/src/api/admin/promotions/utils/rule-attributes-map.ts
```diff
@@ -127,14 +138,22 @@ const buyGetTargetRules = [
 export const getRuleAttributesMap = ({
   promotionType,
   applicationMethodType,
+  applicationMethodTargetType,
 }: {
   promotionType?: string
   applicationMethodType?: string
+  applicationMethodTargetType?: string
```
riqwan, line 145:
```suggestion
  applicationMethodTargetType?: ApplicationMethodTargetTypeValues
```

riqwan: Just for more specificity:
```suggestion
  promotionType?: PromotionTypeValues
  applicationMethodType?: ApplicationMethodTypeValues
  applicationMethodTargetType?: ApplicationMethodTargetTypeValues
```

packages/medusa/src/api/admin/promotions/utils/rule-attributes-map.ts
```diff
@@ -127,14 +138,22 @@ const buyGetTargetRules = [
...
-    "target-rules": [...commonAttributes],
-    "buy-rules": [...commonAttributes],
+    "target-rules":
+      applicationMethodTargetType === "shipping_methods"
+        ? [...shippingAttributes]
+        : [...productAttributes],
+    "buy-rules":
+      applicationMethodTargetType === "shipping_methods"
+        ? [...shippingAttributes]
+        : [...productAttributes],
```
riqwan:
```suggestion
      applicationMethodTargetType === ApplicationMethodTargetType.SHIPPING_METHODS
        ? [...shippingAttributes]
        : [...productAttributes],
    "buy-rules":
      applicationMethodTargetType === ApplicationMethodTargetType.SHIPPING_METHODS
        ? [...shippingAttributes]
        : [...productAttributes],
```

packages/medusa/src/api/admin/promotions/utils/rule-attributes-map.ts
```diff
@@ -46,7 +46,7 @@ const ruleAttributes = [
   },
 ]
 
-const commonAttributes = [
+const productAttributes = [
```
riqwan: They're not product attributes, but attributes you could use for a `applicationMethodTargetType`, which are items, order, shipping_methods

```suggestion
const itemsAttributes = [
```

And then create a new `ordersAttributes = [...itemsAttributes]` since they'd be the same for now.

packages/medusa/src/api/admin/promotions/rule-value-options/[rule_type]/[rule_attribute_id]/route.ts
Comment (olivermrbl) [reply packages/medusa/src/api/admin/promotions/rule-value-options/[rule_type]/[rule_attribute_id]/route.ts:32]: If I understand the previous comment correctly, it questions how this part of the endpoint worked in the first place. The `promotion_type` and `application_method_type` are picked from `req.params`; however, they are not specified as valid params anywhere.

packages/medusa/src/api/admin/promotions/validators.ts
riqwan: You can remove this from here. This would whitelist `application_method_target_type` to show up in the filterable fields

packages/medusa/src/api/admin/promotions/rule-value-options/[rule_type]/[rule_attribute_id]/route.ts
```diff
@@ -43,12 +38,18 @@ export const GET = async (
 
   validateRuleType(ruleType)
   validateRuleAttribute({
-    promotionType,
     ruleType,
     ruleAttributeId,
-    applicationMethodType,
+    promotionType: undefined,
+    applicationMethodType: undefined,
```
Comment (riqwan) [inline packages/medusa/src/api/admin/promotions/rule-value-options/[rule_type]/[rule_attribute_id]/route.ts:48]: iirc this is required for the other cases?

packages/medusa/src/api/admin/promotions/rule-value-options/[rule_type]/[rule_attribute_id]/route.ts
Comment (riqwan) [reply packages/medusa/src/api/admin/promotions/rule-value-options/[rule_type]/[rule_attribute_id]/route.ts:32]: Hm, just checked the frontend implementation. This might be outdated. `promotionType` and `applicationMethodType` aren't being sent at all and hinges on the `rule_type` (rules, buy-rules, target-rules) instead.

packages/medusa/src/api/admin/promotions/rule-value-options/[rule_type]/[rule_attribute_id]/route.ts
Comment (riqwan) [reply packages/medusa/src/api/admin/promotions/rule-value-options/[rule_type]/[rule_attribute_id]/route.ts:55]: We've done this in other places, its kind of a pseudo convention 😄

packages/medusa/src/api/admin/promotions/rule-value-options/[rule_type]/[rule_attribute_id]/route.ts
Comment (willbouch) [reply packages/medusa/src/api/admin/promotions/rule-value-options/[rule_type]/[rule_attribute_id]/route.ts:32]: Yeah, I also noticed they are never passed. That being said, I don't even understand how they could be received even if they were passed. Anyway, took them out 😆

packages/medusa/src/api/admin/promotions/rule-value-options/[rule_type]/[rule_attribute_id]/route.ts
Comment (willbouch) [reply packages/medusa/src/api/admin/promotions/rule-value-options/[rule_type]/[rule_attribute_id]/route.ts:55]: So it is "okay" @riqwan ?

packages/medusa/src/api/admin/promotions/utils/rule-attributes-map.ts
willbouch: Can't really do that because it is a type. But we do get the IDE autocomplete which is nicer
Screenshot 2025-08-20 at 7 41 08 AM

packages/medusa/src/api/admin/promotions/rule-value-options/[rule_type]/[rule_attribute_id]/route.ts
Comment (willbouch) [reply packages/medusa/src/api/admin/promotions/rule-value-options/[rule_type]/[rule_attribute_id]/route.ts:48]: Correct. We add some rules when the applicationMethodType is "fixed" and something else when promotionType is "buyget". It is just that this route never really passed anything

packages/medusa/src/api/admin/promotions/rule-value-options/[rule_type]/[rule_attribute_id]/route.ts
Comment (riqwan) [reply packages/medusa/src/api/admin/promotions/rule-value-options/[rule_type]/[rule_attribute_id]/route.ts:55]: "okay" indeed

olivermrbl (approved):
LGTM!

PR #13251, opened 2025-08-20
fix(): Cart operation should calculate item prices accounting for quantity

### LINKED ISSUE #13262: [Bug]: Price tiers and refreshCartItemsWorkflow are not compatible
### Package.json file

```JSON
{
  "name": "medusa-backend",
  "version": "0.0.1",
  "license": "MIT",
  "keywords": [
    "sqlite",
    "postgres",
    "typescript",
    "ecommerce",
    "headless",
    "medusa"
  ],
  "scripts": {
    "build": "medusa build",
    "start": "medusa start",
    "dev": "medusa develop",
  },
  "dependencies": {
    "@medusajs/admin-sdk": "2.8.7",
    "@medusajs/cli": "2.8.7",
    "@medusajs/framework": "2.8.7",
    "@medusajs/medusa": "2.8.7",
    "@mikro-orm/core": "6.4.3",
    "@mikro-orm/knex": "6.4.3",
    "@perseidesjs/auth-otp": "^2.1.2",
    "awilix": "^8.0.1",
    "lodash": "^4.17.21",
    "pg": "^8.13.0"
  },
  "devDependencies": {
    "@medusajs/test-utils": "2.8.7",
    "@mikro-orm/cli": "6.4.3",
    "@swc/core": "1.5.7",
    "@swc/jest": "^0.2.36",
    "@types/jest": "^29.5.13",
    "@types/lodash": "^4.17.20",
    "@types/node": "^20.0.0",
    "@types/react": "^18.3.2",
    "jest": "^29.7.0",
    "prop-types": "^15.8.1",
    "react": "^18.2.0",
    "react-dom": "^18.2.0",
    "ts-node": "^10.9.2",
    "typescript": "^5.6.2",
    "vite": "^5.2.11",
    "yalc": "^1.0.0-pre.53"
  },
  "engines": {
    "node": ">=20"
  },
  "packageManager": "yarn@4.6.0",
  "resolutions": {
    "@medusajs/core-flows@npm:2.8.7": "patch:@medusajs/core-flows@npm%3A2.8.7#~/.yarn/patches/@medusajs-core-flows-npm-2.8.7-32af44e703.patch"
  }
}
```

### Node.js version

v22.15.0

### Database and its version

PostgreSQL 16.2

MacOS Sequoia

### Browser name

_No response_

### What happended?

Hey team,
I wanted to follow up on my issue, ["which was recently closed due to inactivity"](https://github.com/medusajs/medusa/issues/12706)

⚡ **TLDR of the previous issue**
**PricingModule doesn’t apply quantity based price tiers because item quantities aren’t properly passed into the pricing logic. 
As a result, discounts for buying in bulk don’t work; only the base price is used, no matter the quantity**

--- 

### How to (almost) solve this

Over time, I discovered that it was possible to inject data into the pricing context during different actions (adding to the cart, updating a line item, etc.), and it seemed to me to be the best solution without modifying the Pricing Module!

The idea is simple and is based on the concept of ["Workflow Hooks"](https://docs.medusajs.com/learn/fundamentals/workflows/workflow-hooks#content)

Let's first tackle the `addToCartWorkflow` :

```ts
// src/workflows/cart-hooks.ts

import { StepResponse } from "@medusajs/framework/workflows-sdk";

 // ℹ️ We've imported the core workflows to attach our own custom logic to the `setPricingContext` hook
import {
  addToCartWorkflow,
  updateLineItemInCartWorkflow
} from "@medusajs/medusa/core-flows";


addToCartWorkflow.hooks.setPricingContext(async ({ items }) => {
  // ℹ️ Here we can return a `StepResponse` to update the pricing context that will be used inside the `addToCartWorkflow`
  
  // ℹ️ We know that when we add to cart on the frontend it's always one item
  // Which means that the `items[0]` has the quantity we need to inject inside our pricing context
  const itemAdded = items[0]

 // ℹ️ I've also checked that our variant has some price tiers before returning the `quantity` inside the `StepResponse`
 const hasPriceTiers = /* ... */
 
 if (hasPriceTiers) {
  return new StepResponse({ quantity: itemAdded.quantity })
 }
})

```

You get the idea, once you'll add an item, the pricing context will contains the `quantity` key, allowing us to trigger that ["specific condition inside the Pricing Module"](https://github.com/medusajs/medusa/blob/v2.9.0/packages/modules/pricing/src/repositories/pricing.ts#L127-L149# 

### Let's apply it to the `updateLineItemInCartWorkflow`
Ok now that each time you add an item to the cart there is a quantity check and a retrieval of the correct price, we should be able to apply the same logic when we update a line item's quantity, right ?

**Unfortunately no**, and this is where we start to have problems...

Natively, when updating the quantity, there is no information inside the workflow that indicates which quantity to use, let's dissect the `updateLineItemInCartWorkflow` and ["especially the default pricing context used"](https://github.com/medusajs/medusa/blob/v2.9.0/packages/core/core-flows/src/cart/workflows/update-line-item-in-cart.ts#L123-L134) :

```ts
// packages/core/core-flows/src/cart/workflows/update-line-item-in-cart.ts

export const updateLineItemInCartWorkflow = createWorkflow(
  updateLineItemInCartWorkflowId,
  (
    input: WorkflowData<UpdateLineItemInCartWorkflowInputDTO & AdditionalData>
  ) => {
    // ...
    const cart = transform({ cartQuery }, ({ cartQuery }) => cartQuery.data[0])
    
    const item = transform({ cart, input }, ({ cart, input }) => {
      return cart.items.find((i) => i.id === input.item_id)
    })

   // ... 
   const setPricingContext = createHook(
      "setPricingContext",
      {
        cart,
        item, // ℹ️  We could use this inside our hook, right ?
        variantIds,
        additional_data: input.additional_data,
      },
      {
        resultValidator: pricingContextResult,
      }
    )
  // ...
 })
```

Seeing this and the default context being used, we might think that we could attach a hook in the same way as the `addToCartWorkflow` and use `item.quantity` to determine the price to use : 

```ts
updateLineItemInCartWorkflow.hooks.setPricingContext(async ({ item }) => {
  // ℹ️ We could also check here if that specific item have pricing tiers before using the quantity
  return new StepResponse({
    quantity: item.quantity,
  });
});
```

But if you try this, you'll get the issue ["I've talked about in my previous issue"](https://github.com/medusajs/medusa/issues/12706#issuecomment-3051881677) where we're getting a stale quantity and not the next/fresh one.

It is because that fresh data (the new quantity that the customer chose) is stored inside the `input.update.quantity` object, __which means that we do not have a native way right now to get a fresh quantity without updating the core flow itself.__

### Patching the `@medusajs/core-flows`

And this is the solution I went on at the moment, I've patched the `@medusajs/core-flows` to make sure we have that fresh quantity data :
```diff
--- a/dist/cart/workflows/update-line-item-in-cart.js
+++ b/dist/cart/workflows/update-line-item-in-cart.js
@@ -98,6 +98,7 @@ exports.updateLineItemInCartWorkflow = (0, workflows_sdk_1.createWorkflow)(expor
         item,
         variantIds,
         additional_data: input.additional_data,
+        nextQuantity: input.update.quantity,
     }, {
         resultValidator: schemas_1.pricingContextResult,
     });
```

Which I can now use in my hook :

```ts
// @ts-expect-error - This is a workaround to allow the quantity to be passed to the pricing context since we've patched the workflow
updateLineItemInCartWorkflow.hooks.setPricingContext(async (props: ExtendedProps) => {
  return new StepResponse({
    quantity: props.nextQuantity,
  });
});
```

And everything is now working fine!... At least that's what I've thought...

### The `refreshCartItemsWorkflow`

Now if you try to update the cart for example (adding metadata or changing something in the cart itself, not a line item), this update will trigger a workflow which will trigger the `refreshCartItemsWorkflow` as a step

And you should see your prices return to a default price, this is because this workflow ["refreshes the line items in your cart and reassigns the prices if they are not custom prices"](https://github.com/medusajs/medusa/blob/v2.9.0/packages/core/core-flows/src/cart/workflows/refresh-cart-items.ts#L203-L207) to the found variant prices found

Unfortunately here, it is not possible to pass a quantity, I mean we could use the hook exposed here but since it will refresh all items in your cart, there is no way to pass a specific quantity or all the items prices will be based on that specific quantity

There is a way here, it is to update our line items making sure the `is_custom_price` is set to true, allowing the workflow to not override the price based on the one found on the variant, but there might be some side effects.

### How to fix that issue ?
I'm a bit lost right now on how I should approach this using price tiers and would love to have some guidance on this and especially how should I handle the `refreshCartItemsWorkflow`

### Actual behavior

-

### Link to reproduction repo

-

adrien2p:
RESOLVES CORE-1114

**What**
None of the cart/order workflow are taking into consideration potential items quantity to build the pricing context in order to calculate the prices accurately.

This pr aim to fix the different workflows in charge or relying on the variant calculated prices to make them aware of potential items quantity as part of the calculation context. The fix try to mitigate potential perfomance degradation by applying a groupping strategy in order to minimize the number of calculation calls which can be expensive. That way, in most cases if context are identical then a single calculation will be performed and remapped, on the other hand, if some context differs, then multiple call can be made but still groupped.


carlos-r-l-rodrigues (approved):
LGTM

adrien2p:
few things left:
- create cart operation - which should also take into consideration the quantity. On it
- create order operation - which should calculate prices only if at least one items does not have a unit price.

packages/core/types/src/modules-sdk/remote-query-object-from-string.ts
```diff
@@ -98,7 +98,7 @@ export type RemoteQueryInput<TEntry extends string> =
           /**
            * The maximum number of items to return.
            */
-          take?: number
+          take?: number | null
```
willbouch: nit: we could specify in the explanation on top what null does here? Probably takes as many as possible?

adrien2p: yes exactly, I have a memory of a default limit being applied by default, unless it has been removed (can't find it anymore). but to be extra safe I still pass null. null is a value that we accept down the layers to identify what strategy to use and if limit must be appleid

adrien2p: I ll remove it for now since it appears that we only apply default limit at the http layer, but can't find it anymore at the module layer nor the query tooling. cc @olivermrbl do you remember that we use to have a default limit before hitting mikro orm? can't find it anymore, would you happen to know?

adrien2p: I believe we removed it, I ll revert this type change 👍

adrien2p:
@olivermrbl I ve pushed the changes we discussed to the create order workflow, let me know what you think. 
cc @carlos-r-l-rodrigues

packages/core/types/src/modules-sdk/remote-query-object-from-string.ts
willbouch: Is it expected that it never got reverted?

packages/core/core-flows/src/cart/workflows/refresh-cart-items.ts
```diff
@@ -142,14 +143,20 @@ export const refreshCartItemsWorkflow = createWorkflow(
     )
     const setPricingContextResult = setPricingContext.getResult()
 
-    when({ input }, ({ input }) => {
+    when("force-refresh-calculate-prices", { input }, ({ input }) => {
       return !!input.force_refresh
     }).then(() => {
-      const cart = useRemoteQueryStep({
-        entry_point: "cart",
+      const { data: cartData } = useQueryGraphStep({
```
willbouch: Just a random thought. I see you do a lot of this `useQueryGraphStep` refactor and therefore have to add a transfoorm to get the cart (`data.cartData[0]`). Do you think it would be possible to perhaps specify something to `useQueryGraphStep` so that the type becomes a single element rather than an array of 1 element? Does that make sense?

adrien2p: does that sounds good to you? https://github.com/medusajs/medusa/pull/13251/commits/ca02531d118975c93b4eb3e898e6aba9f14c7311

adrien2p: cc @olivermrbl if you d like to pitch in just in case

willbouch: Yep that works for me

willbouch (approved):
The PR itself looks fine although I would wait for another approval that knows a bit more about the different flows

carlos-r-l-rodrigues (approved):
LGTM

just missing revert the nullable `take`

packages/core/types/src/modules-sdk/remote-query-object-from-string.ts
willbouch: Regarding Carlos' comment

packages/core/core-flows/src/cart/workflows/refresh-cart-items.ts
olivermrbl: LGTM!

adrien2p:
@olivermrbl shall we merge this one?

olivermrbl:
Let's do it

PR #13258, opened 2025-08-20
chore(pricing): Fix excessive db queries during price sets update

adrien2p:
**What**
The current update prices is generating a huge amount of queries to update few prices. This pr is a rework of that same method but in a more optimized way leading to queries reduction.

after some testing batch updating product with prices went from 150 queries to 7 queries

This translate to bottleneck reduction, faster update


packages/modules/pricing/src/services/pricing-module.ts
```diff
@@ -621,73 +621,102 @@ export default class PricingModuleService
...
   ): Promise<InferEntityType<typeof PriceSet>[]> {
-    // TODO: Since money IDs are rarely passed, this will delete all previous data and insert new entries.
-    // We can make the `insert` inside upsertWithReplace do an `upsert` instead to avoid this
     const normalizedData = await this.normalizeUpdateData(data)
 
-    const priceListPrices = await this.priceService_.list({
+    const existingPrices = await this.priceService_.list({
       price_set_id: normalizedData.map(({ id }) => id),
-      price_list_id: { $ne: null },
+      price_list_id: null,
```
olivermrbl: Maybe trivial, but something important seems to have changed here. We are no longer retrieving price list prices. 

Could I get you to explain why?

adrien2p: it seems that the normalized data consist in a price set collection with id and prices properties. Before, the price list prices was retrieved to be re assigned to the price set before upserting them. Now, we manage prices separately and we dont touch those price list prices, therefore we dont need to retrieve them to re assign them later and perform an update to ensure they are still there since we have not touche them. Does that make sense

adrien2p: when you provide a list of prices, we basically do the following:
- list existing prices that are not part of a price list
- create the new prices
- update the prices if exists
- delete the prices from the existing prices that are not part of a price list and are not provided

so there is no need to fetch the price list prices since we dont touch them at any point

adrien2p:
@olivermrbl have you been able to take it to a tour? 😄

willbouch:
> under development

is it ready 😅 ?

adrien2p:
> > under development
> 
> is it ready 😅 ?

it is ready, there is just a point of discussion I ve opened to @olivermrbl That I d like to discuss before moving forward

olivermrbl:
I took it for a tour; looks like a significant improvement in terms of the amount of queries fired against the DB.

Instead of several queries per price update, it collapses those to a single update statement for all updates.

olivermrbl (approved):
LGTM, really nice work! 

Will let @willbouch have a look too

PR #13263, opened 2025-08-20
feat(dashboard,core,modules): free shipping promotion in dashboard

willbouch:
**WHAT**: Finally, this is it, the free shipping promotion!

**WHY**: See ticket

**HOW**: Added a free shipping template that can be selected on UI + adjustments here in there to make it work

**TESTING**: manual on the UI + local storefront

CLOSES CORE-1109 CORE-1131


packages/admin/dashboard/src/routes/promotions/promotion-create/components/create-promotion-form/create-promotion-form.tsx
```diff
@@ -678,60 +678,62 @@ export const CreatePromotionForm = () => {
...
-                                  label={t(
-                                    "promotions.form.value_type.fixed.title"
-                                  )}
-                                  description={t(
-                                    "promotions.form.value_type.fixed.description"
-                                  )}
-                                  className={clx("basis-1/2")}
-                                />
+                    <>
+                      <Divider />
```
willbouch, line 685: Just a small refactor for the dividers. Without that, I had 2 dividers back to back on the page

packages/admin/dashboard/src/routes/promotions/promotion-edit-details/components/edit-promotion-form/edit-promotion-details-form.tsx
```diff
@@ -223,128 +218,139 @@ export const EditPromotionDetailsForm = ({
...
-              }}
-            />
-
-            <Form.Field
-              control={form.control}
-              name="value"
-              render={({ field: { onChange, ...field } }) => {
-                const currencyCode =
-                  promotion.application_method?.currency_code ?? "USD"
+            {promotion.application_method?.target_type !==
```
willbouch, line 221: I decided to hide things for the "shipping promotions":
- The percentage (so it stays 100%)
- The fixed price vs percentage option
- Allocation

packages/core/js-sdk/src/admin/promotion.ts
```diff
@@ -411,32 +412,36 @@ export class Promotion {
   async listRuleAttributes(
     ruleType: string,
     promotionType?: string,
+    applicationMethodTargetType?: string,
```
willbouch, line 415: I guess that's a breaking change. Do you prefer if I add it after the headers argument?

packages/modules/promotion/src/utils/compute-actions/line-items.ts
```diff
@@ -43,29 +43,6 @@ export function getComputedActionsForItems(
   )
 }
 
-export function getComputedActionsForShippingMethods(
```
willbouch, line 46: Idk why these were there. Spoke with Riqwan and it shouldn't have been used. There is another function implementation in the `shipping_methods.ts` file which is more appropriate

packages/modules/promotion/src/utils/compute-actions/shipping-methods.ts
```diff
@@ -123,20 +123,14 @@ export function applyPromotionToShippingMethods(
       }
 
       const promotionValue = applicationMethod?.value ?? 0
-      const applicableTotal = method.subtotal
       const appliedPromoValue = methodIdPromoValueMap.get(method.id) ?? 0
+      const applicableTotal = MathBN.sub(method.subtotal, appliedPromoValue)
```
willbouch, line 127: Without the changes in this file, the tests were simply failing. I believe there was a slight mistake when the allocation is `across`. Basically, I ran a theoretical case (not so theoretical because one of the tests is like this :p) with Claude. The context is that I am applying 2 promotions (200$ each) on 2 shipping methods (500$ and 100$) "across".
Screenshot 2025-08-21 at 10 41 57 AM

Essentially, the code we had did not really handle this well. I believe there was just a small mistake. The subtraction did not happen at the same time, so instead of subtracting the amount and the last applied promotion value (500-166.67), we were subtracting after the division, which does not make a lot of sense. The goal of the division is to have the relative value of the shipping method based on all shipping methods and with the code we had here, we ended up dividing by a smaller number (because the total was updated and shipping method cost was not). It gave >100% relative value, which is weird

packages/admin/dashboard/src/routes/promotions/promotion-create/components/create-promotion-form/templates.ts
```diff
@@ -104,4 +110,21 @@ export const templates = [
...
+    id: "shipping_discount",
+    type: "standard",
+    title: "Free shipping",
+    description: "Applies a 100% discount to shipping fees",
+    hiddenFields: freeShippingHiddenFields,
+    defaults: {
+      is_automatic: "false",
+      type: "standard",
+      application_method: {
+        allocation: "across",
```
willbouch, line 123: This feels like it should be `each`. But I would have to specify a `max_quantity`. I can either set the max_quantity to an arbitrarily high number and just hide it in the dashboard so it can't be edited OR I keep `across` OR. Other ideas or favourite solutions?

.changeset/chilled-tomatoes-vanish.md
```diff
@@ -0,0 +1,10 @@
+---
+"@medusajs/link-modules": patch
+"@medusajs/promotion": patch
+"@medusajs/dashboard": patch
+"@medusajs/core-flows": patch
+"@medusajs/js-sdk": patch
+"@medusajs/types": patch
+---
+
+feat(dashboard,core,modules): free shipping promotion in dashboard
```
fPolic: comment: we usually add names of the packages here so it woudle be `feat(link-modules,promotion, dashboard...)`

packages/modules/promotion/src/utils/compute-actions/shipping-methods.ts
fPolic: think this calculation makes sense to get proportions correctly

packages/modules/promotion/src/utils/compute-actions/shipping-methods.ts
```diff
@@ -123,20 +123,14 @@ export function applyPromotionToShippingMethods(
...
 
       const div = MathBN.eq(totalApplicableValue, 0) ? 1 : totalApplicableValue
-      let applicablePromotionValue = MathBN.sub(
-        MathBN.mult(MathBN.div(applicableTotal, div), promotionValue),
-        appliedPromoValue
-      )
+      let applicablePromotionValue = MathBN.mult(MathBN.div(applicableTotal, div), promotionValue)
```
fPolic: suggestion: while we refactoring, we can remove `div` since it will always be equal to `totalApplicableValue` because of early return at L:116

fPolic (approved):
Nice work @willbouch!

There are two additional things related to free shipping discounts that can be addressed in a follow up:
- would be nice to display shipping promotion in the order summary
- free shipping promotions seems to not work with draft orders

packages/core/js-sdk/src/admin/promotion.ts
willbouch: Tagging you here @olivermrbl

packages/modules/promotion/src/utils/compute-actions/shipping-methods.ts
willbouch: You are right! + tests are passing

packages/core/js-sdk/src/admin/promotion.ts
olivermrbl: Is this breaking? The value is optional no?

willbouch: But it would change the order of arguments I guess

olivermrbl: Ah sorry, I missed the `headers` argument.

I wonder if we should create a query object instead to avoid the breaking change?
```ts
async listRuleAttributes(
    ruleType: string,
    queryOrPromotionType: { promotionType?: string, applicationMethodTargetType?: string} | string,
    headers
)
```

I really would love to get us into a habit of _never_ breaking the public APIs, even though that sounds hard and means we need to mangle things a bit some times.

I don't feel strongly about this specific case though; I doubt many have used this specific endpoint.

willbouch: It will still break it no? Clients using it would need to pass an objet over a string. I believe the only way to not break it is to add my param after the headers, which feels weird. But, if we are to break it no matter what, we could also do it in like you suggested, so at least it scales

olivermrbl: > It will still break it no? Clients using it would need to pass an objet over a string.

No, if you look at the type, I made it an object or string to keep backward compatibility.

olivermrbl: @willbouch, I'm fine to keep this as is here; as I said, probably no one is using this API in a custom implementation :)

willbouch: ~~Nah it's all good. Im doing it. Just having some issues with dependencies between js-sdk and utils package. I asked Carlos~~

It kind of complicates the API layer too in zod. I understand your point, will keep that in mind next times, but I would leave as is on this one

packages/admin/dashboard/src/routes/promotions/promotion-create/components/create-promotion-form/templates.ts
olivermrbl: ~~Did you test having multiple shipping options of the same type and different types in the cart? Just to be sure the `across` behaves as we would expect. If it does, then I would be fine keeping it as is~~

Nvm. read your comment later in the PR

packages/modules/promotion/src/utils/compute-actions/line-items.ts
olivermrbl: It seems `applyPromotionToItems` in this file also accounts for shipping methods; this code seems to be redundant too, as we have `applyPromotionToShippingMethods`. We could clean that up at a later point

willbouch: Yes I noticed that too. We could indeed clean that up. Just didn't want to go on rampage and make the PR bigger than it is already

olivermrbl (approved):
LGTM 💪

PR #13266, opened 2025-08-21
fix(dashboard): fix translations for polish

gladius882:
This PR removes unused translations and adds missing ones. Done based on `yarn i18n:validate pl.json`

### DIFF
--- a/.changeset/fifty-ads-smash.md
+++ b/.changeset/fifty-ads-smash.md
@@ -1,0 +1,5 @@
+---
+"@medusajs/dashboard": patch
+---
+
+add missing polish translations
--- a/packages/admin/dashboard/src/i18n/translations/pl.json
+++ b/packages/admin/dashboard/src/i18n/translations/pl.json
@@ -37,6 +37,8 @@
     "store": "Sklep",
     "details": "Szczegóły",
     "items_one": "{{count}} przedmiot",
+    "items_few": "{{count}} przemioty",
+    "items_many": "{{count}} przemiotów",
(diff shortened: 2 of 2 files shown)


radeknapora (changes requested):
Thanks for the contribution! 🙌 However, removing these keys is not correct for Polish.

Polish is one of the languages with complex pluralization rules. i18next (which Medusa uses) requires the following forms to display correctly:

- `one` - 1 przedmiot
- `few` - 2, 3, 4 przedmioty
- `many` - 5, 6… przedmiotów
- `other` - fallback

Without these keys, the UI will display wrong or incomplete translations (e.g. "2 items" instead of "2 przedmioty"). That’s why I previously added them – they are necessary for proper grammar in Polish.

So I think this PR actually breaks Polish translation instead of improving it. Could you please revert those deletions?

gladius882:
Thank you for explaination, my mistake.
I've just reverted those deletions

radeknapora (approved):
LGTM 👍 Thanks for updating the PR!

I’ll open a [separate PR](https://github.com/medusajs/medusa/pull/13268) to adjust the schema so these keys are recognized and not marked as invalid. That way we can avoid similar issues in the future.

willbouch:
Thanks for the contribution! Can you just quickly add a changeset in the PR?

gladius882:
First time using changeset so if there is something worng let me know

willbouch (approved):
LGTM. Thank you and good job on the changeset ;)

PR #13280, opened 2025-08-22
fix(fulfillment): don't cascade shipping option delete to shipping option type

fPolic:
**What**
- don't cascade delete shipping option type when shipping option is deleted since types can be shared between options
- prevent shipping option type deletion if there are options associated with it


willbouch (dismissed):

packages/core/core-flows/src/shipping-options/workflows/delete-shipping-option-types.ts
```diff
@@ -48,11 +71,30 @@ export const deleteShippingOptionTypesWorkflowId = "delete-shipping-option-types
...
   deleteShippingOptionTypesWorkflowId,
   (input: WorkflowData<DeleteShippingOptionTypesWorkflowInput>) => {
-    const deletedShippingOptionTypes = deleteShippingOptionTypesStep(input.ids)
-    const shippingOptionTypesDeleted = createHook("shippingOptionTypesDeleted", {
-      ids: input.ids,
+    const shippingOptionsQuery = useQueryGraphStep({
+      entity: "shipping_option",
+      filters: { shipping_option_type_id: input.ids },
+      fields: ["id"],
+    }).config({ name: "get-shipping-options" })
```
olivermrbl: thought(non-blocking): This can potentially return a large number of shipping options; I wonder if we should just do `limit: 1` and then tweak the error message?

```
Cannot delete shipping option type because some shipping options are using it
```

I understand the other approach would provide the actual shipping options that are associated with the type, but since this is the IDs, maybe that's not super helpful anyway.

Curious to hear what you and @willbouch think?

willbouch: yeah that is a fair point. Sure, I'm in

fPolic: yeah think it's a good optimisation, will push an update

olivermrbl (approved):

PR #13281, opened 2025-08-22
feat(core, event-bus): Compensate emit event step utility

adrien2p:
RESOLVE CORE-1129

**What**

Currently, when using emitEventStep, the event is added to the corresponding event group. However, if something fails after this step, and no event has been released manually in the meantime, then during workflow compensation the system will repeatedly emit the event that was originally queued during the workflow invocation. This leads to unexpected behavior.

With this PR, once a workflow is compensated, any events emitted during its invocation will be cleared for the given event group ID and event name. This ensures that invocation events are not emitted after the workflow has been compensated.

The only assumption in that pr is that we clear all the events for a particular event name within the event group


olivermrbl:
@adrien2p, is this still WIP?

adrien2p:
> @adrien2p, is this still WIP?

Yes it is, let me finish it and attent to another issue I am investigating right now

adrien2p:
@olivermrbl should be good to review now

integration-tests/modules/package.json
```diff
@@ -5,7 +5,7 @@
   "license": "MIT",
   "private": true,
   "scripts": {
-    "test:integration": "jest --no-cache --maxWorkers=50% --bail --detectOpenHandles --forceExit --logHeapUsage",
+    "test:integration": "jest --no-cache --maxWorkers=50% --bail --detectOpenHandles --forceExit --logHeapUsage -- __tests__/cart/store/cart.completion.ts",
```
olivermrbl: revert

olivermrbl (approved):
LGTM, I believe this was always the intended behavior, but we should probably still highlight it in the release notes right?

adrien2p:
> integration-tests/modules/package.json

Yes I think it is worth higlighting it and make sure pepole sees it and understand what it means

willbouch (approved):
🚀

adrien2p:
@olivermrbl shall we get this bad boy merged as well?

adrien2p:
cc @sradevski mentioning you just in case this could impact you, let me know 💪

sradevski:
@adrien2p I'll upgrade soon and we'll see what happens

PR #13282, opened 2025-08-22
fix(fulfillment): Geozone constraints builder

olivermrbl:
**What**

Update the geo zone constraints (database where-clause) to include the `type` field when filtering shipping options. This ensures that shipping options are correctly limited to the geo zones associated with their respective service zones.

**Why**

Currently, when listing shipping options for a cart based on a customer's address, the geo zone constraints do not account for the type field of the geo zone. This leads to incorrect filtering of shipping options.

Consider the following service zone: 
```
{
  "id": "serzo_1234",
  "name": "Copenhagen",
  "geo_zones": [
    {
      "id": "fgz_1234",
      "type": "city",
      "country_code": "dk",
      "province_code": "SJAELLAND",
      "city": "Copenhagen",
      "postal_expression": null,
    }
  ]
}
```

If a customer provides the following address:

```
{
  "country_code": "dk",
   "province_code": "SJAELLAND",
   "city": "Odense",
   "postal_code": "5230",
}
```

The current geo zone constraint generated for this address is:

```
{
  $or: [
    {
      country_code: "dk",
      province_code: "SJAELLAND",
      city: "Odense",
      postal_expression: "1256",
    },
    {
      country_code: "dk",
      province_code: "SJAELLAND",
      city: "Odense",
    },
    {
      country_code: "dk",
      province_code: "SJAELLAND",
    },
    {
      country_code: "dk",
    },
  ],
}
```

This constraint incorrectly matches the Copenhagen service zone because it includes a broad condition `{ "country_code": "dk" }`. As a result, the customer is shown shipping options from the Copenhagen service zone, which is not intended.

By including the `type` field in the geo zone constraints, we ensure that only the relevant geo zones (e.g., those with type: "city") are matched against the customer’s address.

### DIFF
--- a/packages/modules/fulfillment/src/services/fulfillment-module-service.ts
+++ b/packages/modules/fulfillment/src/services/fulfillment-module-service.ts
@@ -35,6 +35,7 @@ import {
   ModulesSdkUtils,
   promiseAll,
 } from "@medusajs/framework/utils"
+import { isObject } from "@medusajs/utils"
 import {
   Fulfillment,
   FulfillmentProvider,
@@ -59,7 +60,6 @@ import { joinerConfig } from "../joiner-config"
 import { UpdateShippingOptionsInput } from "../types/service"
 import { buildCreatedShippingOptionEvents } from "../utils/events"
 import FulfillmentProviderService from "./fulfillment-provider"
-import { isObject } from "@medusajs/utils"
 
 const generateMethodForModels = {
   FulfillmentSet,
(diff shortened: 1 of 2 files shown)


packages/modules/fulfillment/src/services/fulfillment-module-service.ts
```diff
@@ -2486,15 +2493,22 @@ export default class FulfillmentModuleService
...
-        "province_code",
-        "city",
-        "postal_expression",
-      ],
-      city: ["country_code", "province_code", "city"],
-      province_code: ["country_code", "province_code"],
-      country_code: ["country_code"],
+      postal_expression: {
+        props: ["country_code", "province_code", "city", "postal_expression"],
+        type: "zip",
```
willbouch, line 2498: where are these types coming from?

willbouch (approved):
Missing a changeset :) LGTM

packages/modules/fulfillment/src/services/fulfillment-module-service.ts
olivermrbl: The `type` is part of the geo zone data model and is a pre-defined set of values, so we can safely "hardcode" them here

PR #13283, opened 2025-08-22
feat(utils): define file config

carlos-r-l-rodrigues:
** What
 - Allow auto-loaded Medusa files to export a config object.
 - Currently supports isDisabled to control loading.
 - new instance `FeatureFlag` exported by `@medusajs/framework/utils`
 - `feature-flags` is now a supported folder for medusa projects, modules, providers and plugins. They will be loaded and added to `FeatureFlag`

** Why
 - Enables conditional loading of routes, migrations, jobs, subscribers, workflows, and other files based on feature flags.

```ts
// /src/feature-flags

import { FlagSettings } from "@medusajs/framework/feature-flags"

const CustomFeatureFlag: FlagSettings = {
  key: "custom_feature",
  default_val: false,
  env_key: "FF_MY_CUSTOM_FEATURE",
  description: "Enable xyz",
}

export default CustomFeatureFlag
```

```ts
// /src/modules/my-custom-module/migration/Migration20250822135845.ts

import { FeatureFlag } from "@medusajs/framework/utils"

export class Migration20250822135845 extends Migration {
  override async up(){ }
  override async down(){ }
}

defineFileConfig({
  isDisabled: () => !FeatureFlag.isFeatureEnabled("custom_feature")
})
```


.changeset/dry-geckos-act.md
```diff
@@ -0,0 +1,10 @@
+---
+"@medusajs/modules-sdk": minor
+"@medusajs/framework": minor
+"@medusajs/types": minor
+"@medusajs/utils": minor
+"@medusajs/medusa": minor
+"@medusajs/test-utils": minor
```
olivermrbl: question: what's breaking about this?

olivermrbl (commented):
Overall LGTM, should we add some tests for this?

.changeset/dry-geckos-act.md
carlos-r-l-rodrigues: Nothing really. It was my mistake 😕. I will change them to patch.

carlos-r-l-rodrigues:
> Overall LGTM, should we add some tests for this?

Yes, I will work on them Monday after finishing the manual testing.

integration-tests/http/__tests__/campaigns/admin/campaigns.spec.ts
```diff
@@ -76,7 +76,7 @@ const promotionData = {
   ],
 }
 
-const env = { MEDUSA_FF_MEDUSA_V2: true }
+const env = {}
```
adrien2p, line 79: 🎉 nice cleanup on all those. it was must needed 😂

integration-tests/modules/__tests__/index/sync.spec.ts
```diff
@@ -160,14 +160,14 @@ medusaIntegrationTestRunner({
...
-    id: String
+    id: ID
     title: String
     handle: String
     variants: [ProductVariant]
   }
   
   type ProductVariant @Listeners(values: ["variant.created", "variant.updated", "variant.deleted"]) {
-    id: String
+    id: ID
```
adrien2p: Q: shall we remove that? it is handle in another pr already

packages/core/modules-sdk/src/feature-flags/register-flag.ts
```diff
@@ -0,0 +1,66 @@
...
+}
+
+/**
+ * Registers a feature flag on the provided router.
+ * Resolving precedence:
+ * - env overrides
+ * - project config overrides
+ * - default value
+ */
+export function registerFeatureFlag(options: RegisterFeatureFlagOptions) {
```
adrien2p: Q: Should we move that to the utils package instead?  I found it a bit weird to be in the module sdk when it is something that is being used widely. let me know

packages/core/framework/src/http/middleware-file-loader.ts
```diff
@@ -51,6 +51,10 @@ export class MiddlewareFileLoader {
   async #processMiddlewareFile(absolutePath: string): Promise<void> {
     const middlewareExports = await dynamicImport(absolutePath)
 
+    if (middlewareExports === MEDUSA_SKIP_FILE) {
```
adrien2p: suggestion: we have this check in a lot of places now, I would suggest to maybe create a new util such as `isFileSkipped(exports)`

packages/core/framework/src/jobs/job-loader.ts
```diff
@@ -31,6 +31,10 @@ export class JobLoader extends ResourceLoader {
       config: CronJobConfig
     }
   ) {
+    if (isFileDisabled(path)) {
```
adrien2p: suggestion: I would remove that from here (and other resources loader) and instead only rely on the resource loader, that way the new behaviour is on top of and we can eventually still use the underlying loaders withtout being altered

packages/core/utils/src/common/dynamic-import.ts
```diff
@@ -13,5 +14,12 @@ import { resolveExports } from "./resolve-exports"
  */
 export async function dynamicImport(path: string): Promise<any> {
   const module = require(path)
-  return resolveExports(module)
+
+  const exported = resolveExports(module)
+
+  if (isFileDisabled(path)) {
+    return MEDUSA_SKIP_FILE
+  }
```
adrien2p: suggestion: I am not sure we should entirely replace the return from there, it would mean that programatically you loose the ability to still dynamically load files to do something like logging, force loading partial resources etc. Instead, I would eventually just mutate the `exported` object (which is always an object) to add the flag such as `__MEDUSA_SKIP_FILE__` and then condition based on that. let me know what you think

packages/core/utils/src/dal/mikro-orm/custom-db-migrator.ts
```diff
@@ -0,0 +1,62 @@
+import { Constructor } from "@medusajs/types"
```
adrien2p, line 1: suggestion: should we also update the `MigrationScriptsMigrator`

adrien2p (commented):
few comments to be discussed 💪

packages/core/utils/src/dal/mikro-orm/custom-db-migrator.ts
carlos-r-l-rodrigues: it's there, on the run method.

adrien2p: missed it ahah 👍

adrien2p (approved):
LGTM 🎉 some tests and we should be good

packages/medusa-test-utils/src/module-test-runner.ts
```diff
@@ -135,7 +139,9 @@ export function moduleIntegrationTestRunner<TService = any>({
     databaseConfig: dbConfig,
     joinerConfig,
     preventConnectionDestroyWarning: true,
+    cwd,
   }
+  console.log("INIT MODULES =============================", cwd)
```
olivermrbl:
```suggestion
```

olivermrbl (approved):

packages/medusa-test-utils/src/module-test-runner.ts
carlos-r-l-rodrigues: thanks 👍🏻

PR #13289, opened 2025-08-24
feat(index): $nin and $not operators

carlos-r-l-rodrigues:


martinerko:
Hi @carlos-r-l-rodrigues , thanks for looking into this. Can you please also update the [types package](https://github.com/medusajs/medusa/blob/c7a987daaa0a448d38beb12350e37ad64e95d41a/packages/core/types/src/index-data/index-operator-map.ts) as it's missing quite a few recently added operators (`$and, $or, $not, $nin`).
Please release a snapshot, I will give it a try.

packages/core/types/src/index-data/index-operator-map.ts
```diff
@@ -6,7 +6,11 @@ export type IndexOperatorMap<T> = {
   $gte?: T
   $ne?: T
   $in?: T
+  $nin?: T
```
adrien2p: q: shouldn't `$nin`, `$and` and `$not` be typed as `T[]`?

packages/modules/index/integration-tests/__tests__/query-builder.spec.ts
```diff
@@ -296,7 +297,24 @@ describe("IndexModuleService query", function () {
...
+    const { data: dataNot } = await module.query({
+      fields: ["product.*", "product.variants.*", "product.variants.prices.*"],
+      filters: {
+        product: {
+          variants: {
+            $not: [
+              {
+                sku: { $eq: null },
+              },
+            ],
```
adrien2p: Suggestion: shouldn't `$not` just be an object such as `$not: { $eq: null }` and if you want multiple conditions then you wrap it into a `$and` just like we have with the `OperatorMap` mentionned above? That way we can kind of align the API's as we are in there as it will be a task anyway. wdy?

packages/modules/index/integration-tests/__tests__/query-builder.spec.ts
```diff
@@ -937,4 +963,108 @@ describe("IndexModuleService query", function () {
...
+                },
+              },
+            },
+            {
+              title: {
+                $eq: "Product 2 title",
+              },
+            },
+          ],
+        },
```
adrien2p:
```suggestion
   product: {
    variants: {
      sku: {
        $nin: ["sku 123"],
      },
      title: {
        $eq: "Product 2 title",
      },
    },
  },
```

packages/modules/index/integration-tests/__tests__/query-builder.spec.ts
```diff
@@ -937,4 +963,108 @@ describe("IndexModuleService query", function () {
...
+      },
+    })
+    expect(data).toEqual(expected)
+
+    const { data: data2 } = await module.query({
```
adrien2p: suggestion: this test is supposed to test not in, though we are also testing and, not etc. Should we split these tests a bit differently wdy?

packages/modules/index/integration-tests/__tests__/query-builder.spec.ts
carlos-r-l-rodrigues: I will keep the type as object, but I will let the query builder handle both. I think it makes sense for logical operators of any kind to be grouped.

adrien2p: Yeah def, I was just thinking to make it easy to switch from graph to index in both direction 👌

willbouch (approved):
Can we type `$in` and `$nin` as `T[]` instead of `T`?

willbouch:
Oh and also, don't forget the changeset

olivermrbl (approved):
LGTM

adrien2p (approved):
LGTM

adrien2p:
can we add a little desc for people to be able to understand?

olivermrbl:
Will merge now; we can add the desc spaerately

PR #13296, opened 2025-08-25
chore(): Module Internal Events

adrien2p:
RESOLVES CORE-1140
RESOLVES CORE-1120

**What**
- Introduce **internal event subscribers** to **intercept** all entities **mutation** event and forward them to the module service to aggregate event when a method is wrapped within `@EmitEvents`
  - That means that it is not necessary anymore to aggregate and emit event manually, except for custom event like invite token generated for example
- Fix all module methods not properly handling **transaction flow**
- Fix all module wrongly using serialization
- Remove unused **event builder** that are **not needed anymore**
- **cleanup** module implementation to remove **event related** management to rely on the new interceptor
- **[Product]** Start creating convention to tests events
  - Separating events expectation from business validation in the integration tests
  - To be applied to all modules as we move but potentially separately from that pr
- Add **missing** **context** where needed and/or context decorator
- Fix **event object** shape (like in inventory where `reservation-item` should have been `reservation_item` to follow the convention)
- Fix module service that calls list* method withing a transaction leading to intermediate data serialization where it is not needed
- taxes singe query instead of N


packages/modules/product/integration-tests/__fixtures__/product/data/create-product.ts
```diff
@@ -79,7 +79,10 @@ export const buildProductAndRelationsData = ({
         title: faker.commerce.productName(),
         sku: faker.commerce.productName(),
         options: options
-          ? { [options[0].title]: options[0].values[0] }
+          ? options.reduce((acc, option) => {
```
willbouch, line 82: Just curious, and maybe I am still sleepy haha but is there anything that prevents using the `Object.entries`? Something like:
```
Object.fromEntries(
  options.map(option => [option.title, option.values[0]])
);
```

I tend to think that reduce functions are a biiiit "harder" to understand at first glance

packages/modules/product/integration-tests/__tests__/product-module-service/product-tags.spec.ts
```diff
@@ -409,26 +368,6 @@ moduleIntegrationTestRunner<IProductModuleService>({
           const newTag = productTags.find((t) => t.value === "new")!
           const updatedTag = productTags.find((t) => t.value === "updated")!
```
willbouch, line 369: can probably delete these variables too

packages/modules/product/package.json
```diff
@@ -29,7 +29,7 @@
     "resolve:aliases": "tsc --showConfig -p tsconfig.json > tsconfig.resolved.json && tsc-alias -p tsconfig.resolved.json && rimraf tsconfig.resolved.json",
     "build": "rimraf dist && tsc --build && npm run resolve:aliases",
     "test": "jest --runInBand --bail --forceExit -- src/**/__tests__/**/*.ts",
-    "test:integration": "jest --bail --forceExit -- integration-tests/__tests__/**/*.ts",
+    "test:integration": "jest --bail --forceExit -- integration-tests/__tests__/**/*.spec.ts",
```
willbouch, line 32: I guess that makes sense. Just want to double check that it is not something you forgot to revert 😅

packages/modules/product/src/repositories/product-category.ts
```diff
@@ -452,9 +452,7 @@ export class ProductCategoryRepository extends DALUtils.MikroOrmBaseTreeReposito
         }
         let productCategory = await manager.findOne<
           InferEntityType<typeof ProductCategory>
-        >(ProductCategory.name, {
-          id: categoryData.id,
-        })
+        >(ProductCategory.name, categoryData.id!)
```
willbouch, line 455: it was just the wrong type from the beginning?

adrien2p:
@olivermrbl I feel like this pr might be considered as minor but let me know what you think once you will have done the review 💪

adrien2p:
cc @sradevski, I dont know if that would have any sort of impact in the platform, but mentioning you so that we can check and eventually I can explain things if needed

packages/modules/fulfillment/src/services/fulfillment-module-service.ts
```diff
@@ -1980,9 +1839,9 @@ export default class FulfillmentModuleService
       )
 
     const updatedLabelIds: string[] = []
-    let deletedLabelIds: string[] = []
+    // let deletedLabelIds: string[] = []
```
olivermrbl: nit: Some commented out stuff here, we need to attend to?

packages/modules/user/src/services/user-module.ts
```diff
@@ -390,14 +350,12 @@ export default class UserModuleService
...
-      populate: true,
-    })
+    >(updatedInvites)
 
     moduleEventBuilderFactory({
-      eventName: UserEvents.INVITE_UPDATED,
+      eventName: UserEvents.INVITE_TOKEN_GENERATED,
       source: Modules.USER,
-      action: CommonEvents.UPDATED,
+      action: "token_generated",
```
olivermrbl, line 359: question: this is breaking? if so, should we avoid it

packages/modules/product/src/services/product-category.ts
```diff
@@ -160,14 +176,45 @@ export default class ProductCategoryService {
...
+
+    registerInternalServiceEventSubscriber(sharedContext, subscriber)
+
+    const deletedIds = await this.productCategoryRepository_.delete(
+      ids,
+      sharedContext
+    )
+
+    // Delete are handled a bit differently since we are going to the DB directly, therefore
+    // just like upsert with replace, we need to dispatch the events manually.
```
olivermrbl, line 197: question: could you expand on this? its not immediately obvious what this means

olivermrbl (commented):
What breaking changes are there? If we can avoid the minor with a little bit of code gymnastics, I wouldn't mind it

packages/modules/user/src/services/user-module.ts
adrien2p: thats my bad, I ll fix that

packages/modules/product/src/services/product-category.ts
adrien2p: the categories are going through the tree repository which seems to be handled a bit differenetly, also, since we have a custom service that overrides the internal, we have to do like in the internal service to get the same behaviour

adrien2p:
> What breaking changes are there? If we can avoid the minor with a little bit of code gymnastics, I wouldn't mind it

Some events were malformed so since we are using a convention for naming they are now fixed, also, emitting event manually would duplicate the events if they don't have the same shape (i ve included a deduplication mechanism). And I think that's mainly it 😅

olivermrbl:
> Some events were malformed so since we are using a convention for naming they are now fixed

Is this the inventory events or where there others too?

adrien2p:
> > Some events were malformed so since we are using a convention for naming they are now fixed
> 
> 
> 
> Is this the inventory events or where there others too?

I have in mind the reservation yup

olivermrbl:
How much work would it entail to continue to support the event shapes that were not following the convention? I would really like to be able to merge this as a patch 🙏

adrien2p:
> How much work would it entail to continue to support the event shapes that were not following the convention? I would really like to be able to merge this as a patch 🙏 

I believe I could find a way to keep them non conventional. But since they are not following the convention (like the entire app) probably by mistake, we could argue that it can be considered as a fix that is worth mentioning in the changelog?

olivermrbl:
> > How much work would it entail to continue to support the event shapes that were not following the convention? I would really like to be able to merge this as a patch 🙏
> 
> I believe I could find a way to keep them non conventional. But since they are not following the convention (like the entire app) probably by mistake, we could argue that it can be considered as a fix that is worth mentioning in the changelog?

Yeah, I see. Do you have an overview of those events, so that we can include them if we go down that path?

adrien2p:
> > > How much work would it entail to continue to support the event shapes that were not following the convention? I would really like to be able to merge this as a patch 🙏
> > 
> > 
> > I believe I could find a way to keep them non conventional. But since they are not following the convention (like the entire app) probably by mistake, we could argue that it can be considered as a fix that is worth mentioning in the changelog?
> 
> Yeah, I see. Do you have an overview of those events, so that we can include them if we go down that path?

Yes of course, I’ll do the list 👌

adrien2p:
ok @olivermrbl so just to clarify, it is not the event name that has changed, but the object value (just realised that I ve mention that it was the `object` in the description). Here the list:

```text
inventory-level
inventory-item
reservation-item
```

olivermrbl:
> ok @olivermrbl so just to clarify, it is not the event name that has changed, but the object value (just realised that I ve mention that it was the `object` in the description). Here the list:
> 
> ```
> inventory-level
> inventory-item
> reservation-item
> ```

I am leaning towards making this a "bug" fix instead of a breaking change, which is also practically true. I doubt anyone has relied on this `object` property up until now.

packages/modules/user/src/services/user-module.ts
```diff
@@ -142,22 +140,21 @@ export default class UserModuleService
...
 
+    const serializedInvites = await this.baseRepository_.serialize<
+      UserTypes.InviteDTO[]
+    >(invites)
+
     moduleEventBuilderFactory({
       eventName: UserEvents.INVITE_TOKEN_GENERATED,
       source: Modules.USER,
-      action: "token_generated",
+      action: CommonEvents.UPDATED,
```
olivermrbl: q: think I highlighted this earlier, but shouldn't this be `token_generated`?

carlos-r-l-rodrigues (approved):

packages/modules/user/src/services/user-module.ts
```diff
@@ -303,24 +276,12 @@ export default class UserModuleService
...
-      data: serializedInvites,
-      sharedContext,
-    })
+    >(invites)
 
     moduleEventBuilderFactory({
       eventName: UserEvents.INVITE_TOKEN_GENERATED,
       source: Modules.USER,
-      action: "token_generated",
+      action: CommonEvents.CREATED,
```
olivermrbl: nit: what about this one? shouldn't we revert this too? 🤔

olivermrbl (approved):
One comment, otherwise LGTM!

packages/modules/user/src/services/user-module.ts
adrien2p: I swear, I changed it everywhere and it keeps coming back 🤣 twice now

olivermrbl: haha yeah, my thought too 😄

adrien2p:
I am investigating the weird test behaviour right now, wait my green light before merging

PR #13298, opened 2025-08-25
chore(fulfillment): cleanup old auto-generatred shipping type

willbouch:
**WHAT**: now that we have shipping option types on the dashboard, we need a small cleanup of the old auto-generated ones.

**WHY**: it will look a bit funky for the users on their dashboard otherwise

**HOW**: create a default shipping option type, then soft delete the old auto-generated ones (they have the code test-code) and finally associate the shipping options to that new default type

**TESTING**: run the migration and double-check in the DB


packages/modules/fulfillment/src/migrations/Migration20250825132614.ts
```diff
@@ -0,0 +1,44 @@
...
+      await this.execute(`
+        UPDATE "shipping_option_type" 
+        SET deleted_at = now()
+        WHERE id IN (${typeIdsString});
+      `)
+    }
+  }
+
+  override async down(): Promise<void> {
+    // Not reversible: would require restoring old test-code types + reassignment
```
willbouch: Discussion: what do you guys think here? I guess I can restore the deleted shipping option type, but I will not be able to reassociate with the right shipping options...

olivermrbl: I guess the consequences of this are not catastrophic. You end up with a default type on the shipping options that used to have a dummy type. Probably actually a better state than before 😄

olivermrbl (dismissed):
LGTM!

olivermrbl (approved):
LGTM

adrien2p (approved):

fPolic (approved):

olivermrbl (approved):

Issue #13265, opened 2025-08-21
[Bug]: Buy X Get Y promotion does not apply modulo and ignores max_quantity in application_method


docloulou:
### Package.json file

```JSON
v2.9.0
```

### Node.js version

na

### Database and its version

na

### Operating system name and version

2.9.0

### What happended?

Currently, the Buy X Get Y promotion logic in Medusa does not apply the promotion modulo the quantity in the cart. For example, with a "Buy 2 Get 1" promotion, if a customer has 4 items in the cart, they should get the promotion twice (i.e., "Buy 4 Get 2"), but this is not the case right now.

Instead, the promotion is only applied once, regardless of how many multiples of X the customer buys. This behavior is not expected for most e-commerce platforms, where such promotions are typically applied multiple times for multiples of X.

Additionally, the promotion should respect the `max_quantity` defined in `application_method` to limit the number of times the promotion is applied if required.

Relevant logic can be found in `packages/modules/promotion/src/utils/compute-actions/buy-get.ts`. Improving this would bring Medusa in line with standard e-commerce promotion behaviors and make it more flexible for merchants.

```
{
  "id": "promo_01K35W0Y49791BQF2V3V6H8T3H",
  "code": "qsd",
  "is_automatic": false,
  "is_tax_inclusive": false,
  "type": "buyget",
  "status": "active",
  "created_at": "2025-08-21T08:30:45.389Z",
  "updated_at": "2025-08-21T08:30:45.389Z",
  "deleted_at": null,
  "campaign_id": null,
  "application_method": {
    "id": "proappmet_01K35W0Y4A2FY7GTG1JNDEV0C1",
    "currency_code": null,
    "max_quantity": 100,
    "apply_to_quantity": 1,
    "buy_rules_min_quantity": 2,
    "type": "percentage",
    "target_type": "items",
    "allocation": "each",
    "raw_value": {
      "value": "100",
      "precision": 20
    },
    "created_at": "2025-08-21T08:30:45.388Z",
    "updated_at": "2025-08-21T08:51:28.595Z",
    "deleted_at": null,
    "promotion_id": "promo_01K35W0Y49791BQF2V3V6H8T3H",
    "target_rules": [
      {
        "id": "prorul_01K35W0Y4BX2Y7HE14CC61JDZ5",
        "description": null,
        "attribute": "items.product.id",
        "operator": "in",
        "created_at": "2025-08-21T08:30:45.388Z",
[...]

PR #13305, opened 2025-08-26
feat(promotion): Allow buyget promotion to apply multiple times on cart

riqwan:
what:

Introduces 2 new features to promotion module:

1. Introduce max quantity limit to promotion application - This will limit the application of the promotion based on the quantity of the target products in the cart. 
2. When applying buy get promotions, we will now apply buyget promotion until eligible items are exhausted or max quantity is reached. 

```
- Buy 2 t-shirts, Get 1 sweater
- Max quantity -> 1

This means you can add two t-shirts, and get 1 sweaters for free. However, if you add four t-shirts, you only get one sweater for free.
```

```
- Buy 2 t-shirts, Get 1 sweater
- Max quantity -> 3

This means you can add six t-shirts, and get three sweaters for free. However, if you add eight t-shirts, you only get three sweaters for free
```

```
- Buy 4 t-shirts, Get 2 sweater
- Max quantity -> 1

This should throw on creation, as the max quantity should as a minimum be the same value as the target rule quantity
```

RESOLVES SUP-2357 / https://github.com/medusajs/medusa/issues/13265


packages/modules/promotion/src/utils/validations/application-method.ts
```diff
@@ -68,6 +68,27 @@ export function validateApplicationMethodAttributes(
         `buy_rules_min_quantity is a required field for Promotion type of ${PromotionType.BUYGET}`
       )
     }
+
+    if (!isPresent(maxQuantity)) {
```
olivermrbl, line 72: question: Do these changes make it breaking?

packages/modules/promotion/src/utils/compute-actions/buy-get.ts
```diff
@@ -162,119 +124,352 @@ export function getComputedActionsForBuyGet(
...
-  // Track quantities of items that can't be used as targets because they were used in buy rules
-  const inapplicableQuantityMap = new Map<string, BigNumberInput>()
+  // Calculate remaining quantities for buy items after accounting for what other promotions have consumed
+  const remainingBuyQuantities = new Map<string, BigNumberInput>()
+  for (const buyItem of eligibleBuyItems) {
+    let consumedByOtherPromotions = MathBN.convert(0)
+
+    // Sum up quantities consumed by other promotions (both buy and target consumption)
+    for (const [code, eligibleItems] of eligibleBuyItemMap) {
+      if (code === promotion.code!) continue
```
olivermrbl: nit: use brackets, this doesn't follow our convention

packages/modules/promotion/src/utils/compute-actions/buy-get.ts
```diff
@@ -162,119 +124,352 @@ export function getComputedActionsForBuyGet(
...
-  for (const buyItem of eligibleItemsByPromotion) {
-    const currentValue =
-      inapplicableQuantityMap.get(buyItem.item_id) || MathBN.convert(0)
-    inapplicableQuantityMap.set(
-      buyItem.item_id,
-      MathBN.add(currentValue, buyItem.quantity)
-    )
+    // Also account for target items consumed by other promotions
+    for (const [code, eligibleItems] of eligibleTargetItemMap) {
+      if (code === promotion.code!) continue
```
olivermrbl: nit: same here

packages/modules/promotion/src/utils/compute-actions/buy-get.ts
```diff
@@ -162,119 +124,352 @@ export function getComputedActionsForBuyGet(
...
-      inapplicableQuantity
-    )
+  // Calculate remaining quantities for target items after accounting for what other promotions have consumed
+  const remainingTargetQuantities = new Map<string, BigNumberInput>()
+  for (const targetItem of eligibleTargetItems) {
+    let consumedByOtherPromotions = MathBN.convert(0)
+
+    // Sum up quantities consumed by other promotions (both buy and target consumption)
+    for (const [code, eligibleItems] of eligibleBuyItemMap) {
+      if (code === promotion.code!) continue
```
olivermrbl: same here

packages/modules/promotion/src/utils/compute-actions/buy-get.ts
```diff
@@ -97,57 +97,19 @@ export function getComputedActionsForBuyGet(
...
-  }
-
-  // Get the number of target items that should receive the discount
-  const targetQuantity = MathBN.convert(
+  // Get the number of target items that should receive the discount per application
+  const targetQuantityPerApplication = MathBN.convert(
     promotion.application_method?.apply_to_quantity ?? 0
   )
 
+  // Get the maximum total quantity that can receive the discount across all applications
```
olivermrbl: though: a lot of changes happened in this file, was that necessary? it's not super easy to review ... would maybe be helpful to split this into sub functions, so it is more clear what happens at each stage

packages/modules/promotion/src/utils/compute-actions/buy-get.ts
```diff
@@ -162,119 +124,352 @@ export function getComputedActionsForBuyGet(
...
+          )
+        }
+      }
+    }
 
-    if (MathBN.lte(applicableQuantity, 0)) {
-      continue
+    // Also account for target items consumed by other promotions
+    for (const [code, eligibleItems] of eligibleTargetItemMap) {
+      if (code === promotion.code!) continue
```
olivermrbl: same here

adrien2p (approved):
LGTM, there is just oli's comments to tackle 💪

packages/modules/promotion/src/utils/validations/application-method.ts
riqwan: Not if we run a migration where max_quantity = apply_to_quantity. I think we should provide one.

packages/modules/promotion/src/utils/compute-actions/buy-get.ts
riqwan: It changes the application method to now run in a loop, so we now need to keep track of a few more things on each loop. I'm refactoring it now to split them into functions.

riqwan:
@olivermrbl @adrien2p I've refactored this a bit. Please take a look again

packages/modules/promotion/integration-tests/__tests__/services/promotion-module/compute-actions.spec.ts
```diff
@@ -5336,19 +5341,396 @@ moduleIntegrationTestRunner({
...
+              is_tax_inclusive: false,
+              is_automatic: false,
+              application_method: {
+                allocation: "each",
+                value: 100,
+                max_quantity: 100,
+                type: "percentage",
+                target_type: "items",
+                apply_to_quantity: 1,
+                max_quantity: 100,
```
willbouch: nit: it is defined twice

packages/modules/promotion/src/utils/validations/application-method.ts
willbouch: +1 for the migration script

packages/modules/promotion/src/utils/compute-actions/buy-get.ts
```diff
@@ -260,21 +320,244 @@ export function getComputedActionsForBuyGet(
...
+    - Maximum applicable promotion quantity is reached  
+    - No valid target items can be found for promotion application
+    
+    Each iteration:
+    1. Prepares an application state (selects buy items + eligible target items)
+    2. Applies promotion to the selected target items
+    3. Updates remaining quantities to prevent double-usage in next iteration
+    4. Updates the total eligible items for next iteration
+  */
+  while (true) {
```
willbouch, line 520: nit: I saw other while (true) in the codebase so feel free to disregard this comment, but in my experience, bad things eventually happen in while (true), even when we think that we cover all the use cases for the loop the end. I suggest using a max counter (something ridiculously high for this flow) and if we reach the max counter, we can throw an error

willbouch (approved):

olivermrbl (approved):
LGTM, think @willbouch makes a good point about the while-loop

packages/modules/promotion/src/utils/compute-actions/buy-get.ts
riqwan: good call, added it in there

packages/modules/promotion/src/utils/validations/application-method.ts
riqwan: added script 🤞🏻

riqwan:
Hey @docloulou, could you give this snapshot a quick spin?

docloulou:
@riqwan 

The modulo seems to work on the promotion, but I’m not sure I understand. If I make a 2+1 free rule (/!\ on the same product - look JSON):

* If in the cart: 2 ⇒ no promotion
* If 3 ⇒ 1 on promotion
* If 4 ⇒ 1 on promotion
* If 5 ⇒ 2 on promotion → wrong, since you’d need to reach 6 in the cart to get 2 on promotion
  …
* If 8 ⇒ 3 on promotion
In the loop, it would also be necessary to subtract/add the apply_to_quantity. (if it’s the same product ID or  target_rules/buy_rules are the same products)


The current calculation logic works well if the products that trigger the promotion are not the same as the products impacted by the promotion (hard to generalize, especially if the target_rules/buy_rules are complex).


The max quantity works fine.

```
{
  "id": "promo_01K3QZ3MKJZ0DXJXT9MWC8ZT2E",
  "code": "test2+1",
  "is_automatic": true,
  "is_tax_inclusive": false,
  "type": "buyget",
  "status": "active",
  "created_at": "2025-08-28T09:10:59.444Z",
  "updated_at": "2025-08-28T09:10:59.444Z",
  "deleted_at": null,
  "campaign_id": null,
  "application_method": {
    "id": "proappmet_01K3QZ3MKJAN24T1Z7ZW7380YM",
    "currency_code": null,
    "max_quantity": 500,
    "apply_to_quantity": 1,
    "buy_rules_min_quantity": 2,
    "type": "percentage",
    "target_type": "items",
    "allocation": "each",
    "raw_value": {
      "value": "100",
      "precision": 20
    },
    "created_at": "2025-08-28T09:10:59.443Z",
    "updated_at": "2025-08-28T09:10:59.443Z",
    "deleted_at": null,
    "promotion_id": "promo_01K3QZ3MKJZ0DXJXT9MWC8ZT2E",
    "target_rules": [
      {
        "id": "prorul_01K3QZ3MKK6RTSMCTEDY8N2W7T",
        "description": null,
        "attribute": "items.product.id",
        "operator": "eq",
        "created_at": "2025-08-28T09:10:59.443Z",
        "updated_at": "2025-08-28T09:10:59.443Z",
        "deleted_at": null,
        "values": [
          {
            "value": "prod_01K2YJ78MRHS6AH1NMRB18VRDQ"
          }
        ]
      }
    ],
    "buy_rules": [
      {
        "id": "prorul_01K3QZ3MKKAVKQN66RKQ06A4GY",
        "description": null,
        "attribute": "items.product.id",
        "operator": "eq",
        "created_at": "2025-08-28T09:10:59.443Z",
        "updated_at": "2025-08-28T09:10:59.443Z",
        "deleted_at": null,
        "values": [
          {
            "value": "prod_01K2YJ78MRHS6AH1NMRB18VRDQ"
          }
        ]
      }
    ],
    "value": 100
  },
  "rules": [],
  "campaign": null
}
```

riqwan:
Good catch, it seems the previous target application isn't being done correctly. Let me write a test and fix this case.

docloulou:
Maybe we can use `cart.items.adjustments.id` to count exactly which items are impacted by the promotion. That way, we compute the correct quantity and avoid subtracting Y from X in a **“buy X, get Y”** scenario when **X = Y**.

docloulou:
> Good catch, it seems the previous target application isn't being done correctly. Let me write a test and fix this case.

Maybe we can use `cart.items.adjustments.id` to count exactly which items are impacted by the promotion. That way, we compute the correct quantity and avoid subtracting Y from X in a **“buy X, get Y”** scenario when **X = Y**.

docloulou:
> > Good catch, it seems the previous target application isn't being done correctly. Let me write a test and fix this case.
> 
> Maybe we can use `cart.items.adjustments.id` to count exactly which items are impacted by the promotion. That way, we compute the correct quantity and avoid subtracting Y from X in a **“buy X, get Y”** scenario when **X = Y**.

just thought the adjustement is done AFTER the computation xD

docloulou:
@riqwan is there a snapshot release that include last commit ?

riqwan:
@docloulou present now, give a run and let me know 🥂

docloulou:
> @docloulou present now, give a run and let me know 🥂

Confirm that all my use cases are OK !

PR #13306, opened 2025-08-26
fix: Compute "virtual" adjustments for order previews

olivermrbl:
The approach for refreshing adjustments on order edits introduced in #13189 turned out to be too aggressive. It would modify the original adjustments on the order as well, which could lead to unexpected totals during order edits.

With this PR, we instead generate "virtual" adjustments that only exist as order changes until they have been committed to the original order. Only then, are they applied to the actual order.

---

> [!NOTE]
> Compute promotion adjustments virtually during previews (order edits/exchanges/draft orders) and persist them only on confirmation using versioned adjustments and new workflows.
> 
> - **Core Flows**:
>   - Introduce virtual-adjustment computation for previews via `order-edit/compute-adjustments-for-preview` and `draft-order/compute-draft-order-adjustments` workflows; replace previous refresh logic across order-edit, exchange, and draft-order flows.
>   - Wire workflows to recompute on item/shipping/promotions changes; ensure tax lines updated; `update-tax-lines` now returns computed tax lines; payment collection update condition relaxed to `gte`.
> - **Order Module**:
>   - Preview now applies sorted actions and injects "virtual" item adjustments (not persisted) and tax lines; on confirmation, create actual line-item adjustments for the new version.
>   - Add `includeTaxLinesAndAdjustmentsToPreview` and plumb through `applyChangesToOrder` to collect `lineItemAdjustmentsToCreate`.
>   - Repository: load item adjustments by item/version; support populating `items.item.adjustments`.
> - **Types/Utils**:
>   - Add `ChangeActionType.ITEM_ADJUSTMENTS_REPLACE` and surface adjustments on virtual items; extend adjustment DTOs with `version`.
>   - Promotion compute-actions now include `item_id`/`shipping_method_id` on remove actions; minor doc/option tweaks.
> - **DB**:
>   - Migration adds `version` column to `order_line_item_adjustment` and backfills from latest `order_item` version; model updated.
> - **Tests**:
>   - Add extensive integration tests for order edits, exchanges, and draft orders validating virtual adjustments and totals.
> - **Misc**:
>   - Bump `compression` to `^1.8.1`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit a0335bb96b7979a6e226fcbde64874dfdf92d870. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


packages/core/core-flows/src/order/workflows/order-edit/cancel-begin-order-edit.ts
```diff
@@ -17,7 +17,7 @@ import {
   throwIfIsCancelled,
   throwIfOrderChangeIsNotActive,
 } from "../../utils/order-validation"
-import { refreshOrderEditAdjustmentsWorkflow } from "./refresh-order-edit-adjustments"
+// import { computeAdjustmentsForPreviewWorkflow } from "./compute-adjustments-for-preview"
```
carlos-r-l-rodrigues:
```suggestion
```

carlos-r-l-rodrigues (approved):
LGTM.

Should we also remove `packages/core/core-flows/src/order/workflows/order-edit/refresh-order-edit-adjustments.ts`?

adrien2p (approved):

integration-tests/http/__tests__/order-edits/order-edits.spec.ts
```diff
@@ -1528,9 +1579,23 @@ medusaIntegrationTestRunner({
...
+          await api.get(`/admin/orders/${orderId}`, adminHeaders)
+        ).data.order
+
+        // The promotion is disabled, so what happens is that promotions that were initially applied are removed
+        expect(orderResult2.total).toEqual(23.2)
+        expect(orderResult2.original_total).toEqual(23.2)
       })
 
-      it("should add, remove, and add buy-get adjustment depending on the quantity of the buy rule product", async () => {
+      it.only("should add, remove, and add buy-get adjustment depending on the quantity of the buy rule product", async () => {
```
adrien2p: todo: rm only

packages/core/core-flows/src/order/workflows/order-edit/begin-order-edit.ts
```diff
@@ -101,9 +101,21 @@ export const beginOrderEditOrderWorkflow = createWorkflow(
         created_by: input.created_by,
         description: input.description,
         internal_note: input.internal_note,
+        metadata: {
+          force_all_promotions: false,
```
fPolic: tought: should this be a flag on OrderChange?

packages/core/core-flows/src/order/workflows/order-edit/begin-order-edit.ts
```diff
@@ -101,9 +101,21 @@ export const beginOrderEditOrderWorkflow = createWorkflow(
...
+        },
       }
     })
 
-    return new WorkflowResponse(createOrderChangeStep(orderChangeInput))
+    const orderChange = createOrderChangeStep(orderChangeInput)
+
+    // computeAdjustmentsForPreviewWorkflow.runAsStep({
```
fPolic: should this be computed only when items/shipping change?

packages/core/core-flows/src/order/workflows/order-edit/compute-adjustments-for-preview.ts
```diff
@@ -0,0 +1,139 @@
...
+} from "@medusajs/framework/workflows-sdk"
+import {
+  getActionsToComputeFromPromotionsStep,
+  prepareAdjustmentsFromPromotionActionsStep,
+} from "../../../cart"
+import { previewOrderChangeStep } from "../../steps/preview-order-change"
+import { createOrderChangeActionsWorkflow } from "../create-order-change-actions"
+
+/**
+ * The data to validate that new items can be added to an order edit.
```
fPolic:
```suggestion
 * The data to compute adjustments for an order edit.
```

packages/core/core-flows/src/order/workflows/order-edit/compute-adjustments-for-preview.ts
```diff
@@ -0,0 +1,139 @@
...
+      () => {
+        const actionsToComputeItemsInput = transform(
+          { previewedOrder },
+          ({ previewedOrder }) => {
+            return {
+              currency_code: input.order.currency_code,
+              items: previewedOrder.items.map((item) => ({
+                ...item,
+                // Buy-Get promotions rely on the product ID, so we need to manually set it before refreshing adjustments
+                product: { id: item.product_id },
```
fPolic, line 93: todo: pass shipping_methods as well

packages/core/types/src/promotion/common/compute-actions.ts
```diff
@@ -269,4 +279,8 @@ export interface ComputeActionOptions {
    * automatically. If not provided, the automatic promotions are applied.
    */
   prevent_auto_promotions?: boolean
```
fPolic, line 281: thought: should we prevent this for all order edits by default

packages/core/core-flows/src/order/workflows/order-edit/begin-order-edit.ts
olivermrbl: comment: I guess this is not needed anymore, so we can remove?

packages/core/core-flows/src/order/workflows/order-edit/confirm-order-edit-request.ts
```diff
@@ -272,6 +265,18 @@ export const confirmOrderEditRequestWorkflow = createWorkflow(
...
+    // })
+
+    // refreshDraftOrderAdjustmentsWorkflow.runAsStep({
+    //   input: {
+    //     order: refreshedOrder,
+    //     promo_codes: promoCodes,
+    //     action: PromotionActions.REPLACE,
+    //   },
+    // })
+
```
olivermrbl: comment: I guess this is also not needed anymore?

packages/core/types/src/promotion/common/compute-actions.ts
olivermrbl: would be fine with that assumption

packages/core/types/src/promotion/common/compute-actions.ts
```diff
@@ -269,4 +279,8 @@ export interface ComputeActionOptions {
...
+  /**
+   * Whether to apply inactive and expired promotions.
+   */
+  force_all?: boolean
```
olivermrbl: comment: if we don't use this in this first iteration, let's remove the flag and its usage

packages/modules/order/src/services/order-module-service.ts
```diff
@@ -3536,6 +3554,7 @@ export default class OrderModuleService
       _orderSummaryUpdate,
       orderShippingMethods,
       orderCreditLines,
+      // orderLineItemAdjustments,
```
olivermrbl: rm?

packages/modules/order/src/utils/actions/item-update.ts
```diff
@@ -26,6 +28,29 @@ OrderChangeProcessing.registerActionType(ChangeActionType.ITEM_UPDATE, {
...
+        // @ts-ignore
+        is_tax_inclusive: existing.is_tax_inclusive,
+        // @ts-ignore
+        tax_lines: existing.tax_lines,
+        adjustments: existing.adjustments,
+        // @ts-ignore
+        detail: existing.detail,
+      },
+      {}
+    )
```
olivermrbl: suggestion: would be good to have these ts-ignores removed 😄

packages/modules/order/src/utils/actions/item-update.ts
```diff
@@ -26,6 +28,29 @@ OrderChangeProcessing.registerActionType(ChangeActionType.ITEM_UPDATE, {
...
+
+    const unitTotal = MathBN.div(itemTotals.total, itemTotals.quantity)
```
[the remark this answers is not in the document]
olivermrbl: This seems sensible, wdyt @fPolic?

packages/modules/order/src/utils/actions/item-update.ts
```diff
@@ -42,7 +67,8 @@ OrderChangeProcessing.registerActionType(ChangeActionType.ITEM_UPDATE, {
 
     setActionReference(existing, action, options)
 
-    return MathBN.mult(existing.unit_price, quantityDiff)
+    // @ts-ignore
+    return MathBN.mult(unitTotal, quantityDiff)
```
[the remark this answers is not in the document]
olivermrbl: Think this could be considered too. What I initially did here was a quick test rather than hte long term solution.

Essentially, I found out that we've always blindly been using the unit price of a line item without accounting for adjustments.

packages/modules/order/src/utils/base-repository-find.ts
```diff
@@ -76,13 +77,43 @@ export function setFindMethods<T>(klass: Constructor<T>, entity: any) {
 
     configurePopulateWhere(config, isRelatedEntity, version)
 
+    let loadAdjustments = false
+    if (config.options.populate.includes("items.item.adjustments")) {
```
olivermrbl, line 81: q: I believe we have a shortcut for the relation at the HTTP layer right, `items.adjustments`? Do we need to do something here, or do we transform the fields correctly before reaching this code?

olivermrbl (commented):
Overall good, left a few comments

packages/modules/order/src/utils/base-repository-find.ts
fPolic: it's handled in `mapRepositoryToOrderModel`

packages/modules/order/src/utils/actions/item-update.ts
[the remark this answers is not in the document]
fPolic: It seems that this needs to be fixed in all actions that return `amount`, should we tackle this in a dedicated PR?

packages/core/core-flows/src/order/workflows/order-edit/compute-adjustments-for-preview.ts
```diff
@@ -0,0 +1,139 @@
...
+          }
+        )
+
+        const actions = getActionsToComputeFromPromotionsStep({
+          computeActionContext: actionsToComputeItemsInput,
+          promotionCodesToApply: orderPromotions,
+        })
+
+        const { lineItemAdjustmentsToCreate } =
+          prepareAdjustmentsFromPromotionActionsStep({ actions })
```
olivermrbl, line 114: question: Adjustments are replaced on every edit right? Even when those edits happen in the same order edit. 

Just wanted to confirm, that we don't need to remove anything.

fPolic: There is a test case for that scenario:
`should update adjustments when adding then updating then removing the original item"`.

We rely on action ordering here so the last applied (and newest) action will set adjustments.
Additionally, since previous `ITEM_ADJUSTMENTS_REPLACE` actions of the same edit are useless, we could remove them in `computeAdjustmentsForPreviewWorkflow` when computing a new ones?

```
deleteOrderChangeActions({
  order_change_id: orderChange.id,
  order_id: order.id,
  version: orderChange.version,
  action: ChangeActionType.ITEM_ADJUSTMENTS_REPLACE,
})
```

olivermrbl: > we could remove them in computeAdjustmentsForPreviewWorkflow when computing a new ones?

Yeah, think this could make sense, but no need to include in this PR

packages/core/core-flows/src/cart/steps/get-actions-to-compute-from-promotions.ts
```diff
@@ -1,56 +1,63 @@
...
- *   cart,
+ *   computeActionContext,
  *   promotionCodesToApply: ["10OFF"]
  * })
  */
 export const getActionsToComputeFromPromotionsStep = createStep(
   getActionsToComputeFromPromotionsStepId,
   async (data: GetActionsToComputeFromPromotionsStepInput, { container }) => {
-    const { cart, promotionCodesToApply = [] } = data
+    const { computeActionContext, promotionCodesToApply = [], options } = data
```
olivermrbl, line 50: suggestion: let's maybe make this non-breaking by still allowing both properties but deprecate `cart`? and then do something like:
```
const actionContext = computeActionContext ?? cart
```

packages/modules/order/src/utils/actions/item-update.ts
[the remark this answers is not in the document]
olivermrbl: Yeah, sure let's do that.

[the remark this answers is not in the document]
fPolic: Actually, should we revert this change to be consistent with other actions? Afaik, this is used to calculate `summary.current_order_total` which is based on `unit_price * quantity` since we don't care about the" full" totals at this point in the module.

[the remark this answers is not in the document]
olivermrbl: We can do that, but I think we will have to fix this before releasing, as it fucks up the totals calculations in the UI, if I am not mistaken?

E.g. if you edit an item with a promotion, the new order total will not account for it. At least it was like that when I made the change. But perhaps this has been fixed as part of doing the rest of it?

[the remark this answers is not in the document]
fPolic: We don't use `current_order_total` from `summary` in the UI. If the new total wasn't correct, probably the adjustments weren't properly set. We have a [test](https://github.com/medusajs/medusa/pull/13306/files#diff-3dbfb725ad63f8f53bde956f431e4d0a153d3349760a2561f931bb71e35ffabdR1366) covering that but will test a few more scenarios tomorrow.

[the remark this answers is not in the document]
olivermrbl: Got it!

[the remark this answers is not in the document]
fPolic: You are right, `summary.pending_difference` ends up incorrect on `preview` in both cases, but it's correct after the edit is confirmed. I'm looking into it.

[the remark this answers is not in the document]
olivermrbl: Yeah, I am pretty sure it's because the return value of the actions is used to calculate the order summary totals, and in the actions, we don't consider discounted prices. We just blindly use the unit price and multiply it by the quantity. Therefore, in order to make this right, we would need to revisit all return values of all actions. 

To be clear, this has been an issue for a long time and is really unrelated to the PR. I just noticed it as part of making the initial changes.

We can tackle this in a follow-up PR to not further bloat this one?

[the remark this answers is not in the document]
fPolic: Agree, let's revisit this separately!

packages/modules/order/src/migrations/Migration20251016160403.ts
```diff
@@ -0,0 +1,33 @@
...
+          ON oi.item_id = oli.id
+          AND oi.deleted_at IS NULL
+        GROUP BY oli.id
+      )
+      UPDATE "order_line_item_adjustment" olia
+      SET version = latest_order_item_version.version
+      FROM latest_order_item_version
+      WHERE olia.item_id = latest_order_item_version.item_id
+        AND olia.version <> latest_order_item_version.version;
+    `)
```
olivermrbl, line 25: thought: I wonder if this migration would be more suited as a data migration script and then run it in batches? This could potentially affect A LOT of orders and line items.

@adrien2p @carlos-r-l-rodrigues would maybe be helpful to get your eyes on this

olivermrbl: What do you think @fPolic and @adrien2p?

fPolic: Migrations and data migration scripts are not executed in the same TX? If that's the case, if something fails in the script, line item adjustments will keep `version` 1 and won't be returned for the edited orders properly.

PR #13311, opened 2025-08-27
chore(index): check existing entities

carlos-r-l-rodrigues:
FIXES: CORE-1141


adrien2p (commented):
I ll wait for juan feedback on that as it seems we are missing something right?

PR #13312, opened 2025-08-27
chore(framework,medusa): load custom flags before medusa config

carlos-r-l-rodrigues:
What:
 - Loads custom feature flags from a Medusa project before the medusa-config itself. That makes the flags available into `medusa-config.ts` file and all other resources


adrien2p (approved):
Nice catch 🎉

adrien2p:
shall we catch that in the integration tests?

.changeset/good-meals-watch.md
```diff
@@ -0,0 +1,7 @@
+---
+"@medusajs/framework": patch
```
adrien2p, line 2: todo: update changeset

adrien2p (approved):
LGTM, is there a need to change anything around the test runners?

.changeset/good-meals-watch.md
olivermrbl: Why do we need to do that?

carlos-r-l-rodrigues: just because more packages were modified.
they are probably covered in multiple other PRs, but just to keep the good habit haha

carlos-r-l-rodrigues:
> LGTM, is there a need to change anything around the test runners?

Yes, we have to load the flags before initializing the test, otherwise it will require the medusa-config file before the flags are loaded and it will cause issues if there is any `FeatureFlag` usage in it.

adrien2p:
> > LGTM, is there a need to change anything around the test runners?
> 
> Yes, we have to load the flags before initializing the test, otherwise it will require the medusa-config file before the flags are loaded and it will cause issues if there is any `FeatureFlag` usage in it.

Perfect, just missed the file and thought it was missing 👌

PR #13314, opened 2025-08-27
chore: Enable draft order plugin by default

olivermrbl:
Final todo: I will need to check if removing `@medusajs/medusa` as a peer deps changes anything for the plugin


packages/core/utils/src/common/define-config.ts
```diff
@@ -124,6 +125,24 @@ export function transformModules(
...
 
+function resolvePlugins(
+  configPlugins: InputConfig["plugins"],
+  { isCloud }: { isCloud: boolean }
+): ConfigModule["plugins"] {
+  const defaultPlugins: ConfigModule["plugins"] = [
+    { resolve: "@medusajs/draft-order", options: {} },
+  ]
+
+  if (!!configPlugins?.length) {
```
willbouch: nit: dont think the !! are necessary here? if undefined, then if (undefined) will yield false and if empty array, if (0) will yield false

willbouch (approved):

fPolic (approved):

carlos-r-l-rodrigues (approved):

packages/core/utils/src/common/define-config.ts
```diff
@@ -124,6 +125,24 @@ export function transformModules(
   return remappedModules as Exclude<ConfigModule["modules"], undefined>
 }
 
+function resolvePlugins(
```
adrien2p, line 128: suggestion: this looks good but I believe we should handle merging, like we do for the modules wdyt? What I mean is that if a plugin has default options, maybe the user can register the same plugin with different options instead which in that case here would duplicate the plugin. Instead I would suggest the following as a first stab:

```ts
  const defaultPlugins: Map<string, ConfigModule["plugins"][number]> = new Map([
    ["@medusajs/draft-order", { resolve: "@medusajs/draft-order", options: {} }],
  ])

  if (isCloud) {
    return Array.from(defaultPlugins.values())
  }

  if (configPlugins?.length) {
    configPlugins.forEach((plugin) => {
      if (typeof plugin === "string") {
        defaultPlugins.set(plugin, plugin)
      } else {
        defaultPlugins.set(plugin.resolve, plugin)
      }
    })
  }

  return Array.from(defaultPlugins.values())
```

adrien2p (approved):
LGTM 🚀 just a little suggestion

packages/core/utils/src/common/define-config.ts
olivermrbl: Yeah, good idea

adrien2p (approved):

PR #13321, opened 2025-08-28
chore(core-flows): order payment status fix

carlos-r-l-rodrigues:
What:
 - Consider epsilon to check if an order is fully captured or refunded.


adrien2p (approved):
Good catch 🎉

olivermrbl (approved):

PR #13323, opened 2025-08-28
chore(dashboard): move shipping option type page

willbouch:


fPolic (approved):
Works as expected!

nit: while testing I noticed that we have edit success toast rendering template wrong

PR #13324, opened 2025-08-28
fix(dashboard): rules form operator change

fPolic:
**What**
- fix condition rules form sometimes rendering multi and single selects unrelated to the attribute operator selected
- reset rule value when operator is changed
- disable selecting rule values untill operator is selected
- translate labels


willbouch (approved):
LGTM 🫡

PR #13325, opened 2025-08-28
chore(): faster serialization

adrien2p:
CLOSES CORE-1151

**What**

### Serializer Performance Improvements

The custom MikroORM serializer has been optimized to significantly improve serialization times across varying entity depths.

#### Benchmark Setup
- For each batch size (`entities`), the test looped **10 times**.  
- On each iteration, the specified number of entities was created and then serialized.  
- Reported times are the **average per batch**.  
- The table below compares the old serializer with the new optimized version.

Results for all entity counts:
10 entities: 76.78% improvement (MIKRO ORM NATIVE SERIALIZER: 2.01ms, NEW: 0.47ms)
50 entities: 75.07% improvement (MIKRO ORM NATIVE SERIALIZER: 12.73ms, NEW: 3.17ms)
100 entities: 79.91% improvement (MIKRO ORM NATIVE SERIALIZER: 11.36ms, NEW: 2.28ms)
1,000 entities: 80.85% improvement (MIKRO ORM NATIVE SERIALIZER: 110.87ms, NEW: 21.23ms)
10,000 entities: 79.84% improvement (MIKRO ORM NATIVE SERIALIZER: 1026.05ms, NEW: 206.90ms)


packages/core/utils/src/dal/mikro-orm/mikro-orm-serializer.ts
```diff
@@ -9,65 +14,112 @@ import {
...
+  populate?: string[] | boolean
+  exclude?: string[]
   preventCircularRef?: boolean
-  populate?: [keyof T][] | boolean
+  skipNull?: boolean
+  ignoreSerializers?: boolean
+  forceObject?: boolean
+} = {
+  populate: ["*"],
+  exclude: undefined,
```
willbouch, line 29: why the undefined? If the type says `exclude?: string[]`, we could simply omit no? Not, if that has anything to do with V8 and undefined being set serves a purpose, maybe the correct typing should be `exclude: string[] | undefined`?

packages/core/utils/src/dal/mikro-orm/mikro-orm-serializer.ts
```diff
@@ -88,38 +140,53 @@ function filterEntityPropToSerialize({
...
   static serialize<T extends object, P extends string = never>(
     entity: T,
-    options: CustomSerializeOptions<T, P> = {},
-    parents: string[] = []
+    options = STATIC_OPTIONS_SHAPE,
+    parents: string[] = EMPTY_ARRAY
   ): EntityDTO<Loaded<T, P>> {
-    const parents_ = Array.from(new Set(parents))
+    const parents_ =
+      parents.length > 0 ? Array.from(new Set(parents)) : this.EMPTY_PARENTS
```
willbouch: What is the goal with setting to `this.EMPTY_PARENTS`? Can it not just stay `EMPTY_ARRAY`? Or is the reference to `this.EMPTY_PARENTS` important?

packages/core/utils/src/dal/mikro-orm/mikro-orm-serializer.ts
```diff
@@ -141,62 +208,74 @@ export class EntitySerializer {
     }
 
     const ret = {} as EntityDTO<Loaded<T, P>>
-    const keys = new Set<string>(meta.primaryKeys)
-    Object.keys(entity).forEach((prop) => keys.add(prop))
+
+    const keys = this.EMPTY_KEYS
+    keys.clear()
```
willbouch: Just curious, why the clear here?

packages/core/utils/src/dal/mikro-orm/mikro-orm-serializer.ts
```diff
@@ -141,62 +208,74 @@ export class EntitySerializer {
...
+
+    const primaryKeys = meta.primaryKeys
+    const primaryKeysLen = primaryKeys.length
+    for (let i = 0; i < primaryKeysLen; i++) {
```
willbouch, line 228: Also curious, I assume it is because for loops are faster than for each here? I guess you want to keep everything as low-level as possible to save some overhead?

packages/core/utils/src/dal/mikro-orm/mikro-orm-serializer.ts
```diff
@@ -141,62 +208,74 @@ export class EntitySerializer {
...
     }
 
-    ;[...keys]
-      /** Medusa Custom properties filtering **/
-      .filter((prop) =>
-        filterEntityPropToSerialize({
+    const keysArray = Array.from(keys)
+    const keysLen = keysArray.length
+
+    const className = meta.className
```
willbouch, line 247: Again, just because I am curious, your goal here is to make sure we don't lose time reassigning a variable every iteration of the loop correct? So setting `const className = meta.className` saves us time rather than always re-accessing it multiple times?

packages/core/utils/src/dal/mikro-orm/mikro-orm-serializer.ts
```diff
@@ -206,39 +285,38 @@ export class EntitySerializer {
...
-        (prop) =>
-          (ret[this.propertyName(meta, prop.name, wrapped.__platform)] =
-            this.processProperty(
-              prop.getterName as keyof T & string,
-              entity,
-              options,
-              parents_
-            ))
-      )
+    for (let i = 0; i < metaPropsLen; i++) {
```
willbouch: could we not make this one for loop? We already iterate over `metaProps` at line 291

packages/core/utils/src/dal/mikro-orm/mikro-orm-serializer.ts
```diff
@@ -248,33 +326,53 @@ export class EntitySerializer {
...
+      result = meta.properties[prop].serializedName as string
+    } else if (meta.properties[prop]?.primary && platform) {
+      result = platform.getSerializedPrimaryKeyField(prop) as string
+    } else {
+      result = prop
     }
 
-    if (meta.properties[prop]?.primary && platform) {
-      return platform.getSerializedPrimaryKeyField(prop) as string
+    if (this.propertyCache.size >= this.PROPERTY_CACHE_SIZE) {
```
willbouch: Can you explain a bit more the cache size? Why 2000? Why clear entirely? Faster than clearing a few entries?

packages/core/utils/src/dal/mikro-orm/mikro-orm-serializer.ts
```diff
@@ -248,33 +326,53 @@ export class EntitySerializer {
...
+    parents: string[] = EMPTY_ARRAY
   ): T[keyof T] | undefined {
-    const parents_ = [...parents, entity.constructor.name]
+    const parents_ =
+      parents.length > 0
+        ? [...parents, entity.constructor.name]
+        : [entity.constructor.name]
 
-    const parts = prop.split(".")
+    const parts = prop.split(DOT)
```
willbouch, line 418: Again, question for my personal knowledge, accessing DOT here is faster than recreating a new string I assume?

packages/core/utils/src/dal/mikro-orm/mikro-orm-serializer.ts
```diff
@@ -334,41 +432,73 @@ export class EntitySerializer {
...
-        .filter((field) => field.startsWith(`${prop}.`))
-        .map((field) => field.substring(prop.length + 1))
+      const result: string[] = []
+      const itemsLen = items.length
+
+      for (let i = 0; i < itemsLen; i++) {
+        const field = items[i]
+        if (
+          field.length > propPrefixLen &&
+          field.substring(0, propPrefixLen) === propPrefix
```
willbouch, line 505: I wonder here. Would it be faster to split the string in to directly? Is it faster to run 2 substrings, or one split?

willbouch (approved):
It looks good! I get the idea a lot more now that it's in the morning 😅. Few comments but nothing major. Some of them are questions that you can answer here, or you can call me on Slack if you feel like it is easier

packages/core/utils/src/dal/mikro-orm/mikro-orm-serializer.ts
adrien2p: just to have the property present at its position

packages/core/utils/src/dal/mikro-orm/mikro-orm-serializer.ts
adrien2p: I need to test something around these but we are sharing the same ref, parent has a ref and array another ref and they are used in different places

packages/core/utils/src/dal/mikro-orm/mikro-orm-serializer.ts
adrien2p: all these I will change because we are sharing the same reference in concurrent execution and I would prefer loosing some perf that incresing conflict access

packages/core/utils/src/dal/mikro-orm/mikro-orm-serializer.ts
adrien2p: I ll investigate that 👍

packages/core/utils/src/dal/mikro-orm/mikro-orm-serializer.ts
adrien2p: mayeb could just drop the first one, I need to test

packages/core/utils/src/dal/mikro-orm/mikro-orm-serializer.ts
adrien2p: Screenshot 2025-08-29 at 14 33 38
Screenshot 2025-08-29 at 14 38 09

subsctring is always faster

packages/core/utils/src/dal/mikro-orm/mikro-orm-serializer.ts
willbouch: In this case, doesn't `exclude: string[] | undefined` make more sense?

packages/core/utils/src/dal/mikro-orm/mikro-orm-serializer.ts
willbouch: Thanks for checking!

packages/core/utils/src/dal/mikro-orm/mikro-orm-serializer.ts
adrien2p: yes indeed, that would be better thanks 👍

olivermrbl (approved):
LGTM 🔥

willbouch (approved):
I had no idea you could optimize that much with minor changes like this in JS! Great work

carlos-r-l-rodrigues (approved):
Insane work 🔥 💪🏻 

Q: if we replace the `typeof x`  `"undefined"` with `isDefined`, would it affect the performance?

PR #13353, opened 2025-08-29
fix(core-flows): handle cacluated shipping options on draft orders gracefully

fPolic:
**Why**
- calculated shipping options are not yet supported on draft orders

**What**
- allow adding if a custom amount is provided or handle error explicitly if a calculated option is added


olivermrbl (approved):
LGTM, will you create a ticket in the catch-all improvement project? Not sure we wil tackle it, but good to have on record

fPolic:
yup, created CORE-1161

PR #13360, opened 2025-08-31
fix(draft-order): use correct backend URL in sdk creation

### LINKED ISSUE #13337: baseUrl "/" in draft order plugin causes problems when backend is on a different domain than dashboard
### Describe the bug

Currently, in `packages/plugins/draft-order/src/admin/lib/queries/sdk.ts`, the Medusa js-sdk is initialized with:

```typescript
export const sdk = new Medusa({
  baseUrl: "/",
  auth: {
    type: "session",
  },
})
```

This hardcoded `baseUrl` value of `/` poses a problem when the backend is hosted on a different domain than the dashboard. In such cases, API requests fail due to incorrect routing or CORS issues, since the SDK assumes the backend is on the same domain as the dashboard.

1. Host the backend and dashboard on different domains.
2. Try to perform any API action using the dashboard.
3. Observe failed requests and errors due to incorrect baseUrl resolution.

### Expected behavior
Developers should be able to configure the backend URL separately, allowing the dashboard to communicate with backends on different domains.

### Code snippets
See issue description above for SDK initialization.

### Additional context
It is recommended to add a parameter in `medusa-config.ts` to specify the baseUrl in the plugin config, making the SDK initialization flexible and allowing cross-domain setups.

leobenzol:
Closes #13337 

If the admin server is deployed separately, the sdk object needs the correct url i.e. `__BACKEND_URL__`. This is how it's done already in the `dashboard` package.

### DIFF
--- a/packages/plugins/draft-order/src/admin/lib/queries/sdk.ts
+++ b/packages/plugins/draft-order/src/admin/lib/queries/sdk.ts
@@ -1,7 +1,7 @@
 import Medusa from "@medusajs/js-sdk"
 
 export const sdk = new Medusa({
-  baseUrl: "/",
+  baseUrl: __BACKEND_URL__ || "/",
   auth: {
     type: "session",
   },
--- a/.changeset/hungry-terms-report.md
+++ b/.changeset/hungry-terms-report.md
@@ -1,0 +1,5 @@
+---
+"@medusajs/draft-order": patch
+---
+
+fix(draft-order): use correct backend URL in sdk creation
--- a/packages/plugins/draft-order/src/admin/vite-env.d.ts
+++ b/packages/plugins/draft-order/src/admin/vite-env.d.ts
@@ -1,0 +1,1 @@
+declare const __BACKEND_URL__: string | undefined


willbouch (approved):
@olivermrbl can you validate and check that one out? Would be good to have it in the next release

packages/plugins/draft-order/src/admin/lib/queries/sdk.ts
```diff
@@ -1,7 +1,7 @@
 import Medusa from "@medusajs/js-sdk"
 
 export const sdk = new Medusa({
-  baseUrl: "/",
+  baseUrl: __BACKEND_URL__ || "/",
```
adrien2p, line 4: Q: are you sure about that? shouldn't it comes from the env?

leobenzol: Yes, process.env is forbidden by vite and import.meta.env gets replaced statically at plugin build time(with undefined). It's exactly what the dashboard package does for its sdks and what I had to do in my own plugins.

It also opens the question of how to have a proper system to expose env vars to plugin UIs, but that's for another time and for now luckily only `__BACKEND_URL__` is really needed.

willbouch:
Checked with Oli and all good, we can merge

PR #13371, opened 2025-09-01
fix(dashboard): promotion decimal value definition

fPolic:
**What**
- allow decimal values for percentage and fixed promotions on edit and create

---

CLOSES CORE-1002

### DIFF
--- a/packages/admin/dashboard/src/routes/promotions/promotion-create/components/create-promotion-form/create-promotion-form.tsx
+++ b/packages/admin/dashboard/src/routes/promotions/promotion-create/components/create-promotion-form/create-promotion-form.tsx
@@ -36,7 +36,10 @@ import {
 import { KeyboundForm } from "../../../../../components/utilities/keybound-form"
 import { useCampaigns } from "../../../../../hooks/api/campaigns"
 import { useCreatePromotion } from "../../../../../hooks/api/promotions"
-import { getCurrencySymbol } from "../../../../../lib/data/currencies"
+import {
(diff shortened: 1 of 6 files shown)


packages/admin/dashboard/src/routes/promotions/promotion-create/components/create-promotion-form/create-promotion-form.tsx
```diff
@@ -783,7 +796,7 @@ export const CreatePromotionForm = () => {
                                       onChange(
                                         e.target.value === ""
                                           ? null
-                                          : parseInt(e.target.value)
+                                          : parseFloat(e.target.value)
```
willbouch, line 799: Was that intended? BEcause having a parseInt there looks like something we wanted haha

willbouch (approved):

packages/admin/dashboard/src/routes/promotions/promotion-create/components/create-promotion-form/create-promotion-form.tsx
fPolic: Not sure, maybe @riqwan has a better context here, but don't see a reason why users shouldn't specify promotion value with decimals.

willbouch: Yeah agreed

willbouch (approved):

PR #13373, opened 2025-09-01
fix(dashboard): promotion expired status check when limit is null

pepijn-vanvlaanderen:
This PR fixes an UI issue on the promotion status when limit is null, see also here:

Scherm­afbeelding 2025-09-01 om 17 39 04

### DIFF
--- a/.changeset/bright-spoons-float.md
+++ b/.changeset/bright-spoons-float.md
@@ -1,5 +1,5 @@
 ---
-"@medusajs/dashboard": major
+"@medusajs/dashboard": patch
 ---
 
 Fix promotion expired status check when limit is null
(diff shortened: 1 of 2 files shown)


carlos-r-l-rodrigues (approved):

olivermrbl (approved):

.changeset/bright-spoons-float.md
```diff
@@ -0,0 +1,5 @@
+---
+"@medusajs/dashboard": major
```
olivermrbl: todo: Don't think we need a major bump here though 😄

olivermrbl:
```suggestion
"@medusajs/dashboard": patch
```

fPolic (approved):

.changeset/bright-spoons-float.md
pepijn-vanvlaanderen: My bad was a bit to quick about it :)

PR #13375, opened 2025-09-01
fix(medusa): Use default logger in plugin:develop

olivermrbl:
**What**

Use default logger for plugin command, since there is no `medusa-config.ts` to load the container with (the container that holds the custom logger)


fPolic (approved):

carlos-r-l-rodrigues:
I believe the last PR you force merged will fix this, and the default logger will be used.

olivermrbl:
> I believe the last PR you force merged will fix this, and the default logger will be used.

Alright, lets merge that first and give it a test :)

leobenzol:
I noticed last night that the `plugin:db:generate`also has the same problem, just wanted to let yall know :)

olivermrbl:
@carlos-r-l-rodrigues, are you sure this is the case? I just tested a fresh snapshot, and the issue remains 🤔

olivermrbl:
I believe [this line](https://github.com/medusajs/medusa/blob/ff75641aed38999e6e682ffe2b23b1fc1f04cc70/packages/medusa/src/commands/plugin/develop.ts#L13) will continue to fail, because it attempts to load `medusa-config.ts`

carlos-r-l-rodrigues:
True, this is a special case. let's just use the default logger.

carlos-r-l-rodrigues (approved):

willbouch (approved):

PR #13376, opened 2025-09-01
fix(draft-order, medusa, types): ui polish

fPolic:
**What**
- ensure unique promotion changes are displayed in the activity section
- check linked promotions by id and not code
- list promotions - add id param
- make address payload fields nullish
- Add items view UI tweaks

---

CLOSES SUP-2400


packages/core/types/src/http/order/common.ts
```diff
@@ -1078,6 +1078,11 @@ export interface BaseOrderChange {
  * The order change action details.
  */
 export interface BaseOrderChangeAction {
+  /**
+   * The ordering of the order change action
```
willbouch, line 1082: Was not clear to me at first glance. Maybe adding something to specify that we use it for sorting? Or maybe the field could be called "priority"?

fPolic: it is used to keep track of the order of actions of an order change, it was just useful for sorting in this case

willbouch (approved):

PR #13379, opened 2025-09-01
docs: change digial products to digital products

yoreljenkins:
Small fix for typo I noticed when reading the docs.


shahednasser (approved):
Nice catch thanks!

PR #13385, opened 2025-09-02
feat: Add return type in createPaymentCollectionForCartWorkflow

gladius882:
This PR introduces a small change in createPaymentCollectionForCartWorkflow. By default, it returns nothing. I've added a WorkflowResponse to return the created PaymentCollection.

In case someone needs to work with this collection, it will now be easier to operate on the returned object instead of retrieving it manually.


.changeset/tasty-ghosts-work.md
```diff
@@ -0,0 +1,5 @@
+---
+"@medusajs/core-flows": major
```
carlos-r-l-rodrigues:
```suggestion
"@medusajs/core-flows": patch
```

carlos-r-l-rodrigues (approved):

carlos-r-l-rodrigues:
Thanks @gladius882

carlos-r-l-rodrigues:
@gladius882, could you fix that?

`Error: src/cart/workflows/create-payment-collection-for-cart.ts(135,5): error TS2322: Type 'WorkflowResponse<PaymentCollectionDTO[], []>' is not assignable to type 'WorkflowData<void>'.`

you can run `yarn build` on the root folder from the repo to check if everything is good.

gladius882:
Beginner contributor here, so thanks for pointing that out.
I've pushed the changes.

packages/core/core-flows/src/cart/workflows/create-payment-collection-for-cart.ts
```diff
@@ -130,5 +132,7 @@ export const createPaymentCollectionForCartWorkflow = createWorkflow(
     createRemoteLinkStep(cartPaymentLink).config({
       name: "cart-payment-collection-link",
     })
+
+    return new WorkflowResponse(created)
```
carlos-r-l-rodrigues: nit: this flow only creates one payment collection, we could return just `created[0]`

willbouch (approved):
Thank you for the contribution!

PR #13396, opened 2025-09-02
chore(): only execute race execution checks and publish message only for async workflows

adrien2p:
RESOLVES CORE-1162

**what**
only check race and publish workflow event messages for async workflow

### DIFF
--- a/packages/modules/workflow-engine-redis/src/services/workflow-orchestrator.ts
+++ b/packages/modules/workflow-engine-redis/src/services/workflow-orchestrator.ts
@@ -685,7 +685,7 @@ export class WorkflowOrchestratorService {
       state,
     } = options
 
-    if (publish && !isFlowAsync) {
+    if (publish && isFlowAsync) {
       const channel = this.getChannelName(options.workflowId)
       const message = JSON.stringify({
         instanceId: this.instanceId,
--- a/packages/modules/workflow-engine-redis/src/services/workflow-orchestrator.ts
+++ b/packages/modules/workflow-engine-redis/src/services/workflow-orchestrator.ts
@@ -64,6 +64,7 @@ type IdempotencyKeyParts = {
 
 type NotifyOptions = {
   eventType: keyof DistributedTransactionEvents
+  isFlowAsync: boolean
   workflowId: string
   transactionId?: string
   step?: TransactionStep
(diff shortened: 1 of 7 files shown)


packages/modules/workflow-engine-redis/src/services/workflow-orchestrator.ts
```diff
@@ -689,6 +685,15 @@ export class WorkflowOrchestratorService {
       state,
     } = options
 
+    if (publish && !isFlowAsync) {
```
olivermrbl: thought: this might be obvious if you know the implementation details, but I read this if-statement as such "Publish message if publish is true and **the flow is not asynchronous**" which kind of contradicts the intention. Am I missing something?

adrien2p: oih my bad, you are right 😓 for some reason I didnt see I ve put a negation there nice catch

olivermrbl (approved):
LGTM

packages/modules/workflow-engine-redis/src/services/workflow-orchestrator.ts
```diff
@@ -689,6 +685,15 @@ export class WorkflowOrchestratorService {
       state,
     } = options
 
+    if (publish && isFlowAsync) {
+      const channel = this.getChannelName(options.workflowId)
+      const message = JSON.stringify({
+        instanceId: this.instanceId,
+        data: options,
+      })
+      await this.redisPublisher.publish(channel, message)
```
carlos-r-l-rodrigues, line 694: The only issue I see doing this, is that the subscribers from other nodes won't receive the events for all the steps.

adrien2p: but if the workflow is synchronous, it will run on the node you are running it, and the subsrbiers will be all the locals triggered bellow

carlos-r-l-rodrigues: but when there are multiple instances or worker/server, we can have different subscribers.

adrien2p: if one instance has a wfe.subscribe, all instances will have it as well since they all run the same application code

adrien2p: including the one running the synchronous workflow

carlos-r-l-rodrigues: well, this is true. at the moment we need to cover the core flows and medusa projects, not any blank project that uses the framework/workflows.

carlos-r-l-rodrigues (approved):

PR #13404, opened 2025-09-03
fix(dashboard): edit promotion campaign w/wo currrency

fPolic:
**What**
- a promotion couldn't be added to a campaign without a currency budget (on promotion details screen)
- fix fetching campaigns pagination issue
- move select to Combobox
- fix currency restriction and move warning description to the label hint


olivermrbl:
> a promotion couldn't be added to a campaign without a currency budget (on promotion details screen)

Can you expand upon this a bit? Not immediately obvious what the issue was.

fPolic:
> Can you expand upon this a bit?

If there is a fixed amount promotion (which has a currency defined), it cannot be added to a campaign that has a use limit budget because `currency_code` on that budget is `null` and it doesn't match `currency_code` of the promotion.

olivermrbl:
I can't seem to reproduce. I just tried creating a fixed promotion with currency EUR, and a campaign with a usage limit of 10. I could add the promotion to the campaign. Am I missing a step here?

fPolic:
@olivermrbl - this is working if you try to add promotion to the campaign from the campaign details page, but the issue is on the promotion details page when you try to add it to a campaign:

https://github.com/user-attachments/assets/20e1c8a6-60c2-4480-bb1e-59950bc867c7

olivermrbl:
> @olivermrbl - this is working if you try to add promotion to the campaign from the campaign details page, but the issue is on the promotion details page when you try to add it to a campaign:
> 
>  Screen.Recording.2025-09-24.at.09.17.47.mov

Ah yes, I didn't try it this way around. I see it now!

olivermrbl (approved):

PR #13407, opened 2025-09-03
 fix(dashboard): support more decimals for tx rates

willbouch:
**WHAT**: Allowing 4 decimals in tax rates +. small refactor of the percentage input so that we don't show `10.0000` by default if the tax rate is 10 for example

CLOSES CORE-1166


packages/admin/dashboard/src/components/inputs/percentage-input/percentage-input.tsx
```diff
@@ -37,39 +37,53 @@ DeprecatedPercentageInput.displayName = "PercentageInput"
...
+    },
+    ref
+  ) => {
+    return (
+      <div className="relative">
+        <Primitive
+          ref={ref as any} // dependency is typed incorrectly
+          min={min}
+          autoComplete="off"
+          decimalScale={bypassDefaultDecimalScale ? undefined : decimalScale}
```
willbouch: Ok so the property that was passed everywhere is `decimalScale`. And `decimalsLimit` was set to `decimalScale`. This made the percentage input look like this:
CleanShot 2025-09-03 at 12 50 01@2x


To avoid breaking changes, I did a bit of gymnastics. I made it so that `decimalsLimit` can be set independently from `decimalScale`. Setting `decimalsLimit` to 4 will allow the user to input up to 4 decimals without having the trailing 0s.

packages/admin/dashboard/src/routes/promotions/promotion-create/components/create-promotion-form/create-promotion-form.tsx
```diff
@@ -772,20 +769,17 @@ export const CreatePromotionForm = () => {
                                     disabled={!currencyCode}
                                   />
                                 ) : (
-                                  <DeprecatedPercentageInput
-                                    key="amount"
-                                    className="text-right"
-                                    min={0}
-                                    max={100}
+                                  <PercentageInput
```
willbouch: Figured I would replace the Deprecated from the alst 2 places it is used so that I can delete it for good

packages/admin/dashboard/src/components/inputs/percentage-input/percentage-input.tsx
```diff
@@ -1,75 +1,75 @@
...
+        <Primitive
+          ref={ref as any} // dependency is typed incorrectly
+          min={min}
+          max={max}
+          autoComplete="off"
+          decimalScale={resolvedDecimalScale}
+          decimalsLimit={resolvedDecimalsLimit}
+          value={value}
+          {...props}
+          className={clx(
```
willbouch, line 80: Moved the % from right to left. Looks better with the other inputs

packages/admin/dashboard/src/routes/promotions/promotion-create/components/create-promotion-form/create-promotion-form.tsx
willbouch: Nevermind. Will handle in another PR at some point. It breaks it more than I had expected and I don't want to delay the real fix to be merged

packages/admin/dashboard/src/components/inputs/percentage-input/percentage-input.tsx
```diff
@@ -1,10 +1,22 @@
-import { Input, Text, clx } from "@medusajs/ui"
+import { clx, Input, Text } from "@medusajs/ui"
+import { getNumberOfDecimalPlaces } from "../../../lib/number-helper.ts"
```
fPolic: nit: rm extension

packages/admin/dashboard/src/lib/number-helper.ts
```diff
@@ -0,0 +1,19 @@
...
+ * Gets the number of decimal places in a number
+ * @param num - The number for which we are getting the number of decimal places
+ * @returns The number of decimal places
+ *
+ * @example
+ * getDecimalPlaces(123.456) // 3
+ * getDecimalPlaces(42) // 0
+ * getDecimalPlaces(10.0000) // 0
+ */
+export function getNumberOfDecimalPlaces(num: number | string): number {
```
fPolic: suggestion(non-blocking): might be better to always expect a number here because if a string is passed formatted like: `12.000,50`, for certain locales this might produce an incorrect result. We could separate normalisation/conversion from this or account for formatting differences.

fPolic (commented):
Works as expected!
Left 2 comments.

packages/admin/dashboard/src/lib/number-helper.ts
willbouch: Yes good point!

olivermrbl (approved):
LGTM!

olivermrbl:
@fPolic, could you have a look at this?

fPolic (approved):

PR #13413, opened 2025-09-04
chore(workflows-sdk, utils): Prevent unnecessary serialization

adrien2p:
RESOLVES CORE-1154 ([here](https://github.com/medusajs/medusa/pull/13365) some explanation why only doing that)

**What**

Workflow are heavely relying on the value resolution, which always parse/stringify data to prevent shared references. But in a lot of cases, these value can be primitive only and in that case there is no need to perform such action

It happen to be more than 3X faster to do nothing instead of extraneous operation plus it will have less impact on the memory and GC.

Screenshot 2025-09-04 at 14 48 36


packages/core/utils/src/common/parse-stringify-if-necessary.ts
```diff
@@ -0,0 +1,18 @@
+import { isDefined } from "./is-defined"
+
+/**
+ * Only apply JSON.parse JSON.stringify when we have objects, arrays, dates, etc..
+ * @param result
+ * @returns
+ */
+export function parseStringifyIfNecessary(result: unknown) {
+  if (typeof result !== "object") {
```
adrien2p, line 9: self-review: I purpusfully do not use `isObject` because I want to include Map, Set, Date etc

olivermrbl (approved):

willbouch (approved):
again 🐐

Issue #11386, opened 2025-02-11
[Bug]: Error: Could not find index.html in the admin build directory. Make sure to run 'medusa build' before starting the server.


swatiakkishetti:
### Package.json file

```JSON
{
  "name": "medusa-starter-default",
  "version": "0.0.1",
  "description": "A starter for Medusa projects.",
  "author": "Medusa (https://medusajs.com)",
  "license": "MIT",
  "keywords": [
    "sqlite",
    "postgres",
    "typescript",
    "ecommerce",
    "headless",
    "medusa"
  ],
  "scripts": {
    "dev": "medusa develop",
    "build": "medusa build && ln -s .medusa/server/public/ public",
    "start": "NODE_ENV=production medusa start -H 0.0.0.0",
    "seed": "medusa exec ./src/scripts/user-seed.ts && medusa exec ./src/scripts/seed.ts",
    "migrate": "medusa db:migrate",
    "sync": "medusa db:sync-links",
    "migrate:gen": "npx medusa db:generate",
    "test": "NODE_OPTIONS=--experimental-vm-modules jest --testPathPattern=spec\\.ts$ --maxWorkers=4",
    "test:watch": "jest --watch --testPathPattern=spec\\.ts$",
    "test:coverage:local": "jest --coverage --testPathPattern=spec\\.ts$",
    "test:coverage": "jest --coverage",
    "lint": "yarn eslint",
    "lint:fix": "yarn eslint --fix",
    "prepare": "husky",
    "test:env": "cross-env NODE_ENV=test yarn run dev",
    "test:integration": "jest integration-tests"
  },
  "dependencies": {
    "@aws-sdk/client-s3": "^3.658.1",
    "@aws-sdk/client-ses": "^3.637.0",
    "@aws-sdk/credential-provider-web-identity": "^3.621.0",
    "@aws-sdk/s3-request-presigner": "^3.675.0",
    "@medusajs/admin": "^7.1.17",
    "@medusajs/admin-sdk": "preview",
    "@medusajs/api-key": "preview",
    "@medusajs/auth": "preview",
    "@medusajs/auth-emailpass": "preview",
    "@medusajs/cache-inmemory": "preview",
    "@medusajs/cart": "preview",
    "@medusajs/cli": "preview",
    "@medusajs/currency": "preview",
    "@medusajs/customer": "preview",
    "@medusajs/event-bus-local": "preview",
    "@medusajs/event-bus-redis": "preview",
    "@medusajs/file": "preview",
[...]

PR #13415, opened 2025-09-04
fix(medusa): corrected admin output dir

trevster344:
#11386 notes a problem with starting medusa where the admin output path is slightly incorrect.

I ran into this issue again when I deleted my development database and ran `db:setup`.

Please correct me if I'm wrong but my testing led to success when I adjusted this path to reflect the structure from previous versions.

### DIFF
--- a/.changeset/empty-beers-taste.md
+++ b/.changeset/empty-beers-taste.md
@@ -1,0 +1,5 @@
+---
+"@medusajs/medusa": patch
+---
+
+Corrected the admin relative output dir to reflect previous file structure
--- a/.changeset/empty-beers-taste.md
+++ b/.changeset/empty-beers-taste.md
@@ -2,4 +2,4 @@
 "@medusajs/medusa": patch
 ---
 
-Corrected the admin relative output dir to reflect previous file structure
+fix(medusa): corrected admin output dir
--- a/packages/medusa/src/utils/admin-consts.ts
+++ b/packages/medusa/src/utils/admin-consts.ts
@@ -1,3 +1,3 @@
 export const ADMIN_SOURCE_DIR = "src/admin"
-export const ADMIN_RELATIVE_OUTPUT_DIR = "./public/admin"
+export const ADMIN_RELATIVE_OUTPUT_DIR = ".medusa/server/public/admin"
 export const ADMIN_ONLY_OUTPUT_DIR = ".medusa/admin"


willbouch:
Hello! Thanks for the contribution. Do you think you could give the PR a more meaningful title 😅? Also, please add a changeset when you get the chance

.changeset/empty-beers-taste.md
```diff
@@ -0,0 +1,5 @@
+---
+"@medusajs/medusa": patch
+---
+
+Corrected the admin relative output dir to reflect previous file structure
```
willbouch: could you use your PR name here please? Makes the release doc look better ;)

trevster344: Sure! How about now?

willbouch:
I have never run into that issue though? The issue you linked is also closed. Could you give more details maybe?

trevster344:
@willbouch Sure I stumbled on the issue recently and found that closed issue however I did not see any commits addressing the source of the error.

Recently upgraded my medusa version from < 2.9.0 to the latest release. I deleted my test database and ran `db:create` and `db:setup`.

`error:   Could not find index.html in the admin build directory. Make sure to run 'medusa build' before starting the server.`

I traced the source of the error to this output directory const. Correcting it locally allowed medusa to start once again.

willbouch:
I unfortunately can't seem to reproduce :/

trevster344:
@willbouch 

> I unfortunately can't seem to reproduce :/

Can you place a console.log in the built admin loader file and see if outDir is correct? If so it would be worth figuring out how its correct for you given the changes not yet made. This may lead us to the real culprit if it's not the constant.

willbouch:
> @willbouch
> 
> > I unfortunately can't seem to reproduce :/
> 
> Can you place a console.log in the built admin loader file and see if outDir is correct? If so it would be worth figuring out how its correct for you given the changes not yet made. This may lead us to the real culprit if it's not the constant.



> @willbouch
> 
> > I unfortunately can't seem to reproduce :/
> 
> Can you place a console.log in the built admin loader file and see if outDir is correct? If so it would be worth figuring out how its correct for you given the changes not yet made. This may lead us to the real culprit if it's not the constant.

I was able to reproduce and following those steps made it work for me. Can you make sure to go through the documentation here? https://docs.medusajs.com/learn/build

trevster344:
> > @willbouch
> 
> > 
> 
> > > I unfortunately can't seem to reproduce :/
> 
> > 
> 
> > Can you place a console.log in the built admin loader file and see if outDir is correct? If so it would be worth figuring out how its correct for you given the changes not yet made. This may lead us to the real culprit if it's not the constant.
> 
> 
> 
> 
> 
> 
> 
> > @willbouch
> 
> > 
> 
> > > I unfortunately can't seem to reproduce :/
> 
> > 
> 
> > Can you place a console.log in the built admin loader file and see if outDir is correct? If so it would be worth figuring out how its correct for you given the changes not yet made. This may lead us to the real culprit if it's not the constant.
> 
> 
> 
> I was able to reproduce and following those steps made it work for me. Can you make sure to go through the documentation here? https://docs.medusajs.com/learn/build

Maybe this has to do with plugins. I didn't trace far enough to see if the issue was occurring from a plugin. I'll have to run some tests from a fresh Medusa and plugin.

trevster344:
Update:

A fresh medusa project and fresh plugin yielded no clues. All worked as expected. This must be something related to upgrading my older project to the latest version. All continues to be working. I'll rescind this request.

Thanks for input and patience!

PR #13418, opened 2025-09-05
chore(): Improve some cart operation flows to remove extraneous operations when not required

adrien2p:
RESOLVES CORE-1155

**What**
- prevent update payment collection worklow to fetch data and call steps when there is no payment collection
- prevent refresh Cart Shipping Methods Workflow to fetch data and call steps when there is no shipping methods
- update promotion step to remove extraneous module call


olivermrbl (approved):
LGTM, very nice

PR #13421, opened 2025-09-05
chore(): Improve workflows sdk tooling

adrien2p:
RESOLVES CORE-1171

**What**
 - Reduced async overhead for objects with mixed sync/async properties
 - Lower memory pressure from eliminated promise allocations
 - Faster primitive value processing with early returns
 - Better concurrency through selective batching of truly async operations
 - Event loop friendly behavior preventing artificial delays
 - Reduced memory allocation from eliminated duplicate processing
 - Decreased GC pressure from WeakMap-based caching instead of map

**note**
Now, `resolveValue`always treat every resoltion as sync operation unless it is not, meaning we do not create promise overhead when not necessary and only when actually treating with promises


packages/core/orchestration/src/transaction/distributed-transaction.ts
```diff
@@ -91,7 +91,7 @@ class DistributedTransaction extends EventEmitter {
    *
    * @private
    */
-  #temporaryStorage = new Map<string, unknown>()
+  #temporaryStorage = new WeakMap<{ key: string }, unknown>()
```
willbouch, line 94: I believe this is the first time I see that. You got me reading haha

packages/core/workflows-sdk/src/utils/composer/helpers/create-step-handler.ts
```diff
@@ -80,14 +80,25 @@ export function createStepHandler<
         stepArguments,
       })
 
-      const argInput = input ? await resolveValue(input, stepArguments) : {}
+      let argInput = {}
+      if (input) {
```
willbouch, line 84: why that change ? I tend to like ternary better but is there a reason behind the change?

willbouch (approved):

packages/core/workflows-sdk/src/utils/composer/helpers/create-step-handler.ts
adrien2p: Mental overload 😂 I had more statement that I ve removed

carlos-r-l-rodrigues (approved):

packages/core/workflows-sdk/src/utils/composer/helpers/create-step-handler.ts
adrien2p: do you want me to change it back or is it ok for this time?

willbouch: All good!

PR #13424, opened 2025-09-05
chore(core-flows): lock on cart operations

carlos-r-l-rodrigues:
What:
 - acquire a lock before cart mutations to avoid concurrent changes, e.g applying promotions.


packages/core/core-flows/src/locking/acquire-lock.ts
```diff
@@ -37,6 +41,11 @@ export const acquireLockStep = createStep(
       return new StepResponse(void 0)
     }
 
+    const isSubWorkflow = !!parentStepIdempotencyKey
+    if (isSubWorkflow && data.skipOnSubWorkflow) {
```
adrien2p, line 45: nit: should we skip on subworkflow true by default?

adrien2p (approved):

packages/core/core-flows/src/locking/acquire-lock.ts
carlos-r-l-rodrigues: no because it might be another key that is being locked.

olivermrbl (approved):

PR #13426, opened 2025-09-06
fix(dashboard): add missing translations to spanish file

galdoway:
**Summary:**
Complete Spanish translation file (es.json) by adding missing keys and cleaning up invalid entries.

- Add missing translation keys and values to es.json
- Remove unexpected/invalid keys that don't match schema
- Tests now pass with schema validation successful


willbouch (dismissed):
Thanks for the contribution. Took a look at it and it looks good to me

willbouch:
Do you mind just adding a changeset? You can run `yarn changeset` in the repo. In your case, it would be a patch. I updated the PR name to reflect the package that is being changed ;) you can use that one in the changeset too

galdoway:
> Do you mind just adding a changeset? You can run `yarn changeset` in the repo. In your case, it would be a patch. I updated the PR name to reflect the package that is being changed ;) you can use that one in the changeset too

@willbouch Changeset added! Thanks for your time :)

.changeset/tall-wasps-thank.md
```diff
@@ -0,0 +1,5 @@
+---
+"@medusajs/dashboard": minor
```
willbouch: can you make it a patch? This is not really a breaking change and we tend to reserve minor for breaking changes. Will approve once you do the change and we can get that merged today!

adrien2p (approved):

.changeset/tall-wasps-thank.md
galdoway: Right, I forgot to change that 😅. Done, it's patch now.

willbouch (approved):

PR #13435, opened 2025-09-08
Feat(): distributed caching

adrien2p:
RESOLVES CORE-1153

**What**
- This pr mainly lay the foundation the caching layer. It comes with a modules (built in memory cache) and a redis provider.
- Apply caching to few touch point to test


packages/modules/providers/caching-redis/src/services/redis-cache.ts
```diff
@@ -0,0 +1,276 @@
...
+  #getTagKey(tag: string): string {
+    return `${this.keyNamePrefix}tag:${tag}`
+  }
+
+  #getOptionsKey(key: string): string {
+    return `${this.keyNamePrefix}options:${key}`
+  }
+
+  async get({ key, tags }: { key?: string; tags?: string[] }): Promise<any> {
+    if (key) {
```
adrien2p, line 255: self-review: as carlos mention, we might need to store in memory the pair key/tags-promise and return the same promise for all callers

packages/modules/caching/src/utils/strategy.ts
```diff
@@ -0,0 +1,159 @@
+import {
```
adrien2p, line 7: self-review: add onapplication* hooks and move subscriber creation to onapplicationstart

packages/modules/caching/src/utils/strategy.ts
```diff
@@ -0,0 +1,159 @@
...
+    this.#cacheInvalidationParser = new CacheInvalidationParser(schema)
+    this.#container = container
+    this.#cacheModule = cacheModule
+
+    this.applyReactiveInvalidation()
+  }
+
+  applyReactiveInvalidation() {
+    const eventBus = this.#container[Modules.EVENT_BUS]
+
```
adrien2p, line 48: self-review: as discussed with carlos, this caching could expose a method to publish messages and the event bus if the cache is available could provide the events to it so that they get published as messages (pub/sub without bullmq) through the opened stream and not consumed the lua execution limit. then caching can just rely on messages to invalidate

packages/core/modules-sdk/src/remote-query/query.ts
```diff
@@ -259,17 +261,98 @@ export class Query {
...
+    ttl: extractCacheOptions("ttl"),
+    tags: extractCacheOptions("tags"),
+    autoInvalidate: extractCacheOptions("autoInvalidate"),
+    providers: extractCacheOptions("providers"),
+    container,
+  })(
+    Query.prototype,
+    "index",
+    Object.getOwnPropertyDescriptor(Query.prototype, "index")!
+  ).value
```
adrien2p: Self-review: I ll do some cleanup around here. Pass container to Query and apply decorator to the method directly

packages/core/framework/src/http/utils/refetch-entities.ts
```diff
@@ -1,49 +1,70 @@
...
     fields,
-  })
+    filters,
+    pagination,
+    withDeleted,
+    context: context,
+  }
 
-  return await remoteQuery(queryObject)
+  return (await query.graph(graphOptions, options)) as Omit<
```
adrien2p: Self-review: set metadata to empty object to prevent undefined to pass through and maintain some level of compatibility

packages/core/types/src/caching/index.ts
```diff
@@ -0,0 +1,143 @@
+import { ModuleJoinerConfig } from "../modules-sdk"
+
+type Providers =
+  | string[]
+  | { id: string; ttl?: number }
```
adrien2p: Self-review: rm

packages/core/types/src/joiner/index.ts
```diff
@@ -92,6 +94,39 @@ export interface RemoteJoinerOptions {
   throwIfRelationNotFound?: boolean | string[]
   initialData?: object | object[]
   initialDataOnly?: boolean
+  cache?: {
```
adrien2p, line 97: Self-review: align with the decorator/func api changes

packages/core/utils/src/caching/index.ts
```diff
@@ -0,0 +1,208 @@
...
+/**
+ * This function is used to cache the result of a method call.
+ *
+ * @param options - The options for the cache.
+ * @returns The original method with the cache applied.
+ */
+export function Cached<
+  const Target extends object,
+  const PropertyKey extends keyof Target
+>(options: {
```
adrien2p, line 98: Self-review: re write comments

packages/core/utils/src/modules-sdk/joiner-config-builder.ts
```diff
@@ -150,6 +152,12 @@ export function defineJoinerConfig(
     schema = toGraphQLSchema([...modelDefinitions.values()])
   }
 
+  if (!idPrefixToEntityName) {
+    idPrefixToEntityName = buildIdPrefixToEntityNameFromDmlObjects([
+      ...modelDefinitions.values(),
```
adrien2p, line 157: Self-review: pass values without spreading

packages/modules/cache/src/types/index.ts
```diff
@@ -0,0 +1,60 @@
...
+
+export type InjectedDependencies = {
+  cachingProviderService: CachingProviderService
+  logger?: Logger
+  [CachingDefaultProvider]: string
+  [Modules.EVENT_BUS]: IEventBusModuleService
+}
+
+export type ModuleInjectedDependencies = InjectedDependencies & {
+  cacheProviderService: CacheProviderService
```
adrien2p: Self-review: merge the two interfaces

packages/modules/cache/src/utils/parser.ts
```diff
@@ -0,0 +1,279 @@
...
+
+  /**
+   * Detect entity type based on object structure and GraphQL type map
+   */
+  private detectEntityType(obj: any, suggestedType?: string): string | null {
+    if (obj.id) {
+      const idParts = obj.id.split("_")
+      if (idParts.length > 1) {
+        const idPrefix = idParts[0]
+        if (idPrefix && this.idPrefixToEntityName[idPrefix]) {
```
adrien2p: Self-review: inline with the above

packages/modules/caching/src/utils/parser.ts
```diff
@@ -0,0 +1,279 @@
...
+  private buildAffectedCacheKeys(
+    entity: EntityReference,
+    relatedEntities: EntityReference[],
+    originalKey: string,
+    operation: "created" | "updated" | "deleted" = "updated"
+  ): string[] {
+    const keys = new Set<string>([originalKey])
+
+    // Add keys based on entity type and ID
+    keys.add(`${entity.type}:${entity.id}`)
```
adrien2p, line 232: Self-review: there are some keys that we could potentially get rid off, ill think about it

packages/modules/cache/src/utils/strategy.ts
```diff
@@ -0,0 +1,158 @@
...
+} from "@medusajs/framework/utils"
+import { type CachingModuleService } from "@services"
+import type { ModuleInjectedDependencies } from "@types"
+import crypto from "crypto"
+import stringify from "fast-json-stable-stringify"
+import { CacheInvalidationParser, EntityReference } from "./parser"
+
+export function objectHash(input: any): string {
+  const str = stringify(input)
+  return crypto.createHash("sha1").update(str).digest("hex")
```
adrien2p: Self-review: evaluate md5

packages/modules/cache/src/utils/strategy.ts
```diff
@@ -0,0 +1,158 @@
...
+
+            const tags_ = await this.computeTags(eventData, {
+              entities: [entityReference],
+              operation,
+            })
+
+            tags.push(...tags_)
+          }
+
+          await this.#cacheModule.clear({
```
adrien2p: Self-review: extract out of the if and simplify event handling

packages/modules/event-bus-local/src/services/event-bus-local.ts
```diff
@@ -91,9 +92,10 @@ export default class LocalEventBusService extends AbstractEventBusModuleService
       const options_ = eventData.options as { delay: number }
       const delay = (ms?: number) => (ms ? setTimeout(ms) : Promise.resolve())
 
-      delay(options_?.delay).then(() =>
+      delay(options_?.delay).then(() => {
         this.eventEmitter_.emit(eventData.name, eventBody)
-      )
+        this.eventEmitter_.emit("*", eventBody)
```
adrien2p: Self-review: add check to validate if star listener exists before emitting. We know we have listeners at this point but not specifically if this one exists

packages/core/utils/src/caching/index.ts
```diff
@@ -0,0 +1,238 @@
...
+          }
+        } else {
+          cacheOptions[key] = options[key]
+        }
+      }
+
+      await Promise.all(promises)
+
+      if (!cacheOptions.enable) {
+        return await originalMethod.apply(this, args)
```
carlos-r-l-rodrigues, line 234: I think this can be moved to line 169

carlos-r-l-rodrigues (approved):
Great first iteration 🔥

packages/core/utils/src/caching/index.ts
adrien2p: I still need to process the enable option which is done line 203

adrien2p: let me know what you think @carlos-r-l-rodrigues

olivermrbl (commented):
A few nits before I noticed before review:
- Should we rename the folder `cache` -> `caching` to reflect the package name? This is inline with the rest of our packages
- Should we do the same for the Redis provider? `caching-redis`

packages/modules/providers/caching-redis/src/services/redis-cache.ts
```diff
@@ -0,0 +1,683 @@
...
+
+  #getTagRefCountKey(): string {
+    return `${this.keyNamePrefix}tag:refs`
+  }
+
+  #getTagReverseDictionaryKey(): string {
+    return `${this.keyNamePrefix}tag:reverse_dict`
+  }
+
+  async #internTags(tags: string[]): Promise<number[]> {
```
olivermrbl, line 60: nit: is this mean to say `internal`?

packages/modules/caching/src/utils/strategy.ts
```diff
@@ -0,0 +1,133 @@
...
+      joinerConfigs
+    )
+
+    const eventBus = this.#container[Modules.EVENT_BUS]
+
+    const handleEvent = async (data: Event) => {
+      try {
+        // We dont have to await anything here and the rest can be done in the background
+        return
+      } finally {
```
olivermrbl, line 53: question: What's the ide with the try/finally here? Haven't seen this usage before

packages/modules/caching/src/types/index.ts
```diff
@@ -0,0 +1,56 @@
...
+  strategy: Constructor<ICachingStrategy>
+  [CachingDefaultProvider]: string
+  [Modules.EVENT_BUS]: IEventBusModuleService
+}
+
+export type CachingModuleOptions = Partial<ModuleServiceInitializeOptions> & {
+  /**
+   * The strategy to be used. Default to the inbuilt default strategy.
+   */
+  strategy?: ICachingStrategy
```
olivermrbl: nit: should we remove this for now?

adrien2p:
> A few nits before I noticed before review:
> 
> * Should we rename the folder `cache` -> `caching` to reflect the package name? This is inline with the rest of our packages
> * Should we do the same for the Redis provider? `caching-redis`

done @olivermrbl

packages/core/core-flows/src/cart/steps/find-one-or-any-region.ts
```diff
@@ -15,39 +16,97 @@ export type FindOneOrAnyRegionStepInput = {
...
+  })
+
+  return await useCache(
+    async () =>
+      service.retrieveRegion(regionId, {
+        relations: ["countries"],
+      }),
+    {
+      container,
+      key:
```
olivermrbl: question: We are verifying that the container exists in the `useCache` function, so why do we need to do it here?

packages/core/core-flows/src/cart/steps/find-one-or-any-region.ts
```diff
@@ -15,39 +16,97 @@ export type FindOneOrAnyRegionStepInput = {
...
+        cacheModule &&
+        (await cacheModule.computeKey([
```
olivermrbl: thought: is there a world where this `key` is not explicitly passed, but generated in `useCache`?

packages/core/core-flows/src/cart/steps/find-one-or-any-region.ts
```diff
@@ -15,39 +16,97 @@ export type FindOneOrAnyRegionStepInput = {
   regionId?: string
 }
 
+async function fetchRegionById(regionId: string, container: MedusaContainer) {
+  const service = container.resolve<IRegionModuleService>(Modules.REGION)
+  const cacheModule = container.resolve(Modules.CACHING, {
+    allowUnregistered: true,
+  })
+
+  return await useCache(
```
olivermrbl: The two questions below are relevant to all `useCache` calls.

packages/core/core-flows/src/cart/steps/get-variant-price-sets.ts
```diff
@@ -75,11 +75,18 @@ async function fetchVariantPriceSets(
...
-      entity: "variant",
-      fields: ["id", "price_set.id"],
-      filters: { id: variantIds },
-    })
+    await query.graph(
+      {
+        entity: "variant",
+        fields: ["id", "price_set.id"],
+        filters: { id: variantIds },
+      },
```
olivermrbl, line 83: question: Does this cache entry require the exact `variantIds` passed to be considered a hit? I.e. not a subset of them

packages/core/framework/src/http/utils/refetch-entities.ts
```diff
@@ -1,49 +1,72 @@
...
 export const refetchEntity = async (
   entryPoint: string,
   idOrFilter: string | object,
   scope: MedusaContainer,
-  fields: string[]
+  fields: string[],
+  options?: RemoteJoinerOptions
 ) => {
-  const [entity] = await refetchEntities(entryPoint, idOrFilter, scope, fields)
+  const { data } = await refetchEntities(
```
olivermrbl: comment: would be nice if this accepted an object instead, so we can avoid the `undefined`s 😄

packages/core/types/src/caching/index.ts
```diff
@@ -0,0 +1,161 @@
...
+  //   tags: string[],
+  //   options: { autoInvalidate?: boolean }
+  // ) => Promise<any>
+
+  // traceClear?: (
+  //   cacheClearFn: () => Promise<any>,
+  //   key: string,
+  //   tags: string[],
+  //   options: { autoInvalidate?: boolean }
+  // ) => Promise<any>
```
olivermrbl, line 25: question: What should we do about these?

packages/core/utils/src/product/get-variant-availability.ts
```diff
@@ -177,11 +184,21 @@ const getDataForComputation = async (
...
+      {
+        entity: "sales_channel_locations",
+        fields: ["stock_location_id"],
+        filters: { sales_channel_id: data.sales_channel_id },
+      },
+      {
+        cache: {
+          tags: [
+            `SalesChannel:${data.sales_channel_id}`,
+            "StockLocation:list:*",
```
olivermrbl, line 197: question: Why do we need an explicit tags definition here? Because it is a link?

packages/modules/caching/CHANGELOG.md
```diff
@@ -0,0 +1 @@
+# @medusajs/cache
```
olivermrbl:
```suggestion
# @medusajs/caching
```

packages/modules/pricing/src/services/pricing-module.ts
```diff
@@ -348,6 +350,26 @@ export default class PricingModuleService
...
+
+      return cachingModule.computeKey({
+        filters,
+        config,
+      })
+    },
+    container: function (this: PricingModuleService) {
+      return createContainerLike(this.container_) as MedusaContainer
+    },
+  })
```
olivermrbl: thought(non-blocking): This looks a bit funky; can we leave out caching from services until we have decided to do that? I think it's better to optimize for very narrow usage of cache initially, so we can control the rollout.

olivermrbl (commented):
Very nice work!

packages/modules/providers/caching-redis/src/services/redis-cache.ts
adrien2p: what do you mean by internal? it is a private method that manages the dictionary

packages/modules/caching/src/utils/strategy.ts
adrien2p: the method return without awaiting the execution. common approach to also do clean up which we use in many places in the code base. also, you cant return and then execute code so this allows it easily

packages/core/core-flows/src/cart/steps/find-one-or-any-region.ts
adrien2p: beause we use it to generate the kye

packages/core/core-flows/src/cart/steps/find-one-or-any-region.ts
adrien2p: useCache does not have access to the input, so I dont see another way. we need to pass something

packages/core/core-flows/src/cart/steps/get-variant-price-sets.ts
adrien2p: yes it does, this will be part of the key generation as well

packages/core/framework/src/http/utils/refetch-entities.ts
adrien2p: agree, I didnt want to create too much breaking changes but if you allow me to do it then lets go ;)

packages/core/types/src/caching/index.ts
adrien2p: they are informative, since we didnt add support for static impl type checking

packages/core/utils/src/product/get-variant-availability.ts
adrien2p: yes, it is a link and we want it to be impacted by any change to the sc or any stock location

packages/modules/pricing/src/services/pricing-module.ts
adrien2p: agree, in that case prices cant be cached as of now because it relies on dynamic context which is not the case of those ones 👍

packages/modules/providers/caching-redis/src/services/redis-cache.ts
olivermrbl: What does intern mean?

packages/core/core-flows/src/cart/steps/find-one-or-any-region.ts
olivermrbl: This is related to my other comment; ideally we don't need to do that :)

packages/modules/providers/caching-redis/src/services/redis-cache.ts
adrien2p: it is where we create ids for tags and assign to the dictionary and reverse dictionary and create the tags refs

packages/core/core-flows/src/cart/steps/find-one-or-any-region.ts
olivermrbl: I think we need to see if we can find a way to make this a better DX. Don't have to be in this PR though

packages/core/core-flows/src/cart/steps/find-one-or-any-region.ts
adrien2p: since the useCache does not have access to any input from the callback, I dont see a solution. it is a light wrapper that prevent you from doing all the work yourself of generating key + get cache + check if hit or not + execute + generate tags + set the cache. here we only ask for the key. The other alternative would be to provide an input options which would be used to build the key, but in that case it makes it more hidden and hard to understand no?

adrien2p: for this particular place it could just use the graph query instead of calling the service, more generally speaking

packages/core/framework/src/http/utils/refetch-entities.ts
olivermrbl: If we are already breaking this function, I say let's do it

packages/core/core-flows/src/cart/steps/find-one-or-any-region.ts
adrien2p: for now I dont have an idea to make a better DX than this one for callback caching :) but if you think about something let me know. in general all tools ask you for the key right?

adrien2p: it also allow for custom key caching for custom data object but thats another topic

packages/core/core-flows/src/cart/steps/find-one-or-any-region.ts
olivermrbl: Let's get this merged and we can think about this separately, would that be OK with you? I don't have any good ideas either right now, but just don't think this current approach is super intuitive from a user PoV.

adrien2p: yes lets think about it and try to find something that can match your expectation, but as an axample nextjs except to receive [key](https://nextjs.org/docs/app/api-reference/config/next-config-js/incrementalCacheHandlerPath#set) here and other thing that we dont ask for because we manage that automatically, same for bento and other tools. It is pretty common to pass a key, it just happen that when using the decorator since we are in control of the method we can intercept the input and build a key automatically, but when it is a callback we cant do that and the other alternative is the user to do everything I do in `useCache` manually or ask for something (either the key or the parts that needs to be used for the key so that we hide the usage of the module entirely) let me know

adrien2p: for example, in the current usage for this comment you could imagine something like:

```ts
// before
await useCache(async () => service.retrieveRegion(...args), {
    container,
    key: await cacheModule.computeKey?.(args),
  })

// after
await useCache(async () => service.retrieveRegion(...args), {
    container,
    key: args,
  })
```

adrien2p: ok, I ve done a small improvement in the meantime, here are the scenario:
- key is a string, we use it as is
- key is not a string, we compute the key internally based on what is provided

for the decorator:
- key is a string, we use it as is
- key is a function we execute it
  - the result is a string, we use it as is
  - the result is something else it will be computed later by the above function
- key is something else, we compute the key internally based on what is provided

olivermrbl (approved):
🔥

packages/core/core-flows/src/cart/steps/find-one-or-any-region.ts
olivermrbl: I think this is already an improvement; nice one!

PR #13446, opened 2025-09-09
Chore(): localised improvement to promotion module

adrien2p:
**What**
- Remove overserialization withing transaction and rely more on internal service or protected method instead which does not serialize and work with Infered entity type
- Slightly rework loop complexity

Overall, this gives a good spare of resources and time spent for serialization


olivermrbl (approved):
LGTM

PR #13450, opened 2025-09-09
chore(): upgrade mikro orm

### LINKED ISSUE #6823: Incorrect count query with filters and OneToOne inverse side population
### Describe the bug

When using a filter in combination with populating a `OneToOne` relation (inverse side) and performing a `count` (or `findAndCount`), the generated query is incorrect, causing the error: `SQLITE_ERROR: no such column: u1.address_id`.

**Generated query:**

```sql
select count(*) as `count` from `login` as `l0`
inner join `user` as `u1` on `l0`.`user_id` = `u1`.`id` and `u1`.`deleted_at` is null and `u1`.`deleted_at` is null
inner join `user` as `u2` on `l0`.`user_id` = `u2`.`id`
left join `address` as `a3` on `u1`.`id` = `a3`.`user_id` and `a3`.`deleted_at` is null
where `l0`.`deleted_at` is null and `u1`.`email` like '%@example.com' and (`u1`.`address_id` is null or `a3`.`id` is not null)
```

It looks like the WHERE condition is generated as if user.address were the owning side of the relation - it doesn’t account for it being the inverse side.
Secondly, the join to the user table is duplicated because of the condition on user.email (but for count it is redundant), but the filter is applied only to the first join (2x the same condition for deleted_at) and not to the second.

### Reproduction

https://github.com/Ruby184/mikroorm-reproduction/tree/bug/v6.5.2-one-to-one-count

### What driver are you using?

@mikro-orm/sqlite

### MikroORM version

6.5.2

### Node.js version

22.18.0

### Operating system

Linux

### Validations

- [x] Read the [Contributing Guidelines](https://github.com/mikro-orm/mikro-orm/blob/master/CONTRIBUTING.md).
- [x] Read the [docs](https://mikro-orm.io/docs).
- [x] Check that there isn't [already an issue](https://github.com/mikro-orm/mikro-orm/issues) that reports the same bug to avoid creating a duplicate.
- [x] Check that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/mikro-orm/mikro-orm/discussions) or join our [Discord](https://discord.gg/w8bjxFHS7X).
- [x] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

adrien2p:
RESOLVES CORE-1205

**What**
- issue fixed: https://github.com/mikro-orm/mikro-orm/issues/6823
- waiting for: https://github.com/mikro-orm/mikro-orm/issues/6851
- this should fix an issue I was also experiencing in regard to entity map propagation https://github.com/mikro-orm/mikro-orm/issues/6855

This pr introduce the mikro orm and pg packages upgrade as well as some improvements


olivermrbl (approved):
epic stuff!

olivermrbl:
feel free to merge as you please

PR #13451, opened 2025-09-09
feat(core-flows,dashboard,js-sdk,promotion,medusa,types,utils): limit promotion usage per customer

fPolic:
**What**
- implement promotion usage limits per customer/email
- fix registering spend usage over the limit
- fix type errors in promotion module tests

**How**
- introduce a new type of campaign budget that can be defined by an attribute such as customer id or email
- add `CampaignBudgetUsage` entity to keep track of the number of uses per attribute value
- update `registerUsage` and `computeActions` in the promotion module to work with the new type
- update `core-flows` to pass context needed for usage calculation to the promotion module

**Breaking**
- registering promotion usage now throws (and cart complete fails) if the budget limit is exceeded or if the cart completion would result in a breached limit

---

CLOSES CORE-1172
CLOSES CORE-1173
CLOSES CORE-1174
CLOSES CORE-1175


packages/modules/promotion/src/services/promotion-module.ts
```diff
@@ -548,6 +716,42 @@ export default class PromotionModuleService
...
+        promotion.campaign?.budget?.type === CampaignBudgetType.USE_BY_ATTRIBUTE
+      ) {
+        const attribute = promotion.campaign?.budget?.attribute!
+        const budgetUsageContext =
+          ComputeActionUtils.getBudgetUsageContextFromComputeActionContext(
+            applicationContext
+          )
+        const attributeValue = budgetUsageContext[attribute]
+
+        if (attributeValue) {
```
fPolic: self-review: should we throw here if the attribute value is not passed when computing actions? For example, a promotion has limited usage per email, and a customer applies the promotion but hasn't entered an email yet.

packages/modules/promotion/src/utils/compute-actions/usage.ts
```diff
@@ -34,3 +58,18 @@ export function computeActionForBudgetExceeded(
...
+  computeActionContext: ComputeActionContext
+): CampaignBudgetUsageContext {
+  return {
+    customer_id:
+      computeActionContext.customer_id ??
+      (computeActionContext.customer as any)?.id ??
+      null,
+    customer_email:
+      computeActionContext.email ??
+      (computeActionContext.customer as any)?.email ??
```
fPolic: tought: in this case, maybe only consider the email from the cart and don't fallback to customer's email

packages/modules/promotion/src/services/promotion-module.ts
```diff
@@ -256,7 +340,7 @@ export default class PromotionModuleService
           campaignBudget.limit &&
           MathBN.gt(newUsedValue, campaignBudget.limit)
         ) {
-          continue
+          campaignBudgetData.used = campaignBudget.limit
```
fPolic: tought: since we are allowing some usage over the limit, should we update this value to reflect exactly how much over the limit was spent instead of capping to the limit (same for `use` type)

packages/core/core-flows/src/promotion/steps/register-usage.ts
```diff
@@ -1,38 +1,53 @@
...
   UsageComputedActions,
 } from "@medusajs/framework/types"
 import { Modules } from "@medusajs/framework/utils"
 import { StepResponse, createStep } from "@medusajs/framework/workflows-sdk"
 
 export const registerUsageStepId = "register-usage"
+
+type RegisterUsageStepInput = {
+  computedActions: UsageComputedActions[]
+  registrationContext: CampaignBudgetUsageContext
```
riqwan, line 13: suggestion: use `cartContext` shape here instead even if the whole cart isn't necessary and then use the attributes like - `customer.email`, `email` or `customer.id`. We had a different context object for product, price list and promotions and it caused confusion and unintended bugs. Unifying them to use cartContext turned out to be quite helpful and uniform across modules.

packages/medusa/src/api/admin/campaigns/validators.ts
```diff
@@ -54,6 +55,16 @@ const CreateCampaignBudget = z
       message: `currency_code should not be present when budget type is ${CampaignBudgetType.USAGE}`,
     }
   )
+  .refine(
+    (data) =>
+      (data.type !== CampaignBudgetType.USE_BY_ATTRIBUTE &&
+        data.type !== CampaignBudgetType.SPEND_BY_ATTRIBUTE) ||
+      isPresent(data.attribute),
```
riqwan: nit:

```suggestion
      isPresent(data.attribute)  || [
        CampaignBudgetType.USE_BY_ATTRIBUTE,
        CampaignBudgetType.SPEND_BY_ATTRIBUTE,
      ].exclude(data.type)
```

packages/modules/promotion/src/services/promotion-module.ts
```diff
@@ -292,11 +376,53 @@ export default class PromotionModuleService
...
+        }
+
+        const attribute = campaignBudget.attribute!
+        const attributeValue = registrationContext[attribute]
+
+        if (!attributeValue) {
+          throw new MedusaError(
+            MedusaError.Types.INVALID_ARGUMENT,
+            `Value for attribute: ${attribute} is required for promotions that are part of a "use by attribute" campaign budget.`
+          )
```
riqwan: tip: iirc, registerUsage was supposed to be a non throwing function, as throwing here would revert the entire complete cart flow.

packages/modules/promotion/src/services/promotion-module.ts
riqwan: I would say yes, its required aspect of the promotion application to succeed, so throw

riqwan (approved):
Added some suggestions. Just from glancing through, lgtm!

packages/core/core-flows/src/cart/workflows/complete-cart.ts
```diff
@@ -321,7 +321,13 @@ export const completeCartWorkflow = createWorkflow(
             })
           }
 
-          return promotionUsage
+          return {
+            computedActions: promotionUsage,
+            registrationContext: {
+              customer_id: cart.customer?.id || null,
+              customer_email: cart.email || null,
+            },
```
olivermrbl, line 346: thought: wonder if we should add a hook to populate the context?

packages/modules/promotion/src/utils/compute-actions/usage.ts
```diff
@@ -34,3 +58,16 @@ export function computeActionForBudgetExceeded(
     }
   }
 }
+
+export function getBudgetUsageContextFromComputeActionContext(
```
olivermrbl, line 62: thought: Using a function like this prevents users from specifying other attributes right?

Can't we just pass these properties down as is? I.e. we pass `customer_email` and `customer_id` in the context instead of doing any transformation in the promotion module.

packages/core/core-flows/src/cart/workflows/complete-cart.ts
fPolic: do we need that if we currently support only 2 specific attributes which we read from the cart

packages/modules/promotion/src/utils/compute-actions/usage.ts
fPolic: we can, this feature was about limiting usage per customer attributes so this is hardcoded

if you think about supporting more generalised attributes, believe we could add a context hook as you suggested above and allow defining an attribute key on Admin when creating an attribute campaign

packages/core/core-flows/src/cart/workflows/complete-cart.ts
olivermrbl: It goes hand in hand with my other comment; I think a hook + flexible attributes would be nice. However, not something we have to add now

olivermrbl (approved):

olivermrbl:
@fPolic is this ready to merge? Noticed you removed the automerge label earlier this week.

fPolic:
@olivermrbl yes it is, there was a test that needed to be updated after `develop` was merged, but it's GTM now

PR #13456, opened 2025-09-10
fix(): Prevent ci to fail if lock file does not exists

adrien2p:
**What**
Prevent CI from failing if the package-lock does not exists when removing it


adrien2p:
@olivermrbl if we merge this one we can get the other pipeline green again (plug I ve reverted the mikro orm deps in the ci-custom)

olivermrbl (approved):

PR #13458, opened 2025-09-10
chore(): Remove extra manager fork

adrien2p:
RESOLVE CORE-1182

**What**
Remove unnecessary extra manager fork that need reconciliation under the hood


willbouch (approved):
LGTM. I would appreciate it though if you could give me (a mere mortal) more context 😅

adrien2p:
> LGTM. I would appreciate it though if you could give me (a mere mortal) more context 😅

So the idea is that currently when a method is wrapped into `@InjectManager` we are creating a fork of the main manager, then when we call transaction (either manually or with the decorator) we again fork the main manager and then within the transaction in mikro orm, the manager is yet forked again. After the transction is flushed, the inner manager entity map from its unit of work will be propagated to the outer context. So as you can see there is many manager fork that can be eliminated entirely. which is what I did here, just removing the intermediate fork and preventing the entity map from being propagated to the main manager that we dont rely on for operations anyway. does that make sense?

PR #13459, opened 2025-09-10
chore(): Improve link indexes

adrien2p:
RESOLVE CORE-1183

**What**
Update link indexes to include deleted_at index constraint


carlos-r-l-rodrigues (approved):

willbouch (approved):
This essentially indexes only non deleted values?

adrien2p:
> This essentially indexes only non deleted values?

yes exactly, and since we have a default filter on deleted_at it should gives us some benefit

PR #13460, opened 2025-09-10
fix(js-sdk): skip null values in query qs.stringify

### LINKED ISSUE #12892: [Bug]: js-sdk client effectively treats `null` query param as empty string(not `"null"` or undefined)
### Package.json file

```JSON
{
  "name": "medusa-starter-default",
  "version": "0.0.1",
  "license": "MIT",
  "keywords": [
    "sqlite",
    "postgres",
    "typescript",
    "ecommerce",
    "headless",
    "medusa"
  ],
  "scripts": {
    "build": "medusa build",
    "start": "medusa start",
    "dev": "medusa develop",
  },
  "dependencies": {
    "@medusajs/admin-sdk": "2.8.5",
    "@medusajs/cli": "2.8.5",
    "@medusajs/dashboard": "2.8.5",
    "@medusajs/framework": "2.8.5",
    "@medusajs/medusa": "2.8.5",
    "@mikro-orm/core": "6.4.3",
    "@mikro-orm/knex": "6.4.3",
    "awilix": "^8.0.1",
    "jwt-decode": "^4.0.0",
    "medusa-plugin-tolgee": "1.4.2",
    "pdfmake": "0.2.18",
    "pg": "^8.13.0",
    "postmark": "^4.0.5"
  },
  "devDependencies": {
    "@medusajs/test-utils": "2.8.5",
    "@mikro-orm/cli": "6.4.3",
    "@swc/core": "1.5.7",
    "@swc/jest": "^0.2.36",
    "@types/jest": "^29.5.13",
    "@types/node": "^20.0.0",
    "@types/pdfmake": "^0.2.10",
    "@types/react": "^18.3.2",
    "jest": "^29.7.0",
    "prop-types": "^15.8.1",
    "react": "^18.2.0",
    "react-dom": "^18.2.0",
    "ts-node": "^10.9.2",
    "typescript": "^5.6.2",
    "vite": "^5.4.14",
    "yalc": "^1.0.0-pre.53"
  },
  "peerDependencies": {
    "@tanstack/react-query": "5.64.2"
  },
  "engines": {
    "node": ">=20"
  },
  "packageManager": "pnpm@10.12.1"
}
```

### Node.js version

v22.13.1

### Database and its version

PostgreSQL 17.2

WSL Ubuntu 24.04

### Browser name

_No response_

### What happended?

I noticed while trying to retrieve only top-level categories(by setting the parent_category_id to null), that the sdk client incorrectly handles query params with literal `null` values.

The call:
```
sdk.client.fetch( "/test", {query: {
  c: null,
  a: 1,
  b: undefined
}} )
```
results in the query: `http:    GET /test?c=&a=1`

undefined is correctly being removed from the query params, but null becomes an empty string.

While this doesn't break the request itself(as in: it doesn't crash), it does break the result. For example with /store/product-categories:  
- `parent_category_id: undefined` or absent->`GET /store/product-categories?fields=...` 
Correctly ignores it and fetches all categories
- `parent_category_id: "null"` ->`GET /store/product-categories?parent_category_id=null&fields=...` 
If passed as a string, it's handled correctly and only the top-level categories are retrieved.
- `parent_category_id: null` ->`GET /store/product-categories?parent_category_id=&fields=...` 
It tries to fetch categories with parents with an empty string as id(empty array, was VERY confusing)

### Expected behavior

literal `null` as a query params in the js-sdk client should be translated to a "null" string, not the empty string

### Actual behavior

literal null is treated as the empty string, with unintended results.

### Link to reproduction repo

N/A

leobenzol:
Closes #12892 

As explained in the issue the qs.stringify method turns `null` values into empty string, so they must be escaped beforehand for the endpoints that expect nullable filter values to work properly

I used a quick and easy JSON.parse/stringify to avoid having to write a recursive function(for $and/$or) and since I saw it used elsewhere already.  Let me know if you'd prefer another approach.

I tested it manually with some combination of direct filter, $and/$or and arrays just to be sure that nothing else broke since there aren't tests for the sdk.


willbouch (commented):
@olivermrbl @carlos-r-l-rodrigues @adrien2p @fPolic. We had this discussion in the issue and I tend to prefer sending "null" over an empty string, but there is also the possibility of removing the property if null (like for undefined). Do you guys have a strong preference for removing rather than "null"?

willbouch:
@SteelRazor47 Talked with the team and could we instead make it so that null and undefined behave in the same way? We could strip both of them out :) Thanks and sorry for the change of heart

leobenzol:
I'm totally fine with both, as long as it isn't that empty string non-sense ahah. Just pushed the change :)

willbouch (approved):

PR #13462, opened 2025-09-10
chore(): order module index and some impl details

adrien2p:
RESOLVES CORE-1184

**What**
Fix indexes and some implementation with overserialization


carlos-r-l-rodrigues (approved):
Nice catch. Thanks.

olivermrbl (approved):

PR #13466, opened 2025-09-10
fix(): product image missing index

adrien2p:
**What**
Add product image imssing index

### DIFF
--- a/packages/modules/product/src/models/product-image.ts
+++ b/packages/modules/product/src/models/product-image.ts
@@ -21,6 +21,18 @@ const ProductImage = model
       unique: false,
       where: "deleted_at IS NULL",
     },
+    {
+      name: "IDX_product_image_rank",
+      on: ["rank"],
+      unique: false,
+      where: "deleted_at IS NULL",
+    },
+    {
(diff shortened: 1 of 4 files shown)


packages/modules/product/src/models/product-image.ts
```diff
@@ -21,6 +21,18 @@ const ProductImage = model
       unique: false,
       where: "deleted_at IS NULL",
     },
+    {
```
olivermrbl, line 24: question: Aren't we missing the `product_ID` index here?

adrien2p: it is a foreign key so it should we good no? though, maybe rank and product id would be nice wdyt?

olivermrbl: You're right

olivermrbl (approved):

packages/modules/product/src/models/product-image.ts
adrien2p: Ive updated the pr

willbouch (approved):
changeset and gtg :)

PR #13468, opened 2025-09-10
feat(core-flows): hook to set shipping opt context when listing

willbouch:
CLOSES CORE-1186

Tested on a local project and it looks like it's working!


olivermrbl:
@willbouch nice stuff!

Should we add a few integration tests, following a similar pattern as [these](https://github.com/medusajs/medusa/blob/368c1d1e1737fb9cd1e60632c8a2c46a2ce34f46/integration-tests/modules/__tests__/cart/store/cart.workflows.spec.ts#L94-L117)?

In addition to this, we might need to add this hook to the `listShippingOptionsForCartWithPricingWorkflow` too. 

And finally, if these two workflows are called as sub-workflows, e.g. when adding shipping method to cart, we might need to drill down `additional_data` if we aren't already.

Let me know if this makes sense?

willbouch:
> @willbouch nice stuff!
> 
> Should we add a few integration tests, following a similar pattern as [these](https://github.com/medusajs/medusa/blob/368c1d1e1737fb9cd1e60632c8a2c46a2ce34f46/integration-tests/modules/__tests__/cart/store/cart.workflows.spec.ts#L94-L117)?
> 
> In addition to this, we might need to add this hook to the `listShippingOptionsForCartWithPricingWorkflow` too.
> 
> And finally, if these two workflows are called as sub-workflows, e.g. when adding shipping method to cart, we might need to drill down `additional_data` if we aren't already.
> 
> Let me know if this makes sense?

Yes! I was actually looking for how to test these so thanks!

packages/core/core-flows/src/cart/steps/validate-cart-shipping-options.ts
```diff
@@ -31,6 +31,11 @@ export interface ValidateCartShippingOptionsStepInput {
    * The IDs of the shipping options to validate.
    */
   option_ids: string[]
+  /**
+   * Pre-fetched shipping options. If provided, validation will be done against these
+   * instead of querying the database.
+   */
+  shippingOptions?: Array<{ id: string }>
```
willbouch: To avoid breaking changes, I basically changed that step so that it can take pre-fetched shipping options. If they are defined, we skip the refetching of the shipping options (which is not ideal) and simply run the validation on the ones that we received in arguments

packages/core/core-flows/src/cart/workflows/add-shipping-method-to-cart.ts
```diff
@@ -100,21 +101,21 @@ export const addShippingMethodToCartWorkflow = createWorkflow(
...
       listShippingOptionsForCartWithPricingWorkflow.runAsStep({
         input: {
           options: input.options,
           cart_id: cart.id,
           is_return: false,
+          additional_data: input.additional_data,
         },
       })
 
+    validateCartShippingOptionsStep({
```
willbouch, line 114: Moved the validation to after in order to reuse the already fetched shipping options. The only thing the validation does is checking that the shipping options we want to add to the cart can actually be queried (they fit the defined rules so we don't end up adding a shipping option to the cart that we shouldn't). I don't see why the validation could not happen after the list step

willbouch:
@olivermrbl it should be better now

willbouch:
@shahednasser I guess there will also be doc changes in here. Let me know how you want to handle this

packages/core/core-flows/src/cart/steps/validate-cart-shipping-options.ts
willbouch: Replaced with `{ id: string }[]`

packages/core/core-flows/src/cart/steps/validate-cart-shipping-options.ts
```diff
@@ -31,6 +27,11 @@ export interface ValidateCartShippingOptionsStepInput {
    * The IDs of the shipping options to validate.
    */
   option_ids: string[]
+  /**
+   * Pre-fetched shipping options. If provided, validation will be done against these
+   * instead of querying the database.
+   */
+  shippingOptions?: { id: string }[]
```
olivermrbl: two nits here(non-blocking): 
- should we call this `prefetchedShippingOptions` to make its intent even more obvious? 
- it seems we are mixing casing conventions; should we try to align? we might need to deprecate `option_ids` in favor of `optionIds`

packages/core/core-flows/src/cart/steps/validate-cart-shipping-options.ts
```diff
@@ -51,32 +52,48 @@ export const validateCartShippingOptionsStepId =
 export const validateCartShippingOptionsStep = createStep(
   validateCartShippingOptionsStepId,
   async (data: ValidateCartShippingOptionsStepInput, { container }) => {
-    const { option_ids: optionIds = [], cart, shippingOptionsContext } = data
+    const { option_ids: optionIds = [], cart, shippingOptionsContext, shippingOptions } = data
 
     if (!optionIds.length) {
```
olivermrbl, line 57: question: shouldn't we check if `shippingOptions` is not present too? otherwise, I believe we return early in the case of only passing prefetched options

olivermrbl (approved):
Overall LGTM!

packages/core/core-flows/src/cart/steps/validate-cart-shipping-options.ts
willbouch: Good ideas though. Added it

willbouch: Seems in other steps that the "convention" is snake case. What can I do with context? Leave as is? I cant rename it since it would probably be a breaking change :/

packages/core/core-flows/src/cart/steps/validate-cart-shipping-options.ts
willbouch: I dont think so. In this step, we are validating that the ids in `option_ids` are essentially shipping options that are "usable" based on the context. So, if the user passed prefetched options, we use them to run the validation, otherwise we fetch using the context. So, it kind of makes sense to return early on optionIds being empty. It means we are validating nothing essentially

fPolic (approved):
LGTM!

packages/core/core-flows/src/cart/steps/validate-cart-shipping-options.ts
olivermrbl: yeah I see, we could deprecate it, but i don't feel strongly about doing it now

fine to move forward with this

packages/medusa/src/api/store/carts/[id]/line-items/[line_id]/route.ts
```diff
@@ -32,14 +33,18 @@ export const POST = async (
 }
 
 export const DELETE = async (
-  req: MedusaRequest,
+  req: MedusaRequest<AdditionalData>,
```
Comment (willbouch) [inline packages/medusa/src/api/store/carts/[id]/line-items/[line_id]/route.ts]: @olivermrbl this feels like Cursor is right. What do you think we could do here?

Comment (olivermrbl) [reply packages/medusa/src/api/store/carts/[id]/line-items/[line_id]/route.ts]: Hmm, we probably can't do much here but leave it out. I do agree, payloads are not passed in DELETEs.

Comment (willbouch) [reply packages/medusa/src/api/store/carts/[id]/line-items/[line_id]/route.ts]: Ok, I removed that part @olivermrbl

PR #13469, opened 2025-09-10
chore(medusa): add metadata field to AdminCreateInvite

v0eak:
Adds the possiblity to pass metadata fields when creating invites through API route


willbouch:
> Adds the possiblity to pass metadata fields when creating invites through API route

Hi and thanks for the contribution! May I ask you why? Maybe just give a few examples to better understand the needs behind this?

v0eak:
Well, yeah, currently the admin dashboard does not allow for passing metadata fields, but I have made some customizations to it, so that we can pass "permissions" and "3rd_party_id".

For our use-case its necessary, because we want admins to log into the dashboard via a 3rd party instead of the default email/password and we need a way to make sure that the person trying to log into the admin dashboard is either an admin user or was invited, else we won't authenticate them (our 3rd party provider does not provide us an email associated with the login-attempt).
And we also pass a permissions field so that an Admin User can be assigned roles, which is to limit their access to certain resources (e.g. a user with "support" permissions is
 only supposed to be able view/edit support tickets, but not have the ability to perform changes to products).
 
 image

willbouch (commented):
Could I just ask you to create a changeset? Will get this merged after that!

willbouch (approved):

PR #13477, opened 2025-09-11
fix(): update cart line item route fetching

adrien2p:
**What**
Remove cart fetching before workflow which fetch the cart as well


carlos-r-l-rodrigues (approved):

olivermrbl (approved):
LGTM!

PR #13482, opened 2025-09-11
fix(dashboard): german translation issues

appinteractive:
I fixed those bad translations in German, as they were nearly unbearable. 🙈


carlos-r-l-rodrigues (approved):

willbouch (approved):
Could you please provide a changeset?

PR #13497, opened 2025-09-13
fix(utils,core-flows): subtotal calculation and returns location

carlos-r-l-rodrigues:
What:
 - Line item totals - base subtotal, tax_total, and total on the current quantity (excluding returned/dismissed).
 - Refundable total considering returns request and returns received.
 - Return location validation - Add validation to claim/exchange “request item return” workflows to ensure items can be returned at the specified location_id
 - Optionally accepts location_id to return the items on claims and exchanges, if not provided it picks the first location of the variant stock.

FIXES: CORE-1143


.changeset/rare-colts-warn.md
```diff
@@ -0,0 +1,6 @@
+---
```
fPolic, line 1: nit: this file needs updating

integration-tests/http/__tests__/exchanges/exchanges.spec.ts
```diff
@@ -239,7 +266,7 @@ medusaIntegrationTestRunner({
 
       fulfillmentSet = (
         await api.post(
-          `/admin/fulfillment-sets/${location.fulfillment_sets[0].id}/service-zones`,
+          `/admin/fulfillment-sets/${location.fulfillment_sets[0].id}/service-zones?fields=*location`,
```
fPolic: q: why do we need `location` here

packages/medusa/src/api/admin/exchanges/validators.ts
```diff
@@ -127,6 +127,7 @@ export type AdminPostExchangesAddItemsReqSchemaType = z.infer<
 >
 
 export const AdminPostExchangesReturnRequestItemsReqSchema = z.object({
+  location_id: z.string().optional(),
```
fPolic, line 130: discussion: still a shipping option from another location can be selected for return fulifllment.
Should this be the other way around i.e. when an option is selected, we check if all the items can be returned there.

packages/core/utils/src/totals/line-item/index.ts
```diff
@@ -148,33 +167,66 @@ function getLineItemTotals(
...
+    original_subtotal: new BigNumber(subtotal),
     total: new BigNumber(
-      MathBN.sum(MathBN.sub(subtotal, discountsSubtotal), taxTotal)
+      MathBN.sum(
+        MathBN.sub(currentSubtotal, currentDiscountsSubtotal),
+        taxTotal
+      )
     ),
 
     original_total: new BigNumber(
```
fPolic, line 200: discussion: should `original_total` contain returned and new items?

fPolic (commented):
Nice work overall!
Tested on an example and subtotal/total looks expected but noticed two issues that might be unrelated to the PR changes but have to investigate more 1) new items are not allocated and 2) pending difference is off and out of sync with payment collection.

packages/medusa/src/api/admin/exchanges/validators.ts
carlos-r-l-rodrigues: This one is optional, I imagine if you have 1 item in 3 different warehouses, it could be selected to be returned in any of them. But if this is empty, we'll pick the first available stock location.

packages/core/utils/src/totals/line-item/index.ts
carlos-r-l-rodrigues: Idk 😆 
Is original total, just the total before taxes?

integration-tests/http/__tests__/exchanges/exchanges.spec.ts
carlos-r-l-rodrigues: we don't need it.

packages/medusa/src/api/admin/exchanges/validators.ts
fPolic: Yeah but what if for return shipping, the user selects a SO for a different location?  Then `location_id` on the return and the location of the return shipping don't match. 
Alternatively, we can add a check when creating a return shipping method that it matches the `location_id` if it is set.

packages/core/utils/src/totals/line-item/index.ts
fPolic: and discount but should this value represent the total before modifications, so it should not contain exchanged and edited items?

integration-tests/http/__tests__/claims/claims.spec.ts
```diff
@@ -742,7 +742,7 @@ medusaIntegrationTestRunner({
           expect(paymentCollections[0]).toEqual(
```
olivermrbl: question: Could I get you to add a few sentences explaining why the totals changed here? I presume it is due to the changes, but it would be good to understand exactly which of them corrected this

packages/core/core-flows/src/order/steps/add-order-transaction.ts
```diff
@@ -31,24 +31,43 @@ export const addOrderTransactionStep = createStep(
       return new StepResponse(null)
     }
 
+    const existingQuery: any[] = []
```
olivermrbl, line 34: comment: Could I get you to explain the change in this file? It is not immediately obvious how this relates to the fix

packages/core/core-flows/src/order/workflows/exchange/exchange-request-item-return.ts
```diff
@@ -204,6 +234,51 @@ export const orderExchangeRequestItemReturnWorkflow = createWorkflow(
       status: [OrderChangeStatus.PENDING, OrderChangeStatus.REQUESTED],
     })
 
+    const pickItemLocationId = transform(
```
olivermrbl, line 207: suggestion(non-blocking): Seems like this could be extracted to a reusable step

packages/core/core-flows/src/order/workflows/exchange/exchange-request-item-return.ts
```diff
@@ -106,6 +111,42 @@ export const exchangeRequestItemReturnValidationStep = createStep(
     throwIfIsCancelled(orderReturn, "Return")
     throwIfOrderChangeIsNotActive({ orderChange })
     throwIfItemsDoesNotExistsInOrder({ order, inputItems: items })
+
+    const inputItemIds = new Set(items.map((i) => i.id))
```
olivermrbl: suggestion(non-blocking): Seems like this could be extracted to a reusable step

packages/core/utils/src/totals/__tests__/totals.ts
```diff
@@ -836,8 +836,8 @@ describe("Total calculation", function () {
           tax_lines: [
```
olivermrbl, line 858: question: Same as other comment; would be good to understand why these tests changed so drastically

integration-tests/http/__tests__/claims/claims.spec.ts
carlos-r-l-rodrigues: Sure, the main changes that reflect on the totals are these:

 - Pending difference = current order value − (captured payments − refunds).
 - Returned items reduce the current order value after they are received or dismissed.

Before, returned items were still composing or partially composing the order total, and the pending difference was trying to discount return received items, causing even more discrepancy.

packages/core/core-flows/src/order/steps/add-order-transaction.ts
carlos-r-l-rodrigues: Here it is just avoiding multiple database calls in a loop, and composing 1 single query.

carlos-r-l-rodrigues: plus, if an array with multiple refunds is received, it discard the ones that are already present. But keep the rest.
Before it was stopping if any item that already exist was there.

fPolic (approved):
Tested and works as expected!

olivermrbl:
I noticed a few inconsistencies:

**Pending difference on returns is incorrect:**

CleanShot 2025-09-17 at 09 11 16@2x


**Pending different on exchanges is incorrect:**

CleanShot 2025-09-17 at 09 12 59@2x

Both of these lead to incorrect outstanding amounts when the return/exchange is confirmed, and we eventually end up charging or refunding the customer the wrong amount.

packages/core/utils/src/totals/adjustment/index.ts
```diff
@@ -40,9 +42,23 @@ export function calculateAdjustmentTotal({
     adj["total"] = new BigNumber(adjustmentsTotal)
   }
 
+  const quantity = item?.quantity || MathBN.convert(1)
+  const adjustmentPerItem = MathBN.eq(quantity, 0)
```
willbouch: Nit: might be better a if here rather than 3 ternaries with the same check

willbouch (approved):
Code looks good

packages/core/utils/src/totals/adjustment/index.ts
carlos-r-l-rodrigues: fair enough

PR #13508, opened 2025-09-15
fix(core-flows, medusa): don't allow negative line item quantity

fPolic:
CLOSES CORE-1194


willbouch (approved):

packages/medusa/src/api/store/carts/validators.ts
```diff
@@ -65,15 +65,15 @@ export const StoreCalculateCartTaxes = createSelectParams()
...
+  quantity: z.number().gt(0),
   metadata: z.record(z.unknown()).nullish(),
 })
 
 export type StoreUpdateCartLineItemType = z.infer<
   typeof StoreUpdateCartLineItem
 >
 export const StoreUpdateCartLineItem = z.object({
-  quantity: z.number(),
+  quantity: z.number().gt(0),
```
olivermrbl: thought(non-blocking): I believe it is fairly common to remove an item when the quantity is set to 0 rather than throwing an error. For example, consider a cart with a quantity picker that lets you decrease the quantity all the way to 0; the expected behavior in most cases would be to remove the item.

We might want to consider taking this approach instead of throwing on 0. Curious to hear your thoughts?

fPolic: Yup sounds good to add! Will push an update!

packages/core/core-flows/src/cart/workflows/update-line-item-in-cart.ts
```diff
@@ -217,72 +241,80 @@ export const updateLineItemInCartWorkflow = createWorkflow(
...
+        }
+      )
 
-    updateLineItemsStepWithSelector(lineItemUpdate)
+      updateLineItemsStepWithSelector(lineItemUpdate)
 
-    refreshCartItemsWorkflow.runAsStep({
-      input: { cart_id: input.cart_id },
-    })
+      refreshCartItemsWorkflow.runAsStep({
```
olivermrbl, line 305: question: Do we still need to update promotions when items are removed? E.g. after removing an item, you might not qualify for a Buy 1, Get 2 promotion.

packages/core/core-flows/src/cart/workflows/update-line-item-in-cart.ts
```diff
@@ -217,72 +241,80 @@ export const updateLineItemInCartWorkflow = createWorkflow(
...
         data: { id: input.cart_id },
-      }),
-      releaseLockStep({
-        key: input.cart_id,
-        skipOnSubWorkflow: true,
       })
-    )
+    })
+
+    releaseLockStep({
```
olivermrbl: question: Don't we want to still emit the event in case the item is removed?

olivermrbl (commented):
Overall LGTM, few questions

packages/core/core-flows/src/cart/workflows/update-line-item-in-cart.ts
fPolic: items are refreshed as part of the delete workflow as well

packages/core/core-flows/src/cart/workflows/update-line-item-in-cart.ts
fPolic: Yes, added the event back. I presumed the delete would emit a separate event.

olivermrbl (approved):
LGTM!

PR #13509, opened 2025-09-15
fix(types): cart item totals optional

willbouch:
Totals are not returned by default, so they should be marked as optional


shahednasser (approved):
LGTM

adrien2p (approved):
LGTM 🎉

PR #13527, opened 2025-09-16
fix(core-flows,types,medusa): pass /store/shipping-options fields to workflow

leobenzol:
cc @willbouch since you asked to be tagged if I opened this PR.

The /store/shipping-options is one of only 2 store endpoints to not allow custom `fields` to be passed(other one is /store/returns if you guys want to add it to the backlog). This PR fixes that so that custom linked models can also be retrieved.

Note: This fix reveals a bug in the next.js starter

https://github.com/medusajs/nextjs-starter-medusa/blob/eac359cc8d52f1d33d8712e18e4e0f38940afb3e/src/lib/data/fulfillment.ts#L23-L24

`fields` was previously ignored, now it errors since it tries to parse the misspelled "fulfllment"(the i is missing). It worked before since both fields are already defined by default inside the workflow. So the `fields` line is totally redundant and should be removed(ideally before merging this).   
Maybe in the next release notes it should also warn users to remove it for those that already have a modified copy of the starter.


willbouch:
Hey @leobenzol. Opened a PR [here](https://github.com/medusajs/nextjs-starter-medusa/pull/516) for the issue you described. Good catch. As for the rest, it looks fine. For the release notes, let me check on my end

willbouch (approved):

willbouch:
@leobenzol PR was merged. Are you good with me merging this one?

leobenzol:
@willbouch yep go ahead 👍

PR #13537, opened 2025-09-17
fix(auth-emailpass): better handle identity with same email error

willbouch:
**Testing**: I tested locally. So, I basically create an invite link and before clicking the button on the page, I updated the node modules in my local project to throw an error on the first line in the `admin/invites/accept` route, mocking an error. If I restart my local project after removing that mock error, I can successfully accept the invite, whereas this was not possible before

CLOSES CORE-1199


packages/modules/providers/auth-emailpass/src/services/emailpass.ts
```diff
@@ -185,17 +168,30 @@ export class EmailPassAuthService extends AbstractAuthModuleProvider {
     }
 
     try {
-      await authIdentityService.retrieve({
+      const identity = await authIdentityService.retrieve({
         entity_id: email,
       })
 
+      if (!identity.app_metadata) {
```
olivermrbl: suggestion(non-blocking): Should we add a short comment as to what and why we are doing this? The assumption is that when there is no `app_metadata`, the auth identity is still "claimable"

packages/core/utils/src/common/is-object-empty.ts
```diff
@@ -0,0 +1,5 @@
+import { isObject } from "./is-object"
```
carlos-r-l-rodrigues: I believe we don't need another util for it. We already have isPresent

carlos-r-l-rodrigues (approved):

packages/core/utils/src/common/is-object-empty.ts
willbouch: I was so surprised this didn't exist... Thanks man! Replaced

packages/modules/providers/auth-emailpass/src/services/emailpass.ts
```diff
@@ -185,17 +173,31 @@ export class EmailPassAuthService extends AbstractAuthModuleProvider {
     }
 
     try {
-      await authIdentityService.retrieve({
+      const identity = await authIdentityService.retrieve({
         entity_id: email,
       })
 
+      // If app_metadata is not defined or empty, it means no actor was assigned to the auth_identity yet (still "claimable")
+      if (!identity.app_metadata || isPresent(identity.app_metadata)) {
```
olivermrbl: nit: think `isPresent` already checks for an undefined value

olivermrbl (approved):
LGTM

PR #13538, opened 2025-09-18
feat(): Add support for link filterable relationship

adrien2p:
RESOLVES CORE-1207

**What**
Make it possible to specify internal module relation of first level at least from the link filterable options making it possible to filter by these and get them indexed as well


adrien2p:
not stale

PR #13540, opened 2025-09-18
fix(): Prevent promotion filtering to exceed psql limits

adrien2p:
**What**

When the context can be big, we switch to a simple query to retrieve the available rule attribute in db to stip out the attribute from the context from which we dont care


olivermrbl (approved):
LGTM

PR #13541, opened 2025-09-18
Always enforce is_active and not is_internal

nelo-jdliauw:


packages/medusa/src/api/store/products/middlewares.ts
```diff
@@ -93,6 +93,7 @@ export const storeProductRoutesMiddlewares: MiddlewareRoute[] = [
       }),
       applyDefaultFilters({
         status: ProductStatus.PUBLISHED,
+        categories: { is_internal: false, is_active: true } // Always apply this filter
```
nelo-jdliauw, line 96: The individual product route makes a refetch call to the DB, and these filters should propagate

PR #13549, opened 2025-09-19
chore(medusa): remove empty object type for set cart's customer route

shahednasser:


willbouch (approved):
changeset and :shipit:

PR #13174, opened 2025-08-09
feat(admin): add id filter to collections endpoint

anupam-pandey01:
## Summary
Added support for filtering collections by `id` in the admin API.  
This aligns the collections endpoint with the product categories endpoint, which already supports the `id` parameter.

## Changes
- Updated `AdminGetCollectionsParams` in `validators.ts` to include `id` as an optional query parameter.
- Allows filtering by single `id` or multiple `id` values.

## Why
Consistency across admin endpoints and easier API usage when retrieving specific collections.

## Related Issue
Fixes #13147

PR #13555, opened 2025-09-19
feat(store): add id filtering to store collections endpoint

bqst:
Same as these PRs : #13174 #13495 

Added support for filtering collections by id in the store API.
This aligns the collections endpoint with the product categories/tags/types endpoint, which already supports the id parameter.

- ✅ Add id parameter to StoreGetCollectionsParamsFields
- ✅ Support single ID or multiple IDs filtering
- ✅ Align with other store endpoints (/store/product-categories, /store/product-tags, etc.)

### DIFF
--- a/packages/core/types/src/http/collection/store/queries.ts
+++ b/packages/core/types/src/http/collection/store/queries.ts
@@ -1,5 +1,8 @@
-import { BaseCollectionListParams } from "../common"
+import { BaseFilterable } from "../../../dal"
+import { BaseCollectionListParams, BaseCollectionParams } from "../common"
 
-export interface StoreCollectionFilters
-  extends Omit<BaseCollectionListParams, "id"> {
-}
+export interface StoreCollectionListParams
+  extends BaseCollectionListParams,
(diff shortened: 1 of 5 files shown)


packages/core/types/src/http/collection/store/queries.ts
```diff
@@ -1,5 +1,8 @@
-import { BaseCollectionListParams } from "../common"
+import { BaseFilterable } from "../../../dal"
+import { BaseCollectionListParams, BaseCollectionParams } from "../common"
 
-export interface StoreCollectionFilters
-  extends Omit<BaseCollectionListParams, "id"> {
-}
+export interface StoreCollectionListParams
+  extends BaseCollectionListParams,
+    BaseFilterable<StoreCollectionListParams> {}
```
willbouch: Why the BAseFiltertable here?

willbouch:
Thanks for the contribution! Left a question which is related to Cursor's comment. Also, could you please add a changeset? Thanks

willbouch (approved):

PR #13556, opened 2025-09-19
Chore/promo improve further

adrien2p:
RESOLVES CORE-1203

**What**
Even further improve pre promotions filtering by checking that any unsatisfiable rules accross promotions and its application method (target/buy) results in the promotion being excluded


packages/modules/promotion/src/utils/compute-actions/build-promotion-rule-query-filter-from-context.ts
```diff
@@ -188,18 +185,56 @@ export async function buildPromotionRuleQueryFilterFromContext(
...
+      )
+      SELECT rc.id
+      FROM rule_counts rc
+      WHERE rc.id NOT IN (SELECT id FROM promotions_with_unsatisfiable_rules)
     )
   `.trim()
 
   return {
-    [raw((alias) => notExistsSubquery(alias))]: true,
+    [raw((alias) => cteSubquery(alias))]: true,
```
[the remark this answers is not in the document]
adrien2p: still on it 😂

packages/modules/promotion/src/utils/compute-actions/build-promotion-rule-query-filter-from-context.ts
```diff
@@ -188,18 +185,69 @@ export async function buildPromotionRuleQueryFilterFromContext(
...
-      WHERE ppr.promotion_id = ${alias}.id
-      AND (${joinedConditions})
+      SELECT 1 FROM (
+        -- Promotions with rules for attributes not in context
+        SELECT ppr.promotion_id
+        FROM promotion_promotion_rule ppr
+        JOIN promotion_rule pr ON ppr.promotion_rule_id = pr.id
+        WHERE pr.attribute NOT IN (${attributeKeys})
+
+        UNION ALL
```
carlos-r-l-rodrigues: here (all the ones in the block) it can be UNION too

carlos-r-l-rodrigues (approved):

packages/modules/promotion/src/utils/compute-actions/build-promotion-rule-query-filter-from-context.ts
adrien2p: True, I need to test it but I imagine it should not decrease performance and worst case improve them as it will dedupe agree?

olivermrbl (approved):
LGTM

PR #13558, opened 2025-09-19
fix(types): pluralize settings

willbouch:


adrien2p (approved):
LGTM, we discussed a potential solution that would entirely remove the type inference and rely on the exact generation instead to create the type. we can discuss that later 👍

PR #13571, opened 2025-09-22
feat(dashboard): refactor location list UI to use data table

fPolic:
CLOSES AI-75


willbouch:
Testing result:
- The search field loses focus after the search call is sent
- We cannot search by address (which would be nice perhaps but it might be complicated)

Error in toast
CleanShot 2025-09-23 at 08 16 56@2x

fPolic:
> The search field loses focus after the search call is sent

Seems to be caused by the table rerenders when `isLoading` changes. We can tackle this separately since it's not directly related to the PR changes.

> We cannot search by address

Address is a relation of the location, so not sure if we can filter locations by address.

willbouch:
> > The search field loses focus after the search call is sent
> 
> Seems to be caused by the table rerenders when `isLoading` changes. We can tackle this separately since it's not directly related to the PR changes.
> 
> > We cannot search by address
> 
> Address is a relation of the location, so not sure if we can filter locations by address.

Ok sounds good. It is weird though because I use the Datatable in my PR too and the bug does not seem to happen

willbouch (approved):
LGTM

olivermrbl (approved):

olivermrbl:
> > > The search field loses focus after the search call is sent
> > 
> > 
> > Seems to be caused by the table rerenders when `isLoading` changes. We can tackle this separately since it's not directly related to the PR changes.
> > > We cannot search by address
> > 
> > 
> > Address is a relation of the location, so not sure if we can filter locations by address.
> 
> Ok sounds good. It is weird though because I use the Datatable in my PR too and the bug does not seem to happen

Relations are usually searchable if they use the `searchable()` modifier on the data model definition. I think this is missing on the address of the stock location. We can perhaps tackle this separately?

fPolic:
> It is weird though because I use the Datatable in my PR too and the bug does not seem to happen

hmm, in that case let me investigate this before merging, will also include `searchable` for address

olivermrbl (approved):
Nice

PR #13578, opened 2025-09-23
fix(dashboard, order): set return status on cancel

fPolic:
**What**
- set order return status when cancelled
- display errors in toast when order timeline cancel actions fail

---

CLOSES CORE-1200

### DIFF
--- a/packages/admin/dashboard/src/routes/orders/order-detail/components/order-activity-section/order-timeline.tsx
+++ b/packages/admin/dashboard/src/routes/orders/order-detail/components/order-activity-section/order-timeline.tsx
@@ -1,4 +1,4 @@
-import { Button, Text, Tooltip, clx, usePrompt } from "@medusajs/ui"
+import { Button, Text, Tooltip, clx, toast, usePrompt } from "@medusajs/ui"
 import { Collapsible as RadixCollapsible } from "radix-ui"
 
 import { PropsWithChildren, ReactNode, useMemo, useState } from "react"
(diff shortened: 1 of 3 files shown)


packages/admin/dashboard/src/routes/orders/order-detail/components/order-activity-section/order-timeline.tsx
```diff
@@ -785,7 +785,9 @@ const ReturnBody = ({
       return
     }
 
-    await cancelReturnRequest()
+    await cancelReturnRequest().catch((error) => {
+      toast.error(error.message)
```
willbouch, line 789: I see we do that somewhere, and other places, we translate it. Should we here?

packages/modules/order/src/services/actions/cancel-return.ts
```diff
@@ -90,6 +94,7 @@ export async function cancelReturn(
         {
           data: {
             canceled_at: new Date(),
+            status: ReturnStatus.CANCELED,
```
willbouch, line 97: Interesting 🤔 there was no usage of Canceled at all

willbouch (approved):

packages/admin/dashboard/src/routes/orders/order-detail/components/order-activity-section/order-timeline.tsx
fPolic: You mean like an additional translated label? Think it's fine to pass an error from the server here, indicating something went wrong.

packages/modules/order/src/services/actions/cancel-return.ts
fPolic: we relied on `canceled_at` in some places on the dashboard

olivermrbl (approved):

PR #13580, opened 2025-09-23
chore(utils): make upsert with replace more efficient

adrien2p:
PARTIALLY RESOLVES CORE-1156

**What**
Improve upsertWithReplace to batch as much as possible what can be batched. Performance of this method will be much greater specially for cases with maybe entities and batch (e.g we seen many cases where they bulk product with hundreds variants and options etc)
for example let take the following object:
- entity 1
  - entity 2 []
    - entity 3 []
  - entity 2 []
    - entity 3 []

here all entity 3 will be batched and all entity 2 will be batched

I ve also added a pretty detail test that check all the stage and what is batched or not with many comments so that it is less harder to consume and remember in the future

Also includes:
- mikro orm upgade (issues found and fixes)
- order module hooks fixes

**NOTE**
It was easier for now to do this instead of rewriting the different areas where it is being used, also, maybe it means that we will have closer performance to what we would expect to have natively

**NOTE 2**
Also fix the fact that integration tests of the core packages never ran 😂


packages/modules/order/src/services/order-module-service.ts
```diff
@@ -161,30 +167,25 @@ const generateMethodForModels = {
...
   }
   OnInit()(MikroORMEntity.prototype, "onInit_OrderShipping")
   BeforeCreate()(MikroORMEntity.prototype, "onInit_OrderShipping")
 }
 {
   const MikroORMEntity = toMikroORMEntity(OrderItem)
   MikroORMEntity.prototype["onInit_OrderItem"] = function () {
-    this.version ??= this.order?.version ?? null
+    if (this.order) {
+      this.version ??= this.order?.version ?? null
```
willbouch: this.order? becomes useless no?

packages/modules/order/src/services/order-module-service.ts
```diff
@@ -145,14 +145,20 @@ const generateMethodForModels = {
...
+      this.exchange_id ??= this.order_change.exchange_id ?? null
 
-    if (!this.claim_id && !this.exchange_id) {
-      this.return_id = this.return?.id ?? this.order_change?.return_id ?? null
+      if (
+        !this.claim_id &&
+        !this.exchange_id &&
+        (this.return || this.order_change)
+      ) {
+        this.return_id = this.return?.id ?? this.order_change?.return_id ?? null
```
willbouch: this.order_change? can be replaced by this.order_change no?

willbouch (approved):
Lots of changes, but it looked good to me

packages/modules/order/src/services/order-module-service.ts
adrien2p: Don’t think so, if return is present but not order_change it would break. Since we expect one or the other or both I prefer playing safe

adrien2p:
@olivermrbl shall we merge this one as well?

olivermrbl:
@adrien2p, yes sounds good to me!

olivermrbl (approved):

PR #13581, opened 2025-09-23
feat(js-sdk): allow passing a query parameter to deleteLineItem

shahednasser:


packages/core/js-sdk/src/store/index.ts
```diff
@@ -698,12 +698,14 @@ export class Store {
...
       cartId: string,
       lineItemId: string,
+      query?: SelectParams,
       headers?: ClientHeaders
     ) => {
       return this.client.fetch<HttpTypes.StoreLineItemDeleteResponse>(
         `/store/carts/${cartId}/line-items/${lineItemId}`,
         {
           method: "DELETE",
+          query,
```
willbouch, line 708: It is for the refetch after deleting?

willbouch (approved):

packages/core/js-sdk/src/store/index.ts
shahednasser: It's to retrieve fields in the returned cart

PR #13587, opened 2025-09-23
feat(dashboard): select refund reason in refund form

willbouch:
CLOSES CORE-1211

This PR quickly allows the user to select a refund a reason when creating a refund

CleanShot 2025-09-23 at 14 36 39@2x

CleanShot 2025-09-23 at 14 36 26@2x


packages/admin/dashboard/src/routes/orders/order-create-refund/components/create-refund-form/create-refund-form.tsx
```diff
@@ -35,11 +35,13 @@ const CreateRefundSchema = zod.object({
     float: zod.number().or(zod.null()),
   }),
   note: zod.string().optional(),
+  refund_reason_id: zod.string().optional(),
```
willbouch, line 38: @olivermrbl should this be optional?

olivermrbl: I think it's fine to make this optional; it won't be needed for regular returns, and for ad-hoc refunds, we can leave it up to the merchant to decide.

olivermrbl (approved):
Good stuff!

PR #13589, opened 2025-09-24
fix(dashboard): improve inventory level location management

fPolic:
**What**
- add InfiniteList on location selection for inventory level -> fixes issue with location pagination
- fix removal of location level for an inventory item
- refresh the levels table when locations are updated
- add search input for filtering locations

---

CLOSES CORE-1208

### DIFF
--- a/.changeset/spicy-timers-breathe.md
+++ b/.changeset/spicy-timers-breathe.md
@@ -1,0 +1,5 @@
+---
+"@medusajs/dashboard": patch
+---
+
+fix(dashboard): improve inventory level location management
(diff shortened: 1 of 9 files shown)


willbouch:
Hey, Idk if it is related, but while testing, I noticed there is no toaster after deleting an inventory item. Also, the table needs a refresh

https://github.com/user-attachments/assets/f1478875-5c9b-424b-9515-93b2da8c6d19

The fix to refresh the table when adding one works fine, but maybe we should also do it for remove?

fPolic:
yup good catch, pushed a fix for both

willbouch (approved):
:shipit:

PR #13592, opened 2025-09-25
fix(dashboard): payment providers select

fPolic:
**What**
- use lazy loading for payment providers select on region create/edit

### DIFF
--- a/packages/admin/dashboard/src/components/inputs/combobox/combobox.tsx
+++ b/packages/admin/dashboard/src/components/inputs/combobox/combobox.tsx
@@ -58,6 +58,7 @@ interface ComboboxProps<T extends Value = Value>
   onCreateOption?: (value: string) => void
   noResultsPlaceholder?: ReactNode
   allowClear?: boolean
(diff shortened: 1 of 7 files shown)


packages/admin/dashboard/src/components/inputs/combobox/combobox.tsx
```diff
@@ -152,10 +154,15 @@ const ComboboxImpl = <T extends Value = string>(
       return []
     }
 
+    // do not use mather if input is hidden and no search value can be entered
```
willbouch: mather? You mean matcher?

willbouch (approved):
Tested locally ✅  Created 22 and they all show up

packages/admin/dashboard/src/components/inputs/combobox/combobox.tsx
fPolic:
```suggestion
    // do not use `matcher` if the input is hidden
```

PR #13594, opened 2025-09-25
fix(core-flows): skip locking by default on subworkflows

carlos-r-l-rodrigues:
What: invert the flag to execute acquire and release lock on subworkflows.

### DIFF
--- a/.changeset/cold-impalas-boil.md
+++ b/.changeset/cold-impalas-boil.md
@@ -1,0 +1,5 @@
+---
+"@medusajs/core-flows": patch
+---
+
+fix(core-flows): skip locking on sub workflows by default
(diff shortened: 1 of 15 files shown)


adrien2p (approved):

willbouch (commented):
I guess the PR name should be "stop skip locking"?

willbouch (approved):

carlos-r-l-rodrigues:
> I guess the PR name should be "stop skip locking"?

Why? here the behavior didn't change, it just skips locking by default when running as a subworkflow. To force it respect the lockings, it requires an explicit flag `executeOnSubWorkflow: true`

willbouch:
> > I guess the PR name should be "stop skip locking"?
> 
> Why? here the behavior didn't change, it just skips locking by default when running as a subworkflow. To force it respect the lockings, it requires an explicit flag `executeOnSubWorkflow: true`

Yeah all good. My bad 😵

PR #13597, opened 2025-09-25
feat(medusa): include user_metadata in auth routes jwt

### LINKED ISSUE #13584: Third party auth registration ambiguity
### What Medusa version and documentation are you using?

v2

### Preliminary Checks

- [x] This issue is not a duplicate. Before opening a new issue, please search existing issues: https://github.com/medusajs/medusa/issues

### Issue Summary

Hello,

If you follow the third party auth storefront integration guide, when you reach the point of creating the customer, the docs harcode a example@medusajs.com email, to be sent to the /store/customers route.

This causes ambiguity for the user to determine where to get the actual email from. This problem is accentuated by the fact that the returned token from the validateCallback route, doesn't include the provider_identity representing the google login for that auth_identity, and thus, doesn't include the user_metadata field, which contains the data returned by the provider. **In the case of Google, the email you need to register the customer**

I think there are two solutions, one note exclusive for the other

1. More immediately, remove the harcoded example@medusajs.com email from the /store/customers payload and add a comment instructing where to get this email from, in this code snippet https://docs.medusajs.com/resources/storefront-development/customers/third-party-login#full-code-example-for-third-party-login-callback-page

When solving 1. I think it will become apparent that there is no straightforward way to instruct where to get the email, as one would expect it to be available in the returned token, with no additional call, as the backend has it at this point, it just doesn't return it. Telling the user to put an additional email input is redundant (a user loging in to google would find this unnecessary, as they have logged in with their email already); creating a custom api endpoint to resolve the auth_identity allong with its providers and then filtering the one matching google, seems also redundant.

3. IMO the correct approach would be to modify generateJwtTokenForAuthIdentity method, to include a stripped down version of the provider identities linked to the auth identity, inlcuding: id, provider, user_metadata

What do you think? I could create a PR for 3. provided you think it is correct. I don't think we would be exposing any sensitive information here

- [x] I would like to fix this issue

NicolasGorga:
**What**

Return provider specific user_metadata in jwt token returned by auth/[actor_type]/[auth_provider] routes.

**Why**

When implementing thrid party auth flows, to create the Medusa entity representing the actor type, you will need information obtained from the callback after the user consents. For google for example, you need the email, which so far is stored in the provider user_metadata property but is never returned in the http response.

This causes one to circumvent around this issue by creating an additional http route which gets the auth_identity from the auth_context object, to access the providers array and then look for the specific element matching the current auth_provider route.

**How**

Pass the auth_provider obtained from the route path param and pass it down to `generateJwtTokenForAuthIdentity` and inside it, find the provider matching the auth provider passed from route handler inside the auth identity `provider_identities` property. If found, assign its `user_metadata` value to the property of the same name in the payload of the token to be generated.

Now the user has the user_metadata property included in the obtained token and can use it's information to create the corresponding Medusa entity for the actor type in context. In the case of google and customer actor type, they would use `user_metadata.email` to call the create customer api endpoint.

**Testing**

Haven't found anywhere in the codebase where the http layer of /auth functionality is tested, but tested locally with debugger and test storefront to guarantee the user_metadata is returned as expected, all working correctly.

Fixes #13584 
CLOSES CORE-437 

---

> [!NOTE]
> Propagates auth provider to JWT generation and embeds provider-specific user_metadata in tokens returned by auth routes.
> 
> - **Auth JWT Payload**:
>   - `generateJwtTokenForAuthIdentity` now accepts optional `authProvider` and injects matching provider `user_metadata` into JWT payload.
> - **Auth Routes** (`auth/[actor_type]/[auth_provider]/*`):
>   - Pass `authProvider` from route params to JWT generator in `authenticate`, `register`, and `callback` handlers.
> - **Release**:
>   - Changeset adds patch for `@medusajs/medusa`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 4833faf84533ce7438bc37c15bd572f991988e69. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


willbouch:
Thanks for the contribution! This was actually on my TODO list so you gave me a hand there :). will take a look when I get the chance. In the meantime, can you add a changeset? The package you are changing is `medusa`. In other words, you API layer. I updated your PR desc to reflect that. Also, maybe I would consider this a `feat` rather than a `fix` ? I don't think it is necessarily a bug since it was just not supported

willbouch:
Hi @NicolasGorga. Implementation looks good :) did you check [here](https://github.com/medusajs/medusa/blob/95d282e8ef8f2185737d493dea8b6f1677684543/integration-tests/http/__tests__/auth/admin/auth.spec.ts#L23) for tests? There is probably a way to add something there to test the additional metadata being passed in the token

NicolasGorga:
Hey @willbouch changeset added.

Oh i didn't see that when looking the place to add tests. Let me create some tests for it and make a new commit. Given that the implementation is green lit, i'll go ahead and open a new PR for the doc change around the storefront third party auth guide!

NicolasGorga:
@willbouch regarding the tests, i have the following doubt. 

Since the emailpass, that is the one that all the tests reference doesn't have user_metadata, how do you suggest i go about this? Do i create a fixture for a new auth provider just for the tests that implements AbstractAuthModuleProvider and has user_metadata? 

Or what would be your suggestion?

willbouch:
> @willbouch regarding the tests, i have the following doubt.
> 
> Since the emailpass, that is the one that all the tests reference doesn't have user_metadata, how do you suggest i go about this? Do i create a fixture for a new auth provider just for the tests that implements AbstractAuthModuleProvider and has user_metadata?
> 
> Or what would be your suggestion?

To be honest, the test is not really necessary. It would basically just test that the property is being passed down, which is essentially testing that typescript works haha. I'm good without it

willbouch (approved):

willbouch:
@olivermrbl are you okay with me merging this one like this?

NicolasGorga:
Hey @willbouch yeah thought it would be kind of redundant, so happy if you are. All my tests showed the field returned correctly and with the expected shape to the caller :)

willbouch:
Quick heads up @NicolasGorga I have't forgotten. I am pushing for another review from the team

NicolasGorga:
@willbouch was about to ping you awesome, i'll keep an eye for it

willbouch:
@NicolasGorga big launch coming up, so the team is a bit busy with this. I'll have them check asap so we can get that nice addition merged

NicolasGorga:
Hey @willbouch seeing a lot of movement, exciting!

 If i could get you to also check https://github.com/medusajs/medusa/pull/13633  would be good to fix that bug, haven't been noticed since there were missing tests

olivermrbl (approved):
LGTM

PR #13598, opened 2025-09-25
feat(core-flows): support ad hoc returns

willbouch:
CLOSES CORE-1213


packages/core/core-flows/src/payment/workflows/refund-payment.ts
```diff
@@ -1,89 +1,10 @@
...
- *     // other order details...
- *   },
- *   payment: {
- *     id: "payment_123",
- *     // other payment details...
- *   },
- *   amount: 10
- * })
- */
-export const validateRefundStep = createStep(
```
willbouch, line 61: Essentially, the validation needs to disappear. It made sure that we could not refund more than pending difference, which needs to stop happening if we want to allow the merchant to refund the customer for "good will" for instance

packages/core/core-flows/src/payment/workflows/refund-payment.ts
```diff
@@ -101,6 +22,14 @@ export type RefundPaymentWorkflowInput = {
    * The amount to refund. If not provided, the full payment amount will be refunded.
    */
   amount?: BigNumberInput
+  /**
+   * The note to attach to the refund.
+   */
+  note?: string
```
willbouch, line 28: those were missing

packages/core/core-flows/src/payment/workflows/refund-payment.ts
```diff
@@ -156,7 +85,30 @@ export const refundPaymentWorkflow = createWorkflow(
       list: false,
     }).config({ name: "order" })
 
-    validateRefundStep({ order, payment, amount: input.amount })
+    const creditLineAmount = transform({ order, payment, input }, ({ order, payment, input }) => {
```
willbouch, line 102: The logic here is quite simple. Basically, we need to create credit lines if we end up refunding more than what is missing on the outstanding balance. So there are 2 scenarios:
1. The outstanding balance is >=0, which means we need to create a credit line for the entire refund amount
2. The outstanding balance is negative (there are 2 cases here)
	a. the refund amount covers the outstanding amount exactly -> no credit line
	b. the refund amount covers more than outstanding amount. Then we need to add credit line for the extra amount

packages/modules/payment/src/services/payment-module.ts
```diff
@@ -863,7 +855,7 @@ export default class PaymentModuleService
         amount: data.amount,
         created_by: data.created_by,
         note: data.note,
-        refund_reason_id: data.refund_reason_id,
+        refund_reason_id: data.refund_reason_id || null,
```
willbouch: Something weird is going on. This is just a temporary fix to start a discussion on the PR. There is a mikro orm error when we send an empty string here (which makes sense). We can either make sure at the API level that empty strings become null, but that will only put a bandaid on the problem. We can also somehow add something that applies to everything in the core where if the property is like `*_id` and an empty string, than we set to null. However, I feel like we might overlook scenarios and break something... Other ideas?

packages/core/core-flows/src/payment/workflows/refund-payment.ts
olivermrbl: I wonder if we should figure out a way to make this operation an _either/or_ to ensure the created refund matches either the credit line amount or the value of the returned items.

For example, let's imagine the merchant has returned one item of $100 on a $200 order. Now, the pending difference is $100. However, in addition to the return, the merchant wants to issue an additional $20 refund because the customer had a shitty shipping experience.

With the current implementation, the merchant would likely issue a $120 refund and attach a "Shipping issues" refund reason. Now, the created refund on the order doesn't really reflect the reality; only $20 of the $120 was actually related to "Shipping issues". Would it be more correct to issue two refunds here? One for the returned item and one for the shipping issues?

This is nitpicky, and we can also leave it up to the merchant to handle, but curious to hear your thoughts?

willbouch: I get what you are saying and I agree. It is not easy to decide how the merchant will use it though. I coded it like this to avoid throwing but we could also do that. In my opinion, the merchant should always issue 2 refunds. One for the actual product and then one for the shitty shipping experience. But, I guess we could also argue that it is a bit annoying for merchants and maybe we should let them just create one single refund for the 2 things they are trying to handle

packages/modules/payment/src/services/payment-module.ts
olivermrbl: Hmm, I would expect the client to send the correct payload; in this case, we should make sure we never pass anything but a refund reason ID. I feel like trying to detect the correct ID doesn't really solve the problem either, as you could also accidentally send a non-empty string that doesn't reference a refund reason ID.

packages/core/core-flows/src/payment/workflows/refund-payment.ts
olivermrbl: > But, I guess we could also argue that it is a bit annoying for merchants and maybe we should let them just create one single refund for the 2 things they are trying to handle

I would be OK with this approach initially. After all, I think we are covering an edge case, since these actions would most likely happen at different points in time, i.e. the likelyhood of scenario happening is small.

olivermrbl (approved):

packages/modules/payment/src/services/payment-module.ts
olivermrbl: One alternative would be to validate the passed ID (in this case, the empty string) references a correct refund reason. We would have a preliminary step retrieving the refund reason by ID and throwing if it doesn't exist.

packages/core/core-flows/src/payment/workflows/refund-payment.ts
willbouch: Yeah fair enough. My goal here was mostly to avoid throwing an error and the merchant being confused

packages/modules/payment/src/services/payment-module.ts
willbouch: So, you mean that the client of the API (the admin dashboard), should not return an empty string as refund_reason_id when there is none selected by the user?

olivermrbl: Yes, I would strip it or pass `null`.

packages/core/core-flows/src/payment/workflows/refund-payment.ts
```diff
@@ -156,7 +85,30 @@ export const refundPaymentWorkflow = createWorkflow(
...
+      return amountToRefund
+    })
+
+    when(
+      { creditLineAmount }, ({ creditLineAmount }) => MathBN.gt(creditLineAmount, 0)
+    ).then(() => {
+      createOrderRefundCreditLinesWorkflow.runAsStep({
+        input: {
+          order_id: order.id,
+          amount: creditLineAmount,
```
olivermrbl, line 121: Can we store the reference to the refund reason on the credit line? I believe we have `reference` and `reference_id` properties. This reference would be used in an accounting context when you want to know why a certain set of credit lines was created on an order, e.g. "we want to know many refunds we issue related to shipping issues".

For refund reasons specifically, we ideally store a code or label (something fixed), such that the credit lines are easy to categorise and filter by. Maybe we should consider adding a `code` on the refund reason, what do you think?

packages/admin/dashboard/src/routes/orders/order-create-refund/components/create-refund-form/create-refund-form.tsx
```diff
@@ -63,8 +63,6 @@ export const CreateRefundForm = ({ order }: CreateRefundFormProps) => {
         value: paymentAmount.toFixed(currency.decimal_digits),
         float: paymentAmount,
       },
-      note: "",
```
willbouch, line 66: We were also saving empty strings as note all the time... Figured we could fix it to have null in the DB

packages/core/core-flows/src/payment/workflows/refund-payment.ts
willbouch: As discussed 1-1, will do that :)

willbouch:
@olivermrbl I'd like your opinion on how to display which payment we are creating a refund for. I removed the dropdown, but I am very unsure about what to display on top to make it clear. Do you have an idea? Or maybe we should ask Lubvig?
CleanShot 2025-09-29 at 12 07 18@2x

Also, this is what it looks like for credit lines. If you dont add a refund reason, the default is the payment collection. Not sure if you want to remove this? Or keep as in my screenshot?
CleanShot 2025-09-29 at 12 08 17@2x

olivermrbl (approved):
LGTM!

PR #13623, opened 2025-09-29
feat(core-flows,product,types): scoped variant images

fPolic:
## **What**

This PR introduces the ability to associate product images with specific product variants, enabling more granular image management for products with multiple variants.

 **Scoped Image Management**: Images can now be associated with specific variants instead of being global to the product
**Bulk Operations**: Batch add/remove variant associations for efficient management

---

_**API Changes**_
- `POST /admin/products/:id/variants/:variant_id/images/batch` - Batch manage image associations for a variant
- `POST /admin/products/:id/images/:image_id/variants/batch` - Batch manage variant associations for an image

_**Product Module Changes**_

1. Data model:
- **ProductVariant** model now includes `images` many-to-many relationship
- **ProductImage** model now includes `variants` many-to-many relationship  
- **ProductVariantProductImage** pivot entity for managing associations
- **ProductVariant** now has `thumbnail`

2. Service methods:
- `addImageToVariant()` - Associate images with specific variants
- `removeImageFromVariant()` - Remove image-variant associations
- `getVariantImages()` - Retrieve images for variants - _(general product images + variant specific images)_

_**Core Flows Changes**_
- variant thumbnail will be used for the line item thumbnail if exists

---

CLOSES CORE-1214
CLOSES CORE-1215
CLOSES CORE-1216

---

**Related PRs**: 
- https://github.com/medusajs/medusa/pull/13670
- https://github.com/medusajs/medusa/pull/13731


packages/modules/product/src/services/product-module-service.ts
```diff
@@ -2188,4 +2195,187 @@ export default class ProductModuleService
...
+      )
+      variant.images = variantImages.get(id) || []
+    }
+
+    return this.baseRepository_.serialize(variant)
+  }
+
+  @InjectManager()
+  async addImageToVariant(
+    data: { image_id: string; variant_id: string }[],
```
olivermrbl: nit: should we create a type?

packages/modules/product/src/services/product-module-service.ts
```diff
@@ -2188,4 +2195,187 @@ export default class ProductModuleService
...
+      data,
+      sharedContext
+    )
+
+    return productVariantProductImage as { id: string }[]
+  }
+
+  @InjectTransactionManager()
+  protected async addImageToVariant_(
+    data: { image_id: string; variant_id: string }[],
```
olivermrbl: nit: should we create a type?

olivermrbl: Same comment for other methods

packages/modules/product/src/services/product-module-service.ts
```diff
@@ -2188,4 +2195,187 @@ export default class ProductModuleService
...
+        }
+
+        return img.variants.some((v) => v.id === variantId)
+      })
+
+      result.set(
+        variantId,
+        variantImages as InferEntityType<typeof ProductImage>[]
+      )
+    }
```
olivermrbl: suggestion: We are looping through all product images for every single variant; I believe we can make this more efficient. Here's a suggestion from claude, which I think looks pretty good:
```ts
const variantIdArray = variants.map((v) => v.id)

if (variantIdArray.length === 0) {
  return new Map()
}

const variantProductIdMap = new Map(
  variants.map((v) => [v.id, v.product_id])
)

// Group variant IDs by product ID for efficient lookup
const productVariantsMap = new Map<string, string[]>()
for (const variant of variants) {
  const existing = productVariantsMap.get(variant.product_id) || []
  existing.push(variant.id)
  productVariantsMap.set(variant.product_id, existing)
}

const allProductImages = (await this.listProductImages(
  { product_id: Array.from(variantProductIdMap.values()) },
  {
    relations: ["variants"],
  },
  context
)) as (ProductTypes.ProductImageDTO & {
  product_id: string
  variants: InferEntityType<typeof ProductVariant>[]
})[]

// Initialize result map with empty arrays
const result = new Map<string, InferEntityType<typeof ProductImage>[]>()
for (const variantId of variantIdArray) {
  result.set(variantId, [])
}

// Create a Set for faster variant ID lookups
const variantIdSet = new Set(variantIdArray)

for (const img of allProductImages) {
  if (!img.variants.length) {
    // Image with no variants - add to all variants of this product
    const productVariantIds = productVariantsMap.get(img.product_id) || []
    for (const variantId of productVariantIds) {
      result.get(variantId)!.push(img as InferEntityType<typeof ProductImage>)
    }
  } else {
    // Image with specific variants - add to those variants
    for (const variant of img.variants) {
      if (variantIdSet.has(variant.id)) {
        result.get(variant.id)!.push(img as InferEntityType<typeof ProductImage>)
      }
    }
  }
}

return result
```

packages/core/core-flows/src/product/workflows/batch-image-variants.ts
```diff
@@ -0,0 +1,163 @@
...
+      }
+    })
+
+    const res = parallelize(
+      addImageToVariantsStep(normalizedInput),
+      removeImageFromVariantsStep(normalizedInput)
+    )
+
+    const updateData = when(
+      "shoud-variants-removed",
```
olivermrbl: nit: typo
```suggestion
      "should-remove-variants",
```

packages/core/core-flows/src/product/workflows/batch-image-variants.ts
```diff
@@ -0,0 +1,163 @@
...
+            },
+          }
+        }
+      )
+
+      return updateData
+    })
+
+    when(
+      "shoud-update-variant",
```
olivermrbl: nit: typo
```suggestion
      "should-update-variants",
```

olivermrbl (approved):
Overall LGTM, nice work

packages/core/core-flows/src/product/steps/add-image-to-variants.ts
```diff
@@ -0,0 +1,58 @@
...
+/**
+ * This step adds an image to one or more product variants.
+ *
+ * @example
+ * const data = addImageToVariantsStep({
+ *   image_id: "img_123",
+ *   add: ["variant_123", "variant_456"]
+ * })
+ */
+export const addImageToVariantsStep = createStep(
```
willbouch, line 16: Could we not make one step for both here? We have a step that adds one image to multiple variants and one that adds multiple images to one variant. Why not merge the 2 steps? We could pass a list of images and a list of variants and do it all in one go no?

willbouch: Or am I missing something? I feel like the remove scenario is similar too

fPolic: hmm, and what would behaviour be if we pass `{images_ids: [i1,i2,i3], variant_ids: [v1,v2,v3]}`? We add each image to each variant and delete combinations of image x variant for remove?

think this is one of those cases where a bit of duplication is ok for simpler, specific api, but no strong opinion

willbouch: Yes, this is what I would imagine. No strong opinion either. I believe the customer/customer group duplicated. So leave it as is its fine:)

willbouch (approved):

PR #13626, opened 2025-09-30
fix(): Pricing preference context loss

adrien2p:
**What**
The context reference is being mutated by the repository leading to an empty context. Also, the filter is built from the pricing context instead of pricing context -> context leading to always fetch all preferences all the time

### DIFF
--- a/packages/modules/pricing/src/services/pricing-module.ts
+++ b/packages/modules/pricing/src/services/pricing-module.ts
@@ -9,6 +9,7 @@ import {
   FindConfig,
   InferEntityType,
   InternalModuleDeclaration,
+  MedusaContainer,
   ModuleJoinerConfig,
   ModulesSdkTypes,
   PricePreferenceDTO,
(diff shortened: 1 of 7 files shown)


packages/modules/pricing/src/services/pricing-module.ts
```diff
@@ -395,29 +454,37 @@ export default class PricingModuleService
     )
 
     // We use the price rules to get the right preferences for the price
-    const priceRulesForPrices = await this.priceRuleService_.list(
+    const priceRulesForPrices = await this.listPriceRules(
```
carlos-r-l-rodrigues, line 457: just confirming, are we not using `super.listPriceRules` here because it will be decorated with the cache later?

carlos-r-l-rodrigues (approved):

packages/modules/pricing/src/services/pricing-module.ts
adrien2p: not only, but also to be sure that if is called as part of a transaction you dont get the intermediate serialization for nothing

willbouch (approved):

PR #13634, opened 2025-09-30
feat(admin-bundler): forward env vars to plugin admin extensions

### LINKED ISSUE #11815: [Bug]: cannot expose env vars from backend to plugin admin
### Package.json file

```JSON
{
  "name": "clean-backend",
  "version": "0.0.1",
  "license": "MIT",
  "keywords": [
    "sqlite",
    "postgres",
    "typescript",
    "ecommerce",
    "headless",
    "medusa"
  ],
  "scripts": {
    "build": "medusa build",
    "start": "medusa start",
    "dev": "medusa develop",
  },
  "dependencies": {
    "@medusajs/admin-sdk": "2.6.2-snapshot-20250310153842",
    "@medusajs/cli": "2.6.2-snapshot-20250310153842",
    "@medusajs/framework": "2.6.2-snapshot-20250310153842",
    "@medusajs/icons": "2.6.2-snapshot-20250310153842",
    "@medusajs/medusa": "2.6.2-snapshot-20250310153842",
    "@medusajs/ui": "4.0.8-snapshot-20250310153842",
    "@mikro-orm/core": "6.4.3",
    "@mikro-orm/knex": "6.4.3",
    "awilix": "^8.0.1",
    "medusa-plugin-tolgee": "0.0.3",
    "pg": "^8.13.0"
  },
  "devDependencies": {
    "@medusajs/test-utils": "2.6.2-snapshot-20250310153842",
    "@mikro-orm/cli": "6.4.3",
    "@swc/core": "1.5.7",
    "@swc/jest": "^0.2.36",
    "@types/jest": "^29.5.13",
    "@types/node": "^20.0.0",
    "@types/react": "^18.3.2",
    "jest": "^29.7.0",
    "prop-types": "^15.8.1",
    "react": "^18.2.0",
    "react-dom": "^18.2.0",
    "ts-node": "^10.9.2",
    "typescript": "^5.6.2",
    "vite": "^5.2.11",
    "yalc": "^1.0.0-pre.53"
  },
  "engines": {
    "node": ">=20"
  },
  "packageManager": "pnpm@9.1.2"
}
```

### Node.js version

v22.13.1

### Database and its version

PostgreSQL 17.2

W11

### Browser name

_No response_

### What happended?

First of all great work on #11720. I've been experimenting with the latest snapshot from that PR(`2.6.2-snapshot-20250310153842`) and had almost no problems, except the following:

It is not possible to use import.meta.env to reference env variables inside the main backend inside plugins when running `plugin:publish`. Becuase the variables are supposed to be set by the main backend and missing in the plugin, they are all hard-coded to `undefined` during the bundling. 
This doesn't happen with `plugin:develop`(makes sense).
Since process.env is not allowed by vite, the only solution I found was to expose the required env vars via an admin api route, taking them from the plugin options passed to modules.

Is this the correct approach? Adding API keys to an endpoint's result feels like a sin, but in the end since it's an admin route, an attacker would need the auth token, at which point he could just as easily access the dashboard and see the API keys embedded with (in theory) import.meta.env.

### Expected behavior

There should be a way to access the env variables of the parent project from the plugin admin routes

### Actual behavior

Only way to access any kind of variable in plugin admin routes is via API Routes.

### Link to reproduction repo

https://github.com/SteelRazor47/medusa-plugin-tolgee commit e58cdf with import.meta.env, commit 5601f89 with API route

leobenzol:
Closes #11815

As explained in the issue, the only way to expose env vars to vite in plugins is via the vite-config "define" object, this has to be done by end-users with `admin: vite: () => ({ define: { ... }}) ` in medusa-config.ts
This PR adds a convention that automatically does that for env vars in the form `PLUGIN_<var>`

Note: ~~plugin authors should be careful to check if the variable is actually *declared* with `typeof NAME !== "undefined`, otherwise it could throw a ReferenceError.~~
I was able to create a "fake" process.env that is always defined, so if the env var is mispelled/not supplied by end-users it is simply undefined instead of throwing. There is this weird distinction that one must use import.meta.env only in the actual backend and process.env only inside plugins, but I don't think we can do better.

EXAMPLE:
If you have a plugin and you want to use an env var "TEST"

End-users: to the .env file add PLUGIN_TEST="1234"

Plugin authors: simply use "process.env.TEST" in any widget or extension route

### DIFF
--- a/packages/admin/admin-bundler/src/utils/config.ts
+++ b/packages/admin/admin-bundler/src/utils/config.ts
@@ -76,11 +76,13 @@ export async function getViteConfig(
       .replace(/^@/, "")
       .replace(/[^a-zA-Z0-9]/g, "_")
       .toUpperCase()
+    const pluginEnv: Record<string, string | undefined> = {}
     for (const envVar in process.env) {
       if (envVar.startsWith(`PLUGIN_${normalizedName}_`)) {
-        baseConfig.define![envVar] = JSON.stringify(process.env[envVar]);
+        pluginEnv[envVar] = process.env[envVar];
       }
     }
+    baseConfig.define!["process.env"] = JSON.stringify(pluginEnv);
   }
 
   if (options.vite) {
--- a/packages/admin/admin-bundler/src/utils/config.ts
+++ b/packages/admin/admin-bundler/src/utils/config.ts
@@ -70,6 +70,19 @@ export async function getViteConfig(
     ],
   }
 
+  // Inject plugin environment variables as vite defines
+  for (const plugin of options.plugins ?? []) {
+    const normalizedName = plugin.replace(/\/admin$/, "")
+      .replace(/^@/, "")
+      .replace(/[^a-zA-Z0-9]/g, "_")
+      .toUpperCase()
+    for (const envVar in process.env) {
+      if (envVar.startsWith(`PLUGIN_${normalizedName}_`)) {
+        baseConfig.define![envVar] = JSON.stringify(process.env[envVar]);
+      }
+    }
+  }
+
   if (options.vite) {
     const customConfig = options.vite(baseConfig)
     return mergeConfig(baseConfig, customConfig)
--- a/packages/admin/admin-bundler/src/utils/config.ts
+++ b/packages/admin/admin-bundler/src/utils/config.ts
@@ -71,12 +71,12 @@ export async function getViteConfig(
   }
 
   // Inject plugin environment variables with vite define
-  const pluginEnv = Object.fromEntries(
-    Object.entries(process.env).filter(([key]) =>
-      key.startsWith("PLUGIN_")
-    )
-  )
-  pluginEnv["BACKEND_URL"] = backendUrl
+  const pluginEnv: Record<string, string | undefined> = { BACKEND_URL: backendUrl }
+  for (const [key, value] of Object.entries(process.env)) {
+    if (key.startsWith("PLUGIN_")) {
+      pluginEnv[key.replace(/^PLUGIN_/, "")] = value
+    }
+  }
   baseConfig.define!["process.env"] = JSON.stringify(pluginEnv)
 
   if (options.vite) {
--- a/packages/admin/admin-bundler/src/utils/config.ts
+++ b/packages/admin/admin-bundler/src/utils/config.ts
@@ -70,20 +70,14 @@ export async function getViteConfig(
     ],
   }
 
-  // Inject plugin environment variables as vite defines
-  for (const plugin of options.plugins ?? []) {
-    const normalizedName = plugin.replace(/\/admin$/, "")
-      .replace(/^@/, "")
-      .replace(/[^a-zA-Z0-9]/g, "_")
-      .toUpperCase()
-    const pluginEnv: Record<string, string | undefined> = {}
-    for (const envVar in process.env) {
-      if (envVar.startsWith(`PLUGIN_${normalizedName}_`)) {
-        pluginEnv[envVar] = process.env[envVar];
-      }
-    }
-    baseConfig.define!["process.env"] = JSON.stringify(pluginEnv);
-  }
+  // Inject plugin environment variables with vite define
+  const pluginEnv = Object.fromEntries(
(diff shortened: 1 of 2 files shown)


willbouch (approved):
I like it!  @shahednasser does that make sense to you for documentation?

shahednasser:
Thank you for your contribution @leobenzol ! Can you add to the PR description a small example of setting the env var in `medusa-config.ts` then using it in a plugin's widget? Would help us document it later.

leobenzol:
@willbouch 2 more thoughts:

- i think it'd be nice to include `__BACKEND_URL__` as `process.env.BACKEND_URL` by default. That way if you want to create a js-sdk instance you don't have to create a vite-env.d.ts(to make ts happy) and you could just use process.env.BACKEND_URL. it makes it a little simpler to use the sdk.

- inside the plugins, should the `PLUGIN_<name>` prefix be stripped? So plugin authors can just use process.env.TEST instead of process.env.PLUGIN_MY_MEDUSA_PLUGIN_TEST for example.
I actually prefer the long name, there's less "magic" that way and users can look in the code to see the exact var name they need, but lmk what you think.

packages/admin/admin-bundler/src/utils/config.ts
```diff
@@ -70,6 +70,21 @@ export async function getViteConfig(
...
+      .toUpperCase()
+    const pluginEnv: Record<string, string | undefined> = {}
+    for (const envVar in process.env) {
+      if (envVar.startsWith(`PLUGIN_${normalizedName}_`)) {
+        pluginEnv[envVar] = process.env[envVar];
+      }
+    }
+    baseConfig.define!["process.env"] = JSON.stringify(pluginEnv);
+  }
+
```
adrien2p: suggestion: Here it seems that for every plugin loop we are overriding the `process.env` in define. Also, I propose a little change to this, let me know what you think
```suggestion
const pluginEnv = Object.entries(process.env).filter(([key]) =>
    key.startsWith("PLUGIN_")
  )
  const pluginEnvMap: Record<string, string | undefined> = {}
  if (pluginEnv.length) {
    for (const plugin of options.plugins ?? []) {
      const normalizedName = plugin
        .replace(/\/admin$/, "")
        .replace(/^@/, "")
        .replace(/[^a-zA-Z0-9]/g, "_")
        .toUpperCase()
      for (const [env, value] of pluginEnv) {
        if (env.startsWith(`PLUGIN_${normalizedName}_`)) {
          pluginEnvMap[env] = value
        }
      }
    }
    baseConfig.define!["process.env"] = JSON.stringify(pluginEnvMap)
  }
```

leobenzol: Good catch, I was lazy and only tested it with one, don't know how I didn't see that just by looking at it tho 

This got me thinking: for some reason in my head I thought each plugin had its own process.env, since the are actually all shared in a single environment, should we get rid of "normalizedName" and just use "PLUGIN_" as a prefix? It would make it simpler to use especially for end users.

adrien2p: I think that could be a nice approach, wdyt @willbouch

willbouch: Yep! Sounds good

leobenzol:
should be good to go now, i took the liberty to also add BACKEND_URL as i suggested before

packages/admin/admin-bundler/src/utils/config.ts
```diff
@@ -70,6 +70,15 @@ export async function getViteConfig(
     ],
   }
 
+  // Inject plugin environment variables with vite define
+  const pluginEnv = Object.fromEntries(
```
willbouch: Will that strip `PLUGIN_`? Like you suggested, it would probably be better that for PLUGIN_TEST, the dev inside the plugin can use process.env.TEST

leobenzol: oh I didn't get confirmation on that so I wasn't sure, changed it now

willbouch (approved):

adrien2p (approved):

PR #13646, opened 2025-10-01
chore(): Add instrumentation to deps

adrien2p:
**What**
Add opentelemetry packages to deps package


packages/deps/package.json
```diff
@@ -38,6 +41,9 @@
     "@mikro-orm/knex": "6.5.5",
     "@mikro-orm/migrations": "6.5.5",
     "@mikro-orm/postgresql": "6.5.5",
+    "@opentelemetry/exporter-zipkin": "^2.1.0",
```
sradevski: We have these 5 in platform:
```
 "@opentelemetry/exporter-trace-otlp-proto": "^0.53.0",
    "@opentelemetry/instrumentation-pg": "^0.44.0",
    "@opentelemetry/resources": "^1.26.0",
    "@opentelemetry/sdk-node": "^0.53.0",
    "@opentelemetry/sdk-trace-node": "^1.26.0",
```

The only one that we don't need to necessarily include is zipkin, as that's provider-specific. All the other packages here are generic and can work with any logging/tracing provider

adrien2p: then maybe I should not add the exporter for oltp either right?

sradevski: The OTLP is more generic, so should be fine to include it IMO, but it's ok if we don't. I am fine to include other packages as well btw, it's not like they add a ton of bloat

sradevski (approved):

adrien2p:
@shahednasser just mentioning you as it might impact the doc about the deps 👍

olivermrbl (approved):

PR #13655, opened 2025-10-02
fix(): Transform map

### LINKED ISSUE #13654: [Bug]: Workflow proxy property access triggers hidden transform calls causing WeakMap errors
### Package.json file

```JSON
"@medusajs/admin-sdk": "2.10.3",
    "@medusajs/cli": "2.10.3",
    "@medusajs/framework": "2.10.3",
    "@medusajs/icons": "2.10.3",
    "@medusajs/medusa": "2.10.3",
    "@medusajs/test-utils": "2.10.3",
    "@mikro-orm/cli": "6.4.3",
    "@mikro-orm/core": "6.4.3",
    "@mikro-orm/knex": "6.4.3",
```

### Node.js version

v18.19.0

### Database and its version

16.2

Medusa version: v2.10.3 Node.js version: v18.19.0 Database: PostgreSQL Operating system: macOS/Linux/Windows (all affected)

### Browser name

_No response_

### What happended?

In Medusa 2.10+, accessing properties on workflow step result proxies automatically triggers hidden `transform` calls that fail with "Invalid value used as weak map key" errors. This makes workflows unusable when following standard patterns.

The bug has two parts:
1. Explicit `transform` function fails when used with WorkflowData input
2. Simple property access (e.g., `stepResult.property`) creates hidden transform calls that also fail

1. Create a workflow with multiple steps
2. Try to pass data between steps using property access:
```typescript
const step1Result = fetchDataStep(input);
const step2Result = processStep({
  data: step1Result.data  // This line causes the error!
});
```
3. Run the workflow
4. Observe "Invalid value used as weak map key" error

### Expected behavior

Property access on step results should work without errors. Developers should be able to pass specific properties between workflow steps without triggering hidden transform failures.

### Actual behavior

Any property access on proxy objects triggers a hidden transform call that fails with:
```
TypeError: Invalid value used as weak map key
    at WeakMap.set (<anonymous>)
    at DistributedTransaction.setTemporaryData
```

#### Failing Code (Hidden Transform)
```typescript
// This looks innocent but FAILS
const exportDataWorkflow = createWorkflow(
  "export-data",
  function (input: WorkflowData<ExportDataInput>) {
    const step1Result = fetchExportJobStep(input);

    // Property access triggers hidden transform → WeakMap error
    const step2Result = fetchDataStep({
      exportJob: step1Result.exportJob  // ❌ FAILS HERE
    });
  }
);
```

#### Root Cause in Medusa Source
`/packages/core/workflows-sdk/src/utils/composer/helpers/proxy.ts`:
```typescript
get(target: any, prop: string | symbol): any {
  if (prop in target) {
    return target[prop]
  }
  // THIS CREATES A HIDDEN TRANSFORM!
  const transformer = transform({ target }, function (data) {
    return data.target?.[prop]
  })
  return transformer
}
```

#### Working Workaround
```typescript
// Pass entire proxy objects, no property access
const exportDataWorkflow = createWorkflow(
  "export-data",
  function (input: WorkflowData<ExportDataInput>) {
    const step1Result = fetchExportJobStep(input);

    // ✅ WORKS - Pass entire proxy
    const step2Result = fetchDataStep(step1Result);
  }
);

// Inside fetchDataStep:
async (input: any, context) => {
  const exportJob = input.exportJob;  // Extract at runtime
}


Stack trace showing the error path:
```
TypeError: Invalid value used as weak map key
    at WeakMap.set (<anonymous>)
    at DistributedTransaction.setTemporaryData (/node_modules/@medusajs/orchestration/src/transaction/distributed-transaction.ts:315:28)
    at Proxy.returnFn (/node_modules/@medusajs/workflows-sdk/src/utils/composer/transform.ts:198:38)
    at async resolveProperty (/node_modules/@medusajs/workflows-sdk/src/utils/composer/helpers/resolve-value.ts:22:11)
    at async resolveValue (/node_modules/@medusajs/workflows-sdk/src/utils/composer/helpers/resolve-value.ts:83:7)
    at async Object.invoke (/node_modules/@medusajs/workflows-sdk/src/utils/composer/helpers/create-step-handler.ts:82:32)
```
```

### Actual behavior

  Any property access on proxy objects triggers a hidden transform call that fails with:
  TypeError: Invalid value used as weak map key
      at WeakMap.set ()
      at DistributedTransaction.setTemporaryData

  This section explains:
  1. What triggers it: Any property access on proxy objects (like stepResult.exportJob)
  2. What happens: A hidden transform call is created automatically
  3. The error: "Invalid value used as weak map key"
  4. Where it fails: In the WeakMap.set operation inside setTemporaryData

  The actual behavior is that even simple, innocent-looking property access like step1Result.exportJob causes a critical error
  because Medusa's proxy implementation secretly creates a transform function call, which then fails when trying to store data
  in a WeakMap.

### Link to reproduction repo

https://github.com/medusajs/medusa

adrien2p:
**What**
It seems that for some reason the weak map fail in some scenario, but after investigation, the usage of map would not have a bad impact as it will be released after the Distributed transaction if finished. Therefore, falling back to Map instead

FIXES https://github.com/medusajs/medusa/issues/13654

NOTE: Waiting for the user feedback as he is also using node 18. We also use the exact same pattern in all our core flows without issues 🤔


willbouch (approved):
LGTM. We will wait for feedback I guess but code looks fine

carlos-r-l-rodrigues (approved):

PR #13660, opened 2025-10-02
chore(): Caching improvements

adrien2p:
RESOLVES CORE-1229
RESOLVES CORE-1232

**What**
- Remove the in memory as a default provider, now it must be configured as part of the module options to enable it
- Add support for `maxSize` inmemory cache to prevent bloating the process memory


adrien2p:
cc @sradevski, in memory disabled by default plus support for max size with deafult to 300MB for in memory

packages/modules/caching/src/types/index.ts
```diff
@@ -32,6 +57,14 @@ export type CachingModuleOptions = Partial<ModuleServiceInitializeOptions> & {
    * Time to keep data in cache (in seconds)
    */
   ttl?: number
+
+  /**
+   * Enable and configure the built in memory cache
+   */
+  in_memory?: MemoryCacheModuleOptions & {
+    enable?: boolean
+  }
```
olivermrbl, line 67: thought(non-blocking): should we add the `@private` on this property until we have a clearer picture of how/if we want to use it?

olivermrbl (approved):
💪

PR #13663, opened 2025-10-02
chore(): Default caching configuration and gracefull redis error handling

adrien2p:
RESOLVES CORE-1231
RESOLVES CORE-1230

**What**
Add the caching modules and configuration by default in cloud + support proper gracefull non blocking error handling for read operations but queuing and retry for write operations


adrien2p:
cc @sradevski , module configured by default plus gracefull connection loss handling configured by default in the provider, read ops return prematurely, write ops are run in the background anyway and rely on the queuing and retry

packages/core/utils/src/common/define-config.ts
```diff
@@ -301,6 +301,21 @@ function resolveModules(
...
+    {
+      resolve: MODULE_PACKAGE_NAMES[Modules.CACHING],
+      options: {
+        providers: [
+          {
+            id: "caching-redis",
+            resolve: "@medusajs/medusa/caching-redis",
+            is_default: true,
+            options: {
+              redisUrl: process.env.REDIS_URL,
```
sradevski:
```suggestion
              redisUrl: process.env.CACHE_REDIS_URL,
```

We might use different Redis clusters, so it's better to separate the two, even if they have the same value

packages/core/utils/src/common/define-config.ts
```diff
@@ -301,6 +301,21 @@ function resolveModules(
         ],
       },
     },
+    {
```
sradevski: Let's enable the caching module only if `process.env.CACHE_REDIS_URL` is set

sradevski (commented):
Just a couple of minor comments, LGTM otherwise

packages/core/utils/src/common/define-config.ts
sradevski, line 262: I guess it's a breaking change to remove this one, but I imagine it's for the better? I doubt people use it, but still

adrien2p: I believe some of our users/clients are currently using it, @olivermrbl mentioned something about it sometimes ago

sradevski: Fine to keep it then, I leave it up to you guys to decide here 👍

sradevski (approved):

olivermrbl (approved):

adrien2p:
I ll need to investigate, it seems that the connection get closed on redis and since all our redis modules rely on the same connection, it is then closed for other modules

PR #13668, opened 2025-10-02
fix(core-flows): lock draft order workflows

willbouch:
I added a lock in all workflows of the draft order. I don't think there are drawbacks and it will make sure we don't run into concurrency issues in with draft orders. I don't see why we would not add this in the order workflows, let me know and I can add it too.

I also didn't see any workflow that is long enough to justify adding a timeout of more than 2 seconds, but let me know if you think otherwise, we can discuss adjustments :)

CLOSES-1228


packages/core/core-flows/src/draft-order/workflows/delete-draft-order.ts
```diff
@@ -59,6 +59,8 @@ export const deleteDraftOrderWorkflowId = "delete-draft-order"
 export const deleteDraftOrdersWorkflow = createWorkflow(
   deleteDraftOrderWorkflowId,
   (input: WorkflowData<DeleteDraftOrderStepInput>) => {
+    // TODO - Acquire lock for every order?
```
willbouch: Here, we delete a list of orders, so, I can either forget about the lock for that workflow, or we could create a lock for every order. What do you suggest?

adrien2p: I am not sure a lock is necessary for the delete, the question is more about "what would happen in case of concurrent delete or race deletion", I would say instead of locking, the workflow could be sort of idempotent, meaning that the query does not have to throw if the order does not exist and instead we add a when then such that if there is no order we do nothing. wdyt? cc @carlos-r-l-rodrigues

adrien2p (approved):

packages/core/core-flows/src/draft-order/workflows/delete-draft-order.ts
olivermrbl: +1 for making this idempotent

olivermrbl (approved):

packages/core/core-flows/src/draft-order/workflows/delete-draft-order.ts
willbouch: @adrien2p @olivermrbl might be my understanding of locks missing (especially that I am not 100% sure to understand the difference between the locking mechanism we have in our code and the one that comes with postgres), but the lock in deletes sort of makes sense in my mind when we speak of an order? Essentially, when I want to delete order X and concurrently make a request to update order X, then I would expect my update to fail because the order was deleted. Granted, the result is the same because worst case scenario, we delete something that was just updated. But, if we try to think about it, it does not really make sense to make an update to something that was asked to be deleted. Even if the request was milliseconds ago. But maybe the postgres lock is enough? Feel free to call me on Slack, might be quicker to answer

carlos-r-l-rodrigues:
```suggestion
```

packages/core/core-flows/src/draft-order/workflows/delete-draft-order.ts
```diff
@@ -78,6 +80,8 @@ export const deleteDraftOrdersWorkflow = createWorkflow(
 
     deleteDraftOrdersStep({ orderIds: input.order_ids })
 
+    // TODO - Release
```
carlos-r-l-rodrigues:
```suggestion
```

packages/core/core-flows/src/draft-order/workflows/delete-draft-order.ts
carlos-r-l-rodrigues: The is no purpose to lock the deletes, generally speaking.
I would also not change the behavior of making this flow idempotent in this same PR.
By idempotent, I mean the flag on the workflow, with some `retentionTime` set, so every call made to this workflow would return the same response while the workflow execution is available.

carlos-r-l-rodrigues (approved):

packages/core/core-flows/src/draft-order/workflows/delete-draft-order.ts
adrien2p: I didn’t meant the workflow idempotent, but rather the behavior such that you can call delete and if nothing exists it does not throw 🙃

adrien2p: Especially that the result of this workflow is void 😆

PR #13675, opened 2025-10-03
fix(): caching better identify default provider

adrien2p:
**What**
Consider single provider configuration to be the default one even if not set as default


willbouch (approved):
Missing changeset :)

willbouch (approved):

PR #13684, opened 2025-10-03
chore(medusa-test-utils): shutdown on afterAll

carlos-r-l-rodrigues:


packages/medusa-test-utils/src/module-test-runner.ts
```diff
@@ -193,6 +193,7 @@ export function moduleIntegrationTestRunner<TService = any>({
     beforeEach(beforeEach_)
     afterEach(afterEach_)
     afterAll(async () => {
+      await shutdown()
```
willbouch, line 196: Why on afterall instead of aftereach? It makes sense but what is the reason behind it?

adrien2p: When I fixed the lifecycle execution, I fugred out that some of our modules close the redis connections in the *shutdown hooks, leading to redis being close in subsequent tests as the connection is created in the loaders. So, the idea for now is to kind of revert the fix and we will improve the connection management later which is a bigger work

adrien2p: though, I didnt expect the shutdown entirely to be move @carlos-r-l-rodrigues but rather move line 67-68 into the if here `packages/medusa-test-utils/src/init-modules.ts`

carlos-r-l-rodrigues: The main change on this PR will be reverted as it doesn't really fix the issue we're facing. However I might update it to keep the function to clean the database when tests are stopped manually on our local machine

PR #13693, opened 2025-10-06
feat(core-flows, promotion): limit promotion use per promo code

fPolic:
**What**
- limit promotion use per promo code


packages/admin/dashboard/src/i18n/translations/en.json
```diff
@@ -2262,6 +2262,10 @@
         "usage": {
           "title": "Usage",
           "description": "Set a limit on how many times the promotion can be used."
+        },
+        "use_by_attribute": {
```
willbouch: Why snake case?

willbouch: EDIT: some properties are snake and others are camel... huh haha

packages/admin/dashboard/src/routes/campaigns/campaign-detail/components/campaign-budget/campaign-budget.tsx
```diff
@@ -26,9 +26,9 @@ export const CampaignBudget = ({ campaign }: CampaignBudgetProps) => {
             level="h3"
           >
             {campaign.budget?.type === "use_by_attribute"
-              ? campaign.budget?.attribute === "customer_id"
-                ? t("campaigns.budget.fields.limitBudgetAttributeCustomer")
-                : t("campaigns.budget.fields.limitBudgetAttributeEmail")
+              ? t("campaigns.budget.fields.totalUsedByAttribute", {
+                  attribute: campaign.budget?.attribute,
```
willbouch: The attribute here will always be in English even if we get the translation from another language right?

packages/modules/promotion/src/services/promotion-module.ts
```diff
@@ -412,7 +412,15 @@ export default class PromotionModuleService
...
-        const attributeValue = registrationContext[attribute]
+        /**
+         * `promotion_code` is an exception, since it doesn't apply for the entiere context but per promotion
+         * if multiple promotion are applied on a cart, they share computation context
+         */
+        const attributeValue = (
+          attribute === "promotion_code"
+            ? registrationContext[attribute]
+            : promotion.code
+        ) as string | null
```
olivermrbl, line 423: thought: Tbh. I am not a huge fan of this; it feels dirty to have hard-coded exceptions like this in the code. 

I think we should consider adding a limit directly on the promotion as an alternative solution. That limit could be combined with a campaign budget to solve this use case. 

I am curious to hear what you and others think @medusajs/os

olivermrbl: For context, the use case we are trying to solve here is:
> Some of our users would like to create a single campaign called Customer Service Promos under which their team can create one-time promotions for individual customers. However, the current implementation prevents this because the limit applies at the campaign level rather than per promotion. They are once again left with creating a new campaign for recurring customers (i.e. customers that have already received a code from the campaign in the past).

fPolic: Yeah wasn't happy with that either.

Wdyt about modifying `CampaignBudgetUsageContext` to have scopes that are applied per promotion, for example:

```
export type CampaignBudgetUsageContext = {
 /** applied on all promotions **/
  customer_id: string | null
  customer_email: string | null

   /** applied only on matching promotions **/  
  promotions_context: {
    [promo_id]: {
      code: string | null
    }
  }
}
```

Then, when we iterate over promotions when computing actions or registering usage, we merge the context that is scoped for that promotion?

packages/admin/dashboard/src/i18n/translations/en.json
fPolic: good point, will fix that

packages/admin/dashboard/src/routes/campaigns/campaign-detail/components/campaign-budget/campaign-budget.tsx
fPolic: it is the name of the attribute, it can be anything user defines in theory

fPolic:
closing in favor of https://github.com/medusajs/medusa/pull/13760

PR #13696, opened 2025-10-06
chore: updated TSDocs of the Caching Module service + provider

shahednasser:
Closes DX-2177

### DIFF
--- a/packages/core/types/src/caching/index.ts
+++ b/packages/core/types/src/caching/index.ts
@@ -1,7 +1,20 @@
 import { IModuleService, ModuleJoinerConfig } from "../modules-sdk"
 
-type Providers = string[] | { id: string; ttl?: number }[]
+type Providers = string[] | { 
+  /**
+   * The ID of the provider to use, as set in `medusa-config.ts`.
+   */
+  id: string;
+  /**
+   * Optional custom time-to-live (TTL) (in seconds) for this provider. If not provided, the default TTL configured
+   * in the provider is used.
+   */ 
+  ttl?: number
+}[]
 
+/**
+ * @since v2.11.0
+ */
 export interface ICachingModuleService extends IModuleService {
   // Static trace methods
   // traceGet: (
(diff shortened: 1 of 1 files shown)


packages/core/types/src/caching/index.ts
```diff
@@ -25,37 +38,114 @@ export interface ICachingModuleService extends IModuleService {
...
+   * const data = await cacheModuleService.get({
+   *   key: "products",
+   * }) as { id: string; title: string; }
+   * ```
+   * 
+   * To retrieve by tags:
+   * 
+   * ```ts
+   * const data = await cacheModuleService.get({
+   *   tags: ["products-list"],
```
adrien2p, line 59: nit: should we use a proper conventional tag such as `Product:list:*`?

packages/core/types/src/caching/index.ts
```diff
@@ -25,37 +38,114 @@ export interface ICachingModuleService extends IModuleService {
...
+   * })
+   * ```
+   * 
+   * To store with tags:
+   * 
+   * ```ts
+   * await cacheModuleService.set({
+   *   key: "products-list",
+   *   data: [{ id: "prod_123", title: "Product 123" }],
+   *   tags: ["products"]
```
adrien2p, line 119: Q: when doing that you wont benefit from the auto invalidation, but if you use a conventional tag such as in this example `Product:prod_123` or `Product:list:*` then you can also have the default strategy auto invalidation kicking in, dont know if we want to mention that.

Basically, specifying tags manually without following convention kind of disable auto invalidation as well

packages/core/types/src/caching/index.ts
```diff
@@ -66,77 +156,405 @@ export interface ICachingModuleService extends IModuleService {
...
+   * })
+   * ```
+   * 
+   * This example will clear the item with the key `products` regardless of its `options.autoInvalidate` value.
+   * 
+   * To clear by tags:
+   * 
+   * ```ts
+   * await cacheModuleService.clear({
+   *  tags: ["products-list"]
```
adrien2p, line 223: nit: same as previously, but depending on your decision you can act on all of them :)

packages/core/types/src/caching/index.ts
```diff
@@ -66,77 +156,405 @@ export interface ICachingModuleService extends IModuleService {
...
+   * To clear only items that were set to automatically invalidate:
+   * 
+   * ```ts
+   * await cacheModuleService.clear({
+   *  tags: ["products-list"],
+   *  options: { autoInvalidate: true }
+   * })
+   * ```
+   * 
+   * This example will only clear items with the tag `products-list` that were set with `options.autoInvalidate: true`.
```
adrien2p, line 238: sugg: set with either autoInvalidate true or not set at all, we consider not set as true

packages/core/types/src/caching/index.ts
```diff
@@ -66,77 +156,405 @@ export interface ICachingModuleService extends IModuleService {
...
+   * 
+   * @param input - The input object to compute the key from.
+   * @returns The computed cache key.
+   * 
+   * @example
+   * const key = await cacheModuleService.computeKey({
+   *   id: "prod_123",
+   *   title: "Product 123"
+   * })
+   * // key will be a string like "a1b2c3d4e5f6g7h8i9j0"
```
adrien2p, line 303: sugg: should we mention that the order of the keys does not impact the output?

packages/core/types/src/caching/index.ts
```diff
@@ -66,77 +156,405 @@ export interface ICachingModuleService extends IModuleService {
...
+   * @returns An array of computed cache tags.
+   * 
+   * @example
+   * const tags = await cacheModuleService.computeTags({
+   *   products: [{ id: "prod_123" }, { id: "prod_456" }],
+   *   category: { id: "cat_123" }
+   * }, {
+   *   operation: "updated"
+   * })
+   * // tags might be ["product-prod_123", "product-prod_456", "category-cat_123"]
```
adrien2p, line 321: sugg: the convention is `Entity:id` `Entity:list:*`, the list one is only generated for array

adrien2p (commented):
sorry for the delay 😅

shahednasser:
@adrien2p all good 😄 I'll open another PR to address those

PR #13702, opened 2025-10-07
chore(): Improve caching rollout

adrien2p:
RESOLVES CORE-1243 CORE-1244 CORE-1242

**What**
- remove useCache and move service calls to graph api
- move some remote query to graph api
- update caching inteface
- add some missing cache point


olivermrbl (approved):
Nice, this LGTM!

PR #13703, opened 2025-10-07
fix(medusa,utils,types): inventory management nullable

willbouch:
We should return `inventory_quantity: null` in case of no inventory set. 0 should be reserved for the "Out of stock" scenario, and null will be the "Not available" scenario


olivermrbl (approved):
nice!

PR #13712, opened 2025-10-08
chore(): improve cart operations + Mikro orm 6.4.16

adrien2p:


packages/core/utils/src/modules-sdk/create-pg-connection.ts
```diff
@@ -23,11 +23,16 @@ export function createPgConnection(options: Options) {
       idle_in_transaction_session_timeout:
         (driverOptions?.idle_in_transaction_session_timeout as number) ??
         undefined, // prevent null to be passed
+
+      connectionTimeoutMillis: 5000, // Fail fast on slow connects
+      keepAlive: true, // Prevent connections from being dropped
+      keepAliveInitialDelayMillis: 10000, // Start keepalive probes after 10s
```
olivermrbl: question: should these be configurable?

carlos-r-l-rodrigues (approved):

packages/modules/pricing/src/models/price.ts
```diff
@@ -40,6 +40,10 @@ const Price = model
       on: ["currency_code"],
       where: "deleted_at IS NULL",
     },
+    // {
+    //   on: ["currency_code", "price_set_id", "min_quantity"],
+    //   where: "deleted_at IS NULL",
+    // },
```
olivermrbl: question: looking at the migration file, we can remove this?

olivermrbl:
```suggestion
```

olivermrbl (approved):
lgtm

packages/modules/pricing/src/models/price.ts
adrien2p: Yes we can, it was the index altering the ordering 👌

PR #13729, opened 2025-10-10
fix(admin): support metadata in product images validation

samerdmaidi:
Previously, the product images validator only accepted the 'url' field, which caused the 'id' and 'metadata' fields to be silently stripped out during validation. This prevented users from updating image metadata via the admin API.

This change updates both CreateProduct and UpdateProduct validators to properly support the full image object structure:
- id (optional): for updating existing images
- url (optional for updates, required for creation)
- metadata (optional): for storing custom key-value pairs

This aligns the API validation with the underlying ProductImageDTO type system which already supports these fields.

### DIFF
--- a/packages/medusa/src/api/admin/products/validators.ts
+++ b/packages/medusa/src/api/admin/products/validators.ts
@@ -227,7 +227,14 @@ export const CreateProduct = z
     description: z.string().nullish(),
     is_giftcard: booleanString().optional().default(false),
     discountable: booleanString().optional().default(true),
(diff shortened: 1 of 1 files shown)


willbouch:
Hey and thanks for the contribution! Should we maybe also update it in this [file](https://github.com/medusajs/medusa/blob/6032f3ec40fcd86c88a074d74d1ade52ddbd55ef/packages/core/types/src/http/product/admin/payloads.ts#L183-L188)? For create and update? Otherwise, I believe Typescript will complain if we pass the params from the admin dashboard (which uses the SDK)

PR #13730, opened 2025-10-10
feat(medusa,types): product variant store endpoints

fPolic:
**What**
- add `/store/product-variants`
-  add `/store/product-variants/:id`


packages/medusa/src/api/utils/middlewares/products/set-pricing-context.ts
```diff
@@ -7,10 +7,20 @@ import { MedusaPricingContext } from "@medusajs/framework/types"
 import { MedusaError } from "@medusajs/framework/utils"
 import { NextFunction } from "express"
 
-export function setPricingContext() {
+type PricingContextOptions = {
+  priceFieldPaths?: string[]
+}
+
+const DEFAULT_PRICE_FIELD_PATHS = ["variants.calculated_price"]
```
olivermrbl: suggestion: should this be a global constant?

packages/medusa/src/api/store/product-variants/validators.ts
```diff
@@ -0,0 +1,56 @@
...
+  StoreProductVariantContextFields
+)
+
+export type StoreProductVariantParamsType = z.infer<
+  typeof StoreProductVariantParams
+>
+
+export const StoreProductVariantListParams = createFindParams({
+  offset: 0,
+  limit: 50,
```
olivermrbl: nit(non-blocking): think it would be reasonable to do 20 here to reduce default response size

packages/medusa/src/api/store/product-variants/query-config.ts
```diff
@@ -0,0 +1,39 @@
+export const defaultStoreProductVariantFields = [
+  "id",
+  "title",
+  "sku",
+  "barcode",
+  "ean",
+  "upc",
+  // "*images", TODO: uncomment
```
olivermrbl: todo: we can do this now, i guess

packages/medusa/src/api/store/product-variants/query-config.ts
```diff
@@ -0,0 +1,39 @@
...
+]
+
+export const retrieveProductVariantConfig = {
+  defaults: defaultStoreProductVariantFields,
+  isList: false,
+}
+
+export const listProductVariantConfig = {
+  ...retrieveProductVariantConfig,
+  defaultLimit: 50,
```
olivermrbl: nit(non-blocking): see other comment, think 20 is also a fine default limit

packages/medusa/src/api/store/product-variants/route.ts
```diff
@@ -0,0 +1,59 @@
...
+    )
+  }
+
+  const context: QueryContextType = {}
+
+  if (req.pricingContext) {
+    context["calculated_price"] = QueryContext(req.pricingContext)
+  }
+
+  const { data: variants = [], metadata } = await query.graph({
```
olivermrbl: thought: should we cache this? see `GET /store/products` for a reference implementation

packages/medusa/src/api/store/product-variants/[id]/route.ts
```diff
@@ -0,0 +1,62 @@
...
+    id: req.params.id,
+  }
+
+  if (req.pricingContext) {
+    idOrFilter.context = {
+      calculated_price: QueryContext(req.pricingContext),
+    }
+  }
+
+  const variant = await refetchEntity({
```
Comment (olivermrbl) [inline packages/medusa/src/api/store/product-variants/[id]/route.ts]: nit(non-blocking): we've talked about gradually moving away from using the `refetchEntity` and `refetchEntities` because they are nothing but a wrapper around `query.graph` call and doesn't add much value. Should we use it here?

olivermrbl (approved):
LGTM!

olivermrbl (approved):
LGTM

PR #13736, opened 2025-10-11
fix(): Remove redundant indexes to resolve schema warnings

### LINKED ISSUE #13735: [Bug]: Duplicate indexes on multiple tables in a fresh installation
### Package.json file

```JSON
This issue is not specific to a single user project's `package.json`. It is a schema bug originating from the migration files within the core Medusa repository (`@medusajs/cart`, `@medusajs/product`, etc.). The bug is present in the latest version of Medusa created via `npx create-medusa-app@latest my-medusa-store`.
```

### Node.js version

v21.0.0

### Database and its version

PostgreSQL 16.2

Any

### Browser name

N/A

### What happended?

After a fresh installation of Medusa pointing to a new PostgreSQL database, the database schema contains multiple duplicate indexes. These are identified by database analysis tools like the Supabase Advisor.

The redundant indexes exist on tables across several core modules, including:
- `product_collection`
- `cart_line_item` and other cart-related tables
- `order_item` and `order_shipping`
- `customer_group`
- `inventory_level`

This is caused by legacy database migrations creating indexes that were later replaced by newer migrations with better-named indexes, but the old ones were never dropped

### Expected behavior

The database schema generated by Medusa's migrations should be clean, efficient, and free of redundant objects. No `Duplicate Index` warnings should be present after a fresh installation.

### Actual behavior

The database contains unnecessary duplicate indexes. This adds a small but real performance overhead to all write operations (INSERT, UPDATE, DELETE) on the affected tables and causes warnings in database health-checking tools.

A separate reproduction repository is not required as the issue can be reproduced directly with the Medusa CLI.
Steps to Reproduce:

Create a new Medusa project: `npx create-medusa-app@latest my-medusa-store`
Configure the .env file to point to a fresh, empty PostgreSQL database (e.g., on Supabase).
Inspect the database schema  by running the following SQL query on one of the affected tables:
`SELECT`
    `   tablename,`
    `   indexname,`
    `   indexdef`
`FROM`
   `   pg_indexes`
`WHERE`
    `  tablename = 'customer_group';`
Observe that two functionally identical indexes exist on the name column. The same pattern exists for the other tables mentioned.

### Link to reproduction repo

N/A

DDA1O1:
Fixes #13735

### What

This Pull Request introduces new database migrations to remove multiple redundant indexes across several core modules, including product, cart, order, customer, and inventory.

### Why

As detailed in issue #13735, a fresh Medusa installation produces numerous "Duplicate Index" warnings. These legacy indexes add unnecessary write overhead and provide no performance benefit. This PR cleans up the schema to resolve these warnings and improve database health.

### How

I have added one new, reversible migration file for each of the five affected modules:
- `@medusajs/product`
- `@medusajs/cart`
- `@medusajs/order`
- `@medusajs/customer`
- `@medusajs/inventory`

Each migration's `up()` method safely drops the older, redundant index, and its `down()` method re-creates it, ensuring the change is fully reversible and non-destructive.

### Testing

I have tested these migrations by following the local development workflow outlined in the `CONTRIBUTING.md` guide.

1.  **Setup:**
    *   Cloned my forked Medusa repository locally .
    *   Created a separate, fresh test project using `npx create-medusa-app@latest my-medusa-store`.
    *   The test project's PostgreSQL database, which already contained the schema with the duplicate indexes.

2.  **Linking Local Source Code:**
    *   In the test project's `package.json`, I modified all `@medusajs/*` dependencies and resolutions to point to the local packages in my forked repository (e.g., `"@medusajs/product": "file:../medusa/packages/modules/product"`).
    *   From the test project's directory, I ran `yarn install` to link the local, modified Medusa source code into its `node_modules`.

3.  **Build & Migration:**
    *   Inside my forked Medusa repository, I ran `yarn build` to compile the new TypeScript migration files.
    *   From the root of the **test project**, I then executed the migration command: `npx medusa migration run`.

4.  **Verification:**
    *   The command successfully identified and ran only the new migration files I had created.
    *   I also confirmed via direct SQL queries that the old, redundant indexes were correctly dropped from all affected tables (`product_collection`, `customer_group`, etc.).


.changeset/poor-llamas-march.md
```diff
@@ -0,0 +1,9 @@
+---
+"@medusajs/cart": patch
+"@medusajs/customer": patch
+"@medusajs/inventory": patch
+"@medusajs/order": patch
+"@medusajs/product": patch
+---
+
+fix(db): Remove redundant indexes from product, cart, order, customer, and inventory modules to improve database schema health.
```
willbouch: We usually put the name of the modules affected (db is not a module). In your case, the change affects too many of them to put there, so you can just write `fix()` and that would be fine :)

packages/modules/order/src/migrations/Migration20251010125912.ts
```diff
@@ -0,0 +1,39 @@
+import { Migration } from '@mikro-orm/migrations';
+
+export class Migration20251010125912 extends Migration {
+
+  override async up(): Promise<void> {
+     // 1. On `order_item`: Drops the old composite index.
+    this.addSql(
+      'DROP INDEX IF EXISTS "IDX_order_item_order_id_version";'
```
willbouch: Maybe we should delete `IDX_order_item_version` ? The one you are deleting right now seems to be the one with the most "correct" name

packages/modules/order/src/migrations/Migration20251010125912.ts
```diff
@@ -0,0 +1,39 @@
...
+    );
+
+    // 2. On `order_shipping`: Drops the old composite index.
+    this.addSql(
+      'DROP INDEX IF EXISTS "IDX_order_shipping_order_id_version";'
```
willbouch: Same here, I think we should drop the other one no?

packages/modules/product/src/migrations/Migration20251011090616.ts
```diff
@@ -0,0 +1,18 @@
+import { Migration } from '@mikro-orm/migrations';
+
+export class Migration20251011090616 extends Migration {
+// UP: Adds the feature (the correct index on product_category).
+  override async up(): Promise<void> {
+    this.addSql(
+      'CREATE INDEX IF NOT EXISTS "IDX_product_category_deleted_at" ON "product_category" ("deleted_at");'
```
willbouch: Can you remove this? I am fine with removing indexes that are duplicates, but can we avoid adding new ones? I prefer keeping this a status quo for performance reasons. I am sure this would be ok, but I prefer to be safer. We have people working on performance right now in the team, so if they believe an index is missing here, they will add it :)

packages/modules/order/src/migrations/Migration20251010125912.ts
DDA1O1: hey i removed it cause the latest migration file introduced it so but we can go with the old one

packages/modules/order/src/migrations/Migration20251010125912.ts
DDA1O1: done

packages/modules/product/src/migrations/Migration20251011090616.ts
DDA1O1: sure

DDA1O1:
hey @willbouch , I've made all the changes you suggested.

willbouch (dismissed):

willbouch:
Hey @DDA1O1. It looks like those indexes (at least some I checked) are also defined in the models (so the TS files defining models like "Order" or "OrderShipping". For example, [here](https://github.com/medusajs/medusa/blob/823a5c75ff97cf0acbd4ddf00f43ab3b49f0f210/packages/modules/order/src/models/order-shipping-method.ts#L63-L68) 

WE should clean that up/make sure the stuff there is up to date

DDA1O1:
> Hey @DDA1O1. It looks like those indexes (at least some I checked) are also defined in the models (so the TS files defining models like "Order" or "OrderShipping". For example, [here](https://github.com/medusajs/medusa/blob/823a5c75ff97cf0acbd4ddf00f43ab3b49f0f210/packages/modules/order/src/models/order-shipping-method.ts#L63-L68)
> 
> WE should clean that up/make sure the stuff there is up to date

hey thanks for letting me know definitely going to clean  the dropped indexes

willbouch:
I see you cleaned them up, but the thing is that I feel some are missing no? You cleaned up duplicates, that's fine, but it looks like the model now do not define the IDX that actually live in the database. The model has one IDX defined, which you removed, but the database had 2 IDX and you removed one. That means that the model is missing the actual IDX missing no?

DDA1O1:
I should have talked to you before implementing any change. Is there anything else that i should watch out or should i define  the indexes that actually live  in the database

willbouch:
@DDA1O1 Yeah I did not know before yesterday that the IDS were defined in the model too, sorry on that one. That being said, I believe you should just clean up the ones that are not in the model. Even if the name of the index is not the best one of the 2, it is simpler to get rid of those ones that are not in the model I think

DDA1O1:
ok here is what i found for some indexes like in, order module the indexes are defined in models so i can replace them with the best name as you suggested. For modules like inventory, customer and product the indexes are defined in old migration files itself not in the models so it does not need any change.

packages/modules/order/src/models/order-item.ts
```diff
@@ -34,7 +34,7 @@ const _OrderItem = model
       where: "deleted_at IS NULL",
     },
     {
-      name: "IDX_order_item_version",
+      name: "IDX_order_item_order_id_version",
```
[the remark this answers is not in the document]
DDA1O1: its fine

DDA1O1:
hey @willbouch  can you check it out

packages/modules/cart/src/models/line-item-adjustment.ts
```diff
@@ -24,10 +24,10 @@ const LineItemAdjustment = model
       on: ["promotion_id"],
       where: "deleted_at IS NULL AND promotion_id IS NOT NULL",
     },
-    {
-      name: "IDX_adjustment_item_id",
+     {
+      name: "IDX_cart_line_item_adjustment_item_id",
       on: ["item_id"],
-      where: "deleted_at IS NULL",
+      where: "deleted_at IS NULL AND item_id IS NOT NULL",
```
willbouch: I see you add something in the where here (and you do it for other models) but the migration file does not update the actual index in the DB? So, the model will say `where: "deleted_at IS NULL AND item_id IS NOT NULL"` but what we actually have in the DB is the already existing index we have which seems to only have `where: "deleted_at IS NULL"`. Or am I missing something?

DDA1O1: no you are right the definition should exactly match with the index definition that lives on the DB

willbouch (approved):

PR #13746, opened 2025-10-14
Fix/workflow async concurrency

adrien2p:
**What**
Update workflow engine concurrency save management to ensure consistency and reliability of the workflow data no matter how many concurrent save are made and no matter the order of the save


chuxi:
I had an error when running the workflow with some `transform` functions. one transform workflow data is `undefined` in the steps, even I log out the workflow data in `transform` function body correctly and returned. But when the data applied in the steps, it is `undefined`.

I downgraded the version from `2.10.3` to `2.10.1`, then the error disappeared. `2.10.2` also has the error. 

I guess it is caused by the concurrency. However, I have no time and not familiar to the under layer of medusajs workflows. When I saw this pr, I wish it fixes the error in `2.10.4`.

PR #13763, opened 2025-10-16
feat(admin-*,dashboard): add dashboard i18n extensions

### LINKED ISSUE #13698: [Bug]: react-i18next provider not available in plugins
### Package.json file

```JSON
fresh 2.10.3 backend and plugin starters
```

### Node.js version

v22.13.1

### Database and its version

PostgreSQL 17.2

macOS

### Browser name

_No response_

### What happended?

the "react-i18next" package is not marked as external in the admin-bundler. This means that plugins use a second instance of i18n that is never initialized.
It is currently not possible to use translations already provided by medusa in custom plugins(for example all the strings in a data table: pagination, search, filter, sort, create button etc)

I've added a reproduction repo below(just for convenience, since the plugin starter is quite old) with a minimal test page

Steps:
- [backend] setup the database, seed and create user
- [plugin] `npx medusa plugin:develop`or `npx medusa plugin:publish`
- [backend]`npx medusa plugin:add my-medusa-plugin && yarn && rm -rf node_modules/.vite/deps && yarn dev` 

Solution:
Adding "react-i18next" in these 2 places:
https://github.com/medusajs/medusa/blob/51859c38a728c5f05cf30f09f4c6a55657895de7/packages/admin/admin-bundler/src/utils/config.ts#L38

https://github.com/medusajs/medusa/blob/51859c38a728c5f05cf30f09f4c6a55657895de7/packages/admin/admin-bundler/src/commands/plugin.ts#L26

### Expected behavior

the strings provided by the dashboard should be available

### Actual behavior

they aren't

### Link to reproduction repo

https://github.com/leobenzol/test-plugin-providers

leobenzol:
## Summary

**What**

Adds functionality to directly export custom translations to be used in the admin dashboard, supporting both "direct" extensions from the main backend project as well as from plugins

**Why**  

It is currently possible to extend the react-i18next instance provided by medusa, but the process is clunky: it requires the same boilerplate component in every plugin and project and the "extender" component must be used in each custom route/widget, since one cant access the common parent scope. See one of my plugins [example](https://github.com/leobenzol/medusa-plugin-tolgee/blob/main/package/src/admin/components/PluginI18n.tsx)

**How** 

By creating a new virtual module for i18n, just like other admin features like routes/widgets etc... the DashboardApp now merges all i18n virtual modules into a single "resources" object which is used to initialize the i18n instance directly

**Testing** 

Local testing. Checked extensions in the main project and in plugins(both plugin:develop and plugin:publish) in  development and production env.

I noticed now that the other virtual modules have some form of test for the code generation part, should I add one for i18n as well?

---

## Examples

```ts
// In src/admin/i18n/index.ts:
import en from "./en.json" with { type: "json" } // "with" clause is mandatory
import fr from "./fr.json" with { type: "json" }

export default {
    en: {
        customNs: en,
        translation: en // will be merged with the default namespace
    },
    fr: {
        customNs: fr,
        translation: fr
    }
}
```

```ts
// In any react component:
import { useTranslation } from "react-i18next"
const { t } = useTranslation("customNs")

t("customNsKey") // uses user-defined keys
t("actions.create") // falls back to medusa's translations, unless overridden
```

---

## Checklist

---

## Additional Context

Closes #13698


packages/admin/admin-shared/src/utils/deep-merge.ts
```diff
@@ -0,0 +1,26 @@
+import { isObject } from "./is-object"
+
+export function deepMerge(target: any, source: any) {
```
leobenzol: importing it from "@medusajs/utils" gives a vite error "Uncaught TypeError: util.inherits is not a function"  so i duplicated it, lmk if that's ok

packages/admin/admin-vite-plugin/src/utils.ts
```diff
@@ -49,7 +49,7 @@ export function generateModule(code: string) {
   }
 }
 
-export const VALID_FILE_EXTENSIONS = [".tsx", ".jsx", ".js"]
+export const VALID_FILE_EXTENSIONS = [".tsx", ".jsx", ".js", ".ts"]
```
leobenzol: was there a reason for the ".ts" extension to not be included? I tried to both test and just look at the code and it doesn't seem to be breaking anything.
I could add it as an optional "customExtensions" parameter to crawl, to be safe

willbouch:
@leobenzol if you could leave a few comments on the PR, it would be nice and make the review a bit easier. Some things might seem obvious but it would be nice to have context over what changes and why. Will review today by EOD :)

packages/admin/admin-vite-plugin/src/i18n/generate-i18n-hash.ts
```diff
@@ -0,0 +1,15 @@
+import fs from "fs/promises"
+import { generateHash } from "../utils"
+import { getI18nIndexFilesFromSources } from "./helpers"
+
+export async function generateI18nHash(sources: Set<string>): Promise<string> {
```
leobenzol: the hash is used for vite HMR. I just remembered i haven't seen it work in practice, so I'll investigate, but it can be fixed in the future if I dont find the cause

leobenzol:
@willbouch tysm, everything I *didn't* comment is just virtual module boilerplate. Hope everything makes sense

packages/admin/admin-vite-plugin/src/i18n/generate-i18n.ts
```diff
@@ -0,0 +1,26 @@
...
+    const indexFiles = await getI18nIndexFilesFromSources(sources)
+
+    const imports = indexFiles.map((file, index) => {
+        const normalizedPath = normalizePath(file)
+        return `import i18nTranslations${index} from "${normalizedPath}"`
+    })
+
+    const mergeCode = indexFiles.length > 0
+        ? `deepMerge(${indexFiles.map((_, index) => `i18nTranslations${index}`).join(', ')})`
+        : '{}'
```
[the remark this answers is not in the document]
leobenzol: good job little cursor 🎉  too bad I'd already noticed it but we love the spirit

packages/admin/admin-vite-plugin/src/i18n/generate-i18n.ts
```diff
@@ -0,0 +1,37 @@
+import { deepMerge } from "@medusajs/admin-shared"
+import { outdent } from "outdent"
+import { pathToFileURL } from "url"
+import { getI18nIndexFilesFromSources } from "./helpers"
+import { logger } from "../logger"
+
+export async function generateI18n(sources: Set<string>) {
```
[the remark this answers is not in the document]
leobenzol: Edit: to properly support HMR, the i18n index files must be imported in the generated code. It makes sense since otherwise vite wouldn't have any references to any files, which would make hmr impossible.

since the index files must be imported, they cant be merged outside and the deepMerge call must be moved into the generated code. This only happens in development, when building(either the project or a plugin) for prod you will always only have one index file so mergeCode = i18nTranslations0

willbouch (approved):

willbouch:
@shahednasser most likely will need documentation on this one here :)

willbouch:
> I noticed now that the other virtual modules have some form of test for the code generation part, should I add one for i18n as well?


@leobenzol I'm never gonna say no to tests ;)

fPolic (approved):
Thanks for the contribution @leobenzol, works nicely!

PR #13766, opened 2025-10-16
feat(order,dashboard): version order credit lines

willbouch:
## Summary

Added `version` to the already credit line entity

When issuing an Ad Hoc refund, we systematically create a credit line of the same value to balance out the order total. However, if something goes wrong in the workflow, there was no way to really revert the edits to the order because credit lines did not have version.

The revert flow is pretty thoroughly tested I noticed so I mostly tested manually. I confirmed that before, the credit lines were not reverted in case of an error and now it does.

---

## Checklist

CLOSES CORE-1248


olivermrbl (approved):
LGTM, I guess we could add a test for this, but since we are following an existing pattern,  it's not critical. Will let you decide

willbouch:
> LGTM, I guess we could add a test for this, but since we are following an existing pattern, it's not critical. Will let you decide

Will take a look

packages/modules/order/src/models/credit-line.ts
```diff
@@ -4,14 +4,15 @@ import { Order } from "./order"
...
-    order: model.belongsTo(() => Order, {
-      mappedBy: "credit_lines",
-    }),
+    version: model.number().default(1),
     reference: model.text().nullable(),
     reference_id: model.text().nullable(),
     amount: model.bigNumber(),
     raw_amount: model.json(),
     metadata: model.json().nullable(),
+    order: model.belongsTo<() => typeof Order>(() => Order, {
```
fPolic: nit: isn't the type inferred here?

fPolic (approved):
LGTM!

packages/modules/order/src/models/credit-line.ts
willbouch: Yes, you are absolutely right. That's what happens when you let Claude do too much haha

PR #13769, opened 2025-10-17
fix: workflow async concurrency

adrien2p:


packages/core/core-flows/src/order/workflows/mark-order-fulfillment-as-delivered.ts
```diff
@@ -259,12 +258,11 @@ export const markOrderFulfillmentAsDeliveredWorkflow = createWorkflow(
...
-      markFulfillmentAsDeliveredWorkflow.runAsStep({
-        input: { id: fulfillment.id },
-      }),
-      registerOrderDeliveryStep(deliveryData)
-    )
+    const deliveredFulfillment = markFulfillmentAsDeliveredWorkflow.runAsStep({
+      input: { id: fulfillment.id },
+    })
+
+    registerOrderDeliveryStep(deliveryData)
```
carlos-r-l-rodrigues, line 265: The test was expecting the error from `markFulfillmentAsDeliveredWorkflow` but 
both will throw if already delivered. It is better to prioritiz fulfillment validation first.

packages/modules/workflow-engine-redis/package.json
```diff
@@ -29,7 +29,7 @@
     "resolve:aliases": "tsc --showConfig -p tsconfig.json > tsconfig.resolved.json && tsc-alias -p tsconfig.resolved.json && rimraf tsconfig.resolved.json",
     "build": "rimraf dist && tsc --build && npm run resolve:aliases",
     "test": "jest --passWithNoTests --bail --forceExit -- src",
-    "test:integration": "jest  --forceExit -- integration-tests/**/__tests__/*.ts",
+    "test:integration": "jest  --forceExit --runInBand -- integration-tests/**/__tests__/index.spec.ts && jest --forceExit --runInBand -- integration-tests/**/__tests__/race.spec.ts",
```
olivermrbl: revert

carlos-r-l-rodrigues (approved):

packages/modules/workflow-engine-redis/package.json
olivermrbl, line 32: I guess this should be reverted?

```suggestion
    "test:integration": "jest --forceExit -- integration-tests/**/__tests__/**/*.ts",
```

carlos-r-l-rodrigues: This test has something that running one after the other makes it hang. We could not find yet what is causing this, we just suspect that something is happening on the test runner , or the afterEach/afterAll. So, for now we'll keep it like this, with `&&`

olivermrbl: Okay, is this something we should look into before the release?

carlos-r-l-rodrigues: No, we are confident this is only affecting this test. They run just fine when executed like this manually one after the other, so it has something to do with the test setup than the code itself

PR #13772, opened 2025-10-17
feat: Implement medusa payments provider

sradevski:
The PR introduces a medusa payments provider injected by default in the payment module.

CLOSES CLO-1226

### DIFF
--- a/packages/modules/payment/src/providers/payment-medusa/services/medusa-payments.ts
+++ b/packages/modules/payment/src/providers/payment-medusa/services/medusa-payments.ts
@@ -58,19 +58,22 @@ import {
   UpdateAccountHolderRequest,
 } from "../types/medusa-payments"
 
-export class MedusaPaymentsError extends Error {
-  public statusCode: number
-  public errorType?: string
+type HandledErrorType = { retry: true } | { retry: false; data: any }
+class CloudServiceError extends Error {
+  type: string
+  originalType: string
+  data: any
+  message: string
 
-  constructor(statusCode: number, errorType?: string, message?: string) {
+  constructor(type: string, originalType: string, data: any, message: string) {
     super(message)
-
-    this.statusCode = statusCode
-    this.errorType = errorType
+    this.type = type
+    this.originalType = originalType
+    this.data = data
+    this.message = message
   }
 }
 
-// TODO: Handle errors and retries (similar to Stripe)
 export class MedusaPaymentsProvider extends AbstractPaymentProvider<MedusaPaymentsOptions> {
   static identifier = "medusa-payments"
   protected readonly options_: MedusaPaymentsOptions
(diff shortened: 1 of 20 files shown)


packages/modules/payment/src/providers/payment-medusa/services/medusa-payments.ts
```diff
@@ -0,0 +1,634 @@
...
+        | string
+        | undefined,
+      idempotency_key: context?.idempotency_key,
+      ...additionalParameters,
+    }
+
+    const payment = await this.executeWithRetry(() => {
+      return this.request<{ payment: any }>("/payments", {
+        method: "POST",
+        body: JSON.stringify(intentRequest),
```
olivermrbl: nit(non-blocking): we could bake the stringifying into the `request`, so we can avoid having/remember to do this

packages/modules/payment/src/providers/payment-medusa/services/medusa-payments.ts
```diff
@@ -0,0 +1,634 @@
...
+  }
+
+  constructor(cradle: Record<string, unknown>, options: MedusaPaymentsOptions) {
+    super(cradle, options)
+
+    this.validateOptions(options ?? {})
+    this.options_ = options
+  }
+
+  request<T>(url: string, options: RequestInit): Promise<T> {
```
olivermrbl: question: do we plan to create proper types for the provider?

packages/modules/payment/src/providers/payment-medusa/types/index.ts
```diff
@@ -0,0 +1,18 @@
...
+   */
+  webhookSecret: string
+  /**
+   * The endpoint to use for the payments
+   */
+  endpoint: string
+  /**
+   * The handle of the cloud environment
+   */
+  environmentHandle: string
```
olivermrbl: question: how do we plan to populate this? ideally, happens without the user knowing; maybe with an environment variable as part of the cloud deployment?

olivermrbl (commented):
Overall LGTM, do we want to do the todos before we merge?

packages/modules/payment/src/providers/payment-medusa/types/index.ts
sradevski: We already set an environment variable, we'll use that.

packages/modules/payment/src/providers/payment-medusa/services/medusa-payments.ts
sradevski: Yes, as well as proper error handling (which is why it's still a draft)

packages/modules/payment/src/providers/payment-medusa/services/medusa-payments.ts
sradevski: Yes definitely 👍

sradevski:
@olivermrbl yes I'll do all TODOs before merging, which is why it's still a draft. I just wanted to start testing with snapshots first

olivermrbl (approved):
LGTM

PR #13777, opened 2025-10-18
feat(medusa): add generated .medusa/types to local plugins

leobenzol:
## Summary

**What** 

Copies the generated types inside `.medusa/types` to all "local" plugins(i.e plugins added with plugin:develop)

**Why**

Currently the only way to get type support for `query.graph` etc inside plugins is to manually copy the .medusa/types folder from the main project, this is clunky and prone to confusion if you forget to paste the changed types after editing a model.

**How**

Changed the type generation to a loop over all "local" plugins, detected as plugins with a local admin type(which includes the main project's "local plugin")

**Testing**

Local testing

---

## Checklist

---

> [!NOTE]
> Generates container and GraphQL types into each local plugin’s `.medusa/types` directory during start when types are enabled.
> 
> - **Type generation in `packages/medusa/src/commands/start.ts`**:
>   - Resolve local plugins via `getResolvedPlugins(...)` and iterate over those with `admin.type === "local"`.
>   - For each plugin, target `../../.medusa/types` relative to `plugin.admin.resolve`, clean the directory, then generate:
>     - `generateContainerTypes` with `interfaceName: "ModuleImplementations"`.
>     - `gqlSchemaToTypes` (when `gqlSchema` exists) with `filename: "query-entry-points"` and `interfaceName: "RemoteQueryEntryPoints"`.
> - **Changeset**: adds patch entry for `@medusajs/medusa`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 4ec290c8cfc193b65e1922b28bc5d6f2f0e46320. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


willbouch (approved):
@adrien2p wdyt?

adrien2p:
hey @leobenzol, just want to be sure I understand the needs correctly, did you face this needs in the following context?
- you create a new plugin
- in the meantime you have some customization part of your medusa project, let say a module
- in your plugin you want to query data from that module part of your medusa project but you dont have any types

is the comprehension correct?

leobenzol:
@adrien2p the module would be from the plugin itself. So you create a new plugin -> add module to the plugin -> use query.graph in a http route/workflow in the plugin.

Unless I'm missing something there isnt currently a way to generate types inside plugins, which makes sense since they dont have access to a medusa container without a project to run in. That's why for the PR I had to copy the type definitions from the main project(which also generates types for plugins added with plugin:develop)

adrien2p:
> @adrien2p the module would be from the plugin itself. So you create a new plugin -> add module to the plugin -> use query.graph in a http route/workflow in the plugin.
> 
> Unless I'm missing something there isnt currently a way to generate types inside plugins, which makes sense since they dont have access to a medusa container without a project to run in. That's why for the PR I had to copy the type definitions from the main project(which also generates types for plugins added with plugin:develop)

yeah Understood, just thinking about something, have you tried to configure medusa-config in your plugin (setting up the plugin itself as well) and run yarn build?

leobenzol:
@adrien2p yeah that would work, it'd be pretty much the same as what i do already for integration tests. But I thought it'd be nice to have them update automatically when a file changes and the server restarts, like it happens in the main project.

willbouch:
@adrien2p Any thoughts or is this ready?

adrien2p:
I believe it is good 👍

adrien2p (approved):

adrien2p:
I ve just reverted this pr, in the next release it will be possible to get the types generation with the build commands, which I think will be way better than relying on the start commands which should not be meant to generate types. Which mean that now having the medusa config within your plugin and then run yarn build should solve the root issue. please let me know if that makes sense @leobenzol

leobenzol:
@adrien2p It makes sense and I agree, but i tried it on the latest preview(2.12.4-preview-20251219120148) and it's pretty messy.

Take this plugin for example: https://github.com/leobenzol/medusa-plugin-tolgee It was a monorepo at first with integration-tests in its own full medusa project(which makes types trivial, just generate and copy), but I found a way to run tests in the same project which simplifies the .env, medusa-config, CI and publishing *a lot* and I really like how simple it is.
To generate types I have to:
- run medusa plugin:build(if you dont build first the type generation wont pickup any custom type, idk why)
- cd integration-tests && medusa build
- copy .medusa/types into the parent

This works generally pretty well, `medusa build` generates a lot of errors because of the "lite" .env and medusa-config but it works(unlike medusa develop which hangs forever, probably a missing env var)
If you dont want to copy the types folder and you instead move the medusa-config to the root it gets even worse, as "medusa build" overrides the ".medusa/server" folder, so you have to run "medusa plugin:build -> build -> plugin:build again"

Let me know if I'm missing something and my setup is overly complicated
I'm not sure about the best solution, but probably a --types-only flag on "medusa build" would be enough?

PR #13781, opened 2025-10-20
add Medusa Cloud Email provider

peterlgh7:
CLOSES CLO-1182
This PR introduces a default email notification provider for projects running in Medusa Cloud.


packages/core/types/src/common/config-module.ts
```diff
@@ -865,6 +885,12 @@ export type ProjectConfigOptions = {
       /*admin?: string[]*/
     }
   }
+
+  /**
+   * This property holds configurations for running in Medusa Cloud.
+   * It gets automatically populated in the cloud, and is not needed outside of it.
+   */
+  medusaCloudOptions?: MedusaCloudOptions
```
olivermrbl, line 893: super-nit(non-blocking): Do we need `medusa` and `Options` here? I think it's a little bit awkward. `cloud: {}` would be cleaner imo. You are already in a Medusa project + you are already in a config utility, i.e. prefix and suffix are redundant.

thought:(non-blocking): Wonder if we should keep this config `@private` until we make a public roll out? Just to avoid commenting on it prematurely.

packages/modules/notification/src/loaders/providers.ts
```diff
@@ -40,8 +38,35 @@ export default async ({
...
+    NotificationModuleOptions
 >): Promise<void> => {
+  let providers = options?.providers || []
+
+  // We add the Medusa Cloud Email provider if there is no other email provider configured
+  const hasEmailProvider = options?.providers?.some((provider) =>
+    provider.options?.channels?.some((channel) => channel === "email")
+  )
+  if (!hasEmailProvider) {
+    const { api_key, endpoint, environment_handle } =
```
olivermrbl: suggestion(non-blocking): would make `environment_handle` optional, so we can open this up for use in Bloom; this is also something we can do in a follow up

packages/modules/notification/src/providers/medusa-cloud-email.ts
```diff
@@ -0,0 +1,49 @@
...
+  async send(
+    notification: NotificationTypes.ProviderSendNotificationDTO
+  ): Promise<NotificationTypes.ProviderSendNotificationResultsDTO> {
+    try {
+      const response = await fetch(`${this.options_.endpoint}/send`, {
+        method: "POST",
+        headers: {
+          "Content-Type": "application/json",
+          Authorization: `Basic ${this.options_.api_key}`,
+          "x-medusa-environment-handle": this.options_.environment_handle,
```
olivermrbl, line 25: suggestion(non-blocking): make optional; see other comment

olivermrbl (approved):
Overall LGTM, added a few nits and suggestions

packages/modules/notification/src/loaders/providers.ts
peterlgh7: Sure, I'll do it in a follow-up. We'll probably have decided by then to send some other id or just nothing at all.

packages/core/types/src/common/config-module.ts
peterlgh7: Fair point. Suffix is surely unnecessary. I thought the prefix could remove some ambiguity, but I'll trust your naming over mine in this codebase :)

PR #13843, opened 2025-10-23
feat(admin-*,dashboard): i18n labels for menu item extensions

leobenzol:
## Summary

**What** 

Adds support for translating the label of custom UI routes in the dashboard sidebar, by leveraging the new i18n virtual module to define the needed custom keys.

**Why**

Not currently supported, this is the last step to have 100% localizable custom UIs

**How** 

By adding a new field `translationNs` to the RouteConfig object. If present, the `label` field acts as a translations key, otherwise the literal string is used as before.  

**Testing** 

Tested locally

---

## Examples

```ts
// Example usage
  label: "customLabels.foo",
  translationNs: "customNs",
})

  label: "app.nav.settings.header", // existing "Settings" key from medusa
  translationNs: "translation",     // ns used by medusa
})

  label: "Custom Route" // static label as before
})
```
refer to #13763 for creating the actual namespace and key

---

## Checklist

---

## Additional Context

### DIFF
--- a/packages/admin/dashboard/src/components/layout/nav-item/nav-item.tsx
+++ b/packages/admin/dashboard/src/components/layout/nav-item/nav-item.tsx
@@ -17,6 +17,7 @@ type ItemType = "core" | "extension" | "setting"
 type NestedItemProps = {
   label: string
   to: string
+  translationNs?: string
 }
 
 export type INavItem = {
@@ -27,6 +28,7 @@ export type INavItem = {
   type?: ItemType
   from?: string
   nested?: string
+  translationNs?: string
 }
 
 const BASE_NAV_LINK_CLASSES =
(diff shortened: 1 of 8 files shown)


packages/admin/dashboard/src/components/layout/nav-item/nav-item.tsx
```diff
@@ -90,10 +92,15 @@ export const NavItem = ({
   items,
   type = "core",
   from,
+  translationNs,
 }: INavItem) => {
+  const { t } = useTranslation(translationNs as any)
```
leobenzol: "any" cast necessary since react-i18next tries to be ts-safe, and rejects any namespace/keys that aren't defined in the core i18n files

willbouch (commented):
Hi @leobenzol I am not sure I understand what exactly is the thing we are supporting with this contribution? Could you join screenshots maybe? It might be only me that is missing something

leobenzol:
@willbouch see below. All core routes are localized, but custom routes like "Postmark Templates" have a static label that currently can't change when changing locale.
image

willbouch (approved):
LGTM

PR #13849, opened 2025-10-24
Add config to allow processing to be authorized in stripe base payments

scherddel:
## Summary

Added a configurable option `authorizeOnProcessing` to the `payment-stripe` provider that allows transactions in Stripe that may take longer to be processed by the complete cart workflow. 

Configuration example:
```js
providers: [
  {
    resolve: "@medusajs/medusa/payment-stripe",
    id: "stripe",
    options: {
      authorizeOnProcessing: true
    },
  },
],
```

When enabled, the payment can be automatically captured through the webhook or manually captured in the admin backend. The order remains in a pending/authorized state (shown as "orange") until Stripe changes the payment intent status to `succeeded`.

The solution treats payment intents with status `processing` as `authorized` when the `authorizeOnProcessing` flag is set to `true`. When `false`, it falls back to the previous behavior.

Stripe-based payments that have a delayed capturing cycle currently cause cart completion to fail. This feature provides a workaround for users who need to support such payment methods by explicitly enabling this behavior through configuration.

I tested approximately 5 different payment methods in Stripe to ensure existing functionality works as expected. The changes now allow delayed payment methods to function properly during cart completion.

## Checklist

---

## Additional Comments

I recognize that an ideal solution would introduce a more robust way of handling carts or orders with pending payments (e.g., a "parking" mechanism). However, this PR serves as a pragmatic workaround until a more comprehensive solution can be developed. In the meantime, this feature enables support for payment methods that would otherwise block cart completion.

### DIFF
--- a/packages/modules/providers/payment-stripe/src/core/stripe-base.ts
+++ b/packages/modules/providers/payment-stripe/src/core/stripe-base.ts
@@ -605,7 +605,7 @@ abstract class StripeBase extends AbstractPaymentProvider<StripeOptions> {
         return { status: PaymentSessionStatus.PENDING, data: paymentIntent }
       case "requires_confirmation":
       case "processing":
-        return { status: PaymentSessionStatus.PENDING, data: paymentIntent }
+        return { status: this.options_.authorizeOnProcessing?PaymentSessionStatus.AUTHORIZED:PaymentSessionStatus.PENDING, data: paymentIntent }
       case "requires_action":
         return {
           status: PaymentSessionStatus.REQUIRES_MORE,
--- a/packages/modules/providers/payment-stripe/src/types/index.ts
+++ b/packages/modules/providers/payment-stripe/src/types/index.ts
@@ -16,6 +16,10 @@ export interface StripeOptions {
    */
   automaticPaymentMethods?: boolean
   /**
+   * Automatically authorize the payment when processing (default is false)
+   */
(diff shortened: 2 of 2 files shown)


scherddel:
push

danielgran:
TYSM for providing this. Im not sure about what exactly might brick, but as a quick fix it totally did the job. IMHO Medusa should support a afterpay method, as this is a super common usecase.

For me, it even registeres the orders but does mark the payment as pending / ready to capure by payment with sepa, which is totally fine for me, as there is a capture integration /webhook for that.

Love to you

scherddel:
@danielgran Thank you so much for your feedback.

Ultimately, it comes down to balancing the risk of introducing a 'quick fix' that is difficult to deprecate later versus building a robust solution that handles complex payment event combinations.

This is a specific pain point regarding European compliance, where we must place orders on 'hold' to defer contract acceptance while still acknowledging the payment request via email. Medusa doesn't support this workflow out of the box, necessitating significant customization. I view this as a major functional gap, as it currently blocks 20% of available Stripe payment methods.

What do you think @olivermrbl ?

danielgran:
NOT agree to be stale. This is a major issue.

danielgran:
@shahednasser May we ask you to triage this issue?

scherddel:
.

NicolasGorga:
Closing as duoplicate of #15085

PR #13851, opened 2025-10-25
feat(dashboard): type-safe i18n for UI extensions

leobenzol:
## Summary

**What** 

Exports the type of medusa's i18n keys from the dashboard package, for autocomplete support.

**Why** 

Currently i18next is not able to provide proper ts autocomplete support when adding translations to projects or plugins.   
For example when adding a data table or form, I have to go back multiple times to the source i18n files in the medusa repo and search for the exact key name(is it "actions.confirm" or "general.confirm"? etc) and I forget it right after every single time.

Even if all medusa components were provided already translated, it's still very convinient to use keys from medusa(if context appropriate) for some custom components, since they are already translated into many languages(eg "yes", "no" and other very basic strings). Hence why ts support for the base keys would be very helpful

**How** 

Modified the generateTypes script to also copy the `en.json` file to the dist folder, and export a `Resources` type with the inferred keys in index.d.ts.
This allows users to define their own i18next type definitions including the base "translation" namespace, example below.

**Testing** 

Manual testing

---

## Examples

```ts
// In src/admin/i18next.d.ts
import type enTranslation from "./i18n/en.json"         // custom keys
import type { Resources } from "@medusajs/dashboard"    // medusa keys
declare module "i18next" {
    interface CustomTypeOptions {
        fallbackNS: "translation"
        resources: {
            translation: Resources["translation"]                     
            // all custom namespaces need to merge Resources["translation"] as well
            // otherwise when falling back to "translation", strings will have type "never"
            myCustomNs: typeof enTranslation & Resources["translation"]  
        }
    }
}

```

Then, both `useTranslation()` and `useTranslation("myCustomNs")` will have proper autocomplete support

---

## Checklist


adrien2p (approved):
Sounds good, @willbouch any thought on your side about this one?

willbouch (approved):
I thikn that is pretty useful too. Thanks for your work on plugins' DX recently @leobenzol. I still have a few other of your open PRs to go through. Will be doing that soon

PR #13876, opened 2025-10-27
fix(core-flows): payment error handling

SGFGOV:
## Summary

errors in the provider authorization step wasn't ever propogated up the stack. Instead the message received was mangled.  

*

Many times there are provider specific error codes and message that need to be visible in other parts of the stack via api etc. 

*Please provide answer here*

It's very simple. We check if the error is a medusaError. This implies the provider has sanitised the error. If it is we allow it to propogate up the stack as it is. Else, we continue with the earlier implementations.   

*Please provide answer here*

You can throw an error in a provider and you should see the same in your api response. 

---

## Examples

This isn't really a feature. Just a small bug fix
```ts
// this implies the error is already a MedusaError, so we can throw it safely,
      if (e instanceof MedusaError) {
        throw e
      }
```

## Additional Context

---

> [!NOTE]
> Propagates MedusaError from payment authorization and updates integration test to assert cart completion returns cart with payment authorization error.
> 
> - **Core Flows (`packages/core/core-flows`)**
>   - **Payment**: In `payment/steps/authorize-payment-session.ts`, rethrow provider errors when `MedusaError.isMedusaError(e)` is true to preserve original payment authorization errors.
> - **Integration Tests**
>   - Update `integration-tests/modules/__tests__/cart/store/carts.spec.ts` to mock `authorizePaymentSession` throwing `MedusaError.Types.PAYMENT_AUTHORIZATION_ERROR` and assert `/store/carts/:id/complete` returns a `type: "cart"` response with the payment authorization error.
> - **Changeset**
>   - Add patch release for `"@medusajs/core-flows"`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 098a2d5ea82fe5168c995a16285e2a11df9cb8d8. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


SGFGOV:
@olivermrbl , sorry to ping you directly.  would you mind reviewing and approving. it's blocking a development activity. We need a method of sending an error message up the stack from the the payment provider. This seemed the simplest way to do so

packages/core/core-flows/src/payment/steps/authorize-payment-session.ts
```diff
@@ -58,6 +58,11 @@ export const authorizePaymentSessionStep = createStep(
       logger.error(
         `Error was thrown trying to authorize payment session - ${input.id} - ${e}`
       )
+      // this implies the error is already a MedusaError, so we can throw it safely,
+      if (e instanceof MedusaError) {
+        throw e
+      }
+
```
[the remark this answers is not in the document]
SGFGOV: This is ok, because, when a medusaError is thrown, it signals that the provider has explicitly handled the error, and inturb confidential information won't be leaked.. Otherwise its silently swallowed as confidential information maybe exposed inadvertently,

willbouch:
Hi @SGFGOV ! Can you give a little bit more context as to what is happening? I think the PR makes sense, but I am scared that it might be a breaking change for some people. My reasoning behind this is "what exactly happens with those medusa errors that are "skipped"? You see the error received is mangled? Can you maybe provide an example?

SGFGOV:
When those messages are skipped the old flow continues..so whatever was the older behaviour prior to this PR, that is preserved.

packages/core/core-flows/src/payment/steps/authorize-payment-session.ts
```diff
@@ -58,6 +58,11 @@ export const authorizePaymentSessionStep = createStep(
       logger.error(
         `Error was thrown trying to authorize payment session - ${input.id} - ${e}`
       )
+      // this implies the error is already a MedusaError, so we can throw it safely,
+      if (e instanceof MedusaError) {
+        throw e
+      }
```
adrien2p: suggestion: I prefer using `MedusaError.isMedusaError` just to prevent any issues with instance of just in case. wdyt?

SGFGOV: makes sense. I've pushed an update

adrien2p (approved):

SGFGOV:
@adrien2p thanks for approving. has the been merged into the dev or will it be included in the next release

SGFGOV:
> @adrien2p thanks for approving. has the been merged into the dev or will it be included in the next release

Thanks I figured it out :) it's in the changeset

willbouch:
@SGFGOV It looks like one test is failing. I cannot deploy it without the text fixed :/
CleanShot 2025-10-30 at 08 46 22

SGFGOV:
Ok let me fix it :)

SGFGOV:
@willbouch please check and approve

willbouch (approved):

PR #13894, opened 2025-10-29
chore(): Remove barrel files in favor of imports usage in modules

adrien2p:
RESOLVES CORE-1261

**What**
- Re work module barrel file usage
- add package.json imports
- remove ts config alias usage
- remove tsc-alias usage and scripts


adrien2p:
Non stale

PR #13899, opened 2025-10-29
fix(pricing): price list prices perf issues

adrien2p:
**What**

  This PR addresses performance issues when updating price list prices and price sets prices in the pricing module. The changes focus on optimizing database queries and reducing computational overhead during price normalization and upsert operations.

  Key Changes

  Performance Optimizations in Pricing Module (packages/modules/pricing/src/services/pricing-module.ts)

  1. Optimized Price Hashing
    - Refactored `hashPrice()` function to avoid expensive sort operations
    - Moved from `JSON.stringify()` with sorted entries to direct string concatenation
    - Only includes properties that are actually present in the price object
    - Uses deterministic property order for consistent hashing
  2. Improved Price Normalization Logic
    - Simplified object merging in `normalizePrices()` to reduce redundant object creation
    - Changed rule operator validation from array `.includes()` to `Set .has()` for O(1) lookups
    - Replaced `.map().flat()` with `.flatMap()`
    - Pre-computed validation error messages outside the loop
  3. Targeted Database Queries
    - Added new helper function `buildPreNormalizationPriceConstraintsFromData()` to construct precise database queries
    - Changed `updatePriceListPrices()` and `addPriceListPrices()` to fetch only necessary prices instead of loading entire price lists with all relations
    - Added parallel query execution using `promiseAll()` to fetch price lists and prices concurrently
  4. Optimized `addPricesToPriceSet()` Method
    - Similar improvements as price list methods
    - Pre-normalizes prices before database queries to determine constraints
    - Fetches only relevant existing prices instead of all prices for all price sets

**Impact**
The optimizations reduce:
  - Database query overhead by fetching only necessary data
  - CPU time spent on hashing and validation operations
  - Memory allocation from redundant object creation

### DIFF
--- a/packages/modules/pricing/src/services/pricing-module.ts
+++ b/packages/modules/pricing/src/services/pricing-module.ts
@@ -1817,9 +1817,6 @@ const hashPrice = (
   if ("currency_code" in price) {
     parts.push(`cc:${price.currency_code ?? ""}`)
   }
-  // if ("amount" in price) {
-  //   parts.push(`amount:${price.amount ?? ""}`)
-  // }
   if ("price_set_id" in price) {
     parts.push(`ps:${price.price_set_id ?? ""}`)
   }
@@ -1874,9 +1871,6 @@ const buildPreNormalizationPriceConstraintsFromData = (
     if (price.currency_code !== undefined) {
       constraint.currency_code = price.currency_code
     }
-    // if (price.amount !== undefined) {
-    //   constraint.amount = price.amount
-    // }
     if ("price_set_id" in price && price.price_set_id !== undefined) {
       constraint.price_set_id = price.price_set_id
     }
--- a/packages/modules/pricing/src/services/pricing-module.ts
+++ b/packages/modules/pricing/src/services/pricing-module.ts
@@ -41,7 +41,6 @@ import {
   PricingRuleOperator,
   promiseAll,
   removeNullish,
-  simpleHash,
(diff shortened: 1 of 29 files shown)


packages/modules/pricing/src/services/pricing-module.ts
```diff
@@ -1785,20 +1810,119 @@ const isTaxInclusive = (
...
+  // Build hash using deterministic property order to avoid expensive sort operation
+  // Using a direct string concatenation approach for better performance
+  const parts: string[] = []
+
+  if ("currency_code" in price) {
+    parts.push(`cc:${price.currency_code ?? ""}`)
+  }
+  // if ("amount" in price) {
+  //   parts.push(`amount:${price.amount ?? ""}`)
+  // }
```
olivermrbl: remove?
```suggestion
```

packages/modules/pricing/src/services/pricing-module.ts
```diff
@@ -1785,20 +1810,119 @@ const isTaxInclusive = (
...
+    if (price.id) continue
+
+    const constraint: any = {}
+
+    if (price.currency_code !== undefined) {
+      constraint.currency_code = price.currency_code
+    }
+    // if (price.amount !== undefined) {
+    //   constraint.amount = price.amount
+    // }
```
olivermrbl: remove?
```suggestion
```

olivermrbl (approved):
champ!

PR #13911, opened 2025-10-30
chore(): Throw on migration up/down fail

adrien2p:
**What**
Do not silently handle migration failure, make the process fail and exit with non 0 code

### DIFF
--- a/packages/core/utils/src/modules-sdk/migration-scripts/migration-up.ts
+++ b/packages/core/utils/src/modules-sdk/migration-scripts/migration-up.ts
@@ -2,6 +2,7 @@ import { LoaderOptions, Logger, ModulesSdkTypes } from "@medusajs/types"
 import { mikroOrmCreateConnection } from "../../dal"
 import { loadDatabaseConfig } from "../load-module-database-config"
 import { Migrations } from "../../migrations"
+import { MedusaError } from "../../common/errors"
 
 const TERMINAL_SIZE = process.stdout.columns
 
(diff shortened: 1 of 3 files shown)


carlos-r-l-rodrigues (approved):

sradevski (approved):

packages/core/utils/src/modules-sdk/migration-scripts/migration-up.ts
```diff
@@ -52,6 +53,7 @@ export function buildMigrationScript({ moduleName, pathToMigrations }) {
       }
     } catch (error) {
       logger.error(`Failed with error ${error.message}`, error)
+      throw new MedusaError(MedusaError.Types.DB_ERROR, error.message)
```
sradevski, line 56: Let's just make sure we don't double-log the error somewhere higher up the stack, otherwise LGTM

adrien2p: you mean removing the log entirely?

sradevski: If we log the same error somewhere else, yes, but I don't know if we do

adrien2p: it happens that the command log the error too, but these scripts are also called in other places. Not sure this one or other ones should know about which one is logging. If you feel I should remove it still let me know

olivermrbl (approved):

packages/core/utils/src/modules-sdk/migration-scripts/migration-up.ts
sradevski: No all good, just wanted to raise it as a potential change, but it's not a big deal if we log twice

PR #13932, opened 2025-10-31
chore(): Fix dependencies

adrien2p:
RESOLVES CORE-1263


www/apps/api-reference/package.json
```diff
@@ -53,7 +53,7 @@
   },
   "devDependencies": {
     "@next/bundle-analyzer": "15.3.5",
-    "@types/jsdom": "^21.1.1",
+    "@types/jsdom": "^27.0.0",
```
adrien2p, line 56: @shahednasser do you see any issues with that? I am fixing all vulnerabilities and this rely on form-data < 4.0.4 which as critical issues

adrien2p:
@sradevski do you see any issues with the posthog upgrade?

carlos-r-l-rodrigues (approved):
🔥

sradevski:
@adrien2p these are the breaking changes (https://github.com/PostHog/posthog-js/blob/main/packages/node/CHANGELOG.md):

Breaking changes
1. feat: migrate to native fetch, Node 20+ required
2. PostHog Node now compresses messages with GZip before sending them to our servers when the runtime supports compression. This reduces network bandwidth and improves performance. Network traffic interceptors and test assertions on payloads must handle GZip decompression to inspect the data. Alternatively, you can disable compression by setting disableCompression: true in the client configuration during tests.

I believe we require node 20 so it should be good

adrien2p:
> @adrien2p these are the breaking changes (https://github.com/PostHog/posthog-js/blob/main/packages/node/CHANGELOG.md):
> 
> 
> 
> Breaking changes
> 
> 1. feat: migrate to native fetch, Node 20+ required
> 
> 2. PostHog Node now compresses messages with GZip before sending them to our servers when the runtime supports compression. This reduces network bandwidth and improves performance. Network traffic interceptors and test assertions on payloads must handle GZip decompression to inspect the data. Alternatively, you can disable compression by setting disableCompression: true in the client configuration during tests.
> 
> 
> 
> I believe we require node 20 so it should be good

I am also interested in the impact on the platform if any, but I believe it shouldn't right?

olivermrbl (approved):
LGTM!

sradevski:
@adrien2p no it should be fine, no worries :)

PR #13937, opened 2025-11-02
Inject config module in tests

peterlgh7:
## Summary

Injecting configModule in `moduleIntegrationTestRunner`

**Why**
[These docs](https://docs.medusajs.com/learn/fundamentals/modules/multiple-services) instruct to use `configModule.modules[BLOG_MODULE].options` in internal services to access options, which works fine, but not with the `moduleIntegrationTestRunner`, where the injected `configModule` is `undefined`.

**How**
Just adding `configModule` to the `injectedDependencies`

**Testing**
Added a module integration test for it and other basic expected injections

## Checklist

---

> [!NOTE]
> Inject CONFIG_MODULE into moduleIntegrationTestRunner and add integration tests with fixtures to verify dependency injection and options propagation.
> 
> - **Test Utils (`packages/medusa-test-utils`)**:
>   - Inject `ContainerRegistrationKeys.CONFIG_MODULE` into `moduleIntegrationTestRunner` via `injectedDependencies`, populated with `modulesConfig` to expose module options to services.
> - **Tests**:
>   - Add fixture module (`__fixtures__/test-module`) and internal service to validate dependency injection.
>   - Add integration tests (`__tests__/module-test-runner.spec.ts`) asserting injections on main and internal services, including `CONFIG_MODULE` and options propagation.
> - **Changeset**:
>   - Patch bump for `@medusajs/test-utils`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 47e41951a1fa2294a21ee49f425cbf3c57526e82. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/.changeset/silly-hats-unite.md
+++ b/.changeset/silly-hats-unite.md
@@ -1,0 +1,5 @@
+---
+"@medusajs/test-utils": patch
+---
+
+inject configModule in moduleIntegrationTestRunner
--- a/packages/medusa-test-utils/src/__fixtures__/test-module/index.ts
+++ b/packages/medusa-test-utils/src/__fixtures__/test-module/index.ts
@@ -1,0 +1,8 @@
+import { Module } from "@medusajs/framework/utils"
+import TestService from "./service"
+
+export const TEST_MODULE = "test"
+
+export default Module(TEST_MODULE, {
+  service: TestService,
+})
--- a/packages/medusa-test-utils/src/__fixtures__/test-module/service.ts
+++ b/packages/medusa-test-utils/src/__fixtures__/test-module/service.ts
@@ -1,0 +1,17 @@
+export default class TestService {
+  private dependencies: Record<string, any>
+  private options: any
+
+  constructor(dependencies: Record<string, any>, options: any) {
+    this.dependencies = dependencies
+    this.options = options
+  }
+
+  async getDependencies(): Promise<Record<string, any>> {
+    return this.dependencies
+  }
+
+  async getOptions(): Promise<any> {
+    return this.options
+  }
+}
--- a/packages/medusa-test-utils/src/__fixtures__/test-module/services/internal.ts
+++ b/packages/medusa-test-utils/src/__fixtures__/test-module/services/internal.ts
@@ -1,0 +1,11 @@
+export default class InternalService {
+  private dependencies: Record<string, any>
+
+  constructor(dependencies: Record<string, any>) {
+    this.dependencies = dependencies
+  }
+
+  async getDependencies(): Promise<Record<string, any>> {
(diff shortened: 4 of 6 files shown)


adrien2p:
I think it might be a doc issue, the module options are available as the third argument of the module constructor if I am correct. And the second argument is the module declaration.

if the config module needs to be available though, it can be added the the module dependencies and then it should be available in the cradle

peterlgh7:
let me check that @adrien2p

peterlgh7:
Just checked @adrien2p .
- The main module service's constructor receives `(dependencies, options, moduleDef)`
- The internal services only receive `(dependencies)`, which is why the docs give the `configModule` workaround

The issue is that `configModule` doesn't get injected in tests though, so the workaround is not usable in tests, unless one manually injects the `configModule` (which is what I'm doing now), but that's quite annoying.

What I'm really after here though is getting the options in internal services. Ideally I'd like to get those as second argument, as in the main module, but that's not the case, and the docs make it seem like that's either by design, or due to some limitation. If that's not the case, I could switch the PR to injecting options in internal services, which for my use case would be much cleaner: I don't need the full `configModule`.

adrien2p:
> Just checked @adrien2p .
> 
> - The main module service's constructor receives `(dependencies, options, moduleDef)`
> 
> - The internal services only receive `(dependencies)`, which is why the docs give the `configModule` workaround
> 
> 
> 
> The issue is that `configModule` doesn't get injected in tests though, so the workaround is not usable in tests, unless one manually injects the `configModule` (which is what I'm doing now), but that's quite annoying.
> 
> 
> 
> What I'm really after here though is getting the options in internal services. Ideally I'd like to get those as second argument, as in the main module, but that's not the case, and the docs make it seem like that's either by design, or due to some limitation. If that's not the case, I could switch the PR to injecting options in internal services, which for my use case would be much cleaner: I don't need the full `configModule`.

I see, in that case we could see if we can pass the same constructor signature to internal services as well, wdyt? That way Bose align all signatures, we maintain the current approach while also having a less havky approach?

Though, this pr still makes sense since the config module is a default deps of all modules, the runner should make it Available 👌

peterlgh7:
cool @adrien2p I'll merge this one and align the constructor injection in another PR

adrien2p:
Yes that’s sounds good, let me know if you need any help or review. You can just ping me 👌

peterlgh7:
@adrien2p just need approval here 😄

adrien2p (approved):
LGTM

adrien2p:
> @adrien2p just need approval here 😄

oups, done :D

PR #13938, opened 2025-11-02
Quote column names in indexes

peterlgh7:
## Summary

**What**

Quote column names on index DDL

**Why**

Generated migration is invalid if index contains a column with a name which is a reserved word in Postgres

Just quoting the column names

**Testing**
Unit tests

## Checklist

---


packages/core/utils/src/common/__tests__/create-psql-index-helper.spec.ts
```diff
@@ -53,7 +53,9 @@ describe("createPsqlIndexStatementHelper", function () {
     expect(indexStatement.expression).toEqual(
       `CREATE INDEX IF NOT EXISTS "${options.name}" ON "${
```
peterlgh7, line 54: I think we should refactor these tests to match against exact strings instead of constructing them from the test inputs. If you agree I can include it in this PR.

adrien2p (approved):
LGTM

PR #13940, opened 2025-11-03
chore(): Cleanup and organize deps

adrien2p:
RESOLVES CORE-1264

## Overview
This PR consolidates development dependencies, standardizes package scripts across the monorepo, and optimizes the CI/CD pipeline for better performance and reliability.

---

### Changes
- **Moved most dev dependencies** from individual packages to the root `package.json`
- Deduplicated common dependencies like `@swc/jest`, `@types/*`, `typescript`, `jest`, `vitest`, etc.
- Ensured all packages reference workspace versions for internal dependencies

### Benefits
- ✅ **Reduced duplication**: Single source of truth for tooling versions
- ✅ **Faster installs**: Fewer packages to resolve and install
- ✅ **Easier maintenance**: Update testing/build tools in one place
- ✅ **Smaller package.json files**: Cleaner individual package configurations
- ✅ **Better version consistency**: All packages use the same tool versions

### Build Scripts - Using `yarn run -T`
Updated all build-related scripts to use `yarn run -T` (top-level binaries):

```json
// Before
"build": "rimraf dist && tsc --build"

// After
"build": "yarn run -T rimraf dist && yarn run -T tsc --build"
```

**Binaries updated:**
- `tsc` → `yarn run -T tsc`
- `rimraf` → `yarn run -T rimraf`
- `tsup` → `yarn run -T tsup`
- `tsx` → `yarn run -T tsx`
- `tsc-alias` → `yarn run -T tsc-alias`

**Why:** Uses shared binaries from the root `node_modules`, reducing duplication and ensuring consistent tool versions.

### Test Scripts - Using Relative Binary Paths
Updated all test scripts to use relative paths to binaries instead of `yarn run -T`:

```json
// Before
"test": "jest --passWithNoTests --bail --forceExit"

// After
"test": "../../../node_modules/.bin/jest --passWithNoTests --bail --forceExit"
```

**Why:** Jest/Vitest need to run in the local package context to:
- Properly discover and use local `jest.config.js`
- Resolve dependencies correctly
- Support ESM modules with proper transformation

### Removed Explicit Test Path Patterns
Removed hardcoded test path patterns from test scripts (50 packages affected):

```json
// Before - fails if no tests exist, even with --passWithNoTests
"test": "jest --passWithNoTests -- src/**/__tests__/**/*.ts"

// After - respects --passWithNoTests flag
"test": "../../../node_modules/.bin/jest --passWithNoTests -- src"
```

**Why:** Shell glob expansion happens before jest runs, causing failures for packages without tests. Jest's default test discovery works correctly with `--passWithNoTests`.

### Jest Configuration Updates
Updated `define_jest_config.js`:

1. **Added ESM module transformation support:**
   ```js
   transformIgnorePatterns: ["node_modules/(?!(until-async|msw)/)"]
   ```
   Allows transformation of ESM-only packages like `until-async` and `msw`.

2. **Excluded fixture directories:**
   ```js
   testPathIgnorePatterns: [`dist/`, `node_modules/`, `__fixtures__/`, `__mocks__/`]
   ```
   Prevents jest from treating fixture files as tests.

---

### 1. Enhanced Dependency Caching
**Changed:** Updated `.github/actions/cache-deps/action.yml`

```yaml
# Before - only cached yarn cache
path: |
  .yarn/cache

# After - caches installed dependencies too
path: |
  .yarn/cache
  node_modules
  .pnp.*
```

**Added conditional install:**
```yaml
- run: yarn install --immutable --inline-builds
  if: steps.cache.outputs.cache-hit != 'true'
```

**Benefits:**
- ⚡ **~2-3 minutes saved per job** by not reinstalling node_modules
- ✅ Exact cache hits skip `yarn install` entirely
- ✅ Partial cache hits (different yarn.lock) only update changed packages

### 2. Build Artifacts Management
**Added artifact upload in setup job:**
```yaml
- name: Upload build artifacts
  uses: actions/upload-artifact@v4
  with:
    name: build-artifacts
    path: |
      packages/**/dist
      integration-tests/**/dist  # ← Added integration-tests
      !packages/**/node_modules
```

**Why:** Ensures both `packages/**/dist` and `integration-tests/**/dist` are available for all test jobs.

### 3. Skip Unnecessary Builds
**Added `skip-build` parameter:**
```yaml
- name: Install dependencies
  uses: ./.github/actions/cache-deps
  with:
    skip-build: 'true'  # ← Test jobs don't need to rebuild
```

**Why:** Test jobs download pre-built artifacts from the setup job, saving times per job.

### 4. Fix test scripts

Remove `-- path` to prevent CI test arguments from being wrongly assimilated to path matching

### Reliability Improvements
- ✅ Fixed "No matches found" errors for packages without tests
- ✅ Fixed missing build artifacts in integration tests
- ✅ Proper ESM module transformation for modern dependencies
- ✅ Consistent tool versions across all packages


packages/core/workflows-sdk/package.json
```diff
@@ -28,13 +28,7 @@
...
   "devDependencies": {
     "@medusajs/types": "2.11.2",
-    "@swc/core": "^1.7.28",
-    "@swc/jest": "^0.2.36",
-    "expect-type": "^0.20.0",
-    "jest": "^29.7.0",
-    "rimraf": "^5.0.1",
-    "typescript": "^5.6.2",
-    "zod": "3.25.76"
+    "express": "^4.21.0"
```
riqwan: q: why does this need express?

adrien2p: good catch, I ve put it in dev because we had it in peers but in fact there is no need for it 👍

adrien2p: I ll run a little investigation to check if we have other deps that are there but not being used (probably legacy)

adrien2p: I ve cleaned up a few more packages with similar issues 👍 thanks man

carlos-r-l-rodrigues (approved):
Tested the snapshot and it LGTM.

adrien2p:
@olivermrbl let me know when we are ready to merge this, if we are 💪

packages/cli/medusa-cli/package.json
```diff
@@ -25,20 +25,12 @@
     "node": ">=20"
   },
   "scripts": {
-    "test": "jest --passWithNoTests src",
-    "build": "rimraf dist && tsc --build",
-    "watch": "tsc --watch"
+    "test": "../../../node_modules/.bin/jest --passWithNoTests src",
```
olivermrbl, line 28: suggestion(non-blocking): I think we can use `npx` here; just a bit cleaner IMO. It should look for the binary in local packages (including those higher in the folder hierarchy)

adrien2p: I’ll check, but yarn run should have done the same but it didn’t really work because now the package does not have the dev deps anymore so it don’t try to find it

olivermrbl: Ah okay got it. Fine to leave it as is; this was merely a cosmetic suggestion

olivermrbl (approved):

packages/cli/medusa-cli/package.json
adrien2p: Yeah I tried as best to not have that too, but even yarn run -T would not work because it runs it from top level and can manage the config. I think we can try to find a solution at some point 👌but right now it is the trade of to minimise dev deps conflicts and simplify management

PR #13975, opened 2025-11-05
fix(currency,payment,pricing,region,order,store,cart,core-flows,medusa,utils): repo wide currency_code normalization

### LINKED ISSUE #13956: [Bug]: currency_code validation on create product route does not enforce lowercase
### Package.json file

```JSON
{
  "name": "medusa-b2b-starter",
  "version": "0.0.1",
  "description": "A starter for Medusa B2B projects.",
  "license": "MIT",
  "keywords": [
    "sqlite",
    "postgres",
    "typescript",
    "ecommerce",
    "headless",
    "medusa"
  ],
  "scripts": {
    "build": "medusa build",
    "start": "medusa start",
    "dev": "medusa develop",
  },
  "dependencies": {
    "@medusajs/admin-sdk": "2.11.1",
    "@medusajs/cli": "2.11.1",
    "@medusajs/framework": "2.11.1",
    "@medusajs/js-sdk": "2.11.1",
    "@medusajs/medusa": "2.11.1"
  },
  "devDependencies": {
    "@medusajs/test-utils": "2.11.1",
    "@swc/core": "^1.5.7",
    "@swc/jest": "^0.2.36",
    "@types/jest": "^29.5.14",
    "@types/node": "^20.0.0",
    "@types/react": "^18.3.2",
    "jest": "^29.7.0",
    "prop-types": "^15.8.1",
    "react": "^18.2.0",
    "react-dom": "^18.2.0",
    "ts-node": "^10.9.2",
    "typescript": "^5.6.2",
    "vite": "^5.2.11"
  },
  "engines": {
    "node": ">=20"
  }
}
```

### Node.js version

v21.7.2

### Database and its version

PostgreSQL 17

macOS 15.6.1

### Browser name

_No response_

### What happended?

When creating a product through the Admin API, passing the currency_code in uppercase still succeeds, but breaks everything else because it also gets stored in the DB as uppercase, but all other workflows match it case-sensitive.

### Expected behavior

Both passing lowercase and uppercase currency_code should succeed. enforce lowercase by transforming the input to lowercase.

### Actual behavior

Passing uppercase currency_code succeeds, but stores it as uppercase in the DB. Workflows match case-sensitive and expects lowercase, which breaks everything.

### Link to reproduction repo

Running npx create-medusa-app@latest my-medusa-store

NicolasGorga:
## Summary

Introduces new util function `normalizeCurrencyCode` to be used across the project were currency codes are stored and used for filters. Replaces existent isolated normalizations with the function and adds the missing normalization when handling currency codes in the **PricingModule**.

The objective is to have consistency across currency code handling and avoid potential bugs for storing the currency codes as uppercase in some places and lowercase in others.

New `normalizeCurrencyCode` function acts as the centralized place to normalize the currency codes and should be used whenever storing and or searching a currency code.

Added unit tests for the function + updated existent integration tests.

---

## Examples

```ts
// Example usage
```

---

## Checklist

---

## Additional Context

fixes #13956 

---

> [!NOTE]
> Centralizes currency code lowercasing via `normalizeCurrencyCode`, applies it across services and queries, adds a DB migration to lowercase existing data, and updates tests to expect lowercase codes.
> 
> - **Utils**:
>   - Add `normalizeCurrencyCode` (lowercases codes) with unit tests and export.
> - **Data Migration**:
>   - New script `packages/medusa/src/migration-scripts/migrate-normalize-currency-codes-normalization.ts` lowercases `currency_code` in `cart`, `payment_collection`, `payment_session`, `payment`, `order`, `order_transaction`, `price`, `region`, `store_currency`.
> - **Modules/Services**:
>   - **Pricing**: Normalize on read/write (repository `calculatePrices`, service price hashing/upsert/constraints), propagate lowercase in price lists/sets/preferences.
>   - **Payment**: Normalize on create/update of collections/sessions/payments and provider interactions.
>   - **Order**: Normalize `currency_code` on order creation and order transactions.
>   - **Cart**: Normalize on cart create/update.
>   - **Region**: Normalize input `currency_code`; trim/name handling unchanged.
>   - **Store**: Normalize `supported_currencies[].currency_code` on create/update.
>   - **Core Flows**: Import helper uses `normalizeCurrencyCode` for `variant_price_*` parsing.
>   - **Currency**: Normalize `code` filter in list/retrieve.
> - **Tests**:
>   - Update integration tests to expect lowercase `currency_code` throughout (pricing, payment, indexing/search).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit b07bb4efceb499d5db7a4134f81eeb5d7622ad54. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


packages/core/utils/src/common/normalize-currency-code.ts
```diff
@@ -0,0 +1,12 @@
+import { isString } from "./is-string"
+
+/**
+ * Normalizes `currencyCode` by transforming it to lowercase
+ */
+export function normalizeCurrencyCode(currencyCode: unknown) {
+    if (!isString(currencyCode)) {
+        return
```
adrien2p: suggestion: we might need to return the input directly here? also, we always expet a string so I would type the input as string and here if at runtime a non string is provided then we might probably want to throw?

adrien2p (commented):
I believe we are missing a migration script to update all currency code in the database that are not lower cased no?

NicolasGorga:
> I believe we are missing a migration script to update all currency code in the database that are not lower cased no?

Oh forgot about that, will include it.

packages/core/utils/src/common/normalize-currency-code.ts
NicolasGorga: I think we could go ahead and throw as you mention if the type is not the expected one. Will change that

NicolasGorga:
@adrien2p Hey could you check the migration script? We basically need to migrate `price` table, so variant and price list prices. I searched the current workflows we have and didn't find a straightforward approach to use them in this context, so opted for using knex inside a transaction (since there is no compensation function in the step i created), wyt?

packages/medusa/src/migration-scripts/migrate-normalize-currency-codes-normalization.ts
```diff
@@ -0,0 +1,30 @@
...
+    'normalize-currency-codes',
+    () => {
+        updatePricesStep()
+
+        return new WorkflowResponse(void 0)
+    }
+)
+
+export default async function migrateNormalizeCurrencyCodes({ container }: ExecArgs) {
+    await normalizeCurrencyCodesWorkflow(container).run()
```
adrien2p: suggestion: I am not sure we even need to go through a workflow here, we could exec knex directly right?

packages/modules/pricing/integration-tests/__tests__/services/pricing-module/price-set.spec.ts
```diff
@@ -426,8 +426,8 @@ moduleIntegrationTestRunner<IPricingModuleService>({
             priceSetBefore.id,
             {
               prices: [
-                { amount: 100, currency_code: "USD" },
-                { amount: 200, currency_code: "EUR" },
+                { amount: 100, currency_code:"usd" },
```
adrien2p, line 429: info: you still have a lot of formatting issues in the pr's, we might need to look at it, the risk is that everytime for example I go into the same file I ll end up with hundreds of unexpected changes unrelated to my work because my editor will format everything 😄

packages/modules/pricing/src/services/pricing-module.ts
```diff
@@ -1815,7 +1820,7 @@ const hashPrice = (
   const parts: string[] = []
 
   if ("currency_code" in price) {
-    parts.push(`cc:${price.currency_code ?? ""}`)
+    parts.push(`cc:${normalizeCurrencyCode(price.currency_code) ?? ""}`)
```
adrien2p:
```suggestion
    parts.push(`cc:${normalizeCurrencyCode(price.currency_code ?? "")}`)
```

adrien2p: otherwise you might have an unexpected throw in case currency is not present

packages/modules/pricing/src/services/pricing-module.ts
```diff
@@ -1869,7 +1874,7 @@ const buildPreNormalizationPriceConstraintsFromData = (
     const constraint: any = {}
 
     if (price.currency_code !== undefined) {
-      constraint.currency_code = price.currency_code
+      constraint.currency_code = normalizeCurrencyCode(price.currency_code)
```
adrien2p, line 1877: Q: here we are checking if different than undefined, but it can be null, since null seems to ba an authorized value, the normalizer would throw no?

packages/modules/region/src/services/region-module.ts
```diff
@@ -307,7 +308,7 @@ export default class RegionModuleService
     return regions.map((region) =>
       removeUndefined({
         ...region,
-        currency_code: region.currency_code?.toLowerCase(),
+        currency_code: normalizeCurrencyCode(region.currency_code),
```
adrien2p: Q: here it seems that it was possible that currency_code is not defined (nullish coalescing above), in that case with the new util it would throw, should we handle that?

adrien2p (commented):
Should we also check other modules that are using currency_code? this pr primarely focuse on the pricing and it might be the objective but just want to throw it outloud

NicolasGorga:
> Should we also check other modules that are using currency_code? this pr primarely focuse on the pricing and it might be the objective but just want to throw it outloud

I checked all other modules, and the problematic one is the Pricing module. For other usages, what I did is replace the direct manipulation with the new common dedicated util `normalizeCurrencyCode`.

packages/medusa/src/migration-scripts/migrate-normalize-currency-codes-normalization.ts
NicolasGorga: Yes, you are right. I am not using a compensation function inside the workflow, since the rollback for the transaction will take care of that if anything goes wrong. Will change it!

packages/modules/pricing/integration-tests/__tests__/services/pricing-module/price-set.spec.ts
NicolasGorga: Yes, definitely, I haven't gotten to making it work on save, so I have to remember to do it manually every time. Ever since you raised it, I've been trying to remember to do it, but it is really error-prone. Let's add it to the list of things to see in our huddle :)

packages/modules/pricing/src/services/pricing-module.ts
NicolasGorga: You are absolutely right

packages/modules/pricing/src/services/pricing-module.ts
NicolasGorga: Corrected

adrien2p:
> > Should we also check other modules that are using currency_code? this pr primarely focuse on the pricing and it might be the objective but just want to throw it outloud
> 
> I checked all other modules, and the problematic one is the Pricing module. For other usages, what I did is replace the direct manipulation with the new common dedicated util `normalizeCurrencyCode`.

I was more thinking of cart, order, payment etc, they dont need that or it was in a separate pr?

NicolasGorga:
> > > Should we also check other modules that are using currency_code? this pr primarely focuse on the pricing and it might be the objective but just want to throw it outloud
> > 
> > 
> > I checked all other modules, and the problematic one is the Pricing module. For other usages, what I did is replace the direct manipulation with the new common dedicated util `normalizeCurrencyCode`.
> 
> I was more thinking of cart, order, payment etc, they dont need that or it was in a separate pr?

You are right, I think i didn't miss anything now. Also updated data migration script.

.changeset/free-rings-happen.md
```diff
@@ -0,0 +1,14 @@
+---
+"@medusajs/currency": patch
```
adrien2p, line 2: todo: we can remove one of the changeset

adrien2p (approved):
In that case we should be good 💪

NicolasGorga:
@adrien2p With the latest update i did to the test, this would now be breaking right? Since existing filters using `index.graph` that have an uppercased `currency_code` will no longer return results they previously did.

The thing is tricky, since i know we can hook the normalization probably in `PostgresProvider.quey()` but if we target the `currency_code` field, it means we are potentially normalizing custom modules other than the core `Pricing` module. Whatever we do is a tradeoff, wyt?

adrien2p:
> @adrien2p With the latest update i did to the test, this would now be breaking right? Since existing filters using `index.graph` that have an uppercased `currency_code` will no longer return results they previously did.
> 
> The thing is tricky, since i know we can hook the normalization probably in `PostgresProvider.quey()` but if we target the `currency_code` field, it means we are potentially normalizing custom modules other than the core `Pricing` module. Whatever we do is a tradeoff, wyt?

I can see two options:
- We tell people that they might need to trigger a full resync of the index
- We write a migration to update the index data for specific entity name

wdyt?

shahednasser (approved):

PR #13979, opened 2025-11-05
fix(core-flows): update SendNotificationsStepInput

leobenzol:
## Summary

**What** 

Updated the SendNotificationsStepInput type with the missing fields from CreateNotificationDTO

**Why** 

2.11.2 made the "template" property optional, which causes type errors since the step still expects a non-null template.

## Checklist


packages/core/core-flows/src/notification/steps/send-notifications.ts
leobenzol, line 8: question: should it be simply `SendNotificationsStepInput = CreateNotificationDTO[]`?

I understand that decoupling interfaces is good practice, especially between layers of the stack, but for a simple step that just forwards to the module(and will probably never change) it feels like extra code debt for no gain.

willbouch (approved):

packages/core/core-flows/src/notification/steps/send-notifications.ts
willbouch: I'm good with what you have

PR #13983, opened 2025-11-06
chore(): Remove barrel files in favor of imports usage in core-flows

adrien2p:
RESOLVES CORE-1273


willbouch (commented):
Hey ! Could you expand on the reason ?

adrien2p:
> Hey ! Could you expand on the reason ?

For now it is an experiment, but mainly reducing circular deps references preventing moving to am later (one of the main pain point we had in the past), introducing native imports alias from package json to replace relative cumbersome import, reducing number of files to maintain, size, and load time. Barrel files makes sense in some cases but not all 👌

PR #14010, opened 2025-11-07
fix(core-flows): create reservations on draft order conversion to regular order

### LINKED ISSUE #13773: [Bug]: Draft Order Edit allocates inventory even for draft orders (unexpected behavior)
### Package.json file

```JSON
not relevant
```

### Node.js version

not relevant

### Database and its version

not relevant

not relevant

### Browser name

_No response_

### What happended?

When creating a draft order, no inventory allocation is expected — and this is correct.
However, when performing an edit session on an existing draft order (using the beginEdit / updateItem / addItems / confirmEdit flow), inventory reservations are unexpectedly created for the updated or added items.

This causes draft orders — which should not allocate inventory — to reserve stock in the inventory service after an edit confirmation.

### Expected behavior

Create a draft order with items x, y, z.

Begin an edit session and add new items a and b.
In the same edit session, also update "variant x" quantity

No inventory reservations should ever be created for draft orders, since they are not finalized or confirmed orders.
Draft orders are typically used in POS or admin-side workflows before a sale is completed.

### Actual behavior

After confirmEdit, reservations are created only for the items modified during the edit — in this example:
items a, b, and x
items y, z remain unreserved.

This behavior happens because confirmDraftOrderEditWorkflow internally calls reserveInventoryStep.

### Link to reproduction repo

not relevant

NicolasGorga:
## Summary

Avoid creating reservations when draft order edits are confirmed and rather, create them when the draft order is converted into a regular order.

While the order is a draft, creating reservations would potentially block inventory for regular order requests, when the draft represents a non materialized state of a purchase that might never be completed or at a latter point in time.

Removed the reservation creations inside of `confirmDraftOrderEditWorkflow` and instead do it inside `convertDraftOrderWorkflow`

Added integration tests.

---

## Examples

```ts
// Example usage
```

---

## Checklist

---

## Additional Context

fixes #13773 
closes SUP-2523


.changeset/fancy-banks-move.md
```diff
@@ -0,0 +1,5 @@
+---
+"@medusajs/core-flows": minor
```
fPolic: woudl consider this as a patch/bugfix rather than a minor change

integration-tests/http/__tests__/draft-order/admin/draft-order.spec.ts
```diff
@@ -375,7 +538,7 @@ medusaIntegrationTestRunner({
         ).data.product
       })
 
-      it("should manage reservations on order edit", async () => {
+      it("should not create reservations on draft order edit confirmation", async () => {
```
fPolic, line 546: should we remove this test? this is tested as a part of the expected scenario above

packages/core/core-flows/src/draft-order/workflows/convert-draft-order.ts
```diff
@@ -119,6 +122,37 @@ export const convertDraftOrderWorkflow = createWorkflow(
...
+    const orderItems = useRemoteQueryStep({
+      entry_point: "order",
+      fields: requiredOrderFieldsForInventoryConfirmation,
+      variables: { id: input.id },
+      list: false,
+      throw_if_key_not_found: true,
+    }).config({ name: "order-items-query" })
+
+    const { variants, items } = transform({ orderItems }, ({ orderItems }) => {
+      const items: any[] = orderItems.items
```
fPolic: nit: since we are transforming alrady, should we pass only item props that `prepareConfirmInventoryInput` needs - id, quantity, variant_id

fPolic (commented):
lgtm! 
we just need ✅ on tests

integration-tests/http/__tests__/draft-order/admin/draft-order.spec.ts
NicolasGorga: I think we still need the test I added, as the already existing one in order-edit spec handles edits to a normal order, not a draft one no? In my tests i want to guarantee that for drafts, confirming inventory actions don't create reservations as this should only happen for normal orders (they will be generated if/when it is converted to a normal order)

packages/core/core-flows/src/draft-order/workflows/convert-draft-order.ts
NicolasGorga: done

olivermrbl (approved):
LGTM

integration-tests/http/__tests__/draft-order/admin/draft-order.spec.ts
fPolic: Yeah, but this is something `confirmInventory` tests should ensure then. My suggestion was to remove it since it is implicitly tested in [the other test](https://github.com/medusajs/medusa/pull/14010/files#diff-9ea002eaf89a0f7abc03ec021b1eb51305ef9708aada57989db0a2a42bc69466R387). Not a blocker though.

integration-tests/http/__tests__/draft-order/admin/draft-order.spec.ts
```diff
@@ -248,6 +336,81 @@ medusaIntegrationTestRunner({
...
+                variant_id: product.variants.find((v) => v.title === "M shirt")
+                  .id,
+                quantity: 1,
+              },
+            ],
+          },
+          adminHeaders
+        )
+
+        let reservations = (await api.get(`/admin/reservations`, adminHeaders))
```
fPolic, line 378: this `reservations` are never used, they are refetched after calling `/confirm`. Did we want to check that the length here is 0 as well?

fPolic (approved):
LGTM

integration-tests/http/__tests__/draft-order/admin/draft-order.spec.ts
NicolasGorga: Exactly, missed the assert. Included!

PR #14024, opened 2025-11-10
chore(order): custom display id

adrien2p:
RESOLVES FRMW-2993

**What**
Users can now provide a custom display id generator option to the order module. This generator will be used upon order creation to fill the `custom_display_id` of an order.

example:
```ts
// medusa-config.ts
defineConfig({
  // ...
  modules: [{
    resolve: Modules.ORDER,
    options: {
      generateCustomDisplayId: async function (order: OrderTypes.CreateOrderDTO, sharedContext: Context) {
        // Return your custom display id
      }
    }
  }]
})
```


packages/modules/order/src/migrations/Migration20251113084802.ts
```diff
@@ -0,0 +1,131 @@
+import { Migration } from "@mikro-orm/migrations"
+
+export class Migration20251113084802 extends Migration {
+  override async up(): Promise<void> {
+    this.addSql(
+      `alter table if exists "order" add column if not exists "custom_display_id" text null;`
+    )
+    this.addSql(
+      `CREATE INDEX IF NOT EXISTS "IDX_order_custom_display_id" ON "order" ("custom_display_id") WHERE deleted_at IS NULL;`
```
fPolic: q: should we enforce `unique` here?

fPolic (approved):
LGTM!

Should we display custom_id somewhere in Admin?

packages/modules/order/src/migrations/Migration20251113084802.ts
adrien2p: I wanted to do it and saw that display id wasn't uniq either. So I aligned to that. Wdyt?

adrien2p:
> LGTM!
> 
> Should we display custom_id somewhere in Admin?

I wasn't sure in case it stays an empty column for most projects. But they can still add it as a displayable column in the custom tables no?

packages/modules/order/src/migrations/Migration20251113084802.ts
fPolic: I think we should enforce uniqueness here since it is an id column.

fPolic:
> I wasn't sure in case it stays an empty column for most projects.

I was thinking on the details page maybe, but good point

> But they can still add it as a displayable column in the custom tables no?

yes

packages/modules/order/src/migrations/Migration20251113084802.ts
adrien2p: done ✅

adrien2p:
cc @olivermrbl when you have some bandwidth if you could have a look at this one please 👍

olivermrbl:
Will look now!

olivermrbl (approved):
Nice work @adrien2p

willbouch (approved):
:shipit:

PR #14031, opened 2025-11-11
feat(medusa,types,file-local,file-s3): delete file automatically when possible upon image deletion

### LINKED ISSUE #8548: Deleting an image does not trigger the delete method of AbstractFileProviderService

NicolasGorga:
## Summary

Adds `deleteByUrl` method to the `IFileProvider` interface to allow deleting a file by its url. This is necessary since there is currently no link between our `image` model and its corresponding file representation in the file provider. This includes the implementation of said method in the `S3FileService` and `LocalFileService`.

A core subscriber listening to the `ProductEvents.PRODUCT_IMAGE_DELETED` resolves the currently installed file provider (which could be different from the one the image file was created with, in case it was swapped between the moment it was created and its deletion) and tries to call the method if it is implemented, to delete the corresponding file based on the image url we store.

Messages that are logged are:

- warn message if the `deleteByUrl` method is not implemented in the currently installed file provider
- error message if the method execution throws. For cases where the image was created with a different provider, this would be expected
- info indicating the file was deleted successfully if no problem arises

For the problematic logs, the image url is included to make it obvious to what resource and provider this corresponds to, so any further action can be taken by the user to delete it.

*Please provide answer here*

File deletion isn't performed and "fails" silently from a user's perspective when an image is deleted.

*Please provide answer here*

Added a new `deleteByUrl` method to the file provider interface and we try to call it with a core subscriber when an image deletion event is fired.

*Please provide answer here*

Integration tests

*Please provide answer here*

---

## Examples

```ts
// Example usage
```

---

## Checklist

- [] I have verified the code works as intended locally

---

## Additional Context

fixes #8548 
closes CORE-708

### DIFF
--- a/packages/modules/providers/file-local/src/services/local-file.ts
+++ b/packages/modules/providers/file-local/src/services/local-file.ts
@@ -105,6 +105,11 @@ export class LocalFileService extends AbstractFileProviderService {
     return
   }
 
+  async deleteByUrl(url: string): Promise<void> {
+    const fileKey = this.getFileKey(url)
+    return this.delete({ fileKey })
+  }
+
   async getDownloadStream(
     file: FileTypes.ProviderGetFileDTO
   ): Promise<Readable> {
@@ -175,6 +180,16 @@ export class LocalFileService extends AbstractFileProviderService {
     const baseUrl = new URL(this.backendUrl_)
     baseUrl.pathname = path.join(baseUrl.pathname, fileKey)
     return baseUrl.href
+  }
+
+  private getFileKey = (url: string) => {
+    const parsedUrl = new URL(url)
+    const basePath = new URL(this.backendUrl_).pathname
+    const fullPath = parsedUrl.pathname
+
+    const fileKeyPath = fullPath.startsWith(basePath) ? fullPath.slice(basePath.length) : fullPath
+    const fileKey = fileKeyPath.startsWith('/') ? fileKeyPath.slice(1) : fileKeyPath
+    return fileKey
   }
 
   private async ensureDirExists(baseDir: string, dirPath: string) {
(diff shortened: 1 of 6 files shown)


packages/modules/providers/file-local/src/services/local-file.ts
```diff
@@ -177,6 +182,16 @@ export class LocalFileService extends AbstractFileProviderService {
     return baseUrl.href
   }
 
+  private getFileKey = (url: string) => {
```
NicolasGorga: I guess it would be okay to even simplify this more like:

```js
return url.slice(url.lastIndexOf('/') + 1)
```

NicolasGorga:
I need yet to resolve how to get the `url` from the DB, since the record is already hard deleted by the time the subscriber tries to query it. 

I feel like the cleanest alternative is to make images soft deletable. We can also leverage the core subscriber introduced here to replicate the old behavior and hard delete the image after we get the `url` needed to delete the file.

NicolasGorga:
This PR is on hold until we finish CORE-1280

PR #14037, opened 2025-11-12
fix(): Identify step that should require a save on checkpoint

adrien2p:
RESOLVES CORE-1278

**What**
Instead of purely relying on a step being async to prevent or force a save during a checkpoint, we now rely on more broader configuration


adrien2p:
@NicolasGorga do you think you get this one for a tour?

carlos-r-l-rodrigues (approved):

adrien2p:
cc @olivermrbl tested with @NicolasGorga as well and it fixes the issue for what it is meant to fix at least 🚀

carlos-r-l-rodrigues (approved):

PR #14041, opened 2025-11-12
chore(): Remove lodash.set entirely

adrien2p:
**What**
Remove lodash.set usage entirely

### DIFF
--- a/packages/admin/dashboard/src/components/data-grid/hooks/use-data-grid-form-handlers.tsx
+++ b/packages/admin/dashboard/src/components/data-grid/hooks/use-data-grid-form-handlers.tsx
@@ -151,6 +151,7 @@ function setValue<
       if (index === field.split(".").length - 1) {
         curr[key] = newValue
       }
+      curr[key] ??= {}
       return curr[key]
     }, currentValues)
     return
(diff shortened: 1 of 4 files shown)


carlos-r-l-rodrigues (approved):

packages/admin/dashboard/src/components/data-grid/hooks/use-data-grid-form-handlers.tsx
```diff
@@ -187,11 +192,18 @@ function setValueToggleableNumber(
       : newValue.checked
     : determineChecked(quantity)
 
-  set(currentValues, field, {
```
adrien2p, line 190: @olivermrbl was that expected? it seems that we are shallow copying the entire object to be assigned to the given path alongside the other props

olivermrbl: I really don't know, was I to git blame? 😅

adrien2p: 😂 I dont know, since the person is not part of the company anymore I fallback to you in case you have an idea

olivermrbl: Haha got it, feel free to update this! 😄

PR #14054, opened 2025-11-13
feat(): allow linking a subset of an option's values to the product

willbouch:
This PR expands on the product options redesign and allows linking only a subset of the option values to the products. For example, if I have an option called `Size` with possible values `XS`, `S`, `M`, `L`, `XL` but I have a t-shirt that can only be sold in `XS` and `S`, then I can link only those 2 option values to the product. This is possible using another pivot table between the `product_product_option` table and the `product_option_value` table. 

https://github.com/user-attachments/assets/984e0061-0c4f-4eae-9af8-9a52a39ecb7a


imash96:
@olivermrbl  Will this feature be released in the next update?

PR #14068, opened 2025-11-16
feat(notification): expose providers

leobenzol:
## Summary

**What** 

Makes notification providers available to query.graph, also exposes the getProviderForChannel method for convenience

**Why** 

There isn't currently a way to retrieve notification providers, which is necessary to implement features or retrieve data that are provider specific. For example I'm building a plugin to have a friendly UI interface for wiring up notifications and to be able to retrieve/store generic templates you need the list of providers to choose from.

**How**

The NotificationProvider model was added to the notification's module MedusaService, which is exactly what the [fullfilment](https://github.com/medusajs/medusa/blob/76fd6d14387374336263a22e71f9d53bd9c45998/packages/modules/fulfillment/src/services/fulfillment-module-service.ts#L91) and [tax](https://github.com/medusajs/medusa/blob/76fd6d14387374336263a22e71f9d53bd9c45998/packages/modules/tax/src/services/tax-module-service.ts#L46) modules do to make providers queryable.

## Checklist

### DIFF
--- a/.changeset/dry-frogs-scream.md
+++ b/.changeset/dry-frogs-scream.md
@@ -1,0 +1,5 @@
+---
+"@medusajs/notification": patch
+---
+
+feat(notification): expose providers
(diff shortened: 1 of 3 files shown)


leobenzol:
I went on a different route and don't need access to specific providers anymore. I'll reopen if I see someone or myself need it in the future

PR #14100, opened 2025-11-20
fix(medusa): Correct MedusaRequest query type

NicolasGorga:
## Summary

Updates `MedusaRequest` generics in /store/currencies get endpoints.

To align the body and query types to what is actually expected.

Corrected `MedusaRequest` generics adding undefined as the first type parameter so the already passed type is correctly mapped to `query` instead of `body`.

---

## Examples

```ts
// Example usage
```

---

## Checklist

---

## Additional Context

closes ENTSUP-279

### DIFF
--- a/.changeset/yellow-wings-design.md
+++ b/.changeset/yellow-wings-design.md
@@ -1,0 +1,5 @@
+---
+"@medusajs/medusa": patch
+---
+
+fix(medusa): Correct MedusaRequest generics in /store/currencies endpoint
--- a/.changeset/yellow-wings-design.md
+++ b/.changeset/yellow-wings-design.md
@@ -2,4 +2,4 @@
 "@medusajs/medusa": patch
 ---
 
-fix(medusa): Correct MedusaRequest generics in /store/currencies endpoint
+fix(medusa): Correct MedusaRequest query type
--- a/packages/core/framework/src/http/__tests__/validate-body.spec.ts
+++ b/packages/core/framework/src/http/__tests__/validate-body.spec.ts
@@ -21,7 +21,7 @@ describe("validateAndTransformBody", () => {
       body: {
         additional_data: {},
       },
-    } as MedusaRequest
+    } as unknown as MedusaRequest
 
     const mockResponse = {} as MedusaResponse
     const nextFunction = jest.fn()
@@ -57,7 +57,7 @@ describe("validateAndTransformBody", () => {
     let mockRequest = {
       query: {},
       body: {},
-    } as MedusaRequest
+    } as unknown as MedusaRequest
 
     const mockResponse = {} as MedusaResponse
     const nextFunction = jest.fn()
@@ -90,7 +90,7 @@ describe("validateAndTransformBody", () => {
       body: {
         additional_data: {},
       },
-    } as MedusaRequest
+    } as unknown as MedusaRequest
 
     const mockResponse = {} as MedusaResponse
     const nextFunction = jest.fn()
--- a/packages/core/framework/src/http/types.ts
+++ b/packages/core/framework/src/http/types.ts
@@ -218,7 +218,7 @@ export interface AuthenticatedMedusaRequest<
 
 export interface MedusaStoreRequest<
   Body = unknown,
-  QueryFields = Record<string, unknown>
+  QueryFields extends Record<string, unknown> = Record<string, unknown>
 > extends MedusaRequest<Body, QueryFields> {
   auth_context?: AuthContext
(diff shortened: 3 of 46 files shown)


olivermrbl (commented):
pretty sure we have done this all over the place; are you sure this is correct? if so, maybe we need to update more places

NicolasGorga:
> pretty sure we have done this all over the place; are you sure this is correct? if so, maybe we need to update more places

I'm positive. Included fixes to all the incorrect usages.

adrien2p (commented):
I think it is fine, I am just wondering if it would make sense to do something like this

```ts
export interface AuthenticatedMedusaRequest<
  BodyOrQueryFields = unknown,
  QueryFields extends Record<string, unknown> = {}
> extends MedusaRequest<
    {} extends QueryFields ? undefined : BodyOrQueryFields,
    {} extends QueryFields
      ? BodyOrQueryFields & Record<string, unknown>
      : QueryFields
  > {
  auth_context: AuthContext
  publishable_key_context?: PublishableKeyContext
}
```

```ts
export interface MedusaRequest<
  Body = unknown,
  QueryFields extends Record<string, unknown> = {}
> extends Request<
    { [key: string]: string },
    any,
    {} extends QueryFields ? Body : undefined
  > {
  validatedBody: {} extends QueryFields ? undefined : Body
  validatedQuery: {} extends QueryFields
    ? RequestQueryFields
    : RequestQueryFields & QueryFields
//...
```

Here we would allow for 
- single template args -> query fields
- both -> body, fields

Let me know if that makes sense guys (cc @olivermrbl)

NicolasGorga:
> I think it is fine, I am just wondering if it would make sense to do something like this
> 
> ```ts
> export interface AuthenticatedMedusaRequest<
>   BodyOrQueryFields = unknown,
>   QueryFields extends Record<string, unknown> = {}
> > extends MedusaRequest<
>     {} extends QueryFields ? undefined : BodyOrQueryFields,
>     {} extends QueryFields
>       ? BodyOrQueryFields & Record<string, unknown>
>       : QueryFields
>   > {
>   auth_context: AuthContext
>   publishable_key_context?: PublishableKeyContext
> }
> ```
> 
> ```ts
> export interface MedusaRequest<
>   Body = unknown,
>   QueryFields extends Record<string, unknown> = {}
> > extends Request<
>     { [key: string]: string },
>     any,
>     {} extends QueryFields ? Body : undefined
>   > {
>   validatedBody: {} extends QueryFields ? undefined : Body
>   validatedQuery: {} extends QueryFields
>     ? RequestQueryFields
>     : RequestQueryFields & QueryFields
> //...
> ```
> 
> Here we would allow for
> 
> * single template args -> query fields
> * both -> body, fields
> 
> Let me know if that makes sense guys (cc @olivermrbl)

I like this since I think users might not consistently understand that they should put `undefined` as the first type parameter for requests without a body. Additionally, it is more convenient not to have to do this in these cases and simply specify the query parameter type. Let me know wyt Oli, and I'll proceed with the update.

NicolasGorga:
Hey @adrien2p I've updated the types with your suggestion, with a small correction on the `MedusaRequest` type if you could check?

Also, if you could help me know how we could handle the following error when building in `packages/core/framework/src`

```bash
src/http/utils/validate-body.ts:31:7 - error TS2322: Type 'ZodRawShape' is not assignable to type 'undefined'.

31       req.validatedBody = await zodValidator(schema, req.body)
         ~~~~~~~~~~~~~~~~~


Found 1 error.
```

This happens when not providing any types. QueryFields is the default value, which treats the Body param as the query one and evaluates to undefined.

olivermrbl (dismissed):
Neat

NicolasGorga:
Yes, added the changes to the type as per Adrien suggestion, though that introduces the issue I mentioned in my last comment when we don't pass the types and use defaults, while try to assign to validatedBody, since in these cases it is treated as undefined (as per the "if only one type provided, we assume no body and only query"). I will wait for his input :)

adrien2p:
> Yes, added the changes to the type as per Adrien suggestion, though that introduces the issue I mentioned in my last comment when we don't pass the types and use defaults, while try to assign to validatedBody, since in these cases it is treated as undefined (as per the "if only one type provided, we assume no body and only query"). I will wait for his input :)

Yeah I ve suggeted undefined but it can be a record or unknown or something like that, the main idea is to route the correct template args to the correct location 👍

adrien2p (approved):

NicolasGorga:
I am still having build issues, thinking of how to solve this with the changes we did

PR #14111, opened 2025-11-24
fix(core-flows): lock mark as delivered fulfillment/order flows

NicolasGorga:
## Summary

Add locking to `markFulfillmentAsDeliveredWorkflow` and `markOrderFulfillmentAsDeliveredWorkflow` flows to avoid race conditions.

Without locking, race conditions could arise from multiple workers executing the workflows at the same time. If both execute at the same time before the other updates the fulfillment status to delivered, the check we put for undelivered workflows will be true for both and we'll duplicate the underlying order changes.

Included locking by fulfillment id inside `markFulfillmentAsDeliveredWorkflow` when used as standalone.

Included locking by order id inside `markOrderFulfillmentAsDeliveredWorkflow` so that it guarantees `markFulfillmentAsDeliveredWorkflow` locks when called as a sub workflow.

---

## Examples

```ts
// Example usage
```

---

## Checklist

---

## Additional Context

closes ENTSUP-399

### DIFF
--- a/.changeset/new-owls-heal.md
+++ b/.changeset/new-owls-heal.md
@@ -1,0 +1,5 @@
+---
+"@medusajs/core-flows": patch
+---
+
+fix(core-flows): lock mark as delivered fulfillment/order flows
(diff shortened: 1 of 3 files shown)


NicolasGorga:
@adrien2p let me know what you think of this. I added locking to `markOrderFulfillmentAsDeliveredWorkflow` since it is my understanding that without this, the locking inside `markFulfillmentAsDeliveredWorkflow` would have no effect as it is being called as a sub workflow no?

adrien2p:
> @adrien2p let me know what you think of this. I added locking to `markOrderFulfillmentAsDeliveredWorkflow` since it is my understanding that without this, the locking inside `markFulfillmentAsDeliveredWorkflow` would have no effect as it is being called as a sub workflow no?

yes thats correct, by default we dont run the lock on sub workflows 👍 so it makes sense

adrien2p (approved):

PR #14138, opened 2025-11-26
fix(dashboard,medusa,types): improve performance for price list prices retrieval

NicolasGorga:
## Summary

Price lists prices didn't have a dedicated method to query them and instead, relied on being returned as part of price lists. This, however, introduces optimization issues that for price lists with many prices, could cause crashes. The reason being that relations are not paginated and thus, all prices linked to the price list would be returned.

This PR aims to solve this by introducing a dedicated endpoint and avoiding returning the `prices` as part of price lists by default. The idea being that it is up to the user to explicitly express this, which, for small price lists no issues will arise, but for bigger ones, they will easily recognize the performance impact.

Users with large enough price lists would have serious performance issues or even crashes when querying the `/admin/price-lists` endpoints. This is also true when navigating to the price list section of the Admin UI since it queries this same endpoints.

- Removed the `prices` relation to be part of the default fields returned by the `/admin/price-lists/` endpoints. User may still request it by passing it in `fields` query param.
- Added new `/admin/price-lists/[id]/prices` GET endpoint to be able to retrieve a price list prices with pagination.

Integration tests.

---

## Examples

```ts
// Example usage
```

---

## Checklist

---

## Additional Context

The current state of the PR fixes the issue on the price list list and detail component. It still doesn't solve the issue for the following screens: Edit Prices & Add Prices

All the prices are still retrieved from the `/admin/price-lists/` endpoint for these. I want first some feedback before changing it to the new endpoint, since the current DataGrid implementation doesn't support pagination and it seems we are passing a default limit for the products to show there, an arbitrarily large number 9999 and there is also a TODO comment of changing that.

This previous point, though, could be implemented in a later PR, so we can already fix the issue in the price list list and detail pages, so at least for large price lists these screens don't explode and smaller price lists can still have its product prices edited, while only large ones will explode when trying to perform this action. @adrien2p @fPolic thoughts?

closes ENTSUP-265, CORE-1239


adrien2p:
@olivermrbl would using the changeset experimental flag `onlyUpdatePeerDependentsWhenOutOfRange` help us with those major bump on peer deps bump?

olivermrbl (approved):
LGTM, think we should solve this for Edit Prices & Add Prices too before we release

olivermrbl (commented):
~~Oh, and why is this a minor? I didn't see any breaking changes~~

Nevermind, I see it now. We removed `prices` from the default fields.

adrien2p (approved):

carlos-r-l-rodrigues (approved):

packages/admin/dashboard/src/hooks/api/price-lists.tsx
```diff
@@ -124,6 +129,29 @@ export const useDeletePriceList = (
...
+  options?: UseQueryOptions<
+    HttpTypes.AdminPriceListPriceListResponse,
+    FetchError,
+    HttpTypes.AdminPriceListPriceListResponse,
+    QueryKey
+  >
+) => {
+  const { data, ...rest } = useQuery({
+    queryFn: () =>
+      sdk.client.fetch<HttpTypes.AdminPriceListPriceListResponse>(
```
fPolic: nit: should we add an sdk method for this?

packages/admin/dashboard/src/routes/price-lists/price-list-detail/components/price-list-general-section/price-list-general-section.tsx
```diff
@@ -15,8 +16,11 @@ export const PriceListGeneralSection = ({
   priceList,
 }: PriceListGeneralSectionProps) => {
   const { t } = useTranslation()
+  const { count, isLoading, error } = usePriceListPrices(priceList.id, {
+    limit: 1,
+  })
 
-  const overrideCount = priceList.prices?.length || 0
+  const overrideCount = count
```
fPolic: nit: no need to define new var, you can just use/reneme `count`

packages/medusa/src/api/admin/price-lists/query-config.ts
```diff
@@ -2,7 +2,8 @@ export enum PriceListRelations {
   PRICES = "prices",
 }
 
-export const adminPriceListPriceRemoteQueryFields = [
+// Note: renamed to avoid referencing remoteQuery which i legacy
```
fPolic:
```suggestion
// Note: renamed to avoid referencing remoteQuery which is legacy
```

fPolic (approved):
LGTM!

adrien2p:
Tests are failing can you have a look

PR #14145, opened 2025-11-27
chore: Update medusa exec worker mode

adrien2p:
RESOLVES CORE-1292
**What**

Currently, using `medusa exec` start the application in worker mode, which sometimes leads to unexpected behaviour in the sense that potential events get emitted and the worker might start to pull them but then get killed once the script is done, this can lead to partially handled event and do not allow for recovery of such

### DIFF
--- a/.changeset/cuddly-numbers-camp.md
+++ b/.changeset/cuddly-numbers-camp.md
@@ -1,0 +1,5 @@
+---
+"@medusajs/medusa": patch
+---
+
+chore: Update medusa exec worker mode
--- a/packages/medusa/src/commands/exec.ts
+++ b/packages/medusa/src/commands/exec.ts
@@ -41,7 +41,6 @@ export default async function exec({ file, args }: Options) {
       throw new Error(`File doesn't default export a function to execute.`)
     }
 
-    // set worker mode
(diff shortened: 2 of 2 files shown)


adrien2p:
@shahednasser do you think it might be worth letting users know about scripts limitation. For example the one I ve fixed here, but also if someone tries to run async workflows they might face unexpected behaviour from their POV. let me know if you want more details

shahednasser:
@adrien2p yup we definitely should.

I wonder though if starting it in server mode would have side effects. Would it try to start the server at port 9000 for example? Because that would cause an error if an actual server is running. Another thing is Cloud users may have been using a script in their `build` command, do you know if that could cause side effects for them?

adrien2p:
> @adrien2p yup we definitely should.
> 
> I wonder though if starting it in server mode would have side effects. Would it try to start the server at port 9000 for example? Because that would cause an error if an actual server is running. Another thing is Cloud users may have been using a script in their `build` command, do you know if that could cause side effects for them?

- The exec command only runs the loaders and does not start any servers under the hood, it was already the case
- It shouldnt impact those users, it will do all the same thing, but it should rather prevent nasty side effect rather than adding more

olivermrbl (approved):

PR #14148, opened 2025-11-28
feat(): default to select-in strat

adrien2p:


MicahBCode:
If you consider upgrading mikroorm to version ^6.5.0 you could also use the balanced loading strategy per default which seems to give the best results out of both.

adrien2p:
> If you consider upgrading mikroorm to version ^6.5.0 you could also use the balanced loading strategy per default which seems to give the best results out of both.

We had a pr upgrading to mikro orm latest but it came with it’s own set of issues that we need to fix. But definitely a goal 👌

Issue #13783, opened 2025-10-20
[Bug]: Product updates with many variants and images memory usage/problem


mikkel-lindstrom:
### Package.json file

```JSON
{
  "name": "my-medusa-performance-store",
  "version": "0.0.1",
  "description": "A starter for Medusa projects.",
  "author": "Medusa (https://medusajs.com)",
  "license": "MIT",
  "keywords": [
    "sqlite",
    "postgres",
    "typescript",
    "ecommerce",
    "headless",
    "medusa"
  ],
  "scripts": {
    "build": "medusa build",
    "seed": "medusa exec ./src/scripts/seed.ts",
    "start": "medusa start",
    "dev": "medusa develop",
    "test:integration:http": "TEST_TYPE=integration:http NODE_OPTIONS=--experimental-vm-modules jest --silent=false --runInBand --forceExit",
    "test:integration:modules": "TEST_TYPE=integration:modules NODE_OPTIONS=--experimental-vm-modules jest --silent=false --runInBand --forceExit",
    "test:unit": "TEST_TYPE=unit NODE_OPTIONS=--experimental-vm-modules jest --silent --runInBand --forceExit"
  },
  "dependencies": {
    "@medusajs/admin-sdk": "2.10.3",
    "@medusajs/cli": "2.10.3",
    "@medusajs/framework": "2.10.3",
    "@medusajs/medusa": "2.10.3",
    "@mikro-orm/core": "6.4.3",
    "@mikro-orm/knex": "6.4.3",
    "@mikro-orm/migrations": "6.4.3",
    "@mikro-orm/postgresql": "6.4.3",
    "awilix": "^8.0.1",
    "pg": "^8.13.0"
  },
  "devDependencies": {
    "@medusajs/test-utils": "2.10.3",
    "@mikro-orm/cli": "6.4.3",
    "@swc/core": "1.5.7",
    "@swc/jest": "^0.2.36",
    "@types/jest": "^29.5.13",
    "@types/node": "^20.0.0",
    "@types/react": "^18.3.2",
    "@types/react-dom": "^18.2.25",
    "jest": "^29.7.0",
    "prop-types": "^15.8.1",
    "react": "^18.2.0",
    "react-dom": "^18.2.0",
    "ts-node": "^10.9.2",
    "typescript": "^5.6.2",
    "vite": "^5.2.11",
    "yalc": "^1.0.0-pre.53"
  },
  "engines": {
    "node": ">=20"
  },
[...]

PR #14150, opened 2025-11-28
fix(product): Update performance issue

adrien2p:
**What**

After many runs, we can see that with a medium catalog the performances for updating a product are as follow:
```
before:      3500-5300ms
after:       70-100ms
```

The primary change in this PR addresses a performance issue related to the query strategy used during data fetching.

Previously, the service layer was fetching data without specifying a strategy, which defaulted to the joined strategy. This approach retrieved all related data in a single query with
multiple JOINs, putting significant stress on the database when dealing with large datasets — resulting in substantially increased query times.

The repository layer already had logic to apply pagination based on the input size and force the select-in strategy in those cases. This PR ensures the service layer also leverages
this optimization, preventing the joined strategy from being used when it would be detrimental to performance.

Key improvement: By using the select-in strategy, the database handles smaller, more efficient queries instead of a single heavy JOIN query, dramatically reducing query execution time.


NicolasGorga:
Relates to #13783 . We can close the issue after this and the PR for CORE-1296 is merged

olivermrbl (approved):
LGTM

PR #14153, opened 2025-11-28
chore: improve completeCartWorkflow TSDocs

shahednasser:
## Summary

Improve the TSDocs of the `completeCartWorkflow` to mention that it's idempotent, with tips on how to handle common use cases.

> Note: this will be part of the generated references in the docs.

Closes DX-2335

*Please provide answer here*

*Please provide answer here*

*Please provide answer here*

---

## Examples

```ts
// Example usage
```

---

## Checklist

---

## Additional Context


packages/core/core-flows/src/cart/workflows/complete-cart.ts
```diff
@@ -75,6 +75,212 @@ export const completeCartWorkflowId = "complete-cart"
  * You can use this workflow within your own customizations or custom workflows, allowing you to wrap custom logic around completing a cart.
  * For example, in the [Subscriptions recipe](https://docs.medusajs.com/resources/recipes/subscriptions/examples/standard#create-workflow),
  * this workflow is used within another workflow that creates a subscription order.
+ * 
+ * ## Cart Completion Idempotency
+ * 
+ * This workflow is idempotent, meaning that if it is executed multiple times with the same input, it will not create duplicate orders. The
+ * same order will be returned for subsequent executions with the same cart ID. This is necessary to avoid rolling back payments or causing
+ * other side effects if the workflow is retried or fails due to transient errors.
```
adrien2p, line 83: Do you think we should specifically mention concurrency usage of the workflow, such that the lock manages the queuing but they will run one after each other leading 100% error if not managed idempotently in the sense that errors will occur already because we would duplicate things?

packages/core/core-flows/src/cart/workflows/complete-cart.ts
```diff
@@ -75,6 +75,212 @@ export const completeCartWorkflowId = "complete-cart"
...
+ * 
+ * So, if you use this workflow within your own, make sure your workflow's steps are idempotent as well to avoid unintended side effects. The following
```
adrien2p: I think it is important to dissociate idempotent behaviour from idempotent configuration, one might be confused no?

adrien2p: should the lock be mentionned in the doc from there or not?

adrien2p (approved):

packages/core/core-flows/src/cart/workflows/complete-cart.ts
shahednasser: Do you mean if you don't use locks? Or is this the case right now if the `completeCartWorkflow` is used concurrently?

packages/core/core-flows/src/cart/workflows/complete-cart.ts
shahednasser: > I think it is important to dissociate idempotent behaviour from idempotent configuration, one might be confused no?

Makes sense, will add it

> should the lock be mentionned in the doc from there or not?

It is mentioned in the generated docs because we automatically add a note about using locks if a workflow uses the `acquireLockStep`, but I'll look into if it makes sense to mention it here as well.

packages/core/core-flows/src/cart/workflows/complete-cart.ts
adrien2p: the case right now if used concurrently, like with stripe/store for example

PR #14177, opened 2025-12-01
fix(core-flows): Avoid checking inventory items on fulfillment cancel for unmanaged inventory variants

NicolasGorga:
## Summary

Avoid having an error when trying to cancel a fulfillment and one of the item variants changed its `manage_inventory` status from `true` to `false`.

In cases like the above, the inventory item is still linked to the product variant, but the fulfillment_item has no linked inventory item. This causes an issue when we expect we will find a match between the variant linked inventory items and the fulfillment item inexistent inventory item link.

Avoiding making performing this macth between fulfillment item inventory items and product variant inventory items when the variant has `manage_inventory=false`

Existent integration tests.

---

## Examples

```ts
// Example usage
```

---

## Checklist

---

## Additional Context

closes CORE-1298

---

> [!NOTE]
> Skips inventory matching on fulfillment cancellation for unmanaged variants and auto-unlinks variant–inventory links when variants are updated to manage_inventory=false.
> 
> - **Core Flows**:
>   - **Order Fulfillment Cancel**: In `order/workflows/cancel-order-fulfillment.ts`, only consider `variant.inventory_items` when `variant.manage_inventory` is true to avoid inventory matching for unmanaged variants.
> - **Product Workflows**:
>   - **New Step**: Add `dismiss-product-variants-inventory` step to remove links between variants and inventory items.
>   - **Integration**:
>     - In `product/workflows/update-product-variants` and `product/workflows/update-products`, invoke the dismiss step for variants updated with `manage_inventory=false`.
>     - Export the new step in `product/steps/index.ts`.
> - **Tests**:
>   - Add admin API tests asserting that setting `manage_inventory=false` on variants unlinks all `inventory_items` while keeping inventory items intact.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 7d56f8779e85184a430c6da0f5267b5e255e5d25. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


NicolasGorga:
I wonder if, apart from doing this (for already existent orders and fulfillments in this state), we should unlink the product variant from its inventory items when it goes from managed to unmanaged inventory-wise. This way, the inventory item is retained, for the user to do whatever they need to do in practice before deleting or just keeping it there altogether, but the link is also faithful to the actual inventory management state of the variant. wyt?

adrien2p (commented):
I agree, if the variants is unmanaged we should clean up inventories for it right?

NicolasGorga:
> I agree, if the variants is unmanaged we should clean up inventories for it right?

Yes, will add that in this PR then :)

olivermrbl (dismissed):

packages/core/core-flows/src/order/workflows/cancel-order-fulfillment.ts
```diff
@@ -160,7 +160,9 @@ function prepareCancelOrderFulfillmentData({
         (i) => i.id === lineItemId
       ) as OrderItemWithVariantDTO
       // find inventory items
-      const iitems = orderItem!.variant?.inventory_items
+      const iitems = orderItem!.variant?.manage_inventory
```
fPolic, line 163: q: this change has effectivley the same behaviour?

packages/core/core-flows/src/product/steps/dismiss-product-variants-inventory.ts
```diff
@@ -0,0 +1,87 @@
...
+  link: any
+): Promise<Record<string, string[]>> {
+  const dismissedVariantInventoryItems: Record<string, string[]> = {}
+
+  for (const variantId of variantIds) {
+    if (!variantId) {
+      continue
+    }
+
+    const { data: variantInventoryItems } = await query.graph({
```
fPolic: suggestion: we can query all variants in a single query

packages/core/core-flows/src/product/steps/dismiss-product-variants-inventory.ts
```diff
@@ -0,0 +1,87 @@
...
+  "dismiss-product-variants-inventory"
+
+export type DismissProductVariantsInventoryStepInput = {
+  variantIds: string[]
+}
+
+async function dismissVariantsInventory(
+  variantIds: string[],
+  query: any,
+  link: any
```
fPolic: suggestion: we can type these since they are Link and Query modules

packages/core/core-flows/src/product/steps/dismiss-product-variants-inventory.ts
```diff
@@ -0,0 +1,87 @@
...
+      },
+    })
+
+    dismissedVariantInventoryItems[variantId] = variantInventoryItems.map(
+      (item: { inventory_item_id: string }) => item.inventory_item_id
+    )
+
+    await Promise.all(
+      variantInventoryItems.map(async (item: { inventory_item_id: string }) => {
+        await link.dismiss({
```
fPolic: suggestion: you can pass an arrayof links directly to `dismiss`

packages/core/core-flows/src/product/workflows/update-products.ts
```diff
@@ -444,6 +445,35 @@ export const updateProductsWorkflow = createWorkflow(
...
+    const variantsToDismissInventory = transform(
+      { input, updatedProducts },
+      (data) => {
+        const variantIds: string[] = []
+
+        if ("products" in data.input) {
+          for (const product of data.input.products) {
+            if (product.variants?.length) {
+              variantIds.push(
+                ...product.variants.filter((v) => v.id).map((v) => v.id!)
```
fPolic: todo: should filter for `manage_inventory === false` here as well otherwise we will unlink all product variants from inventory on variant update

packages/core/core-flows/src/order/workflows/cancel-order-fulfillment.ts
NicolasGorga: Not really. When a variant goes from managed to unmanaged inventory-wise wise and it already had inventory items linked to it, then it will have the `inventory_items` relation, which we don't want to handle here. After this PR gets merged, then we would be able to assume that, as we would unlink the variant from its inventory items when its `manage_inventory` property is updated to false. But even then, if you don't use the core workflows to update the product then we wouldn't be sure this is the case

integration-tests/http/__tests__/product/admin/variant.spec.ts
```diff
@@ -649,6 +649,123 @@ medusaIntegrationTestRunner({
...
+          )
+        ).data.product
+
+        const updatedVariant = updatedProduct.variants.find(
+          (v) => v.id === variantWithInventory.id
+        )
+
+        expect(updatedVariant.manage_inventory).toBe(false)
+
+        expect(updatedVariant.inventory_items).toHaveLength(0)
```
fPolic, line 746: todo: this seems to be failing in tests

fPolic (approved):
LGTM

integration-tests/http/__tests__/product/admin/product.spec.ts
```diff
@@ -2356,6 +2356,125 @@ medusaIntegrationTestRunner({
           )
         })
 
+        it("updates variant to manage_inventory false and unlinks inventory items", async () => {
```
adrien2p: suggestion: If we could always start test case with `should` just for consistency :)

packages/core/core-flows/src/product/steps/dismiss-product-variants-inventory.ts
```diff
@@ -0,0 +1,89 @@
...
+}
+
+async function dismissVariantsInventory(
+  variantIds: string[],
+  query: Query,
+  link: Link
+): Promise<Record<string, string[]>> {
+  const dismissedVariantInventoryItems: Record<string, string[]> = {}
+
+  const { data: variantInventoryItems } = await query.graph({
```
adrien2p, line 23: suggestion: just as a safe guard can we add an early return if !variantIds.length just to avoid mistakes of calling it with an empty array which would fetch for nothing

packages/core/core-flows/src/product/steps/dismiss-product-variants-inventory.ts
```diff
@@ -0,0 +1,89 @@
...
+    },
+  })
+
+  const dismissLinks: LinkDefinition[] = []
+  for (const variantId of variantIds) {
+    if (!variantId) {
+      continue
+    }
+
+    dismissedVariantInventoryItems[variantId] = variantInventoryItems
```
adrien2p: suggestion: we could probably create a little map of it after the query

packages/core/core-flows/src/product/steps/dismiss-product-variants-inventory.ts
```diff
@@ -0,0 +1,89 @@
...
+      .filter((item) => item.variant_id === variantId)
+      .map((item: { inventory_item_id: string }) => {
```
adrien2p: suggestion: `map` return a new array, which here we dont need as we just want to loop through. I would either suggest `forEach` or `for of` loop

packages/core/core-flows/src/product/steps/dismiss-product-variants-inventory.ts
```diff
@@ -0,0 +1,89 @@
...
+    )
+    return new StepResponse(void 0, dismissedVariantInventoryItems)
+  },
+  async (dismissedVariantInventoryItems, { container }) => {
+    if (!dismissedVariantInventoryItems) {
+      return
+    }
+
+    const link = container.resolve(ContainerRegistrationKeys.LINK)
+    await Promise.all(
```
adrien2p: suggestion: instead of promise.all I prefer that we rely on promiseAll util, also, would it be possible to make one call to the create like we do for the dismiss?

packages/core/core-flows/src/product/workflows/update-product-variants.ts
```diff
@@ -151,6 +154,32 @@ export const updateProductVariantsWorkflow = createWorkflow(
...
+
+        if ("product_variants" in data.input) {
+          variantIds.push(
+            ...data.input.product_variants
+              .filter((v) => v.id && v.manage_inventory === false)
+              .map((v) => v.id!)
+          )
+        } else if (
+          data.input.update &&
+          data.input.update.manage_inventory === false
```
adrien2p:
```suggestion
          data.input.update?.manage_inventory === false
```

packages/core/core-flows/src/product/workflows/update-product-variants.ts
```diff
@@ -151,6 +154,32 @@ export const updateProductVariantsWorkflow = createWorkflow(
...
+      { input, updatedVariants },
+      (data) => {
+        const variantIds: string[] = []
+
+        if ("product_variants" in data.input) {
+          variantIds.push(
+            ...data.input.product_variants
+              .filter((v) => v.id && v.manage_inventory === false)
+              .map((v) => v.id!)
+          )
```
adrien2p: nit: loop -> push instead of loop -> filter -> map -> create array -> spread array

packages/core/core-flows/src/product/workflows/update-products.ts
```diff
@@ -444,6 +445,37 @@ export const updateProductsWorkflow = createWorkflow(
...
 
+    const variantsToDismissInventory = transform(
+      { input, updatedProducts },
+      (data) => {
+        const variantIds: string[] = []
+
+        if ("products" in data.input) {
+          for (const product of data.input.products) {
+            if (product.variants?.length) {
+              variantIds.push(
```
adrien2p: nit: same here

adrien2p: and bellow

packages/core/core-flows/src/product/steps/dismiss-product-variants-inventory.ts
NicolasGorga: But we don't just loop through, we assign to `dismissedVariantInventoryItems[variantId]` since we return this to the caller of `dismissVariantsInventory` function.

packages/core/core-flows/src/product/steps/dismiss-product-variants-inventory.ts
NicolasGorga: @adrien2p you mean instead of creating `dismissedVariantInventoryItems` by looping the `variantIds` to create it looping `variantInventoryItems`? Or you mean something else?

packages/core/core-flows/src/product/steps/dismiss-product-variants-inventory.ts
adrien2p: Sorry, it was a comment i thought i ve cancelled 🤣

packages/core/core-flows/src/product/steps/dismiss-product-variants-inventory.ts
adrien2p: no I meant to remove the filter in case we end up with many items.

NicolasGorga: The thing is i still need to filter at some point the inventory items for a particular variant to construct this map i am returning no?

adrien2p: if you have a map you will do `map.get` o(1) and dont have to filter anything, I am missing something?

NicolasGorga: Oh i get what you mean. Okay i can do that, just need to make sure that i don't return this map, or that at least, i transform it in the step that is calling this function, since i need to return a serializable object as a `StepResponse` right?

carlos-r-l-rodrigues (approved):

PR #14189, opened 2025-12-02
feat(): Introduce translation module and preliminary application of them

adrien2p:
Summary

  This PR introduces the Translation Module - a new first-party module for multi-language support in Medusa. The module provides a centralized way to manage translations for any entity
  type (products, categories, etc.) with support for BCP 47 locale codes.

  Key Features

  - Locale Management: Pre-populated with 48 common locales (en-US, fr-FR, de-DE, etc.) with support for custom locales
  - Entity Translation Storage: Store translations as JSONB with a unique constraint on (entity_id, entity_type, locale_code)
  - JSONB Search: Full-text search within translated content using the q filter
  - Locale Resolution Middleware: Automatically resolves locale from Accept-Language header or locale query parameter
  - Read-only Product Links: Connect products to their translations via the link modules system
  - Feature Flagged: Disabled by default, enable via MEDUSA_FF_TRANSLATION=true for both API layer and Admin UI
  - Store supported locales: You can select which of the available locales your Store supports
  - Translations batch API: manage translations with a single batch API admin endpoint
  - Apply translations on the fly for products (including relations translations, automanaged with bulk fetching baked in)

  Changes

  New Module (@medusajs/translation)
  - Locale model: Stores available locales (code, name)
  - Translation model: Stores translations with entity reference and JSONB content
  - TranslationModuleService: Full CRUD operations with JSONB search capability
  - Integration tests covering all happy/unhappy paths

  Update Store Module
  - Link to Translation Module Locale to define supported_locales
  - New UI in detail and edit components to update the relation

  Core Types (@medusajs/types)
  - ITranslationModuleService interface
  - DTOs: LocaleDTO, TranslationDTO, CreateTranslationDTO, UpdateTranslationDTO
  - Filterable props with q parameter for JSONB search

  Core Flows (@medusajs/core-flows)
  - createTranslationsStep / createTranslationsWorkflow
  - updateTranslationsStep / updateTranslationsWorkflow
  - deleteTranslationsStep / deleteTranslationsWorkflow
  - validateTranslationsStep
  - batchTranslationsWorkflow

  Framework
  - applyLocale middleware: Resolves locale from request headers/query params
  - normalizeLocale: Applies normalization on write/read operations for the local code
  - Module registered in Modules.TRANSLATION

  Medusa Package
  - Feature flag: translation.ts
  - Product middleware integration for locale-aware responses
  - Read-only link: product-translation.ts
  - applyTranslations
    - batch querying
    - auto apply translation to input objects (including nested)
  - applied the above to product retrieve and list store api routes


packages/core/framework/src/http/middlewares/apply-locale.ts
```diff
@@ -0,0 +1,39 @@
+import { normalizeLocale } from "@medusajs/utils"
+import type {
+  MedusaNextFunction,
+  MedusaRequest,
+  MedusaResponse,
+} from "../types"
+
+const CONTENT_LANGUAGE_HEADER = "content-language"
```
olivermrbl, line 8: todo: So, this actually seems to be a response header, not a request header, so it is relatively unusual to be using it like this. Instead we should be using `Accept-Language` and extract the locale from there. The locales in this header typically come with weights, we can use for the selection. That is of cause, if the locale is not singular.

packages/core/js-sdk/src/admin/locale.ts
```diff
@@ -0,0 +1,120 @@
...
+   */
+  private client: Client
+  /**
+   * @ignore
+   */
+  constructor(client: Client) {
+    this.client = client
+  }
+
+  // TODO: Ask Adrien/Shahed how do I know how to construct the api docs url when implementing new methods
```
olivermrbl: comment: there is a TODO here

packages/medusa/src/api/admin/translations/middlewares.ts
```diff
@@ -0,0 +1,73 @@
...
+import {
+  ContainerRegistrationKeys,
+  MedusaError,
+  MedusaErrorTypes,
+  Modules,
+} from "@medusajs/framework/utils"
+import { IFlagRouter } from "@medusajs/framework/types"
+import TranslationFeatureFlag from "../../../feature-flags/translation"
+
+const isTranslationsEnabledMiddleware = async (
```
olivermrbl: thought: wouldn't it be simpler to just omit the route files when the feature flag is not enabled? using the utility that @carlos-r-l-rodrigues created

packages/core/framework/src/http/middlewares/apply-locale.ts
adrien2p: Accept-language is not for language selection I believe, but more if no locale is being provided, then we can pick from their and choose the one with the higher weight if available. This header is meant as a Hint for the server if it can't find the prefered locale for the user

packages/modules/translation/src/migrations/Migration20251202230658.ts
```diff
@@ -0,0 +1,19 @@
+import { Migration } from '@mikro-orm/migrations';
+
+export class Migration20251202230658 extends Migration {
+
+  override async up(): Promise<void> {
+    this.addSql(`alter table if exists "locale" drop constraint if exists "locale_pkey";`);
```
olivermrbl: question: should we merge the  two migration files, now that we can?

olivermrbl (approved):
great stuff 💪

packages/core/framework/src/http/middlewares/apply-locale.ts
adrien2p: https://learn.microsoft.com/en-us/aspnet/core/fundamentals/localization/select-language-culture?view=aspnetcore-10.0 here they also use it for both req/res

olivermrbl: Fine to start with this, and we can reevaluate if need be.

packages/modules/translation/src/migrations/Migration20251202124048.ts
```diff
@@ -0,0 +1,37 @@
...
+    )
+
+    this.addSql(
+      `create table if not exists "translation" ("id" text not null, "entity_id" text not null, "entity_type" text not null, "locale_code" text not null, "translations" jsonb not null, "created_at" timestamptz not null default now(), "updated_at" timestamptz not null default now(), "deleted_at" timestamptz null, constraint "translation_pkey" primary key ("id"));`
+    )
+    this.addSql(
+      `CREATE INDEX IF NOT EXISTS "IDX_translation_deleted_at" ON "translation" ("deleted_at") WHERE deleted_at IS NULL;`
+    )
+    this.addSql(
+      `CREATE UNIQUE INDEX IF NOT EXISTS "IDX_translation_entity_id_entity_type_locale_code_unique" ON "translation" ("entity_id", "entity_type", "locale_code") WHERE deleted_at IS NULL;`
```
carlos-r-l-rodrigues: I believe we can remove `entity_type` from this index, or create a new one without it.
I believe we won't use this to retrieve the translations.

carlos-r-l-rodrigues (approved):
Left a small comment about the DB index.
Nice PR

packages/core/js-sdk/src/admin/locale.ts
shahednasser: Fine to remove the TODO and leave as is. I'll go over all of the TSDocs before release to fix them up.

packages/medusa-test-utils/src/medusa-test-runner.ts
```diff
@@ -287,6 +288,7 @@ class MedusaTestRunner {
         cwd: this.cwd,
       })
       await medusaAppLoader.runModulesLoader()
+      await createDefaultsWorkflow(copiedContainer).run()
```
adrien2p, line 291: cc @carlos-r-l-rodrigues @olivermrbl I found an issue herer while writing my tests for the other pr, the first test running can have access to the default store being created upon starting the app. Since the after each tear down the database, the defaults are lost and never re created since we only reload modules. should that be considered a breaking and do you see any issue in having the default always re created?

packages/medusa/src/api/admin/translations/middlewares.ts
olivermrbl: Wdyt about this @adrien2p?

packages/core/utils/src/common/define-config.ts
```diff
@@ -185,6 +185,8 @@ function resolveModules(
     { resolve: MODULE_PACKAGE_NAMES[Modules.ORDER] },
     { resolve: MODULE_PACKAGE_NAMES[Modules.SETTINGS] },
 
+    { resolve: MODULE_PACKAGE_NAMES[Modules.TRANSLATION] },
```
olivermrbl: question: should we leave this out of the shared modules for now? otherwise, migrations will be applied right? and we probably don't want that to happen before we are closer to finishing the feature

packages/medusa/src/api/admin/translations/middlewares.ts
adrien2p: Yes I agreed with that, I thought @NicolasGorga was going over the remaining reviews but he was caught on support. I’ll handle the remaining ones 👌

packages/core/utils/src/common/define-config.ts
adrien2p: Yes agree 👌

adrien2p: I’ll put a todo for later to re add that when Time comes, the config and feature flags are not yet loaded at this point.

packages/core/js-sdk/src/admin/locale.ts
NicolasGorga: Removed

packages/medusa/src/api/admin/translations/middlewares.ts
NicolasGorga: Done. Had to push a little fix since the routes were registered anyway when using `defineFileConfig`, just that they had an empty handler. This gave unexpected results, for example, when a validator fails, as it will return 400 instead of 404, as expected in these cases. I merged it already in here, but created a separate PR here https://github.com/medusajs/medusa/pull/14236 wyt? @adrien2p @olivermrbl

PR #14190, opened 2025-12-02
feat(): Pluralized props for list readonly link

adrien2p:
**What**
When setting a readonly link to `isList` then the alias/prop will be pluralized


packages/core/utils/src/modules-sdk/define-link.ts
```diff
@@ -504,24 +504,57 @@ ${serviceBObj.module}: {
...
+        fieldAlias: buildFieldAlias(readOnlyLinkOptions?.shortcut),
+        relationship: {
+          serviceName: serviceBObj.module,
+          entity: serviceBObj.entity,
+          primaryKey: serviceBObj.primaryKey,
+          foreignKey: serviceAObj.field,
+          alias:
+            readOnlyLinkOptions?.isList ?? serviceAObj.isList
+              ? pluralize(serviceBObj.alias)
+              : serviceBObj.alias,
```
adrien2p: just pluralize

packages/core/utils/src/modules-sdk/define-link.ts
```diff
@@ -504,24 +504,57 @@ ${serviceBObj.module}: {
...
+        relationship: {
+          serviceName: serviceBObj.module,
+          entity: serviceBObj.entity,
+          primaryKey: serviceBObj.primaryKey,
+          foreignKey: serviceAObj.field,
+          alias: serviceBObj.alias,
+          isList: readOnlyLinkOptions?.isList ?? serviceAObj.isList,
+        },
+      })
+    }
```
adrien2p: move out of the if/else

adrien2p:
@olivermrbl we should be good for this one 💪

adrien2p:
@carlos-r-l-rodrigues are you find with this one ? (cc @olivermrbl )

carlos-r-l-rodrigues (approved):

shahednasser (commented):
Should this be labeled as a breaking change? It doesn't seem like it's backwards compatible

adrien2p:
@shahednasser it is backward compatible, we still generate the singular alias in both cases like it used to be, but we add the plural alias on top of it as well. Let me know what makes you think it is not backward before I merge 👍

olivermrbl:
Should we merge this?

johnymanuelli:
@adrien2p @shahednasser  Not sure if this is only our unique problem, but we have a custom translation module 

````typescript
export default defineLink(
  {
    linkable: ProductModule.linkable.productCategory,
    field: "id",
    isList: true,
  },
  {
    linkable: {
      serviceName: TRANSLATION_MODULE,
      alias: "category_translations",
      primaryKey: "category_id",
    },
  },
  {
    readOnly: true,
  }
)

````

And after the upgrade to 2.12.3 it broke the way we did get those data via fields queries (*category_translations). After some investigation I have found what works, is that i need to rename the alias to category_translation, and then it will correctly work again.

PR #14191, opened 2025-12-02
feat: add non-sortable fields to entity mappings

adevinwild:
## Summary

- Introduced `nonSortableFields` to the entity mappings to specify fields that cannot be sorted, enhancing the data handling capabilities.
- Updated the `generateEntityColumns` function to respect the new `nonSortableFields` logic, ensuring proper sorting behavior across entities.

```ts
export const ENTITY_MAPPINGS = {
  orders: {
    serviceName: "order",
    graphqlType: "Order",
    defaultVisibleFields: [
      "display_id",
      "created_at",
      "payment_status",
      "fulfillment_status",
      "total",
      "customer_display",
      "country",
      "sales_channel.name",
    ],
    // Here the non sortable fields by default as they will throw a 500 error
    nonSortableFields: ["total", "payment_status", "fulfillment_status"], 
```

If you use the view configurations FF it will show that the total, fulfillment status or payment status can be sorted ASC or DESC, which is impossible as it is and will throw a 500 error on the server (virtual field)

This add a new key to the entity mapping `nonSortableFields` that will be used in the helper to determine the `sortable` bool

*Please provide answer here*

---

## Examples

https://github.com/user-attachments/assets/0937b7d0-98b2-448b-9250-3d0cd4ceed7b

**Fix on orders with nonSortableFields**

https://github.com/user-attachments/assets/8b35193d-763f-465f-93f3-83e64ab42bd9

---

## Checklist

- [X] I have added a **changeset** for this PR
- [X] I have linked the related issue(s) if applicable

---

## Additional Context

We can't inject the total, fulfillment status and payment status on the computedColumns from what I've understood, so I just kept it that way and added that logic on top of it, tell me if that's not the way 🙌

### DIFF
--- a/.changeset/wet-bags-arrive.md
+++ b/.changeset/wet-bags-arrive.md
@@ -1,0 +1,5 @@
+---
+"@medusajs/medusa": patch
+---
+
+Added `nonSortableFields` on the entity mappings for the view_configuration ff to disable sortable fields on computed columns
--- a/packages/medusa/src/api/admin/views/[entity]/columns/entity-mappings.ts
+++ b/packages/medusa/src/api/admin/views/[entity]/columns/entity-mappings.ts
@@ -20,6 +20,8 @@ export const ENTITY_MAPPINGS = {
       // Specific field names to exclude
       excludeFields: ["order_change"],
     },
+    // Fields that cannot be sorted (virtual/computed fields)
+    nonSortableFields: ["total", "payment_status", "fulfillment_status"],
     computedColumns: {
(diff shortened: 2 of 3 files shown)


adevinwild:
issue
https://github.com/medusajs/medusa/issues/14192

adevinwild:
Now that I think about it we should add every virtual fields that exists as a column, should we just throw a bunch of keys inside ? or is there a more elegant way that will follow the current architecture ?

PR #14202, opened 2025-12-03
chore(): Run some commands in server mode only

adrien2p:
Summary

  - Set MEDUSA_WORKER_MODE="server" for CLI database commands (db:migrate, db:rollback, run-scripts) and the user command
  - Move subscribers and jobs loading to occur AFTER MedusaAppLoader completes, ensuring all modules are fully initialized before subscribers/jobs are registered

  Implications of the Loading Order Change

  Before

```ts
  loadEntrypoints() {
    subscribersLoader()  // ← Loaded here
    jobsLoader()         // ← Loaded here
    // Then early return if worker mode
    // Then load API routes...
  }
```

  After

```ts
  // In main loader function:
  MedusaAppLoader().load()  // All modules initialized
  WorkflowLoader().load()   // Workflows loaded
  subscribersLoader()        // ← Now loaded here (after modules)
  jobsLoader()               // ← Now loaded here (after modules)
  loadEntrypoints()          // API routes loaded separately
```

  Key implications:

  1. Modules are guaranteed to be fully initialized before subscribers/jobs - This prevents race conditions where a subscriber might try to access a module service that isn't yet available
  2. CLI commands run in "server" mode - This means:
    - shouldLoadBackgroundProcessors() returns false → scheduled jobs are not loaded
    - Subscribers are loaded but won't process events (no worker running)
    - Appropriate for one-shot CLI operations that should exit cleanly
  3. For user command specifically - Changed from "worker" to "server" mode:
    - Previously: Jobs would load (potentially unwanted background work)
    - Now: Jobs don't load, cleaner execution for user creation


sradevski (approved):

olivermrbl (approved):
LGTM

PR #14207, opened 2025-12-03
feat: validate product <> option assignments and updates

fPolic:
**What**
- validate product <> options assignments and updates

---

CLOSES CORE-1305


integration-tests/modules/__tests__/product/workflows/product-options-removal-validation.spec.ts
```diff
@@ -0,0 +1,365 @@
+import { createAndLinkProductOptionsToProductWorkflow } from "@medusajs/core-flows"
+import { medusaIntegrationTestRunner } from "@medusajs/test-utils"
+import { IProductModuleService } from "@medusajs/types"
+import { MedusaError, Modules } from "@medusajs/utils"
+
+jest.setTimeout(50000)
+
+medusaIntegrationTestRunner({
+  env: {},
```
adrien2p: nit: rm

integration-tests/modules/__tests__/product/workflows/product-options-removal-validation.spec.ts
```diff
@@ -0,0 +1,365 @@
...
+      let productModule: IProductModuleService
+
+      beforeAll(async () => {
+        appContainer = getContainer()
+        productModule = appContainer.resolve(Modules.PRODUCT)
+      })
+
+      describe("removeProductOptionFromProduct", () => {
+        it("should throw error when trying to remove ProductOption used by variant", async () => {
+          // Create product with option
```
adrien2p, line 21: nit: those comments can be removed, the actions are pretty self explanatory in themselves :D

integration-tests/modules/__tests__/product/workflows/product-options-removal-validation.spec.ts
```diff
@@ -0,0 +1,365 @@
...
+            title: "Test Product No Variants",
+            options: [
+              {
+                title: "Color",
+                values: ["Red", "Blue"],
+              },
+            ],
+          })
+
+          const optionId = product.options[0].id
```
adrien2p, line 61: suggestion: just for stability, are we sure the returned order is predictable here?

integration-tests/modules/__tests__/product/workflows/product-options-removal-validation.spec.ts
```diff
@@ -0,0 +1,365 @@
...
+            .filter((v) => v.value !== "S")
+            .map((v) => v.id)
+
+          await expect(
+            productModule.addProductOptionToProduct({
+              product_id: product.id,
+              product_option_id: option.id,
+              product_option_value_ids: remainingValueIds,
+            })
+          ).rejects.toThrow(MedusaError)
```
adrien2p: suggestion: should we check the message as well?>

integration-tests/modules/__tests__/product/workflows/product-options-removal-validation.spec.ts
```diff
@@ -0,0 +1,365 @@
...
+
+          await expect(
+            productModule.addProductOptionToProduct({
+              product_id: product.id,
+              product_option_id: option.id,
+              product_option_value_ids: remainingValueIds,
+            })
+          ).rejects.toThrow(
+            "Cannot unassign option values from product because the following variant(s) are using it"
```
adrien2p, line 113: Q: the message seems to be a bit miss leading, when I read `the following variant(s)` I thought I ll see their ids later :) maybe we could go with `Cannot unassign option values from product ${id) because the some variant(s) are using it`

integration-tests/modules/__tests__/product/workflows/product-options-removal-validation.spec.ts
```diff
@@ -0,0 +1,365 @@
...
+
+          try {
+            await productModule.addProductOptionToProduct({
+              product_id: product.id,
+              product_option_id: option.id,
+              product_option_value_ids: remainingValueIds,
+            })
+          } catch (error) {
+            expect((error as MedusaError).message).toContain("Variant S")
+          }
```
adrien2p: todo: try catch are generally not a good practice, if it doesnt fail, the expectation never run and the test would succeed. two options, either adding an expectation in the try such as
```ts
try {
  await productModule.addProductOptionToProduct({
    product_id: product.id,
    product_option_id: option.id,
    product_option_value_ids: remainingValueIds,
  })
  expect(true).toBe(false)
} catch (e) {
  expect((error as MedusaError).message).toContain("Variant S")
}
```

or alternatively

```ts
const res = await productModule.addProductOptionToProduct({
  product_id: product.id,
  product_option_id: option.id,
  product_option_value_ids: remainingValueIds,
}).catch(e => e)
expect((error as MedusaError).message).toContain("Variant S")
```

adrien2p: I let you check if there is any other place like that

integration-tests/modules/__tests__/product/workflows/product-options-removal-validation.spec.ts
```diff
@@ -0,0 +1,365 @@
...
+                "options.values",
+                "product_options",
+                "product_options.values",
+              ],
+            }
+          )
+          // Check the filtered option values (should only show linked values)
+          const updatedOption = updatedProduct.options[0]
+          expect(updatedOption.values).toHaveLength(2)
+          expect(updatedOption.values.map((v) => v.value)).toEqual(
```
adrien2p, line 186: Q: is the order deterministic?

integration-tests/modules/__tests__/product/workflows/product-options-removal-validation.spec.ts
```diff
@@ -0,0 +1,365 @@
...
+          })
+
+          const variant2 = await productModule.createProductVariants({
+            product_id: product.id,
+            title: "Variant Medium",
+            options: { Size: "M" },
+          })
+
+          try {
+            await productModule.removeProductOptionFromProduct({
```
adrien2p: todo: same as previous

integration-tests/modules/__tests__/product/workflows/product-options-removal-validation.spec.ts
```diff
@@ -0,0 +1,365 @@
+import { createAndLinkProductOptionsToProductWorkflow } from "@medusajs/core-flows"
```
adrien2p, line 1: general comment: here most of the tests are purely testing the product module itself, I would have expected them to go in the product module tests suite no? aqnd keep the workflow tests here as this is an integration relying on the product module. let me know what you think

packages/modules/product/src/repositories/product.ts
```diff
@@ -220,4 +220,76 @@ export class ProductRepository extends DALUtils.mikroOrmBaseRepositoryFactory(
...
+    Map<
+      string,
+      Array<{ variant_id: string; title: string | null; product_id: string }>
+    >
+  > {
+    if (pairs.length === 0) {
+      return new Map()
+    }
+
+    // Filter out pairs with empty value arrays
```
adrien2p: todo: there is a lot of unnecessary comments that just explain what is already very explicit, could we remove them following the DRY principle please? :)

packages/modules/product/src/repositories/product.ts
```diff
@@ -220,4 +220,76 @@ export class ProductRepository extends DALUtils.mikroOrmBaseRepositoryFactory(
...
+  /**
+   * Checks if any variants of products use the specified option values.
+   * Queries the product_variant_option pivot table directly in bulk.
+   *
+   * @param pairs - Array of { productId, optionValueIds } pairs to check
+   * @param context - The context
+   * @returns Map keyed by `${productId}_${valueId}` with arrays of variant info (id and title) that use those values
+   */
+  async checkVariantsUsingOptionValues(
+    pairs: Array<{ productId: string; optionValueIds: string[] }>,
```
adrien2p:
```suggestion
    pairs: Array<{ productId: string; optionValueIds: string[] }> = [],
```

packages/modules/product/src/repositories/product.ts
```diff
@@ -220,4 +220,76 @@ export class ProductRepository extends DALUtils.mikroOrmBaseRepositoryFactory(
...
+
+    const manager = this.getActiveManager<SqlEntityManager>(context)
+    const connection = manager.getConnection()
+    const knex = connection.getKnex()
+
+    // Collect all unique product IDs and value IDs
+    const allProductIds = [...new Set(validPairs.map((p) => p.productId))]
+    const allValueIds = [
+      ...new Set(validPairs.flatMap((p) => p.optionValueIds)),
+    ]
```
adrien2p, line 262:
```suggestion
    const allProductIds = new Set<string>()
    const allValueIds = new Set<string>()
    for (const pair of pairs) {
      allProductIds.add(pair.productId)
      for (const valueId of pair.optionValueIds) {
        allValueIds.add(valueId)
      }
    }

    if (allValueIds.size === 0) {
      return new Map()
    }

    const manager = this.getActiveManager<SqlEntityManager>(context)
    const connection = manager.getConnection()
    const knex = connection.getKnex()
```

packages/modules/product/src/repositories/product.ts
```diff
@@ -220,4 +220,76 @@ export class ProductRepository extends DALUtils.mikroOrmBaseRepositoryFactory(
...
+    for (const row of result) {
+      const key = `${row.product_id}_${row.option_value_id}`
+      if (!resultMap.has(key)) {
+        resultMap.set(key, [])
+      }
+      resultMap.get(key)!.push({
+        variant_id: row.variant_id,
+        title: row.title,
+        product_id: row.product_id,
+      })
```
adrien2p, line 292:
```suggestion
      const key = `${row.product_id}_${row.option_value_id}`
      const results = resultMap.get(key) ?? []
      results.push({
        variant_id: row.variant_id,
        title: row.title,
        product_id: row.product_id,
      })
      resultMap.set(key, results)
```

packages/modules/product/src/services/product-module-service.ts
```diff
@@ -1210,14 +1210,111 @@ export default class ProductModuleService
...
+      optionId: string
+      valueIdsToCheck: string[]
+    }> = []
+
+    for (const pair of pairs) {
+      if (pair.product_option_value_ids) {
+        const existingProductOption = existingProductOptions.find(
+          (epo) =>
+            (epo as any).product_id === pair.product_id &&
+            (epo as any).product_option_id === pair.product_option_id
```
adrien2p: todo: We can probably remove the cast to any here?

packages/modules/product/src/services/product-module-service.ts
```diff
@@ -1210,14 +1210,111 @@ export default class ProductModuleService
...
+              productId: pair.product_id,
+              optionId: pair.product_option_id,
+              valueIdsToCheck: removedValueIds,
+            })
+          }
+        }
+      }
+    }
+
+    if (validationPairs.length > 0) {
```
```suggestion
    if (validationPairs.length) {
```

packages/modules/product/src/services/product-module-service.ts
```diff
@@ -1210,14 +1210,111 @@ export default class ProductModuleService
...
+      await this.validateOptionRemoval_(validationPairs, sharedContext)
+    }
+
+    // Separate pairs: those that need PPO creation vs those that just need value updates
+    const pairsNeedingPPOCreation: ProductTypes.ProductOptionProductPair[] = []
+    for (const pair of pairs) {
+      const hasExistingPPO = existingProductOptions.some(
+        (epo) =>
+          (epo as any).product_id === pair.product_id &&
+          (epo as any).product_option_id === pair.product_option_id
```
adrien2p: same here

packages/modules/product/src/services/product-module-service.ts
```diff
@@ -1210,14 +1210,111 @@ export default class ProductModuleService
...
+      )
+      // Only create PPO if it doesn't exist
```
adrien2p: Q: for the overall create/update/delete, would the upsertWithReplace could have been used now that we improve its intrinsinc performance? and if yes what would be the impact overall

packages/modules/product/src/services/product-module-service.ts
```diff
@@ -1250,11 +1349,96 @@ export default class ProductModuleService
...
-          valuePairsToCreate.push({
-            product_product_option_id: ppo.id,
-            product_option_value_id: value.id,
-          })
+        // If updating with specific value_ids, remove old values that aren't in the new list
+        if (originalPair.product_option_value_ids) {
+          const existingProductOption = existingProductOptions.find(
+            (epo) =>
+              (epo as any).product_id === originalPair.product_id &&
+              (epo as any).product_option_id === originalPair.product_option_id
```
adrien2p: same here

packages/modules/product/src/services/product-module-service.ts
```diff
@@ -1250,11 +1349,96 @@ export default class ProductModuleService
...
+              })
+            }
+          }
+        } else {
+          // No specific value_ids, create all value links (only if PPO is newly created)
+          if (
+            !existingProductOptions.find(
+              (epo) =>
+                (epo as any).product_id === pair.product_id &&
+                (epo as any).product_option_id === pair.product_option_id
```
adrien2p: same here

packages/modules/product/src/services/product-module-service.ts
```diff
@@ -1305,9 +1499,31 @@ export default class ProductModuleService
     const productOptionsProducts = await this.productProductOptionService_.list(
       {
         $or: pairs,
-      }
+      },
+      { relations: [] },
```
adrien2p: Q: is that necessary? I would have thought that the relations are not eager loaded through the internal service 🤔

adrien2p (commented):
Global comments, overall I believe there is too much things going on inside the service itself which could be moved in utilities functions instead, just to reduce the complexity of the service which becomes hard to read and reason about. I also feel (but I might be wrong) that there is a lot of unnecessary complexity in different places which could be cleaned up or simplified. I would be interested to know if you ran some performance tests overall, in a simple fashion of course, usually for that I create some temporary integration tests where I create a bunch of tests data (configurable) to run against. otherwise good work man

adrien2p:
Question, it seems the tests pipeline did not run have you run the product module integration tests locally?

integration-tests/modules/__tests__/product/workflows/product-options-removal-validation.spec.ts
fPolic: no but all the test with that expression have a single product option

PR #14222, opened 2025-12-05
RFC - feat: add event emission for product category updates in `batchLinkProductsToCategoryWorkflow`

adevinwild:
## Summary
Emit the `product-category.updated` event when we add products in batch to a category

Added a new step `emitEventStep` in the `batchLinkProductsToCategoryWorkflow`

This will allow devs to listen to a product category update, allowing to revalidate storefront cached products for example

Added the `emitEventStep` with the `ProductCategoryWorkflowEvents.UPDATED` event name

Locally with a dev server :

```
http:    POST /admin/product-categories/pcat_01KBD9410TDE304R6F4APZYS7F/products ← http://localhost:9000/app/categories/pcat_01KBD9410TDE304R6F4APZYS7F/products (200) - 54.302 ms
info:    Processing product-category.updated which has 0 subscribers
```

CleanShot 2025-12-05 at 09 51 54

---

## Examples

- Add product(s) to category

---

## Checklist

- [X] I have added a **changeset** for this PR
- [X] The changes are covered by relevant **tests**
- [] I have linked the related issue(s) if applicable

---

## Additional Context

Related to RFC
https://github.com/medusajs/medusa/discussions/14223

### DIFF
--- a/packages/core/core-flows/src/product/workflows/batch-products-in-category.ts
+++ b/packages/core/core-flows/src/product/workflows/batch-products-in-category.ts
@@ -1,6 +1,10 @@
 import type { ProductCategoryWorkflow } from "@medusajs/framework/types"
-import { WorkflowData, createWorkflow } from "@medusajs/framework/workflows-sdk"
+import {
+  WorkflowData, createWorkflow
+} from "@medusajs/framework/workflows-sdk"
 import { batchLinkProductsToCategoryStep } from "../steps/batch-link-products-in-category"
+import { emitEventStep } from "../../common"
+import { ProductCategoryWorkflowEvents } from "@medusajs/framework/utils"
 
 export const batchLinkProductsToCategoryWorkflowId =
   "batch-link-products-to-category"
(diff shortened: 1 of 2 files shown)


adevinwild:
bump

packages/core/core-flows/src/product/workflows/batch-products-in-category.ts
```diff
@@ -30,6 +34,10 @@ export const batchLinkProductsToCategoryWorkflow = createWorkflow(
     // eslint-disable-next-line max-len
     input: WorkflowData<ProductCategoryWorkflow.BatchUpdateProductsOnCategoryWorkflowInput>
   ): WorkflowData<void> => {
-    return batchLinkProductsToCategoryStep(input)
+    batchLinkProductsToCategoryStep(input)
+    emitEventStep({
+      eventName: ProductCategoryWorkflowEvents.UPDATED,
```
adrien2p, line 39: I am wondering if it should be this event, I am afraid people might thing the category itself have been updated wdyt? Otherwise it could be product update? 🤔 I am not sure though

adevinwild: @adrien2p  Initially, that was the goal!🙌 
My mental model was that products were linked to a category, and therefore, once products were added to a category, that category was considered updated; from the storefront perspective, this is quite convenient for invalidating data

**But** I agree with you; it could have been an event triggered on the product side too, and I would have invalidated the categories linked to that product as well

The problem, though, is that I wouldn't have known which category was added (_precisely_) in order to invalidate the right category page without invalidating all the others 🤔

Not sure which side should be the owner in that situation, depends on the POV

adrien2p: Yes I agree that it depends 😅 @olivermrbl what are your takes here? I think we can go with it but want your input

NicolasGorga:
@olivermrbl would be good to have your take on this. I agree that semantically it is not entirely correct to use the category updated event, as this has to do more with the attachment/detachment of linked records. Would it be better to  add these new methods?

PR #14225, opened 2025-12-05
chore(utils): currency epsilon

carlos-r-l-rodrigues:
## Summary

Use currency decimal precision to calculate some totals and payment statuses.

Very small numbers, due to big number precision calculation, will lead to odd numbers displayed.

At the moment we use the `defaultCurrencies` to get the decimal precision of a given `currency_code` to determine the epsilon to consider the value as zero


adrien2p (approved):
LGTM, does that mean that now the expsilon correspond to the currency decimal dynamically right?

carlos-r-l-rodrigues:
> LGTM, does that mean that now the expsilon correspond to the currency decimal dynamically right?

Yes, but only for the very last totals calculation of currency values.

PR #14226, opened 2025-12-05
feat(): sync cart translation synced

adrien2p:
ref: https://github.com/medusajs/medusa/pull/14189

  **Summary**

  This PR extends the translation module to support automatic translation syncing for cart line items based on the cart's locale.

  Key changes:
  - Added locale field to the Cart model to store the cart's locale preference
  - Created new workflow steps:
    - getTranslatedLineItemsStep - Translates line items when adding to cart or creating a cart
    - updateCartItemsTranslationsStep - Re-translates all cart items when the cart's locale changes
  - Integrated translation logic into cart workflows:
    - createCartWorkflow - Applies translations to initial line items
    - addToCartWorkflow - Applies translations when adding new items
    - updateCartWorkflow - Re-translates all items when locale_code is updated
    - refreshCartItemsWorkflow - Maintains translations during cart refresh
  - Added applyTranslationsToItems utility to map variant/product/type/collection translations to line item fields (title, subtitle, description, etc.)


NicolasGorga:
Cookin' 👨‍🍳

packages/modules/cart/src/models/cart.ts
```diff
@@ -12,6 +12,7 @@ const Cart = model
     sales_channel_id: model.text().nullable(),
     email: model.text().nullable(),
     currency_code: model.text(),
+    locale_code: model.text().nullable(),
```
olivermrbl: suggestion: `locale` would probably suffice

yarn.lock
```diff
@@ -5,10 +5,10 @@ __metadata:
   version: 6
   cacheKey: 8c0
 
-"@acemir/cssom@npm:^0.9.19":
-  version: 0.9.19
-  resolution: "@acemir/cssom@npm:0.9.19"
-  checksum: baef216086b4c6d32ac02fff6621e5343969710d8262be3a84494564754424f60d3add8313c05a76465310e02bf72faa65abcb9febee15c2dc01d9b08c044d5a
+"@acemir/cssom@npm:^0.9.23":
```
olivermrbl, line 8: question: did we need all of these updates? 🤔

packages/core/core-flows/src/cart/workflows/refresh-cart-items.ts
```diff
@@ -233,6 +243,28 @@ export const refreshCartItemsWorkflow = createWorkflow(
       },
     })
 
+    // TODO: Currently the prepareLineItemData function is overriding the translated item data
```
olivermrbl: question: not sure I understand this TODO?

olivermrbl (approved):
Nice, this LGTM

packages/core/core-flows/src/cart/workflows/refresh-cart-items.ts
adrien2p: This correspond to what I ve asked you about on slack, currently, this function override all item fields with variant/product info. I would have expected that we first get them on item and use product/variant info otherwise. Which means that instead of translating a single item when it gets added for example, we always need to re translate everything otherwise it will be reset to the original data. I would prefer not to always have to re translate all items if we can avoid it.

We can think about an approach to solve this (I have a few in mind)

yarn.lock
adrien2p: I ve just re ran yarn, but since the version constraints is starting with a carret it means that we accept patch upgrade automatically, which is fine. Did you see something else?

packages/core/core-flows/src/cart/utils/apply-translations-to-items.ts
```diff
@@ -0,0 +1,73 @@
+import { ProductVariantDTO } from "@medusajs/framework/types"
+
+const productFields = [
```
adrien2p: Todo: rename file to apply-translation-to-item. Use fields mapping instead of fields array plus loop conditions

packages/modules/cart/src/models/cart.ts
NicolasGorga: Will we change this @adrien2p? Would need to change `store_locale` model too if so to align it, currently `locale_code` there as well.

adrien2p: Yes we will 👌

packages/medusa/src/api/store/carts/validators.ts
```diff
@@ -23,6 +23,7 @@ export const CreateCart = z
     sales_channel_id: z.string().nullish(),
     promo_codes: z.array(z.string()).optional(),
     metadata: z.record(z.unknown()).nullish(),
+    locale: z.string().nullish().optional(),
```
olivermrbl: nit(non-blocking): think `nullish` is already optional

packages/medusa/src/api/store/carts/validators.ts
```diff
@@ -53,6 +54,7 @@ export const UpdateCart = z
     sales_channel_id: z.string().nullish(),
     metadata: z.record(z.unknown()).nullish(),
     promo_codes: z.array(z.string()).optional(),
+    locale: z.string().nullish().optional(),
```
olivermrbl: same here

olivermrbl (approved):
LGTM, strong work

olivermrbl:
Feel free to merge as you please

carlos-r-l-rodrigues (approved):

adrien2p:

PR #14243, opened 2025-12-08
feat: order export and upload stream

carlos-r-l-rodrigues:
## Summary

This PR introduces streaming file upload capabilities to the File Module and Order export functionality on the dashboard.

The file providers (S3 and Local) needed a standardized way to handle stream-based write operation.

Orders can be exported from the dashboard.

* Added `getUploadStream` method to the IFileProvider interface

Exporting a large number of orders can consume significant memory. By streaming the CSV generation and upload, we avoid loading the entire dataset or file content into memory, preventing potential OOM crashes.

Integration tests and manual testing


packages/modules/providers/file-local/integration-tests/__tests__/services.spec.ts
```diff
@@ -0,0 +1,117 @@
...
+
+jest.setTimeout(10000)
+
+describe("Local File Plugin", () => {
+  let localService: LocalFileService
+  let fixtureImagePath: string
+  let uploadDir: string
+
+  beforeAll(async () => {
+    fixtureImagePath =
```
adrien2p: suggestion: should we use the `FileSystem` class util?

packages/modules/providers/file-local/integration-tests/__tests__/services.spec.ts
```diff
@@ -0,0 +1,117 @@
...
+        backend_url: "http://localhost:9000/static",
+      }
+    )
+  })
+
+  afterAll(async () => {
+    await fs.rm(uploadDir, { recursive: true, force: true })
+  })
+
+  it(`uploads, reads, and then deletes a public file successfully`, async () => {
```
adrien2p: suggestion: if we can keep all test case starting with `should` :)

adrien2p (approved):
LGTM 🎉

olivermrbl:
Epic stuff! 

I just noticed the `Export` button doesn't exist when you use the feature flagged "View Configurations". Can we also add it there?  🙏

carlos-r-l-rodrigues:
@olivermrbl, export button is there when `MEDUSA_FF_VIEW_CONFIGURATIONS` is true.

olivermrbl:
LGTM

olivermrbl:
Would be cool to allow field selection for the export using the view configurations, but we can do that in a follow-up

olivermrbl (approved):

PR #14244, opened 2025-12-08
chore(): Move event bus local logging

adrien2p:
**what**
Log only when actually processing events


packages/modules/event-bus-local/src/services/event-bus-local.ts
```diff
@@ -35,11 +35,28 @@ export default class LocalEventBusService extends AbstractEventBusModuleService
     // eslint-disable-next-line prefer-rest-params
     super(...arguments)
 
-    this.logger_ = logger
+    this.logger_ = logger ?? console
     this.eventEmitter_ = eventEmitter
     this.groupedEventsMap_ = new Map()
   }
 
+  static logProcessingEvent(
```
shahednasser: suggestion: why not make this a private method and use the logger class member (instead of passing it as a parameter)?

packages/modules/event-bus-local/src/services/event-bus-local.ts
```diff
@@ -35,11 +35,28 @@ export default class LocalEventBusService extends AbstractEventBusModuleService
...
+    options: Record<string, unknown> = {},
+    totalSubscribers: number,
+    logger: Logger
+  ) {
+    if (
+      totalSubscribers &&
+      !options?.internal &&
+      !eventData.options?.internal
+    ) {
+      logger?.info(
```
shahednasser: nit: `?` is not necessary here
```suggestion
      logger.info(
```

shahednasser (approved):

adrien2p:
@olivermrbl are you alright merging that one? I think iti is fine

olivermrbl (approved):

PR #14248, opened 2025-12-08
feat: validate product option (value) deletion

fPolic:
**What**
- prevent option deletion if there are associated products
- prevent option value deletion if there is associated product_product_option link


packages/modules/product/src/services/product-module-service.ts
```diff
@@ -1100,6 +1101,51 @@ export default class ProductModuleService
       )
     }
 
+    // Check if any option values are being removed and if they're associated with products
+    for (const opt of data) {
+      if (opt.values !== undefined) {
```
carlos-r-l-rodrigues:
```suggestion
      if (!isDefined(opt.values)) {
         continue
      }
```

packages/modules/product/src/services/product-module-service.ts
```diff
@@ -1100,6 +1101,51 @@ export default class ProductModuleService
...
+        const dbOption = dbOptions.find(({ id }) => id === opt.id)
+        if (dbOption) {
```
carlos-r-l-rodrigues:
```suggestion
        if (!dbOption) {
           continue
        }
```

packages/modules/product/src/services/product-module-service.ts
```diff
@@ -1100,6 +1101,51 @@ export default class ProductModuleService
...
+
+          for (const existingValue of dbOption.values || []) {
+            if (!newValues.has(existingValue.value)) {
+              removedValueIds.push(existingValue.id)
+            }
+          }
+
+          if (removedValueIds.length > 0) {
+            const productProductOptionValues =
+              await this.productProductOptionValueService_.list(
```
carlos-r-l-rodrigues: Could we just store `removedValueIds`, and check the db outside the main loop to avoid the db call every iteration?

packages/modules/product/src/services/product-module-service.ts
```diff
@@ -1175,6 +1221,48 @@ export default class ProductModuleService
...
+    TReturnableLinkableKeys extends string = string
+  >(
+    primaryKeyValues: string | object | string[] | object[],
+    config?: SoftDeleteReturn<TReturnableLinkableKeys>,
+    @MedusaContext() sharedContext: Context = {}
+  ): Promise<Record<string, string[]> | void> {
+    const optionIds = Array.isArray(primaryKeyValues)
+      ? primaryKeyValues.map((v) => (typeof v === "string" ? v : (v as any).id))
+      : [
+          typeof primaryKeyValues === "string"
```
carlos-r-l-rodrigues: `isString`

packages/modules/product/src/services/product-module-service.ts
fPolic: 🤦 good point

olivermrbl (approved):
LGTM

PR #14262, opened 2025-12-09
chore(): Accept an extra agument 'all-or-nothing' on the migrate command

adrien2p:
**What**
Accept an extra option on the 'db:migrate' command named 'all-or-nothing' which would basically revert all executed migration from all modules during that run if one failed.


adrien2p:
Cc @olivermrbl @shahednasser just so you know. I ve also used the same naming as orm's

olivermrbl (approved):

carlos-r-l-rodrigues (approved):

shahednasser (approved):

PR #14264, opened 2025-12-09
fix(core-flows): conditionally create customer on order email update if unset

NicolasGorga:
## Summary

Conditionally find or create a customer when an order's email is updated and this was previously unset. This allows to cover more flows, like for example, if an order is created from a sync to an external system that doesn't provide customer information and this should be filled manually afterwards, we now make it possible to implement said flow.

If the email is already set, we don't create nor link a customer to the order. This flow is already possible using the transfer order workflow.

If an order doesn't have an email set, calling the update order endpoint and passing the email only updates the field, but the customer linked to the order is still `null`. This will cause issues in the Admin UI when clicking on the customer email to navigate to the customer detail, since no customer exists. It also makes it impossible to transfer the order, as source is `null`

Take the same approach as cart flows where we find or create a customer when one is not linked to the cart but we have an email. In the `updateOrderWorkflow` we check if the update input contains an `email` and the `order.email` is unset and in this case, we find or create a customer matching the input email and link it to the order.

Integration tests.

---

## Examples

```ts
// Example usage
```

---

## Checklist

---

## Additional Context

closes CORE-1308

### DIFF
--- a/integration-tests/http/__tests__/order/admin/order.spec.ts
+++ b/integration-tests/http/__tests__/order/admin/order.spec.ts
@@ -165,7 +165,10 @@ medusaIntegrationTestRunner({
         order = seeder.order
 
         order = (
-          await api.get(`/admin/orders/${order.id}?fields=+email`, adminHeaders)
+          await api.get(
+            `/admin/orders/${order.id}?fields=+email,+customer_id`,
+            adminHeaders
+          )
         ).data.order
       })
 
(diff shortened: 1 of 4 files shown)


carlos-r-l-rodrigues (approved):
LGTM

integration-tests/http/__tests__/order/admin/order.spec.ts
```diff
@@ -477,6 +480,215 @@ medusaIntegrationTestRunner({
...
+              email: newEmail,
+            },
+            adminHeaders
+          )
+
+          expect(response.data.order.email).toBe(newEmail)
+          expect(response.data.order.customer_id).toBeDefined()
+
+          const customersAfter = await customerService.listCustomers()
+          expect(customersAfter.length).toBe(customerCountBefore + 1)
```
adrien2p, line 539: Q: do we need this list check since we retrieve it just after to check the email?

adrien2p (approved):
LGTM

PR #14265, opened 2025-12-09
feat: product option UI + storefront API changes

fPolic:
**What**

**_UI_**
- create and select existing options on product create
- fix product loading for options
- fix combobox onOptionCreate logic
- disallow values creation in product create flow
- fix placeholder
- fix badge

**_API_**
- allow filtering products by option in API
- product options storefront endpoints
- filter products by option values


packages/admin/dashboard/src/components/inputs/combobox/combobox.tsx
```diff
@@ -240,7 +241,13 @@ const ComboboxImpl = <T extends Value = string>(
...
       setOpen={handleOpenChange}
       selectedValue={selectedValues}
-      setSelectedValue={(value) => handleValueChange(value as T)}
+      setSelectedValue={(value) => {
+        // TMP: fix issue where sometimes first element of the array is empty string
+        if (Array.isArray(value) && value[0] === "") {
+          value.shift()
+        }
+        handleValueChange(value as T)
+      }}
```
olivermrbl, line 250: question: anything we should follow-up with here?

packages/admin/dashboard/src/i18n/translations/en.json
```diff
@@ -859,6 +859,11 @@
     "values": {
       "header": "Values"
     },
+    "products": {
+      "list": {
+        "noRecords": "Product that have this option will show up here."
```
olivermrbl: nit: typo
```suggestion
        "noRecords": "Products with this option will show up here."
```

olivermrbl: Maybe we could even do:
```
There are no products with this option
```

packages/modules/product/src/services/product-module-service.ts
```diff
@@ -314,6 +322,13 @@ export default class ProductModuleService
     config?: FindConfig<ProductTypes.ProductDTO>,
     sharedContext?: Context
   ): Promise<[ProductTypes.ProductDTO[], number]> {
+    const { filters: normalizedFilters, shouldReturnEmpty } =
+      await this.applyOptionValueFilter_(filters, sharedContext)
+
+    if (shouldReturnEmpty) {
+      return [[], 0]
+    }
```
olivermrbl, line 330: question: does this account for the case where there is no option value filter (and therefore `shouldReturnEmpty` is true), but we still want to retrieve products? I am not sure I understand this if-statement; it feels overly specific

olivermrbl (approved):
LGTM

packages/admin/dashboard/src/components/inputs/combobox/combobox.tsx
fPolic: yup will create a ticket to revisit this component

packages/modules/product/src/services/product-module-service.ts
fPolic: When there is no options filter `shouldReturnEmpty` is false. `shouldReturnEmpty` flag is only used to signal when there are passed option value filters but there are no matches so we want to return [].

PR #14266, opened 2025-12-10
chore(): Add translations/locale integration tests and fix locale endpoint

adrien2p:
**What**
- Fix locale end point (throw if key does not exists require an id to be provided)
- Add integration tests


olivermrbl (approved):
looks nice!

PR #14272, opened 2025-12-10
fix(translation): add index file for models

shahednasser:
Add an index file to the Translation Module that exports the data models. This follows existing conventions and is necessary for how we generate the data model references in the docs


NicolasGorga (approved):
LGTM

adrien2p (approved):

PR #14276, opened 2025-12-11
fix(): Improve store currency readonly link

adrien2p:
**What**
Improve readonly link such that a user can do:
`supported_currencies.currency.symbol_native`


packages/modules/link-modules/src/definitions/readonly/store-locale.ts
```diff
@@ -20,7 +20,7 @@ export const StoreLocales: ModuleJoinerConfig = {
         foreignKey: "locale_code",
         alias: "locale",
         args: {
-          methodSuffix: "StoreLocales",
+          methodSuffix: "Locales",
```
shahednasser, line 23: q: should this be here?

shahednasser (approved):

packages/modules/link-modules/src/definitions/readonly/store-locale.ts
adrien2p: yeah I quickly fixed it as the other one I ve merged to quickly

adrien2p: do you prefer to split the pr?

shahednasser: No fine by me, just checking

PR #14297, opened 2025-12-12
chore: sync Figma resources

fPolic:
**What**
- sync Figma resources


packages/design-system/icons/src/components/circle-minus-solid.tsx
```diff
@@ -1,6 +1,6 @@
 import * as React from "react"
 import type { IconProps } from "../types"
-const CircleMinus = React.forwardRef<SVGSVGElement, IconProps>(
```
fPolic, line 3: this one is deleted but not used internally

olivermrbl (approved):

PR #14299, opened 2025-12-12
feat(): Translation statistics

adrien2p:
### Overview

  This PR adds a translation statistics API that allows administrators to track translation completion progress across entity types (products, variants, categories, etc.) and locales.

###  Key Features

  1. New GET /admin/translations/statistics endpoint
    - Accepts locales and entity_types as query parameters
    - Returns statistics per entity type with both aggregated and per-locale breakdowns
  2. Statistics output structure:
```json
  {
    "product": {
      "expected": 16,
      "translated": 3,
      "missing": 13,
      "by_locale": {
        "fr-FR": { "expected": 8, "translated": 2, "missing": 6 },
        "de-DE": { "expected": 8, "translated": 1, "missing": 7 }
      }
    }
  }
```
  4. Performance optimization - Uses a precomputed translated_field_count column to avoid expensive JSONB iteration at query time


NicolasGorga:
Looking good

packages/medusa/src/api/admin/translations/statistics/route.ts
```diff
@@ -0,0 +1,75 @@
...
+            cache: { enable: true },
+          }
+        )
+        .catch((e) => {
+          const normalizedMessage = e.message.toLowerCase()
+          if (
+            normalizedMessage.includes("service with alias") &&
+            normalizedMessage.includes("was not found")
+          ) {
+            return { metadata: { count: 0 } }
```
NicolasGorga, line 50: If we want to return the count as 0 for an entity which wasn't found, should we return  `{ entityType, count: 0 }`. Since later we are doing 

```js
for (const { entityType, count } of entityCounts) {
    entities[entityType] = { count }
  }
  ```
  
  This would resolve to: 
  
  ```json
 {
"entities": {
"metadata": {
"count": 0
}
// Other ....
/
}
}
  ```

File: packages/modules/translation/src/utils/translatable-fields.ts [hunk 2]
```diff
@@ -0,0 +1,8 @@
+export const PRODUCT_TRANSLATABLE_FIELDS = [
+  "title",
+  "description",
```
NicolasGorga, line 3: Shall we add `material` here as well?

packages/medusa/src/api/admin/translations/statistics/route.ts
adrien2p: nope because here it is the result of the `query.graph` which returns the `metadata` and just after the graph query we return `entityType` and `count`

packages/modules/translation/src/utils/translatable-fields.ts
adrien2p: oh missed it, yes I ll add it

packages/core/types/src/http/translations/admin/responses.ts
```diff
@@ -33,3 +33,58 @@ export interface AdminTranslationsBatchResponse {
...
+
+/**
+ * Statistics for an entity type.
+ */
+export interface AdminTranslationEntityStatistics
+  extends AdminTranslationLocaleStatistics {
+  /**
+   * Breakdown of statistics by locale.
+   */
+  byLocale: Record<string, AdminTranslationLocaleStatistics>
```
olivermrbl: suggestion: I think HTTP responses are typically snake_case regardless of their level in the object

packages/core/types/src/http/translations/admin/responses.ts
```diff
@@ -33,3 +33,58 @@ export interface AdminTranslationsBatchResponse {
...
+  /**
+   * A mapping of entity types to their translatable field names.
+   *
+   * @example
+   * {
+   *   "product": ["title", "description", "subtitle", "status"],
+   *   "product_variant": ["title", "material"]
+   * }
+   */
+  translatableFields: Record<string, string[]>
```
olivermrbl: suggestion: I think HTTP responses are typically snake_case regardless of their level in the object

packages/medusa/src/api/admin/translations/middlewares.ts
```diff
@@ -29,4 +30,14 @@ export const adminTranslationsRoutesMiddlewares: MiddlewareRoute[] = [
...
     middlewares: [validateAndTransformBody(AdminBatchTranslations)],
   },
+  {
+    method: ["GET"],
+    matcher: "/admin/translations/statistics",
+    middlewares: [validateAndTransformQuery(AdminTranslationStatistics, {})],
+  },
+  {
+    method: ["GET"],
+    matcher: "/admin/translations/translatable-fields",
```
olivermrbl: suggestion:(non-blocking): do we need the "translatable" prefix here?
```
GET /admin/translations/fields
```

olivermrbl (approved):
Lgtm, nice!

packages/medusa/src/api/admin/translations/middlewares.ts
adrien2p: I was thinking, maybe it should be settings, and the response has translatable_fields? To prepare the future?

olivermrbl: Can you show an example of that?

adrien2p: currently:

```ts
// GET /admin/translations/translatable-fields
{ 
  translatableFields: {
    product: ['title, 'subtitle', ...],
    product_variant: ['title', ...]
  }
}
```

suggestion:

```ts
// /admin/translations/settings
{ 
  translatable_fields: {
    product: ['title, 'subtitle', ...],
    product_variant: ['title', ...]
  }
}
```

NicolasGorga: I like that more

olivermrbl: yeah I like that too

adrien2p:

PR #14300, opened 2025-12-12
chore(): Remove default_locale from StoreLocale

NicolasGorga:
## Summary

Remove `default_locale` from Store <> Locale relation

*Please provide answer here*

*Please provide answer here*

*Please provide answer here*

---

## Examples

```ts
// Example usage
```

---

## Checklist

---

## Additional Context


packages/core/framework/src/http/middlewares/apply-locale.ts
```diff
@@ -54,9 +54,8 @@ export async function applyLocale(
   )
 
   if (store?.supported_locales?.length) {
-    req.locale = store.supported_locales.find(
-      (locale) => locale.is_default
-    )?.locale_code
+    // Use the first locale as the default
```
adrien2p: Suggestion: I think this would be wrong, it might select a random locale that can be different at any point. We better remove this entire section and it will return the default values

packages/modules/store/src/services/store-module-service.ts
```diff
@@ -259,24 +257,27 @@ export default class StoreModuleService
...
 
-    let seenDefault = false
-    items.forEach((item) => {
-      if (item.is_default) {
-        if (seenDefault) {
-          throw new MedusaError(
-            MedusaError.Types.INVALID_DATA,
-            `Only one default ${typeName} is allowed`
-          )
+    const hasDefaultProperty = items.some((item) => "is_default" in item)
```
adrien2p: A: why do we have still default check? It is not part of the db anymore nor the types and validator right?

packages/modules/store/src/services/store-module-service.ts
```diff
@@ -259,24 +257,27 @@ export default class StoreModuleService
...
-      }
-    })
+      })
 
-    if (!seenDefault) {
-      throw new MedusaError(
-        MedusaError.Types.INVALID_DATA,
-        `There should be a default ${typeName} set for the store`
-      )
+      if (!seenDefault) {
```
adrien2p: A: do we?

packages/core/framework/src/http/middlewares/apply-locale.ts
NicolasGorga: I agree

packages/modules/store/src/services/store-module-service.ts
NicolasGorga: Because I had consolidated the check to verify we don't have duplicates and that only one default is set for supported currencies/locales. Since now locale doesn't have `is_default`, the part of the check that is common to both is the duplicate one and for the other check, only enforce it when the object contains `is_default` which will be the case just for `supported_currencies` now. wyt?

packages/modules/store/src/services/store-module-service.ts
NicolasGorga: Check my previous comment, but maybe it would be clearer remove this function and perform the duplicate + unique default validation for `supported_currencies` and just the duplicate validation for `supported_locales` spearately.

packages/modules/store/src/services/store-module-service.ts
adrien2p: So here we can clearly see a problem of responsibility, I would separate the validation in to two function or put them back in their respective ones. as we can see that they evolved differently. Just looking at the method name I couldn’t clearly see that it was for both currency and locale. Would you mind doing that? But as you prefer really, I try to follow SOLID when it makes sense

packages/modules/store/src/services/store-module-service.ts
adrien2p: Yes I do agree with you 👌

packages/modules/store/src/services/store-module-service.ts
NicolasGorga: Yeah, i agree with you, this latest change i felt was kind of forceful, so naturally going back to what we had before, of separate functions feels like the right thing. WIll update!

adrien2p (approved):

olivermrbl (approved):

PR #14306, opened 2025-12-15
feat(): Add support for locale to the js sdk

adrien2p:
### overview

Add support for locale to the js-sdk:
- set locale will store in the object and locale store under `medusa_locale`
- getting locale will return it from the local storage if available and fallback to the local object property

### DIFF
--- a/packages/core/js-sdk/src/client.ts
+++ b/packages/core/js-sdk/src/client.ts
@@ -10,6 +10,7 @@ import {
 } from "./types"
 
 export const PUBLISHABLE_KEY_HEADER = "x-publishable-api-key"
+export const LOCALE_STORAGE_KEY = "medusa_locale"
 
 // We want to explicitly retrieve the base URL instead of relying on relative paths that differ in behavior between browsers.
 const getBaseUrl = (passedBaseUrl: string) => {
@@ -112,6 +113,18 @@ export class Client {
   private DEFAULT_JWT_STORAGE_KEY = "medusa_auth_token"
   private token = ""
 
+  private locale_ = ""
+
+  get locale() {
+    if (hasStorage("localStorage")) {
+      const storedLocale = window.localStorage.getItem(LOCALE_STORAGE_KEY)
+      if (storedLocale) {
+        return storedLocale
+      }
+    }
+    return this.locale_
+  }
+
   constructor(config: Config) {
     this.config = { ...config, baseUrl: getBaseUrl(config.baseUrl) }
     const logger = config.logger || {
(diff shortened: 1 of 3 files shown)


packages/core/js-sdk/src/client.ts
```diff
@@ -226,10 +250,12 @@ export class Client {
       // We always want to fetch the up-to-date JWT token before firing off a request.
       const headers = new Headers(defaultHeaders)
       const customHeaders = {
+        "content-language": this.locale,
```
olivermrbl, line 253: thought: Shouldn't we only set the locale, if it is not explicitly passed? Otherwise, we could be overriding it.

Something like:
```ts
setRequestLocale(url, headers) {
    const carriesLocale =
      url.searchParams.has("locale") ||
      headers.get("content-language")

    if (!carriesLocale && this.locale) {
      url.searchParams.set("locale", this.locale)
    }

    return url
  }
```

packages/core/js-sdk/src/client.ts
```diff
@@ -112,6 +113,18 @@ export class Client {
   private DEFAULT_JWT_STORAGE_KEY = "medusa_auth_token"
   private token = ""
 
+  private locale_ = ""
+
+  get locale() {
```
NicolasGorga, line 118: Suggestion: I don't necessarily know if the speed is a thing here, but wouldn't it be better to check first:

```js
if (this.locale_) {
   return this.locale_
}
```

And otherwise try to get it from localStorage if available?

packages/core/js-sdk/src/client.ts
adrien2p: oh you mean for the query param, yes I ll add that, but the header itself will get overridden already if the user provide one for it

NicolasGorga: @olivermrbl Because of the `applyMiddleware` locale, if the locale is in the query, this will have priority over the header. It is true that here we would be overriding the header one.

packages/core/js-sdk/src/client.ts
adrien2p: the speed is not an issue here for this type of access, though, I did that on purpose just to always prioritize local storage over memory, so local storage always take precedence in case there is desync

packages/core/js-sdk/src/client.ts
adrien2p: yes thats true as well 👍

adrien2p: just to summarize, the query param already take precedence over the header, so if specified, we are using the query param, otherwise take the header.

Also, the header is set but if the user pass the same header, then the user header will take precedence as it is applied after this one on purpose to override the default

adrien2p: @olivermrbl I ll wait for your answer before proceeding on anything

olivermrbl: Okay got it, sounds good to me then

olivermrbl (approved):

adrien2p:

PR #14307, opened 2025-12-15
feat(): Add support for store locales end point

adrien2p:
**What**
Add support for `store/locales` which return the configured supported locales

### DIFF
--- a/packages/medusa/src/api/store/locales/route.ts
+++ b/packages/medusa/src/api/store/locales/route.ts
@@ -14,7 +14,15 @@ export const GET = async (req: MedusaRequest, res: MedusaResponse) => {
     },
   })
 
+  const locales = store?.supported_locales.reduce((acc, locale) => {
+    acc.push({
+      code: locale.locale_code,
+      name: locale.locale.name,
+    })
+    return acc
+  }, [])
+
   res.json({
-    locales: store?.supported_locales ?? [],
+    locales,
   })
 }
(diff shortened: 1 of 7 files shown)


packages/medusa/src/api/store/locales/route.ts
```diff
@@ -0,0 +1,20 @@
...
+  } = await query.graph({
+    entity: "store",
+    fields: ["supported_locales.*", "supported_locales.locale.*"],
+    pagination: {
+      take: 1,
+    },
+  })
+
+  res.json({
+    locales: store?.supported_locales ?? [],
```
olivermrbl: thought(non-blocking): should we map it to something a bit more simple?
```
{ code: "en-US", name: "English (US)" }
```

Don't think you need anything else from the locales. Makes it easier to consume in the storefront.

olivermrbl (approved):

packages/medusa/src/api/store/locales/route.ts
adrien2p: you have that in locals[n].locale -> { code, name }, are you saying we should stip the first level entirely?

olivermrbl: Yes exactly. There is no valuable information at the first level

adrien2p: ✅

adrien2p:

PR #14311, opened 2025-12-15
feat(): Add more translatable core entity

adrien2p:
**What**
product type, collection, categories, tags, options, options values, shipping options and type, , return reason translation

### DIFF
--- a/packages/modules/link-modules/src/definitions/readonly/product-translation.ts
+++ b/packages/modules/link-modules/src/definitions/readonly/product-translation.ts
@@ -1,12 +1,8 @@
 import { ModuleJoinerConfig } from "@medusajs/framework/types"
-import {
-  FeatureFlag,
-  MEDUSA_SKIP_FILE,
-  Modules,
-} from "@medusajs/framework/utils"
+import { MEDUSA_SKIP_FILE, Modules } from "@medusajs/framework/utils"
 
 export const ProductTranslation: ModuleJoinerConfig = {
-  [MEDUSA_SKIP_FILE]: !FeatureFlag.isFeatureEnabled("translation"),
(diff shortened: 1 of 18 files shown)


packages/modules/link-modules/src/definitions/readonly/product-translation.ts
```diff
@@ -1,12 +1,8 @@
...
-import {
-  FeatureFlag,
-  MEDUSA_SKIP_FILE,
-  Modules,
-} from "@medusajs/framework/utils"
+import { MEDUSA_SKIP_FILE, Modules } from "@medusajs/framework/utils"
 
 export const ProductTranslation: ModuleJoinerConfig = {
-  [MEDUSA_SKIP_FILE]: !FeatureFlag.isFeatureEnabled("translation"),
+  [MEDUSA_SKIP_FILE]: process.env.MEDUSA_FF_TRANSLATION !== "true",
```
NicolasGorga, line 5: would this incorrectly skip the file if the feature flag was set through medusa-config.ts and not env variable?

adrien2p: the feature flag are not loaded when reaching that part, thats why for now I ve done that otherwise we cant tests with it

adrien2p: but usually feature flags are managed through env, and the feature flag in the config rely on it, otherwise it would require code changes, deploy to production to change one featuer flag. and since it is temporary I thought why not

NicolasGorga: great was curious if it was a timing thing so good to know :)

NicolasGorga (approved):
LGTM

PR #14314, opened 2025-12-16
feat(js-sdk,types,medusa): add list locales store method to JS SDK

shahednasser:
- Add a `sdk.store.locale.list` method to retrieve the supported locales in a store
- Add HTTP types for store locale types
- Use response type in list locales API route
- Other: Add missing locale parameter for the update order HTTP payload type


olivermrbl (approved):
nice

PR #14315, opened 2025-12-16
chore(): Migration generator fix generated import

adrien2p:
**What**

Fix migration generator to properly handle import from medusajs framework instead of mikro orm directly.
also include fixes for existing migrations

### DIFF
--- a/packages/core/utils/src/dal/mikro-orm/mikro-orm-create-connection.ts
+++ b/packages/core/utils/src/dal/mikro-orm/mikro-orm-create-connection.ts
@@ -41,7 +41,12 @@ export class CustomTsMigrationGenerator extends TSMigrationGenerator {
       diff.up.unshift(sql)
     }
 
-    return super.generateMigrationFile(className, diff)
+    let migrationFileContent = super.generateMigrationFile(className, diff)
+    migrationFileContent = migrationFileContent.replace(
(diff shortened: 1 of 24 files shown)


packages/core/utils/src/dal/mikro-orm/mikro-orm-create-connection.ts
```diff
@@ -41,7 +41,12 @@ export class CustomTsMigrationGenerator extends TSMigrationGenerator {
       diff.up.unshift(sql)
     }
 
-    return super.generateMigrationFile(className, diff)
+    let migrationFileContent = super.generateMigrationFile(className, diff)
+    migrationFileContent = migrationFileContent.replace(
+      "import { Migration } from '@mikro-orm/migrations';",
+      "import { Migration } from '@medusajs/framework/mikro-orm/migrations';"
```
olivermrbl: super nit(non-blocking): we typically don't use single quotes nor semi colon 😄

olivermrbl (approved):

packages/core/utils/src/dal/mikro-orm/mikro-orm-create-connection.ts
adrien2p: oups that's my bad :)

PR #14319, opened 2025-12-16
chore(): Apply translation feature flag broaderly

adrien2p:
**What**
Prevent some fields to be queries in the backend when feature flag is not enabled


adrien2p:
@olivermrbl can we merge that one?

PR #14324, opened 2025-12-16
fix(types,medusa): use HTTP types in request type parameters of product options routes

shahednasser:
- Use HTTP types instead of validators in request type arguments of product option routes
- Add missing `metadata` property to HTTP type

### DIFF
--- a/packages/medusa/src/api/store/product-options/[id]/route.ts
+++ b/packages/medusa/src/api/store/product-options/[id]/route.ts
@@ -9,7 +9,7 @@ import {
 } from "@medusajs/framework/http"
 
 export const GET = async (
-  req: AuthenticatedMedusaRequest<HttpTypes.SelectParams>,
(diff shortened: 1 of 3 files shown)


packages/medusa/src/api/store/product-options/[id]/route.ts
```diff
@@ -9,7 +9,7 @@ import {
 } from "@medusajs/framework/http"
 
 export const GET = async (
-  req: AuthenticatedMedusaRequest<HttpTypes.StoreProductOptionParams>,
+  req: AuthenticatedMedusaRequest<HttpTypes.SelectParams>,
```
Comment (fPolic) [inline packages/medusa/src/api/store/product-options/[id]/route.ts]: q: shouldn't this be `req: AuthenticatedMedusaRequest<{}, HttpTypes.SelectParams>,`

Comment (shahednasser) [reply packages/medusa/src/api/store/product-options/[id]/route.ts]: Good point, fixed

fPolic (approved):
lgtm - thanks!

PR #14337, opened 2025-12-17
feat(): Auto generated types generation upon build

adrien2p:
**What**
Auto generate the auto generated types during the build

https://github.com/user-attachments/assets/674c4a86-9c34-4bb6-b7c7-cb167ff305ae

---

> [!NOTE]
> Automatically generates container and GraphQL types during `medusa build` using a new migration-only module loading path.
> 
> - **CLI (build)**:
>   - Runs `generateTypes` before compiling, emitting container and GraphQL types to `.medusa/types`.
>   - New `packages/medusa/src/commands/utils/generate-types.ts` loads plugins/links, validates module names, partially loads app, cleans output dir, and writes types.
> - **Framework / Modules SDK**:
>   - Add `migrationOnly` option to `MedusaApp`/`MedusaAppLoader.load` to partially load modules (skip full loaders) and expose joiner configs/schema.
>   - `MedusaAppLoader.load` conditionally skips shared resource prep when `migrationOnly` is true and forwards the flag to `MedusaApp`.
>   - Minor signature cleanup in `MedusaApp_` to include `migrationOnly` in `MedusaAppOptions`.
> - **Release**:
>   - Patch bumps for `@medusajs/medusa`, `@medusajs/framework`, `@medusajs/modules-sdk`, and `@medusajs/cli`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit c8c3a17b82e03b93cf61ab6adcb3fd6e0a76e703. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


adrien2p:
cc @shahednasser

carlos-r-l-rodrigues (approved):

packages/cli/medusa-cli/src/create-cli.ts
```diff
@@ -466,6 +466,11 @@ function buildLocalCommands(cli, isLocalProject) {
           type: "boolean",
           describe:
             "Only build the admin to serve it separately (outDir .medusa/admin)",
+        }).option("types", {
```
shahednasser: Q: is there a use case where you don't want to generate types? If not I think we should remove this.

shahednasser (approved):

packages/cli/medusa-cli/src/create-cli.ts
adrien2p: when you build for production, the types are runtime does not server any purposes

shahednasser: But you still need the types to avoid errors in compilation.

For example, i have this line:

```
const algoliaModuleService = container.resolve(ALGOLIA_MODULE)
```

And I run the build command with `types=false`, I get a type error because TypeScript can't resolve the type of the service.

So, even if you're building for production, you need to generate the types to avoid the compilation errors related to generated types.

adrien2p:
cc @olivermrbl

PR #14338, opened 2025-12-17
fix(core-flows): fix type of getTranslatedLineItemsStep

shahednasser:
Fix the return type of the `getTranslatedLineItemsStep`, otherwise our reference generator can't pick up the step's typing correctly

### DIFF
--- a/packages/core/core-flows/src/common/steps/get-translated-line-items.ts
+++ b/packages/core/core-flows/src/common/steps/get-translated-line-items.ts
@@ -43,4 +43,6 @@ const step = createStep(
  */
 export const getTranslatedLineItemsStep = <T>(
   data: GetTranslatedLineItemsStepInput<T>
-): ReturnType<StepFunction<any, T[]>> => step(data)
+): ReturnType<StepFunction<any, T[]>> => step(data) as unknown as ReturnType<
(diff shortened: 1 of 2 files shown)


packages/core/core-flows/src/common/steps/get-translated-line-items.ts
```diff
@@ -43,4 +43,6 @@ const step = createStep(
  */
 export const getTranslatedLineItemsStep = <T>(
   data: GetTranslatedLineItemsStepInput<T>
-): ReturnType<StepFunction<any, T[]>> => step(data)
+): ReturnType<StepFunction<any, T[]>> => step(data) as unknown as ReturnType<
```
adrien2p, line 46: Q: we have to cast it for the reference to be picked up?

shahednasser: Yes it's mainly because the inner `step` function isn't exported, so the tools we use can't pick up the step or its type. By casting it we can pick up the type. We do the same for `useQueryGraphStep`.

The alternative would be to remove the step separation and just define the step in `getTranslatedLineItemsStep`, but not sure what the context of this separation is

adrien2p (approved):

PR #14355, opened 2025-12-18
feat(): Translation settings + user configuration + admin hook and js sdk + dashboard

adrien2p:
### what

  This PR introduces database-backed translatable field configuration for the translation module, allowing dynamic management of which fields are translatable per entity type. This is also a preliminary step before fully configurable translation from the UI

  - New Settings model
  - Config loader updates
    - Merges hardcoded defaults with user-provided module options
    - Upserts settings to database on module initialization
  - Module service updates
    - Updated getTranslatableFields() method to fetch config from database
    - Added ensureReferenceFieldInConfig() to fix field selection issues in case reference is not requested
    - listTranslations, listAndCountTranslations, and retrieveTranslation now filter returned translations based on configured fields
  - New /admin/translations/entities endpoint
    - Returns paginated entities with only their translatable fields
    - Useful for translation UIs to show what content needs translation
  - SDK updates
    - JS SDK: Added sdk.admin.translation.entities() method
    - Admin dashboard: Added useTranslationEntities hook and query keys
  - Dashboard update to consume data from the new endpoint
 
### Add support for `ModulesOptions`
type augmentation to allow modules to define expected types in medusa config defineConfig

https://github.com/user-attachments/assets/246b2bbc-0f24-4100-a34a-0a58bb929a0c

**Loader Updates**

  - Config loader now merges hardcoded defaults with user-provided options
  - Upserts settings to database on startup
  - User can extend translatable fields via module options:
 ```ts
  {
    entities: [
      { type: "custom_entity", fields: ["title", "description"] }
    ]
  }
```

### DIFF
--- a/packages/medusa/src/api/admin/translations/settings/route.ts
+++ b/packages/medusa/src/api/admin/translations/settings/route.ts
@@ -25,12 +25,12 @@ export const GET = async (
   const translationService = req.scope.resolve<ITranslationModuleService>(
     Modules.TRANSLATION
   )
-  const translatable_fields = await translationService.getTranslatableFields(
+  const translatableFields = await translationService.getTranslatableFields(
     req.validatedQuery.entity_type
   )
 
   res.json({
-    translatable_fields,
+    translatable_fields: translatableFields,
   })
 }
 
--- a/packages/medusa/src/api/admin/translations/settings/route.ts
+++ b/packages/medusa/src/api/admin/translations/settings/route.ts
@@ -25,7 +25,7 @@ export const GET = async (
   const translationService = req.scope.resolve<ITranslationModuleService>(
     Modules.TRANSLATION
   )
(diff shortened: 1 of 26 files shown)

### CODE REVIEW

File: packages/medusa/src/api/admin/translations/settings/route.ts [hunk 1]
```diff
@@ -25,7 +25,7 @@ export const GET = async (
   const translationService = req.scope.resolve<ITranslationModuleService>(
     Modules.TRANSLATION
   )
-  const translatable_fields = translationService.getTranslatableFields(
+  const translatable_fields = await translationService.getTranslatableFields(
```
olivermrbl: nit: should be camelCased

olivermrbl (approved):
Clean!

NicolasGorga (commented):
Fire

PR #14358, opened 2025-12-18
feat(translation,core-flows): translate remaining core entities and sync shipping option <> method translations

NicolasGorga:
Add remaining core entities translation configuration:

- shipping_option: name
- region: name
- customer_group: has name
- shipping_option_type: label and description
- tax_rate: name

Sync shipping_option translation to shipping_method


packages/core/core-flows/src/order/steps/update-order-shipping-methods-translations.ts
```diff
@@ -0,0 +1,70 @@
...
+  updateOrderShippingMethodsTranslationsStepId,
+  async (
+    data: UpdateOrderShippingMethodsTranslationsStepInput,
+    { container }
+  ) => {
+    const isTranslationEnabled = FeatureFlag.isFeatureEnabled("translation")
+    const query = container.resolve(ContainerRegistrationKeys.QUERY)
+    const orderModuleService = container.resolve(Modules.ORDER)
+
+    if (isTranslationEnabled && !!data.locale && data.shippingMethods.length) {
```
olivermrbl: suggestion(non-blocking): I think an early return would a bit easier to read

packages/core/core-flows/src/common/steps/get-translated-shipping-option.ts
```diff
@@ -0,0 +1,28 @@
...
+  shippingOptions: ShippingOptionDTO[]
+  locale: string
+}
+
+export const getTranslatedShippingOptionsStep = createStep(
+  getTranslatedShippingOptionsStepId,
+  async (data: GetTranslatedShippingOptionsStepInput, { container }) => {
+    const isTranslationEnabled = FeatureFlag.isFeatureEnabled("translation")
+
+    if (isTranslationEnabled && !!data.locale && data.shippingOptions.length) {
```
olivermrbl: Same here

olivermrbl (approved):
LGTM, nice work!

packages/core/core-flows/src/common/steps/get-translated-shipping-option.ts
NicolasGorga: Removed the check here since it was redundant. We already check if the feature flag is enabled inside the `applyTranslations` util and also handle missing locale and empty array of objects to translate.

packages/core/core-flows/src/order/steps/update-order-shipping-methods-translations.ts
NicolasGorga: Done

olivermrbl (approved):
LGTM, strong work!

PR #14381, opened 2025-12-22
fix(medusa): allow passing null to cart API routes

shahednasser:
Allow passing `null` as the locale for cart API route, as `optional` only allow passing `undefined`.


olivermrbl (approved):

adrien2p (approved):
I believe it is the same for orders etc

PR #14395, opened 2025-12-24
feat: add Medusa Cloud OAuth provider

peterlgh7:
## Summary

Added a Medusa Cloud auth provider that gets registered automatically in Cloud.

To make it easier for Cloud users to log into their deployed environments Medusa admin. They can just use their Cloud credentials now.

A new auth provider, using OAuth2 and very similar to the Google and Github ones

Added module tests

---

## Checklist

## Future work
We should move all Cloud related code to a plugin and install it by default, otherwise that code is unlocalized, and we pollute the codebase. Will do in another PR.

---

> [!NOTE]
> Introduces Cloud-based authentication across backend, config, and admin UI.
> 
> - Adds `cloud` OAuth provider in `@medusajs/auth` (auto-registers when Cloud options present; supports `disabled`), including token exchange and identity creation; extends types/options.
> - Extends project config with Cloud OAuth fields (authorize/token/callback/disabled) and injects provider options into modules; enables `cloud` user auth method when running in Cloud.
> - New endpoints: `GET /cloud/auth` (reports enabled) and `POST /cloud/auth/users` (creates or links admin user; uses `setAuthAppMetadataWorkflow`); adjusts auth middleware to allow unregistered actors and include `user_metadata`.
> - Admin dashboard: adds hooks and `CloudAuthLogin` to initiate OAuth, handle callback, create user if needed, and refresh token; adds i18n strings.
> - Core flows: exports `set-auth-app-metadata` workflow.
> - Tests: integration tests for provider authenticate/callback and HTTP user creation/linking; changeset added.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 11da8a6e0054970adf212928f0433ad5034bcee9. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


packages/core/utils/src/common/define-config.ts
```diff
@@ -504,6 +526,24 @@ function applyCloudOptionsToModules(
...
+        }
+        module.options = {
+          cloud: {
+            oauth_authorize_endpoint: config.oauthAuthorizeEndpoint,
+            oauth_token_endpoint: config.oauthTokenEndpoint,
+            environment_handle: config.environmentHandle,
+            sandbox_handle: config.sandboxHandle,
+            api_key: config.apiKey,
+            callback_url: callbackUrl,
+            disabled: config.oauthDisabled,
```
sradevski, line 542: How can users manage this flag?

sradevski (approved):
LGTM 👍

packages/core/utils/src/common/define-config.ts
peterlgh7: From their medusa-config, or even passing the env var MEDUSA_CLOUD_OAUTH_DISABLED

packages/admin/dashboard/src/routes/login/components/cloud-auth-login.tsx
```diff
@@ -0,0 +1,105 @@
...
+import { useMutation } from "@tanstack/react-query"
+import { useEffect, useMemo } from "react"
+import { decodeToken } from "react-jwt"
+import { useNavigate, useSearchParams } from "react-router-dom"
+import { useCreateCloudAuthUser } from "../../hooks/api/cloud"
+import { sdk } from "../../lib/client"
+
+const CLOUD_AUTH_PROVIDER = "cloud"
+
+export const CloudAuthLogin = () => {
```
sradevski, line 12: NIT: This is the path where the user gets redirected with the code? Maybe the name is a bit misleading if that's the case

packages/admin/dashboard/src/routes/login/components/cloud-auth-login.tsx
```diff
@@ -0,0 +1,105 @@
...
+        throw new Error("Authentication failed")
+      }
+
+      const decodedToken = decodeToken(token) as {
+        actor_id: string
+        user_metadata: Record<string, unknown>
+      }
+
+      const userExists = decodedToken?.actor_id !== ""
+      if (!userExists) {
```
sradevski: If it exists it means it's just a login right?

packages/admin/dashboard/src/routes/login/components/cloud-auth-login.tsx
```diff
@@ -0,0 +1,105 @@
...
+        // Create user account for this auth identity
+        await createCloudAuthUser({
+          email: decodedToken.user_metadata.email as string,
+          first_name: decodedToken.user_metadata.given_name as any,
+          last_name: decodedToken.user_metadata.family_name as any,
```
sradevski: Where is the user metadata populated from? If it's Cloud, maybe we go with `first_name` `last_name` instead of given/family name as that's what we use in most places. Very minor though

packages/admin/dashboard/src/routes/login/components/cloud-auth-login.tsx
```diff
@@ -0,0 +1,105 @@
...
+        }
+      }
+
+      return true
+    },
+    onSuccess: () => {
+      navigate("/app")
+    },
+    onError: (error) => {
+      toast.error(error.message || "Authentication failed")
```
sradevski: Should we translate the errors?

packages/admin/dashboard/src/routes/login/components/cloud-auth-login.tsx
```diff
@@ -0,0 +1,105 @@
...
+    },
+  })
+
+  // Check if we're returning from the OAuth callback
+  const hasCallbackParams = useMemo(() => {
```
sradevski: I don't think we need the `useMemo` since the result is a boolean

packages/admin/dashboard/src/routes/login/login.tsx
```diff
@@ -70,6 +73,11 @@ export const Login = () => {
     form.formState.errors.email?.message ||
     form.formState.errors.password?.message
 
+  const loginAfterWidgets = getWidgets("login.after")
+  if (cloudAuth?.enabled) {
+    loginAfterWidgets.push(CloudAuthLogin)
```
sradevski, line 78: Not sure if mutating the array returned from getWidgets is ok or not, might be better to create a new one instead or at the bottom do `[...getWidgets, ...(cloudAuth ? [cloudAuth] : [])]`

packages/medusa/src/api/auth/middlewares.ts
```diff
@@ -11,7 +11,7 @@ export const authRoutesMiddlewares: MiddlewareRoute[] = [
   {
     method: ["POST"],
     matcher: "/auth/session",
-    middlewares: [authenticate("*", "bearer")],
+    middlewares: [authenticate("*", "bearer", { allowUnregistered: true })],
```
sradevski, line 14: How did this work before?

packages/medusa/src/api/cloud/auth/users/route.ts
```diff
@@ -0,0 +1,91 @@
...
+
+  // If user already exists for this auth identity, reject
+  if (req.auth_context?.actor_id) {
+    throw new MedusaError(
+      MedusaError.Types.INVALID_DATA,
+      "The user is already registered and cannot create a new account."
+    )
+  }
+
+  if (!req.auth_context?.auth_identity_id) {
```
sradevski: This shouldn't be necessary as the middleware should do it right?

packages/medusa/src/api/cloud/auth/users/route.ts
```diff
@@ -0,0 +1,91 @@
...
+      entity: "user",
+      fields: ["id"],
+      filters: {
+        email: req.body.email,
+      },
+    })
+    .then((result) => result.data[0])
+
+  if (user) {
+    await setAuthAppMetadataWorkflow(req.scope).run({
```
sradevski, line 67: It's better to do wfe.run (but I know in core we call workflows directly like this as well)

packages/medusa/src/api/cloud/auth/route.ts
```diff
@@ -0,0 +1,11 @@
+import { MedusaRequest, MedusaResponse } from "@medusajs/framework"
+import { ContainerRegistrationKeys } from "@medusajs/framework/utils"
+
+export const GET = async (req: MedusaRequest, res: MedusaResponse) => {
```
sradevski, line 4: NIT: Should we make this generic for all providers?

packages/medusa/src/api/cloud/middlewares.ts
```diff
@@ -0,0 +1,25 @@
+import {
+  authenticate,
+  MiddlewareRoute,
+  validateAndTransformBody,
+} from "@medusajs/framework/http"
+import { CreateCloudAuthUserSchema } from "./auth/users/route"
+
+export const cloudRoutesMiddlewares: MiddlewareRoute[] = [
```
sradevski, line 3: Is it not possible to reuse the existing `[actor_type]/[auth_provider]` routes? It's better if we don't add custom routes like this if possible

packages/admin/dashboard/src/routes/login/components/cloud-auth-login.tsx
peterlgh7: yep

packages/admin/dashboard/src/routes/login/components/cloud-auth-login.tsx
peterlgh7: This is not a route, it's just a component that is used in login.tsx

peterlgh7: i'll put it under component to avoid confusion

peterlgh7: but ideally this becomes a widget once we move it to a plugin

packages/medusa/src/api/cloud/auth/users/route.ts
```diff
@@ -0,0 +1,91 @@
...
+      "Authentication required to create a user account."
+    )
+  }
+
+  const user = await query
+    .graph({
+      entity: "user",
+      fields: ["id"],
+      filters: {
+        email: req.body.email,
```
sradevski: req.validatedBody?

packages/medusa/src/api/cloud/auth/users/route.ts
```diff
@@ -0,0 +1,91 @@
...
+    return
+  }
+
+  const { result: createdUser } = await createUserAccountWorkflow(
+    req.scope
+  ).run({
+    input: {
+      authIdentityId: req.auth_context.auth_identity_id,
+      userData: {
+        email: req.body.email,
```
sradevski: Same, req.validatedBody

packages/medusa/src/api/auth/middlewares.ts
peterlgh7: I wonder the same, I'll check now

packages/medusa/src/api/cloud/auth/users/route.ts
peterlgh7: oh this was Cursor, I hadn't done this before actually. I'll check other routes and see what is prevalent

packages/medusa/src/api/cloud/auth/route.ts
peterlgh7: I planned to bring it up on Monday, but I think we should

packages/medusa/src/api/cloud/middlewares.ts
peterlgh7: This is what the docs recommend for a custom provider. But, I also think we could make this generic and lower the barrier for creating a custom provider. I just wasn't sure if we want that, cause it may be an overgeneralization. I also thought if it wasn't done already there might be a reason, so I'd rather discuss this one on Monday too.

packages/admin/dashboard/src/routes/login/components/cloud-auth-login.tsx
peterlgh7: Google and Github providers are both using given/family name too. I think that's just because the OAuth2 spec uses those and they were copied down to our user metadata.

packages/admin/dashboard/src/routes/login/components/cloud-auth-login.tsx
peterlgh7: You're right, I copied this from the docs, but being boolean + the computation being such a simple one doesn't really deserve a useMemo.

packages/admin/dashboard/src/routes/login/login.tsx
peterlgh7: Yeah this was already fixed, CursorBot noticed

packages/medusa/src/api/cloud/auth/users/route.ts
peterlgh7: good point, that was also on the docs and i didn't review it

packages/medusa/src/api/cloud/auth/users/route.ts
peterlgh7: Actually not Cursor, also copied from the docs

peterlgh7: discussed offline, decided to leave as is as it's the dominant pattern right now

packages/admin/dashboard/src/routes/login/components/cloud-auth-login.tsx
peterlgh7: I checked and none of the "Authentication failed" are currently translated, probably cause most come from the BE

packages/medusa/src/api/auth/middlewares.ts
peterlgh7: @sradevski the reason why this hasn't been necessary so far:
- For Oauth providers, we recommend using `AUTH_TYPE=jwt`, so probably no one's hit this endpoint
- For email pass the user is created when the accept invite endpoint is called, i.e. before calling auth/session

packages/modules/auth/src/providers/medusa-cloud-auth.ts
```diff
@@ -0,0 +1,206 @@
...
+      if (error.type === MedusaError.Types.NOT_FOUND) {
+        const createdAuthIdentity = await authIdentityService.create({
+          entity_id,
+          user_metadata: userMetadata,
+        })
+        authIdentity = createdAuthIdentity
+      } else {
+        return { success: false, error: error.message }
+      }
+    }
```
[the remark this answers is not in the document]
peterlgh7: that's ok for now

packages/core/utils/src/common/define-config.ts
```diff
@@ -504,6 +511,22 @@ function applyCloudOptionsToModules(
...
+        module.options = {
+          cloud: {
+            oauth_authorize_endpoint: config.oauthAuthorizeEndpoint,
+            oauth_token_endpoint: config.oauthTokenEndpoint,
+            environment_handle: config.environmentHandle,
+            sandbox_handle: config.sandboxHandle,
+            api_key: config.apiKey,
+            callback_url:
+              config.oauthCallbackUrl ||
+              `${adminConfig?.backendUrl}/auth/user/cloud/callback`,
```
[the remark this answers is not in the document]
peterlgh7: MEDUSA_BACKEND_URL is always set in Cloud

packages/admin/dashboard/src/routes/login/components/cloud-auth-login.tsx
```diff
@@ -0,0 +1,114 @@
...
+        if (!decodedToken?.actor_id) {
+          await createCloudAuthUser()
+
+          if (__AUTH_TYPE__ === "jwt") {
+            // Refresh token to get the updated token with actor_id
+            const newToken = await sdk.auth.refresh()
+            if (!newToken) {
+              throw new Error("Failed to refresh token after user creation")
+            }
+          }
```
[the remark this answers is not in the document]
peterlgh7: not really

packages/admin/dashboard/src/routes/login/login.tsx
```diff
@@ -21,6 +23,7 @@ export const Login = () => {
   const location = useLocation()
   const navigate = useNavigate()
   const { getWidgets } = useExtension()
+  const { data: cloudAuth } = useCloudAuthEnabled()
```
olivermrbl, line 26: Couldn't this be an env var injected in the admin dashboard?

packages/core/utils/src/common/define-config.ts
olivermrbl: I presume Cloud auth will be enabled by default on new projects; it would be nice to make this option configurable from the project settings page, so users don't have to set an env var. It is effectively the same thing happening under the hood, but just with a slightly better UX.

packages/medusa/src/api/cloud/auth/users/route.ts
```diff
@@ -0,0 +1,82 @@
...
+      authIdentityId: req.auth_context.auth_identity_id,
+      userData: {
+        email: req.auth_context.user_metadata.email as string,
+        first_name: req.auth_context.user_metadata.given_name as string,
+        last_name: req.auth_context.user_metadata.family_name as string,
+      },
+    },
+  })
+
+  res.status(201).json({ user: createdUser })
```
olivermrbl: nit: should be a 200 to follow general conventions in the core ... we use 201 for creates without content in the response

peterlgh7: We should update [these docs](https://docs.medusajs.com/resources/how-to-tutorials/how-to/admin/auth#create-user-api-route) then.

olivermrbl: Yup, think that would make sense. Although this is a custom route, I still think we can recommend following the core conventions. cc @shahednasser

packages/admin/dashboard/src/routes/login/login.tsx
```diff
@@ -70,6 +73,11 @@ export const Login = () => {
     form.formState.errors.email?.message ||
     form.formState.errors.password?.message
 
+  const loginAfterWidgets = [...getWidgets("login.after")] // cloning to avoid mutating the original array below
+  if (cloudAuth?.enabled) {
```
olivermrbl, line 77: thought(non-blocking): we don't really need this to be a widget, but I guess it doesn't hurt ... it could just be conditionally rendered directly on the page based on whether it is enabled or not

peterlgh7: well it's not really a widget here, just added to the list of widgets cause I envisioned all the cloud stuff going to a plugin in the near future, and `CloudAuthLogin` becoming an actual widget.

olivermrbl: yeah, but that's sort of my point ... since it's not a widget and we don't need it to be a widget, we could just inline the rendering to not "abuse" the concept

however, not a strong opinion; think it's fine to keep it as is, if we have plans to move it to a plugin

packages/admin/dashboard/src/routes/login/login.tsx
peterlgh7: @olivermrbl that was my first attempt at it, but we don't pass env variables at build time in Cloud right now. I'll probably work on that this week even, but I still have doubts whether it's enough to determine this at build time. So far we're allowing configuration of all cloud options either via env vars or through `ProjectConfig`. Correct me if I'm wrong, but I thought I wouldn't be able to access `ProjectConfig` at build time.
I do think this HTTP call is super annoying, so we could just decide to only allow this configuration through. env vars, and change this once Cloud injects env vars.

packages/admin/dashboard/src/routes/login/login.tsx
peterlgh7: I was intentionally keeping it widget-like as a reminder that it would be better to not add ad-hoc Cloud stuff directly on Medusa

packages/admin/dashboard/src/routes/login/login.tsx
olivermrbl: > we could just decide to only allow this configuration through. env vars, and change this once Cloud injects env vars.

Yeah, think this would be fine. Let's give this a shot once we have support for it in Cloud

olivermrbl (approved):

PR #14405, opened 2025-12-25
chore: Ensure isolated order deletion

olivermrbl:
## Summary

Ensure that address deletion during order deletion is handled correctly with respect to cascading deletes. Currently, when deleting an order that does not have a shipping or billing address, we incorrectly attempt to delete all order addresses. This happens because `undefined` address IDs are not handled properly during deletion.

More specifically, when deleting an order without addresses, the following method is called with these arguments:
```ts
await deleteOrderAddresses([undefined, undefined])
```
This triggers deletion of all order addresses.

To make matters worse, because we have a cascade delete defined from order addresses to orders, the same call also deletes all associated orders.

The root cause is the following filtering logic:
```ts
const orderAddressIds = orders
      .map((order) => [order.shipping_address_id, order.billing_address_id])
      .flat(1)
```

For orders without addresses, this produces `[undefined, undefined]` as input filter, which our delete methods interpret as:
```ts
await delete({ '$or': [ { id: undefined }, { id: undefined } ] })
```

This effectively translates to “delete all addresses.”

In this PR, we make two changes to prevent this going forward:
1. Filter out undefined address IDs before attempting deletion
2. Remove the cascade delete from address to order, as this is an aggressive constraint

The latter change is open for discussion, but cascading deletes from a child entity to a parent is slightly off IMO. Let me know your thoughts.

## Checklist

---

## Additional Context

---

> [!NOTE]
> Strengthens order deletion to avoid unintended cascades and validates behavior with tests.
> 
> - Update FK constraints in migration to `ON DELETE SET NULL` for `order.shipping_address_id` and `order.billing_address_id` (was `CASCADE`)
> - In `order-module-service.ts` `deleteOrders`, filter falsy address IDs and conditionally batch-delete `order_address`/`order_change` via `promiseAll`
> - Add integration tests `integration-tests/__tests__/delete-order.spec.ts` covering deletion of orders and associated entities, deleting orders without addresses (no cross-order impact), and address deletion setting `NULL` on the order
> - Add changeset marking `@medusajs/order` patch
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 8e4ab59af407ec865f73fbf286ec93e710915c8e. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


shahednasser (approved):
LGTM, original issue is resolved

fPolic (approved):
LGTM

olivermrbl:
@carlos-r-l-rodrigues any thoughts here?

packages/modules/order/src/services/order-module-service.ts
```diff
@@ -931,8 +932,13 @@ export default class OrderModuleService
...
 
-    await this.orderAddressService_.delete(orderAddressIds, sharedContext)
-    await this.orderChangeService_.delete(orderChangeIds, sharedContext)
+    if (orderAddressIds.length) {
+      await this.orderAddressService_.delete(orderAddressIds, sharedContext)
+    }
+
+    if (orderChangeIds.length) {
+      await this.orderChangeService_.delete(orderChangeIds, sharedContext)
+    }
```
carlos-r-l-rodrigues, line 948: suggestion/nit: we could run that in parallel and await.

carlos-r-l-rodrigues (approved):
LGTM
good catch

PR #14443, opened 2026-01-05
feat(create-medusa-app): add support for pnpm and specifying package manager

shahednasser:
## Summary

Adds pnpm support to create-medusa-app with flexible package manager choices and detection.

Closes DX-2415

Currently, `create-medusa-app` tries to use `yarn` and defaults to `npm`. Users don't have the flexibility to choose the package manager to user. Users also can't use `pnpm` with `create-medusa-app`. So, they're forced to create a project with yarn / npm, then re-install dependencies manually with pnpm.

By providing support for pnpm and providing flexibility in choosing package managers, we improve the DX of create-medusa-app and provide a more intuitive experience using it.

1. Automatically detect the package manager based on the executed command.
    - `pnpm dlx create-medusa-app`, `pnpm create medusa-app`, `pnpx create-medusa-app` -> use `pnpm`.
    - `yarn dlx create-medusa-app`, `yarn create medusa-app` -> use `yarn`
    - default: npm
2. Add `--use-npm`, `--use-yarn`, and `--use-pnpm` options to override the detection method and explicitly specify the package manager.

These changes affect both project and plugin installations, with Next.js storefront starter installation as well.

Other changes: Added tests for the `PackageManager` class. Note that the `create-medusa-app` has no tests set up at the moment, so we need to address adding more tests in the future.

These changes have been tested rigorously across package managers (unchecked are still to be tested):

- [x] Project installation with `npx create-medusa-app` -> installs Medusa with NPM
- [x] Project installation with `--use-npm` (regardless of package manager used) -> installs Medusa with NPM
- [x] Project installation with `pnpm dlx create-medusa-app` -> installs Medusa with PNPM
- [x] Project installation with `pnpx create-medusa-app` -> installs Medusa with PNPM
- [x] Project installation with `--use-pnpm` (regardless of package manager used) -> installs Medusa with PNPM
- [x] Project installation with `yarn dlx create-medusa-app` -> Installs Medusa with Yarn
- [x] Project installation with `--use-yarn` (regardless of package manager used) -> installs Medusa with Yarn
- [x] All above testing with Next.js storefront
- [x] All above testing for plugins with `--plugin` option
- [x] Project installation with package manager flag (such as `--use-pnpm`) while it's not installed -> should throw

**Testing Considerations**

1. I wasn't able to specify snapshot version with `yarn create medusa-app` so I wasn't able to test with it. That means if you're using yarn v1 you might not be able to test this without the `--use-yarn` option.
2. If you test the installation with the Next.js starter, open the starter in the browser after the installation, and get an error related to `@medusajs/icons`, this is being addressed by [this PR](https://github.com/medusajs/nextjs-starter-medusa/pull/557) on the Next.js starter repo and should be resolved once it's merged.

---

## Checklist


shahednasser:
@claude review this

olivermrbl (approved):
LGTM, clean PR @shahednasser!

packages/cli/create-medusa-app/src/utils/prepare-project.ts
```diff
@@ -210,59 +222,59 @@ async function prepareProject({
...
+      }
+    )
+
+    if (client) {
+      // check the migrations table is in the database
+      // to ensure that migrations ran
+      let errorOccurred = false
+      try {
+        const migrations = await client.query(
+          `SELECT * FROM "mikro_orm_migrations"`
```
adrien2p: nit-nonblocking:
```sql
SELECT count(tablename) from pg_tables WHERE tablename = 'mikro_orm_migrations'
```

adrien2p (approved):
LGTM

shahednasser:
@claude review this

shahednasser:
@claude review this

packages/cli/create-medusa-app/src/utils/execute.ts
```diff
@@ -72,8 +72,6 @@ const execute = async (
       ...options,
       env: {
         ...process.env,
-        // Disable corepack download prompt - automatically accept downloads
-        COREPACK_ENABLE_DOWNLOAD_PROMPT: "0",
```
adrien2p: what this was producing?

packages/cli/create-medusa-app/src/utils/clone-repo.ts
```diff
@@ -104,14 +104,17 @@ function deleteGitDirectory(projectDirectory: string) {
  * Useful for deleting directories when fs methods fail (e.g., with Yarn v3)
  */
 function deleteWithCommand(projectDirectory: string, dirName: string) {
-  const dirPath = path.join(projectDirectory, dirName)
+  const dirPath = path.normalize(path.join(projectDirectory, dirName))
+  if (!dirPath.startsWith(projectDirectory)) {
```
adrien2p: Q: here dirPath already join projectDirectory, so in theory we will never enter that condition no?

packages/cli/create-medusa-app/src/utils/execute.ts
shahednasser: This was an attempt for a fix related to conflicting yarn versions, I removed it as it was unnecessary

packages/cli/create-medusa-app/src/utils/clone-repo.ts
shahednasser: Correct, just a security recommendation from Claude 😄

adrien2p: I believe it must have hallucinate, his comments mention dirPath without account for its construction which already include that parameter which means that condition will never resolves to true 😅

shahednasser: Ok I'll remove the condition 😄

adrien2p: and furthermore, it could break instead, because we look if path starts with the project dir except that it have been normalized so it could defer because of the normalization right>?

olivermrbl (approved):

PR #14444, opened 2026-01-05
chore: fix eslint config in root

shahednasser:
Fix the ESLint config in the root of the repo to resolve the TSConfig of each project correctly.


olivermrbl (approved):
nice catch, wonder how we didn't see this earlier

PR #14446, opened 2026-01-05
fix: product pivots id format migration

fPolic:
**What**
- ensure product options pivot tables have ulid ids
- update ids in batches with a migration script

---

CLOSES CORE-1307


fPolic:
@claude review this PR

adrien2p (approved):

packages/medusa/src/migration-scripts/migrate-product-option-link-ids.ts
```diff
@@ -0,0 +1,103 @@
...
+    const { rows } = await pgConnection.raw<{ id: string }>(
+      `select id from "${tableName}" where id not like ? order by id limit ?`,
+      [idPrefixPattern, BATCH_SIZE]
+    )
+
+    if (!rows.length) {
+      break
+    }
+
+    batchCount += 1
```
adrien2p, line 53: suggestion: should we invert both?

fPolic: could you elaborate please, you men invert their order?

adrien2p: Nvm, I think I read something else, my brains must have played with me 🤣

packages/medusa/src/migration-scripts/migrate-product-option-link-ids.ts
```diff
@@ -0,0 +1,103 @@
...
+  try {
+    await pgConnection.transaction(async (trx) => {
+      await updateIdsInBatches({
+        tableName: "product_product_option",
+        idPrefix: "prodopt",
+        pgConnection: trx,
+      })
+
+      await updateIdsInBatches({
+        tableName: "product_product_option_value",
```
olivermrbl, line 94: question: Just be sure, do we need to deal with the FK from `product_product_option_value.product_product_option_id` to `product_product_option.id`?

fPolic: there is a cascade update on FK: https://github.com/medusajs/medusa/blob/chore/product-pivots-migrations-id/packages/modules/product/src/migrations/Migration20251110180907.ts#L19 so `product_product_option_value.product_product_option_id` is updated when we set new `product_product_option.id` (if that is your question?)

PR #14457, opened 2026-01-06
docs: auth in cloud + restructure deployment guides

shahednasser:
1. Move existing access deployment section to its own document for easy access and reference
2. Add details on logging in with Cloud


olivermrbl (approved):
LGTM

PR #14462, opened 2026-01-06
fix(create-medusa-app): fix version option not working with Next.js storefront

shahednasser:
1. Fix error when passing `--version` and installing Next.js storefront. The error is because the UI package doesn't follow the same versioning as other Medusa packages, so this will skip updating the version of the UI package.
2. Other: added error logging for installation errors


shahednasser:
@claude review this

olivermrbl (approved):

PR #14476, opened 2026-01-07
feat(events): Implement default priority-based event processing

adrien2p:
RESOLVES CORE-1356

**What**
In order to prevent internal events to overload the queue and prevent more meaningful event to be processed we decided on the following:
- Internal events will have the lowest priority, allowing other events to be processed first
- All other events will have a priority of 100
- order placed events during emition specify a priority of 10 to always be processed first


adrien2p:
@claude review this

carlos-r-l-rodrigues (approved):
LGTM

adrien2p:
@claude review this

adrien2p:
@shahednasser just pinging you in case we want to document that somewhere (cc @olivermrbl )

olivermrbl (commented):
thought(non-blocking): One thought that struck me this morning was that this implementation is somewhat specific to BullMQ and our Redis event bus. That is, the notion of a priority on an event is not necessarily a concept shared across event bus providers, e.g. Kafka. For that reason, there is something about it that feels slightly off. 

I don't have a great alternative right now, but I wanted to share this in case you had some thoughts around it?

For example, let's imagine we build a Kafka provider in the near future, the `options.priority` configuration will be made redundant, which to me feels like bad design from our side, since we shouldn't be exposing configurations that are provider specific.

adrien2p:
> thought(non-blocking): One thought that struck me this morning was that this implementation is somewhat specific to BullMQ and our Redis event bus. That is, the notion of a priority on an event is not necessarily a concept shared across event bus providers, e.g. Kafka. For that reason, there is something about it that feels slightly off.
> 
> I don't have a great alternative right now, but I wanted to share this in case you had some thoughts around it?
> 
> For example, let's imagine we build a Kafka provider in the near future, the `options.priority` configuration will be made redundant, which to me feels like bad design from our side, since we shouldn't be exposing configurations that are provider specific.

Yes I agree with you and it is inherent to our current event module architecture as I mentioned in our last call. Currently the only module using it is the event bus redis and therefore there is no conflict or notion of other provider to conflict with if you see what I mean. Furthermore, those options are event-bus-redis related which is the only one existing at the moment.

On the other project I ve mentioned, we would have a provider approach and we might need to have a specific contract that we design for it with certain options and naming and then the provider would have the knowledge of that contract and can remap the opinionated options to its own usage and API. Really like an adapter pattern and similar to payment etc. Also, once we will move to the new architecture, like for caching it will likely be a new module with its provider I believe so it will have its own API. But again, that requires discussion and alignment and part of a bigger project

edit: Also, kafka is a message system, which is different from bullmq. But priority could be a naming we use but instead of numbers we can just use string like critical, high, medium, low, etc and that can be remapped for bullmq but also for kafka and other systems. It removes granularity in favor of groupping by levels sort of speak.

olivermrbl (approved):
LGTM

olivermrbl:
@claude provide me with a concise summary of these changes for the release notes

PR #14477, opened 2026-01-07
feat(config): Default event worker concurrency to 3 on cloud

adrien2p:
RESOLVES CORE-1357

**What**
For cloud deployement, unless manually configured, the default event bus redis worker concurrency will be set to 3


adrien2p:
@olivermrbl is that one ok to merge as well?

olivermrbl (approved):

PR #14494, opened 2026-01-08
feat(DML): Add a new translatable property modifier applicable on text

adrien2p:
**What**
Add a new `translatable` property modifier to be able to mark an entity property as translatable.
All entities that have translatable properties are stored in a global storage and can be retrieved using `DmlEntity.getTranslatableEntities()` which would return an array of `{ entity, fields }` .

**Usage**
The translation service for example, can now get all translatable entities and their translatable fields without knowing about other modules. Furthermore, the entity name is following conventioned and therefore can be manipulated by the translation module to fit its requirements

PR #14495, opened 2026-01-08
feat(translation,fulfillment,customer,product,region,tax,core-flows,medusa,types): Implement dynamic translation settings management

NicolasGorga:
## Summary

Add an endpoint to batch translation settings to allow managing configurations for translatable entities and fields via API layer.

To move away from static config at the `medusa-config` level and enable creating a UI to interact with said functionality.

Added a new endpoint to batch translation settings for entities. You can specify the translatable fields and also update the general entity translatable status to activate/deactivate it.

Integration tests.

---

## Examples

```ts
// Example usage
```

---

## Checklist

---

## Additional Context


NicolasGorga:
@claude review this

packages/core/types/src/translation/common.ts
```diff
@@ -168,13 +173,30 @@ export interface FilterableTranslationProps
...
+  extends BaseFilterable<FilterableTranslationSettingsProps> {
+  /**
+   * The IDs to filter the translation settings by.
+   */
+  id?: string[] | string | OperatorMap<string | string[]>
+
+  /**
+   * Filter translation settings by entity type.
+   */
+  entity_type?: string | string[] | OperatorMap<string>
```
adrien2p:
```suggestion
  entity_type?: string | string[] | OperatorMap<string | string[]>
```

adrien2p (commented):
as we discussed, we still need a to handle the settings update in the hooks of the service and the validation 👍

.changeset/fine-wings-stick.md
```diff
@@ -0,0 +1,8 @@
+---
+"@medusajs/translation": patch
+"@medusajs/core-flows": patch
+"@medusajs/types": patch
+"@medusajs/medusa": patch
+---
+
+feat(translation,medusa,core-flows,types): New translations settings batch endpoint
```
adrien2p: todo: remove one of the changeset :)

adrien2p:
requires https://github.com/medusajs/medusa/pull/14494/changes to be merged first

packages/modules/product/src/models/product-category.ts
```diff
@@ -4,8 +4,8 @@ import Product from "./product"
 const ProductCategory = model
   .define("ProductCategory", {
     id: model.id({ prefix: "pcat" }).primaryKey(),
-    name: model.text().searchable(),
-    description: model.text().searchable().default(""),
+    name: model.text().searchable().translatable(),
+    description: model.text().searchable().translatable().default(""),
     handle: model.text().searchable(),
```
adrien2p, line 9: Q: should we allow handle to be translatable>?

packages/modules/product/src/models/product-collection.ts
```diff
@@ -4,7 +4,7 @@ import Product from "./product"
 const ProductCollection = model
   .define("ProductCollection", {
     id: model.id({ prefix: "pcol" }).primaryKey(),
-    title: model.text().searchable(),
+    title: model.text().searchable().translatable(),
```
adrien2p, line 7: Q: same Q as above

packages/modules/translation/src/services/translation-module.ts
```diff
@@ -492,4 +493,50 @@ export default class TranslationModuleService
...
+    )
+
+    const settingsToUpsert: (
+      | CreateTranslationSettingsDTO
+      | UpdateTranslationSettingsDTO
+    )[] = []
+
+    for (const setting of currentTranslationSettings) {
+      if (
+        !translatableEntitiesMap.has(setting.entity_type) &&
```
adrien2p: todo: this will never match I believe, we stored them in snake cased, the entity is camel case you need to transform it before comparaison

adrien2p (commented):
Some comments, we also need to remove the modules options typings for the entities configuration right?

packages/modules/translation/src/services/translation-module.ts
adrien2p: todo: snake case no?

packages/modules/product/src/models/product-category.ts
NicolasGorga: Naturally, I would say yes, but it comes with its challenges. I would think many would assume they could send the translated handle in retrieval operations, where only the original language handle (the one that was set originally against the entity) is the one that actually make sense for said operation.

packages/modules/product/src/models/product-collection.ts
NicolasGorga: see reply above

.changeset/fine-wings-stick.md
NicolasGorga: didn't understand this could you expand? :)

NicolasGorga:
@adrien2p I see the module tests are failing ,and while debugging, i see that this is because only the translations module gets loaded, which makes sense since they are it's module tests. Since we need all modules to load for the tests, or at least), the ones containing the `entity_type`'s we use in the tests, do you have any suggestion of what we could do here?

adrien2p:
> @adrien2p I see the module tests are failing ,and while debugging, i see that this is because only the translations module gets loaded, which makes sense since they are it's module tests. Since we need all modules to load for the tests, or at least), the ones containing the `entity_type`'s we use in the tests, do you have any suggestion of what we could do here?

since the settings cant be computed based on other modules during the tests due to the isolation nature of the tests, I would recommend to mock the Dml service method returns result and run the on application start manually in the before each

packages/core/core-flows/src/translation/steps/validate-translation-settings.ts
```diff
@@ -0,0 +1,87 @@
...
+import { createStep, StepResponse } from "@medusajs/framework/workflows-sdk"
+
+export const validateTranslationSettingsStepId = "validate-translation-settings"
+
+export type ValidateTranslationSettingsStepInput = (
+  | CreateTranslationSettingsDTO
+  | UpdateTranslationSettingsDTO
+)[]
+
+export const validateTranslationSettingsStep = createStep(
```
adrien2p: Q: I am wondering if that validation shouldnt be in the module directly? wdyt?

packages/core/core-flows/src/translation/workflows/batch-translation-settings.ts
```diff
@@ -0,0 +1,48 @@
...
+
+    const [created, updated, deleted] = parallelize(
+      createTranslationSettingsStep(input.create),
+      updateTranslationSettingsStep(input.update),
+      deleteTranslationSettingsStep(input.delete)
+    )
+
+    return new WorkflowResponse(
+      transform({ created, updated, deleted }, (data) => data)
+    )
```
adrien2p: suggestion: you can remove the transform here and return { created, updated, deleted } directly

packages/modules/translation/integration-tests/__tests__/translation-module-service.spec.ts
```diff
@@ -1,15 +1,41 @@
...
+      ])
+    },
+  },
   testSuite: ({ service }) => {
     describe("Translation Module Service", () => {
+      afterAll(() => {
+        // Restore the mock after all tests complete
+        if (mockGetTranslatableEntities) {
+          mockGetTranslatableEntities.mockRestore()
+        }
```
adrien2p: Q: isnt `mockGetTranslatableEntities` always present?

packages/modules/translation/integration-tests/__tests__/translation-module-service.spec.ts
```diff
@@ -247,7 +273,7 @@ moduleIntegrationTestRunner<ITranslationModuleService>({
         })
       })
 
-      describe("Translation", () => {
+      describe.only("Translation", () => {
```
adrien2p: todo: rm only everywhere

.changeset/fine-wings-stick.md
adrien2p: no worries, it was because you didnt change the base branch here and therefore there was my changeset as well

packages/core/core-flows/src/translation/steps/validate-translation-settings.ts
olivermrbl: +1, also seems more reasonable to me

olivermrbl (commented):
Overall LGTM, will let @adrien2p have the final say :)

packages/modules/translation/src/services/translation-module.ts
```diff
@@ -377,6 +382,58 @@ export default class TranslationModuleService
...
+    const updatedSettings = await this.settingsService_.update(
+      dataArray,
+      sharedContext
+    )
+
+    const serialized = await this.baseRepository_.serialize<
+      TranslationTypes.TranslationSettingsDTO[]
+    >(updatedSettings)
+
+    return Array.isArray(data) ? serialized : serialized[0]
```
adrien2p:
```suggestion
   return await super.updateTranslationSettings(data, sharedContext)
```

packages/modules/translation/src/services/translation-module.ts
```diff
@@ -377,6 +382,58 @@ export default class TranslationModuleService
...
+    const createdSettings = await this.settingsService_.create(
+      dataArray,
+      sharedContext
+    )
+
+    const serialized = await this.baseRepository_.serialize<
+      TranslationTypes.TranslationSettingsDTO[]
+    >(createdSettings)
+
+    return Array.isArray(data) ? serialized : serialized[0]
```
adrien2p:
```suggestion
  return await super.createTranslationSettings(data, sharedContext)
```

packages/modules/translation/src/services/translation-module.ts
```diff
@@ -492,4 +549,128 @@ export default class TranslationModuleService
...
+    )
+
+    const settingsToUpsert: (
+      | CreateTranslationSettingsDTO
+      | UpdateTranslationSettingsDTO
+    )[] = []
+
+    for (const setting of currentTranslationSettings) {
+      if (
+        !translatableEntitiesSet.has(setting.entity_type) &&
```
adrien2p: suggestion-nit: can you move that higher up like just after the constructor, just so that we have the info straight away please

adrien2p (approved):

adrien2p (approved):
LGTM

PR #14502, opened 2026-01-09
fix: Add schema only flag on Medusa app loader

adrien2p:
**What**
Introduce a new flag on medusa app `schemaOnly` for the purpose of only generate gql schema which purely rely on joiner config and therefore ensure we isolate that responsiblity when needed like in the build/dev command generate types


adrien2p:
@claude review this

olivermrbl (approved):

adrien2p:
the unreachable code block is unreachable in some cases on purpose and will be reach on the other cases on purpose

PR #14521, opened 2026-01-12
chore(medusa): middleware policies

carlos-r-l-rodrigues:
# Introduce Middleware Policies for Role-Based Access Control

## Overview

This PR adds a new `policies` property to both middleware configurations and route exports, enabling fine-grained role-based access control (RBAC) at the middleware and route level.

### 1. New `policies` Property for Middlewares

Middlewares can now define policies to restrict access based on user roles:

```typescript
// Single policy
{
  matcher: "/admin/campaigns",
  handler: someHandler,
  policies: {
    resource: "campaign",
    operation: "read"
  }
}

// Multiple policies
{
  matcher: "/admin/campaigns",
  handler: someHandler,
  policies: [
    { resource: "campaign", operation: "read" },
    { resource: "promotion", operation: "read" }
  ]
}
```

### 2. Permission Checking Middleware

- Added `wrapWithPermissionCheck()` function that automatically checks user permissions before executing handlers
- Extracts user roles from JWT token's `app_metadata.roles`
- Uses the existing `hasPermission()` utility from `@medusajs/utils`
- Returns 401 Unauthorized if user lacks required permissions

#### Benefits

- **Centralized Authorization**: Policies are defined alongside middlewares they protect
- **Flexible Configuration**: Support for per-method policies or global policies
- **Automatic Enforcement**: Framework automatically applies permission checks


packages/core/framework/src/http/types.ts
```diff
@@ -1,12 +1,18 @@
+import type {
```
olivermrbl, line 1: nit: can we make `policies` private until the feature is ready for release?

cc @shahednasser

packages/core/types/src/common/common.ts
```diff
@@ -437,6 +437,10 @@ export type RawRounding = {
  * @ignore
  */
 export type QueryConfig<TEntity> = {
+  /**
+   * The main entity to retrieve. For example, `product`.
+   */
+  entryPoint?: TEntity | string
```
olivermrbl: question: is this used anywhere? couldn't see it in the PR

if so, I would maybe argue we should name this `entity` to align with the query graph API, wdyt?

olivermrbl (approved):
Overall LGTM!

packages/core/framework/src/http/types.ts
shahednasser: @carlos-r-l-rodrigues can you add an `@ignore` TSDoc flag for the `policies` properties? This will make sure it's not generated as part of any reference

packages/core/framework/src/http/middlewares/check-permissions.ts
```diff
@@ -0,0 +1,85 @@
...
+  }
+
+  const authContext = req.auth_context
+  // Get roles from JWT token's app_metadata
+  const roleIds = (authContext.app_metadata?.roles as string[]) || []
+
+  if (!roleIds.length) {
+    throw new MedusaError(
+      MedusaError.Types.UNAUTHORIZED,
+      "User has no roles assigned"
```
adrien2p: Q: should we just say action fobidden or similar to not leak information?

packages/core/framework/src/http/middlewares/check-permissions.ts
```diff
@@ -0,0 +1,85 @@
...
+  })
+
+  if (!hasAccess) {
+    const policyKeys = policyList
+      .map((p) => `${p.resource}:${p.operation}`)
+      .join(", ")
+
+    throw new MedusaError(
+      MedusaError.Types.UNAUTHORIZED,
+      `Insufficient permissions. Required policies: ${policyKeys}`
```
adrien2p, line 50: Q: same as previously, cause I think for 401 we dont change the message like we do for others like 500. maybe thats something to change as well now that I think about it

packages/core/framework/src/http/router.ts
```diff
@@ -194,12 +202,18 @@ export class ApiLoader {
...
+      }
+
       const handler = ApiLoader.traceMiddleware
-        ? (ApiLoader.traceMiddleware(wrapHandler(route.handler), {
+        ? (ApiLoader.traceMiddleware(wrapHandler(handlerToUse), {
             route: route.matcher,
             method: method,
           }) as RequestHandler)
-        : wrapHandler(route.handler)
+        : wrapHandler(handlerToUse)
```
adrien2p, line 216: Q: I am wondering the impact of doing the policy check on each middleware rather than first thing after auth but once. But also the effort to do that? we might discuss it to be sure

packages/core/types/src/common/common.ts
adrien2p: +1 this is true we are using entity for graph. We wanted to align with graph and thought it was entry point but in fact that was the remote query wording :)

packages/core/framework/src/http/router.ts
carlos-r-l-rodrigues: It's cached in memory, but we can always run the load test and compare.

packages/core/types/src/common/common.ts
carlos-r-l-rodrigues: This will be used in the next PRs where we'll have to know what is the main entity of the requested fields.

adrien2p: it is more about the wording :)

PR #14528, opened 2026-01-12
fix(core-flows): Avoid throwing if no prices found for variant when adding to cart custom price item

NicolasGorga:
## Summary

When adding to cart a line item with custom unit price via workflow, avoid throwing an error if no price is found for the linked variant.

Without this, it is not possible to have product variants whose price is going to be resolved exclusively dynamically.

Avoid throwing when adding to cart if the variant doesn't have prices defined and the line item has a custom price.

Integration test.

---

## Examples

```ts
// Example usage
```

---

## Checklist

---

## Additional Context

Closes CORE-1362

### DIFF
--- a/packages/core/core-flows/src/cart/workflows/get-variants-and-items-with-prices.ts
+++ b/packages/core/core-flows/src/cart/workflows/get-variants-and-items-with-prices.ts
@@ -125,6 +125,7 @@ export const getVariantsAndItemsWithPrices = createWorkflow(
               context: {
                 ...baseContext,
                 quantity: item.quantity,
+                is_custom_price: !!item.unit_price,
               },
             }
           })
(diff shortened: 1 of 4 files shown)


packages/core/core-flows/src/cart/workflows/get-variants-and-items-with-prices.ts
```diff
@@ -180,7 +180,10 @@ export const getVariantsAndItemsWithPrices = createWorkflow(
             calculatedPriceSet = calculatedPriceSets[item_.variant_id!]
           }
 
-          if (!calculatedPriceSet && item_.variant_id) {
+          const isCustomPrice =
```
adrien2p, line 184: Q: Does that mean if a customer post an item with a custom price and the variant does not have any price then the customer can set any price?

NicolasGorga: The custom price can only be set via workflow, not through http call, so it is the owner of the application the one that is explicitly allowing items to have custom prices or not. This is the same that currently happens; this just allows the variant not to have any base price set, if it represents a product that is 100% dynamic price-wise. Let me know what you think.

adrien2p (approved):

olivermrbl (approved):
LGTM

PR #14538, opened 2026-01-14
docs: document incompatibility for Next.js storefront + Node v25

shahednasser:
Added troubleshooting guides and notes + clarified prerequisites to mention that the Next.js Starter Storefront is incompatible with Node v25

### DIFF
--- a/www/apps/book/app/learn/installation/page.mdx
+++ b/www/apps/book/app/learn/installation/page.mdx
@@ -49,6 +49,12 @@ npx create-medusa-app@latest my-medusa-store
 
 Where `my-medusa-store` is the name of the project's directory and PostgreSQL database created for the project. When you run the command, you'll be asked whether you want to install the Next.js Starter Storefront.
 
+<Note title="Installations with Next.js Starter Storefront">
+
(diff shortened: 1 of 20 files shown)


stuartpb:
Is there an issue on https://github.com/medusajs/nextjs-starter-medusa/issues or discussion in https://github.com/medusajs/nextjs-starter-medusa/discussions to link to for the underlying problem here? Without a description of what would be blocking this, this sounds like a transient problem with an outdated dependency (one that can probably be worked around by passing `--no-webstorage` to the `node` command / putting `export NODE_OPTIONS="--no-webstorage"` in the environment, [which is how vitest works around it](https://github.com/vitest-dev/vitest/issues/8757)).

For example, [this similar issue to the one described](https://github.com/vbenjs/vue-vben-admin/issues/6832) was fixed by bumping `vite-plugin-vue-devtools` to `^8.0.3`; I'd expect something similar to fix the underlying problem here.

shahednasser:
Hi @stuartpb, you can read more about the original issue on [Discord](https://discord.com/channels/876835651130097704/1460831362973171742). Next.js 15 isn't compatible with Node 25, so it needs to be updated. You're welcome to try a workaround in your own setup, but to avoid unexpected issues we prefer documenting the incompatibility and work on updating the Next.js Starter Storefront separately.

PR #14548, opened 2026-01-15
fix: optimise list product options fetching

fPolic:
**What**
- fix loading time for the list product operation
- refactor `filterOptionValues` helper
- remove loading of uneeded relations in a few places

**Why**
- The list methods were loading both `options` and `product_options` relations when options were requested. That would cause an "explosion" in number of rows retreived due to Cartesian product. For each `product x option x value` another `product_option x values` number of rows would be added resulting in 20k rows instead of 1k.
- reduced time from ~**5300ms** -> ~**950ms**
  - _tested on a create product flow which then lists products with (50 products, 10options, 2 values)_


fPolic:
@claude review this PR

packages/modules/product/src/repositories/product.ts
```diff
@@ -430,4 +430,41 @@ export class ProductRepository extends DALUtils.mikroOrmBaseRepositoryFactory(
...
+
+    const rows = await knex("product_product_option_value as ppov")
+      .select("ppo.product_id", "ppov.product_option_value_id")
+      .innerJoin(
+        "product_product_option as ppo",
+        "ppo.id",
+        "ppov.product_product_option_id"
+      )
+      .whereIn("ppo.product_id", productIds)
+      .whereNull("ppo.deleted_at")
```
carlos-r-l-rodrigues: This `ppo.delete_at is null` could be added to the inner join directly.
I don't believe it will change the performace because it is very likely the query planner will do that, but it worth trying.

carlos-r-l-rodrigues (approved):

packages/modules/product/src/repositories/product.ts
fPolic: added, getting similar results but worth trying 👍

PR #14588, opened 2026-01-20
feat(medusa): filter query fields

carlos-r-l-rodrigues:
Summary
This PR introduces RBAC-based field filtering for API responses and enhances the permission system to support wildcard resources ("*"), enabling super admin roles with full access.

🔧 RBAC Field Filtering
Added field filtering based on user permissions for API responses
Filters out requested fields (e.g., prices, tags) when users lack permissions
Configurable via new feature flag MEDUSA_FF_RBAC_FILTER_FIELDS

🎯 Wildcard Resource Support
Enhanced hasPermission function to support resource: "*" for all resources
Super admin roles can now access all resources and operations with *:* permissions

⚙️ Feature Flag
New feature flag: rbac_filter_fields (disabled by default)
Environment variable: MEDUSA_FF_RBAC_FILTER_FIELDS

---

> [!NOTE]
> Enables policy-aware field selection and super-admin access.
> 
> - Adds RBAC field filtering pipeline: `FieldParser`, `AllowedFieldFilter`/`RestrictedFieldFilter`, and `RBACFieldFilter` (GraphQL schema aware); filters disallowed `fields` at query-build time when `MEDUSA_FF_RBAC_FILTER_FIELDS` is enabled
> - Refactors `prepareListQuery`/`prepareRetrieveQuery` to async, accept `req` with `policies`/auth, and include `entity` in `remoteQueryConfig`; updates `validateAndTransformQuery` to await these
> - Enhances `hasPermission` to support wildcard resources/ops (`*`) and centralizes under `policies`; middleware wrapper `check-permissions` now uses it
> - Middleware loader now attaches `policies` even without explicit middlewares
> - JWT generation optionally embeds role IDs; new feature flag `rbac_filter_fields`
> - Aligns pricing policy/resource names (`prices` -> `price`) and updates types/policies
> - Updates routes/tests to new behavior, including integration tests validating field filtering and super-admin access
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 0672098022a5590a18b26cd50a7e14ea38951479. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


integration-tests/http/__tests__/rbac/admin/rbac-field-filtering.spec.ts
```diff
@@ -0,0 +1,414 @@
...
+      testUser = user
+
+      const rbacModule = container.resolve(Modules.RBAC)
+
+      // Create RBAC policies for product operations
+      productReadPolicy = await rbacModule.createRbacPolicies([
+        {
+          key: "read:product",
+          resource: "product",
+          operation: PolicyOperation.read,
```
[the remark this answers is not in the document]
adrien2p: wrong, global var

packages/core/framework/src/http/middleware-file-loader.ts
```diff
@@ -124,8 +124,15 @@ export class MiddlewareFileLoader {
...
 
-        if (route.middlewares) {
-          route.middlewares.forEach((middleware) => {
+        if (route.middlewares || route.policies) {
+          const middlewares = route.middlewares ?? []
+          if (route.policies && !route.middlewares?.length) {
+            middlewares.push((_, __, next) => {
+              next()
+            })
+          }
```
adrien2p, line 133: Q: any reason for pushing a noop there?

packages/core/utils/src/modules-sdk/policy-to-types.ts
```diff
@@ -59,7 +59,7 @@ ${resourceKeys
    * import { PolicyOperation } from '@medusajs/framework/utils'
    * 
    * const readOp = PolicyOperation.read // "read"
-   * const writeOp = PolicyOperation.write // "write"
+   * const writeOp = PolicyOperation.create // "create"
```
adrien2p, line 62: Q: write op are used for create and update no?

packages/medusa/src/api/store/carts/[id]/complete/route.ts
```diff
@@ -44,15 +44,16 @@ export const POST = async (
 
     // If we end up with errors outside of statusOKErrors, it means that the cart is not in a state to be
     // completed. In these cases, we return a 400.
+    const cartReq = await prepareRetrieveQuery(
+      {},
+      {
+        defaults: defaultStoreCartFields,
+      }
+    )
```
Comment (adrien2p) [inline packages/medusa/src/api/store/carts/[id]/complete/route.ts:53]: Q: wouldnt we have gone through the middleware to prepare the fields already there?

adrien2p (commented):
🔥

packages/core/framework/src/http/middleware-file-loader.ts
carlos-r-l-rodrigues: Because we wrap the policy checks and run the original middleware, in this case, there were none.
Potentially I can check on the wrapper if there is any original function to be executed, but I was affraid of breaking something unexpected.
Will run some real tests and if I can have this removed, I'll do it in a next PR.

packages/core/utils/src/modules-sdk/policy-to-types.ts
carlos-r-l-rodrigues: We can discuss that, if it makes sense to have `create` and `update`, or we just use `write` for everything.
cc: @olivermrbl

packages/medusa/src/api/store/carts/[id]/complete/route.ts
Comment (carlos-r-l-rodrigues) [reply packages/medusa/src/api/store/carts/[id]/complete/route.ts:53]: This is an exception. Here the prepared one is for "Order", and in this condition we return "Cart". I also thought that when changing this part.

packages/core/utils/src/modules-sdk/policy-to-types.ts
adrien2p: yeah my concern here is that write is changed to create, but then what happen to update? could exist a case where someone cannot create something and only update existing objects?

packages/medusa/src/api/store/carts/[id]/complete/route.ts
Comment (adrien2p) [reply packages/medusa/src/api/store/carts/[id]/complete/route.ts:53]: oh I see, in case of an error we return the cart, should we though? I am wondering why we return the cart if we fail to complete, as it will be the same as we are suppose to have already at hand one completing no?

packages/core/utils/src/modules-sdk/policy-to-types.ts
olivermrbl: Hmm, a simple example that comes to mind is that customer service staff might not be allowed to create entire new orders, but only update them with e.g. shipping or customer info. 

And if we think about this outside the context of commerce (e.g. on Cloud), I think these scenarios become even more realistic. 

I would maybe go with create and update as separate actions for now, and see how that fits.

packages/core/framework/src/http/utils/get-query-config.ts
```diff
@@ -165,30 +173,46 @@ export function prepareListQuery<T extends RequestQueryFields, TEntity>(
 
   let notAllowedFields: string[] = []
 
+  // RBAC field filtering if request and policies are available
+  if (req?.policies && entity) {
```
olivermrbl: question: should this check the feature flag?

packages/core/framework/src/http/utils/get-query-config.ts
```diff
@@ -165,30 +173,46 @@ export function prepareListQuery<T extends RequestQueryFields, TEntity>(
...
-      MedusaError.Types.INVALID_DATA,
-      `Requested fields [${Array.from(notAllowedFields).join(
-        ", "
-      )}] are not valid`
-    )
+    // Remove not allowed fields from request
+    notAllowedFields.forEach((field) => {
+      allFields.delete(field)
+      starFields.delete(field)
+    })
```
olivermrbl: comment: would maybe keep the existing logic and add this behind a feature flag, otherwise it will be breaking

integration-tests/http/__tests__/rbac/admin/rbac-field-filtering.spec.ts
```diff
@@ -0,0 +1,532 @@
...
+        const response = await api.get(
+          `/admin/products/${baseProduct.id}`,
+          copyAdminHeaders
+        )
+
+        expect(response.status).toEqual(200)
+        expect(response.data.product).toBeDefined()
+        expect(response.data.product.tags).toBeDefined()
+        expect(response.data.product.variants).toBeDefined()
+        expect(response.data.product.variants[0].prices).toBeDefined()
```
olivermrbl, line 426: question: shouldn't we assert some of the fields that shouldn't be there to ensure it is not a false positive?

olivermrbl (commented):
Overall LGTM, will let @adrien2p review the filtering algo

packages/core/framework/src/http/utils/get-query-config.ts
carlos-r-l-rodrigues: `req.policies` will only exist if the flag is enabled. That will call `wrapWithPoliciesCheck`
https://github.com/medusajs/medusa/blob/325080c4b9df3d05e8386b6610f2446c2b5716ad/packages/core/framework/src/http/middlewares/check-permissions.ts#L63-L73

olivermrbl: Makes sense!

adrien2p:
I am currently on the review @carlos-r-l-rodrigues, I might have some changes to propose but i ll let you know 👍

adrien2p (approved):

packages/medusa/src/api/store/carts/[id]/complete/route.ts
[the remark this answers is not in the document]
Comment (adrien2p) [reply packages/medusa/src/api/store/carts/[id]/complete/route.ts:53]: true we might need to pass the req as the last param? 🤔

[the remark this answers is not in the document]
Comment (carlos-r-l-rodrigues) [reply packages/medusa/src/api/store/carts/[id]/complete/route.ts:53]: yes. nice catch

PR #14593, opened 2026-01-20
feat(dashboard,framework,rbac,js-sdk,types,utils,medusa): rbac admin dashboard utils

fPolic:
**What**

- added a new set of utilities for permissions `packages/admin/dashboard/src/lib/permissions/*`
- Introduced permissions providers, hooks, and guard components
- `ProtectedRoute` now fetches RBAC role policies, builds a policy, and wraps the dashboard in `PermissionsProvider`.
- added access-denied UI
- example of permission integration in the customer domain

**Newly Introduced Permission Utils**
- `checkPermission`, `checkAnyPermission`, `checkAllPermissions`, and `can` enforce permission checks with wildcard support
- `getRoutePermission`, `matchRoutePattern`, and `canAccessRoute` enable route-level inference
- `filterByPermission` filters data by permission requirements
- `RESOURCE_ROUTE_MAP` and `ROUTE_PERMISSIONS` define the resource <-> route mapping

**Guard Components and Hooks (Usage Examples)**
```tsx
// Route guard with explicit permissions
{
  path: "/customers/create",
  element: <RoutePermissionGuard permissions={["customer:create"]} />,
  children: [{ path: "", lazy: () => import("./customer-create") }],
}

// Route guard with automatic inference (uses route map)
{
  path: "/customers/:id/edit",
  element: <RoutePermissionGuard />,
  children: [{ path: "", lazy: () => import("./customer-edit") }],
}
```

```tsx
// Component guard
<PermissionGuard resource="customer" operation="delete" fallback={<Text>No access</Text>}>
  <DeleteButton />
</PermissionGuard>

// Hook-based checks
const { can, hasPermission } = usePermissions()
if (can("customer", "update")) {
  // show edit
}
if (hasPermission("customer:read")) {
  // show list
}

// Resource-specific hook
const { canRead, canCreate } = useCustomerPermissions()
```

```tsx
// Register required permissions for a subtree (for displaying permissions needed)
<PermissionsRequirementsProvider>
  <PermissionsRequirement permissions={["customer:update"]} source="customer-detail">
    <CustomerDetail />
  </PermissionsRequirement>
</PermissionsRequirementsProvider>
```


packages/admin/dashboard/src/components/authentication/route-permission-guard/route-permission-guard.tsx
```diff
@@ -0,0 +1,142 @@
...
+ * {
+ *   path: "/customers/create",
+ *   element: <RoutePermissionGuard />,
+ *   children: [...]
+ * }
+ * ```
+ */
+export const RoutePermissionGuard = ({
+  permissions,
+  requireAll = false, // TODO: should be true by default ?
```
adrien2p: yes I believe we went originally with all, and we discussed potentially let the user configure that but I am not sure it was part of the first scope

packages/admin/dashboard/src/lib/permissions/constants.ts
```diff
@@ -0,0 +1,265 @@
+import type { PermissionOperation, PermissionResource } from "./types"
```
adrien2p: Q: will not that file be hard to maintain, especially in case users assign/update permission, override routes etc? Would it make sense for the backend to maybe manage that as it has full awareness and maybe use that as a way to get all the informations?

adrien2p (commented):
I havnt finished the full review but I had a general question, what would happen if an admin that has the permission to change the permission of another admin, would that admin be out of date between the permission it has locally and what the server knows? would it be alright to update the user permission as soon as the user hit a restricted resource that seems allowed in the browser but is not on the server?

packages/admin/dashboard/src/lib/permissions/constants.ts
fPolic: yes 100%, this was temporary to have the concept working while BE was still in development,t but agree that BE should be the source of truth here

fPolic:
good points @adrien2p, considered some of these issues but decided to discuss those when we get closer to implementing the permission management view.

> would it be alright to update the user permission as soon as the user hit a restricted resource that seems allowed in the browser but is not on the server

you mean if we get a 403 / insufficient permission error, we refetch roles and permissions?

adrien2p:
> good points @adrien2p, considered some of these issues but decided to discuss those when we get closer to implementing the permission management view.
> 
> > would it be alright to update the user permission as soon as the user hit a restricted resource that seems allowed in the browser but is not on the server
> 
> you mean if we get a 403 / insufficient permission error, we refetch roles and permissions?

Ok undeerstand 👍 
Yes thats what I mean, in the sense that if you access from the UI an end point that you dont have permission it could mean that your local permission are not up to date and we can refresh them

packages/admin/dashboard/src/providers/permissions-provider/permissions-provider.tsx
```diff
@@ -0,0 +1,103 @@
...
+    },
+    [hasPermission]
+  )
+
+  const can = useCallback(
+    (resource: PermissionResource, operation: PermissionOperation): boolean => {
+      return !!permissionsMap[buildPermission(resource, operation)]
+    },
+    [permissionsMap]
+  )
```
NicolasGorga, line 102: Already satisfied by `hasPermission`

packages/admin/dashboard/src/components/authentication/route-permission-guard/route-permission-guard.tsx
```diff
@@ -0,0 +1,166 @@
...
+  if (!rawPermissions) {
+    return undefined
+  }
+
+  const permissions = Array.isArray(rawPermissions)
+    ? rawPermissions
+    : [rawPermissions]
+  if (!permissions.length) {
+    return undefined
+  }
```
shahednasser, line 62: should we confirm that it is an array of strings? The check above is too permissive accepting any object

packages/admin/dashboard/src/components/common/permission-guard/permission-guard.tsx
```diff
@@ -0,0 +1,156 @@
...
+
+  if ("permission" in props && props.permission) {
+    hasAccess = hasPermission(props.permission)
+  } else if ("permissions" in props && props.permissions) {
+    hasAccess = props.requireAll
+      ? hasAllPermissions(props.permissions)
+      : hasAnyPermission(props.permissions)
+  } else if ("resource" in props && "operation" in props) {
+    hasAccess = can(props.resource, props.operation)
+  }
```
shahednasser: nit: Can't we just use the `requiredPermissions` variable and either apply hasAll / hasAny based on requireAll?

packages/admin/dashboard/src/components/layout/main-layout/main-layout.tsx
```diff
@@ -43,12 +44,14 @@ const MainSidebar = () => {
...
-            <Divider variant="dashed" />
+        <PermissionGuard resource="store" operation="read">
+          <div className="bg-ui-bg-subtle sticky top-0">
+            <Header />
+            <div className="px-3">
+              <Divider variant="dashed" />
+            </div>
           </div>
-        </div>
+        </PermissionGuard>
```
shahednasser, line 54: thought: isn't excessive hiding the header entirely? I think it's fine to show the store name for example. Instead you can't go to its settings page, so we can guard the link to it

packages/admin/dashboard/src/routes/customers/customer-detail/customer-detail.tsx
```diff
@@ -33,26 +35,29 @@ export const CustomerDetail = () => {
...
+          before: getWidgets("customer.details.before"),
+          after: getWidgets("customer.details.after"),
+          sideAfter: getWidgets("customer.details.side.after"),
+          sideBefore: getWidgets("customer.details.side.before"),
+        }}
+        data={customer}
+        hasOutlet
+        showJSON
+        showMetadata
+        requiredPermissionsSection={<RequiredPermissionsSection />}
```
shahednasser: thought: any reason why we don't make this the default?

packages/core/utils/src/modules-sdk/define-policies.ts
```diff
@@ -38,10 +45,11 @@ const PolicyOperation: Record<string, string> & {
   readonly delete: "delete"
   readonly "*": "*"
   readonly ALL: "*"
-} = global.PolicyOperation ?? { ALL: "*" }
+} = global.PolicyOperation ?? { ALL: WILDCARD }
 
 for (const operation of defaultOperations) {
-  const operationKey = operation === "*" ? "*" : toSnakeCase(operation)
+  const operationKey =
+    operation === WILDCARD ? WILDCARD : toSnakeCase(operation)
```
shahednasser: nit: we repeat this conditioning a lot in this file, maybe worth making it a utility funciton and reusing it?

shahednasser (approved):
@NicolasGorga LGTM with nits

packages/admin/dashboard/src/components/layout/main-layout/main-layout.tsx
NicolasGorga: If you don't have `store:read` policy, then the request to fetch the store to get the name will fail

PR #14599, opened 2026-01-21
fix(translation): recompute entity translations translated_field_count upon setting update

NicolasGorga:
## Summary

Recompute `translated_field_count` for translation entries whenever its corresponding entity `translation_setting` has updates to its status and/or translatable fields.

Otherwise, you end up with stale metrics for the translatable fields counts for a given entity. Example:

- Deactivating the entity: should show 0 translatable count for existing translation entries
- Activating the entity: should show the count of the translatable fields present in the translation object
- Adding translatable fields to setting: should increase the count in cases where there are existing translation entries, where their translation object contains the field (this represents a scenario where the field was initially present in the setting, translations were created, and then it was later removed from the setting)
- Removing translatable fields from setting: Inverse of the previous point.

Handle the recomputation of the `translated_field_count` whenever updating the corresponding entity translation setting.

Integration tests.

---

## Examples

```ts
// Example usage
```

---

## Checklist

---

## Additional Context

fixes CORE-1373


packages/modules/translation/src/services/translation-module.ts
```diff
@@ -487,10 +491,63 @@ export default class TranslationModuleService
...
+              translatableFields
+            ),
+          }
+        })
+
+        await this.translationService_.update(toUpdate)
+      })
+    )
+
+    return updatedSettings
```
adrien2p: todo: O(3N) -> O(3) operations, many operation within a loop. we should avoid making queries within loop as it increase the number of time we target the db to as many time as we loop and here we to 2/3 queries per iteration when we can combine those queries and make it constant

packages/modules/translation/src/services/translation-module.ts
```diff
@@ -683,4 +727,27 @@ export default class TranslationModuleService
...
+      | (UpdateTranslationSettingsDTO & { entity_type: string })
+    )[]
+  > {
+    return await Promise.all(
+      settings.map(async (setting) => {
+        const entityType = await this.resolveEntityType_(setting, sharedContext)
+        return { ...setting, entity_type: entityType }
+      })
+    )
+  }
```
adrien2p, line 753: todo: O(N) -> O(1) many queries in loop + potentially unnecessary queries

adrien2p (approved):

olivermrbl (approved):

PR #14612, opened 2026-01-22
fix(medusa): use http type for batch translation settings request

shahednasser:
use http type for batch translation settings request to ensure it's generated in the API reference correctly.


adrien2p (approved):

NicolasGorga (approved):
LGTM, didn't know about this Shahed, will make sure to import this way in the future.

olivermrbl (approved):

PR #14620, opened 2026-01-22
fix(dashboard): fix view configurations order filters

NicolasGorga:
## Summary

Fix order filters when view configurations feature flag is true.

You are unable to filter orders when the view configurations is configured and applying a filter breaks the UI.

Implemented the filters with the expected shape the new `DataTable` component needs, basically now identifying filters with `id` instead of `key` and different shapes for certain filter types, like `multiselect`.

Also cleaned up hook duplication of the legacy implementation of the hook implemented here, which we still need while we have usages of the deprecated `DataTable` component

Validated locally the filters are applied correctly.

---

## Examples

```ts
// Example usage
```

---

## Checklist

---

## Additional Context

fixes CORE-1377


adrien2p (approved):

packages/admin/dashboard/src/hooks/table/filters/use-order-table-filters.tsx
```diff
@@ -17,138 +21,140 @@ export const useOrderTableFilters = (): Filter[] => {
...
+        type: "select",
+        options: regions.map((r) => ({
+          label: r.name,
+          value: r.id,
+        })),
+        multiple: true,
+        searchable: true,
+      }
+
+      filters = [...filters, regionFilter]
```
fPolic: nit: since this is first assignment, filters won't have any items before this

fPolic (approved):
lgtm

packages/admin/dashboard/src/hooks/table/filters/use-order-table-filters.tsx
adrien2p: also, cant we just push in filters?

packages/admin/dashboard/src/hooks/table/filters/use-order-table-filters.tsx
```diff
@@ -17,138 +21,140 @@ export const useOrderTableFilters = (): Filter[] => {
...
+        type: "select",
+        multiple: true,
+        searchable: true,
+        options: sales_channels.map((s) => ({
+          label: s.name,
+          value: s.id,
+        })),
+      }
+
+      filters = [...filters, salesChannelFilter]
```
adrien2p: Q: same here?

adrien2p: and other places

adrien2p (approved):

PR #14642, opened 2026-01-26
chore(medusa): super admin user and rbac invite link

carlos-r-l-rodrigues:
## Summary

This PR enhances the RBAC module with super admin role management and invite-to-role linking capabilities.

### Core Features

- **Super Admin Role Auto-Creation**: RBAC module loader now automatically creates the "Super Admin" role with wildcard (`*:*`) policy on startup
- **Invite-Role Linking**: Invites can now include role assignments that are automatically applied when users accept invitations
- **User-Role Linking**: Added workflow steps to link users to RBAC roles during user creation

### New Components

- `link-users-to-roles` step for user creation workflows
- `link-invites-to-roles` and `get-invite-roles` steps for invite workflows
- `invite-rbac-role` link module definition
- Migration script (`create-super-admin-role.ts`) to assign super admin role to existing users

### Policy Type Generation

- Split policy type generation into two files:
  - `policy-types.d.ts` - Global ambient type declarations
  - `policy-bindings.d.ts` - Module augmentation for `@medusajs/framework/utils`

### API Updates

- Invite endpoints now accept `roles` array in payloads
- User creation via CLI (`medusa user`) links new users to super admin role when RBAC is enabled
- Simplified middleware policies for products, draft-orders, and users

### Test Coverage

- Added integration tests for RBAC role auto-creation
- Updated invite tests to verify role assignment flow
- Fixed RBAC field filtering and roles tests to account for auto-created super admin role


integration-tests/http/__tests__/invite/admin/invite.spec.ts
```diff
@@ -208,5 +210,248 @@ medusaIntegrationTestRunner({
...
+        const userRoles = await linkService.list({
+          user_id: acceptedUser.id,
+        })
+
+        expect(userRoles).toHaveLength(0)
+      })
+
+      it("should handle invite with non-existent role gracefully", async () => {
+        // Try to create invite with non-existent role
+        await api
```
adrien2p: suggestion: in general I prefer that we avoid expectations in catch because if there is no fail then there is no expectation and the test will also pass. Instead you can add `expect.hasAssertions();` at the begining of your test to ensure at least one expectation as been triggered or, alternatively you can return the error from the catch and do the expect outside

packages/core/core-flows/src/invite/steps/get-invite-roles.ts
```diff
@@ -0,0 +1,41 @@
...
+export const getInviteRolesStepId = "get-invite-roles-step"
+/**
+ * This step retrieves the roles associated with an invite.
+ *
+ * @example
+ * const data = getInviteRolesStep({
+ *   invite_id: "invite_123"
+ * })
+ */
+export const getInviteRolesStep = createStep(
```
adrien2p, line 17: nit-opti: I believe there is multiple places where we need to get info from the link, would it make sense to make a common step like we have for graph? wdyt? maybe it is premature but it feels right :)

packages/core/core-flows/src/invite/steps/link-invites-to-roles.ts
```diff
@@ -0,0 +1,55 @@
...
+/**
+ * This step links invites to RBAC roles using the remote link service.
+ *
+ * @example
+ * const data = linkInvitesToRolesStep({
+ *   invite_id: "invite_123",
+ *   role_ids: ["role_245asd156"]
+ * })
+ */
+export const linkInvitesToRolesStep = createStep(
```
adrien2p: Q: would it make sense to use `createRemoteLinkStep` directly instead?

packages/core/core-flows/src/user/steps/link-users-to-roles.ts
```diff
@@ -0,0 +1,55 @@
...
+/**
+ * This step links users to RBAC roles using the remote link service.
+ *
+ * @example
+ * const data = linkUsersToRolesStep({
+ *   user_id: "user_123",
+ *   role_ids: ["role_245asd156"]
+ * })
+ */
+export const linkUsersToRolesStep = createStep(
```
adrien2p: suggestion: same as previous

packages/core/utils/src/common/define-config.ts
```diff
@@ -195,8 +196,14 @@ function resolveModules(
     { resolve: MODULE_PACKAGE_NAMES[Modules.ORDER] },
     { resolve: MODULE_PACKAGE_NAMES[Modules.SETTINGS] },
 
-    // TODO: re-enable this once we have the final release
-    // { resolve: MODULE_PACKAGE_NAMES[Modules.TRANSLATION] },
+    {
```
adrien2p, line 199: Thanks :)

packages/medusa/src/api/admin/users/middlewares.ts
```diff
@@ -13,45 +12,43 @@ import {
...
+        resource: Entities.user,
+        operation: PolicyOperation.read,
+      },
+    ],
   },
   {
     method: ["GET"],
     matcher: "/admin/users/:id",
     middlewares: [
-      authenticate("user", ["bearer", "session"]),
```
adrien2p, line 43: Q: How are the policies working if the auth middleware is not being set in general?

packages/medusa/src/migration-scripts/create-super-admin-role.ts
```diff
@@ -0,0 +1,185 @@
+#!/usr/bin/env node
```
adrien2p, line 1: todo: will this script works with our runner? I am not sure this flag is necessary as it wont be run through bash

packages/medusa/src/migration-scripts/create-super-admin-role.ts
```diff
@@ -0,0 +1,185 @@
...
+
+async function assignSuperAdminRoleToUsers(container: any): Promise<void> {
+  try {
+    const logger = container.resolve(ContainerRegistrationKeys.LOGGER)
+    logger.info("🚀 Starting super admin role assignment script...")
+
+    const userModuleService = container.resolve(Modules.USER)
+    const rbacModuleService = container.resolve(Modules.RBAC)
+
+    const users: User[] = await userModuleService.listUsers({})
```
adrien2p, line 38: suggestion: should we batch?

packages/modules/rbac/integration-tests/__tests__/rbac.spec.ts
```diff
@@ -0,0 +1,304 @@
...
+
+moduleIntegrationTestRunner<IRbacModuleService>({
+  moduleName: Modules.RBAC,
+  debug: true,
+  moduleOptions: {},
+  injectedDependencies: {
+    [Modules.EVENT_BUS]: new MockEventBusService(),
+  },
+  testSuite: ({ service }) => {
+    it.skip(`should export the appropriate linkable configuration`, () => {
```
adrien2p, line 19: Q: why is it skpped?

packages/medusa/src/migration-scripts/create-super-admin-role.ts
carlos-r-l-rodrigues: I tested it, and it works fine

packages/medusa/src/api/admin/users/middlewares.ts
carlos-r-l-rodrigues: All `/admin` endpoints are protected by this already. But before the router loader be reworked, `users` and `invites` were disabling all the middlewares and redefining them here. This is no longer required.

adrien2p: but now wouldnt be impacted by the global admin auth requirement since we removed the exported authenticate toogle from the route files?

packages/medusa/src/migration-scripts/create-super-admin-role.ts
carlos-r-l-rodrigues: do you believe that there is anyone with hundreds of thousands of users?

adrien2p: yeah I dont think so, was more a force of habits in case some people just have a lot of test users in their staging or test env

packages/medusa/src/api/admin/users/middlewares.ts
carlos-r-l-rodrigues: but this is what we want, no?

adrien2p: oh you mean it was miss configured? because some end point should be publically available to accept for example an invite no?

packages/modules/rbac/integration-tests/__tests__/rbac.spec.ts
carlos-r-l-rodrigues: There is an issue while loading the Module/linkable fields that has to be investigated.

adrien2p (approved):

packages/core/core-flows/src/invite/steps/validate-roles-exist.ts
```diff
@@ -0,0 +1,37 @@
...
+
+    const rbacService = container.resolve(Modules.RBAC)
+    const existingRoles = await rbacService.listRbacRoles({
+      id: roleIds,
+    })
+
+    const existingRoleIds = new Set(existingRoles.map((r) => r.id))
+    const missingRoles = roleIds.filter((id) => !existingRoleIds.has(id))
+
+    if (missingRoles.length > 0) {
```
fPolic:
```suggestion
    if (missingRoles.length) {
```

packages/medusa/src/migration-scripts/create-super-admin-role.ts
```diff
@@ -0,0 +1,185 @@
...
+
+    if (users.length === 0) {
```
fPolic: nit: `!users.length`

fPolic (approved):
LGTM!

PR #14670, opened 2026-01-29
fix(core-flows): only consider captures amounts for credit line amount computation

NicolasGorga:
## Summary

Fix the credit line amount computation logic upon order cancellation.

Currently wehen cancelling an order, the credit line amount is incorrectly computed based on the amounts of the payments linked to the order, regardless of their statuses.

Only take into consideration the amounts of the captures (if any) to resolve the total credit line amount. This prevents situations like we currently have, where, for an order with multiple cancelled/pending payments you end up with really big negative pending differences.

Integration tests.

---

## Examples

```ts
// Example usage
```

---

## Checklist

---

## Additional Context

fixes CORE-1382

### DIFF
--- a/packages/core/core-flows/src/order/workflows/cancel-order.ts
+++ b/packages/core/core-flows/src/order/workflows/cancel-order.ts
@@ -138,6 +138,7 @@ export const cancelOrderWorkflow = createWorkflow(
         "payment_collections.payments.refunds.amount",
         "payment_collections.payments.captures.id",
         "payment_collections.payments.captures.amount",
+        "payment_collections.payments.captures.raw_amount",
       ],
       filters: { id: input.order_id },
       options: { throwIfKeyNotFound: true },
@@ -174,7 +175,7 @@ export const cancelOrderWorkflow = createWorkflow(
       return payments
         .flatMap((payment) => payment.captures)
         .reduce(
-          (acc, capture) => MathBN.sum(acc, capture.amount),
+          (acc, capture) => MathBN.sum(acc, capture.raw_amount),
           MathBN.convert(0)
         )
     })
--- a/packages/core/core-flows/src/order/workflows/cancel-order.ts
+++ b/packages/core/core-flows/src/order/workflows/cancel-order.ts
@@ -171,10 +171,12 @@ export const cancelOrderWorkflow = createWorkflow(
         ({ payments }) => payments
       )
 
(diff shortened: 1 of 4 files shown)


adrien2p (approved):
LGTM 🔥

adrien2p:
@NicolasGorga looks like the tests are failing, if you could fix them and we can probably then get that in

adrien2p:
@carlos-r-l-rodrigues could you throw on eye on this one as well please, for me it looks good but just want a double confirmation :)

packages/core/core-flows/src/order/workflows/cancel-order.ts
carlos-r-l-rodrigues, line 140:
```suggestion
        "payment_collections.payments.captures.amount",
        "payment_collections.payments.captures.raw_amount",
```

packages/core/core-flows/src/order/workflows/cancel-order.ts
```diff
@@ -171,10 +171,12 @@ export const cancelOrderWorkflow = createWorkflow(
...
       )
 
-      return payments.reduce(
-        (acc, payment) => MathBN.sum(acc, payment.amount),
-        MathBN.convert(0)
-      )
+      return payments
+        .flatMap((payment) => payment.captures)
+        .reduce(
+          (acc, capture) => MathBN.sum(acc, capture.amount),
```
carlos-r-l-rodrigues:
```suggestion
          (acc, capture) => MathBN.sum(acc, capture.raw_amount),
```

carlos-r-l-rodrigues (approved):
LGTM
just a small suggestion

packages/core/core-flows/src/order/workflows/cancel-order.ts
NicolasGorga: hey guys @carlos-r-l-rodrigues @adrien2p would like to understand why it is better here to do `raw_amount` instead of `amount`?

adrien2p: the raw amount is the bignumber meta, so it would use all precision, values etc and since we rely on the bignumber math it makes more sense, is more reliable and future proof

NicolasGorga: good to know thanks, will make this changes now guys

PR #14674, opened 2026-01-30
feat(dashboard,admin-shared,core-flows,js-sdk,types,medusa,link-modules): RBAC admin dashboard CRUD

fPolic:
**WIP**
- [x] RBAC CRUD UI
- [x] permission management for roles
- [x] assign role when creating user invitation
- [x] role <> user management


packages/admin/dashboard/src/components/data-table/data-table.tsx
```diff
@@ -520,14 +523,28 @@ const DataTableAction = ({
...
+      <Tooltip content={tooltip}>{linkButton}</Tooltip>
+    ) : (
+      linkButton
+    )
   }
 
-  return (
+  return tooltip ? (
+    <Tooltip content={tooltip}>
+      <Button {...buttonProps} onClick={props.onClick}>
```
shahednasser, line 543: q: should we pass `disabled` here since we allow passing it? Otherwise a button would be active. (Same goes without the tooltip)

packages/admin/dashboard/src/dashboard-app/routes/get-route.map.tsx
```diff
@@ -1082,7 +1080,16 @@ export function getRouteMap({
                   children: [
                     {
                       path: "invite",
-                      lazy: () => import("../../routes/users/user-invite"),
+                      element: <RoutePermissionGuard />,
+                      handle: {
+                        permissions: ["invite:create", "invite:read"],
```
shahednasser, line 1085: q: this reads to me that we require both to access the invites, but shouldn't you be able to access if you have either?

packages/admin/dashboard/src/routes/policies/policy-detail/policy-detail.tsx
```diff
@@ -0,0 +1,69 @@
...
+    }
+  )
+
+  if (isLoading || !policy) {
+    return <SingleColumnPageSkeleton sections={2} showJSON showMetadata />
+  }
+
+  if (isError) {
+    throw error
+  }
```
shahednasser: Wouldn't `policy` be unset if an error occurred? `isError` check should be before

packages/admin/dashboard/src/routes/roles/role-add-users/components/add-users-form/add-users-form.tsx
```diff
@@ -0,0 +1,313 @@
...
+  roleId: string
+}
+
+type UserWithRbacRoles = HttpTypes.AdminUser & {
+  rbac_roles?: HttpTypes.AdminRbacRole[] | null
+}
+
+export const AddUsersSchema = zod.object({
+  user_ids: zod.array(zod.string()).min(1),
+})
```
shahednasser: nit: do we need to export this?

packages/admin/dashboard/src/routes/roles/role-add-users/components/add-users-form/add-users-form.tsx
```diff
@@ -0,0 +1,313 @@
...
+const PAGE_SIZE = 10
+
+export const AddUsersForm = ({ roleId }: AddUsersFormProps) => {
+  const { t } = useTranslation()
+  const { handleSuccess } = useRouteModal()
+  const { hasAllPermissions } = usePermissions()
+  const canManageRole = hasAllPermissions([
+    "user:update",
+    "rbac_role:update",
+  ])
```
shahednasser: thought: this doesn't make a lot of sense for me, maybe I'm misunderstanding it. Isn't the `rbac_role:update` sufficient, or does it refer to something else as well?

shahednasser: Also shouldn't the form be inaccessible if you don't have these permissions?

packages/admin/dashboard/src/routes/roles/role-create/components/create-role-form/create-role-form.tsx
```diff
@@ -0,0 +1,414 @@
...
+
+    if (tab === Tab.DETAILS) {
+      currentState[Tab.DETAILS] = "in-progress"
+      currentState[Tab.PERMISSIONS] = "not-started"
+    }
+
+    if (tab === Tab.PERMISSIONS) {
+      currentState[Tab.DETAILS] = "completed"
+      currentState[Tab.PERMISSIONS] = "in-progress"
+    }
```
shahednasser: can be simplified to single conditional assignment for each

packages/admin/dashboard/src/routes/roles/role-create/components/create-role-form/create-role-form.tsx
```diff
@@ -0,0 +1,414 @@
...
+      } catch (error) {
+        toast.error((error as Error).message)
+      }
+    }
+
+    toast.success(
+      t("roles.create.successToast", {
+        name: createdRole.name,
+      })
+    )
```
shahednasser: if an error occurs while adding the policies, an error message is shown but then a success message is shown, which is confusing.

Why don't we allow passing the policies to the create role route? That way it's all handled in a single workflow that handles the compensation and status needed here. From a user perspective, it also doesn't make sense for one part of the operation to succeed and the other to fail.

packages/admin/dashboard/src/routes/roles/role-create/components/create-role-form/create-role-form.tsx
```diff
@@ -0,0 +1,414 @@
...
+                getRowId={(row) => row.id}
+                rowCount={count}
+                pageSize={PAGE_SIZE}
+                isLoading={isPoliciesLoading}
+                prefix={PREFIX}
+                rowSelection={{
+                  state: rowSelection,
+                  onRowSelectionChange: onRowSelectionChange,
+                  enableRowSelection: (row) =>
+                    canAssignPolicy(row.original, hasPermission),
```
shahednasser: thought: if you don't have the permission to assign a policy, maybe it's better to not show it in the first place.

packages/admin/dashboard/src/routes/roles/role-create/role-create.tsx
```diff
@@ -0,0 +1,26 @@
...
+
+  useEffect(() => {
+    if (!isRbacEnabled) {
+      navigate(-1)
+    }
+  }, [isRbacEnabled, navigate])
+
+  if (!isRbacEnabled) {
+    return null
+  }
```
shahednasser: nit, not blocking: we repeat this a lot, maybe worth having a single hook for it like `useRequireRbacFeature`

packages/admin/dashboard/src/routes/roles/role-detail/components/role-general-section/role-general-section.tsx
```diff
@@ -0,0 +1,135 @@
...
+                  disabledTooltip: (
+                    <Hint variant="error">
+                      {t("permissions.accessDenied.action")}
+                    </Hint>
+                  ),
+                },
+              ],
+            },
+          ]}
+        />
```
shahednasser: thought: I prefer hiding actions if they're not accessible, but I'm not sure if we already established this pattern across

packages/admin/dashboard/src/routes/roles/role-detail/components/role-users-section/role-users-section.tsx
```diff
@@ -0,0 +1,320 @@
...
+const PAGE_SIZE = 10
+
+export const RoleUsersSection = ({ role }: RoleUsersSectionProps) => {
+  const [rowSelection, setRowSelection] = useState<RowSelectionState>({})
+  const { t } = useTranslation()
+  const prompt = usePrompt()
+  const { offset, order } = useQueryParams(["offset", "order"])
+  const { hasAllPermissions } = usePermissions()
+
+  const canManageRole = hasAllPermissions(["user:update", "rbac_role:update"])
```
shahednasser, line 34: Same question I had before about requiring both roles

packages/admin/dashboard/src/routes/roles/role-detail/role-detail.tsx
```diff
@@ -0,0 +1,68 @@
...
+    return null
+  }
+
+  if (isLoading || !role) {
+    return <SingleColumnPageSkeleton sections={2} showJSON showMetadata />
+  }
+
+  if (isError) {
+    throw error
+  }
```
shahednasser: Same comment about moving the `isError` condition before the above condition

packages/admin/dashboard/src/routes/roles/role-permissions/role-permissions.tsx
```diff
@@ -0,0 +1,37 @@
...
+            <Heading>
+              {t("roles.permissions.header", { name: role.name })}
+            </Heading>
+          </RouteDrawer.Title>
+          <RouteDrawer.Description className="sr-only">
+            {t("roles.permissions.hint")}
+          </RouteDrawer.Description>
+        </RouteDrawer.Header>
+      )}
+      {!isPending && role && <EditRolePermissionsForm role={role} />}
```
shahednasser: we're running the same condition twice

packages/core/core-flows/src/user/steps/get-assignable-roles.ts
```diff
@@ -0,0 +1,118 @@
...
+      ],
+      filters: filters ?? {},
+      pagination: pagination ?? {},
+    })
+
+    const assignable: AssignableRole[] = []
+
+    for (const role of candidates ?? []) {
+      const actions = (role.policies ?? [])
+        .filter((p: any) => p?.resource && p?.operation)
```
shahednasser: thought: we can apply this condition at the `query.graph` level

shahednasser (commented):
General observation: we're not consistent in the dashboard how we treat an inaccessible action. In some cases we disable it with a tooltip, and in others we hide it. I think we should be consistent in this all across (not just in this PR), and I prefer hiding inaccessible actions. However, not sure whether this was agreed on differently, so I'm fine with whatever the approach chosen is, but we need to keep it consistent.

Also, there are multiple of lint errors in this PR in the dashboard package that should be straightforward to fix.

packages/admin/dashboard/src/dashboard-app/routes/get-route.map.tsx
NicolasGorga: It's because of the pending invites component seen in the focus modal, without the read permission, you wouldn't see this nor the invite you have just created. But maybe we could leave the `invite:create` at the route level and the `invite:read` just protecting the pending invites table, wyt?

CleanShot 2026-05-25 at 15 42 55

packages/admin/dashboard/src/components/data-table/data-table.tsx
NicolasGorga: It's part of `buttonProps`

packages/admin/dashboard/src/routes/roles/role-add-users/components/add-users-form/add-users-form.tsx
NicolasGorga: No we don't thanks

packages/admin/dashboard/src/routes/roles/role-add-users/components/add-users-form/add-users-form.tsx
NicolasGorga: The form is protected at the route level via `RoutePermissionGuard` [here](https://github.com/medusajs/medusa/blob/df5a52f29f1841ef7734226ab3ec591166496e32/packages/admin/dashboard/src/dashboard-app/routes/get-route.map.tsx#L1181-L1194). 

Regarding `user:update` the underlying route had that permission check, so i replicated it in here as well to align it and prevent 403. It makes sense to me, since you are also updating the user itself when you update the role it has. lmkwyt

packages/admin/dashboard/src/routes/roles/role-detail/components/role-general-section/role-general-section.tsx
NicolasGorga: It had a mix of this before i took this on. What I like about the disabling + tooltip is you get more visibility into what's possible via the UI, so it would be easier for you to know where you can perform an action you don't have access to now but that you are later granted, wyt? I do agree we need consistency, so it is a matter of deciding if we should hide actions/components/labels or disable them + informational tooltips.

packages/admin/dashboard/src/routes/roles/role-detail/components/role-users-section/role-users-section.tsx
NicolasGorga: Same reply as in the other comment, keeping unresolved until we reach a conclusion there.

packages/admin/dashboard/src/routes/roles/role-permissions/role-permissions.tsx
NicolasGorga: Thanks missed fixing that

packages/admin/dashboard/src/routes/roles/role-create/components/create-role-form/create-role-form.tsx
NicolasGorga: I think this also ties to reaching a conclusion if we want to hide or show disabled actions in general. Here in particular, I like the fact that you are able to see what policies are out there, so you can directly request the exact policy to your admin.

packages/admin/dashboard/src/dashboard-app/routes/get-route.map.tsx
shahednasser: But then if you have read without create access it's redundant. I think we should either allow a way to apply "any" permissions for the routes, or remove the permission from here and instead apply it in the route at `packages/admin/dashboard/src/routes/users/user-invite/user-invite.tsx` where if the user doesn't have either permissions we redirect them / return null. Then in the component we show the parts relevant to the user's permissions.

packages/admin/dashboard/src/routes/roles/role-add-users/components/add-users-form/add-users-form.tsx
shahednasser: If you only have `rbac_role:update`, what actions can you perform other than updating user roles?

NicolasGorga: I don't know if this would be a real use case, but aside from assigning users to roles, you could change the policies linked to the role.

packages/admin/dashboard/src/routes/roles/role-detail/components/role-general-section/role-general-section.tsx
NicolasGorga: Agreed on hiding

packages/admin/dashboard/src/routes/roles/role-create/components/create-role-form/create-role-form.tsx
NicolasGorga: Agreed on hiding

packages/admin/dashboard/src/dashboard-app/routes/get-route.map.tsx
NicolasGorga: Sorry maybe I missed something, but we are already applying both permissions at the route level with `RoutePermissionGuard` @shahednasser  lmk

packages/admin/dashboard/src/routes/policies/policy-detail/components/policy-roles-section/policy-roles-section.tsx
```diff
@@ -0,0 +1,103 @@
...
+    policy.id,
+    {
+      offset: offset ? parseInt(offset) : 0,
+      limit: PAGE_SIZE,
+    },
+    {
+      placeholderData: keepPreviousData,
+      enabled: canReadRoles,
+    }
+  )
```
shahednasser, line 40: q: is there a reason we don't return early if `canReadRoles` is disabled?

packages/admin/dashboard/src/routes/policies/policy-edit/components/edit-policy-form/edit-policy-form.tsx
```diff
@@ -0,0 +1,155 @@
...
+  operation: z.string().min(1),
+  name: z.string().optional(),
+  description: z.string().optional(),
+})
+
+type EditPolicyFormProps = {
+  policy: HttpTypes.AdminRbacPolicy
+}
+
+export const EditPolicyForm = ({ policy }: EditPolicyFormProps) => {
```
shahednasser, line 25: q: shouldn't we check permissions in this component? We do it in read components.

packages/admin/dashboard/src/routes/policies/policy-edit/policy-edit.tsx
```diff
@@ -0,0 +1,29 @@
+import { Heading } from "@medusajs/ui"
+import { useTranslation } from "react-i18next"
+import { useParams } from "react-router-dom"
+
+import { RouteDrawer } from "../../../components/modals"
+import { useRbacPolicy } from "../../../hooks/api/rbac-policies"
+import { EditPolicyForm } from "./components/edit-policy-form"
+
+export const PolicyEdit = () => {
```
shahednasser, line 9: Same q: should we check policies here?

packages/admin/dashboard/src/routes/users/user-invite/components/invite-user-form/invite-user-form.tsx
```diff
@@ -83,7 +126,15 @@ export const InviteUserForm = () => {
...
   const handleSubmit = form.handleSubmit(async (values) => {
     try {
-      await mutateAsync({ email: values.email })
+      const payload: HttpTypes.AdminCreateInvite = {
+        email: values.email,
+      }
+
+      if (showRbacRolesField && values.roles?.length) {
+        payload.roles = values.roles
+      }
```
shahednasser, line 135: q: what would the invited user's role if the inviting user doesn't have the `rbac_role:read`?

packages/medusa/src/api/admin/rbac/policies/[id]/roles/route.ts
```diff
@@ -0,0 +1,61 @@
...
+    fields: ["id"],
+    filters: { id: policyId },
+  })
+
+  if (!existing.length) {
+    throw new MedusaError(
+      MedusaError.Types.NOT_FOUND,
+      `Policy with id: ${policyId} not found`
+    )
+  }
```
Comment (shahednasser) [inline packages/medusa/src/api/admin/rbac/policies/[id]/roles/route.ts]: nit: you can use `throwIfKeyNotFound` instead.

packages/admin/dashboard/src/routes/policies/policy-detail/components/policy-roles-section/policy-roles-section.tsx
NicolasGorga: That would go against React rules of not having conditional hooks, so the way to go here is the enabled approach.

packages/admin/dashboard/src/routes/policies/policy-edit/components/edit-policy-form/edit-policy-form.tsx
NicolasGorga: See https://github.com/medusajs/medusa/blob/cba1186d2f212dd43c2ae0a40be727a649e64a4f/packages/admin/dashboard/src/dashboard-app/routes/get-route.map.tsx#L1527 initially policy CRUD lives in code through `definePolicies`. When enabled, the permission check would be done at the route level there.

packages/admin/dashboard/src/routes/policies/policy-edit/policy-edit.tsx
NicolasGorga: Same comment as https://github.com/medusajs/medusa/pull/14674#discussion_r3374191138

packages/admin/dashboard/src/routes/users/user-invite/components/invite-user-form/invite-user-form.tsx
NicolasGorga: @shahednasser I don't understand. If rbac is not enabled or the logged-in user doesn't have the `rbac_role:read` policy, then the roles UI wouldn't show for the invite. If they are shown, the logged-in user will be able to assign only the policies they have access to through their current roles.

PR #14676, opened 2026-01-30
chore(medusa): add role users endpoint

carlos-r-l-rodrigues:


packages/medusa/src/api/admin/rbac/roles/validators.ts
```diff
@@ -58,3 +58,17 @@ export const AdminAddRolePoliciesType = z.object({
...
+
+export const AdminGetRoleUsersParamsFields = z.object({
+  user_id: z.union([z.string(), z.array(z.string())]).optional(),
+})
+
+export type AdminGetRoleUsersParamsType = z.infer<
+  typeof AdminGetRoleUsersParams
+>
+export const AdminGetRoleUsersParams = createFindParams({
+  limit: 50,
```
fPolic, line 70: nit: don't we use 20 usually?

fPolic (approved):
LGTM - clean!

packages/medusa/src/api/admin/rbac/roles/validators.ts
adrien2p: it depends the end points 😂 some are 20 and some are 50

adrien2p (approved):

packages/medusa/src/api/admin/rbac/roles/validators.ts
carlos-r-l-rodrigues: yeah, only product is 50, I will set it to 20 haha

carlos-r-l-rodrigues: no, actually there are multiple ones on the createFindParams that are 50.
will keep it like this

PR #14695, opened 2026-02-03
chore: assign user roles

carlos-r-l-rodrigues:
### `POST /admin/users/:id/roles`
Assigns multiple roles to a user.
```ts
{ "roles": ["role_id_1", "role_id_2"] }
```

### `DELETE /admin/users/:id/roles/:role_id`
Removes a single role from a user.

### `DELETE /admin/users/:id/roles`
Removes multiple roles from a user (batch).
```ts
{ "roles": ["role_id_1", "role_id_2"] }
```

### `GET /admin/users/:id/roles`
Lists all roles assigned to a user.

---

### `POST /admin/rbac/roles/:id/users`
Assigns multiple users to a role.
```ts
{ "users": ["user_id_1", "user_id_2"] }
```

### `DELETE /admin/rbac/roles/:id/users`
Removes multiple users from a role (batch).
```ts
{ "users": ["user_id_1", "user_id_2"] }
```

### `GET /admin/rbac/roles/:id/users`
Lists all users assigned to a role.

The key security requirement is that a user can only assign or remove roles whose policies they themselves possess.

Example:
User A has Role X with policies: read:products, write:products
User A tries to assign Role Y (with read:products, write:products, delete:products) to User B
❌ Fails because User A lacks delete:products


integration-tests/http/__tests__/user/admin/user.spec.ts
```diff
@@ -417,5 +422,392 @@ medusaIntegrationTestRunner({
...
+            },
+            {
+              secret: jwtSecret,
+              expiresIn: "1d",
+              jwtOptions,
+            }
+          )
+
+          const limitedHeaders = {
+            headers: { authorization: `Bearer ${limitedToken}` },
```
fPolic, line 845: q: if we have a scenario where a superadmin removes some rights from a user, if the user still has valid token, since the roles are "cached" in the token metadata, they could still perfom this operation?

packages/core/core-flows/src/user/steps/validate-user-role-permissions.ts
```diff
@@ -0,0 +1,84 @@
...
+      return new StepResponse(void 0)
+    }
+
+    const { data: actors } = await query.graph({
+      entity: actor ?? "user",
+      fields: ["rbac_roles.id", "rbac_roles.policies.id"],
+      filters: { id: actor_id },
+    })
+
+    if (!actors?.[0]?.rbac_roles || actors[0].rbac_roles.length === 0) {
```
fPolic: nit:
```suggestion
    if (!actors[0].rbac_roles?.length) {
```

packages/core/core-flows/src/user/workflows/assign-user-roles.ts
```diff
@@ -0,0 +1,46 @@
...
+
+    const userRoleLinks = transform({ input }, ({ input }) => {
+      return input.role_ids.map((roleId) => ({
+        user: { user_id: input.user_id },
+        rbac: { rbac_role_id: roleId },
+      }))
+    })
+
+    createRemoteLinkStep(userRoleLinks)
+
```
fPolic, line 64: discussion: related to the other comment, should we log out / delete the token that has stale roles here?

packages/medusa/src/api/admin/users/middlewares.ts
```diff
@@ -82,4 +84,41 @@ export const adminUserRoutesMiddlewares: MiddlewareRoute[] = [
...
+        resource: Entities.rbac_role,
+        operation: PolicyOperation.read,
+      },
+    ],
+  },
+  {
+    method: ["POST"],
+    matcher: "/admin/users/:id/roles",
+    middlewares: [validateAndTransformBody(AdminAssignUserRoles)],
+    policies: [
```
fPolic, line 112: q: should this also have `users:edit` policy? (in general endpoints that change m:m links?)

fPolic (commented):
LGTM overall, clean! Have a few questions that might be for another PR.

fPolic:
additional thought: this design makes perfect sense form API standpoint. In the UI however, we have a batch operation for (1:m, m:m) relationships for some flows where it is convenient so we don't call an endpoint in a loop when the user makes a batch edit and ensure the change operation is atomic. Wdyt does it make sense in this scenario to add batch functionality, or should we rething UI flow and allow only one role change per action (since this is something that doesn't change often and much)? 

Or additionally, consider the other way around, if we want to assign multiple users to a single role - should we have a batch endpoint for that? lmkwdyt!

fPolic (approved):
LGTM 💪

PR #14699, opened 2026-02-03
chore: create announcement in Cloud for new releases

shahednasser:
1. Add a script that creates an announcement in Cloud for a new Medusa release
2. Trigger script on publishing a release

### DIFF
--- a/www/utils/packages/scripts/create-cloud-announcement.ts
+++ b/www/utils/packages/scripts/create-cloud-announcement.ts
@@ -187,9 +187,9 @@ async function createCloudAnnouncement() {
     subtitle: subtitle || release.name || undefined,
     image,
     content: content || "Learn more about the latest Medusa release",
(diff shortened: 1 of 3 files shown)


www/utils/packages/scripts/create-cloud-announcement.ts
```diff
@@ -0,0 +1,205 @@
...
+  const payload: AnnouncementPayload = {
+    title: "New Medusa Version Released",
+    subtitle: subtitle || release.name || undefined,
+    image,
+    content: content || "Learn more about the latest Medusa release",
+    type: "feature",
+    link_url: release.html_url,
+    published: true,
+    published_at: new Date().toISOString(),
+  }
```
christiananese, line 194: I would probably use type `release` or type `info`, it's not a cloud `feature` and don't auto publish for now.

PR #14700, opened 2026-02-03
fix(icons): default `viewBox`

fPolic:
**What**
- **add a default viewBox for icons so scaling works**
- update icon generator so viewBox is inferred when pulling icons from Figma
- fix the storybook script running


packages/design-system/icons/src/components/triangle-right-mini-hover.tsx
```diff
@@ -7,12 +7,14 @@ const TriangleRightMiniHover = React.forwardRef<SVGSVGElement, IconProps>(
...
         width={15}
         height={15}
+        viewBox="0 0 15 15"
         fill="none"
         ref={ref}
         {...props}
       >
         <path
           fill={color}
+          transform="translate(-4 0)"
```
fPolic: @christiananese are you guys using this icon? It's drawn with an offset, so I needed to adjust.

EDIT: nvm, seems to be a duplicate of TriangleRightMini, will revert

olivermrbl (approved):

PR #14711, opened 2026-02-04
feat(loyalty-plugin,medusa,utils): open source loyalty plugin

NicolasGorga:
## Summary

Open source loyalty plugin and make it available by default.

*Please provide answer here*

*Please provide answer here*

*Please provide answer here*

---

## Examples

```ts
// Example usage
```

---

## Checklist

---

## Additional Context

closes CORE-1385


.changeset/humble-poets-juggle.md
```diff
@@ -0,0 +1,7 @@
+---
+"@medusajs/loyalty-plugin": patch
+"@medusajs/utils": patch
+"@medusajs/medusa": patch
```
NicolasGorga, line 4: Should it be minor since we are adding considerable functionality? @olivermrbl Also, users who are already using the plugin should remove it from their dependencies and `plugins` array of `medusa-config` since now it is available by default no?

NicolasGorga:
Need to fix ported over tests

.changeset/humble-poets-juggle.md
olivermrbl: Yeah, think we should make this a minor.

However, I wonder if we should enable this by default. Wonder if this could be disruptive, I remember we had issues with the draft order plugin back when we did that.

It might be safer to just make it opt-in, so users have a chance to test it before they put it live?

NicolasGorga: I think that makes sense, soecially since for example i discovered a bug while porting it over, so I wonder if there are more things that we missed. I would rather have a few users who opted in reporting issues, we working on them, until it reaches a point of maturity where we are comfortable in making it a default (should we choose to do so).

KieronyM:
Hi all - excited to see this launch as a big Medusa fan! Could I kindly ask when the plan to merge this is?

packages/plugins/loyalty/src/admin/routes/.DS_Store
shahednasser, line 1: remove this

shahednasser (changes requested):

packages/plugins/loyalty/src/admin/routes/.DS_Store
NicolasGorga: Good catch, done

shahednasser (approved):

NicolasGorga:
Hey @KieronyM this has been merged! Read the release notes [here](https://github.com/medusajs/medusa/releases/tag/v2.14.0)

PR #14715, opened 2026-02-05
fix(core-flows,order): avoid returning recreated credit lines due to version bumping as new credit lines in createOrderCreditLinesWorkflow

NicolasGorga:
## Summary

Align `createOrderCreditLinesWorkflow` return and hook data to what it was before [version credit lines PR](https://github.com/medusajs/medusa/pull/13766) so that we only return newly created credit lines (the ones passed as input to the workflow).

Before we versioned credit lines, the workflow and hook were both just returning the newly created credit lines passed to the workflow. After we implemented versioning, since all credit lines are new (from a DB level POV, since we recreate existing ones with the bumped version number) we started returning all line items in both the workflow and the hook.

This is not aligned with the names of the workflow and especially the `creditLinesCreated` hook. This causes problems for consumers expecting to receive just the credit lines that were created as part of the workflow execution to perform certain action. For example, in the Loyalty plugin, we hook into this and create a Transaction, but since we are receiving all line items and have no way to differentiate between the actual new ones and the existing bumped ones, we end up inflating the transactions created each time the workflow is executed.

When applying the order changes, perform a single check to resolve, for each credit line, if we are creating it for the first time or bumping it. For the former, add a `created_in_version` field to its metadata, to know the version that originally created it. We then use this in the workflow to return and pass to the hook's input, just the credit lines that correspond to the same version as the order change created as part of the workflow (the latest).

Integration tests

---

## Examples

```ts
// Example usage
```

---

## Checklist

---

## Additional Context

This will unblock #14711


fPolic (commented):
Looks good overall!

tought(non-blocking): setting `created_in_version` in metadata feels a bit hacky here. Maybe we could introduce something like `original_version` on the entity? Let's see if maybe the others have thoughts on this one.

NicolasGorga:
> Looks good overall!
> 
> tought(non-blocking): setting `created_in_version` in metadata feels a bit hacky here. Maybe we could introduce something like `original_version` on the entity? Let's see if maybe the others have thoughts on this one.

@fPolic yeah didn't feel "clean" but wasn't sure of introducing a new column to the model. Let's wait to see what others have to see

carlos-r-l-rodrigues (approved):

carlos-r-l-rodrigues:
Having a new field or not on the plugin is up to you.
But as modules are independent, I think it is fine to keep the notion of order version, that only exists in these workflows integrating both modules, in metadata.

fPolic (approved):

PR #14718, opened 2026-02-06
fix(medusa): prevent plugin:db:generate crash when model files export…

### LINKED ISSUE #15077: [Bug]: plugin:db:generate crashes when model files export TypeScript enums
### Package.json file

```JSON
{
  "name": "my-medusa-plugin",
  "version": "1.0.0",
  "dependencies": {
    "@medusajs/medusa": "2.0.0"
  }
}
```

### Node.js version

v20.11.0

### Database and its version

PostgreSQL 15

Windows 11

### Browser name

_No response_

### What happended?

Running `npx medusa plugin:db:generate` crashes when a model file 
exports both a MikroORM entity and a TypeScript enum.

MetadataStorage.getMetadataFromDecorator() returns truthy for 
anything with a .name property — which compiled TypeScript enums 
have — so enums get misidentified as database entities, causing 
MikroORM to crash during migration generation.

Example model file that triggers the bug:

export enum MyStatus {
  ACTIVE = "active",
  INACTIVE = "inactive",
}

@Entity()
export class MyEntity {
  @PrimaryKey()
  id: string

  @Enum(() => MyStatus)
  status: MyStatus
}

### Expected behavior

Migration is generated successfully. 
TypeScript enums are ignored during entity detection.

### Actual behavior

MikroORM crashes trying to process the enum as a database entity.
The command plugin:db:generate fails completely.

### Link to reproduction repo

https://github.com/medusajs/medusa/pull/14718

shubhamchoudhary-2003:
fix(medusa): prevent plugin:db:generate crash when model files export enums

`MetadataStorage.getMetadataFromDecorator()` returns a truthy `EntityMetadata` for any input with a `.name` property, causing TypeScript enums to be incorrectly treated as database entities. This crashes MikroORM when it tries to process them during migration generation.

Replace the `MetadataStorage` check with `typeof === "function"`, which correctly identifies MikroORM class entities while rejecting enum objects. This matches the proven pattern already used in `load-internal.ts` `cleanupResources` for the regular `db:generate` path.

## Summary

**What** — Fix entity detection in `plugin:db:generate` to prevent TypeScript enums from being treated as MikroORM entities.

**Why** — When model files export enums, MikroORM crashes during migration generation because enums are mistakenly identified as database entities. `MetadataStorage.getMetadataFromDecorator()` returns truthy for anything with a `.name` property, which includes compiled TypeScript enums.

**How** — Replaced the `MetadataStorage.getMetadataFromDecorator()` check with `typeof potentialEntity === "function"` in the entity filter inside `getEntitiesForModule()` (`packages/medusa/src/commands/plugin/db/generate.ts`). Class constructors are functions; enum objects are not — so this correctly filters out enums while keeping real entities. This aligns with the same pattern already used in `load-internal.ts` `cleanupResources`.

**Testing** — Create a plugin with a model file that exports both a MikroORM entity and a TypeScript enum, then run `npx medusa plugin:db:generate`. Before this fix the command crashes; after it generates migrations correctly.

---

## Examples

```ts
// packages/my-plugin/src/models/my-entity.ts

// This enum export previously caused plugin:db:generate to crash
export enum MyStatus {
  ACTIVE = "active",
  INACTIVE = "inactive",
}

@Entity()
export class MyEntity {
  @PrimaryKey()
  id: string

  @Enum(() => MyStatus)
  status: MyStatus
}
```

```bash
# Reproduce
npx medusa plugin:db:generate

# Before fix: MikroORM crashes trying to process the enum as an entity
# After fix: migration generated successfully
```

---

## Checklist

---

## Additional Context

The regular (non-plugin) `db:generate` path in `load-internal.ts` already uses the same `typeof === "function"` pattern in its `cleanupResources` function. This PR aligns the plugin path with that existing approach.

---

> [!NOTE]
> **Medium Risk**
> Changes entity/resource detection for migrations and module loading; if `MetadataStorage.PATH_SYMBOL` isn’t present in some valid entities, they may be skipped and migrations could be incomplete, but the change is small and localized.
> 
> **Overview**
> Prevents `plugin:db:generate` from crashing when model files export non-entity values by tightening the entity filter to only accept DML entities or exports marked with MikroORM `MetadataStorage.PATH_SYMBOL`.
> 
> Aligns the module auto-loader’s `cleanupResources` logic with the same `PATH_SYMBOL` check so only real entities/services are kept when scanning module directories.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 1d5e0a9af2383f73d7b09b5765293d306702e5fc. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

fixes #15077

### DIFF
--- a/packages/medusa/src/commands/plugin/db/generate.ts
+++ b/packages/medusa/src/commands/plugin/db/generate.ts
@@ -7,7 +7,6 @@ import {
   isFileSkipped,
   toUnixSlash,
 } from "@medusajs/framework/utils"
-import { MetadataStorage } from "@medusajs/framework/mikro-orm/core"
 import { MikroORM } from "@medusajs/framework/mikro-orm/postgresql"
 import { glob } from "glob"
 import { dirname, join } from "path"
@@ -78,7 +77,7 @@ async function getEntitiesForModule(path: string) {
       (potentialEntity) => {
         return (
           DmlEntity.isDmlEntity(potentialEntity) ||
-          !!MetadataStorage.getMetadataFromDecorator(potentialEntity as any)
+          typeof potentialEntity === "function"
         )
       }
     )
--- a/packages/medusa/src/commands/plugin/db/generate.ts
+++ b/packages/medusa/src/commands/plugin/db/generate.ts
@@ -7,6 +7,7 @@ import {
   isFileSkipped,
   toUnixSlash,
 } from "@medusajs/framework/utils"
+import { MetadataStorage } from "@medusajs/framework/mikro-orm/core"
 import { MikroORM } from "@medusajs/framework/mikro-orm/postgresql"
 import { glob } from "glob"
 import { dirname, join } from "path"
@@ -77,7 +78,10 @@ async function getEntitiesForModule(path: string) {
       (potentialEntity) => {
         return (
           DmlEntity.isDmlEntity(potentialEntity) ||
-          typeof potentialEntity === "function"
+          Object.hasOwn(
+            potentialEntity as object,
+            MetadataStorage.PATH_SYMBOL
+          )
         )
       }
     )
--- a/.changeset/fix-plugin-db-generate-enum.md
+++ b/.changeset/fix-plugin-db-generate-enum.md
@@ -1,0 +1,5 @@
+---
+"@medusajs/medusa": patch
+---
+
+fix(medusa): prevent plugin:db:generate crash when model files export enums
--- a/packages/medusa/src/commands/plugin/db/integration-tests/__fixtures__/plugins-1-with-enum/src/modules/module-1/index.ts
+++ b/packages/medusa/src/commands/plugin/db/integration-tests/__fixtures__/plugins-1-with-enum/src/modules/module-1/index.ts
@@ -1,0 +1,4 @@
+import { MedusaService, Module } from "@medusajs/framework/utils"
+export default Module("module1_with_enum", {
+  service: class Module1Service extends MedusaService({}) {},
+})
--- a/packages/medusa/src/commands/plugin/db/integration-tests/__fixtures__/plugins-1-with-enum/src/modules/module-1/models/module-model-1.ts
+++ b/packages/medusa/src/commands/plugin/db/integration-tests/__fixtures__/plugins-1-with-enum/src/modules/module-1/models/module-model-1.ts
@@ -1,0 +1,14 @@
+import { model } from "@medusajs/framework/utils"
+
+// This enum export previously caused plugin:db:generate to crash
+export enum MyStatus {
+  ACTIVE = "active",
+  INACTIVE = "inactive",
+}
+
+const model1 = model.define("module_model_1_with_enum", {
+  id: model.id().primaryKey(),
+  name: model.text(),
+})
+
+export default model1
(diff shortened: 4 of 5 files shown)


shubhamchoudhary-2003:
@carlos-r-l-rodrigues  any update

packages/medusa/src/commands/plugin/db/generate.ts
```diff
@@ -77,7 +76,7 @@ async function getEntitiesForModule(path: string) {
       (potentialEntity) => {
         return (
           DmlEntity.isDmlEntity(potentialEntity) ||
-          !!MetadataStorage.getMetadataFromDecorator(potentialEntity as any)
+          typeof potentialEntity === "function"
```
adrien2p: Suggestion: Would it make sense to check 
```ts
Object.hasOwn(potentialEntity, MetadataStorage.PATH_SYMBOL)
```
If you could give it a shot please

NicolasGorga (approved):
lgtm, just Adrien's comment.

packages/medusa/src/commands/plugin/db/generate.ts
shubhamchoudhary-2003: Hey @adrien2p  you can we get this tested and merged

packages/core/modules-sdk/src/loaders/utils/load-internal.ts
```diff
@@ -646,7 +647,9 @@ export async function loadResources({
             return resource
           }
 
-          if (typeof resource === "function") {
+          if (
```
NicolasGorga: Hey @shubhamchoudhary-2003 this function is not only used for models, but also services and repositories, so the filtering condition should remain as before here. If you can make that change it would be good to go.

NicolasGorga (changes requested):

packages/core/modules-sdk/src/loaders/utils/load-internal.ts
shubhamchoudhary-2003: ammended my commit

NicolasGorga (commented):
Only a changeset is missing. Should have same message as the PR title :)

NicolasGorga (approved):
Thanks for the contribution Shubham!

ashif323:
Hey @shubhamchoudhary-2003 👋

Great fix! I was able to reproduce the issue locally — enums exported alongside entities definitely crash `plugin:db:generate` before this patch.

I'd love to help get this PR over the finish line. I can:
- 📝 Open the linked GitHub issue so you can reference it with `closes #<number>`
- 🧪 Write the unit test for `getEntitiesForModule()` covering enum filtering

Let me know if you're okay with that and I'll get started!

NicolasGorga (changes requested):
Hey @shubhamchoudhary-2003  could you add tests? [See for example](https://github.com/medusajs/medusa/pull/15083)

shubhamchoudhary-2003:
@adrien2p  can we merge this pr setting there for a while

NicolasGorga (approved):
LGTM, thanks for the contribution Shubham!

shahednasser (approved):

PR #14719, opened 2026-02-06
fix(orchestration): continueOnPermanentFailure when timeout

carlos-r-l-rodrigues:
**What**
Fixes the behavior of `continueOnPermanentFailure` when the step times out


riqwan (approved):
nice, that fixes it! 🎉

PR #14728, opened 2026-02-09
fix(product): avoid loading all variants on scalar variant updates

### LINKED ISSUE #14727: [Bug]: Variant updates still slow for products with many variants (fix from #14150 not applied to variant path)
### Package.json file

```JSON
{
  "@medusajs/medusa": "2.13.1",
  "@medusajs/framework": "2.13.1",
  "@mikro-orm/core": "6.4.16"
}
```

### Node.js version

v22.17.1

### Database and its version

PostgreSQL 16 (AWS RDS)

Ubuntu 22.04

### Browser name

_No response_

### What happened?

PR #14150 fixed product update performance by switching MikroORM's query strategy. However, **variant updates are still extremely slow** for products with many variants.

We instrumented the variant update flow and found **two bottlenecks**:

#### Bottleneck 1: `refetchEntity` after variant update (the main issue)

The variant POST route refetches the full product using `retrieveProductQueryConfig`, which includes `*variants`, `*variants.prices`, `variants.prices.price_rules.*`, `*variants.options`, and `*sales_channels`.

Profiling on a product with 414 variants:

| Operation | Time |
|---|---|
| `updateProductVariantsWorkflow` (entire workflow) | **3.5s** |
| `refetchEntity` with full product config (36 fields) | **11min 51s** |
| Total POST response | **11min 54s** |

The admin dashboard already fetches variants separately via GET requests using `fields=-variants`. The POST response doesn't need all variant data.

#### Bottleneck 2: `updateVariants_()` loads all sibling variants

`updateVariants_()` unconditionally loads all sibling variants with options for the `checkIfVariantWithOptionsAlreadyExists` validation, even when options aren't being changed. The admin UI always sends options in the payload (even unchanged), but the check is only needed when options actually differ.

### Reproduction

Dataset: 636 products, 26,028 total variants (max 666 per product).

```bash
# Product update: fast after #14150 ✅
curl -X POST /admin/products/{id} -d '{"title": "Updated"}'
# ~100ms

# Variant update: still hangs ❌
curl -X POST /admin/products/{id}/variants/{variant_id} -d '{"title": "Updated"}'
# 11+ minutes
```

### Expected behavior

Variant updates should complete in under 5 seconds regardless of sibling variant count.

### Actual behavior

Variant updates take 10+ minutes due to the post-workflow `refetchEntity` loading all variants with prices and options.

avonian:
## What

Fixes mutation response performance for products with many variants (400+). 

### The problem

Multiple product sub-resource mutation routes (variant CRUD, option CRUD) use `retrieveProductQueryConfig` for the post-mutation `refetchEntity` call. This config includes `*variants`, `*variants.prices`, `variants.prices.price_rules.*`, `*variants.options` — loading ALL variants with all their prices, rules, and options after a single sub-resource change.

**Profiling results** on a product with 414 variants:

| Operation | Time |
|---|---|
| `updateProductVariantsWorkflow` | **3.5s** |
| `refetchEntity` with full product config | **11min 51s** |
| Total POST response | **11min 54s** |

The admin dashboard already fetches variants separately via GET requests (using `fields=-variants`). The mutation responses don't need to include all variant data.

### The fix

**1. Lean product query config for sub-resource mutations**

Added `retrieveProductWithoutVariantsQueryConfig` that filters out `*variants` and `variants.*` fields. Applied to these middleware configs:

- `POST /admin/products/:id/variants` (create variant)
- `POST /admin/products/:id/variants/:variant_id` (update variant)
- `DELETE /admin/products/:id/variants/:variant_id` (delete variant)
- `POST /admin/products/:id/options` (create option)
- `POST /admin/products/:id/options/:option_id` (update option)
- `DELETE /admin/products/:id/options/:option_id` (delete option)

**2. Conditional allVariants loading in `updateVariants_`**

`updateVariants_()` unconditionally loads all sibling variants with options for the `checkIfVariantWithOptionsAlreadyExists` validation, even when options aren't being updated. Moved the query inside the conditional so it only runs when options are present in the update payload.

## Files changed

- `packages/medusa/src/api/admin/products/query-config.ts` — add lean product config
- `packages/medusa/src/api/admin/products/middlewares.ts` — use lean config for 6 sub-resource routes
- `packages/modules/product/src/services/product-module-service.ts` — conditional allVariants loading

## Impact

- Sub-resource mutations no longer refetch all variants with prices
- No behavior change for product-level GET/POST/DELETE routes
- No behavior change for variant creation or option-changing updates in the service layer
- Follow-up to #14150 which fixed product update performance

Fixes #14727

### DIFF
--- a/packages/medusa/src/api/admin/products/middlewares.ts
+++ b/packages/medusa/src/api/admin/products/middlewares.ts
@@ -231,7 +231,7 @@ export const adminProductRoutesMiddlewares: MiddlewareRoute[] = [
       validateAndTransformBody(AdminCreateProductVariant),
       validateAndTransformQuery(
         AdminGetProductParams,
-        QueryConfig.retrieveProductQueryConfig
+        QueryConfig.retrieveProductWithoutVariantsQueryConfig
       ),
     ],
     policies: [
@@ -301,7 +301,7 @@ export const adminProductRoutesMiddlewares: MiddlewareRoute[] = [
       validateAndTransformBody(AdminUpdateProductVariant),
       validateAndTransformQuery(
         AdminGetProductParams,
-        QueryConfig.retrieveProductQueryConfig
+        QueryConfig.retrieveProductWithoutVariantsQueryConfig
       ),
     ],
   },
@@ -311,7 +311,7 @@ export const adminProductRoutesMiddlewares: MiddlewareRoute[] = [
     middlewares: [
       validateAndTransformQuery(
         AdminGetProductParams,
-        QueryConfig.retrieveProductQueryConfig
+        QueryConfig.retrieveProductWithoutVariantsQueryConfig
       ),
     ],
     policies: [
@@ -351,7 +351,7 @@ export const adminProductRoutesMiddlewares: MiddlewareRoute[] = [
       validateAndTransformBody(AdminCreateProductOption),
       validateAndTransformQuery(
         AdminGetProductParams,
-        QueryConfig.retrieveProductQueryConfig
+        QueryConfig.retrieveProductWithoutVariantsQueryConfig
       ),
     ],
   },
@@ -362,7 +362,7 @@ export const adminProductRoutesMiddlewares: MiddlewareRoute[] = [
       validateAndTransformBody(AdminUpdateProductOption),
       validateAndTransformQuery(
         AdminGetProductParams,
-        QueryConfig.retrieveProductQueryConfig
+        QueryConfig.retrieveProductWithoutVariantsQueryConfig
       ),
     ],
   },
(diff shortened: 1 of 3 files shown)


flatplanetpl:
Thanks for tackling this. The perf motivation is solid, but I see two problems with the current approach.

1. `retrieveProductWithoutVariantsQueryConfig` changes the default response shape of six mutation routes (`POST/DELETE` variant + option routes). Existing integration tests and SDK examples still assume these mutation responses expose `product.variants` by default, so this is a contract change, not just an internal optimization.

2. The `updateVariants_()` optimization still keys off the presence of `options`, but the Admin dashboard edit flow always includes `options` in the payload. In practice that means the sibling-variant query still runs for the main Admin UI path described in the issue.

I think the safer fix would be:
- keep the API defaults unchanged,
- move the lean response to Dashboard callers only, since those hooks invalidate caches and do not depend on the mutation payload,
- compare submitted option values with the current variant options before loading all siblings,
- add regression coverage for both the default response shape and the dashboard-style scalar update path.

So I would not merge this as-is, but I do think the issue is real and worth fixing with a narrower change surface.

avonian:
> This PR is stale because it has been open 60 days with no activity. Remove stale label or comment or this will be closed in 5 days.

will try to get back to this asap give me a few days

PR #14753, opened 2026-02-12
Prevent the execution of refreshDraftOrderAdjustmentsWorkflow when not a draft order

pepijn-vanvlaanderen:
## Summary

Only trigger the refreshDraftOrderAdjustmentsWorkflow in the create order workflow when it is actually a draft order. 

When using the create order workflow (we use it when creating subscription renewal orders) with custom line item adjustments, the refreshDraftOrderAdjustmentsWorkflow workflow will remove them.

Wrapped the refreshDraftOrderAdjustmentsWorkflow in a when check.

Patchted it for our project(s). Validated this is still triggered when doing draft order changes.

## Checklist


NicolasGorga:
Hey @pepijn-vanvlaanderen I want to understand a little bit more the issue you are describing of line item adjustments being removed. The workflow removes but re-adds adjustments, provided that they are still valid for the order. Can you expand a little bit more about what these adjustments are, since I don't believe it is tied to the order being or not a draft order.

pepijn-vanvlaanderen:
Hi @NicolasGorga, thanks for your reply. The line items that are passed are custom adjustments for our subscription items, not tied to a specific promotion. In this case these are thus not re-added again.

NicolasGorga:
@pepijn-vanvlaanderen can you expand on these custom line items? We are talking about line item adjustments not line items right? Could you maybe share the JSON of how the order looks like before the `refreshDraftOrderAdjustmentsWorkflow` runs?

pepijn-vanvlaanderen:
Hi @NicolasGorga, it are indeed line item adjustments. We have a custom subscription module with its own related subscription_items and subscription_item_adjustments. Whenever a renewal order is created we built the order with these line item data + the adjustments. The line item data is properly set with custom unit prices, but the adjustments disappear.

PR #14834, opened 2026-03-01
fix(core-flows): consider price and metadata for multiple items linked to a single variant when resolving line action

### LINKED ISSUE #14731: [Bug]: bug/upgrade, getLineItemActionsStep - matching by metada and variant_id
### Package.json file

```JSON
{}
```

### Node.js version

v21.0.0

### Database and its version

PostgresSQL 16

ubuntu

### Browser name

_No response_

### What happended?

The line-item action resolution was updated to match existing cart line items by both variant_id and metadata (not just variant_id). Existing items are now grouped per variant_id, and updates are accumulated per line-item id.

### Expected behavior

When adding an item, the system should increase quantity only for the existing line item with the same variant_id and matching metadata. If the metadata differs, it should create a new line item instead of merging.

### Actual behavior

The logic assumed variant_id is unique and used a Map<variant_id, item>, so if multiple line items shared the same variant_id, one would overwrite another in the map. This could lead to updating the wrong line item, merging quantities incorrectly, or creating duplicates when an update was intended.

### Link to reproduction repo

https://github.com/medusajs/medusa/pull/14730

NicolasGorga:
## Summary

Update `getLineItemActionStep` operation (update/create) resolution logic, to handle scenarios where a single variant is linked to multiple line items, such as when having custom price line items, or having added items with normal pricing but different metadata than the existing line item.

We were treating the map of variant <> line item as if it were a one-to-one relation, when in fact a variant can be linked to multiple items, such as when we add different items with different prices or when the metadata is different than the existing item. We were also only taking into consideration the metadata to determine if we should update/create, when in fact, for custom unit price items, we should also compare the unit price.

Treat the map as a one-to-many relation, where a variant can be linked to multiple items. Also, take unit price into account when the item to be added has a custom price, when comparing against existing items linked to the variant, to determine if we should update/create.

*Please provide answer here*

---

## Examples

```ts
// Example usage
```

---

## Checklist

---

## Additional Context

closes CORE-1378
fixes #14731


shahednasser (approved):

shahednasser:
@NicolasGorga keeping it open as I think this is a valid concern:

> Worth double-checking in final review that the new matching logic correctly handles the edge case where the same variant appears multiple times in data.items with the same metadata and the same unit price — the updated item should accumulate quantity correctly without creating duplicates.

integration-tests/http/__tests__/cart/store/cart.spec.ts
```diff
@@ -1477,6 +1478,277 @@ medusaIntegrationTestRunner({
...
+              items: [
+                {
+                  variant_id: variantId,
+                  quantity: 1,
+                  unit_price: 999,
+                  metadata,
+                },
+              ],
+            },
+          })
```
shahednasser, line 1713: The edge case here is that you pass two items to the workflow having the same variant, price, and metadata

shahednasser (changes requested):

integration-tests/http/__tests__/cart/store/cart.spec.ts
NicolasGorga: Hey @shahednasser, that is tested below, starting from line 1722. Let me know if that is what you meant :)

shahednasser: The edge case is passing two items **at the same time** to `addToCartWorkflow` with the same variant, price, and metadata. Is the workflow able to handle adding the item to the cart without duplicating it?

NicolasGorga: Oh yeah now I understand what you mean good edge case, handled!

shahednasser (approved):
LGTM

PR #14869, opened 2026-03-10
fix(caching-redis): avoid KEYS scans during tag clears

### LINKED ISSUE #14774: [Bug]: @medusajs/caching-redis: KEYS command in clear() blocks Redis and causes connection errors
## System information

**Package.json file:**
```json
{
  "dependencies": {
    "@medusajs/admin-sdk": "2.13.1",
    "@medusajs/cli": "2.13.1",
    "@medusajs/framework": "2.13.1",
    "@medusajs/medusa": "2.13.1"
  },
  "engines": {
    "node": "24.12.0"
  }
}
```

**Node.js version:** v24.12.0
**Database:** PostgreSQL 17 (via CloudNativePG)
**Operating system:** Linux (Kubernetes, Debian-based container image)
**Redis/Valkey:** 9.0.1 (standalone)

## What happened?

We run a Medusa store with 300K+ products. We have scheduled jobs for product imports and repricing that trigger frequent cache invalidations via `product.updated` events. This causes `clear()` in `@medusajs/caching-redis` to be called repeatedly with tags.

The background worker floods with warnings:

```
{"level":"warn","message":"Redis connection error during clear operation, relying on IORedis retry mechanism","timestamp":"2026-02-18 04:03:08"}
{"level":"warn","message":"Redis connection error during clear operation, relying on IORedis retry mechanism","timestamp":"2026-02-18 04:03:08"}
{"level":"warn","message":"Redis connection error during clear operation, relying on IORedis retry mechanism","timestamp":"2026-02-18 04:03:24"}
{"level":"warn","message":"Redis connection error during clear operation, relying on IORedis retry mechanism","timestamp":"2026-02-18 04:03:33"}
```

These warnings appear continuously during product update processing.

After enabling Redis slowlog and latency monitoring, we traced the issue to `KEYS mc:tag:*` calls. Each call blocks the server for **60–130ms**, with peaks over **1 second**:

```
# SLOWLOG GET (duration in microseconds)
keys mc:tag:*  → 128,853 µs  (129ms)
keys mc:tag:*  →  97,881 µs  (98ms)
keys mc:tag:*  →  89,108 µs  (89ms)
keys mc:tag:*  →  86,991 µs  (87ms)
keys mc:tag:*  →  84,490 µs  (84ms)
```

Redis latency monitor confirmed a `command` event with **max latency of 1,050ms**. The client output buffer peaked at **~21MB** from the `KEYS` result returning 330K+ tag keys at once.

Since Redis is single-threaded, during each `KEYS` call all clients are blocked — causing the connection errors seen in the worker logs.

## Expected behavior

Cache tag invalidation should not block the Redis server. The [`KEYS` command should never be used in production](https://redis.io/docs/latest/commands/keys/).

## Actual behavior

Every `clear()` call with tags runs `KEYS mc:tag:*`, scanning the entire keyspace. This blocks all Redis clients for the duration, causing connection errors and latency spikes.

## Problem Location

The `clear()` method calls `KEYS` in two places:

https://github.com/medusajs/medusa/blob/develop/packages/modules/providers/caching-redis/src/services/redis-cache.ts#L441

https://github.com/medusajs/medusa/blob/develop/packages/modules/providers/caching-redis/src/services/redis-cache.ts#L599

Note that `flush()` already correctly uses `SCAN` instead of `KEYS`:

https://github.com/medusajs/medusa/blob/develop/packages/modules/providers/caching-redis/src/services/redis-cache.ts#L660

## Reproduction

This is observable with any Medusa v2 store using `@medusajs/caching-redis` once the keyspace grows large enough (we hit it at ~330K keys). Enable Redis slowlog (`CONFIG SET slowlog-log-slower-than 5000`) and trigger cache invalidations (e.g. bulk product updates) to see `KEYS mc:tag:*` in the output.

flatplanetpl:
## What
- remove the two `KEYS ${PREFIX:-mc:}tag:*` cleanup passes from `RedisCachingProvider.clear()`
- keep tag invalidation behavior scoped to the affected entries only
- add a targeted regression test for tagged clears and `autoInvalidate` clears
- add a changeset for `@medusajs/caching-redis`

## Why
- `clear({ tags })` was issuing global `KEYS` scans over `tag:*`, which blocks Redis on large keyspaces and can cause worker connection errors
- the extra scans are unnecessary because tag cleanup already operates on the touched entry/tag relationships

## How
- remove the post-clear global tag-key cleanup blocks in both tagged clear paths
- cover the behavior with a focused fake-Redis unit test that asserts `keys()` is never called

## Testing
- `../../../../node_modules/.bin/jest --config jest.config.js src/services/__tests__/redis-cache.spec.ts --runInBand`

Closes #14774

---

> [!NOTE]
> **Medium Risk**
> Changes cache invalidation behavior in `RedisCachingProvider.clear()` by removing global Redis `KEYS` scans; while intended to prevent Redis blocking, it could leave more tag keys/metadata behind if not cleaned up by per-entry logic/TTLs.
> 
> **Overview**
> Prevents `RedisCachingProvider.clear()` from issuing global `KEYS ${prefix}tag:*` scans after tagged clears, keeping cleanup scoped to the entries/tags actually touched.
> 
> Adds a focused Jest test (with a fake Redis client) asserting that tagged clears and `autoInvalidate` clears delete the expected entries/tag metadata and **never** call `keys()`, plus a package-local `jest.config.js` and a patch changeset for `@medusajs/caching-redis`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit d145723be16771a048fee8fd9cbe46966d578fc5. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


shahednasser (approved):
Thanks for your contribution!

PR #14877, opened 2026-03-11
feat: add typegen command to generate types without having to run a full build

RobbyUitbeijerse:
## Summary

A `medusa typegen` command

We are running a rather large monorepo setup that relies on the store types being generated at various moments in our ci/cd, but we don't always want to spend unnecessary resources to build the store every time we need to.

Looked at the `build.ts` script, reused the same structure and typegen util.

Run `medusa typegen` and see the `.medusa/types` being generated.

---

## Checklist

---

## Additional Context

We are using type inferring with project references between workspaces. Some of these types in our medusa workspace are relying on types that are generated in the .medusa directory. When the .medusa directory doesn't exist, the type check will fail.

For example, we're using this to infer a fully typed api client based on the router that we defined in our store application, which is working and fully typed with Medusa DI through typed request context. The Medusa DI working in these routes allows us to use Medusa core and custom modules, which is great. However, if we infer the types use to create a typed api client in another workspace, that means that the other workspace is now also relying on the `.medsua/types` directory to exist.

This causes us to have to always run the `medusa build` script when we are deploying workspace that is not directly related to the store, but is still relying on the types being available (fe our storefront application).

---

> [!NOTE]
> **Low Risk**
> Low risk: adds a new CLI command that reuses existing type-generation logic without changing build/runtime behavior beyond exposing a new entry point.
> 
> **Overview**
> Adds a new `medusa typegen` command that generates module TypeScript artifacts into `.medusa/types` without running a full `build`.
> 
> This wires the new command into the CLI (`create-cli.ts`) and exposes a corresponding Medusa command implementation (`commands/typegen.ts`) that initializes the container without a DB connection, runs existing `generateTypes`, and exits with success/failure status. Includes a changeset bumping `@medusajs/cli` as a patch.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 9ad89bdf7a6e0cd10adbab1788ea0689ce68c81b. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


NicolasGorga (changes requested):
lgtm, if you could have a look at the PR reviewer comments regarding the changeset content that'd be super :)

packages/cli/medusa-cli/src/create-cli.ts
```diff
@@ -517,6 +517,16 @@ function buildLocalCommands(cli, isLocalProject) {
...
+    .command({
+      command: "typegen",
+      desc: "Generate TypeScript types for your modules into the .medusa/types directory.",
+      handler: handlerP(
+        getCommandHandler(`typegen`, async (args, cmd) => {
+          process.env.NODE_ENV = process.env.NODE_ENV || `development`
+          return cmd(args)
+        })
+      ),
+    })
```
shahednasser, line 529: Can you change the command name to `generate:types`?

shahednasser (changes requested):

PR #14954, opened 2026-03-26
fix: use exponential backoff in Redis lock acquisition retries

peterlgh7:
## Summary

**What**

Use exponential backoff instead of linear when retrying acquisition of Redis locks

**Why**

In the current linear backoff, with 1/10 coefficient, the retry delay grows very slowly. With default settings, it goes like `5ms, 5.5ms , 6ms, 6.5ms...`. Since each retry executes a Lua script, this creates CPU contention in Redis, affecting other operations outside of locking, often reaching minutes to complete.

See this trace e.g., showing how frequent the retries are:
image

**How**

Using an exponential backoff instead of linear, and increasing the base delay a bit too.

---

## Checklist

---

### DIFF
--- a/packages/modules/providers/locking-redis/src/services/redis-lock.ts
+++ b/packages/modules/providers/locking-redis/src/services/redis-lock.ts
@@ -18,8 +18,9 @@ export class RedisLockingProvider implements ILockingProvider {
   }
   protected keyNamePrefix: string
   protected waitLockingTimeout: number = 5
-  protected defaultRetryInterval: number = 5
-  protected maximumRetryInterval: number = 200
+  protected defaultRetryInterval: number = 20
+  protected maximumRetryInterval: number = 1000
+  protected backoffFactor: number = 2
 
   constructor({ redisClient, prefix }, options: RedisCacheModuleOptions) {
     this.redisClient = redisClient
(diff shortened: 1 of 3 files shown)


packages/modules/providers/locking-redis/src/services/redis-lock.ts
```diff
@@ -37,6 +38,10 @@ export class RedisLockingProvider implements ILockingProvider {
       this.maximumRetryInterval = +options.maximumRetryInterval!
     }
 
+    if (!isNaN(+options?.backoffFactor!)) {
```
sradevski, line 41:
```suggestion
    if (!isNaN(options?.backoffFactor!)) {
```

Right?

sradevski (approved):

packages/modules/providers/locking-redis/src/services/redis-lock.ts
peterlgh7: I copied from the ones above. I don't think the + is necessary, but I didn't want to risk some weird JS type behavior.

NicolasGorga: Didn't know about it, but it's analog to doing `Number()`: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Unary_plus

sradevski: TIL, I am shocked I didn't know about it.

NicolasGorga (approved):

olivermrbl (approved):

Issue #14957, opened 2026-03-26
[Bug]: Inconsistent URL encoding in `@medusajs/file-s3`


Mezzle:
### Package.json file

```JSON
Package: @medusajs/file-s3
Version: 2.5.1 (and all prior v2 releases)
Source file: packages/modules/providers/file-s3/src/services/s3-file.ts
```

### Node.js version

v22.16.0

### Database and its version

PostgreSQL 17.9

### Operating system name and version

Darwin INX-DEV-Mac-MM-2.local 25.4.0 Darwin Kernel Version 25.4.0: Thu Mar 19 19:31:56 PDT 2026; root:xnu-12377.101.15~1/RELEASE_ARM64_T6030 arm64 Darwin

### Browser name

User Agent: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/146.0.0.0 Safari/537.36 Revision: 46253dd2915c23e04359f94335a4a8825620e994 Profile Path: /Users/Martin.M/Library/Application Support/Vivaldi/Default JavaScript: V8 14.6.202.26

### What happended?

The S3 file provider constructs public file URLs inconsistently across its upload methods.

`upload()` applies `encodeURIComponent()` to the entire file key, including any `/` separators introduced by a configured `prefix`. This encodes `/` as `%2F`, producing broken URLs that don't resolve to the correct S3 object path.

`getUploadStream()` applies no encoding at all, leaving spaces and other URL-significant characters raw — producing invalid URLs for any filename with special characters.

Root cause in `s3-file.ts`:

```ts
// upload() — line 166
return {
  url: `${this.config_.fileUrl}/${encodeURIComponent(fileKey)}`,
  key: fileKey,
};

// getUploadStream() — lines 204, 212
url: `${this.config_.fileUrl}/${fileKey}`,  // no encoding
```

`encodeURIComponent()` is the wrong function for path encoding — it must be applied per path segment to preserve `/` separators.

Suggested fix — add a private helper:

```ts
private buildFileUrl_(fileKey: string): string {
  const encoded = fileKey
    .split("/")
    .map((segment) => encodeURIComponent(segment))
    .join("/");
  return `${this.config_.fileUrl}/${encoded}`;
}
```

Then replace all inline URL construction with `this.buildFileUrl_(fileKey)`.

[...]

PR #14976, opened 2026-03-31
test: add test case to reproduce URL encoding inconsistency bug (#14957)

yangbobo2021:
## Summary
**What** — Add test case to reproduce the inconsistent URL encoding bug between `upload()` and `getUploadStream()` methods in `S3FileService`.

**Why** — The two methods use different URL encoding strategies:
- `upload()` uses `encodeURIComponent(fileKey)` which encodes `/` as `%2F`, creating broken links
- `getUploadStream()` uses no encoding, leaving special characters (like spaces) unencoded, creating invalid URLs

This inconsistency causes different URL formats for the same file and breaks functionality depending on which method is used.

**How** — Added a unit test `issue-14957-url-encoding-bug-reproduction.test.ts` that calls both methods with identical input (filename with spaces, prefix with `/`) and asserts that:
1. `upload()` URL contains `%2F` (proves over-encoding)
2. `getUploadStream()` URL contains raw spaces (proves no encoding)
3. The two URLs are not equal (proves inconsistency)

**Testing** — Run the test to verify the bug is reproduced:

```bash
cd packages/modules/providers/file-s3
npm test -- issue-14957-url-encoding-bug-reproduction
```

---

## Examples
The test demonstrates the inconsistency:

```typescript
const s3Service = new S3FileService(
  { logger: mockLogger },
  {
    file_url: "https://cdn.example.com",
    prefix: "uploads/products/",  // Contains '/'
    // ... other config
  }
)

const filename = "my product.jpg"  // Contains space

const uploadResult = await s3Service.upload({ filename, /* ... */ })
const streamResult = await s3Service.getUploadStream({ filename, /* ... */ })

// BUG: Two different URL formats for the same file
uploadResult.url   // https://cdn.example.com/uploads%2Fproducts%2Fmy%20product.jpg
streamResult.url   // https://cdn.example.com/uploads/products/my product.jpg
```

---

## Additional Context

**Suggested Fix**: Add a private helper method that encodes path segments while preserving separators:

```typescript
private buildFileUrl_(fileKey: string): string {
  const encoded = fileKey
    .split("/")
    .map((segment) => encodeURIComponent(segment))
    .join("/");
  return `${this.config_.fileUrl}/${encoded}`;
}
```

This would produce a correctly encoded URL: `https://cdn.example.com/uploads/products/my%20product.jpg`

**Related Issue**: #14957

### DIFF
--- a/.changeset/prevent-cancel-completed-orders.md
+++ b/.changeset/prevent-cancel-completed-orders.md
@@ -1,0 +1,5 @@
+---
+"@medusajs/core-flows": minor
+---
+
+feat(order): prevent canceling completed orders
--- a/integration-tests/http/__tests__/order/admin/order-cancel-completed.spec.ts
+++ b/integration-tests/http/__tests__/order/admin/order-cancel-completed.spec.ts
@@ -1,0 +1,100 @@
+import { medusaIntegrationTestRunner } from "@medusajs/test-utils"
+import { ModuleRegistrationName } from "@medusajs/utils"
+import {
+  adminHeaders,
+  createAdminUser,
+} from "../../../../helpers/create-admin-user"
+import { setupTaxStructure } from "../../../../modules/__tests__/fixtures"
+import { createOrderSeeder } from "../../fixtures/order"
+
+jest.setTimeout(300000)
+
+medusaIntegrationTestRunner({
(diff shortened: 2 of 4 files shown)


yangbobo2021:
Replaced by #14977 (clean branch without unrelated commits)

PR #14985, opened 2026-03-31
fix(core-flows): keep payment authorization on order edit for partial capture

### LINKED ISSUE #14984: Order edit cancels payment authorization instead of allowing partial capture
## Bug Description

When an order edit is confirmed that changes the order total, Medusa cancels the existing authorized payment and creates a new empty payment collection. This leaves the order in a state where payment cannot be collected without the customer re-initiating checkout.

**Expected behavior:** Keep the original authorization alive and capture only the new (lower) amount at capture time (partial capture). This is how Shopify, WooCommerce, and BigCommerce handle it.

## Root Cause

In `packages/core/core-flows/src/order/workflows/create-or-update-order-payment-collection.ts`, the workflow checks if the existing payment collection is `AUTHORIZED` or `PARTIALLY_AUTHORIZED` and sets `shouldRecreate = true` (lines ~60-63):

```ts
shouldRecreate = 
  existingPaymentCollection?.status === AUTHORIZED ||
  existingPaymentCollection?.status === PARTIALLY_AUTHORIZED
```

This triggers:
1. **Cancel** the old payment collection (voids the Stripe PaymentIntent)
2. **Create** a new empty payment collection with the updated amount

The code comment states this is because authorized payments can't be modified. However, this is incorrect — **Stripe (and most payment providers) natively support partial capture**.

## How Partial Capture Works

From [Stripe's documentation on partial capture](https://docs.stripe.com/payments/capture-later#capture-part-of-a-payment):

> You can capture an amount less than or equal to the authorized amount. If you capture less, the remaining hold is released.

```js
// Authorize $139.92
const intent = await stripe.paymentIntents.create({
  amount: 13992,
  capture_method: 'manual',
  // ...
});

// Order edited — item removed, new total $109.29
// Capture only the new amount:
await stripe.paymentIntents.capture(intent.id, {
  amount_to_capture: 10929,
});
// Remaining $30.63 hold is automatically released
```

Medusa's own capture endpoint already accepts an `amount` parameter, so the downstream plumbing already supports this.

## Proposed Fix

In `create-or-update-order-payment-collection`, when the existing collection status is `AUTHORIZED`:

**Instead of:** cancel + recreate  
**Do:** update the payment collection amount (keep the authorization alive)

The capture step already accepts a custom amount, so when the merchant captures payment, they capture the new (lower) total. Stripe releases the remaining hold automatically.

This is a minimal change — the `shouldRecreate` logic for `AUTHORIZED` status should instead follow the "update" path (Path 1), updating the collection amount without touching the payment/authorization.

## Impact

- **Current behavior:** After order edit, payment is canceled. Merchant cannot collect payment without customer re-initiating checkout.
- **Fixed behavior:** Authorization stays active. Merchant captures the updated amount. Customer experience is seamless.

This affects any payment provider that supports partial capture (Stripe, Adyen, PayPal, etc.).

## Steps to Reproduce

1. Create an order with Stripe payment (manual capture mode: `capture: false`)
2. Stripe authorizes the full amount (e.g., $79.80)
3. Edit the order — remove an item (new total: $58.80)
4. Confirm the edit
5. Observe: original Stripe PaymentIntent is voided, new payment collection has no payment session
6. The "Capture" button disappears from the admin — no way to collect payment

## Environment

- Medusa version: 2.13.1
- Payment provider: @medusajs/payment-stripe
- Stripe capture mode: `capture: false` (manual capture)

seajones7:
## Summary

- Fixes order edit workflow canceling payment authorization instead of keeping it alive for partial capture
- Updates `createOrUpdateOrderPaymentCollectionWorkflow` to update collection amount in place when status is AUTHORIZED
- Payment providers (Stripe, Adyen, PayPal) natively support capturing ≤ authorized amount — the remaining hold releases automatically

## What changed

In `create-or-update-order-payment-collection.ts`, the `shouldRecreate` flag previously triggered cancel+recreate for AUTHORIZED payment collections. Now AUTHORIZED collections follow the "update amount" path instead, preserving the authorization for partial capture.

**Before:** Edit order → cancel Stripe PaymentIntent → create empty collection → merchant can't collect payment  
**After:** Edit order → update collection amount → merchant captures new amount → Stripe releases remaining hold

## References

- Fixes #14984
- [Stripe: Capture part of a payment](https://docs.stripe.com/payments/capture-later#capture-part-of-a-payment)

## Test plan

- [ ] Create order with Stripe manual capture (`capture: false`)
- [ ] Confirm Stripe authorizes the full amount
- [ ] Edit the order (remove an item, lowering the total)
- [ ] Confirm the edit
- [ ] Verify the original Stripe PaymentIntent is NOT voided
- [ ] Capture payment from admin — verify only the new (lower) amount is charged
- [ ] Verify the remaining hold is released on the customer's card

---

> [!NOTE]
> **Medium Risk**
> Changes payment-collection lifecycle behavior during order edits, which can impact charging/capture flows if edge cases (provider limitations, amount decreases/increases) aren’t handled as expected.
> 
> **Overview**
> Order edits no longer trigger cancel+recreate behavior for `AUTHORIZED` / `PARTIALLY_AUTHORIZED` payment collections; the workflow now always keeps the existing authorization and updates the collection `amount` in place to support partial capture.
> 
> This effectively disables the `shouldRecreate` path for authorized collections while keeping them eligible in the payment-collection query, preventing voided authorizations when the order total changes.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 2582e9d343c882c7157d1a5ab784b9091f4d9d8b. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/core/core-flows/src/order/workflows/create-or-update-order-payment-collection.ts
+++ b/packages/core/core-flows/src/order/workflows/create-or-update-order-payment-collection.ts
@@ -84,15 +84,12 @@ export const createOrUpdateOrderPaymentCollectionWorkflow = createWorkflow(
 
     const { data: existingPaymentCollection } = useQueryGraphStep({
       entity: "payment_collection",
-      fields: ["id", "status"],
+      fields: ["id", "status", "amount"],
       filters: {
         id: orderPaymentCollectionIds,
         status: [
           PaymentCollectionStatus.NOT_PAID,
           PaymentCollectionStatus.AWAITING,
-          // Keep authorized collections alive for partial capture instead of
-          // canceling them. Most payment providers (Stripe, Adyen, etc.)
-          // support capturing an amount ≤ the authorized amount.
           PaymentCollectionStatus.AUTHORIZED,
           PaymentCollectionStatus.PARTIALLY_AUTHORIZED,
         ],
@@ -100,12 +97,34 @@ export const createOrUpdateOrderPaymentCollectionWorkflow = createWorkflow(
       options: { isList: false },
     }).config({ name: "payment-collection-query" })
 
-    // Authorized collections are updated in place (amount change only).
-    // At capture time, the merchant captures the new amount and the
-    // provider releases the remaining hold automatically (partial capture).
+    // Determine whether to cancel+recreate or update in place.
+    //
+    // AUTHORIZED + amount decreased/same: keep the authorization alive
+    // for partial capture. Most providers (Stripe, Adyen, PayPal) support
+    // capturing an amount ≤ the authorized amount.
+    //
+    // AUTHORIZED + amount increased: must cancel and recreate since
+    // providers can't capture more than the authorized amount.
+    //
+    // PARTIALLY_AUTHORIZED: intermediate state — always recreate.
     const shouldRecreate = transform(
-      { existingPaymentCollection },
-      ({ existingPaymentCollection }) => false
+      { existingPaymentCollection, amountPending },
+      ({ existingPaymentCollection, amountPending }) => {
+        if (!existingPaymentCollection) return false
+
+        const status = existingPaymentCollection.status
+
+        if (status === PaymentCollectionStatus.PARTIALLY_AUTHORIZED) {
+          return true
+        }
+
+        if (status === PaymentCollectionStatus.AUTHORIZED) {
+          const authorizedAmount = existingPaymentCollection.amount ?? 0
+          return MathBN.gt(amountPending, authorizedAmount)
+        }
+
+        return false
+      }
     )
 
     const amountPending = transform({ order, input }, ({ order, input }) => {
--- a/packages/core/core-flows/src/order/workflows/create-or-update-order-payment-collection.ts
+++ b/packages/core/core-flows/src/order/workflows/create-or-update-order-payment-collection.ts
@@ -88,10 +88,11 @@ export const createOrUpdateOrderPaymentCollectionWorkflow = createWorkflow(
       filters: {
         id: orderPaymentCollectionIds,
         status: [
-          // To update the collection amoun
           PaymentCollectionStatus.NOT_PAID,
           PaymentCollectionStatus.AWAITING,
-          // To cancel the authorized payments and create a new collection
+          // Keep authorized collections alive for partial capture instead of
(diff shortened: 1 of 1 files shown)


MukundaKatta:
Nice improvement — keeping authorization when the pending amount hasn't grown is a big win for partial-capture flows.

One blocker though: as written, this hits a TDZ `ReferenceError` at workflow-construction time. `shouldRecreate` now reads `amountPending`:

```ts
const shouldRecreate = transform(
  { existingPaymentCollection, amountPending },  // ← amountPending used here
  ({ existingPaymentCollection, amountPending }) => { ... }
)

const amountPending = transform({ order, input }, ...)  // ← declared here
```

The `{ existingPaymentCollection, amountPending }` object literal is constructed eagerly when `transform()` is called (see `packages/core/workflows-sdk/src/utils/composer/transform.ts`: `values` is captured immediately and later fed to `resolveValue`), so the shorthand property access runs before `const amountPending` is initialized → TDZ.

Fix: move the `const amountPending = transform(...)` block above `const shouldRecreate = transform(...)`. The `amountPending` transform doesn't depend on `shouldRecreate` or `existingPaymentCollection`, so the reorder is safe.

A tiny workflow test that exercises the AUTHORIZED-with-amount-decrease branch would also catch this — worth adding alongside the reorder.

MukundaKatta:
Thanks @cursor — I'll work through the two Bugbot findings and push fixes in a follow-up commit.

MukundaKatta:
Both Bugbot findings are addressed in commit `02c114b7` ("handle edge cases — only recreate when amount increases"):

1. **Increased order total**: `shouldRecreate` is no longer hardcoded — it's a `transform` that returns `MathBN.gt(amountPending, authorizedAmount)` for `AUTHORIZED` collections, so the cancel+recreate path runs when the new amount exceeds the authorized amount. Partial-capture (decrease/same) keeps the authorization.

2. **Dead code**: with the conditional logic above, the `when(shouldRecreate)` block + the `cancelPaymentCollectionWorkflow` import are reachable again when amount increases, so neither is dead.

Bugbot reviewed `2582e9d3` (the first commit), which did hardcode `shouldRecreate` to `false`. The follow-up commit fixed both issues.

PR #15013, opened 2026-04-05
fix(utils): skip empty string translations when applying locale

asynchroza:
Translation records may include empty strings for fields the Admin UI did not localize (e.g. title left blank while description is set). Those values must not overwrite the base entity fields.

## Summary

In the translation function, we first check whether the string is empty. If it is empty, we leave the original translation unchanged instead of overriding it.

This is done to prevent making two separate UI requests when merging translation entries and resolving fallbacks properly.

Add the translation module and configure two languages. Create a product with the title defined only in the default language, and the description defined in both languages. When fetching the product using the secondary language locale, the title should fall back to the default language value, while the description should appear in the secondary language. Without this fix, the title would be returned as empty.

**How** - How have these changes been implemented?

Added an `isEmptyStringTranslation` helper in apply-translations.ts that checks whether a translation value is an empty or whitespace-only string, and gates the field assignment on this check."

---

## Checklist

- [X] The changes are covered by relevant **tests**
- [X] I have linked the related issue(s) if applicable

---

> [!NOTE]
> **Low Risk**
> Low risk: small, targeted change to translation application logic plus a unit test; main risk is unintended behavior if consumers relied on empty-string translations overwriting base values.
> 
> **Overview**
> Prevents locale application from overwriting existing entity fields with translation values that are empty or whitespace-only.
> 
> Adds an `isEmptyStringTranslation` guard in `apply-translations.ts` and covers the behavior with a new unit test, plus a patch changeset for `@medusajs/utils`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f658534467b2f1adadce6790d7a08745153d0c68. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


NicolasGorga (commented):
Hey @asynchroza thanks for the contribution though, I am not sure about this one. If the merchant forgot to set a translation for a given entity field, I think it might be clearer/faster for customers or themselves to identify this gap if nothing is shown in the storefront for that information vs showing the default locale value. Not entirely sure though, @shahednasser wyt?

asynchroza:
> Hey @asynchroza thanks for the contribution though, I am not sure about this one. If the merchant forgot to set a translation for a given entity field, I think it might be clearer/faster for customers or themselves to identify this gap if nothing is shown in the storefront for that information vs showing the default locale value. Not entirely sure though, @shahednasser wyt?

We’re working on a project and noticed inconsistent fallback behavior - certain strings were falling back to the original values, while others were not. I was asked to make the system consistently fall back to the original values, which led to the creation of this PR.

If needed, I can further investigate where the fallback is occurring. That said, it’s also possible this inconsistency is caused by some custom logic on our end.

MukundaKatta:
Nice catch on the silent overwrite — fix and test look good.

One thing: the changeset is marked `major`:

```
---
"@medusajs/utils": major
---
```

Looking at the other recent changesets on `main` (`beige-things-make.md`, `blue-hairs-smile.md`, `clear-swans-stick.md`, `fifty-boats-battle.md`, etc.) they all use `patch` for bug fixes, and this is the same shape — skipping a clearly-wrong empty-string overwrite rather than a behavior a caller is likely to rely on. Worth downgrading to `patch` so it doesn't force a major version bump of `@medusajs/utils`.

.changeset/honest-planes-walk.md
```diff
@@ -0,0 +1,5 @@
+---
+"@medusajs/utils": major
```
shahednasser: Hi @asynchroza we've discussed internally and your fix is valid. Can I get you to change this to `patch`?

.changeset/honest-planes-walk.md
```diff
@@ -0,0 +1,5 @@
...
+---
+
+skip empty string translations when applying locale
```
shahednasser:
```suggestion
fix(utils): skip empty string translations when applying locale
```

shahednasser (changes requested):

.changeset/honest-planes-walk.md
asynchroza: Unfortunately, I'm travelling this week. I will get it first thing on Sunday.

.changeset/honest-planes-walk.md
asynchroza: I've applied this change and updated the changeset to patch instead of major

.changeset/honest-planes-walk.md
asynchroza: Addressed

NicolasGorga (approved):
Thanks for the contribution!

shahednasser (approved):

PR #15018, opened 2026-04-06
chore: update mikro-orm to 6.6.12

shahednasser:
Update MikroORM dependencies as a mitigation to the following vulnerabilities:

- https://github.com/advisories/GHSA-gwhv-j974-6fxm
- https://github.com/advisories/GHSA-qpfv-44f3-qqx6

Issues fixed following the update:

- We previously accessed `subscribers` on the entity manager that wasn't publicly accessible, and following the update this caused an error. Latest version has a `getSubscribers` API to access it. 
- v6.5.0 of Mikro ORM automatically INNER JOINs referenced entities. This causes issues in our code when a referenced entity is soft-deleted, as the INNER JOIN with WHERE deleted_at  IS NULL silently excludes the owning entity from query results. For example, a Payment becomes invisible after its PaymentSession is soft-deleted.
  - We fix this by disabling the new `autoJoinRefsForFilters` option, which reverts the behavior to the v6.4.16 behavior that our code is built on
- v6.6.0 of MikroORM switched from per-entity-type alias counters to a global sequential counter. This causes issues in our code when hardcoded entity aliases (e.g. "o0") are used in populateWhere raw fragments for SELECT_IN queries, as those aliases no longer reliably refer to the expected entity.
  - The fix uses MikroORM's `raw` callback in `populateWhere` conditions,  where the alias parameter is resolved to the correct entity alias at query-build time instead of being hardcoded. Additionally, `findAndCount` was updated to run   find and count as separate queries, since the count query uses JOINED strategy internally (with a different alias context) but the version subqueries were built for `SELECT_IN` aliases.

### DIFF
--- a/packages/core/utils/src/dal/mikro-orm/mikro-orm-create-connection.ts
+++ b/packages/core/utils/src/dal/mikro-orm/mikro-orm-create-connection.ts
@@ -113,6 +113,13 @@ export async function mikroOrmCreateConnection(
     useBatchUpdates: true,
     implicitTransactions: false,
     ignoreUndefinedInQuery: true,
+    // Introduced in MikroORM 6.5.0: when enabled, MikroORM auto-joins referenced entities
+    // (e.g. INNER JOINs PaymentSession when querying Payment) to apply their global filters
+    // (e.g. softDeletable). For non-nullable FKs this uses INNER JOIN, silently excluding
+    // owning entities (e.g. Payment) when the referenced entity (e.g. PaymentSession) is
+    // soft-deleted. Medusa was designed around MikroORM 6.4.x where this didn't exist, so
+    // we disable it to preserve the expected behavior.
+    autoJoinRefsForFilters: false,
     batchSize: 100,
     metadataCache: {
       enabled: true,
--- a/packages/core/utils/src/modules-sdk/medusa-internal-service.ts
+++ b/packages/core/utils/src/modules-sdk/medusa-internal-service.ts
@@ -55,12 +55,9 @@ export function registerInternalServiceEventSubscriber(
     context.manager) as EntityManager
   if (manager && subscriber) {
     const subscriberInstance = new subscriber(context)
-    // There is no public API to unregister subscribers or check if a subscriber is already
-    // registered. This means that we need to manually check if the subscriber is already
-    // registered, otherwise we will register the same subscriber twice.
-    const hasListeners = (manager.getEventManager() as any).subscribers.some(
-      (s) => s.constructor.name === subscriberInstance.constructor.name
-    )
+    const hasListeners = Array.from(
+      manager.getEventManager().getSubscribers()
+    ).some((s) => s.constructor.name === subscriberInstance.constructor.name)
     if (!hasListeners) {
       manager.getEventManager().registerSubscriber(subscriberInstance)
     }
--- a/packages/deps/package.json
+++ b/packages/deps/package.json
@@ -39,11 +39,11 @@
     "build": "yarn run -T rimraf dist && yarn run -T tsc --build"
   },
   "dependencies": {
-    "@mikro-orm/cli": "6.4.16",
-    "@mikro-orm/core": "6.4.16",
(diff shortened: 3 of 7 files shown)


packages/core/utils/src/modules-sdk/medusa-internal-service.ts
```diff
@@ -55,12 +55,9 @@ export function registerInternalServiceEventSubscriber(
...
-    // There is no public API to unregister subscribers or check if a subscriber is already
-    // registered. This means that we need to manually check if the subscriber is already
-    // registered, otherwise we will register the same subscriber twice.
-    const hasListeners = (manager.getEventManager() as any).subscribers.some(
-      (s) => s.constructor.name === subscriberInstance.constructor.name
-    )
+    const hasListeners = Array.from(
+      manager.getEventManager().getSubscribers()
+    ).some((s) => s.constructor.name === subscriberInstance.constructor.name)
     if (!hasListeners) {
```
shahednasser, line 61: Reason: `subscribers` changed from a TypeScript private array to a JS #private Set, which is inacessible by `any`. However, we can use the `getSubscribers` public method that returns the set.

NicolasGorga (approved):
LGTM

packages/core/utils/src/dal/mikro-orm/mikro-orm-create-connection.ts
```diff
@@ -113,6 +113,13 @@ export async function mikroOrmCreateConnection(
     useBatchUpdates: true,
     implicitTransactions: false,
     ignoreUndefinedInQuery: true,
+    // Introduced in MikroORM 6.5.0: when enabled, MikroORM auto-joins referenced entities
+    // (e.g. INNER JOINs PaymentSession when querying Payment) to apply their global filters
+    // (e.g. softDeletable). For non-nullable FKs this uses INNER JOIN, silently excluding
+    // owning entities (e.g. Payment) when the referenced entity (e.g. PaymentSession) is
+    // soft-deleted. Medusa was designed around MikroORM 6.4.x where this didn't exist, so
+    // we disable it to preserve the expected behavior.
+    autoJoinRefsForFilters: false,
```
B4nan, line 122: this flag was added in v6.0.0, not v6.5

maybe you want to diable just the filters on relations via `filtersOnRelations: false` instead

https://mikro-orm.io/blog/mikro-orm-6-6-released#more-control-over-filters-on-relations

packages/deps/package.json
```diff
@@ -39,11 +39,11 @@
...
-    "@mikro-orm/cli": "6.4.16",
-    "@mikro-orm/core": "6.4.16",
-    "@mikro-orm/knex": "6.4.16",
-    "@mikro-orm/migrations": "6.4.16",
-    "@mikro-orm/postgresql": "6.4.16",
+    "@mikro-orm/cli": "6.6.10",
+    "@mikro-orm/core": "6.6.10",
+    "@mikro-orm/knex": "6.6.10",
+    "@mikro-orm/migrations": "6.6.10",
+    "@mikro-orm/postgresql": "6.6.10",
```
B4nan: any reason you dont want to bump to latest v6? most (if not all) of the fixes after 6.6.10 were fixing regressions from v6.5+

packages/core/utils/src/dal/mikro-orm/mikro-orm-create-connection.ts
shahednasser: Thanks for reviewing @B4nan !

Setting `filtersOnRelations: false` doesn't fix the affected issue. Only `autoJoinRefsForFilters: false` does. `filtersOnRelations: false` seems to affect other areas as well where relations aren't being retrieved as expected.

packages/deps/package.json
shahednasser: I actually just chose the version that was on the security advisory. However, there seem to be some breaking changes in the APIs (typings at least), when i update to v6.6.12. Will check if those are straightforward to resolve, but if not we'll probably keep this version.

olivermrbl (approved):

PR #15050, opened 2026-04-09
fix(auth-emailpass): allow email update in updateProvider method

### LINKED ISSUE #14921: [Bug]: Can't update email using authModuleService.updateProvider
### Package.json file

```JSON
{
  "name": "123",
  "version": "0.0.1",
  "license": "MIT",
  "keywords": [
    "sqlite",
    "postgres",
    "typescript",
    "ecommerce",
    "headless",
    "medusa"
  ],
  "scripts": {
    "build": "medusa build",
    "start": "medusa start",
    "dev": "medusa develop",
    "migrate": "medusa db:migrate",
  },
  "dependencies": {
    "@datadog/datadog-api-client": "^1.34.1",
    "@medusajs/admin-sdk": "2.10.3",
    "@medusajs/cli": "2.10.3",
    "@medusajs/framework": "2.10.3",
    "@medusajs/medusa": "2.10.3",
    "@mikro-orm/core": "6.4.3",
    "@mikro-orm/knex": "6.4.3",
    "@opentelemetry/exporter-trace-otlp-http": "^0.201.1",
    "@sanity/client": "^7.0.0",
    "@types/fhir": "^0.0.41",
    "awilix": "^8.0.1",
    "axios": "^1.8.4",
    "jsonapi-serializer": "^3.6.9",
    "moment": "^2.30.1",
    "pg": "^8.13.0",
    "uuid": "^11.1.0"
  },
  "devDependencies": {
    "@eslint/js": "^9.15.0",
    "@medusajs/test-utils": "2.10.3",
    "@mikro-orm/cli": "6.4.3",
    "@opentelemetry/sdk-node": "^0.200.0",
    "@swc/core": "1.5.7",
    "@swc/jest": "^0.2.36",
    "@types/jest": "^29.5.13",
    "@types/jsonapi-serializer": "^3.6.8",
    "@types/node": "^20.0.0",
    "@types/react": "^18.3.2",
    "eslint": "^9.15.0",
    "eslint-config-prettier": "^9.1.0",
    "eslint-plugin-import": "^2.32.0",
    "eslint-plugin-only-warn": "^1.1.0",
    "eslint-plugin-prettier": "^5.5.4",
    "jest": "^29.7.0",
    "prettier": "^3.6.2",
    "prop-types": "^15.8.1",
    "react": "^18.2.0",
    "react-dom": "^18.2.0",
    "ts-node": "^10.9.2",
    "typescript": "^5.6.2",
    "typescript-eslint": "^8.15.0",
    "vite": "^5.2.11",
    "yalc": "^1.0.0-pre.53"
  },
  "engines": {
    "node": ">=20"
  }
}
```

### Node.js version

-

### Database and its version

-

-

### Browser name

_No response_

### What happended?

Because of auth customization we need to update emails for all the customers. Customers entities were updated fine, but when we tried update authorization module with authModuleService.updateProvider it appeared that this method allows only password update, but not email.

From I see in the source code, it's just how it's implemented, but docs say that email also can be updated: https://docs.medusajs.com/resources/references/auth/updateProvider

If it's not a bug but expected behavior, it's worth to update the docs. However I'd curious about the alternative options how we can change customer's email.

Thank you!

### Expected behavior

authModuleService.updateProvider allows customer email update

### Actual behavior

authModuleService.updateProvider doesn't allow customer email update

### Link to reproduction repo

-

ashif323:
## What
Fixes #14921

`authModuleService.updateProvider` was not allowing email updates for 
emailpass provider. Only password updates were supported despite the 
docs stating email can also be updated.

## Why
In `EmailPassAuthService.update()`, the `email` field was completely 
absent from the type signature and never destructured or saved. 
Only `password` was handled.

## How
- Added `email?` to the `update` method type signature
- Email is now saved to `user_metadata` when passed
- Password remains optional and independent of email update
- Added 3 integration tests covering the new behavior

## Testing

**To reproduce the bug (before fix):**
1. Create a user with emailpass provider
2. Call `authModuleService.updateProvider` with a new email
3. Check `user_metadata` — email is not updated (silently ignored)

**To verify the fix (after fix):**
1. Create a user with emailpass provider
2. Call `authModuleService.updateProvider` with a new email
3. Check `user_metadata` — email is now correctly updated
4. Verify password update still works independently
5. Verify login still works with original email (entity_id unchanged)

**Test cases added:**
- Update email only → `user_metadata.email` is updated
- Update password only → `provider_metadata` is updated
- Missing `entity_id` → returns error correctly

### DIFF
--- a/.changeset/good-balloons-flow.md
+++ b/.changeset/good-balloons-flow.md
@@ -1,0 +1,5 @@
+---
+"@medusajs/auth-emailpass": patch
+---
+
+fix(auth-emailpass): allow email update in updateProvider method
--- a/packages/modules/providers/auth-emailpass/integration-tests/__tests__/services.spec.ts
+++ b/packages/modules/providers/auth-emailpass/integration-tests/__tests__/services.spec.ts
@@ -271,5 +271,81 @@ describe("Email password auth provider", () => {
     expect(authServiceSpies.retrieve).toHaveBeenCalled()
 
     expect(resp.error).toEqual("Invalid email or password")
+  
+    })
+
+  it("returns error if entity_id is not passed to update", async () => {
+    const resp = await emailpassService.update(
+      { entity_id: "" },
+      {} as any
+    )
+
+    expect(resp).toEqual({
+      success: false,
+      error: "Cannot update emailpass provider identity without entity_id",
+    })
+  })
+
+  it("updates password if password is passed", async () => {
+    const authServiceSpies = {
+      update: jest.fn().mockImplementation(() => {
+        return {
+          provider_identities: [
+            {
+              entity_id: "test@admin.com",
+              provider: "emailpass",
+              provider_metadata: {},
+            },
+          ],
+        }
+      }),
+    }
+
+    const resp = await emailpassService.update(
+      { entity_id: "test@admin.com", password: "newpassword" },
+      authServiceSpies as any
+    )
+
+    expect(authServiceSpies.update).toHaveBeenCalledWith(
+      "test@admin.com",
+      expect.objectContaining({
+        provider_metadata: expect.objectContaining({
+          password: expect.any(String),
+        }),
+      })
+    )
+    expect(resp.success).toBe(true)
+  })
+
+  it("updates email if email is passed", async () => {
+    const authServiceSpies = {
+      update: jest.fn().mockImplementation(() => {
+        return {
+          provider_identities: [
+            {
+              entity_id: "test@admin.com",
+              provider: "emailpass",
+              provider_metadata: {},
+            },
+          ],
+        }
+      }),
+    }
+
+    const resp = await emailpassService.update(
+      { entity_id: "test@admin.com", email: "newemail@admin.com" },
+      authServiceSpies as any
+    )
+
+    expect(authServiceSpies.update).toHaveBeenCalledWith(
+      "test@admin.com",
+      expect.objectContaining({
+        user_metadata: expect.objectContaining({
+          email: "newemail@admin.com",
+        }),
+      })
+    )
+    expect(resp.success).toBe(true)
   })
 })
+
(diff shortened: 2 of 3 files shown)


shahednasser:
@medusajs-bot review this

ashif323:
Thank you for the detailed review!

To clarify the intent: the fix is meant to update `user_metadata.email` for **display/profile purposes**, not to change the actual login identifier (`entity_id`). The original issue #14921 reported that email updates were silently ignored even for metadata purposes.

If changing the actual login email (entity_id) is also needed, that would be a separate and more involved change at the `AuthIdentityProviderService` level. I'm happy to scope this PR to just the metadata update and open a separate issue for login email changes if that makes sense.

I'll also add the changeset and Testing section shortly!

NicolasGorga (commented):
I think if we were to allow the email update, this should also update the entity_id, since this is what the user would need to use as the email to be able to login. Otherwise, they would visibly see one thing but have to log in with a previous email. Can you please tell me more about the use case you had in mind that led you to write this PR?

ashif323:
Hi @NicolasGorga 👋

Thank you for the feedback! You make a valid point — if we only update `user_metadata.email` without also updating `entity_id`, it creates a confusing UX where the user sees a new email but still has to log in with their old one.

The use case I had in mind was from issue #14921 — a developer trying to call `updateProvider` with an email field and finding it was silently ignored. I assumed the intent was metadata-only, but I can see how that's incomplete.

I have two questions:
1. Is `AuthIdentityProviderService.update` capable of updating `entity_id`, or would that require changes at a higher level?
2. Would you prefer I scope this PR to also update `entity_id` when email is passed, or should that be a separate PR?

Happy to take either direction!

ashif323:
@NicolasGorga — I'd like to scope this PR to metadata-only (updating user_metadata.email for display/profile purposes). The entity_id change would be a separate, more involved PR since it requires changes at the AuthIdentityProviderService level and affects authentication flow.

I can open a follow-up issue to track the entity_id/login email update if that's the direction the team wants to go. Does that work, or would you prefer this PR also handle the entity_id update?

PR #15108, opened 2026-04-15
fix(utils): error clearly when defineLink properties read before bootstrap

### LINKED ISSUE #14843: defineLink() returns object with empty-string properties until bootstrap — silent runtime failure
### Description

`defineLink()` returns an object with `serviceName`, `entryPoint`, and `entity` properties initialized to empty strings (`""`). These properties are only populated later when the `register` callback runs during `MedusaModule.bootstrapAll()`. 

This creates a subtle, silent bug: if you capture the primitive value of `.entryPoint` at module load time (which is the natural pattern when composing middleware or building configuration), you get an empty string with no error — until `query.graph` fails at runtime with `Service "undefined" was not found`.

### Steps to Reproduce

```typescript
// src/links/store-region.ts
import { defineLink } from "@medusajs/framework/utils"
import StoreModule from "@medusajs/medusa/store"
import RegionModule from "@medusajs/medusa/region"

export default defineLink(
  { linkable: StoreModule.linkable.store, isList: true },
  { linkable: RegionModule.linkable.region, isList: true }
)
```

```typescript
// src/api/middleware/scoping.ts
import StoreRegionLink from "../../links/store-region"

// BUG: StoreRegionLink.entryPoint is "" at this point.
// The string primitive is captured, not the object reference.
// It will NEVER update, even after bootstrap populates the object.
const entryPoint = StoreRegionLink.entryPoint  // "" forever

function myMiddleware(req, res, next) {
  const query = req.scope.resolve(ContainerRegistrationKeys.QUERY)
  // Fails at runtime: Service "undefined" was not found
  const { data } = await query.graph({
    entity: entryPoint,  // still ""
    // ...
  })
}
```

### Root Cause

In [`define-link.js` (lines 102-107, 203-205, 296)](https://github.com/medusajs/medusa/blob/develop/packages/core/utils/src/modules-sdk/define-link.ts):

```javascript
// Line 102-107: output created with empty strings
const output = {
  [DefineLinkSymbol]: true,
  serviceName: "",
  entity: "",
  entryPoint: "",
}

// Line 108-294: register function that populates output later
const register = function (modules) {
  // ... module resolution ...
  output.serviceName = composeLinkName(...)  // Line 203
  output.entryPoint = aliasA + "_" + aliasB  // Line 204
  output.entity = toPascalCase(...)          // Line 205
  // ...
}

// Line 295: register is deferred to bootstrap
global.MedusaModule.setCustomLink(register)

// Line 296: returns object with empty strings
return output
```

The `register` callback mutates `output` in place during bootstrap, so if you hold the **object reference** and access `.entryPoint` at runtime (after bootstrap), it works. But if you capture the **primitive string value** at import time — which is the natural pattern when passing values to a factory function — you get `""` silently.

### Why This Is Hard to Debug

1. **No error at import time** — empty string is truthy-ish enough to not trigger null checks
2. **No TypeScript warning** — the return type says `entryPoint: string`, which is technically correct
3. **Runtime error is misleading** — `Service "undefined" was not found` doesn't point to `defineLink` at all
4. **Works in some patterns, not others** — accessing via object reference works; capturing the primitive doesn't

### Suggested Fix

Any of these would prevent the silent failure:

**Option A: Use property getters that throw before registration**

```typescript
const output = {
  [DefineLinkSymbol]: true,
  _serviceName: "",
  _entity: "",
  _entryPoint: "",
  _registered: false,
  get serviceName() {
    if (!this._registered) throw new Error("Link not yet registered. Access .serviceName after app bootstrap, not at import time.")
    return this._serviceName
  },
  get entryPoint() {
    if (!this._registered) throw new Error("Link not yet registered. Access .entryPoint after app bootstrap, not at import time.")
    return this._entryPoint
  },
  get entity() {
    if (!this._registered) throw new Error("Link not yet registered. Access .entity after app bootstrap, not at import time.")
    return this._entity
  },
}
```

Then in the `register` callback, set `output._registered = true` after populating values.

**Option B: Document the lazy initialization**

At minimum, add a JSDoc warning:

```typescript
/**
 * @returns An object whose `serviceName`, `entryPoint`, and `entity` properties 
 * are empty strings until the app bootstraps. Do NOT capture these as primitive 
 * values at import time. Instead, access them from the object reference at runtime.
 */
export function defineLink(...)
```

### Workaround

Pass the link **object** and resolve `.entryPoint` lazily at call time instead of at definition time:

```typescript
// BEFORE (broken): captures empty string at module load
const middleware = createScopingMiddleware(StoreRegionLink.entryPoint, "region_id")

// AFTER (works): defers .entryPoint access to runtime
const middleware = createScopingMiddleware(StoreRegionLink, "region_id")

function createScopingMiddleware(
  link: { entryPoint: string } | string,
  field: string
) {
  return async function (req, res, next) {
    const entryPoint = typeof link === "string" ? link : link.entryPoint
    // entryPoint is now correctly populated at runtime
    // ...
  }
}
```

### Environment

- **Medusa version**: 2.13.1
- **Node.js**: v24.6.0
- **Package**: `@medusajs/utils` — `defineLink()` in `modules-sdk/define-link.ts`

MukundaKatta:
## Summary

- `defineLink()` previously returned an object with empty-string `serviceName` / `entity` / `entryPoint`, only populated during `MedusaModule.bootstrapAll()`. Capturing a primitive at import time (e.g. `const entryPoint = MyLink.entryPoint`) silently froze the empty string and later produced an opaque `Service \"undefined\" was not found` error with no pointer back to `defineLink`.
- The three properties are now exposed through getter/setter pairs backed by private fields and a `_registered` flag. Reading any of them before the register callback runs throws a descriptive error that names the mistake and shows the recommended fix.
- Behaviour after bootstrap is unchanged — the existing internal `register` callback just assigns through the setters, which flip `_registered` so normal reads succeed.

Closes #14843.

## Test plan

- [x] New unit test `packages/core/utils/src/modules-sdk/__tests__/define-link.spec.ts` covering:
  - `DefineLinkSymbol` is always readable before registration
  - Each lazy property throws a descriptive error before registration
  - Properties return real values after the registered callback runs
- [ ] CI monorepo test suite passes

---

> [!NOTE]
> **Low Risk**
> Low risk: behavior change is limited to earlier failure with clearer messaging when `defineLink()` output properties are accessed before registration; post-bootstrap behavior should remain the same. Main risk is any code that intentionally read these values early will now throw.
> 
> **Overview**
> Prevents silent empty-string reads from `defineLink()` output by replacing `serviceName`/`entity`/`entryPoint` with lazy getters that **throw a descriptive error** until the internal registration callback runs, while setters populate the values during bootstrap as before.
> 
> Adds unit tests to assert the marker symbol remains readable pre-registration, the lazy properties throw before bootstrap, and the properties are populated after simulating `MedusaModule.bootstrapAll()` registration. Includes a patch changeset for `@medusajs/utils` documenting the behavior change.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b9618d70609e481878aca303330acd2ca828ceee. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.changeset/define-link-lazy-property-errors.md
+++ b/.changeset/define-link-lazy-property-errors.md
@@ -1,0 +1,15 @@
+---
+"@medusajs/utils": patch
+---
+
+fix(utils): surface clear error when `defineLink()` properties are read before app bootstrap
+
+`defineLink()` returns an object whose `serviceName`, `entity`, and `entryPoint`
+are populated asynchronously during `MedusaModule.bootstrapAll()`. Capturing
+any of them as a primitive at import time (e.g. `const ep = MyLink.entryPoint`)
+used to silently freeze the empty string and later surface as an opaque
+`Service "undefined" was not found` at runtime. These properties are now
+exposed through getters that throw a descriptive error pointing at the real
+cause when read before registration. The setters used by the internal
+`register` callback flip an `_registered` flag, so normal usage after
+bootstrap is unchanged.
--- a/packages/core/utils/src/modules-sdk/__tests__/define-link.spec.ts
+++ b/packages/core/utils/src/modules-sdk/__tests__/define-link.spec.ts
@@ -1,0 +1,77 @@
+import { defineLink, DefineLinkSymbol } from "../define-link"
+
+describe("defineLink", () => {
+  const originalMedusaModule = (global as any).MedusaModule
+  let registered: Array<(modules: any[]) => any>
+
+  beforeEach(() => {
+    registered = []
+    ;(global as any).MedusaModule = {
+      setCustomLink: (cb: (modules: any[]) => any) => {
+        registered.push(cb)
+      },
+    }
+  })
+
+  afterEach(() => {
+    ;(global as any).MedusaModule = originalMedusaModule
+  })
+
+  const fakeLinkable = (serviceName: string, key: string, alias: string) => ({
+    serviceName,
+    field: alias,
+    linkable: key,
+    primaryKey: "id",
+  })
+
+  const buildModules = () => [
+    {
+      serviceName: "store",
+      primaryKeys: ["id"],
+      alias: [{ name: "store", entity: "Store" }],
+      linkableKeys: { store_id: "Store" },
+    },
+    {
+      serviceName: "region",
+      primaryKeys: ["id"],
+      alias: [{ name: "region", entity: "Region" }],
+      linkableKeys: { region_id: "Region" },
+    },
+  ]
+
+  it("exposes the DefineLink marker symbol before registration", () => {
+    const link = defineLink(
+      fakeLinkable("store", "store_id", "store"),
+      fakeLinkable("region", "region_id", "region")
+    )
+
+    // The marker symbol is always readable — only the lazy properties throw.
+    expect((link as any)[DefineLinkSymbol]).toBe(true)
+  })
+
+  it("throws a descriptive error when a lazy property is read before registration", () => {
+    const link = defineLink(
+      fakeLinkable("store", "store_id", "store"),
+      fakeLinkable("region", "region_id", "region")
+    )
+
+    expect(() => link.entryPoint).toThrow(/not available until the app bootstraps/)
+    expect(() => link.serviceName).toThrow(/entryPoint|serviceName/)
+    expect(() => link.entity as any).toThrow(/not available until the app bootstraps/)
+  })
+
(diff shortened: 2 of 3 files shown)


NicolasGorga (commented):
I see what you mean, but I haven't been able to reproduce as per my comment in the linked issue. Waiting on reproduction steps before continuing

PR #15124, opened 2026-04-16
fix(medusa): register project API middleware before core routes

### LINKED ISSUE #15035: [Bug] Custom middleware on exact route paths (e.g. /admin/products) is silently ignored
### Describe the bug

When registering custom middleware via `defineMiddlewares` in an app's `src/api/middlewares.ts`, middleware with an exact path matcher like `/admin/products` is silently ignored — the middleware function is never called. Using a wildcard matcher like `/admin/*` on the same middleware works correctly.

### To reproduce

1. Create `src/api/middlewares.ts`:

```ts
import { defineMiddlewares } from "@medusajs/framework/http"

function myMiddleware(req, res, next) {
  console.log("MIDDLEWARE HIT")
  next()
}

export default defineMiddlewares({
  routes: [
    {
      matcher: "/admin/products",
      methods: ["POST"],
      middlewares: [myMiddleware],
    },
  ],
})
```

2. Start the dev server and POST to `/admin/products`
3. Observe: "MIDDLEWARE HIT" is never logged
4. Change the matcher to `/admin/*` — middleware now fires

### Observations

- The middleware module IS loaded at startup (top-level `console.log` fires)
- The middleware function is never called regardless of whether `methods` is specified or omitted
- Plugin middleware on the same exact path (e.g. GET `/admin/products`) works fine
- A wildcard matcher (`/admin/*`) works fine as a workaround
- This is **not** an Express issue — `app.post("/admin/products", mw)` followed by `app.post("/admin/products", handler)` works correctly in isolation

### Expected behavior

Middleware registered with `matcher: "/admin/products"` should fire on requests to `/admin/products`, the same way it works for wildcard matchers.

### Environment

- Medusa version: 2.13.1
- Node version: 20.18.2
- OS: macOS

### Workaround

Use `matcher: "/admin/*"` with a path guard inside the middleware:

```ts
if (!req.path.endsWith("/admin/products")) {
  return next()
}
```

Anshul439:
### What

This PR fixes exact-path project API middleware registration order.

Project middleware defined in `src/api/middlewares.ts` with an exact matcher like `/admin/products` is now registered before Medusa core API routes and route middleware.

### Why

Custom project middleware using exact route matchers could be skipped in practice because Medusa core routes were registered before project/plugin API resources.

For example, a project middleware on `POST /admin/products` would not run before Medusa's built-in product validation middleware. Wildcard matchers such as `/admin/*` worked as a workaround because they were sorted ahead of static core routes.

Fixes #15035.

### How

The API loader now scans plugin/project API source paths before Medusa core API paths. Since the project app is represented as the project plugin, its `src/api/middlewares.ts` is collected before core API middleware and routes.

A new HTTP integration regression test boots a fixture Medusa project with `src/api/middlewares.ts`, registers middleware on `POST /admin/products`, and verifies it runs before the built-in product validation middleware.

### Testing

```bash
yarn test:integration:http -- --testPathPattern="routing.spec.ts"
```

---

## Checklist

- [ ] I have added a changeset for this PR
- [x] The changes are covered by relevant tests
- [x] I have fixed the related issues if applicable

---

> [!NOTE]
> **Medium Risk**
> Changes route/middleware registration order, which can alter request handling precedence for existing projects/plugins. Covered by a targeted integration test, but could surface unexpected overrides in custom setups.
> 
> **Overview**
> Fixes API routing precedence so **project/plugin-defined middlewares and routes are registered before Medusa core API routes**, ensuring exact-match middlewares (e.g. `POST /admin/products`) can intercept requests before built-in validators.
> 
> Adds an HTTP integration regression test fixture with a `src/api/middlewares.ts` override and a new `routing.spec.ts` to assert the custom middleware runs first, plus a patch changeset for the Medusa package.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2b1ae44635ba26b094f5b2301f73ecfd63108fc1. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.changeset/cool-lines-attend.md
+++ b/.changeset/cool-lines-attend.md
@@ -1,0 +1,5 @@
+---
+"@medusajs/medusa": patch
+---
+
+fix(medusa): register project API middleware before core routes
--- a/integration-tests/http/__fixtures__/middlewares/medusa-config.js
+++ b/integration-tests/http/__fixtures__/middlewares/medusa-config.js
@@ -1,0 +1,10 @@
+const baseConfig = require("../../medusa-config")
+
+module.exports = {
+  ...baseConfig,
+  projectConfig: {
+    ...baseConfig.projectConfig,
+    databaseUrl: process.env.DB_TEMP_NAME || "postgres://localhost/medusa-test",
+    databaseType: "postgres",
+  },
+}
--- a/integration-tests/http/__fixtures__/middlewares/src/api/middlewares.ts
+++ b/integration-tests/http/__fixtures__/middlewares/src/api/middlewares.ts
@@ -1,0 +1,17 @@
+import { defineMiddlewares } from "@medusajs/framework/http"
+
+export default defineMiddlewares({
+  routes: [
+    {
+      matcher: "/admin/products",
+      methods: ["POST"],
+      middlewares: [
+        (req, res, next) => {
+          // If our bugfix works, this user middleware will hit first and we can intercept
+          // before the built-in validators throw a 400 Bad Request error.
+          res.status(200).json({ custom_middleware_hit: true })
+        },
+      ],
+    },
+  ],
+})
--- a/integration-tests/http/__tests__/middlewares/routing.spec.ts
+++ b/integration-tests/http/__tests__/middlewares/routing.spec.ts
@@ -1,0 +1,33 @@
+import { medusaIntegrationTestRunner } from "@medusajs/test-utils"
+import {
+  adminHeaders,
+  createAdminUser,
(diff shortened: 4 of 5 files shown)


shahednasser:
Thank you for contributing. This isn't a bug, this is by design. Read more in our docs: https://docs.medusajs.com/learn/fundamentals/api-routes/middlewares#overriding-middlewares

The bot reviewer / triager failed to identify that this is by design. I'll update its instructions to make sure this doesn't happen again.

PR #15142, opened 2026-04-18
fix(order): include shipping, tax, and subtotal in order_summary totals

### LINKED ISSUE #15125: [Bug]: order_summary totals missing shipping and tax breakdowns
### What happended?

**Summary**
The order_summary table's totals JSONB column does not include shipping or tax (or their raw_* equivalents), making it impossible to report on these figures using the order summary alone.

### Expected behavior

The totals column should include a full breakdown of all order cost components, including shipping and taxes.

### Actual behavior

No shipping_total or tax_total are present.

### Impact

Any analytics or reporting that reads from order_summary cannot break down order value into its components. We had to remove shipping and tax columns from our finances summary export entirely, and can only derive discounts indirectly as original_order_total - current_order_total.
                                                                                                                                                                                           
### Workaround

None that uses order_summary alone. The order table's raw_shipping_total / raw_tax_total columns appear to have been deprecated in favour of order_summary, but the replacement doesn't carry the same data.

Anshul439:
## Summary

The `order_summary.totals` JSONB column now correctly persists `shipping_total`, `original_shipping_total`, `tax_total`, `original_tax_total`, `subtotal`, `discount_total`, and `discount_tax_total` after order creation.

These fields were computed correctly at runtime by `decorateCartTotals()` but were never forwarded into `getSummary()` before saving the snapshot to the database. This caused silent data loss for merchants using analytics or reporting. Fixes #15125.

- Updated `getSummary()` in `calculate-order-change.ts` to accept the decorated order and read breakdown totals from it
- Updated `getSummaryFromOrder()` the same way for post-change/preview flows
- Updated the call site in `order-module-service.ts` to pass `orderWithTotals` into `getSummary()`
- Added the missing fields to `OrderSummaryDTO` and the internal `OrderSummaryCalculated` type

**Testing** — How have the changes been tested?

Integration test and unit test added. Run:

```bash
cd integration-tests/http && npx jest --testPathPattern="cart/store/cart" --testNamePattern="shipping_total, tax_total, and subtotal in order summary"
cd packages/modules/order && npx jest --testPathPattern="get-summary.spec"
```

---

## Checklist

- [x] I have linked the related issue(s): #15125

---

## Additional Context

No database migration required. The `totals` column is JSONB so it accepts any fields. The data was always computed correctly in memory — it was just never written into the summary object before persisting.


shahednasser (approved):
LGTM thank you for your contribution!

Anshul439:
@shahednasser I've fixed the failing unit test. Not sure if it needs approval again from your side

pepijn-vanvlaanderen:
How would we want to backfill existing order summaries? Or should one create a script for it?

shahednasser (changes requested):
That's a good point @pepijn-vanvlaanderen . @Anshul439 can you add a migration script for existing orders?

Anshul439:
@shahednasser done!

Anshul439:
I’ve added the migration script and tested it locally.
@shahednasser would appreciate a review when you get a chance.

packages/medusa/src/migration-scripts/migrate-order-summary-totals.ts
shahednasser, line 1: Please create a [migration script](https://docs.medusajs.com/learn/fundamentals/data-models/write-migration#data-migration-scripts) under `packages/medusa/src/migration-scripts`

shahednasser (changes requested):

packages/medusa/src/migration-scripts/migrate-order-summary-totals.ts
shahednasser, line 1: A few things about this migration script:

1. It should execute the changes in a workflow, making it possible to revert the changes if an issue occurs. Check `packages/medusa/src/migration-scripts/migrate-normalize-currency-codes-normalization.ts` as an example (make sure you pull the latest changes)
2. To improve the performance, we can filter out deleted rows. For example:

```
FROM "order_shipping_method_adjustment"        
  WHERE deleted_at IS NULL                                              
  GROUP BY shipping_method_id, version
```

3. `line_item_adjustments` handles tax inclusiveness, but not `shipping_adjustments`.
4. The migration script relies on `order_line_item.unit_price`, but the order may be changed and the actual unit price for the latest version will be in `order_item`. We should coalesce between the unit price in the order item and order line item.

shahednasser (changes requested):

NicolasGorga (changes requested):
Hey @Anshul439 could you try the following instead of the pure SQL approach?

- Resolve each version of the different orders
- Call `decorateCartTotals` passing each version
- Update the corresponding versioned order summary entry, updating just the new totals to be persisted, resolved from the last point

If this works, it has the benefit of greatly simplifying the compensation logic you would need to implement in the workflow.

It could make sense, after you test if this is a valid approach, for you to create quite a few dummy orders with a few versions each and run the migration script, monitoring memory and time it takes to execute, to flag any potential performance issues.

Anshul439:
Makes sense, will work on refactoring this using `decorateCartTotals` and workflow. Will push updates soon.

Anshul439:
The CI failure seems to be in `cart.completion.ts` (Failed to acquire lock for key "cart_...") which doesn't look related to the order summary changes , but not 100% sure. Could someone please confirm? @NicolasGorga @shahednasser

packages/medusa/src/migration-scripts/migrate-order-summary-totals.ts
```diff
@@ -0,0 +1,195 @@
...
+        raw_original_shipping_total:
+          existingTotals.raw_original_shipping_total ??
+          (orderWithTotals as any).raw_original_shipping_total,
+        raw_discount_total:
+          existingTotals.raw_discount_total ??
+          (orderWithTotals as any).raw_discount_total,
+        raw_discount_tax_total:
+          existingTotals.raw_discount_tax_total ??
+          (orderWithTotals as any).raw_discount_tax_total,
+      }
```
shahednasser, line 150: This can be simplified by setting the fields (without checking for existingTotals values), then spreading existingTotals after. Fields in existingTotals will then be used if they're set.

shahednasser (changes requested):

pepijn-vanvlaanderen:
Not stale

PR #15150, opened 2026-04-20
fix(loyalty-plugin): remove delete gift cards action + cleanup

shahednasser:
1. Remove delete gift card button from admin. we don't have API route for it.
2. Clean up of loyalty plugin workflows by removing unused imports, unused variables, and fixing type errors where compensation function inputs were cast incorrectly


NicolasGorga (approved):
lgtm

PR #15153, opened 2026-04-21
fix(core-flows): cancel-order credit line ignores pre-existing refunds

langovoi:
## Summary

**What**

Fixes `cancelOrderWorkflow` so the credit line it creates accounts for refunds that already existed on the payment.

**Why**

When an order with partial refunds on captured payments is cancelled, the credit line was issued for the **full** captured amount instead of the remaining (not-yet-refunded) portion, double-counting the prior refund on the order's credit ledger.

Repro: order total \$1272, captured \$1272, already refunded \$500 (via `refundPaymentWorkflow`, which created its own \$500 credit line). Cancel the order — `refundCapturedPaymentsWorkflow` correctly refunds the remaining \$772, but the cancel then creates a credit line for **\$1272** instead of **\$772**. Net: sum of credit lines = \$1772 > the \$1272 ever captured.

**How**

Replaced the post-refund `useQueryGraphStep` on `captures` with one on `refunds`, and compute the credit line as *(total refunds on the affected payments after cancel) − (refunds that existed on them before cancel)*. The pre-cancel value comes from the order snapshot already fetched at the top of the workflow (added `refunds.raw_amount` to that query). This pairs with `refundedPaymentIds` — which only contains payments where refunds actually landed — so partial-success cases stay correct.

**Testing**

Added `should not double-count pre-existing refunds when cancelling an order` to `integration-tests/http/__tests__/order/admin/order-cancel-credit-line.spec.ts`. Against unpatched code it fails with `Expected: 1272, Received: 1772`; after the fix it passes. The two existing tests in that file still pass.

---

## Checklist

---

> [!NOTE]
> **Medium Risk**
> Touches order-cancel refund/credit-line math, which impacts financial/accounting totals and could mis-credit customers if wrong. Change is localized and covered by an added integration test for partial-refund scenarios.
> 
> **Overview**
> Fixes `cancelOrderWorkflow` credit-line calculation so canceling an order only credits the *new* refunds created by the cancel, not refunds that already existed on the payment.
> 
> The workflow now queries `refunds` and computes the credit line as *(refunds after cancel − refunds before cancel)*, and expands the initial order query to include `payment_collections.payments.refunds.raw_amount` to support that subtraction. Adds an integration test ensuring partial refunds made before cancel aren’t double-counted, plus a changeset to ship the patch.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 8ba5736cdf30bf0495768f04cfe6fd3838e5f4d6. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


shahednasser (approved):
@NicolasGorga can you also take a look since you worked on this?

NicolasGorga (approved):
LGTM

PR #15159, opened 2026-04-21
fix(core-flows): fix incorrect stock location picked for item with backorder in a sales channel with multiple locations

### LINKED ISSUE #15132: [Bug]: reserveInventoryStep picks location_ids[0] blindly — cart fails when items are stocked at different locations within one sales channel
### Package.json file

```JSON
{
  "dependencies": {
    "@medusajs/admin-sdk": "2.13.4",
    "@medusajs/cli": "2.13.4",
    "@medusajs/framework": "2.13.4",
    "@medusajs/medusa": "2.13.4"
  }
}
```

> The relevant code is unchanged on `develop` as of [`0929b8d`](https://github.com/medusajs/medusa/blob/0929b8d192833e10b6b91795f1ffc478938c3fb7/packages/core/core-flows/src/cart/steps/reserve-inventory.ts#L84) — `location_id: item.location_ids[0]`.

### Node.js version

v24.14.0 (also reproduced on v20.x)

### Database and its version

PostgreSQL 16 — bug is DB-agnostic (pure JS logic in `core-flows`).

macOS 26.4.1 — bug is OS-agnostic.

### Browser name

N/A — server-side workflow bug surfaced on `POST /store/carts/:id/complete`.

### What happended?

`reserveInventoryStep` picks `item.location_ids[0]` blindly without checking that the chosen location actually has an inventory level for the item. When a cart contains items that are stocked at **different locations within the same sales channel**, the workflow tries to reserve one of the items against a location where it has no level and fails:

```
Item iitem_XXX is not stocked at location sloc_YYY
```

— even though the item is in stock at another location linked to the same sales channel.

This is a different scenario from #10658 / #10661. That fix narrowed the candidate set to locations linked to the cart's sales channel; it did not address how a single location is *picked from* that set. When the candidate set has more than one element (very common — most channels link multiple locations), selection still falls through to `location_ids[0]`.

**Root cause (two locations in code):**

1. [`prepareConfirmInventoryInput`](https://github.com/medusajs/medusa/blob/0929b8d192833e10b6b91795f1ffc478938c3fb7/packages/core/core-flows/src/cart/utils/prepare-confirm-inventory-input.ts#L253-L271) builds a per-item `location_ids` candidate list. When `locationsWithAvailability` is empty, it falls back to **all** channel-linked locations — including locations where the item has no level at all. That fallback is reasonable (e.g. for `allow_backorder` items, or items currently oversold) provided the consumer then picks a location that has a level.

2. [`reserveInventoryStep`](https://github.com/medusajs/medusa/blob/0929b8d192833e10b6b91795f1ffc478938c3fb7/packages/core/core-flows/src/cart/steps/reserve-inventory.ts#L84) just does `location_id: item.location_ids[0]` — no check that a level exists at the chosen location. The picked location ends up depending on the iteration order of a `Set` populated by `deepFlatMap`, which is non-deterministic from the caller's perspective.

### Expected behavior

For each line item, `reserveInventoryStep` should reserve at a location where the inventory item actually has a level — preferring one with sufficient available quantity, falling back to any location with a level (acceptable when `allow_backorder` is true). Two items in the same cart that are stocked at different locations within the same sales channel should each reserve at their respective locations and the cart should complete.

### Actual behavior

Both items try to reserve at the same location (whichever location happens to be `location_ids[0]`). The item that doesn't have a level at that location fails the reservation, the workflow compensates, and the customer sees:

```
POST /store/carts/cart_XXX/complete  →  500
Error setting up the request: Item iitem_XXX is not stocked at location sloc_YYY
```

#### Real-world diagnostic output from a failing cart

Sales channel **Webshop** has two linked stock locations:
- `sloc_…GT5` (Local Vault)
- `sloc_…ST8V0` (CodesWholesale)

Cart contains two managed-inventory items, each stocked at a different location:

```
--- Line item Battlefield 6 (qty=1, manage_inventory=true, allow_backorder=false) ---
  inventory_item iitem_…JS  required=1
    level @ sloc_…ST8V0 (CodesWholesale): stocked=4 reserved=3 avail=1  ← only level
    locationsWithAvailability = [sloc_…ST8V0]
    → reserveInventoryStep picks sloc_…ST8V0  ✓ (works)

--- Line item Minecraft (qty=1, manage_inventory=true, allow_backorder=true) ---
  inventory_item iitem_…WSR  required=1
    level @ sloc_…GT5 (Local Vault): stocked=6 reserved=19 avail=-13  ← only level
    locationsWithAvailability = []  (no level meets required≥1)
    fallback location_ids = [sloc_…GT5, sloc_…ST8V0]  (all channel locations)
    → reserveInventoryStep picks location_ids[0] = sloc_…ST8V0
    ✗ Fails: "Item iitem_…WSR is not stocked at location sloc_…ST8V0"
```

The Minecraft item is `allow_backorder: true` and has a level (with negative availability) at Local Vault, so the reservation should succeed there. Instead it picks CodesWholesale and fails.

### Link to reproduction repo

A separate repo is not necessary — the bug is purely in `@medusajs/core-flows` and reproduces against any Medusa 2.x install with the following setup. (Happy to provide a minimal repo on request.)

**Steps to reproduce on a stock Medusa starter:**

1. Create one sales channel **"Webshop"**.
2. Create two stock locations **A** and **B**. Link both to "Webshop".
3. Create two products with `manage_inventory: true`, both assigned to "Webshop":
   - **P1** — inventory level **only at A** (e.g. `stocked_quantity: 5`).
   - **P2** — inventory level **only at B** (e.g. `stocked_quantity: 5`).
4. As a customer in the Webshop channel, add one P1 + one P2 to a cart, fill in address/shipping/payment.
5. `POST /store/carts/:id/complete`.

**Result:** the request fails with `Item iitem_<P1 or P2> is not stocked at location sloc_<A or B>` (which one depends on iteration order). Whichever item is matched against the wrong location fails the reservation and the entire cart completion is rolled back.

A simpler single-item variant exhibits the same bug: one `allow_backorder: true` product stocked only at A but currently oversold (`reserved > stocked`), in a channel that links both A and B → `locationsWithAvailability = []` → fallback set is `[A, B]` → blind `[0]` may pick B → "not stocked at location B".

### Suggested fix

In `reserveInventoryStep`, before assigning `location_id`, query `inventoryService.listInventoryLevels({ inventory_item_id: […], location_id: […] })` for the candidate `(inventory_item_id, location_id)` pairs and pick the first candidate that actually has a level — preferring one with `available >= required`, falling back to any candidate with a level. One extra batched query per `complete-cart` invocation; tested in production for our store.

Sketch:

```ts
const pickLocationId = (item, perItem /* Map<location_id, level> */) => {
  if (!perItem?.size) return item.location_ids[0]
  const required = MathBN.mult(item.required_quantity, item.quantity)
  for (const id of item.location_ids) {
    const lvl = perItem.get(id)
    if (lvl && MathBN.gte(
      MathBN.sub(lvl.raw_stocked_quantity ?? lvl.stocked_quantity ?? 0,
                 lvl.raw_reserved_quantity ?? lvl.reserved_quantity ?? 0),
      required,
    )) return id
  }
  for (const id of item.location_ids) if (perItem.has(id)) return id
  return item.location_ids[0]
}
```

Alternatively the fix could live in `prepareConfirmInventoryInput` by intersecting the fallback location set with locations that have *any* level for the inventory item.

Happy to open a PR with whichever approach maintainers prefer.

shahednasser:
## Summary

Fix use cases where incorrect stock location is picked, causing an error for unavailable quantity on checkout

Closes #15132

Consider you have:

1. Sales Channel with locations A and B
2. Product A that has inventory in location A
3. Product B that has allowed backorder = true and negative availability in B

Then, as a customer, when you add product A to the cart, then product B to the cart, and choose shipping option from location A, then proceed to checkout, you'll get an error when completing the order that product B doesn't have stock in location A

This occurred because:

1. When retrieving the item's stock and inventory details, we try to retrieve the locations where the item has quantity greater than the required quantity. However, we don't take into account that the item may have allowed_backorder = true
2. If there is no location that has availability greater than the required quantity, we return all locations (including locations where other items in the cart has levels, but not this item)
3. When confirming the inventory, we look at the first location only. In the case where we return all locations, the item may not have levels at the chosen location, resulting in an error (even though allowed backorder is enabled) 

In this fix, if an item doesn't have available quantity, we only return the locations where the item has levels in. That way, when confirming the quantity availability, we're not comparing against a location where the item doesn't have levels in.

Added necessary test for this use case. Tested locally as well

---

## Checklist

### DIFF
--- a/.changeset/silly-worms-stand.md
+++ b/.changeset/silly-worms-stand.md
@@ -1,0 +1,5 @@
+---
+"@medusajs/core-flows": patch
+---
+
+fix(core-flows): fix incorrect stock location picked for item with backorder in a sales channel with multiple locations
--- a/packages/core/core-flows/src/cart/utils/__tests__/prepare-confirm-inventory-input.spec.ts
+++ b/packages/core/core-flows/src/cart/utils/__tests__/prepare-confirm-inventory-input.spec.ts
@@ -463,6 +463,103 @@ describe("prepareConfirmInventoryInput", () => {
     })
   })
 
+  it("should only return locations where the inventory item has a level when falling back (backorder item stocked at a different location than another item in the same channel)", () => {
+    const input = {
+      sales_channel_id: "sc_1",
+      variants: [
+        {
+          id: "pv_1",
+          manage_inventory: true,
+          allow_backorder: false,
+          inventory_items: [
+            {
+              inventory_item_id: "ii_1",
+              variant_id: "pv_1",
+              required_quantity: 1,
+              inventory: [
+                {
+                  location_levels: {
+                    stocked_quantity: 5,
+                    reserved_quantity: 0, // available=5 >= 1
+                    location_id: "sl_A",
+                    stock_locations: [
+                      {
+                        id: "sl_A",
+                        sales_channels: [{ id: "sc_1" }],
+                      },
+                    ],
+                  },
+                },
+              ],
+            },
+          ],
+        },
+        {
+          id: "pv_2",
+          manage_inventory: true,
+          allow_backorder: true,
+          inventory_items: [
+            {
+              inventory_item_id: "ii_2",
+              variant_id: "pv_2",
+              required_quantity: 1,
+              inventory: [
+                {
+                  location_levels: {
+                    stocked_quantity: 6,
+                    reserved_quantity: 19, // available=-13 < 1: no availability
+                    location_id: "sl_B",
+                    stock_locations: [
+                      {
+                        id: "sl_B",
+                        sales_channels: [{ id: "sc_1" }],
+                      },
+                    ],
+                  },
+                },
+              ],
+            },
+          ],
+        },
+      ],
+      items: [
+        {
+          variant_id: "pv_1",
+          quantity: 1,
+          id: "item_1",
+        },
+        {
+          variant_id: "pv_2",
+          quantity: 1,
+          id: "item_2",
+        },
+      ],
+    }
+
+    const result = prepareConfirmInventoryInput({ input })
+
+    expect(result).toEqual({
+      items: [
+        {
+          id: "item_1",
+          inventory_item_id: "ii_1",
+          required_quantity: 1,
+          quantity: 1,
+          allow_backorder: false,
+          location_ids: ["sl_A"], // has availability at sl_A
+        },
+        {
+          id: "item_2",
+          inventory_item_id: "ii_2",
+          required_quantity: 1,
+          quantity: 1,
+          allow_backorder: true,
+          location_ids: ["sl_B"], // no availability, but only has a level at sl_B — must not pick sl_A
+        },
+      ],
+    })
+  })
+
   it("should throw an error if any variant has no inventory items", () => {
     const input = {
       sales_channel_id: "sc_1",
--- a/packages/core/core-flows/src/cart/utils/prepare-confirm-inventory-input.ts
+++ b/packages/core/core-flows/src/cart/utils/prepare-confirm-inventory-input.ts
@@ -259,6 +259,12 @@ const formatInventoryInput = ({
         )
       )
 
+      const locationsWithLevel = location_ids.filter((locId) =>
+        stockAvailability
+          .get(locId)
+          ?.has(variantInventoryItem.inventory_item_id)
+      )
+
       itemsToConfirm.push({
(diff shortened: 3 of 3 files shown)


NicolasGorga (approved):
LGTM

